PR #2, opened 2016-01-17
Add a configuration file so we can use Travis CI

chrisseaton:
I'll need to also connect the repo to Travis after this PR is merged before it starts being run.


smarr:
Should this also include `checkheaders`, `checkcopyright`, and `checkstyle`?

However, I think `checkheaders` does not work currently.

PR #3, opened 2016-01-19
Fix pylint install for travis.

chumer:


.travis.yml
```diff
@@ -1,6 +1,12 @@
 language: java
+python:
+  - "2.7"
 jdk:
   - oraclejdk8
```
jtulach: I don't know what this line means, but I remind you that Truffle has to build with JDK7.

PR #6, opened 2016-01-20
Test case for a potential bug in the polyglot engine.

smarr:
This test case fails to access the global object of SL.
It raises `IllegalStateException("Accessor.findLanguage access to vm")`, while it probably shouldn't.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 ## Introduction
 
 Truffle is a framework for implementing languages as simple interpreters.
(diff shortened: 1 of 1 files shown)


jtulach:
Perfect, I'll fix this by Tuesday next week.

jtulach:
Here is the fix: https://github.com/graalvm/truffle/commit/a5258c3736edd36d395df96f836df095eb72e370

smarr:
Thanks!

PR #7, opened 2016-01-21
README: github.com/OracleLabs/GraalVM is deprecated.

zapster:
We could also link to http://openjdk.java.net/projects/graal/ if that is preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 ## Introduction
 
 Truffle is a framework for implementing languages as simple interpreters.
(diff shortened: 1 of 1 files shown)


smarr:
Most user friendly might be a link to https://wiki.openjdk.java.net/display/Graal/Instructions

woess:
+1 for https://wiki.openjdk.java.net/display/Graal/Instructions

zapster:
Ok will change it to https://wiki.openjdk.java.net/display/Graal/Instructions!

PR #9, opened 2016-01-21
Potential bug, can't call getEnv() before doing an eval()

smarr:
This is a test case, for something that looks like an issue with initialization order.
The thread-local field for the VM/polyglot engine does not yet seem to be set when calling getEnv(.).


jtulach:
Why it should be possible to call getEnv? It is not a public API.

PR #10, opened 2016-01-21
Enable passing of arguments to the polyglot engine

smarr:
This change allows to pass an array of strings to the polyglot engine and retrieve it in the environment object.

The goal is to provide a straight-forward way to pass implementation-level arguments, as the typically come from the command line to the `PolyglotEngine` and the languages.

At the moment, we need to jump through many many hoops.
To realize the same effect, in a language-specific way, it seems we need to do roughly the following:

1. we need some wrapper class to be able to get to the value later on in the language:

  ```Java
  public static class CmdArgWrapper<T> implements TruffleObject {
    private final T obj;
    public CmdArgWrapper(final T obj) {
      this.obj = obj;
    }

    public T getObject() {
      return obj;
    }

    @Override
    public ForeignAccess getForeignAccess() {
      throw new UnsupportedOperationException();
    }
  }
  ```

2. we need to pass the arguments to the engine. In the case of SOM, I need these arguments when creating the language context, because the context contains the object system, which depends on the source to be used. And, might also have instruments that need to be configured from the very start.

  ```Java
  String[] args = new String[] {"--kernel", "Kernel.som", "--instrument", "dyn-metrics"};
  Builder builder = PolyglotEngine.newBuilder();
  builder.globalSymbol("CMD_ARGS", new CmdArgWrapper<String[]>(args));
  PolyglotEngine vm = builder.build();
  ```

  I believe, the whole notion of global TruffleObjects does not apply here because these arguments are strictly at the implementation level and not language-level objects. But even if there are arguments that become visible to the language, it is the languages job to figure that out and do the necessary wrapping.

3. in the `TruffleLanguage.createContext(...)` method, I need to do this:

```Java
protected VM createContext(final Env env) {
    Object argsObj = env.importSymbol("CMD_ARGS");
    String[] args = null;
    if (argsObj instanceof CmdArgWrapper<?>) {
      CmdArgWrapper<String[]> argsWrapper = (CmdArgWrapper<String[]>) argsObj;
      args = argsWrapper.getObject();
    }
    assert args != null;
  // ...
```

All this is non-trivial for a very basic use case that any practical language would want to support: take some parameters to control its initialization process/startup.

With the provided patch, all this shrinks to:

```Java
String[] args = new String[] {"--kernel", "Kernel.som", "--instrument", "dyn-metrics"};
Builder builder = PolyglotEngine.newBuilder();
builder.setArguments(args);
PolyglotEngine vm = builder.build();

protected VM createContext(final Env env) {
    String[] args = env.getArguments();
// ...
```

I hope this is compelling enough for inclusion. I know the API should be minimal, but this is a super basic use case, and while there might be static-state solutions that are shorter than this, I believe we should provide an API that allows users to do 'the right thing' easily.

I am not entirely sure about my patch thought. The whole code based involved here is extremely complex and rather hard to comprehend. I mean, it adds about 40 lines of code for something very very trivial: another field + getter. I did not expect that much change for this simple feature.

### DIFF
--- a/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
+++ b/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
@@ -25,14 +25,18 @@ package com.oracle.truffle.api.vm;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.List;
 
 import org.junit.Test;
 
 import com.oracle.truffle.api.TruffleLanguage.Env;
 import com.oracle.truffle.api.source.Source;
+import com.oracle.truffle.api.vm.ImplicitExplicitExportTest.Ctx;
 
 public class EngineTest {
     protected PolyglotEngine.Builder createBuilder() {
@@ -120,17 +124,41 @@ public class EngineTest {
     }
 
     @Test
-    public void initializePolyglotEngineWithArguments() throws IOException {
-        PolyglotEngine vm = createBuilder().config("application/x-test-import-export-1", "cmd-line-args", new String[]{"1", "2"}).build();
+    public void initializePolyglotEngineWithConfiguration() throws IOException {
+        PolyglotEngine vm = createBuilder().config("application/x-test-import-export-1", "cmd-line-args", new String[]{"1", "2"}).config("application/x-test-import-export-2", "hello", "world").build();
         PolyglotEngine.Language language1 = vm.getLanguages().get("application/x-test-import-export-1");
 
-        // TODO: remove once initialization issue is solved for
-        // https://github.com/graalvm/truffle/pull/9
-        language1.eval(Source.fromText("return=arr", "get the array")).as(AccessArray.class);
+        assertNotNull("Lang1 found", language1);
 
-        Env env = language1.getEnv(true);
-        String[] args = (String[]) env.getConfig().get("application/x-test-import-export-1").get("cmd-line-args");
+        Ctx ctx1 = language1.getGlobalObject().as(Ctx.class);
+        String[] args = (String[]) ctx1.env.getConfig().get("application/x-test-import-export-1").get("cmd-line-args");
+        assertNotNull("Founds args", args);
+
         assertEquals("1", args[0]);
         assertEquals("2", args[1]);
+
+        assertNull("Can't see settings for other language", ctx1.env.getConfig().get("application/x-test-import-export-2"));
+
+        // make sure configuration is read-only
+        try {
+            ctx1.env.getConfig().get("application/x-test-import-export-1").put("hi", "there!");
+            fail("The map should be readonly");
+        } catch (UnsupportedOperationException ex) {
+            // OK
+        }
+
+        try {
+            ctx1.env.getConfig().put("new-mime-type", new HashMap<String, Object>());
+            fail("The map should be readonly");
+        } catch (UnsupportedOperationException ex) {
+            // OK
+        }
+
+        PolyglotEngine.Language language2 = vm.getLanguages().get("application/x-test-import-export-2");
+        assertNotNull("Lang2 found", language2);
+
+        Ctx ctx2 = language2.getGlobalObject().as(Ctx.class);
+        assertEquals("world", ctx2.env.getConfig().get("application/x-test-import-export-2").get("hello"));
+        assertNull("Cannot find args", ctx2.env.getConfig().get("cmd-line-args"));
     }
 }
--- a/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
+++ b/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
@@ -124,7 +124,6 @@ public class EngineTest {
         assertEquals(3, arr[2].intValue());
     }
 
-
     @Test
     public void engineConfigBasicAccess() throws IOException {
         Builder builder = createBuilder();
@@ -228,6 +227,5 @@ public class EngineTest {
         Ctx ctx2 = language1.getGlobalObject().as(Ctx.class);
         assertNull(ctx2.env.getConfig().get("hello"));
     }
-
 
 }
--- a/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
+++ b/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
@@ -121,7 +121,7 @@ public class EngineTest {
 
     @Test
     public void initializePolyglotEngineWithArguments() throws IOException {
-        PolyglotEngine vm = createBuilder().setArguments("application/x-test-import-export-1", new String[]{"1", "2"}).build();
+        PolyglotEngine vm = createBuilder().config("application/x-test-import-export-1", "cmd-line-args", new String[]{"1", "2"}).build();
         PolyglotEngine.Language language1 = vm.getLanguages().get("application/x-test-import-export-1");
 
         // TODO: remove once initialization issue is solved for
@@ -129,7 +129,7 @@ public class EngineTest {
         language1.eval(Source.fromText("return=arr", "get the array")).as(AccessArray.class);
 
         Env env = language1.getEnv(true);
-        String[] args = (String[]) env.getArguments().get("application/x-test-import-export-1");
+        String[] args = (String[]) env.getConfig().get("application/x-test-import-export-1").get("cmd-line-args");
         assertEquals("1", args[0]);
         assertEquals("2", args[1]);
     }
--- a/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
+++ b/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
@@ -31,6 +31,7 @@ import java.util.List;
 
 import org.junit.Test;
 
+import com.oracle.truffle.api.TruffleLanguage.Env;
 import com.oracle.truffle.api.source.Source;
 
 public class EngineTest {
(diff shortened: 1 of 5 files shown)


eregon:
Is Env per-language?
If so, we just need a way to add the arguments to the proper Env I think.

smarr:
The `env` object references a specific language yes. But the arguments API I added does not take that into account currently. I'll draft something that allows one to specify the language the arguments should belong to, as well.

smarr:
Ehm, I don't see how that could be working. To be able to construct a global object, I need to have the language initialized, no?

So, I would propose:

```Java
Builder builder = PolyglotEngine.newBuilder();
builder.setArguments(Language1.class, args1);
builder.setArguments(Language2.class, args2);
```

And then, the following would provide me with the correct `argsN`:
```Java
protected VM createContext(final Env env) {
    String[] args = env.getArguments();
```

I need the arguments in `createContext(.)`, and I don't want to hack around it.

smarr:
This is now implemented in the last commit.

truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
```diff
@@ -117,4 +119,19 @@ public void wrappedAsArray() throws Exception {
         assertEquals(2, arr[1].intValue());
         assertEquals(3, arr[2].intValue());
     }
+
+    @Test
+    public void initializePolyglotEngineWithArguments() throws IOException {
+        PolyglotEngine vm = createBuilder().setArguments(ExportImportLanguage1.class, new String[]{"1", "2"}).build();
```
jtulach: Alas, I don't think people should reference the implementation class of a language. They certainly don't have to do so now. Probably the language should be identified by a MIME type.

smarr: Ok, used the MIME type. What's the reason for this? Potential class loading issues or something? Intuitively, I would assume that this kind of interaction still happens in tightly bound places, like a main method or something similar. You envision a more dynamic and modular use case?

woess:
Isn't the actual problem here that you can't get global symbol values back out without wrapper?

smarr:
I think there are two separate concepts here:

1. Such arguments/parameters as I propose to introduce are on the implementation level, this is data a language runtime might need for initialization.

2. Global Symbols/Values are language-level objects that are used by an application. As such, I agree with the current design. Actually, I even consider the whole notion of unboxing on those value problematic (for primitives there is a use cases yes). Generally, this is something a specific language knows how to handle and deal with. This is for applications.

So, I would argue we need this second way for implementation-level purpose only to pass data to a language implementation. Unifying these two ways does not seem the right solution for me, because this weakens the nice semantic properties we got for interop.

woess:
@smarr: Good point. I was thinking that perhaps one might want to pass something else than a `String[]` which is already possible with a global symbol; but I agree it's not the right tool.

smarr:
hm, yeah, I was only briefly thinking of other things than strings. Perhaps it should be just `Object` instead of String[]? My current use case is really typical command line arguments. But, other languages also need access to environment variables from the host system. So, perhaps this should be more general.

jtulach:
Right, this is the kind of discussion that lead me to believe there shouldn't be any generic setArguments. The languages may have different needs and it should be up to them to decide what data and in what format they want.

I've created https://github.com/graalvm/truffle/pull/11 to demonstrate how (simple) it feel to access array of parameters in current API.

smarr:
Ok, well, so there is at least one lesson we should learn from this. The current API is absolutely not discoverable, and the implementation is way to involved to get any idea of intentions. I am really suffering over here :(

In your version #11, making the jump from `TruffleObject` to `JavaInterop.asJavaObject(.)` is just something that is not discoverable. Not sure how much documentation is going to do about that.

jtulach:
I have a bug in codesnippets that I will fix on Monday. Then we will see...

Right. The trick isn't easy to discover and that's the reason it took me so long and I had to test it first.

smarr:
Taking Andreas' comment into account, I generalized the String[] to an Object. Benefits of this vs. the Java Interop based approach:
 - does not require reflection
 - is a separate mechanism for implementation-level data
 - might be easer to find for new users

jtulach:
Object == the JVM style of typing things. I don't like it.

I've closed the #10 request, because I realized that in this case we don't want to wrap Java types into TruffleObject. Thus we need different method. Rather than setArguments, I'd however prefer:

Builder Builder.config(String mimeType, String key, Object value);
Map<String,Object> Env.getConfig();

I think this would work for your usecase as well and gives languages a chance to expose just names, not real objects.

smarr:
@jtulach I added the change to have the API structure like this:
```Java
Builder Builder.config(String mimeType, String key, Object value);
Map<String,Object> Env.getConfig(); // can be called from createContext(), value is not wrapped
```

At the other languages (@woess, @ansalond, @eregon, @chrisseaton, @lukasstadler, @chumer), does this look useful/sufficient to you?

jtulach:
We should improve our workflow. I've just added implementation of my request too: https://github.com/jtulach/truffle/commit/32d4240880fd2a97100c3eb1776e274d7c8bbfbb

truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
```diff
@@ -117,4 +118,19 @@ public void wrappedAsArray() throws Exception {
...
+    @Test
+    public void initializePolyglotEngineWithArguments() throws IOException {
+        PolyglotEngine vm = createBuilder().setArguments("application/x-test-import-export-1", new String[]{"1", "2"}).build();
+        PolyglotEngine.Language language1 = vm.getLanguages().get("application/x-test-import-export-1");
+
+        // TODO: remove once initialization issue is solved for
+        // https://github.com/graalvm/truffle/pull/9
+        language1.eval(Source.fromText("return=arr", "get the array")).as(AccessArray.class);
+
+        Env env = language1.getEnv(true);
```
jtulach: Isn't this broken? I am getting exceptions from here. I suggest to use my version of the test - it uses just public methods.

smarr:
Hm, ok, since I started the pull request, I assumed it to be my responsibility to fix the issues.

truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
smarr: ok, will change the test. Do you have any strong opinions on the Map<> vs. List<> difference between my and your implementation?

truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/EngineTest.java
```diff
@@ -117,4 +121,43 @@ public void wrappedAsArray() throws Exception {
...
+        assertNotNull("Lang1 found", language1);
+
+        Ctx ctx1 = language1.getGlobalObject().as(Ctx.class);
+        String[] args = (String[]) ctx1.env.getConfig().get("application/x-test-import-export-1").get("cmd-line-args");
+        assertNotNull("Founds args", args);
+
+        assertEquals("1", args[0]);
+        assertEquals("2", args[1]);
+
+        assertNull("Can't see settings for other language", ctx1.env.getConfig().get("application/x-test-import-export-2"));
```
jtulach: I don't think there is any use for two levels of the map. Just merge values for all mimetypes of a language together into one map.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
```diff
@@ -437,6 +439,13 @@ public OutputStream err() {
         public Instrumenter instrumenter() {
             return instrumenter;
         }
+
+        /**
+         * @return arguments used to create the polyglot engine
+         */
+        public Map<String, Map<String, Object>> getConfig() {
+            return config;
```
jtulach: This should return Map<String,Object>. Javadoc should connect this method with the config one. See
https://github.com/jtulach/truffle/blob/32d4240880fd2a97100c3eb1776e274d7c8bbfbb/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java#L445

truffle/com.oracle.truffle.api.vm/src/com/oracle/truffle/api/vm/PolyglotEngine.java
```diff
@@ -265,6 +270,29 @@ public Builder onEvent(EventConsumer<?> handler) {
         }
 
         /**
+         * Provide configuration data to initialize the {@link PolyglotEngine} for a specific
+         * language. These arguments can be used by languages to initialize and configure their
+         * initial execution state.
```
jtulach: It may be good idea to connect the method to the getConfig one. See
https://github.com/jtulach/truffle/blob/32d4240880fd2a97100c3eb1776e274d7c8bbfbb/truffle/com.oracle.truffle.api.vm/src/com/oracle/truffle/api/vm/PolyglotEngine.java#L263

smarr:
For documentation: some discussion on the relationship between MIME types and config options: https://github.com/jtulach/truffle/commit/32d4240880fd2a97100c3eb1776e274d7c8bbfbb#commitcomment-15653479

smarr:
Ok, the latest change merges the per-mime-type options to per-language options from the view of the languages.

Anything else left to do before a merge can be done? (at some point, I'll rebase the change set again on master)

jtulach:
If you could use the codesnippet tag to enrich Javadoc of the two new methods with real sample, it would be great. Then I am ready to merge.

smarr:
Hm, perhaps this is me being lazy, but code examples for getter/setter methods? I mean, the javadoc references the two elements that belong together, which is very helpful. Don't know but code example feel too much here.

jtulach:
OK. I'll do it myself and merge.

jtulach:
Your solution with Map<String,Map> doesn't guarantee order. It is not really predictable which key wins if there is a conflict.

smarr:
Silent merging is just a source of bugs. I think, I would prefer an exception on conflict.

jtulach:
Builder pattern style is to use the latest value provided when build method is called. I'll follow that pattern in this case as well.

smarr:
From my point of view, the problem is that this is not related to the builder.
In the builder we do not even know the connection between mime types and language implementations.

The problem is that we at some point start 'interpreting' a given configuration and do something funky with it. Something where apparently everyone has different expectation. I asked Lukas, and he had another surprising expectation about what should happen. He suggested that it should be merged based on the order in which the mime types are specified for the language.

So, well, I guess there are other options as well.
That's why I initially kept the configuration for different mime types apart.

jtulach:
Yes, Lukas's line of thinking is possible as well. However it shows that Lukas is a language writer, not (pure) user. The user of PolyglotEngine has no way of knowing what is the order of MIME types in a TruffleLanguage.Registration. For the user the only meaningful order is the order he used. I've clarified that in:
https://github.com/jtulach/truffle/commit/d5f92ae7860a2d0abc7498885bc2475736b271f7

jtulach:
Here is the state I am OK with: https://github.com/jtulach/truffle/commit/f4756bf491a874f14c8ef47622617bd48ad5b8af - can you pull the changeset and push it into your repository so we can reuse this pull request or shall I start a new one?

PR #11, opened 2016-01-22
Showing the current approach to pass in language arguments

jtulach:
In contrast to https://github.com/graalvm/truffle/pull/10 this kind of arguments access doesn't require any API changes right now and it seems to be straightforward enough (once documented) in Javadoc.

### DIFF
--- a/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/ImplicitExplicitExportTest.java
+++ b/truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/ImplicitExplicitExportTest.java
@@ -48,9 +48,13 @@ import com.oracle.truffle.api.TruffleLanguage.Env;
 import com.oracle.truffle.api.frame.MaterializedFrame;
 import com.oracle.truffle.api.instrument.Visualizer;
 import com.oracle.truffle.api.instrument.WrapperNode;
+import com.oracle.truffle.api.interop.TruffleObject;
+import com.oracle.truffle.api.interop.java.JavaInterop;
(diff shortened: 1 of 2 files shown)


truffle/com.oracle.truffle.api.test/src/com/oracle/truffle/api/vm/ImplicitExplicitExportTest.java
```diff
@@ -270,6 +273,27 @@ public ExportImportLanguage1() {
...
         @Override
+        protected Ctx createContext(Env env) {
+            // BEGIN: arguments.read
+            Object args = env.importSymbol("args");
+            List<?> argsList = null;
+            if (args instanceof TruffleObject) {
+                // convert arguments to list
+                argsList = JavaInterop.asJavaObject(
+                    List.class, (TruffleObject) args
+                );
```
eregon: Why does the String[] magically becomes a List?
Also, it seems not so practical to use, looking at the lines below (get, cast, etc).

woess:
Alas, one of our requirements for basic features like this (argument passing) is that they work also without JavaInterop.

jtulach:
Yeah, I know. We have to be able to run without reflection to achieve reasonable speed. I'll think about ways to avoid it, because in this case it isn't needed - the type (List) is well known in advance.

jtulach:
Looks like there is a difference in the usage of globalSymbol vs. this kind of parameter. In case of global symbol, we want to wrap Java objects into Truffle. In this situation, we don't want to wrap anything. It probably doesn't make sense to try to use the same API for these two situations.

PR #13, opened 2016-01-25
Adding check of javadoc broken links into travis pull request run

jtulach:
Run more tests to ensure documentation doesn't have broken links and missing tags.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,6 @@ env:
   - TEST_COMMAND='mx/mx checkoverlap'
   - TEST_COMMAND='mx/mx canonicalizeprojects'
   - TEST_COMMAND='mx/mx checkstyle'
-  
+  - TEST_COMMAND='mx/mx javadoc --unified'  
 
 


chumer:
I tried this before, but failed. Unfortunately mx javadoc requires to have "mx build" run before. In my opinion mx javadoc should just do the build automatically if not yet built. What do you think @gilles-duboscq ?

chrisseaton:
Or you can have multiple commands as in `mx/mx build && mx/mx javadoc --unified`, or call a shell script with many commands in it.

chumer:
`mx/mx build && mx/mx javadoc --unified` does not work on travis. don't know the reason. stopped investigating

jtulach:
In the light of #15 this change is obsolete.

gilles-duboscq:
Regarding javadoc: `mx javadoc` already does the build by default. What was the problem?

smarr:
The build/gate did not fail when I introduced imperfect JavaDoc.

PR #15, opened 2016-01-25
Making 'mx javadoc --unified' part of the gate

jtulach:
Calling mx.javadoc(...) from the gate.


smarr:
I fix the underlying JavaDoc issue and pushed/merged the change. So, hopefully, this will pass now.

PR #16, opened 2016-01-25
minor javadoc cleanup and access modifier fixes

woess:
just some javadoc cleanup and object model access modifier fixes in the implementation package.


truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/source/Source.java
```diff
@@ -411,8 +411,8 @@ private Source() {
 
     /**
      * Returns a short version of the name of the resource holding a guest language program (as
-     * described in @getName). For example, this could be just the name of the file, rather than a
```
woess: @jtulach: how come `@getName` is not a javadoc error?

smarr:
Thanks for the visibility changes.
And, looks good to me.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/source/Source.java
jtulach: I don't know.

PR #17, opened 2016-01-25
Add Env#getSupportedMimeTypes()

chrisseaton:
I want to know what languages are supported in a `PolyglotEngine`. The pressing use-case is in JRuby+Truffle where I want to know if I have a JS interpreter available. I'm currently doing this by just trying to eval with `application/javascript` and catching any error as an indication that it isn't available. This patch instead checks what languages are available first.

https://github.com/jruby/jruby/compare/truffle-execjs...truffle-execjs-list-languages#diff-0c9caf96978e9602deec3f1fca2d97a8L68

The workaround before this patch is slow, even if I end up not using the JavaScript interpreter after finding that it is available. This is 3rd party code for the most part so I can't redesign much.

This is a new API, so a 'large' review. @jtulach @chumer @woess

### DIFF
--- a/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
+++ b/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
@@ -388,11 +388,11 @@ public abstract class TruffleLanguage<C> {
         }
 
         /**
-         * Allows it to be determined if this {@link PolyglotEngine} can execute code written in a
-         * language with a given MIME type.
+         * Allows it to be determined if this {@link com.oracle.truffle.api.vm.PolyglotEngine} can
+         * execute code written in a language with a given MIME type.
          *
-         * @see Source#withMimeType()
-         * @see #parse()
+         * @see Source#withMimeType(String)
+         * @see #parse(Source, String...)
          *
          * @return a boolean that indicates if the MIME type is supported
          */
--- a/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
+++ b/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
@@ -232,8 +232,8 @@ public abstract class TruffleLanguage<C> {
     protected abstract Visualizer getVisualizer();
 
     /**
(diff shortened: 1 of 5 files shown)


jtulach:
The test coverage seems to be zero. Please increase it a bit.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
```diff
@@ -386,6 +386,19 @@ public Object importSymbol(String globalName) {
...
+         * Produces an array of strings for the MIME types supported by this {@link PolyglotEngine}.
+         * There may be multiple MIME types for a single language. Each MIME type will appear only
+         * once, even if multiple language declare it.
+         *
+         * MIME types from this array can be used in methods such as {@link #parse}.
+         *
+         * @return a non-null array of a string for each supported MIME type
+         */
+        public String[] getSupportedMimeTypes() {
+            return API.getSupportedMimeTypes(vm);
```
jtulach: As far as I understand your use-case a method like
```
public boolean isMimeTypeSupported(String mimeType);
```
would work for you as well. If so, I prefer it to the array of all mime types.

chrisseaton:
I switched from `getSupportedMimeTypes` to `isMimeTypeSupported`.

Where is the best place to test this? I couldn't see a collection of `Env` tests. Do I need to set up a dummy language of my own to get at `Env`?

smarr:
I added a test to `EngineTest` class. But be careful not to use internal APIs in the tests (which are available in that package). @jtulach does not like that :)

jtulach:
There are few testing languages along EngineTest. Feel free to improve them for testing the mime types.

jtulach:
Nice test - no hacks, close to the way people will use this API. Once the formatting is fixed, please merge. I'll work on tagging the release then.

chrisseaton:
I'm not sure how to fix these javadoc warnings. When I run `mx javadoc` locally I get tons of other warnings, and Eclipse doesn't have a problem with the classes and methods I've referenced.

Am I missing something?

https://travis-ci.org/graalvm/truffle/jobs/105453975

jtulach:
The right command is:

mx javadoc

PR #21, opened 2016-01-28
better message for UnsupportedSpecializationException: include data types

lukasstadler:
the actual data types that weren't successfully specialized on are not always visible from the "toString()" of the values - this change adds a list of type names.


chumer:
Looks good.

chumer:
You can assign me for the change the next time so I get notified.

PR #23, opened 2016-01-29
Add capability to reload Source objects.

chumer:
I needed the capability to reload sources. Especially with url caching this can be hacky. So I introduced a new method reload in Source that keeps the original source unmodified but reloads the code from the URL or File.

### DIFF
--- a/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/source/Source.java
+++ b/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/source/Source.java
@@ -399,6 +399,11 @@ public abstract class Source {
     private String mimeType;
     private TextMap textMap;
 
+    /**
+     * Reloads a source by creating a new instance. The original source instance remains unchanged.
+     */
+    public abstract Source reload() throws IOException;
+
     abstract void reset();
 
     /**
@@ -712,6 +717,13 @@ public abstract class Source {
         }
 
         @Override
+        public Source reload() {
+            AppendableLiteralSource literals = new AppendableLiteralSource(description);
+            literals.codeList.addAll(codeList);
+            return literals;
+        }
+
+        @Override
         public String getName() {
             return description;
         }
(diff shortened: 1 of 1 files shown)


mlvdv:
What's the use case?

woess:
Seems a bit strange to me. Wouldn't you rather want an uncached URL/File Source? You could then simply create a new one when you need it up-to-date.

jtulach:
There was d9294237e1518b8fab5c7668945967bbbf0b2df5 and I got a feeling that Source is supposed to be immutable. That seems to be OK in this proposal, but the general motivation for the change is unclear.

chumer:
I need the in R to auto reload source files on changes. There is no way to reload a source from with a given URL atm. Its always returning the same Source object which caches the code as field and never reloads its contents.

We would not need this change if URL source would not cache their instances globally. I am happy to remove the cache instead.

@mlvdv I think you introduced the URL source cache. Do you remember its rationale?

mlvdv:
No, I'm not sure what the conversation was about that.  The original Source object wasn't much more than a collection of use cases in the different languages, since every language handled sources differently.

woess:
I think we should remove the current (static!) cache from Source. Perhaps we could split it out into a separate API that builds on top of Source that languages can optionally use if they need caching (e.g. "SourceCache"), perhaps tied to a PolyglotEngine instance?

jtulach:
OK. I prefer to remove the cache rather than trying to introduce an API for reload.

jtulach:
How such change affects Source.equals then? Right now it assumes that if Source comes from the same File/URL it is the same one. That will no longer be true - the equals will have to be modified to really compare the content.

chumer:
I don't think Source#equals should compare the code for File sources or URL sources as they are mutable. For literal sources it might be a good idea to compare the sources itself on equals.

chumer:
Closing. Will open another one where I remove the caching capability.

jtulach:
So what should Source#equals compare? Identity?

PR #24, opened 2016-01-29
Small convenience improvements for Instrumentation API

smarr:


truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/source/SourceSection.java
```diff
@@ -229,12 +231,17 @@ public String getShortDescription() {
...
         if (source == null) {
-            return kind + ": " + identifier;
+            result = kind + ": " + identifier;
         } else {
-            return "source=" + source.getShortName() + " pos=" + charIndex + " len=" + charLength + " line=" + startLine + " col=" + startColumn +
+            result = "source=" + source.getShortName() + " pos=" + charIndex + " len=" + charLength + " line=" + startLine + " col=" + startColumn +
                             (identifier != null ? " identifier=" + identifier : "") + " code=" + getCode();
         }
+        if (tags != null && tags.length > 0) {
+            result = " tags: " + Arrays.toString(tags);
```
chumer: this should be a += ,right?

smarr: How did that ever work? hm, probably it didn't. yes, you are right. will fix.

smarr:
I think this needs more discussion before merging.
Points raised elsewhere:
 - should tags be kept here as a mutable field?
 - should source sections be unique?

For the last point, this is something I wonder, but doesn't seem to make a lot of sense. Unique source sections, i.e., only one object and one set of tags for a give range. This seems problematic because it might be language dependent or specific to parser/AST structure. Also overlapping source sections are easily constructed, also for identical range. Would be hard to guarantee uniqueness.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/source/SourceSection.java
```diff
@@ -292,6 +299,16 @@ public boolean equals(Object obj) {
     }
 
     /**
```
jtulach: Missing Javadoc.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/source/SourceSection.java
```diff
@@ -292,6 +299,16 @@ public boolean equals(Object obj) {
...
+     * @param newTags is a new set of tags
+     * @return copy with the provided tags set
+     */
+    public SourceSection cloneWithTags(String... newTags) {
```
jtulach: withTags(...) is more consistent with naming used in this package (see Source#withMimeType).

chumer:
> should tags be kept here as a mutable field?
No. SourceSection must be immutable. It is used in HashMaps as key a lot. Since nodes are mutable already its not a big deal if Node#sourceSection is mutable instead.

> should source sections be unique?
Unique for a particular Source, or globally unique? I don't think we need that. A proper Equality check is good enough I believe.

mlvdv:
@chumer you mentioned using SourceSection in HashMaps, but haven't responded to my question in the instrumentation review about how tags affect equality.  The current overrides of `hashcode()` and `equals()` ignore tags.

smarr:
For my use cases, I actually added tags to the hash (`prime * result + Arrays.hashCode(tags)`) and to the quality (`return Arrays.equals(tags, other.tags);`).

With respect to the 'uniqueness' discussion above, I feel that makes most sense.

mlvdv:
Might adding tags to the hash make `SourceSection` unsuitable for building maps of source locations that ignore tags?  In the new Instrumentation code there is no longer any way to "remember" locations in guest language code other than `SourceSection` querying. So if the hashcode includes tags, then changing tags causes existing queries stop matching places where they originally matched, and this is true for all clients.

I continue to see evidence that forcing storage of tags into the `SourceSection` is inappropriate. A "tag" is conceptually a different kind of information, as I've described in a Skype chat about Instrumentation.  It doesn't say anything fundamental about the program, only that some tool has been instructed to behave a certain way in this particular situation. Some clients will care about this, others will not. We're already discovering that the binding times for the two kinds of information are quite different; the need to "replace" information that has already been recorded is a clue.

smarr:
Hm, source section querying? Do you have an example for what that is used?

Building the DynamicMetrics tool, I actually found a couple of things that surprised me.

For instance, I annotated message sends, and I annotated field reads. In Newspeak (SOMns) there is however no separate syntax for field reads. So, I end up with two different source sections for the same piece of source code. One annotated as message send, and one annotated as field read.

In the tool that actually works nicely. I render field reads as blue, and sends as italic. That combines nicely. And, as a reader of this highlighted code, I get a lot of information.

However, some might say, I abuse the instrumentation framework by exposing such dynamic information. But I tend to disagree. It works very nicely for me.

mlvdv:
Source section querying:  user sets a breakpoint (e.g. "the statement on line 42 in some source") and expects feedback whether the specified location actually exists.  "Don't know yet" is sometimes the right answer. The  debugger must (and already does) do extra work to manage "unresolved" breakpoints for files that haven't been loaded.  But the feedback is an important usability issue.  If the user thinks the location exists, but incorrectly, behavior is mysterious: an instance of the well-known "hidden state" usability failure.

mlvdv:
> For instance, I annotated message sends, and I annotated field reads. In Newspeak (SOMns) there is however no separate syntax for field reads. So, I end up with two different source sections for the same piece of source code. One annotated as message send, and one annotated as field read.

This came up when I was working with the Irvine Python people:  two nodes can (correctly) have the same `SourceSection`. In your case, querying with one tag or the other gets you the right location.
Getting multiple "locations" is also a reasonable result that can reveal something in the case you mention, for example if your query had both tags or just specified "any" tag.

I'll propose a stronger requirement:  a client must be able to distinguish between the case where (when all SourceSections are the same) a single node has two tags and where two separate node each have one of the tags.

mlvdv:
> However, some might say, I abuse the instrumentation framework by exposing such dynamic information. But I tend to disagree. It works very nicely for me.

I strongly disagree with what "some might say"!  I've been trying to get information out of source code and in front of programmers *usefully* for many years; that's the whole point of the Instrumentation framework. It is also why I argue constantly for *flexibility* and *openness* in the APIs:  I named it a "Framework" and I refer to it as a "tool kit". Anybody should be able to create, as easily as possible, developer tools we have not yet imagined.

chumer:
I will add hashCode, toString and equals implementation to SourceSection for tags. I missed that, thanks.
A cloneWithTags makes sense as well will add that. But with a little more documentation and testing.

Ok if I close this pull-request?

@mlvdv with the EventNodeFactory you can map mulitple "AST location" to one and the same SourceSection. If you care about same SourceSections beeing different in a Map you can use an IdentityHashMap in your tool implementation.

smarr:
Ok, closing it.

jtulach:
Veto: I am against the name **cloneWithTags**, I am OK with **withTags**.

chumer:
Ok. Will add withTags.

mlvdv:
@chumer 
> @mlvdv with the EventNodeFactory you can map mulitple "AST location" to one and the same SourceSection. If you care about same SourceSections beeing different in a Map you can use an IdentityHashMap in your tool implementation.

If I were to use an IdentityHashMap in that fashion, what guarantee would I be depending on about SourceSection identity:
- where would this guarantee be documented?
- how would this guarantee be enforced for language implementations, since they are responsible for creating and assigning SourceSection objects?

PR #28, opened 2016-02-02
Convert incoming values to SL values in SLForeignWriteNode

grimmerm:


jtulach:
Is there a way to test this in SL or in TCK so we avoid these ad-hoc fixes in future?

grimmerm:
maybe TCK could write values with different primitive types into an object (e.g. the SL object) and then read again - SL would need to provide the expected type of the read value.
not sure how these test should look like though.

chumer:
Looks good. Integrating.

chumer:
On a side note. The ForeignPropertyWriteNode can be declared using the DSL.

jtulach:
TCK has tests like that. The question is why they don't fail in case of SL.

woess:
why do you need to convert the value?

grimmerm:
We need to convert foreign primitive values to SL primitive values.
For example: if JavaScript tries to write a double value (i.e., a JS number) to an SL object, then we need to convert this double value to a long value (SL only supports the following primitive types: long, BigInteger). Hence, a SL object always contains either a long, BigInteger, or TruffleObject member.

smarr:
Hm, should this really become a `long`, or should it be a `TruffleObject` that knows that it is a JavaScript number? Going from `double` to `long` does not seem to maintain sufficient information in the generic case to be a straightforward transition I'd do automagically.

grimmerm:
Good point @smarr.
For SL I've decided that we simply convert double --> long.
Of course, your language can apply different rules. For example boxing the value, storing it as a TruffleObject (which itself supports an unbox message).
That's up to your language.
But not applicable to any language - for example C: in C we cannot store a boxed value, there needs to be a conversion to e.g. long, and if this conversion cannot be done, one can throw a com.oracle.truffle.api.interop.UnsupportedTypeException

smarr:
Ah, ok, I see. Thanks for the clarification.

woess:
I would have assumed that SL would do the simplest thing, that is, storing values verbatim in objects and only converting on-demand when a valid SL value is needed. But I agree it's a language choice.

PR #29, opened 2016-02-02
Add deprecation message to ForeignAccess#execute

grimmerm:


jtulach:
Maybe the message shall state the whole truth. That new specialized methods were added with more typesafety and the send is only good for rare need of 1:1 mapping...

PR #31, opened 2016-02-02
Fix virtual frame abbreviations

eregon:
`VirtualFrame` parameters seem to be abbreviated in different and inconsistent ways: `vFrame`, `vf` and `frame`. I believe only the last one is good for readability and is also the most used.
One of the obvious place where it looks strange is in the [JavaDoc](http://lafo.ssw.uni-linz.ac.at/javadoc/truffle/latest/com/oracle/truffle/api/interop/AcceptMessage.html).

If this create large conflicts with existing in-flights PR, I'm happy to redo it on top when they are merged.

### DIFF
--- a/truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/interop/WriteNode.java
+++ b/truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/interop/WriteNode.java
@@ -30,7 +30,7 @@ public final class WriteNode extends BaseWriteNode {
 
     @SuppressWarnings({"static-method", "unused"})
     @ExpectError({"The first argument must be a com.oracle.truffle.api.frame.VirtualFrame- but is java.lang.String"})
-    protected int access(String frame, Object receiver, Object name, Object value) {
+    protected int access(String string, Object receiver, Object name, Object value) {
         return 0;
     }
 }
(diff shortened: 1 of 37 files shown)


truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/interop/WriteNode.java
```diff
@@ -30,7 +30,7 @@
 
     @SuppressWarnings({"static-method", "unused"})
     @ExpectError({"The first argument must be a com.oracle.truffle.api.frame.VirtualFrame- but is java.lang.String"})
-    protected int access(String vf, Object receiver, Object name, Object value) {
+    protected int access(String frame, Object receiver, Object name, Object value) {
```
woess: The first parameter is not a frame here + 2 other occurrences.

eregon: Yes, although it's a "wrong frame parameter". Should we name that one "string" then?

eregon: Fixed in the new commit.

eregon:
Thanks!

PR #33, opened 2016-02-03
Importing truffle instrument rather than midi one

jtulach:
Impacts only Javadoc, but still.


jtulach:
Assigning to Christian, cannot find Michael in the list if assignees.

chumer:
No sound for breakpoints? :+1: merging

PR #34, opened 2016-02-04
Truffle language env lookup instrumenter

jtulach:
The simplest way to implement Env.lookup(Instrumenter.class) is to use Accessor. No need to invent any fancy public registration API - the accessor gives us internal way to connect these  pieces together.

Unit tests seem to work OK, but I am not sure about the eagerness - I create the LanguageInstrumenter everytime Env is created. Maybe that is too early...


smarr:
Not sure whether this is the best place to discuss this, but I am wondering how tools are parameterized.

There is this `InstrumentationInstrumenter`, which is a sibling of `LanguageInstrumenter`.
With this change, `LanguageInstrumenter` seems to get access to the TruffleLanguage.Env object.
The `InstrumentationInstrumenter` however has only access to a separate `TruffleInstrument.Env`, which is passed to `onCreate()`.

Now, I am wondering how to get parameters to the `onCreate()` method, for instance to set the port a debugger should listen on, the output file for my dynamic metric tool, etc.

Currently, the rewritten instrumentation API relies very much on Java properties, which seems to circumvent the whole Polyglot infrastructure. Not sure this is on purpose. I am also not entirely clear on the relation of these instrumenters and `PolyglotEngine` instances. I remember a couple of complex ideas (AST sharing) that might prevent a locality of these instrumenters and tools with respect to instances of `PolyglotEngine`.

chumer:
@jtulach promised me that this service API is going to be usable for instrumentations and for polyglot-engine users as well. So the way to paramatrize instrumentations is going to be with services. This will be done in another changeset. This is changeset is only to finalize the instrumentation API for guest languages. So they don't immediatly need to change their languages again after the service API comes along.

chumer:
Looks good merging.

truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentationHandler.java
```diff
@@ -621,7 +621,7 @@ boolean isInstrumentable(Node node) {
 
         @Override
         void initialize() {
-            language.installInstrumentations(context, this);
+            // language.installInstrumentations(env, this);
```
chumer: commented code. I can fix that.

jtulach:
Here is the sketch of similar concept and a demonstration how it can be used to configure debugger: https://github.com/jtulach/truffle/commit/eb6666e2424a6db961c2abb3c0c64c7f68ab90b5

Issue #25, opened 2016-01-29
SourceSection handling in Node


smarr:
What is the future of [`Node.assignSourceSection()`](https://github.com/graalvm/truffle/blob/master/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java#L84)?

Currently, the reassigning of a source section is protected by an `IllegalStateException`.
However, the API also provides a `public clearSourceSection()`, which enables reassignment.

The main reason for this issue is that the API is marked for revision, so, I hope we can settle on an API considering for instance the needs for the new instrumentation API.

@chumer suggested as solution a `protected setSourceSection(.)`, which could be used depending on a language's needs.

Other interested parties are most likely @woess and @jtulach.

A little bit of background: the new instrumentation API currently manages tags as part of `SourceSection`s. For me this creates the need to reassign the source sections on a node.
The situation in SOMns is, that I got a recursive descent parser, and thus, for instance when parsing expression trees recursively, I might only know after creating a full node with source section and state etc that I needs a tag for some tool. So, I got the need to update the tag and reassign the source section.
Using `clearSourceSection()` the use case is supported.

jtulach:
There is an attempt to lower the memory consumption and I am surprised we still carry the SourceSection field for all nodes - in spite just a few of them really use it. I would keep just the getSourceSection method and remove all the other methods. I am sure every language can create its own 
```
@CompilationFinal private SourceSection sourceSection;
```
and work with it properly.

smarr:
To address the point of the usage of the source section: This might evolve based on the instrumentation API and how tagging is going to work.

As I see it, trying to implement a tool with the API, actually a great many nodes need the source section because that's where the tags go. So, not sure removing `sourceSection` is really an improvement at this point.

chumer:
@jtulach the instrumentation API needs Node#getSourceSection(). I guess you would make getSourceSection() return null by default if we remove the field?

chumer:
I'd be in favor of removing the Node#sourceSection field.

chumer:
@woess What do you think about this?

woess:
looping in @mlvdv

mickjordan:
[...]

Issue #19, opened 2016-01-27
Include code snippets in sources Jars


pniederw:
A very common way to read Javadoc is through the sources (Jars) attached to the IDE project. (It's much more convenient to search/browse sources in an IDE than Javadoc on a web page, and seeing both documentation and source code at the same time is very useful.) However, seeing Javadoc such as `Usage example: {@codesnippet BranchProfileSample}` isn't helpful - instead I'd like to see the code snippet itself. Therefore, please embed code snippets in sources before creating sources Jars. (And please make sure line numbers in class files remain intact.)

smarr:
Hm, I kind of share the sentiment.
Not sure about the technical solution though. Seems problematic.

@jtulach, you got an idea?
Generally, I like that the code examples are checked by the compiler. With the changes we have this is absolutely necessary.

@pniederw, the relevant project is [codesnippet4javadoc](https://github.com/jtulach/codesnippet4javadoc), pull requests are probably welcome.

chumer:
I agree. Its also my favorite way of browser external libraries and code in general. While I understand the advantages of codesnippets its lack of IDE support kills it.

pniederw:
To be fair, after attaching Javadoc Jars to the IDE project (which at least Gradle doesn't do by default), documentation popup/window does reveal the sample code (in IntelliJ). But it's not the same experience as seeing it in the source code.

jtulach:
I was hoping attaching Javadoc will solve the most critical problem.

chumer:
In the current situation for me the disadvantages outweigh the advantages. I'd rather live with the redundancy of having the test case defined in a TestClass as well.

mlvdv:
I don't have a strong opinion about the tradeoff here, but I disagree with the choice of keywords "BEGIN" and "END".  They appears completely mysterious to the uninformed (human) reader, and we've seen examples in recent code reviews.  Keywords that are more self-explanatory would help.

woess:
@mlvdv +1

jtulach:
I don't understand what is wrong on sentences like "begin of sample xyz" and "end of sample xyz", but you are of course allowed to propose alternatives. Contribute to development of the codesnippets plugin by forking https://github.com/jtulach/codesnippet4javadoc ...

jtulach:
[...]

PR #35, opened 2016-02-04
Source section less node

jtulach:
According to discussion in #25 it seems that there is a general consensus that SourceSections should be handled by individual languages and the interface in Truffle API simplified to:

- Node.getSourceSection() that can be implemented by subclasses
- RootNode(SourceSection, ...) constructor accepting SourceSection as majority of RootNode instances have source section

The other methods are deprecated while the functionality is kept. Once the deprecated methods are removed, the Node#sourceSection field will be removed as well. RootNode will then get such field to hold the value provided in constructor.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,2 +1,12 @@
-Please see http://openjdk.java.net/contribute/ for details of how to contribute
-to OpenJDK projects including GraalVM.
+To submit pull requests for truffle, you need to sign the [Oracle Contributor Agreement][1].
+There is a white-list for contributors who have signed the OCA so please add a comment
+to your first pull request indicating the name you used to sign the OCA if it isn't clear
+from your github profile.
+
+Pull requests can only be merged by committers and a committer cannot merge
+his or her own pull requests.
+
+When creating a pull request, make sure you designate an assignee to ensure
+the request is processed in a timely manner.
+
+[1]: http://www.oracle.com/technetwork/community/oca-486395.html
--- a/truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/SourceSectionTest.java
+++ b/truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/SourceSectionTest.java
@@ -22,13 +22,10 @@
  */
 package com.oracle.truffle.api.dsl.test;
 
-import static com.oracle.truffle.api.dsl.test.TestHelper.createRoot;
+import com.oracle.truffle.api.CompilerDirectives;
 import static com.oracle.truffle.api.dsl.test.TestHelper.createRootPrefix;
 import static com.oracle.truffle.api.dsl.test.TestHelper.executeWith;
-import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.sameInstance;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertThat;
 
 import org.junit.Test;
 import org.junit.experimental.theories.DataPoints;
@@ -41,13 +38,15 @@ import com.oracle.truffle.api.dsl.Fallback;
 import com.oracle.truffle.api.dsl.NodeChild;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.dsl.internal.SpecializationNode;
-import com.oracle.truffle.api.dsl.test.SourceSectionTestFactory.SourceSection0Factory;
-import com.oracle.truffle.api.dsl.test.SourceSectionTestFactory.SourceSection1Factory;
 import com.oracle.truffle.api.dsl.test.TypeSystemTest.ArgumentNode;
 import com.oracle.truffle.api.dsl.test.TypeSystemTest.TestRootNode;
 import com.oracle.truffle.api.dsl.test.TypeSystemTest.ValueNode;
 import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.api.source.SourceSection;
+import static com.oracle.truffle.api.dsl.test.TestHelper.createRoot;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
 
 @RunWith(Theories.class)
 public class SourceSectionTest {
@@ -56,9 +55,9 @@ public class SourceSectionTest {
 
     @Theory
     public void testSourceSections(int value0, int value1, int value2) {
-        TestRootNode<SourceSection0> root = createRoot(SourceSection0Factory.getInstance());
+        TestRootNode<MutableSourceSectionNode> root = createRoot(SourceSectionTestFactory.MutableSourceSectionNodeFactory.getInstance());
         SourceSection section = SourceSection.createUnavailable("a", "b");
-        root.getNode().assignSourceSection(section);
+        root.getNode().changeSourceSection(section);
         expectSourceSection(root.getNode(), section);
         assertThat((int) executeWith(root, value0), is(value0));
         expectSourceSection(root.getNode(), section);
@@ -81,7 +80,21 @@ public class SourceSectionTest {
     }
 
     @NodeChild("a")
-    static class SourceSection0 extends ValueNode {
+    static class MutableSourceSectionNode extends ValueNode {
+        // BEGIN: source.section.mutable
+        @CompilerDirectives.CompilationFinal private SourceSection section;
+
+        final void changeSourceSection(SourceSection sourceSection) {
+            CompilerDirectives.transferToInterpreterAndInvalidate();
+            this.section = sourceSection;
+        }
+
+        @Override
+        public SourceSection getSourceSection() {
+            return section;
+        }
+
+        // END: source.section.mutable
 
         @Specialization(guards = "a == 1")
         int do1(int a) {
@@ -108,18 +121,27 @@ public class SourceSectionTest {
     public void testCreateCast() {
         SourceSection section = SourceSection.createUnavailable("a", "b");
         assertNull(section.getSource());
-        TestRootNode<SourceSection1> root = createRootPrefix(SourceSection1Factory.getInstance(), true, section);
+        TestRootNode<NodeWithFixedSourceSection> root = createRootPrefix(SourceSectionTestFactory.NodeWithFixedSourceSectionFactory.getInstance(), true, section);
         expectSourceSection(root.getNode(), section);
         assertThat((int) executeWith(root, 1), is(1));
         expectSourceSection(root.getNode(), section);
     }
 
     @NodeChild("a")
-    static class SourceSection1 extends ValueNode {
+    static class NodeWithFixedSourceSection extends ValueNode {
+        // BEGIN: source.section.fixed
+        private final SourceSection section;
 
-        public SourceSection1(SourceSection section) {
-            super(section);
+        public NodeWithFixedSourceSection(SourceSection section) {
+            this.section = section;
         }
+
+        @Override
+        public SourceSection getSourceSection() {
+            return section;
+        }
+
+        // END: source.section.fixed
 
         @CreateCast("a")
         public ValueNode cast(ValueNode node) {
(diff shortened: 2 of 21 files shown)


CONTRIBUTING.md
```diff
@@ -1,2 +1,12 @@
...
+to your first pull request indicating the name you used to sign the OCA if it isn't clear
+from your github profile.
+
+Pull requests can only be merged by committers and a committer cannot merge
+his or her own pull requests.
+
+When creating a pull request, make sure you designate an assignee to ensure
+the request is processed in a timely manner.
+
+[1]: http://www.oracle.com/technetwork/community/oca-486395.html
```
smarr: This looks like it should not be part of this pull request?

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
```diff
@@ -119,14 +127,31 @@ public NodeCost getCost() {
...
+     *
+     * {@codesnippet source.section.fixed}
+     *
+     * To create a node which can associate and change the {@link SourceSection} later at any point
+     * of time use:
+     *
+     * {@codesnippet source.section.mutable}
      *
      * @return the source code represented by this Node
      */
```
smarr: Does this comment need a transitory note? Mostly, because the implementation does not yet return null?

jtulach: If you use just non-deprecated API it returns null, so I think updating the Javadoc "in advance" is proper thing to do.

----- Reply message -----
Od: "Stefan Marr" <notifications@github.com>
Komu: "graalvm/truffle" <truffle@noreply.github.com>
Kopie: "Jaroslav Tulach" <jaroslav.tulach@oracle.com>
Předmět: [truffle] Source section less node (#35)
Datum: čt, úno 4, 2016 12:06

In truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java:
>      public void clearSourceSection() {
>          this.sourceSection = null;
>      }
>  
>      /**
> -     * Retrieves the segment of guest language source code that is represented by this Node.
> +     * Retrieves the segment of guest language source code that is represented by this Node. The
> +     * default implementation of this method returns <code>null</code>. If your node represents a
> +     * segment of the source code, override this method and return a <code>final</code> or
> +     * {@link CompilationFinal} field in your node to the caller.
> +     *
> +     * To define node with <em>fixed</em> {@link SourceSection} that doesn't change after node
> +     * construction use:
> +     *
> +     * {@codesnippet source.section.fixed}
> +     *
> +     * To create a node which can associate and change the {@link SourceSection} later at any point
> +     * of time use:
> +     *
> +     * {@codesnippet source.section.mutable}
>       *
>       * @return the source code represented by this Node
>       */

Does this comment need a transitory note? Mostly, because the implementation does not yet return null?

&mdash;

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
```diff
@@ -119,14 +127,31 @@ public NodeCost getCost() {
...
+     * To define node with <em>fixed</em> {@link SourceSection} that doesn't change after node
+     * construction use:
+     *
+     * {@codesnippet source.section.fixed}
+     *
+     * To create a node which can associate and change the {@link SourceSection} later at any point
+     * of time use:
+     *
+     * {@codesnippet source.section.mutable}
      *
```
chumer: I am still not convinced by the usefulness of these codesnippets. I would even say that most GL devs don't read javadocs but are browsing using the IDE where codesnippets dont translate nicely.

smarr: Hm, well, the issue would be little less problematic if at least the names would be proper JavaDoc names (like for `@link`) that one can click on to navigate to the example.

woess: yes, including at least the classname would help a bit to find the snippet within an IDE...

jtulach: I would prefer to leave the snippet discussion in #19 - I understand they are source of tension. That said the idea to make them behave like a link or at least include the name of source file is certainly worth exploring.

woess:
@jtulach I'd like to hear other people's opinion about this as well, but I think getSourceSection() should stay public.

mlvdv:
`getSourceSection()` should be public.  There will be tools (none yet) for which precise source attribution of nodes will be a requirement.  This translates to the the invariant that a node was produced by parsing the text represented by its `SourceSection`.  In contrast, `getEncapsulatingSourceSection()` should mean something else:  what is the nearest enclosing syntactic unit that this `Node` is part of?  
[Aside: `getEncapsulatingSourceSection()`  only makes sense reliably when ASTs are "well-behaved" with respect to source attribution, something discussed in some detail elsewhere.  Well-behaved ASTs are important for some kinds of tools, but this imposes a constraint on AST design that I've often seen violated, e.g. in *javac*.]
@mickjordan may want to comment; we spoke about this while the FastR AST was under development.

jtulach:
I see. A real use-case for a tool that we yet have to imagine.

If I accept as given that getSourceSection is needed - do we need getEncapsulatingSourceSection() helper method at all? It seems to me like utility method that iterates over chain of nodes. Utility methods don't belong into core API. Each tool write the loop on their own.

mickjordan:
On 2/4/16 11:38 AM, Michael Van De Vanter wrote:
>
> |getSourceSection()| should be public. There will be tools (none yet) 
> for which precise source attribution of nodes will be a requirement. 
> This translates to the the invariant that a node was produced by 
> parsing the text represented by its |SourceSection|. In contrast, 
> |getEncapsulatingSourceSection()| should mean something else: what is 
> the nearest enclosing syntactic unit that this |Node| is part of?
>
> [Aside: |getEncapsulatingSourceSection()| only makes sense reliably 
> when ASTs are "well-behaved" with respect to source attribution, 
> something discussed in some detail elsewhere. Well-behaved ASTs are 
> important for some kinds of tools, but this imposes a constraint on 
> AST design that I've often seen violated, e.g. in /javac/.]
> @mickjordan <https://github.com/mickjordan> may want to comment; we 
> spoke about this while the FastR AST was under development.
>
>
In FastR we removed all uses of getEncapsulatingSourceSection and throw 
an exception if anyone uses it. Our model is that you first find the 
RSyntaxNode (getRSyntaxNode) and then call getSourceSection, which must 
not fail. We actually try hard to avoid accessing the source section 
until we really need it, like printing an error with context. We do 
support a hierarchical search by default for getRSyntaxNode, but a node 
can override if it's node structure is not hierarchical. As you recall I 
tried to get this model into truffle without success, so we have a 
special subclass of Node that contains the new methods and overrides of 
the Truffle ones. It's a slight pain to have to remember to use this in 
preference to Node. You can subclass Node directly but you have to 
follow some rules to be compliant with the model.

Mick

jtulach:
I see that FastR [overrides getEncapsulatingSourceSetion](http://lahoda.info/index/ui/index.html#/showCode?path=graal&relative=fastr%2Fcom.oracle.truffle.r.runtime%2Fsrc%2Fcom%2Foracle%2Ftruffle%2Fr%2Fruntime%2Fnodes%2FRBaseNode.java&goto=METHOD:com.oracle.truffle.r.runtime.nodes.RBaseNode:getEncapsulatingSourceSection:%28%29Lcom%2Foracle%2Ftruffle%2Fapi%2Fsource%2FSourceSection%3B) to delegate to **protected** method that finds the right node and extracts the section from it. I see [four overrides of the method](http://lahoda.info/index/ui/index.html#/usages?signature=METHOD:com.oracle.truffle.r.runtime.nodes.RBaseNode:getRSyntaxNode:%28%29Lcom%2Foracle%2Ftruffle%2Fr%2Fruntime%2Fnodes%2FRSyntaxNode%3B). FastR overrides *getSourceSection*, but only to throw exceptions in unexpected cases. FastR calls getSourceSection - but mostly only because it passes the section somewhere else (*assignSourceSection* or constructor of other node).

As far as I can tell FastR would benefit from language controlled *find/getSourceSection* for a node.

chumer:
I like the change. +1

I'd leave getSourceSection public for now, but maybe revisit this in another changeset? I think the instrumentation tools could use accessors to access the source section. Tools can access source sections using the EventContext via public API.

@chrisseaton @woess please also accept or deny this change 

@jtulach After that please rebase and merge.

truffle/com.oracle.truffle.api.dsl/src/com/oracle/truffle/api/dsl/internal/DSLShare.java
```diff
@@ -146,7 +146,8 @@ public T call() {
 
     private static void updateSourceSection(Node oldNode, Node newNode) {
         if (newNode.getSourceSection() == null) {
-            newNode.assignSourceSection(oldNode.getSourceSection());
+            throw new IllegalArgumentException("Can't update source: " + oldNode);
+            // newNode.assignSourceSection(oldNode.getSourceSection());
```
chrisseaton: Comment left in by mistake?

truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/SourceSectionTest.java
```diff
@@ -108,19 +121,28 @@ Object do4(Object a) {
...
+        TestRootNode<NodeWithFixedSourceSection> root = createRootPrefix(SourceSectionTestFactory.NodeWithFixedSourceSectionFactory.getInstance(), true, section);
         expectSourceSection(root.getNode(), section);
         assertThat((int) executeWith(root, 1), is(1));
         expectSourceSection(root.getNode(), section);
     }
 
     @NodeChild("a")
-    static class SourceSection1 extends ValueNode {
+    static class NodeWithFixedSourceSection extends ValueNode {
+        // BEGIN: NodeWithFixedSourceSection
```
chrisseaton: Are these comments directives to some tool? What do they do?

chrisseaton:
I'm not sure what the final proposed change is, as @chumer's last comments suggests more changes which may or may not be accepted, so it's hard to accept right now.

I currently use `getEncapsulatingSourceSection` on arbitrary foreign nodes to print backtraces for other language's stack frames in Ruby, so I'd need that functionality to remain or an alternative.

I don't mind maintaining the `SourceSection` field myself in Ruby, that seems like a good idea so languages can do it in different ways, but I think `getSourceSection` (maybe with the semantics of `getEncapsulatingSourceSection`) should stay, given the example of my backtraces.

truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/SourceSectionTest.java
chumer: See https://github.com/graalvm/truffle/issues/19

CONTRIBUTING.md
woess: yes, please rebase.

woess:
@chumer @jtulach 
I accept the change (needs to be rebased though).
For now, getSourceSection(), getEncapsulatingSourceSection(), and RootNode should be kept as is, and any suggestion to change them is out of scope of this PR.

truffle/com.oracle.truffle.api.dsl/src/com/oracle/truffle/api/dsl/internal/DSLShare.java
jtulach: I kept the comment as I expected the exception to be raised at some places, but it wasn't raised yet. But we haven't yet executed all our language tests. Maybe it is wiser to keep it for now.

jtulach:
OK, this change request will just deprecate Node(SourceSection) constructor and Node#assignSourceSection and Node#cleanSourceSection. Unless there are objections, let's merge tomorrow.

jtulach:
No objections. Merging.

PR #38, opened 2016-02-05
Make Message.WRITE final

eregon:
All `Message` statics are `final` except `WRITE`.

* The various `INSTANCE` fields are also not final, but accessed only here, should we make them `final` as well?


woess:
@eregon yes, the `INSTANCE` fields should be final, too.

eregon:
OK, will do.

eregon:
Done.
`sigtest` data had to be adapted but I don't see any valid use case to set `Message.WRITE`, not to mention it's highly inconsistent. So I believe changing has no effect on real users.

jtulach:
Changing WRITE field to final is correct, but incompatible change. I'd like to do it, but please get wider approval. Write to mailing list and to known language maintainers to confirm they are OK with the change. (Assuming they are) merge then.

woess:
I approve of the change. It's an API error and _very_ unlikely somebody depends on it (nobody should ever write to the (WRITE) field), so it should be fixed.

jtulach:
I'll merge this tomorrow.

woess:
ready for merging, but needs to be rebased; @eregon

eregon:
Rebased.

PR #40, opened 2016-02-07
Add a DSL for generating layouts for the object model

chrisseaton:
Ruby has a relatively Java-style class hierarchy, and many complex built-in classes which we implement in Java and need many fields to store state. The OM is more suited to simpler prototypical objects with few internal properties, so we found it a little verbose to use in JRuby. This DSL allows the declaration of Java-style classes, but backed using the OM. In JRuby this made it possible to move to an OM-only object style, where previously we had our own classes for internal fields and so two objects for every Ruby object (the OM object and the instance of our class).

This PR isn't ready for merge yet as the implementation of the generator is not yet high enough quality, but I'd like a review of the API and its semantics before I spend time refining the implementation, so please take a look at the `com.oracle.truffle.api.object.dsl` and `com.oracle.truffle.api.object.dsl.test` packages. Please ignore the implementation in the `com.oracle.truffle.api.object.dsl.generator` packages.

The API already works very well for JRuby.

@chumer @jtulach please review for API, and @woess please check I'm using the OM correctly. Please don't merge (unless you're happy with the poor implementation of the generator).


truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/layout/LayoutGenerator.java
```diff
@@ -0,0 +1,839 @@
...
+                    if (property.isVolatile()) {
+                        if (property.getType().getKind() == TypeKind.INT) {
+                            stream.printf("        return ((AtomicInteger) %s_PROPERTY.get(object, true)).get();%n", NameUtils.identifierToConstant(property.getName()));
+                        } else if (property.getType().getKind() == TypeKind.BOOLEAN) {
+                            stream.printf("        return ((AtomicBoolean) %s_PROPERTY.get(object, true)).get();%n", NameUtils.identifierToConstant(property.getName()));
+                        } else {
+                            stream.printf("        return ((AtomicReference<%s>) %s_PROPERTY.get(object, true)).get();%n", property.getType(), NameUtils.identifierToConstant(property.getName()));
+                        }
+                    } else {
+                        stream.printf("        return (%s) %s_PROPERTY.get(object, true);%n", property.getType(), NameUtils.identifierToConstant(property.getName()));
```
woess: The condition should not be `true`, but probably `is%s(object)`, where `%s` is `layout.getName()`. (applies to all the gets)

chrisseaton: I presume ideally it would be a boolean data edge from the earlier `is%s(object)` call that the user is presumably doing? But we don't have access to that in the DSL.

Can you tell me about the difference between doing `PROPERTY.get(object, true)`,  `PROPERTY.get(object, false)` and  `PROPERTY.get(object, isFoo(object))`. I think there are some subtleties here that I may not be aware of.

truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/ObjectTypeSuperclassTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.junit.Assert;
+
+public class ObjectTypeSuperclassTest {
+
+    public static class ObjectTypeSuperclass extends ObjectType {
+
+    }
+
+    @Layout(objectTypeSuperclass = "com.oracle.truffle.api.object.dsl.test.ObjectTypeSuperclassTest.ObjectTypeSuperclass")
```
chumer: Why not a class literal for objectTypeSuperclass?

chumer:
In general this is a good idea.

One problem of this approach is that layout implementations cannot speculate because they are not implemented in the context of a Node. Or do I miss something?

For the object model accesses I think I'd prefer standardized but still flexible nodes that languages reuse to perform certain operations like reading writing, iterating.

I had in mind some abstract datatype pattern for OM, similar to interop but simpler, that lets you create nodes instead of calling methods directly. We could still use a DSL to generate the node constructors invocations, but a bit different like this:

```
   public final class BasicTestLayoutImpl implements BaseTestLayout {

        AllocateBasicTestNode createBasicTest(int value);

        public PropertyGetNode getValue(DynamicObject object) {
            return PropertyGetNode.create(object, "value")
       ]

        public PropertySetNode setValue(DynamicObject object, int value) {
           return PropertySetNode.create(object, "value")
      }
      
      public abstract static class AllocateBasicTestNode extends DynamicObjectConstrutor {
            @Child PropertySetNode setValue;
          
            public DynamicObject allocate(int value) {
               DynamicObject object = super.allocate();
               setValue.executeSet(object, value);
               return object;
            }
       }   


    }

      public abstract static class PropertyGetNode extends Node {
         
         public abstract Object executeGet(DynamicObject object)
          ...
       }   

public abstract static class PropertySetNode extends Node {
         
         public abstract Object executeSet(DynamicObject object, Object value)
          ...
       }   
```

@chrisseaton  What do you think?

truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/ObjectTypeSuperclassTest.java
chrisseaton: I think when I wrote this in JRuby some compilation order problem meant that was hard. I'll try again.

chrisseaton:
The main motivation here is for something that works without nodes and doesn't need to speculate. I could just store all my implementation fields (by which I mean things like the `byte[]` for a `String` - things the user never sees but I need to store in objects) like normal Ruby instance variables but with special names, and use normal instance variable field nodes to read and write them. That would have two problems.

First of all, it would mean an explosion of nodes and code. All the places where I currently get the size of an array, or its storage field, or the bytes in a string, go from being one line to being at least another `@Cached` parameter.

Imagine this code using nodes instead of `Layouts.ARRAY.get...`: https://github.com/jruby/jruby/blob/c2787507d3f07ee9c847f6b89cf2174fc5241a56/truffle/src/main/java/org/jruby/truffle/core/array/ArrayAppendManyNode.java#L142-L158. That's not even a `@Specialization`, so the nodes would either have to be fields (more code) or passed in as parameters (yet more code).

This one method alone would need nine new nodes (or one slow, polymorphic one) https://github.com/jruby/jruby/blob/ff9e6968d47127a0f5bc4beccb032b9f75074e6a/truffle/src/main/java/org/jruby/truffle/core/proc/ProcNodes.java#L141-L149. If I have lots of fields why not store them in a single object? Because then that's another object to allocate.

Secondly, I can't execute nodes in code behind a `@TruffleBoundary`. Imagine have to somehow hoist code like this out from behind a boundary in order to be able to execute a property read node: https://github.com/jruby/jruby/blob/28a97d92915a90edf71255d0985c2b6d7eaaaba9/truffle/src/main/java/org/jruby/truffle/core/encoding/EncodingNodes.java#L193-L193. I'd have to split the method into two - one that did property reads, and other that was the boundary method. I'm not sure that would be practical in many cases.

Graal.js already uses the same approach as this DSL, just manually written out (see the implementation of `JSString` and the `Property` members). Graal.js would also have the same problems with the approach you are suggesting - such as property reads behind boundaries.

So I don't think using nodes for implementation fields is tractable.

What's more. I believe that my current implementation wouldn't even benefit from any speculation, as everything is as static as the OM allows and the code should compile to not much more than a Java field or array access. Maybe @woess can confirm or deny?

nirvdrum:
I'm perhaps biased by having used this in JRuby, but I'd like to be able to access these fields easily without having to rely on passing nodes around. From what I've seen, the static helpers compile down to simple field loads. I don't know if doing that in a node would allow speculative optimizations, but we're mostly representing mutable state in a guest language object . . . it seems unlikely that we'll see values that are constant enough to even speculate on. JRuby has a handful of these (e.g., a String's encoding), but we always end up using them as guards in specializations.

Having said that, if the node approach can make things faster I don't think there's a problem in also generating helper nodes. I don't see the two approaches as being at odds with one another.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/layout/LayoutGenerator.java
woess: yes, but Graal can GVN to an earlier isFoo guard.

PROPERTY.get(object, _condition_):
* false: fixed in control flow, always safe
* true: floating unguarded (unsafe)
* isFoo(object) or result of an earlier guard: dependency edge on the guard (GVN-able)
* _shape_: equivalent to object.getShape() == _shape_

truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/InheritanceTest.java
```diff
@@ -0,0 +1,253 @@
...
+        BigInteger getC(DynamicObject object);
+
+        void setC(DynamicObject object, BigInteger value);
+
+    }
+
+    private static final BaseLayout BASE_LAYOUT = BaseLayoutImpl.INSTANCE;
+    private static final MiddleLayout MIDDLE_LAYOUT = MiddleLayoutImpl.INSTANCE;
+    private static final TopLayout TOP_LAYOUT = TopLayoutImpl.INSTANCE;
+
```
jtulach: I'd like all the implementations in one package to be generated into single class, with a default name, for example **Layouts**

chumer:
@chrisseaton  There is a solution for property reads (or nodes) behind boundaries in FastR
https://github.com/graalvm/fastr/blob/master/com.oracle.truffle.r.nodes/src/com/oracle/truffle/r/nodes/access/vector/ExtractVectorNode.java#L147-L172

Regarding:
https://github.com/jruby/jruby/blob/c2787507d3f07ee9c847f6b89cf2174fc5241a56/truffle/src/main/java/org/jruby/truffle/core/array/ArrayAppendManyNode.java#L142-L158

This is a case where you redo the caching in the operation instead of the property accesses. This way you could probably express this code with a generic cache without most of the specializations. I can offer doing a sketch if needed. 

BTW.: As a rule of thumb: Many specializations that all forward to the same method indicates that a generic cache might be useful.

Having nodes for property accesses might be useful also for instrumenting the AST (eg instrument all object writes).

Another interesting thing is that without nodes you cannot speculate on constant fields. We might want to decide to do such kind of optimizations in Ruby in the future.

Regarding footprint: You can reuse the standard single read and write node implementation for property accesses. You can and probably should preinitialize the node with the specialization for the shape and attribute on creation. So the call to the node can become monomorph in the interpreter. The class can even be shared across language implementations. Additional nodes need to be allocated in this case which is a clear downside with the nodes approach.

> So I don't think using nodes for implementation fields is tractable.

To sum up: 
* On the use side its an additional child field or @Cached attribute for the property access nodes.
* You can read a property from any object without type check which enables more generic implementations of some operations closer to Ruby code.
* Property accesses become instrumentable
* You can use TruffleBoundaryNodes to access files behind boundaries.
* Yes nodes need to be allocated and require additional space.
* Using generic field access nodes hides more details about the OM. No need to know about shapes at all.

I think the most important issue is that we can hide bigger parts of the OM implementation.

You have good points, but I think regarding future evolution generic attribute access nodes might be a better choice.  I also don't agree with the code that is in graal.js in that regard.

truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/InheritanceTest.java
chrisseaton: Do you mean just the instance fields, so it would be `Layouts.BASE` or something like that? I could do that. Why do you prefer it though?

chrisseaton:
If the proposed benefits of your approach, such as speculative constants and instrumentation, are nice-to-have rather than nessecary immediately for correctness or performance, then maybe the DSL could provide my static methods now and in the future, but add node getters and setters later on? Then we could transition code gradually.

chumer:
Yes that could be a way forward. We should look on the R side if we can use this as well. I will have another look at the implementation tomorrow.

woess:
> maybe the DSL could provide my static methods now and in the future, but add node getters and setters later on

I think that's a good compromise. Currently, there does not seem to be much benefit in channeling all accesses through nodes.

woess:
> What's more, I believe that my current implementation wouldn't even benefit from any speculation, as everything is as static as the OM allows and the code should compile to not much more than a Java field or array access.

I agree, although type checks could benefit from specialization.

What I like most about the DSL is implementation detail hiding and less boilerplate code.

woess:
@chrisseaton I see that you have commented code in various places. We should try to avoid that.

chrisseaton:
Yes I'll re-do the implementation (where the comments are) from scratch with `CodeTree`.

truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Volatile.java
```diff
@@ -0,0 +1,51 @@
...
+ * boolean compareAndSetNAME(DynamicObject object, TYPE expected, TYPE newValue);
+ *
+ * TYPE getAndSet(DynamicObject object, TYPE value);
+ * </pre>
+ *
+ * @see Layout
+ */
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.CLASS)
+public @interface Volatile {
```
jtulach: Retention "class"? That surprising. There is an annotation processor, so "source" ought to be enough, I think.

chrisseaton: Yes, looks like it works as `SOURCE`.

mlvdv:
I vote for any approach that allows straightforward instrumentation of all assignments.  This would permit a reasonable implementation of debugger watchpoints, for example, with the same performance advantages that we have for breakpoints.  Thomas asked me once, quite a while back, whether we could ever do "value tracing"; I hadn't thought about it since then, but this would also be the most practical way to approach that as well.  This would be a pretty impressive capability of the Truffle platform.

chrisseaton:
> I vote for any approach that allows straightforward instrumentation of all assignments.

But these aren't source-level assignments. They're internal, implementation fields. They aren't visible to the person writing a Ruby program. It's like wanting to set a breakpoint on the header work in Java objects.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/layout/LayoutGenerator.java
chrisseaton: Done.

truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/ObjectTypeSuperclassTest.java
chrisseaton: Done.

truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Volatile.java
chrisseaton: Done.

truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/InheritanceTest.java
chrisseaton: I looked into this and can't see any way to implement it correctly. I don't have any control over when my processing rounds are run, or when new source files with `@Layout` annotations in them may be generated by other processors that I don't know about, so I can't see any way to generate one file for all layouts in a package at the same time.

Do you know a way to make it possible?

jtulach: Yes, I know. One always needs to process the whole package. I thought I even wrote a blog post about that, but I cannot find it. I used the code at http://hg.netbeans.org/html4j/file/ed4b25eb66f3/boot/src/main/java/org/netbeans/html/boot/impl/JavaScriptProcesor.java which also needs to generate single `$JsCallbacks$` file per package.

truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Volatile.java
```diff
@@ -0,0 +1,51 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.truffle.api.object.dsl;
+
```
jtulach: Is there strong reason for the DSL being in separate module, package? I know the "node DSL" is separated, but that is because it is big, right? The `AcceptMessage` annotation on the other hand is next to other interop APIs. I know the object.api itself hasn't yet been polished, but if it was, I would prefer these three annotations to go next to it.

Of course the current layout makes it easier to deprecate these annotations while keeping the object.api untouched. I am not sure that is our goal, however.

chrisseaton: > Is there strong reason for the DSL being in separate module, package?

It just seemed like separate functionality so it made sense to isolate it. I did have a vague idea that we might want to build some layers here - where people can understand that there is core functionality, and then things built on top of that.

I don't mind moving it to wherever, so if someone with a stronger opinion and a justification asks me to move it that's fine.

truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Layout.java
```diff
@@ -0,0 +1,299 @@
...
+ * Inheritance of layout interfaces allows you to model classical class inheritance, such as in a
+ * language like Java. Use normal interface inheritance to make one layout inherit from another. You
+ * then need to add the parameters for super-layouts at the beginning of sub-layout constructor
+ * methods.
+ *
+ * Inherited shape properties work in a similar way.
+ *
+ * <pre>
+ * &#64;Layout
+ * interface BaseLayout {
```
jtulach: Following comment depends on result of #19: I want all our code samples to be provided via the code snippets infrastructure - e.g. to be error proof, hyperlinked and not printed in a hard to read uniform font/color.

truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Volatile.java
woess: I'm in favor of the separation. It makes it clear what belongs to the annotation processor based DSL and what to the underlying API (with a clear one-way dependency from the former to the latter). We followed a similar approach for the Truffle DSL (`com.oracle.truffle.api.dsl`). Everything that the DSL does for you could potentially be implemented by hand.

jtulach:
I am trying to provide the per-package generator, but the gate isn't clean and I don't know if there are changes in my code or in your branch. Can you please fix your code to pass gate?

chrisseaton:
I haven't finished tidying up the implementation yet to a standard where it passes the gate. I'm working on that today.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/layout/LayoutGenerator.java
```diff
@@ -0,0 +1,845 @@
...
+
+import java.io.PrintStream;
+import java.util.ArrayList;
+import java.util.List;
+
+public class LayoutGenerator {
+
+    private static final boolean GENERATE_SHAPE_BOUNDARY_CONSTANT_CHECKS = Boolean.valueOf(
+                    System.getProperty("org.jruby.truffle.om.dsl.processor.layout.LayoutGenerator.GENERATE_SHAPE_BOUNDARY_CONSTANT_CHECKS", Boolean.FALSE.toString()));
+
```
jtulach: You may want to remove this JRuby like named constant. If you keep it, it has to be documented and tested.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/layout/LayoutGenerator.java
```diff
@@ -0,0 +1,845 @@
...
+    public LayoutGenerator(LayoutModel layout) {
+        this.layout = layout;
+    }
+
+    public void generate(final PrintStream stream) {
+        stream.printf("package %s;%n", layout.getPackageName());
+        stream.println();
+        stream.println("import java.util.EnumSet;");
+        stream.println("import com.oracle.truffle.api.object.*;");
+        stream.println("import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;");
```
jtulach: Truffle uses FQN imports and no wildcards. I guess the generated code could do the same.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/layout/LayoutGenerator.java
chrisseaton: Yes I'll do that. As I say, when I opened this review I was just ready for the API and semantics to be reviewed. Now that people seem to generally approve of that I'll go back and tidy up these implementation classes.

jtulach:
Here is a sketch of the "single generated file per package" - it compiles, but it doesn't pass the tests yet: https://github.com/jtulach/truffle/commit/93a7da4c7a2e3824f84eac02d01bff2e6938d9ba

woess:
@jtulach @chumer does this "single generated file per package" approach work correctly with incremental compilation (eclipse)?

jtulach:
Of course it works (unless somebody reports it doesn't).

chumer:
Looks like this could work. Definitely needs verification that it works with eclipse incremental compilation. Writing tests for this is rather hard unfortunately. Will have a more in depth look soonish.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/layout/LayoutGenerator.java
chrisseaton: Done.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/layout/LayoutGenerator.java
chrisseaton: Done.

truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Layout.java
chrisseaton: Done.

chrisseaton:
I don't think single file does work. I merged @jtulach's branch into mine https://github.com/chrisseaton/truffle/tree/om-dsl-one-file, and it builds fine from mx but the moment you open Eclipse it rewrites it with a file with no implementations in.

I don't understand how @jtulach's changes will get the compiler to process files that it wouldn't do normally. If I change a single file, and it compiles that, why would it also send me files that haven't changed to be compiled. I don't see the mechanism that does that.

chrisseaton:
Tests are green. I don't have nodes (planning to defer to a subsequent commit) or single file support (not been explained why it's a good idea, and can't get it to work).

I'm just writing a few more tests and fixing a couple of issues that JRuby has exposed now.

chrisseaton:
@jtulach @woess @chumer and anyone else interested, can I get a fresh opinion on this PR? I've finished implementation work so it can all be reviewed now. As I said earlier the only two comments I haven't done anything for are nodes for getters and setters, as I think this can be done as a separate PR later, and I haven't done the single-file version, as I'm not sure what the benefits are or if it can be made to work.

woess:
I believe the single-file version cannot work reliably.
Overall, I'm happy with the cleanup and changes. I'll look at it a bit more tomorrow. I didn't see documentation for the `DynamicObjectFactory` parameter to the `create` method, but maybe I just overlooked it.

chrisseaton:
@woess it's talked about here https://github.com/graalvm/truffle/pull/40/files#diff-f0c09dee90b1682f275e703e074a2535R148 - easier to understand with the snippets expanded.

jtulach:
"will get the compiler to process files that it wouldn't do normally. If I change a single file, and it compiles that, why would it also send me files that haven't changed to be compiled"

My change is asking for a package for the compile file(s) and then asks for all types in that package(s). That is something Java compiler has to do to resolve package private references, so it should work. I'll do some debugging.

woess:
@chrisseaton Ah, right, thanks. About the snippets: While I appreciate how they're done (as inner classes so they're easy to find and close to the javadoc where they're used; checked by javac and formatted), I feel like these classes shouldn't be in the truffle jar.

chrisseaton:
Yes and the source does become a bit of a mess - lots of helpers in order to be able to write what you want to. I don't mind though.

truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Layout.java
```diff
@@ -0,0 +1,412 @@
...
+ *
+ * {@codesnippet customid}
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.SOURCE)
+public @interface Layout {
+
+    Class<? extends ObjectType> objectTypeSuperclass() default ObjectType.class;
+
+    class Snippets {
```
jtulach: Every member of an interface is public, so you can't hide Snippets at this place. But you could move the `class Snippets` definition to the end of this file. A .java file can contain many subsequent class definitions. At most one of them can be public. So if you add `class LayoutSnippets` to the end (or even beginning) of this file it should be gone from the Javadoc.

If we agree on same naming scheme, we can even modify mx to not include **/*Snippet*.class files in the final JAR. With that in place most of the issues raised in #19 might be solved without negative impact on the Javadoc and final JAR bits.

chrisseaton:
The snippets class has been moved to top-level and package-private at the end of the file containing the `Layout` interface.

Final opinions, please. I'll merge at close of play tomorrow unless there are objections.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/LayoutParser.java
```diff
@@ -0,0 +1,494 @@
...
+            if (layoutElement.getInterfaces().size() > 1) {
+                processor.reportError(layoutElement, "@Layout interfaces can have at most one super-interface");
+            }
+
+            final DeclaredType superInterface = (DeclaredType) layoutElement.getInterfaces().get(0);
+            parseSuperLayout((TypeElement) superInterface.asElement());
+        }
+
+        for (AnnotationMirror annotationMirror : layoutElement.getAnnotationMirrors()) {
+            if (annotationMirror.getAnnotationType().toString().equals(Layout.class.getCanonicalName())) {
```
chumer: You are lucky that this works. toString is undefined and in fact works different between ECJ and the eclipse compiler. Here is the version I use to in ElementUtils of Truffle DSL.
```
    public static String getQualifiedName(TypeElement element) {
        String qualifiedName = element.getQualifiedName().toString();
        if (qualifiedName.contains("$")) {
            qualifiedName = qualifiedName.replace('$', '.');
        }
        return qualifiedName;
    }
```

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/LayoutGenerator.java
```diff
@@ -0,0 +1,901 @@
...
+            stream.println("        ");
+        }
+
+        stream.println("    }");
+        stream.println("    ");
+    }
+
+    private void generateAllocator(final PrintStream stream) {
+        if (layout.getSuperLayout() == null) {
+            stream.println("    protected static final Layout LAYOUT = Layout.newLayout().addAllowedImplicitCast(Layout.ImplicitCast.IntToLong).build();");
```
chumer: `addAllowedImplicitCast(Layout.ImplicitCast.IntToLong)`
is Ruby specific isn't it? Is there a way to configure this? There should be.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/LayoutParser.java
chrisseaton: Fixing this. What is the correct way to get a `Name` as a `String`. `toString` isn't specifically defined there either.

chumer:
CH1) I'd like to see the hardcoded addAllowedImplicitCast somehow fixed (no good idea how though). 

CH2) Also I don't feel very good with TypeMirror#toString uses in the generator. Maybe you can copy getQualifiedName(TypeMirror) from ElementUtils?

CH3) (can be done later as well) We introduce a new distribution for this DSL that needs to be configured and distributed. Maybe we should put this processor into the TRUFFLE_DSL_PROCESSOR distribution as well? So we always have one truffle-api jar and one processor jar in the end. For a guest language developer granularity this fine does not make much sense. @chrisseaton What do you think?

CH4) (definitely for later) recast the generator with CodeTreeBuilder. Maybe we should track this with an issue?

Besides these small issues, I think this is a significant simplification compared to using the object model API directly. 
Accept.

chrisseaton:
CH1 resolved in df3c888ca86f5d1adc34e4dec941488992243d39

CH2 resolved in 90424f3751b6b8dc233eeb5004cdb7a38f0005d2

jtulach:
Re. CH3 - I must have been blind to not notice this myself. The vision so far was to have one *truffle-api.jar* and one *truffle-dsl-processor.jar* - originally the DSL, but also all interop, TruffleLanguage and co. processors are in the second one. I believe there is no need to increase the number of distributions for Truffle - two is enough. Please include the new processor into the existing  *truffle-dsl-processor.jar*.

Btw. should the new processor be in a separate JAR, we need to control imported/exported API between the new processor and existing *truffle-dsl-processor.jar* as they can be distributed separately and it could happen one is using 0.11 of *truffle-dsl-processor.jar* with 0.12 of the new one, or even some wilder combinations. That is a lot of headache which we can avoid by packaging all the processors into one distribution.

woess:
I feel like it's in a good shape and ready for inclusion.
Two more comments, though:

There seems to be a problem with generating the factory method (`create*`) for the (Inherited)ShapePropertiesTest when the processor is run from eclipse (4.4).

Do you think it could make sense to create the `*LayoutImpl` with optionally, say, a `Layout` parameter in the future? e.g.,
`CAST_LAYOUT = CastLayoutImpl.INSTANCE`
could then potentially be:
`CAST_LAYOUT = CastLayoutImpl.create(Layout.newLayout().addAllowedImplicitCast(IntToLong).build())`
I don't really see a need for it right now (your solution for CH1 is fine), but I'm thinking that maybe we should still use an instantiation method instead of a field from an API flexibility perspective. Initially it would simply return INSTANCE (private field). We could then more easily change it later to not return a singleton. Feel free to ignore or defer it, though.

chrisseaton:
CH3 resolved in 8967b831f179253b9bc8aa6828e652acf2ad0a89.

Can't test with JRuby until the iterate frames discussion is resolved.

chumer:
CH1, CH2, and CH3 resolved. Thanks. I am good with the change now.

woess:
Please have a look at the eclipse problem, though.

chrisseaton:
@woess I did see the problem you're describing before, but not after I fixed CH2. Tell me if you see it again.

woess:
Yes, it's still there (eclipse 4.4.2).
expected (after mx build):
```
    public DynamicObject createShapeBase(DynamicObjectFactory factory) {
        assert factory != null;
        CompilerAsserts.partialEvaluationConstant(factory);
        assert createsShapeBase(factory);
        return factory.newInstance();
    }
```
actual (after cleaning the object.dsl.test project in eclipse):
```
    public DynamicObject createShapeBase(
            DynamicObjectFactory factory,
            com.oracle.truffle.api.object.DynamicObjectFactory factory,
        assert factory != null;
        CompilerAsserts.partialEvaluationConstant(factory);
        assert createsShapeBase(factory);
        assert factory.getShape().hasProperty(FACTORY_IDENTIFIER);
        assert factory != null;
        return factory.newInstance(
            factory,
    }
```

chrisseaton:
Yes sorry it broke all the builders as well :( I'll revert.

PR #41, opened 2016-02-09
TruffleInstrument can provide services

jtulach:
Giving TruffleInstrument a chance to provide multiple services when its onCreate method is called.

### DIFF
--- a/truffle/com.oracle.truffle.api.instrumentation.test/src/com/oracle/truffle/api/instrumentation/examples/DebuggerController.java
+++ b/truffle/com.oracle.truffle.api.instrumentation.test/src/com/oracle/truffle/api/instrumentation/examples/DebuggerController.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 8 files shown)


jtulach:
Travis build failed, but in the parts of the code (SourceSection) that hasn't been touched by this change at all. Consider merging into graalvm:feature/instrumentation_api regardless of that.

jtulach:
@jtulach: If you are OK with the diff, I can merge
@chumer:  Really? onCreate returns an Object[]? I would have expected something in Env
@jtulach: return value from onCreate is automatically immutable. Env.registerService(Object... services) could be called multiple times. Our lookup mechanism is not designed to support dynamic changes.
@chumer:  I have a bad feeling for using a return value for optional api.  We could also enforce service registration only on startup dynamically as well right?

@jtulach: : One more option is to put the objects into (one of) constructor of TruffleInstrument: protected TruffleInstrument(Object... services) and one would
```java
public DebuggerExample() {
  super(new DebuggerControllerImpl());
}
```
@chumer: Not very happy with that either, because you should have access to Env. Otherwise you create your services in the constructor and then you need to pass in Env in an extra step.
@chumer: We could have a Env#registerService(Object) and it cannot be called after onCreate was invoked. We could set a flag in Env.

jtulach:
I've started from scratch to address @chumer requirements: https://github.com/jtulach/truffle/commit/ed2871fb0b37161d879ee8c400b1c6818b368bd2 - Christian is this what you wanted?

chumer:
Yes that was what I wanted :-) thanks

PR #43, opened 2016-02-12
use Objects.requireNonNull instead of obj.getClass for null checks

woess:
Use the preferable self-documenting `Objects.requireNonNull` method instead of `obj.getClass()` to null-check arguments.


smarr:
Looks good to me. 
Could you kick travis? Looks like the trigger got lost, the PR didn't get listed for a test run.

jtulach:
Feel free to integrate anytime.

smarr:
Hurray, and without merge commit. I like that! :+1:

PR #46, opened 2016-02-12
Putting the class into correct file, so the name of the class and nam…

jtulach:
…e of the file do match.


nirvdrum:
Was this a long-lived branch? The merge commit seems unnecessary.

dougxc:
Why do you think the merge is unnecessary @nirvdrum ? I asked @jtulach to make the merge since the PR had the text `"This branch is out-of-date with the base branch"` which means it cannot be merged.

nirvdrum:
I guess I don't know what our policy on rebasing and all that is, but the primary commit here doesn't seem to conflict with anything. It looks like if the branch were just made at a slightly later time, the merge wouldn't be required. I'm a bit surprised because I would have expected a `git pull` to do a fast-forward merge, which also would avoid the merge commit here.

Put another way, it looks like the merge is here because some other unrelated work happened to go on at the same time and was pushed before this commit. If that continues to happen, you'd just wind up with a series of merge commits really adding nothing to this branch.

dougxc:
I assume the merge was done on github via the `Update Branch` button and github doesn't do a fast-forward merge (for whatever reason).

smarr:
I would also strongly prefer that we avoid unnecessary merge commits, even if that means we can't use the shiny github buttons. Think, it makes following changes much easier.

And perhaps it is just me and I don't know what I am doing, but finding the commit that introduced a bug is also much easier when rebasing is used IMHO, instead of trying unravelling complex parallel histories with merges.

dougxc:
I agree with this and am working towards the workflow you describe. It's just a matter of getting a little more familiar with git.

jtulach:
I know how to avoid merges when I have not only a right to commit, but also a right to push. However with the review system we practice it is hard to avoid merges easily. We switched to Github to adopt github style - and that means merges - there is "update branch" button and not "rebase branch" one.

Moreover I don't see a reason for prevent merges. Git is distributed system and distributed systems handle merges as first class citizens. Merge has no overhead, it is just an entry in log, it can be easily filtered out.

Repository with merges [correctly reflects the situation during development](http://wiki.apidesign.org/wiki/Mercurial_vs._Subversion) - unlike rebase when the original state of commit is lost.

dougxc:
Which github style are you referring to @jtulach? I'm aware of both [git-flow](http://nvie.com/posts/a-successful-git-branching-model/) and [github-flow](http://scottchacon.com/2011/08/31/github-flow.html). In any case, our move to github was mainly about it being a good collaboration platform for our open source code compared to in-house systems, not so much about adopting a specific workflow.

I'm curious - what's the advantage of the commit history reflecting the chronological details of development? Or asked another, what important state is lost by rebasing?

woess:
> I know how to avoid merges when I have not only a right to commit, but also a right to push. However with the review system we practice it is hard to avoid merges easily. We switched to Github to adopt github style - and that means merges - there is "update branch" button and not "rebase branch" one.

But you *have* the right to push to the PR branch in your fork; it's not hard to avoid merges. In fact, our current CI setup requires it to be up to date with master so that it can be merged, so a manual step is required either way. It's unfortunate that github PR UI doesn't offer a rebase option (yet), otherwise it would be just a matter of clicking the right button.
There's no such thing as "github style". The choice is ours.

> Moreover I don't see a reason for preventing merges. [...] Merge has no overhead, it is just an entry in log [...] correctly reflects the situation during development [...]

That's still overhead and (more importantly) needless clutter as far as I am concerned.
There're actually two flavors of merges here:
1. the final merge of your branch into master.
2. merges to bring your branch up-to-date with master, as created by the update branch button.
I see no value of having the latter reflected in the commit history at all, and you might end up with a couple of superfluous merges for a simple non-conflicting change if your pull request isn't merged right away. The goal here really is to keep the history clean.

jtulach:
The goal is imho, to reflect history of development properly. If you want clean history, use:
```bash
$ git log --no-merges
```

woess:
What history of development? Updating your branch is hardly "development" and nobody really wants to see these merges reflected in history. Merges are useful on a shared branch but for the majority of our pull requests you won't have such a situation.

`git log --no-merges` does not solve anything. It doesn't and cannot differentiate between useful and useless merges, and the commits will appear in a seemingly arbitrary order with your commits and unrelated commits you merged in spliced together in all but trivial cases. Not to mention that most GUIs don't even have such a feature. So what you're stuck with in the end is a cluttered history.

jtulach:
"nobody really wants to see these merges reflected in history" - I have nothing against straightforward history, but I don't mind seeing merges either. I see their value and I am far for supporting (your?) position that (most of) merges in history should be avoided.

PR #49, opened 2016-02-14
Add a default value of an empty array to TypeSystem.

### LINKED ISSUE #48: Should TypeSystem#value have a default value?
Isn't it the case that the list of types used in the DSL type system is largely redundant now? So at least should it be given a default value of `{}` so that we can write:

```java
@TypeSystem
```

instead of

```java
@TypeSystem({})
```

chrisseaton:
Fixes #48.


jtulach:
No comment. I don't have enough experience in this area.

woess:
looks good to me

chumer:
Looks good. Feel free to rebase and merge.

PR #51, opened 2016-02-15
Remove Message.valueOf usage in generated Factory classes for interop

grimmerm:


grimmerm:
Yes, that is correct. accessMessage should not contain code that is already covered by the other methods (e.g. accessRead).
This method then will throw an UnsupportedmessageException in all cases, except when there are new/custom messages.

However, is it ok to compare hashCodes in accessMessage?

woess:
does that mean hash codes have to be unique?

grimmerm:
Yes, hashCodes need to be unique.
For KnownMessages `DelegatingFactory#accessMessage` already relies on that.

jtulach:
Comparing hashes is only possible for KnownMessage subclasses, imho. For anything else, it is not guaranteed to succeed.

grimmerm:
I've fixed the handling of unknown messages now.
Also, `accessMessage` now only contains cases, which were not covered by `Factory10`.
For example, the processor will create an `accessMessage` that looks as follows:
```
    public CallTarget accessMessage(Message unknown) {
      if (unknown instanceof com.oracle.truffle.dsl.processor.MyMessage) {
        return Truffle.getRuntime().createCallTarget(com.oracle.truffle.api.dsl.test.interop.AcceptMyMessage.createRoot(com.oracle.truffle.api.dsl.test.interop.TestTruffleLanguage.class));
      }
      throw UnsupportedMessageException.raise(unknown);
    }
```

jtulach:
Using instanceof is the right solution in current state of interop. I can merge the change.

I would feel more confident if there was a test to permanently verify that the change "improved something".

jtulach:
I fixed the compilation problem in https://github.com/jtulach/truffle/commit/2936a49c14b434a8c5464e69222098eeb167b7c9 - please review and continue from there - ecj still complains about something...

Thanks.

woess:
why was this merged despite the compilation errors in eclipse?

jtulach:
Yes, my observation yesterday was that there is some problem with AcceptMyMessage2 - but Matthias did something, the merge button become green, so somehow the travis had to succeed!?

woess:
@chumer may be able to help.

jtulach:
Fix of `AcceptMyMessage2` ecj problem available in #60

PR #58, opened 2016-02-17
DSL: Fix bug in error reporting which generated code even if the parsed node was invalid.

chumer:


woess:
is it possible to test this?

chumer:
This was happening in a super rare case on JDT only. I tried to reproduce it but failed. I could only reproduce in real code.

woess:
ok, looks good. but please rebase; I don't want two "Merge branch 'master' into dsl_fix" commits.

PR #59, opened 2016-02-17
Temporary fix for SourceSectionFilterTest execution for running inside graal classloaders

chumer:
Trying to solve this problem:
https://travis-ci.org/graalvm/graal-core/jobs/109844273

I could not solve it with:
@RunWith(SeparateClassloaderTestRunner.Theories.class)

Its the reason I cannot integrate:
https://github.com/graalvm/graal-core/pull/39


chumer:
Suggestions how else I could fix this?

woess:
truffle-api.jar is on the boot class path but the tests are not

jtulach:
Reflection is probably the most safe change to make the tests pass. Feel free to integrate once the formatting is OK.

PR #60, opened 2016-02-18
Add methods into the class generated in case of error to make the cod…

jtulach:
…e more compilable and help users to focus on the real error message


jtulach:
I'd like the assignee to be @grimmerm, please review my change.

jtulach:
I could verify the error previously via
```bash
$ mx clean
$ mx build --jdt-warning-as-error --jdt ~/bin/ecj-4.4.2.jar
```
after my change the error is gone and ecj compilation finishes without problems.

grimmerm:
I like the change.
The eclipse compilation error was caused by the missing `continue top` in L 175.
Also, generating abstract methods in case of an error makes sense as we can put `@Override` on the access messages consistently in the tests.

PR #65, opened 2016-02-19
CallTarget initialization must not throw NPE if PolyglotEngine is not initialized

woess:


smarr:
Hm, this raises for me the general question, do we support execution outside of PolyglotEngine?

woess:
That's a good question. In any case, it used to work so there was an incompatible API change in #1.

ldaynes:
Not everyone would want their Truffle Engine to run though the polyglot engine API. So I would ward off any changes that make Truffle unusable outside of polyglot engine.

jtulach:
Yes, Truffle execution without PolyglotEngine wrapper shall be supported. The fix is OK. Can we also have a test to make sure the fix never regresses?

chumer:
We should remove the InstrumentatinMode from the graal truffle PE tests. That these tests will crash when running without wrapper is broken.

jtulach:
The change falls into "documenting existing behaviour", so yes, we could integrate it. However the documentation is far from perfect. First of all I can't verify the code is compilable. Second, I believe the API classes in the code should hyperlinked.

mlvdv:
This sounds like a platform policy decision (new to me) that code run outside PolyglotEngine is un-instrumentable (which also means un-debuggable, un-profileable, etc. etc.).  Has the strategic implication of this been discussed more widely?

chumer:
Yes this has been discussed before. My latest policy update was the following: Yes the minimal Truffle example without instrumentation and polyglot engine, ie just using TruffleRuntime and CallTarget directly, must remain working and simple because Truffle is also used by non-dynamic-language use cases. All the nice instrumentation, debugging, polyglot engine stuff is allowed to not work for these use-cases.

chumer:
Thanks for the change.

jtulach:
Bugfixes in Truffle should be accompanied with a unit test, if it is possible to write one.

PR #68, opened 2016-02-21
Make old instrument API abstract methods in RubyLanguage concrete

chrisseaton:
Even if you aren't using the old `.instrument` API anymore you can't avoid it because of these three abstract methods in `TruffleLanguage`. This PR makes them concrete and just throws `UnsupportedOperationException` if they're used, which they won't be if you aren't using the API.

### DIFF
--- a/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
+++ b/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
@@ -228,7 +228,9 @@ public abstract class TruffleLanguage<C> {
      * Gets visualization services for language-specific information.
      */
     @Deprecated
-    protected abstract Visualizer getVisualizer();
+    protected Visualizer getVisualizer() {
+        return null;
+    }
 
     /**
      * Returns {@code true} for a node can be "instrumented" by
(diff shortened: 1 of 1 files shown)


smarr:
Guess, I am again hijacking an issue for something only somewhat related:
Should the old instrumentation API in its entirety be marked deprecated and scheduled for removal?

/cc @jtulach, @chumer

entlicher:
I'd prefer to return null from getVisualizer() instead of throwing an exception. Calling that method was valid until now. If you want to break existing code, you can remove the method as such with a similar result. Returning null will keep the API contract.

chrisseaton:
@entlicher that's fine by me. We could also return `false` from `isInstrumentable`.

entlicher:
@chrisseaton +1

chumer:
@smarr old instrumentation will get deprecated in the next release. But we wait with that until the languages have fully adopted the new API and we are sure there are no major issues. If we would make it deprecated now, most guest language would get through the gate without adopting the new API (because using deprecated API is fatal).

jtulach:
OK. Make the method non-abstract now. Deprecate them (together with rest of old instrumentation API) in early March.

chrisseaton:
Suggested modifications made. Waiting for a review by @mlvdv.

jtulach:
Reviewed by @chumer, @entlicher, @mlvdv, @jtulach. Approved. @chrisseaton, feel free to integrate tomorrow.

mlvdv:
Agreed.

thomaswue:
Looks good.

PR #69, opened 2016-02-22
Check API signatures against most recent API snapshot

jtulach:
I'd like the travis to download the most recent API snapshot and compare the signatures in the pull request with it. This way we can find out whether there are (signature) API changes in the PR or not.


jtulach:
@chrisseaton, please verify this is the right way to integrate with travis and if you agree merge the change. The check should never fail - it is there just to help us do manual inspection of PR's signature changes.

PR #70, opened 2016-02-22
Make loop counts reportable on Nodes instead of RootNodes; 

chumer:
Two improvements:
* Deprecated LoopCountReceiver that was never intended to be visible to guest languages and use the accessor pattern to call into the runtime instead. Tried to implement it in a compatible way, please verify.
* Move reportLoopCount from RootNode to Node to make the context(Node) of the reported loop count visible  to the Truffle runtime implementation. This is important to do a better OSR decision in the Graal runtime. Should be a compatible change.

### DIFF
--- a/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
+++ b/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
@@ -45,7 +45,6 @@ import com.oracle.truffle.api.TruffleLanguage.Env;
 import com.oracle.truffle.api.frame.MaterializedFrame;
 import com.oracle.truffle.api.instrument.Instrumenter;
 import com.oracle.truffle.api.instrument.Probe;
-import com.oracle.truffle.api.instrument.Visualizer;
 import com.oracle.truffle.api.instrument.WrapperNode;
 import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.api.nodes.RootNode;
@@ -98,12 +97,6 @@ public abstract class Accessor {
 
             @Override
             protected WrapperNode createWrapperNode(Node node) {
-                return null;
-            }
-
-            @SuppressWarnings("deprecation")
-            @Override
-            protected Visualizer getVisualizer() {
                 return null;
             }
 
@@ -317,8 +310,7 @@ public abstract class Accessor {
     private static Assumption oneVM = Truffle.getRuntime().createAssumption();
 
     @TruffleBoundary
-    @SuppressWarnings("unused")
-    protected Closeable executionStart(Object vm, int currentDepth, boolean debugger, Source s) {
+    protected Closeable executionStart(Object vm, @SuppressWarnings("unused") int currentDepth, boolean debugger, Source s) {
         CompilerAsserts.neverPartOfCompilation("do not call Accessor.executionStart from compiled code");
         Objects.requireNonNull(vm);
         final Object prev = CURRENT_VM.get();
@@ -418,6 +410,30 @@ public abstract class Accessor {
 
     protected String toString(TruffleLanguage<?> language, Env env, Object obj) {
         return API.toString(language, env, obj);
+    }
+
+    protected boolean supportsOnLoopCount() {
+        return false;
+    }
+
+    @SuppressWarnings("deprecation")
+    protected void onLoopCount(Node source, int count) {
+        // optimized calltarget is not existent on default runtimes
+        if (OPTIMIZEDCALLTARGET != null) {
+            if (OPTIMIZEDCALLTARGET.supportsOnLoopCount()) {
+                OPTIMIZEDCALLTARGET.onLoopCount(source, count);
+            } else {
+                // needs an additional compatibilty check so older graal runtimes
+                // still run with newer truffle versions
+                RootNode root = source.getRootNode();
+                if (root != null) {
+                    RootCallTarget target = root.getCallTarget();
+                    if (target instanceof com.oracle.truffle.api.LoopCountReceiver) {
+                        ((com.oracle.truffle.api.LoopCountReceiver) target).reportLoopCount(count);
+                    }
+                }
+            }
+        }
     }
 
     static <T extends TruffleLanguage<?>> T findLanguageByClass(Object vm, Class<T> languageClass) {
--- a/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
+++ b/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
@@ -600,6 +600,12 @@ public abstract class Node implements NodeInterface, Cloneable {
         protected void probeAST(RootNode rootNode) {
             super.probeAST(rootNode);
         }
+
+        @Override
+        protected void onLoopCount(Node source, int iterations) {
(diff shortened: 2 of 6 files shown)


chumer:
On a second though maybe reportLoopCount should be protected. 
@jtulach whats the best way of changing this in a compatible way?

woess:
@chumer Right now it's source (but not binary) compatible. Unless you're allowed to break backward compatibility, if you change reportLoopCount to protected, you'd have to override it in RootNode with a public reportLoopCount and remove the final from the former to still be compatible. I'd keep it public.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
```diff
@@ -416,6 +419,26 @@ protected String toString(TruffleLanguage<?> language, Env env, Object obj) {
...
+        // optimized calltarget is not existent on default runtimes
+        if (OPTIMIZEDCALLTARGET != null) {
+            if (OPTIMIZEDCALLTARGET.supportsOnLoopCount()) {
+                OPTIMIZEDCALLTARGET.onLoopCount(source, count);
+            } else {
+                // needs an additional compatibilty check so older graal runtimes
+                // still run with newer truffle versions
+                RootNode root = source.getRootNode();
+                if (root instanceof LoopCountReceiver) {
+                    root.reportLoopCount(count);
```
woess: I think this should be `getRootNode().getCallTarget() instanceof LoopCountReceiver`.

chumer: Yes! indeed!

chumer:
I decided I don't care about the protected visiblity in this changeset. can be done in a separate one.

woess:
looks good to me

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
```diff
@@ -420,6 +422,29 @@ protected String toString(TruffleLanguage<?> language, Env env, Object obj) {
         return API.toString(language, env, obj);
     }
 
+    protected boolean supportsOnLoopCount() {
+        return false;
+    }
+
+    protected void onLoopCount(Node source, int count) {
+        // optimized calltarget is not existent on default runtimes
+        if (OPTIMIZEDCALLTARGET != null) {
```
jtulach: I am not able to find out who provides OPTIMIZEDCALLTARGET instance, but I am starting to get scared that it is someone outside of Truffle repository. That isn't good idea as that makes Accessor part of API and we would have to evolve it compatibly to prevent linkage errors.

Using Accessor pattern for the loopcount reporting is good idea, just the way to communicate this with runtime as to be different.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
```diff
@@ -517,6 +517,17 @@ public String getLanguage() {
         return "";
     }
 
+    /**
+     * Reports the execution count of a loop that is a child of this node. The optimization
+     * heuristics can use the loop count to guide compilation and inlining.
+     *
+     * @count count the number iterations to report to the runtime system
+     */
+    public final void reportLoopCount(int iterations) {
```
jtulach: From my naive, external point of view this looks very something similar to Profile. Shouldn't the API look like the profiles one?

I know that Profile doesn't need any special support from the runtime and apparently loop count reporting needs it. That means internally these two things are probably very different, but from user point of view I don't see much difference in loop counting and branch probability profiling.

smarr: Well, technically, the branch profile needs runtime support to be useful, and that is provided by the runtime by knowing what `CompilerDirectives.injectBranchProbability(.)` is.

The difference from an implementation perspective is probably, that the `ConditionProfile` communicates information directly to the compiler, while the loop count is report to the more general runtime system which wants to decide earlier on when to compile stuff.

chumer: @jtulach I see the concern but I have no good idea how to merge these two concepts. Not every ConditionProfile needs to be reported only loops are interesting for the runtime to compile faster.

jtulach:
I am assignee, but I don't have enough context to evaluate the change. No tests, no usecase, the javadoc talks about runtime that "might have do something". There is the Accessor that is implemented elsewhere. Passing back to Christian.

chumer:
Its going to be tested in graal-core. With a follow-up changeset.
Reassigning to @woess.

@jtulach is the API change fine with you?

woess:
@chumer examples/tests that show how `reportLoopCount` is supposed to be used would indeed be good. Could you please add some?

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
woess: how about renaming it to CALLTARGET, with an AccessorCallTarget implementation in DefaultCallTarget and OptimizedCallTarget each?

chumer: We did not need the Accesor in the default runtime yet. But generally this is a good idea. Can we defer this into another changeset?

chumer:
@woess  agreed. will add next week.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
woess: sure

woess:
great, thanks!

jtulach:
Re: "@jtulach is the API change fine with you?" According to [contributing](https://github.com/graalvm/truffle/pull/44/files#diff-6a3371457528722a734f3c51d9238c13R43) document
1. the change *is compatible*
2. it is *not documented* sufficiently
3. it is *not tested*
4. by putting new public method into such central class as `Node` it may *negatively affect future evolution* 
I am not sure about point 4, as I don't understand the use-case (e.g. point 2). Not having point 2 and 3 prevents us to consider any alternatives. My favorite one currently would be:

```java
public final class LoopCounter {
  private LoopCounter() {}
  public static LoopCounter create(Node forNode);
  public void report(int count);
}
```

but I have no idea whether this can work or not. And whether additional class is appropriate to the frequency of usage of the *counter pattern*.

jtulach:
Christian has just explained the problem domain to me and based on that conversation I can say that static method in the `LoopNode` class:
```java
public class LoopNode {
  public static void reportLoopCount(Node anyNode, int count) {....}
}
```
looks like the best solution to me. It removes the need to use `LoopNode` for loops, yet it keeps locality of the API to `LoopNode`. 

The names used in the sample aren't a requirement, they are subject to @chumer choice...

chumer:
@jtulach 
1. the change is still compatible
2. the change in LoopNode#reportLoopCount is now documented with example
3. its now tested with LoopNodeTest, I also added some more tests.
4. by hiding the method in LoopNode it will not affect future evolution as much negatively

@woess I've added the example and tests you requested.

woess:
Looks good to me now.

jtulach:
OK.

PR #72, opened 2016-02-23
CallTarget initialization should not throw NPE when not using PolyglotEngine

woess:
Follow up to #65, this time for the Graal Truffle Runtime.


woess:
Added a test as suggested by @jtulach in #65.

chumer:
Thanks for the fix. Can you give me access to your fork? So next time I can rebase and integrate?

PR #73, opened 2016-02-23
Improve error message and exception when a missing type is specified …

thomaswue:
…in the @NodeInfo annotation.

### DIFF
--- a/truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/parser/NodeParser.java
+++ b/truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/parser/NodeParser.java
@@ -457,20 +457,21 @@ public class NodeParser extends AbstractParser<NodeData> {
                 String name = ElementUtils.firstLetterLowerCase(ElementUtils.getAnnotationValue(String.class, mirror, "name"));
                 TypeMirror type = ElementUtils.getAnnotationValue(TypeMirror.class, mirror, "type");
 
-                NodeFieldData field = new NodeFieldData(typeElement, mirror, new CodeVariableElement(type, name), true);
-                if (name.isEmpty()) {
-                    field.addError(ElementUtils.getAnnotationValue(mirror, "name"), "Field name cannot be empty.");
-                } else if (names.contains(name)) {
-                    field.addError(ElementUtils.getAnnotationValue(mirror, "name"), "Duplicate field name '%s'.", name);
+                if (type != null) {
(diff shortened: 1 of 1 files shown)


truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/parser/NodeParser.java
```diff
@@ -463,6 +463,11 @@ private NodeData parseNodeData(TypeElement templateType, List<TypeElement> typeH
                 } else if (names.contains(name)) {
                     field.addError(ElementUtils.getAnnotationValue(mirror, "name"), "Duplicate field name '%s'.", name);
                 }
+
+                if (type == null) {
+                    throw new AssertionError("The type specified in the NodeField annotation cannot be found. A possible cause could be a missing import statement for the type.");
```
mbrantner: This does not look like an error/invariant that should raise an AssertionError. It looks more like an error that is relevant for the user. Also, it would be nice to actually include name of the type in the error.

thomaswue: OK. Agreed that AssertionError is probably not the best choice here. Do you have a preference for an existing exception class from JDK or a custom exception class extending RuntimeException?

mbrantner: I think that there should be a custom exception. Truffle (and the DSL) should define a hierarchy of custom exceptions. 

BTW: the same applies to https://github.com/graalvm/truffle/blob/master/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/frame/FrameDescriptor.java#L134

mbrantner: Also, I suggest introducing a consistent style for error messages with respect to format, style, capitalization, punctuation and grammar. For example,

"%s: type used in NodeField annotation not found (missing import?)"

Separating the variables contained in messages would also allow for i18n.

thomaswue:
The error is now reported by the Java compiler and not by the annotation processor. This will ensure that the error message is the standard error message for unresolved types in Java. Here the output from a test run:
```
SLAddNode.java:66: error: cannot find symbol
@NodeField(name = "test", type = FrameSlot.class)
                                 ^
  symbol: class FrameSlot
1 error
```
@mbrantner let me know if this addresses your comments. Regarding better error messages for other parts of the code, please create separate issues.

mbrantner:
Great solution.

PR #75, opened 2016-02-24
Fix a possible overflow in the counting condition profile.

thomaswue:
Makes sure the sum of the counters of the true and false branch of a condition profile cannot overflow.


thomaswue:
@mbrantner fyi.

thomaswue:
@mbrantner fyi.

mbrantner:
Great, thanks!

chumer:
Looks good. Thanks for the fix.

PR #76, opened 2016-02-24
Make DefaultCallTarget final. 

thomaswue:
The class is not part of the API and should also not be extensible.


woess:
looks good to me.

jtulach:
Final is better than non-final and this class isn't part of any API. It is just implementation. Merging.

PR #79, opened 2016-02-25
Fixed missing stackframe if CallTarget#call was called without a call node with the default runtime;

chumer:
Fix TruffleRuntime#iterateStackTrace was not iterating the topmost frame;
Fixed missing stackframe if CallTarget#call was called without a call node with the default runtime;
Added stack frame tests.


chumer:
Here is the graal runtime change:
https://github.com/graalvm/graal-core/pull/81

thomaswue:
Looks great!

PR #84, opened 2016-02-29
Source section filter extension

chumer:
This should address https://github.com/graalvm/truffle/issues/80 with rootSourceSectionEquals.

Additionally I've added indexIn, indexNotIn, lineIn and lineNotIn with a set of ranges to allow arbitrary index and line based checks. 

@mickjordan please comment if this fits your needs.

### DIFF
--- a/truffle/com.oracle.truffle.api.instrumentation.test/src/com/oracle/truffle/api/instrumentation/SourceSectionFilterTest.java
+++ b/truffle/com.oracle.truffle.api.instrumentation.test/src/com/oracle/truffle/api/instrumentation/SourceSectionFilterTest.java
@@ -150,6 +150,61 @@ public class SourceSectionFilterTest {
         Assert.assertNotNull(SourceSectionFilter.newBuilder().lineIn(2, 2).build().toString());
     }
 
+    @Test
+    public void testLineNotIn() {
+        Source sampleSource = Source.fromText("line1\nline2\nline3\nline4", null);
+        SourceSection root = sampleSource.createSection(null, 0, 23);
+        Assert.assertFalse(isInstrumented(SourceSectionFilter.newBuilder().lineNotIn(IndexRange.byLength(2, 1)).build(),
+                        root, sampleSource.createSection(null, 6, 5, tags())));
+
+        Assert.assertFalse(isInstrumented(SourceSectionFilter.newBuilder().lineNotIn(IndexRange.byLength(1, 2)).build(),
+                        root, sampleSource.createSection(null, 2, 1, tags())));
+
+        Assert.assertTrue(isInstrumented(SourceSectionFilter.newBuilder().lineNotIn(IndexRange.byLength(1, 1)).build(),
+                        root, sampleSource.createSection(null, 6, 5, tags())));
+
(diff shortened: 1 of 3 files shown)


mickjordan:
This seems comprehensive; it's probably good to have the indexNotIn in addition to rootSourceSectionIs

jtulach:
Tested? Seems so.

Documented? Yes, but:
* missing `@param` and `@return` tags in many methods

Compatible? Yes.

Evolution: do we have use-cases for public `IndexRange.getXYZ` methods? I can't imagine any (**JT01**). This is *feed-in only* API and exposing getters may limit us in the future.

Consistency: the new API is mixing `Builder` pattern with factory pattern in `IndexRange`. The advantage of builder is *fluent* like usage. I would prefer to keep such style: 
* one option is to make `lineIn` and `indexIn` cumulative - e.g. be able to call them multiple times - that would remove the need for `IndexRange` altogether - however such approach may not be easy to merry with the `and/or` semantics and also `notIn` style.
* then there is *subbuilder* pattern: a method that returns a `RangeBuilder` which then returns back the original `Builder`. Example:
```java
builder.lineIn().
  range(0,10).range(20,30).range(40,50).anyOf().
indextIn().
  range(5, 15).range(25,35).range(45,45).noneOf(). 
build();
```
The `lineIn` and `indexIn` methods would return new builder `RangeBuilder` and the `anyOf` and `noneOf` methods would close the *subbuilder* and return back the original `Builder` so one can continue providing configuration for the *higher level*.

From my comments **JT01** is of higher importance, the rest are just thoughts and opinions.

chumer:
Will fix JT01.

Regarding consistency: I like the sub builder pattern idea. The problem is that everything is currently already built like canoncial boolean expressions (see isSourceSection(SourceSection...)). Multiple builder elements are always AND and each parameter is considered OR. The fluent builder mostly improves writeability but not readability. So regarding readability (which I care most) both versions fluent and parameter based are equivalent. For these reasons I will not change it. Good suggestion tough.

mlvdv:
Looks good  I agree with JT01, there could be implementations that forget the originally specified ranges but correctly implement what's specified.  Also agree that this particular application of _subbuilder_ would be suboptimal for client friendliness.

jtulach:
API changes seem to be as expected:
```
Added Classes
-------------

SourceSectionFilter$IndexRange

Added Nested Classes
--------------------

SourceSectionFilter: nested public final static SourceSectionFilter$IndexRange

Added Methods
-------------

public final SourceSectionFilter$Builder SourceSectionFilter$Builder.indexIn(SourceSectionFilter$IndexRange[])
public final SourceSectionFilter$Builder SourceSectionFilter$Builder.indexNotIn(SourceSectionFilter$IndexRange[])
public final SourceSectionFilter$Builder SourceSectionFilter$Builder.lineIn(SourceSectionFilter$IndexRange[])
public final SourceSectionFilter$Builder SourceSectionFilter$Builder.lineNotIn(SourceSectionFilter$IndexRange[])
public final SourceSectionFilter$Builder SourceSectionFilter$Builder.rootSourceSectionEquals(SourceSection[])
```

PR #85, opened 2016-02-29
Moving FindContextNode implementation next to PolyglotEngine so it ha…

jtulach:
…s access to advanced APIs including profiles

Next step is to do some value profiling to make sure the code re-optimizes when multiple polyglot engines are used in one JVM.


eregon:
Why is the whole conversation not done here? It would be much more practical to get background on this change.

truffle/com.oracle.truffle.tck/src/com/oracle/truffle/tck/TruffleTCK.java
```diff
@@ -995,33 +1023,38 @@ public void testPrimitiveIdentityForeignObject() throws Exception {
...
-                assert res instanceof Number : "expecting number: " + res + "\n" + log;
-                ++prev2;
-                assert ((Number) res).intValue() == prev2 : "expecting " + prev2 + " but was " + res + "\n" + log;
+        PolyglotEngine.Builder builder = PolyglotEngine.newBuilder().executor(Executors.newSingleThreadExecutor());
+        PolyglotEngine vm2 = prepareVM(builder);
+        try {
+            PolyglotEngine.Value count2 = vm2.findGlobalSymbol(countMethod);
+
+            PolyglotEngine.Value count1 = findGlobalSymbol(countMethod);
+            PolyglotEngine vm1 = tckVM;
```
eregon: How does it work here, since it seems both `vm1` and `vm2` are used from the same thread?

jtulach: Yes, it does work.

Although it was tough to find out how to make it work. Please note that *vm2* uses *executor* - which means that while the commands are send to *vm2* from the main thread, the actual execution happens in another one.

chumer:
timed out

PR #87, opened 2016-03-03
clear text map when updating code in ClientManagedFileSource

lukasstadler:
without this change, methods that use the text map (to query line numbers, etc.) may fail or produce wrong results after the source's contents are updated.


jtulach:
JT01: could a test demonstrating the problem be part of the fix?

lukasstadler:
I added some tests and fixed a second, related problem that showed up.
(Source.fromFileName did not update the contents if the file name is relative)

PR #94, opened 2016-03-04
Instrumentation javadoc

mlvdv:
New Instrumentation cleanup:
- Remove obsolete class names from Javadoc
- Realign variable names around correct distinctions and uniform terminology
- Rename an especially awkwardly named (non-public) class
- Rework some Javadoc for clarity, diction, and spelling


truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentationHandler.java
```diff
@@ -536,17 +538,18 @@ protected void visitInstrumented(Node node, SourceSection section) {
     }
 
     /**
-     * Instrumenter implementation for {@link TruffleInstrument}.
+     * Instrumenter implementation for use by {@linkplain TruffleInstrument external clients} of
+     * instrumentation.
      */
-    final class InstrumentationInstrumenter extends AbstractInstrumenter {
+    final class ClientInstrumenter extends AbstractInstrumenter {
```
chumer: I think the term Client is too generic. Is there something wrong with InstrumentInstrumenter. There is also a LanguageInstrumenter.

chumer:
Besides one minor remark, good changes as far as I can tell.

Thanks!

truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentationHandler.java
chumer: Its optional to change this.

mlvdv: I understand the internal consistency behind the name, but it unfortunately leads to a word that visually scans very poorly and is distracting.  Fortunately it is not public.  How about changing both names:  `LanguageClientInstrumenter` and `InstrumentClientInstrumenter`?

The source of the awkwardness is the name `Instrumenter`, which now that I've thought about it more should really be called something like `Binder`.  This would be much more descriptive, since it creates instances of `EventBinding`, but that's a breaking change.

chumer: If you think InstrumentClientInstrumenter is better than InstrumentInstrumenter so feel free to change it. I have no strong opinion. I also had no problem with InstrumentInstrumenter, given its an internal class.

smarr: As a second opinion: I'd prefer `InstrumentClientInstrumenter`. The first time I saw `InstrumentInstrumenter` I didn't really have an idea what it was.

mlvdv: Any opinions on an earlier thought: renaming both internal subclasses to` InstrumentClientInstrumenter` and `Language ClientInstrumenter`

chumer: Good with me. If that helps your understanding.

mlvdv:
@chumer please have a look at this merge; when I tried it I encountered build failures somewhere in the mx code for managing distributions, where there results of some DSL processing seems not to pass an assertion that looks like a versioning problem of some kind.

PR #95, opened 2016-03-04
Embedding code snippets

jtulach:
Once https://github.com/graalvm/mx/pull/65 is accepted, we can start using it. This is the rewrite of `{@codesnippet ...}` tags to `{@link ...Snippets...}` tags that can be easily navigated from the IDE. This will address problems reported in https://github.com/graalvm/truffle/issues/19


jtulach:
Ready for integration.

truffle/com.oracle.truffle.api.profiles/src/com/oracle/truffle/api/profiles/BranchProfile.java
```diff
@@ -131,3 +132,22 @@ public String toString() {
     }
 
 }
+
+class BranchProfileSnippets {
+    class Node {
+    }
```
chumer: Why don't you use the real Node?

jtulach:
True, in this case we can use real Node. Probably copy paste error. OK now?

truffle/com.oracle.truffle.api.vm/src/com/oracle/truffle/api/vm/PolyglotEngine.java
```diff
@@ -1285,3 +1285,21 @@ protected String toString(TruffleLanguage language, TruffleLanguage.Env env, Obj
...
+        // BEGIN: com.oracle.truffle.api.vm.PolyglotEngineSnippets#initializeWithParameters
+        String[] args = {"--kernel", "Kernel.som", "--instrument", "dyn-metrics"};
+        PolyglotEngine.Builder builder = PolyglotEngine.newBuilder();
+        builder.config(YourLang.MIME_TYPE, "CMD_ARGS", args);
+        PolyglotEngine vm = builder.build();
+        // END: com.oracle.truffle.api.vm.PolyglotEngineSnippets#initializeWithParameters
+        // @formatter:on
+        return vm;
+    }
+}
```
woess: No newline at the end of the file. How come checkstyle does not complain about it? Is it not enabled for the truffle.api.vm project?

woess: It really isn't checked by checkstyle. Please enable it (add the following line in suite.py).
`"checkstyle" : "com.oracle.truffle.api",`

jtulach: Good catch.

chumer:
Its a shame that eclipse cannot properly handle doclets. Having the snippets that close definitely makes it simpler to find the examples when browsing code. While this use-case still is not perfect (eg. clicking on the link will show you the javadoc of the snippet and not the code), I think its good enough now to be used. The formatted code in javadoc looks very good.

I've also verified that the snippets code does not appear in the distribution jars.

PR #96, opened 2016-03-04
Support java.util.Map in the interop

jtulach:
Java Interop already supports java.util.List - it is important for it to support Map as well. One of the usecases is to be able to obtain Map view of a language's global context: 

```java
lang1.getGlobalObject().as(Map.class)
```


jtulach:
Trivial Map <-> TruffleObject implementation. What do you think @grimmerm, OK? I am not very happy about the unsupport entrySet() method, but I cannot do anything unless we define Message.PROPERTIES to enumerate all keys...

Merge as it is or define the new message?

grimmerm:
I would very much like a message to get all properties. That is a good idea! It would also allow us to easily implement iterators for `TruffleObjects`.
`Message.PROPERTIES` could return a range for arrays and identifiers for objects.
+1 for this new message.

grimmerm:
Changes look good to me.

truffle/com.oracle.truffle.api.interop.java/src/com/oracle/truffle/api/interop/java/TruffleMap.java
```diff
@@ -0,0 +1,74 @@
...
+            return valueType.cast(ToJavaNode.message(Message.READ, obj, key));
+        } catch (InteropException e) {
+            throw new IllegalStateException(e);
+        }
+    }
+
+    @Override
+    public V put(K key, V value) {
+        keyType.cast(key);
+        valueType.cast(value);
```
grimmerm: This will throw a ClassCastException is the cast is not possible - but shouldn't this be a conversion (using the ToJavaNode)?
Same question for the cast in the `get` method.

truffle/com.oracle.truffle.api.interop.java/src/com/oracle/truffle/api/interop/java/TruffleMap.java
```diff
@@ -0,0 +1,74 @@
...
+        this.obj = obj;
+    }
+
+    static <K, V> Map<K, V> create(Class<K> keyType, Class<V> valueType, TruffleObject foreignObject) {
+        return new TruffleMap<>(keyType, valueType, foreignObject);
+    }
+
+    @Override
+    public Set<Entry<K, V>> entrySet() {
+        throw new UnsupportedOperationException();
```
grimmerm: Can't we implement this now using `Message.PROPERTIES `

truffle/com.oracle.truffle.api.interop.java/src/com/oracle/truffle/api/interop/java/TruffleMap.java
jtulach: I think you are right. The current state isn't ideal.

Can you help me write a test that would fail with current code? I would feel way more confident if the changes were supported by real use-cases.

jtulach: Btw. I should point out that `TruffleList` doesn't have any conversions either:

```java
public T get(int index) {
    return type.cast(ToJavaNode.message(Message.READ, array, index));
}
```
e.g. if there is a bug in `TruffleMap`, it is very likely in `TruffleList` as well. We need the sample use-cases to see what is failing (if anything).

chumer:
Assume the following js code:
```
foo = {}
foo.bar = {}
foo.bar.baz = {}
```

```
Value foo = ... // get  foo somehow
Map m = foo.as(Map.class);
```
Should I be able to do?
`m.get("foo").get("bar").get("baz") `

or something like:
`m.get("foo").as(Map.class).get("bar").as(Map.class).get("baz") `?

Can we do this already? How would it look like?
Whats the longer term perspective?

truffle/com.oracle.truffle.api.interop.java/src/com/oracle/truffle/api/interop/java/TruffleMap.java
jtulach: The https://github.com/graalvm/truffle/pull/96/commits/650f1aabae65985033c261da8b37f7671bbc7c1b change shows that the conversion is already happening.

jtulach:
I was trying to type Christian's example, but all I can say is that it seems easier to traverse the object with raw `PolyglotEngine.Value`:
```java
Value foo = ... // get  foo somehow
Value baz = foo.get("bar").get("baz");
baz.// do something with baz
```
that would require 
```java
class Value {
  public Value get(String fieldName);
}
````
method in `PolyglotEngine.Value`. 

I don't see how `Map` could make it easy unless you type it recursively as `typedef X = Map<String,X>` and I don't think it is possible. One can of course type it as `Map<String,Object>` and change the system to automatically wrap every `TruffleObject` into yet another `Map`, but then your example requires casts:
```java
baz = ((Map<String,Object>)((Map<String,Object>)root.get("foo")).get("bar")).get("baz");
```
This looks complicated.

Maybe we should separate two usecases. The unsafe one (every access/invocation is using a string): use `PolyglotEngine.Value`. The safe one (one describes structure of objects via interface first): use `JavaInterop`. The proposed interop `Map` is useful for access to named terminal leaf values in the safe access approach.

woess:
The `PROPERTIES` message lacks javadoc, and it's unclear to me what it should do for objects that aren't maps or map-like objects (e.g., @grimmerm suggested it could return the indices for arrays).

Naming: maybe consider renaming `PROPERTIES` to `KEYS` (cf. `Map#keySet`).

PR #97, opened 2016-03-04
Om dsl

chrisseaton:
Sorry this is a repeat of #40, which failed on other test systems. It turns out that my implementation was dependent on the order in which the compiler passed methods to my annotation processor. I've fixed that.

### DIFF
--- a/truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Layout.java
+++ b/truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Layout.java
@@ -1,0 +1,424 @@
+/*
+ * Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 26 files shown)


truffle/com.oracle.truffle.api.object.dsl/src/com/oracle/truffle/api/object/dsl/Layout.java
```diff
@@ -0,0 +1,424 @@
...
+ *
+ * <h2>Custom Identifiers</h2>
+ *
+ * By default, internal {@link HiddenKey} identifiers with descriptive names will be created for
+ * your properties automatically. You can also specify the identifiers to use by defining them as
+ * public static final fields in the interface.
+ *
+ * {@codesnippet customid}
+ *
+ * @since 0.8 or earlier
```
woess: should be `@since 0.12` (+2 other occurrences)

chrisseaton: Ah yes of course - thanks - fixed.

woess:
Nitpicking: the commits and their order are a bit confusing. I would have expected a 1 commit that is the "unrevert" followed by another one that is the bug fix.

chrisseaton:
@woess done

woess:
thanks

PR #98, opened 2016-03-05
KillException and QuitException handling

mlvdv:
Clarify and document required behavior for WrapperNodes in the instrumentation framework. Fix a related bug in ProbeNode.  Add tests for correct handling.
@chumer please have a look at tests failing for automatically generated wrappers


chumer:
It was intentional that KillException and QuitException was not supported for normal instruments only for language instruments.

mlvdv:
Those exceptions are required by the Debugger, which is not a language instrument and which is now broken.

@chumer suggests replacing the exceptions and I agree; see new issue #109.

Meanwhile, this patch is needed to keep the current debugger working correctly until a replacement is in place.

I could simplify the current situation by removing any dependence on` QuitException`.  Only the functionality of `KillException` is actually required. Is that worth doing?

chumer:
I cannot really come up with a better solution.
So lets only support KillException (CH1) and deprecate QuitException (CH2).

truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/ProbeNode.java
```diff
@@ -280,6 +282,8 @@ final void onDispose(EventContext context, VirtualFrame frame) {
         final void onEnter(EventContext context, VirtualFrame frame) {
```
chumer: CH3) onDispose should also support KillException. Any reason not to?

truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/ProbeNode.java
```diff
@@ -280,6 +282,8 @@ final void onDispose(EventContext context, VirtualFrame frame) {
...
             try {
                 innerOnEnter(context, frame);
+            } catch (KillException | QuitException ex) {
+                throw ex;
             } catch (Throwable t) {
```
chumer: Can we implement it like this? I don't think it makes every sense to have Kill on the fastpath. Also we should merge the control flow edge with other exceptions in case of langauge bindings.

```
 static void failEventForInstrumentation(EventBinding<?> b, String eventName, Throwable t) {
        if (t instanceof KillException) {
            throw (KillException) t;
        }
  ...
}

```

truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentableFactory.java
```diff
@@ -27,49 +27,80 @@
...
+     * onReturnExceptional(Frame,Throwable)}: an <em>execute</em> method on the delegate has just
+     * thrown an exception.</li>
+     * </ul>
+     * </p>
      * <p>
-     * Returns a new, never adopted, unshared instrumentable {@link WrapperNode wrapper} node
-     * implementation for a particular node of the guest language AST. The returned wrapper
-     * implementation must extend the same type that is annotated with {@link Instrumentable} and
-     * forward all events to the given {@link ProbeNode probe}.
+     * A wrapper must pass through a {@link KillException} or {@link QuitException} without any
```
chumer: CH5) If that is true we need to change the wrapper generation as well.
Currently the generated wrapper pattern looks like this

```
 @Override
        public Object executeGeneric(VirtualFrame frame) {
            try {
                probeNode.onEnter(frame);
                Object returnValue = delegateNode.executeGeneric(frame);
                probeNode.onReturnValue(frame, returnValue);
                return returnValue;
            } catch (Throwable t) {
                probeNode.onReturnExceptional(frame, t);
                throw t;
            }
        }
```

Why is it important that onReturnExceptional is not invoked in case of KillException? I'd say that onReturnValue or onReturnExceptional is **always** invoked even if a KillException was thrown. This would make this specification simpler, because the wrapper can treat any exception the same way.

truffle/com.oracle.truffle.api.instrumentation.test/src/com/oracle/truffle/api/instrumentation/InstrumentationTest.java
```diff
@@ -667,4 +669,94 @@ public void onReturnValue(EventContext context, VirtualFrame frame, Object resul
...
+    }
+
+    @Test
+    public void testQuitExceptionOnEnter() throws IOException {
+        engine.getInstruments().get("testKillQuitException").setEnabled(true);
+        TestKillQuitException.exceptionOnEnter = new QuitException();
+        TestKillQuitException.exceptionOnReturnValue = null;
+        TestKillQuitException.returnExceptionalCount = 0;
+        try {
+            run("STATEMENT");
```
chumer: CH6) It seems that if we throw the KillExcepion all through PolyglotEngine, we need to adapt either the documentation in PolyglotEngine. (Which is an incompatible change)
I'd prefer if we wrap the KillException into an IOException inside PolyglotEngine.

chumer: CH6 all ways of evaluating code in PolyglotEngine require a test.

chumer:
@mlvdv  Please address CH1-6.

truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentableFactory.java
woess: +1

mlvdv:
CH1-2) Already suggest this possibility, so agree.
CH3) Not clear to me this would occur, but for consistency agree
CH4) Agree, as long as semantically equivalent.
CH5) Yes, fixing wrapper generation is what I was requesting.  No, a Kill exception must not generate an instrumentation event.  Kill means that guest language execution terminates immediately; a subsequent guest language execution event would be meaningless and misleading.
CH6) Either way would work for me, but this makes several of the above moot so there's no point acting on them yet. 

MLVDV1)  Is there a pattern already documented for what IOException means and how it should be used?  Should it hold referred exceptions as you suggest, but why not subclassing? Is such a discussion out of scope for this PR, which is a bug fix on existing code?

woess:
> CH5) Yes, fixing wrapper generation is what I was requesting. No, a Kill exception must not generate an instrumentation event.

Isn't this something onReturnExceptional can and should handle? I'd rather have the logic in the handler, and the wrapper as simple as possible.

chumer:
+1 with @woess . Having it in the wrapper would propagate complexity to all languages.

smarr:
Perhaps I am missing something, but I'd actually argue the other way.

If there is an exit/quit 'event', why do I need to handle that in my instrumentation?

This seems to be something that should be handled by the framework/wrappers. I think, I only want language-specific events in the onReturnExceptional(). At least for my current tool, these seems to be the ones I care about. Having to handle exit/quit in tool-specific nodes also seems to ensure that the tools break, because until the recent discussions, I wasn't even aware of these exceptions.

woess:
@smarr: this exception should be handled in the framework (inside ProbeNode) and not propagated to the actual instrumentation observers' onReturnExceptional methods. This way we don't duplicate the catch/rethrow (there are multiple wrapper classes (at least one per language) but only 1 ProbeNode) and effectively make it an implementation detail *here*.

smarr:
Ah, right, forgot about the `ProbeNode`, yes, that sounds like the right place. Sorry for the noise.

chumer:
> MLVDV1) Is there a pattern already documented for what IOException means and how it should be used? Should it hold referred exceptions as you suggest, but why not subclassing? Is such a discussion out of scope for this PR, which is a bug fix on existing code?

To answer your question: yes it is out of scope of this PR. Lets discuss it here: https://github.com/graalvm/truffle/issues/116

chumer:
I don't agree to 3. and 4.. It makes sense to have the option to cleanly dispose any open resources one might have opened because of a started event. How would you do that in case of a kill event?

5. what help do you need with the tests? Your list looks already like the perfect specification.

mlvdv:
> I don't agree to 3. and 4.

- Your objection to 3 is easily addressed.  A language _implementation_  is free to catch a `KillException` and do what it needs as long as it (1) performs no more guest language execution, (2) generates no more execution event notifications, and (3) re-throws the exception when finished.  These are the Debugger's requirements at this time.
- Your objection to 4 can only be based on a misunderstanding.  From the perspective of a client-generated `KillException` the event notification method `onReturnExceptional()` is exactly the same as the other two event notification methods and must behave exactly the same way to such an exception. 

> what help do you need with the tests? Your list looks already like the perfect specification.

- I  don't believe I need any help with tests; I have included a test that correctly identifies a bug in WrapperNode generation.
- What I am requesting is a fix for the bug in WrapperNode auto-generation. 
- My fix for a similar bug in the new `ProbeNode` implementation is included.

The Debugger continues to be broken (9 days now) until these two fixes are in place.

woess:
@mlvdv I think there's a misunderstanding here about what it means to handle KillException in ProbeNode.onReturnExceptional. Just because the WrapperNodes call this method, doesn't mean it has to generate an actual return exceptional event (i.e., notify event handlers). So all your 3 points can be addressed in this way. FWIW, the old implementation used to it this way (i.e., catch and rethrow KillException in ProbeNode.returnExceptional).

I'd like to hide the existence of KillException from the guest language as much as possible. Ideally, it would be just an implementation detail of the instrumentation framework and PolyglotEngine. But at least we should avoid duplication of the catch-rethrow logic and leaking it to the guest language (which wrapper nodes are still part of, even when they're generated). It'll also give us more flexibility to do changes around it in the future.

entlicher:
I just got an idea to use ThreadDeath for that purpose.
If KillException extends ThreadDeath, it does not have to be in the APIs.
Language implementors should be advised to let the ThreadDeath propagate and we can catch it at any time as a KillException, which differentiates it from the real ThreadDeath.
Thoughts?

entlicher:
The Javadoc to ThreadDeath says: "An application should catch instances of this class only if it must clean up after being terminated asynchronously. If ThreadDeath is caught by a method, it is important that it be rethrown so that the thread actually dies."
Therefore it's a good practice to rethrow it already, we just can take the advantage of that.

mlvdv:
>CH3) onDispose should also support KillException. Any reason not to?

No reason not to; fixed.

mlvdv:
>I'd like to hide the existence of KillException from the guest language as much as possible. Ideally, it would be just an implementation detail of the instrumentation framework and PolyglotEngine. But at least we should avoid duplication of the catch-rethrow logic and leaking it to the guest language (which wrapper nodes are still part of, even when they're generated). It'll also give us more flexibility to do changes around it in the future.

@woess Can this part of the discussion be forwarded to #109 or #116?

woess:
@mlvdv ok

mlvdv:
Apologies to @woess and others.  My persistent misinterpretation of an earlier comment kept me from understanding the follow-up until his most recent comment caused me to look back at the code more closely.  I withdraw the request for changes to auto-generation of wrappers, and I'll clean up the handling of the exceptions.

chumer:
KillException is still not specified in PolyglotEngine.
Given that KillException was thrown before by PoylgotEngine we can go along I think. We need to do a bigger change with https://github.com/graalvm/truffle/issues/116 so we can probably also specify this properly with that change.

mlvdv:
>We need to do a bigger change with #116 so we can probably also specify this properly with that change.

Agreed.

PR #101, opened 2016-03-05
Test downstream users (JRuby to start with) of Truffle

chrisseaton:
This adds a new command `mx testdownstream` and runs that on Travis as an allowed-failure (it doesn't fail the overall commit if it fails, but you still see the result). Only JRuby so far, and runs 64k of JRuby's fastest tests so it only takes a couple of minutes.

This allows you to see if your changes are immediately compatible with JRuby. For example it would have caught the problems with the `iterateFrames` changes.

Potential problems are if the JRuby build is broken anyway (it very rarely is, but sometimes), or if people ignore it because it's allowed-failure.

@mickjordan am I using mx correctly here?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,7 @@ env:
   - TEST_COMMAND='mx/mx canonicalizeprojects'
   - TEST_COMMAND='mx/mx checkstyle'
   - TEST_COMMAND='sh .travis.sigtest.sh'
+  - TEST_COMMAND='mx/mx testdownstream'
+matrix:
+  allow_failures:
+    - env: TEST_COMMAND='mx/mx testdownstream'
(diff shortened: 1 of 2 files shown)


chrisseaton:
Example passing log https://s3.amazonaws.com/archive.travis-ci.org/jobs/113954147/log.txt

chumer:
I think this makes sense to try. If we get problems getting changes in because of these tests, but not caused by a Truffle problem, we can consider disabling this again.

chrisseaton:
You should never get problems integrating changes as the configuration means you can still push if it fails. It's just informational.

jtulach:
I would prefer if downstream projects "listen on" pull requests for Truffle and run the tests themselves and then annotated the PR with their results. It would scale better. On the other hand I understand that it is harder to implement such system from an infrastructure point of view.

This is good start.

PR #102, opened 2016-03-05
Ignore dists/truffle-mx.jar

chrisseaton:
This file is created by `mx maven-install`. I frequently have git confused by its presence.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ workingsets.xml
 findbugs.results
 .metadata/
 /eclipseformat.backup.zip
+dists/truffle-mx.jar


woess:
This is the only jar that ends up in `dists/`, isn't it? All the other jars seem to be in `mxbuild/dists/`. Perhaps this is a bug in mx? Removing the accept label until this is clarified.

jtulach:
Nine days without any clarification.

woess:
It should be fixed in mx.

jtulach:
Amazing. Glad to hear that.

woess:
I've created https://github.com/graalvm/mx/issues/89 to track this issue.

PR #105, opened 2016-03-07
Fix double-checked locking patterns

eregon:
* Only publish the DefaultCallTarget as initialized when it is.
* Only publish the InstrumentationHandler as initialized when it is.
* `intialized` needs to be volatile or the store to it might move above the actual initialization (compiler or CPU reordering).


smarr:
Random question: is there a reason why the `synchronized(this){}` might be preferable to a `synchronized` keyword on the method? Saw that in https://github.com/graalvm/graal-core/pull/102 as well.

eregon:
My personal preference would also be an explicit `synchronized ()` like this, because the function modifier is very easy to miss when reading the code.

woess:
+1 for explicit `synchronized`

chumer:
I have no strong opinion on the synchronized block issue. But I know that synchronized in a method has less footprint in terms of bytecodes.

PR #107, opened 2016-03-07
added TRUFFLE_TEST distribution

dougxc:


rschatz:
lgtm

mx.truffle/mx_truffle.py
```diff
@@ -93,3 +94,58 @@ def _truffle_gate_runner(args, tasks):
...
+            lines = []
+            counter = 1
+            for enum in sorted(properties.viewkeys()):
+                assert enum.startswith(metainfFile)
+                newEnum = metainfFile + str(counter)
+                counter += 1
+                for prop in properties[enum]:
+                    lines.append(newEnum + prop)
+
+            content = '\n'.join(lines)
```
chumer: Please use os.linesep instead of '\n'

rschatz: Really? That means the content of the jar file will be different depending on the platform where it was built. Isn't the jar file supposed to be platform independent?

(but probably it doesn't make any difference, the code that reads it should be robust for different line endings anyway...)

dougxc: `Properties.store0` uses the platform dependent line separator. On the other hand, when loading a properties file, Java handles both. In any case, I will implement your request Christian.

mx.truffle/mx_truffle.py
```diff
@@ -93,3 +94,58 @@ def _truffle_gate_runner(args, tasks):
...
+    def __opened__(self, arc, srcArc, services):
+        self.settings = {}
+        self.arc = arc
+
+    def __add__(self, arcname, contents):
+        metainfFile = self._truffle_metainf_file(arcname)
+        if metainfFile:
+            properties = self.settings.setdefault(metainfFile, {})
+            for line in contents.strip().split(os.linesep):
+                if not line.startswith('#'):
```
woess: I think we shouldn't rely on the file to have windows line endings on windows.

dougxc: Good catch - will fix.

PR #110, opened 2016-03-07
REPL debugger:  reflective access to SourceSection tags

mlvdv:
The REPL command line debugger supports features that are useful for working on language implementations and the debugger itself; this require access to *all* tags at each program location.  New instrumentation code does not make this available, so per @chumer suggestion the REPL now uses reflection.


ldaynes:
Can you hide use of reflection behind an internal interface ? This will 
make it easier for SVM to
substitute unsupported reflexive access with something else.

chumer:
I think we misunderstood. I did not suggest to commit this.

I recommended iterating over a list of known tags and check whether they are available. There is going to be away to get a list of all tags in the system (not for a particular source section) in the end, but its not landed yet. I'd suggest to postpone this or check for a fixed set of tags in your debugger with hasTags.

PR #111, opened 2016-03-08
Debugging support:  TruffleLanguage.evalInContext() reports when unsupported

mlvdv:
Usability improvement:  Languages not supporting eval() return an UnsupportedOperationException so that the debugger can distinguish between this and an ordinary failure to avoid user confusion.

### DIFF
--- a/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
+++ b/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
@@ -102,7 +102,7 @@ public abstract class Accessor {
 
             @Override
             protected Object evalInContext(Source source, Node node, MaterializedFrame mFrame) throws IOException {
-                return null;
+                throw new UnsupportedOperationException();
             }
         };
         lng.hashCode();
@@ -163,6 +163,16 @@ public abstract class Accessor {
         return API.eval(l, s, cache);
     }
 
+    /**
+     * @param vm
+     * @param ev
+     * @param code text to evaluate
+     * @param node node in halted context
+     * @param frame frame in halted context
+     * @return result of language evaluation
+     * @throws IOException
+     * @throws UnsupportedOperationException if not supported by the language implementation
+     */
     protected Object evalInContext(Object vm, Object ev, String code, Node node, MaterializedFrame frame) throws IOException {
         return API.evalInContext(vm, ev, code, node, frame);
     }
--- a/truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/Debugger.java
+++ b/truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/Debugger.java
@@ -62,7 +62,7 @@ import com.oracle.truffle.api.vm.PolyglotEngine;
  * Instance of this class is delivered via {@link SuspendedEvent#getDebugger()} and
  * {@link ExecutionEvent#getDebugger()} events, once {@link com.oracle.truffle.api.debug debugging
  * is turned on}.
- * 
+ *
  * @since 0.9
  */
 public final class Debugger {
@@ -71,7 +71,7 @@ public final class Debugger {
      * A {@link SourceSection#withTags(java.lang.String...) tag} used to mark program locations
      * where ordinary stepping should halt. The debugger will halt just <em>before</em> a code
      * location is executed that is marked with this tag.
-     * 
+     *
      * @since 0.9
      */
     public static final String HALT_TAG = "debug-HALT";
@@ -232,7 +232,7 @@ public final class Debugger {
     /**
      * Gets all existing breakpoints, whatever their status, in natural sorted order. Modification
      * save.
-     * 
+     *
      * @since 0.9
      */
     @TruffleBoundary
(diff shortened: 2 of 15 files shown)


truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
```diff
@@ -163,6 +163,16 @@ protected Object eval(TruffleLanguage<?> l, Source s, Map<Source, CallTarget> ca
...
+     * @param vm
+     * @param ev
+     * @param code text to evaluate
+     * @param node node in halted context
+     * @param frame frame in halted context
+     * @return result of language evaluation
+     * @throws IOException
+     * @throws UnsupportedOperationException if not supported by the language implementation
+     */
     protected Object evalInContext(Object vm, Object ev, String code, Node node, MaterializedFrame frame) throws IOException {
```
woess: this is not an API method and doesn't need javadoc

chumer:
I am not sure this is the right choice. I did not find other places where we used UnsupportedOperationException to indicate that something is not supported. The alternative would be to add boolean isEvalInContextSupported(). Also this change is not compatible, where it was previously IOException indicating that it was not supported, its now UnsupportedOperationException.

Couldn't we just improve the user message on the thrown IOException instead?

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
mlvdv: Such methods, however, can be very confusing because of the way the Accessor pattern works.

"doesn't need" ==? "must not have "

mlvdv:
Are you suggesting a protocol that involves parsing strings?  This sort of thing has been discouraged in other contexts.

Or are you suggesting that each language implementor should have sole responsible for the usability of debugger behavior in this situation, leading of course to different usability for different languages?

The fact that there are no other places where we would use this does not mean that there will be none in the future; it is a pretty basic concept to say that there are things some languages do not support.

	mlvdv

chumer:
CH1) This change is incompatible
CH2) The behavior is untested
CH3) Why not boolean isEvalInContextSupported? Or we could make this feature an optional service that can get registered by the language? Something like a InPlaceEvaluator service. I think we should work towards retiring SuspendedEvent and ExecutionEvent in favor of directly interacting with the Debugger class instead. For that we need to extend EventContext to support eval in place.

mlvdv:
CH1) yes it is
CH2) agree to add tests, but see CH3)
CH3) Would agree to most of these; pick one?

MLVDV1) "work towards" out of scope?

chumer:
CH1,CH3)
Incompatibility is a problem. Suggest one that is compatible.

MLVDV1) "work towards" out of scope?
Yes. Lets discuss this when we talk next time. What I wanted to say is that we should not invest too much time into this but instead try to evolve Debugger so it can work without SuspendedEvent and ExecutionEvent. Maybe collect whats missing in the instrumentation for that?

jtulach:
Can't we rather invest in fixing SL to support some form of eval?

```diff
truffle$ git diff
diff --git a/truffle/com.oracle.truffle.sl/src/com/oracle/truffle/sl/SLLanguage.java b/truffle/com.oracle.truffle.sl/src/com/oracle/truffle/sl/SLLanguage.java
index 438849f..ce754ff 100644
--- a/truffle/com.oracle.truffle.sl/src/com/oracle/truffle/sl/SLLanguage.java
+++ b/truffle/com.oracle.truffle.sl/src/com/oracle/truffle/sl/SLLanguage.java
@@ -58,6 +58,7 @@ import com.oracle.truffle.api.Truffle;
 import com.oracle.truffle.api.TruffleLanguage;
 import com.oracle.truffle.api.dsl.NodeFactory;
 import com.oracle.truffle.api.dsl.UnsupportedSpecializationException;
+import com.oracle.truffle.api.frame.FrameSlot;
 import com.oracle.truffle.api.frame.MaterializedFrame;
 import com.oracle.truffle.api.frame.VirtualFrame;
 import com.oracle.truffle.api.instrument.Visualizer;
@@ -510,7 +511,12 @@ public final class SLLanguage extends TruffleLanguage<SLContext> {
 
     @Override
     protected Object evalInContext(Source source, Node node, MaterializedFrame mFrame) throws IOException {
-        throw new IllegalStateException("evalInContext not supported in SL");
+        for (FrameSlot slot : mFrame.getFrameDescriptor().getSlots()) {
+            if (source.getCode().equals(slot.getIdentifier())) {
+                return mFrame.getValue(slot);
+            }
+        }
+        throw new IOException("Unknown local variable " + source.getCode());
     }
 
     public Node createFindContextNode0() {
```

That is enough for a toy debugger and certainly better than complicating API because we fear to touch SL.

Btw. Ideally this test would be in `TruffleTCK`.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/Accessor.java
woess: No, but it seems redundant (if you change the API method's javadoc, you'd have to change it in two places), and the `@param vm` and `@param ev` parts are useless without description. Alternatively, you could put a `@see APIClass#documentedMethod` there instead.

jtulach: I think I have an idea how to clean up the accessor pattern a bit and reduce the number of indirections. If you keep remind me, I'll do it at the end. What @kamosh thinks about that?

PR #112, opened 2016-03-09
Ignore JavaInteropSpeedTest.

zapster:
The test tries to find performance regressions in the java interop implementation by comparing invocation time which is highly system dependent and fails regularly on different machines. Doing this in a unit test is sub-optimal. It should be converted into a microbenchmark.

### DIFF
--- a/truffle/com.oracle.truffle.api.interop.java.test/src/com/oracle/truffle/api/interop/java/test/JavaInteropSpeedTest.java
+++ b/truffle/com.oracle.truffle.api.interop.java.test/src/com/oracle/truffle/api/interop/java/test/JavaInteropSpeedTest.java
@@ -31,11 +31,13 @@ import java.util.Random;
 
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import com.oracle.truffle.api.interop.TruffleObject;
 import com.oracle.truffle.api.interop.java.JavaInterop;
 
+@Ignore("Convert into a microbenchmark!")
 public class JavaInteropSpeedTest {
     private static final int REPEAT = 10000;
     private static int[] arr;


jtulach:
Is there anyone who knows how to "Convert into a microbenchmark!"? Who will be responsible for that?

zapster:
@jtulach I can do that (actually I have already done it and I wanted to add it to this PR but then we stumbled over #113). But there are two question: First, the benchmark makes only sense if executed with the graal truffle implementation, right? If so I would put the benchmark inside the graal source tree. The infrastructure is already there. (If we put it into truffle we need to add a JMH dependency.) Second and even more important, who is going to monitor the benchmark? The unittest variant was flaky for weeks and it turned out that it was not doing anything (because of #113).

jtulach:
As far as the monitoring goes, I'd suggest to follow work done in https://github.com/graalvm/truffle/pull/101 - it allows downstream projects to run the test they need and still provide feedback to each truffle pull request.

zapster:
Sounds promising! But we need to keep in mind that this is not a test that says OK or NG. It is a performance number. Deciding automatically whether there is a regression is not trivial IMHO.
But back to question 1: so the bench should be in graal, right?

jtulach:
The alternative is to have the test code in the truffle repository and only execute it if the [optimizing runtime](https://github.com/graalvm/truffle/issues/47) is present. Plus configure travis to get the latest binary of graal-core and run `mx unittest` against it.

zapster:
Well, it is not a unittest, it is an `mx microbech`. I'm also fine with having it in truffle (although I just opened graalvm/graal-core#126) but this means we pull the JMH dependency into truffle. Would this be a problem @jtulach?

jtulach:
Truffle repository contains various distributions. Adding the JMH dependency to `truffle-api` would be a problem, but if you only add it to `truffle-test`, it should be OK, as far as I can tell.

PR #115, opened 2016-03-09
use '\n' instead of os.linesep when splitting properties file into lines

dougxc:


woess:
we don't have to strip the lines (of `\r`), do we? then this LGTM.

in `__closing__`:
```python
content = os.linesep.join(lines)
self.arc.zf.writestr(arcname, content + '\n')
```
we're potentially mixing different line separators here

dougxc:
@woess "we don't have to strip the lines (of \r), do we" - not really since `java.util.Properties` strips out trailing `\r`'s when loading property files. However, we should be consistent so I'll make that fix and merge it.

PR #117, opened 2016-03-10
Fix Truffle archive merging for multiple files.

chumer:


chumer:
The service keys (instrument1, instrument2) are not unique across the projects so they were overwriting each other.

chumer:
I can verify that this fixes https://travis-ci.org/graalvm/graal-core/jobs/115052321
to integrate https://github.com/graalvm/graal-core/pull/100

PR #118, opened 2016-03-10
Temporarily disable instrumentation tests as they seem to fail on JDK9.

chumer:
I needed to disable the instrumentation tags as they were failing on jdk9 for unknown reasons.
This is the job that failed: https://travis-ci.org/graalvm/graal-core/jobs/115116308

@mlvdv We decided to disable this as a new Truffle version is desperately needed in graal-core. Do you have an idea what could be the cause?


mlvdv:
No idea why this would fail on jdk9, but it seems moot because this relates to the old instrumentation framework.

PR #119, opened 2016-03-10
REPL debugger:  fix display bug for frame with no slots

mlvdv:


truffle/com.oracle.truffle.tools.debug.shell/src/com/oracle/truffle/tools/debug/shell/client/REPLRemoteCommand.java
```diff
@@ -542,13 +542,17 @@ void processReply(REPLClientContext context, REPLMessage[] replies) {
...
                 Integer frameNumber = replies[0].getIntValue(REPLMessage.FRAME_NUMBER);
                 context.selectFrameNumber(frameNumber);
-                context.displayReply("Frame " + frameNumber + ":");
-                for (REPLMessage message : replies) {
-                    final StringBuilder sb = new StringBuilder();
-                    sb.append("#" + message.get(REPLMessage.SLOT_INDEX) + ": ");
-                    sb.append(message.get(REPLMessage.SLOT_ID) + " = ");
-                    sb.append(message.get(REPLMessage.SLOT_VALUE));
-                    context.displayInfo(sb.toString());
+                if (replies[0].get(REPLMessage.SLOT_INDEX) == null) {
```
chumer: I find this contract weird. Why is it assumed that if the first reply has a null slot index, there are no more replies with valid slot index? Other replies might still have a valid slot index, no?

Won't block this change, as it seems minor.

PR #121, opened 2016-03-11
Debug: update node visualizing support

mlvdv:
Clean up AST printing, use temporary "known tags" strategy to replace former functionality


chumer:
I've a hard time reviewing this, as I don't know this code very well. But the use pattern for enumerating tags looks like we discussed it. We can replace it with a better solution as soon as my tags improvement is through.

PR #122, opened 2016-03-11
Debug: Ability to attach EventConsumers on demand and initialize Debu…

entlicher:
After the changes associated with instrumentation rewrite, NetBeans Truffle debugging does not work at all. We get no calls to `PolyglotEngine.dispatch*Event()`.
The reason is, that when there's no `EventConsumer` registered, there are no events. But debugger has no control of `PolyglotEngine.Builder`, because it's not creating the `PolyglotEngine`. It's the application code that creates the `PolyglotEngine`, not debugger.
For flexibility reasons, I believe that we need the ability to add/remove the `EventConsumer` on demand. Therefore I propose to introduce two methods:
```
PolyglotEngine.addEventConsumer(EventConsumer<?> handler)
PolyglotEngine.removeEventConsumer(EventConsumer<?> handler)
```
I'm also introducing an entry access point, which replaces `dispatch*Event()` methods that are non-functional as described above. The new access point is `PolyglotEngine.initJPDADebugger()`. Having a breakpoint in this method, JPDA debugger is notified about a new `PolyglotEngine` instance (unfortunately, JPDA debugger does not have a notification mechanism for new instances created, we'd not need this access point if it has one) and can add event consumers to that new instance.

Currently, registering an `EventConsumer` does not initialize the `Debugger` (`Accessor.DEBUG` is null), therefore we need to force the Debugger initialization. I did not find another suitable place than when an event consumer is registered, force to load the `Debugger` class and it's accessor.
This looks hacky, if there are better ideas how to accomplish this, I'll gladly change it.

### DIFF
--- a/truffle/com.oracle.truffle.api.vm/src/com/oracle/truffle/api/vm/PolyglotEngine.java
+++ b/truffle/com.oracle.truffle.api.vm/src/com/oracle/truffle/api/vm/PolyglotEngine.java
@@ -64,6 +64,9 @@ import com.oracle.truffle.api.interop.java.JavaInterop;
 import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.api.nodes.RootNode;
 import com.oracle.truffle.api.source.Source;
+import java.lang.reflect.Field;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Gate way into the world of {@link TruffleLanguage Truffle languages}. {@link #buildNew()
@@ -107,13 +110,14 @@ public class PolyglotEngine {
     static final boolean JAVA_INTEROP_ENABLED = !TruffleOptions.AOT;
     static final Logger LOG = Logger.getLogger(PolyglotEngine.class.getName());
     private static final SPIAccessor SPI = new SPIAccessor();
+    private static final AtomicBoolean DBG_INITIALIZED = new AtomicBoolean(false);
     private final Thread initThread;
     private final Executor executor;
     private final Map<String, Language> langs;
     private final InputStream in;
     private final OutputStream err;
     private final OutputStream out;
-    private final EventConsumer<?>[] handlers;
+    private final List<EventConsumer<?>> handlers;
     private final Map<String, Object> globals;
     private final Instrumenter instrumenter; // old instrumentation
     private final Object instrumentationHandler; // new instrumentation
@@ -154,7 +158,7 @@ public class PolyglotEngine {
     /**
      * Real constructor used from the builder.
      */
-    PolyglotEngine(Executor executor, Map<String, Object> globals, OutputStream out, OutputStream err, InputStream in, EventConsumer<?>[] handlers, List<Object[]> config) {
+    PolyglotEngine(Executor executor, Map<String, Object> globals, OutputStream out, OutputStream err, InputStream in, List<EventConsumer<?>> handlers, List<Object[]> config) {
         assertNoTruffle();
         this.executor = executor;
         this.out = out;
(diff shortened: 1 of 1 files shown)


entlicher:
Please provide a suggestion on how to fix it when I'm not allowed to use sun.misc.Unsafe.
Thanks.

chumer:
@mlvdv whats missing so you cannot support this as it was?

@entlicher @mlvdv 
The API direction I'd prefer would be something like:
`
 Debugger.find(engine).addBreakpoint(new BreakPoint(...), new BreakPointCallback() {/* necessary JPDA callback.*/})
`
So the PolygotEngine API for event dispatching becomes obsolete. I'd like the debugging to become an instrumentation without custom hooks in PolyglotEngine. Also I think we should avoid intrinsics magic for the debugger (initJPDADebugger). What API's would you need in Debugger to achieve this? Please ping me to discuss this offline.

jtulach:
Call to `Debugger.find(engine)` should be all JPDA debugger needs to initialize the debugging.

entlicher:
@mlvdv does not have problems like this, the REPLServer creates the PolyglotEngine and attaches the event consumers in the Builder.
OTOH I connect to a starting/running JVM, which is processing some client code already. And if it happens to instantiate PolyglotEngine, I need to hook up and intercept the events.

After some more investigations, I think I will be able to live without initJPDADebugger() method and hook up to all return statements in all constructors instead. That should work, I'll check it today.

Adding an extra callback for every breakpoint (and step) is probably an overkill.
Would it make sense to leave `ExecutionEvent` on `PolyglotEngine` (there might be other uses besides debugger?) and move `SuspendedEvent` to `Debugger`? I mean, `Debugger` will have `add/removeSuspendedEventListener()` and `PolyglotEngine` will take care of `ExecutionEvent` itself.

chumer:
> OTOH I connect to a starting/running JVM, which is processing some client code already. And if it happens to instantiate PolyglotEngine, I need to hook up and intercept the events.

I see why you are doing this. But could you do it by intercepting maybe the PolyglotEngine builder call and then just use public API? You need to ensure that the debugger is enabled, its what Debugger.find(PolyglotEngine) does for you.

> Adding an extra callback for every breakpoint (and step) is probably an overkill.

I don't think so. It would be very flexible and easy to implement. Also at the moment multiple execution events are mixed into a single callback mechanism it can lead to difficult to read code.

> Would it make sense to leave ExecutionEvent on PolyglotEngine (there might be other uses besides debugger?

I think its what the instrumentation API is for. Its not the job of the PolyglotEngine.

jtulach:
Here is my attempt to fix the deficiency and let original debugger code work again: https://github.com/jtulach/truffle/tree/ExecutionEventAlways - Martin @Entlicher has to verify if it really works.

entlicher:
> But could you do it by intercepting maybe the PolyglotEngine builder call and then just use public API?

Well, true, I can hook up at `Builder.build()` and call `onEvent()` myself. I'm only afraid of the package-private constructor of PolyglotEngine together with some future changes. As long as `Builder.build()` is the only code that calls the constructor, it's fine. Probably, I should rewrite it that way for now.

> You need to ensure that the debugger is enabled, its what Debugger.find(PolyglotEngine) does for you.

The problem with this is, that nothing forces me to do so. IMHO when someone registers a consumer for an event, they expect to get some events and do not expect a necessity to call some magic code to enable these events. If the magic code is known, it should be done automatically IMHO, this is why I've added the `Class.forName(...), etc.`

> I think its what the instrumentation API is for. Its not the job of the PolyglotEngine.

Maybe this is the key point. Perhaps Debugger should be rewritten not to use PolyglotEngine at all and use just the instrumentation API directly (if it's possible to live without high-level execute start/stop notifications, or if it's possible to get them from the instrumentation API).

jtulach:
#124 is in and NetBeans Truffle debugger seems to work again, so I think we can close this request for now.

The way to *attach debugger* on demand is to call `Debugger.find`. That works fine when no execution is in progress, but it would be very nice if one could turn debugging on while `PolyglotEngine.eval` is currently in progress. @mlvdv, do you think you could look at that?

Martin @entlicher could then attach the JPDA debugger get all instances of `PolyglotEngine` and turn debugging on for all/some of them. That would give us true *attach on demand* behavior.

entlicher:
Thanks, I'm closing this pull request.

PR #123, opened 2016-03-14
Revise tagging mechanism. Add the notion of @ProvidedTags.

chumer:
Since there was bad feedback on the SourceSection#hasTag solution, I tried to come up with a more flexible approach to tagging. I changed to a protected method Node#isTaggedWith(String). So its now up to the language how tagging is implemented. 

Also new and noteworthy:
- ProvidedTags to specify for a language which tags are provided by a language. (isTaggedWith is never called with other tags than that)
- RequiredTags to specify which tags are required by an instrument. One cannot create queries for tags they do not require anymore.
- PolyglotEngine#Instrument#isCompatibleWith(Language) to check wheter an instrument is compatible to a language. (contains all tags)
- Instrumenter#isNodeTaggedWith(Node, String) to query other nodes not instrumented for a specific tag.

Since the behavior for provided and required tags is incompatible with the original version, I've introduced a temporary compatibility mode flag `-Dtruffle.instrumentation.compatibility=true`. For a language to verify that it is compatible to the new required/provided tags schema you need to run all you tests with compatibility set to `false`. 

Also please note that in compatibility mode this change is fully compatible also SourceSection#hasTag is still supported.

@mickjordan @chrisseaton @woess please review for use in language
@mlvdv @jtulach please review for everything else.

@mlvdv 
As you requested: here is a sketch to display all tags of a node:
```
    public static void displayAllTags(PolyglotEngine engine, Instrumenter instrumenter, Node node) {
        Set<String> allTags = new HashSet<>();
        for (Instrument instrument : engine.getInstruments().values()) {
            allTags.addAll(instrument.getRequiredTags());
        }
        for (Language language : engine.getLanguages().values()) {
            allTags.addAll(language.getProvidedTags());
        }

        for (String tag : allTags) {
            if (instrumenter.isNodeTaggedWith(node, tag)) {
                System.out.println(String.format("Node %s is tagged with %s.", node, tag));
            }
        }
    }
```


jtulach:
I don't understand why one should be able to display all tags of a node.

truffle/com.oracle.truffle.sl/src/com/oracle/truffle/sl/nodes/SLStatementNode.java
```diff
@@ -88,6 +83,19 @@ public SLStatementNode getNonWrapperNode() {
...
     }
 
+    public void setDebugHalt(boolean isDebugHalt) {
+        this.isDebugHalt = isDebugHalt;
+    }
+
+    @Override
+    protected boolean isTaggedWith(String tag) {
+        switch (tag) {
+            case Debugger.HALT_TAG:
```
lukasstadler: couldn't this be solved much simpler now, by just checking whether the parent is a sequence node?

smarr:
In responds to @lukasstadler question about 'just checking a parent' for something.
This brings up a problem, which might or might not have been discussed before, don't remember:

How much support is needed/desired to 'just check' a node, a parent, or child for something in the potential presence of wrapper nodes?

Do wrapper nodes delegate questions about tags to their wrapped nodes? When asking a parent node, this could give very confusing results. Do you expect every usage side to guard with `node instanceof WrapperNode`? I got quite a number of these checks in SOMns, and I am not sure I put all in that are necessary.

chumer:
@smarr Yes this is the downside of using wrappers. The alternative would be to make each leaf node class instrumentable individually without wrappers. 

> Do wrapper nodes delegate questions about tags to their wrapped nodes?
Yes since wrappers derive the automatically inherit the tags behavior. But wrapper nodes are never asked for tags by the instrumentation framework.

> Do you expect every usage side to guard with node instanceof WrapperNode?
I don't think its good practice to rely on the parent/childs type. If you do, eg for pattern matching, you can work with helpers that filter wrapper nodes.

truffle/com.oracle.truffle.sl/src/com/oracle/truffle/sl/nodes/SLStatementNode.java
chumer: I would need to check for the parent beeing a wrapper. Also isTaggedWith is defined as unchanged after AST creation, its hard to gurantee that if the parent is changing. Also its not true that each statement of a block is tagged halt in SL. For conditionals you want to step before the condition and not the full if expression. So the boolean solution is a lot easier to fine-tune.

chumer:
@jtulach @mlvdv wants to have that for instrument debugging.

woess:
I believe relying on a parent node is fragile and thus should be avoided.

lukasstadler:
> I believe relying on a parent node is fragile and thus should be avoided.
Why? as long as there's a way to identify the next "real" parent, it should be fine.
Are you thinking about situations where the parent of still-executing code is changed to null?

smarr:
Method-level recursion and AST-level self-optimization can lead to very surprising results when relying on certain invariants wrt. parent nodes.

This bit me because some information I need for tagging only becomes available during execution, i.e., after self-optimization in dynamic languages. So, the tags-are-stable-after-AST creation is very optimistic.

chumer:
@lukasstadler no parent do never become null after adoption. this is necessary also for multithreading.
I agree with @smarr that the parent is always that stable is optimistic. It might lead to hard to debug errors and its hard to verify. Wheras having a simple state in the node is easy to verify unchanged after AST creation, but it needs state. But for FastR it might make sense to do that. The good thing is that each language can now choose how to do it.

woess:
@lukasstadler Simply because there's no real guarantee that your parent is always going to be what you think it will be. You're right, it can work if you can identify the "real" parent. I'm thinking more about potential refactoring hazards: someone might later change the AST structure a bit or introduce a new parent node and forget to update the tag detection. (e.g., you might skip the SequenceNode for a single statement and suddenly the statement tagging is gone.)
Obviously, these aren't obstacles, and it's totally up to the guest language implementation.

chrisseaton:
Works for JRuby https://github.com/jruby/jruby/compare/truffle-head...truffle-new-tagging

woess:
It would have been good to split this into 2 separate changes (which I believe are mostly independent):
1. isTaggedWith
2. ProvidedTags, RequiredTags

1. I like it and I think this is the right approach. 
2. doesn't really allow for optional tags, does it? I'm not sure all these restrictions are really necessary and if not implementing one tag should already mean that the instrument is incompatible.

Please also add a changelog entry. Thanks!

chumer:
@woess ad 2: Its never going to get as easy as now to integrate this. I think adding @OptionalTags as well could solve most of our problems, right?

Ad changelog. Will add it.

woess:
@chumer I'm more concerned that we're introducing something that we don't actually need. That said I don't object to the annotations.

mickjordan:
In FastR we have some utililty functions to "unwrap" both down and up that takes care of wrapper nodes.

jtulach:
Objections:
1. instrumentation presents itself as an optional feature - it used to be once (in early January) - but since then it sneaked into core API and now it even has its central method in `Node`. Something is rotten in the kingdom of Denmark!
2. while the list of supported tags per language/instrument Require/Provide tags is nice, we don't need it for anything right now. E.g. there is some truth on Andreas @woess comment that this could be separated into two changes - and only one of them is incompatible -e.g. needs to be done in somoe hurry.
3. discussion about parent is completely out of scope of this change as that aspect isn't influenced at all.
4. listing all tags is a non-goal - e.g. `Instrument.getRequiredTags` shouldn't be public. If we really want to use that for verification of compatibility between language/instrument, such methods should be at most `protected`.

chumer:
1. Its still optional in the sense that you don't need to implement anything at all. In the current state its sneaking into the SourceSection API which is an even more central API. Do you have any suggestions on how to fix that? I think its a good compromise.
2. Will separate into separate PR.  Will reuse this PR for the tagging mechanism. I don't agree that any change in here is incompatible to the previous Truffle version. Since we introduce an all new instrumentation framework its now that we can introduce such an api like require/provide very easily. Later its going to be more difficult.
3. Its not at all out of scope as the languages want to know how they should implement isTaggedWith. How getParent() behaves is very relevant to this question.
4. Agreed. Will remove Instrument#getRequiredTags from the API.

woess:
1. Yes, it's better than the SourceSection solution. But we could extract it into an interface.
4. Agreed. There's no reason to get all tags.

chumer:
@woess 
(1.) So you suggest an interface InstrumentableNode? We had this in the first prototype. @jtulach is that better in your opinion as well?

jtulach:
It is really interesting to see how the design walks in spiral (hopefully it is a spiral, not a circle).

In December we had relatively working design based on annotations. Then we found it too constraining (Ruby had a node which might or might not be a statement) and switched to string tags to give us flexibility. Now we see that strings are bit too flexible, and we need a way to find out if language/instrument can agree on a set of tags. Thus we build a new protocol (provides/requires annotation) to find out if the tags match.

This whole round trip makes me ask: What was the core reason for rejecting annotations? My answer: the inflexibility shown by a Ruby wanting to represent statement/non-statement by a single class.

Well, that can easily be fixed by `Node.isTaggedWith` or similar. Then the dynamism is there, yet we can keep using standard Java meta programming facility - e.g. annotations. Here is my rewrite of the API: https://github.com/jtulach/truffle/commit/7b30201efb440636858daf695c03539d32706fc1 - it is not fully polished, but following works:

```bash
truffle$ mx clean && mx build && mx unittest sl.test && mx javadoc
```

The test execution includes `SLDebugTest` which makes me believe it is working correctly for the debugger usecase and adopting languages to this API will be straightforward. 

If we switch to the annotations again, we'll get static typing between instrument and a language and ability to perform static analysis during compilation - for example find out which annotations a language nodes support automatically. We'll be able to use standard IDE tools as home made protocols like provides/requires tags will no longer be needed or will operate on typed annotations.

chrisseaton:
> My answer: the inflexibility shown by a Ruby wanting to represent statement/non-statement by a single class.

The problem isn't that the Ruby team is just stubborn about adding a new class to represent statements. I can still do that if it's wanted. I would just wrap each expression that is also a statement with a new `StatementNode`, which could be annotated with the right tags.

The problem is that this new node is essentially a wrapper node - a node that just takes up space and slows down partial evaluation - and this is what the original rewrite of instrumentation was supposed to remove. That was my only objection.

If we want to go back to wrapper nodes and annotations on nodes, that's fine I can do that any time.

smarr:
From my experience with the dynamic metrics tool, annotations alone are not sufficient. You don't want to restructure your AST nodes just because you want to support a new tool. Also, it is unclear that tools are not going to have conflicting requirements when it comes to tags, so, we need more flexibility there.

woess:
@chumer: It's a possibility if @jtulach thinks it should not be part of the Node API. However, I don't think it has to be extracted, now that it's just one method querying a tag.

lukasstadler:
From the FastR perspective, what constitutes a statement cannot be tied to specific node types, because it depends on _where_ something is, not _what_ it is.
Whether the tag itself is a string or an annotation type should be decided based upon whether the tighter coupling of class references as opposed to matching stings is desired and/or possible.
The main difference is that you actually have to compile against the source of whoever provides the tags, right?

jtulach:
Hello @chrisseaton, no I don't want to re-introduce wrappers into Ruby. The system that I am proposing should have the same level of flexibility. See https://github.com/jtulach/jruby/commit/8aaed63d02eb5407a46e9cae19f0701bd4d4b724

Good point, @lukasstadler. The compilation makes the difference. Please note that in both systems (e.g. `@ProvidedTags` vs. annotations) the list of tags supported by a language is hardcoded during compilation of the language binary. Thus I think `@ProvidedTags` and annotations have the same flexibility.

chrisseaton:
`isAnnotationPresent` is a new method - this isn't related to Java's `Class#isAnnotationPresent` is it? And I guess there's a default that does the obvious thing (presumably using `Class#isAnnotationPresent`)?

Seems fine to me.

woess:
@jtulach: seems like a good compromise, although it's a bit of an abuse of annotations.

chumer:
After some discussions with @jtulach we agreed on the following changes:
* We keep Node#isTaggedWith but we use Class<?> instead of strings for tags. Classes are implicitly interned and their names can be refactored / support tooling.
* We discard RequiredTags and the instrumentation compatibility API into the next release. So instruments can still access all available tags.
* We keep ProvidedTags but with Class<?> instead of string.
* Instead of Instrumenter#isNodeTaggedWith there is now Instrumenter#queryTags(Node) which will only return tags which were provided by the language.

As opposed to my original design this PR is now not compatible anymore to the previous tagging solution. So updating to this version will require changes. It should however still compile with the old API.

@mlvdv jtulach @chrisseaton @lukasstadler  @woess  please review again

woess:
I don't quite understand why there needs to be a tie between tags and language at all. Can you clarify?

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
```diff
@@ -524,9 +525,65 @@ public final void atomic(Runnable closure) {
...
+     * </code>
+     * </pre>
+     *
+     * <p>
+     * Often it is impossible to just rely on the node's Java type to implement tagging. This
+     * example shows how to use local state to implement tagging for a node.
+     *
+     * <pre>
+     * <code>
+     * private boolean isDebuggerHalt;
```
jtulach: Can we use code snippets, please? You can copy one from: https://github.com/jtulach/truffle/commit/7b30201efb440636858daf695c03539d32706fc1#diff-390d3d3d22a24a2b4a58c1918871c59fR704

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
```diff
@@ -524,9 +525,65 @@ public final void atomic(Runnable closure) {
...
+     * calling this method for a particular tag remains always the same.
+     *
+     * @param tag the class {@link com.oracle.truffle.api.instrumentation.ProvidedTags provided} by
+     *            the {@link TruffleLanguage language}
+     * @return <code>true</code> if the node should be considered tagged by a tag else
+     *         <code>false</code>.
+     * @since 0.12
+     */
+    protected boolean isTaggedWith(Class<?> tag) {
+        return false;
```
jtulach: Does the Javadoc state what is the default value? Of course, my favourite default implementation would be:
```java
return Annotation.class.isAssignableFrom(tag) && getClass().isAnnotationPresent(Annotation.class.asSubclass(tag));
```
but we can leave up until the moment when there is an instrument that could benefit from annotating node classes more than the debugger `HaltTag` and `CallTag` can.

jtulach:
OK from my perspective.

chumer:
@woess  because we don't want another interface that encodes some magic. wheras having it in Node is very simple. SourceSection is also in Node. Why isn't that an interface? Its also an optional feature.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
chumer: Lets see if languages want to use that first. Would be interesting know if the languages  want to use tags as annotations on node classes? (cc @lukasstadler @chrisseaton @woess)

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
chumer: Yes. I will change it.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
smarr: For SOMns and the dynamic metrics tool tags work for some things, and I'd likely use them. Can prototype something next week.

chumer:
Will wait a bit with integrating to give languages time to prepare patches.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
lukasstadler: IMO, this interface should be structured in one of two ways, either:
- accept only annotation classes and detect annotations on the node class if we think that this is the way it's used most of the time (and isTaggedWith is just a fallback for the odd case that cannot be tied to the node class), or:
- not make any assumptions about the class being an annotation if we think that it's going to be specified with isTaggedWith most of the time anyway.

I favor the second option, because it has a smaller interface, and because it looks like at least R and Ruby will use isTaggedWith anyway.
In that cast it makes sense to introduce a base class/interface which documents a class being designed to be used as a tag: `isTaggedWith(Class<? extends SyntaxTag> tag)`

chumer: I am also in favor of the second option. 

@lukasstadler @jtulach  and I discussed if we should introduce a base class. What benefit would that bring except documentation? Why not reusing some existing classes as tags? I cannot see a problem with that.

lukasstadler: Just reusing an existing class for that means that, without looking at documentation, I don't know that this class is used in this way. Good luck finding the bug that appears because you're looking for `isTaggedWith(Statement.class)` instead of `isTaggedWith(StatementNode.class)` or because you're testing for `return tag == Statement.class` instead of `return tag == StatementNode.class`. Type safety would prevent both of them.

woess:
@chumer: you misunderstood my comment, it was probably not clear. I'm completely fine with having it in Node. What I'm wondering is, why do we need to filter the tags based on the language's provided tags.

@chumer @lukasstadler I was thinking about an interface for tag classes as well, but it would preclude use of annotations (afaik, annotations cannot implement interfaces). So I'd leave it as is.
@chumer I probably won't use annotations as tags myself, but I can imagine it to be useful (e.g., for the call tag).

chrisseaton:
@chumer is now the right time to try using this in Ruby again?

chumer:
@woess It fixes the tags for a language at javac compile time.
It enables us to enumerate supported tags of a language. The language can rely on the fact that isTaggedWith is never called by something else. We optimize instrumentation to not traverse ASTs that do not support the filtered tag. We can later support compatibility checks similar to RequiredTags (but probably implemented by the instrument).

@chrisseaton  yes this is the right time.

woess:
Note that tags stored in node fields need to be transferred to the new node when a node replaces itself. The `Node.onReplace` hook can be used for that matter (if the flag is not passed to the constructor already). Like so:
```java
private boolean isDebuggerHalt;
@Override
protected void onReplace(Node newNode, CharSequence reason) {
    super.onReplace(newNode, reason);
    ((StatementNode) newNode).isDebuggerHalt = isDebuggerHalt;
}
```

chrisseaton:
Works for JRuby.

mickjordan:
I have merged the version from mid-March with FastR and it seems work ok and is definitely simpler overall in the code. However I am concerned about the recent change to using a class instead of a string. In particular, I had a proposal that we could use string pattern matching to identify,e.g statements,
and be able to get ride of Debugger.HALT. I really don't want to statically depend on the Debugger at all in FastR code so will there be either some subclassing solution to tag classes or finally bring back the StandardTags of the old instrumentation framework?

chrisseaton:
> I really don't want to statically depend on the Debugger at all in FastR code

Yes this was a key reason to use strings and I missed the reason why we moved to classes.

jtulach:
Re. "I really don't want to statically depend on the Debugger at all" - why? Is there some rational reason or is it a matter of preferences?

I can imagine rational reasons. For example packaging. If debugger was a separate JAR, then depending on its tags would force you to always run with the debugger JAR on classpath. However debugger API isn't separate - it is standard part of Truffle API, so this can't be the explantation.

Re. "finally bring back the StandardTags" - yes, defining common set of tags is the ultimate goal. However there isn't consensus what they should be these days. As such we have to live with tags defined by each instrument in forthcoming release 0.12.

woess:
@mickjordan @jtulach Is this perhaps is a problem of having the debugger implementation in the `api.debug` project rather than a separate project?

chumer:
@woess yes at the moment the debugger API uses the Accessor API. But I think we should make the Debugger only dependent on the instrumentation longer term. There is still more stuff in instrumentation to do to support this.

Regarding the change for classes instead of tags. Having strings has significant disadvantages as they are not checked at all. Classes are statically checked, always interned and refactorable. It comes with the downside of having to depend on the tool. This does not mean that tools cannot start sharing certain tags in a project which then the language can then depend on. Longer term we are going to share the evolved standard tags in the Truffle API.

woess:
@chumer :+1:
I don't see how strings would solve this problem anyway, you'd still want to use an exported constant.

mickjordan:
Last I checked I couldn't trivially compile using Debugger without using another dist (that wasn't exported?) But I just object (a) to having to import a tool API into FastR and (b) why is it so darn hard to get some consensus on something as simple as a standard statement and call tag? Remember we used to have a consensus on this in the old framework. Whoever decided that we don't have one anymore needs to explain why. 

I am obliged to remind you all that in FastR we actually use the debugger, or rather we would if it wasn't broken so much of the time. Fortunately we have the internal R debugger (over which we have control, thankfully)  but that is missing some key features like line breakpoints.

chrisseaton:
The reason we agreed not to standardise on tags right now was that we couldn't come to a conclusion on what tags we needed. At one point we had a long speculative list, but to me and at least some other people it seemed obvious that some tags from that list would turn out to be useful and others wouldn't. The idea therefore was to use strings for a couple of versions, figure out what tags were actually common, and to then make them into an API. Part of it I think was a way to get the work merged sooner without committing ourselves to long-term API.

mlvdv:
>What was the core reason for rejecting annotations?

The core reasons were discussed many times during review:
1. Class Annotations are not sufficiently flexible for non-trivial language implementations, as all experience has shown.
2. Class Annotations commit a fundamental architectural mistake: coupling AST implementation (where priorities are on semantics and performance) with the behavior of user-facing  tools.  Tool behavior depends, among many other things on "user expectations", evolving patterns of use, and the ongoing need for fine-grained tuning.  @smarr summarized nicely:  "You don't want to restructure your AST nodes just because you want to support a new tool."  To that I would add:  you don't want to restructure your AST nodes just because a usability bug was filed against the debugger.  In practice, any usability bug requiring changes to AST Node classes would likely never get fixed.

Effort spent basing any part of tagging on annotations is not time well spent.

mlvdv:
>The idea therefore was to use strings for a couple of versions, figure out what tags were actually common, and to then make them into an API. Part of it I think was a way to get the work merged sooner without committing ourselves to long-term API.

Agreed.  That lack of consensus was the "show stopper",  but there are other requirements for tags that have been pushed aside ("on hold") during this "experimental" phase.  I agreed to that decision only because it bought time for more people to get experience with tags (which has now started with attention to Debugger behavior).  More experience will help address the real functionality for which tags are designed. We should make no API commitments about tags in this release that restrict future choices.

I suggest we move in a positive direction toward understanding tags by adopting a temporary, minimal set of "standard" tags, e.g. "STATEMENT", "EXPRESSION", "CALL", and "ROOT".  It is important to keep open the flexibility we chose for "a couple of versions" when we made that decision.

chumer:
@mickjordan we reached this consensus in the review meetings. there is not just one person responsible for that. 

```I suggest we move in a positive direction toward understanding tags by adopting a temporary, minimal set of "standard" tags, e.g. "STATEMENT", "EXPRESSION", "CALL", and "ROOT". It is important to keep open the flexibility we chose for "a couple of versions" when we made that decision.```

Ok lets do this temporary minimal set of tags again to make everybody happy. We can still deprecate and remove them later if they turn out wrong. Any objections?

**Comment by chrisseaton** [comment]
There's so many questions to resolve to add something like `CALL` though. Does it mean explicit calls? What about things that look like operators but may be calls? What about things that make implicit calls such as to convert types?

Is there an application right now for `EXPRESSION` and `CALL`? What is `ROOT`?

`STATEMENT` makes sense straight away.

**Comment by smarr** [comment]
No objection here, and perhaps that should be discussed on a separate issue: what are the precise definitions of these tags? With that I mean, what can I expect as a tool builder? What's the difference between statement and expression, etc.

Ha, Chris beat me to it.

**Comment by chumer** [comment]
> Does it mean explicit calls? What about things that look like operators but may be calls? What about things that make implicit calls such as to convert types?

At the moment it means any call that should be visible to a language user. Implicit calls for conversions can be seen as such. However it is not a problem if some calls are not instrumentable. The debugger will just skip over in such a case.

**Comment by chumer** [comment]
Here is my suggestion of tags. EXPRESSION has no use-case yet and will not be included in this release. Maybe the next. I tried to add use-case descriptions to each tag to make it easier to implement the tags for languages. If it turns out that some of the definitions are ambiguous (coverage STATEMENT vs debugger STATEMENT) then we can deprecate and introduce specific tags.

```
package com.oracle.truffle.api.instrumentation;
/**
 * Set of standard tags usable by language agnostic tools. Language should {@link ProvidedTags
 * provide} an implementation of these tags in order to support a wide variety of tools.
 *
 * @since 0.12
 */
public final class StandardTags {

    private StandardTags() {
        /* No instances */
    }

    /**
     * Marks program locations that represent a statement of a language.
     * <p>
     * Use case descriptions:
     * <ul>
     * <li><b>Debugger:</b> Marks program locations where ordinary stepping should halt. The
     * debugger will halt just <em>before</em> a code location is executed that is marked with this
     * tag.</li>
     * </ul>
     *
     * @since 0.12
     */
    public final class StatementTag {
        private StatementTag() {
            /* No instances */
        }
    }

    /**
     * Marks program locations that represent a call to other guest language functions, methods or
     * closures.
     * <p>
     * Use case descriptions:
     * <ul>
     * <li><b>Debugger:</b> Marks program locations where <em>returning</em> or
     * <em>stepping out</em> from a method/procedure/closure call should halt. The debugger will
     * halt at the code location that has just executed the call that returned.</li>
     * </ul>
     *
     * @since 0.12
     */
    public final class CallTag {
        private CallTag() {
            /* No instances */
        }
    }

    /**
     * Marks program locations as root of a function, method or closure. The root prolog should be
     * executed before and the epilog after the source location marked as root is executed.
     * <p>
     * Use case descriptions:
     * <ul>
     * <li><b>Profiler:</b> Marks every root that should be profiled.</li>
     * </ul>
     *
     * @since 0.12
     */
    public final class RootTag {
        private RootTag() {
            /* No instances */
        }
    }

}
```

chrisseaton:
So if I have a deep tree of nodes that form my method prologue, `ROOT` should be set on the first user node, at the bottom of that tree? So if you broke there all the arguments are loaded into local variables and things like that?

chumer:
@chrisseaton think of it as: you need to be able to run code that uses local variables at that location. If you would do that before the prolog your expression might not be evaluatable.

lukasstadler:
tas can be defined by what the debugger and other tools do with this information or at an abstract language level, e.g., create a definition for what a statement is.
the former requires tool writers to collaborate on creating a set of tags, while the latter additionally requires involvement form the language writers.

I'd go with the former - what difference  make whether something is a statement, call or expression, other than what tools do with it?

+1 though on discussion this as a separate issue.

chumer:
> I'd go with the former - what difference does it make whether something is a statement, call or expression, other than what tools do with it?

Are you opting for no tag sharing between tools? We want a standard set of tags shared between the languages independent of tools. There is/was consensus about that in the meetings. How exactly the tags look like, we don't know. But we can evolve them easily. So it might turn out that we have a DEBUGGER_HALT standard tag in the future. Lets wait and see.

> +1 though on discussion this as a separate issue.
Yes we should continue the discussion. For this release I will go with this set of tags (or maybe some minor modifications) because we have use-cases. If it turns out they are wrong we deprecate them. I do this now to fix concerns raised by @mickjordan .

smarr:
Was there a specific reason not to introduce a common super class for tags? If I am not mistaken someone else also asked for it. Could make finding tags easier, I think. (and I think typing was brought up as another benefit.)

woess:
Not that _statement_ is not necessarily the same as _debugger-halt_ (i.e., you might want to halt on something that's not a statement (in the language) and not halt on some specific statements). Can this be a problem?

woess:
> Was there a specific reason not to introduce a common super class for tags? If I am not mistaken someone else also asked for it. Could make finding tags easier, I think. (and I think typing was brought up as another benefit.)

One reason not to have a common superclass could be to allow annotation classes to be used as tags.

smarr:
@woess, I think it is more of an indication that tags will likely be rather tool specific. For my tool, I see only the root node as useful.

wrt. superclass: ok, I see.

lukasstadler:
> Are you opting for no tag sharing between tools?
Quite the opposite. My point is that it doesn't make sense for the languages to add tags and hope that they are useful. The set of tags should result from discussions on what would be useful for tools, and individual tags should be defined in terms of the expected behavior (as opposed to names of programming language concepts, which can be interpreted in wildly different ways by different languages).

woess:
@smarr Not sure if that's relevant to users, though. The standard tags aren't annotation classes, so these can't be used as annotations right now.

mlvdv:
This is a good discussion, but sometimes the most important requirements get lost in a sea of implementation detail.   This is a big deal because tags are the single most important functional thing about Truffle Instrumentation (runtime overhead seems well under control now, and everything else is secondary).

Here is the first of three comments about the actual requirements for tags in the Instrumentation Framework 

First:  the usefulness of Truffle Instrumentation depends almost completely on their _flexibility_.

Here’s why.  Tags provide the essential _level of indirection_ between language engineering and tool behavior.  Tags make it possible to configure tool behavior with the least possible impact on language engineering. Their concerns are orthogonal:
- _Language engineering_ is strongly constrained by correctness and mainly motivated by performance.
- _Tool behavior_, on the other hand, is mainly sensitive to user expectations, part of an evolving tool set, and subject to frequent fine-tuning.

So, do not restrict how or when language implementors apply tags any more than absolutely necessary for fast path performance. Every bit of coupling between tool behavior and  language engineering reduces the usefulness of tags, and this includes _anything_ that relates to AST node types. This is true for the standard tools we are building now, and it will becomes more true for tools not yet built.  

Other than fast path concerns, flexibility is the _**only**_ thing that matters.  A tag gets bound to one or more source locations; the rest is implementation detail.

mlvdv:
A second comment on the real requirements for tags. 

The critical functional requirement for Truffle tags is _**flexibility**_, but this discussion is also about how to use them.  Instrumentation/tools will only get adopted in practice if we get the tags right.  Our strategy for adoption is:
- provide workable tools to both language implementors and end users
- for very little effort
- that can be incrementally fine-tuned and improved as needed.

This strategy creates two hard requirements:

1. The platform must define _standard tags_ that configure a collection of standard tools (debugger, profiler, etc.), and it must be as easy as possible for language implementors to apply tags and to adjust them frequently.

2. The platform must provide _standard tools_ that are _**useful**_ (if not optimal) for every language implementation that applies standard tags, and it must be as easy as possible to fine-tune tool behavior using tool-specific tags.

Adoption depends on this positive feedback loop. A language implementor adds a few standard tags, and the Truffle platform rewards her/him immediately with tools that do something useful.  Since they are useful, the language implementor and colleagues use tools during development and notice rough edges.  This motivates him/her to improve tool behavior, possibly in collaboration with tool builders, by adjusting standard tags and fine-tuning tool-specific tags.

mlvdv:
A third comment on the real requirements for tags.

As recent discussion shows, we are still stuck on how to define _standard tags_.  The only way forward is to keep working with tags collaboratively, tool builders together with language implementors.  We have to do this, even if this costs some API stability.  

@chumer  want to return to tags as objects, with some new (to me) features, and detach them from SourceSection.  This is the right thing to do (except for not enough tagging flexibility!). 

In order to move forward and get some experience, we should start now with an absolutely minimal set of standard tags.  For Debugger only STATEMENT and CALL.  I know that CALL can be tricky, so let’s get started working through those issues with Debugger and Profiler as clients.  There seems to be an understood use case for ROOT, so that could also be included. All other tags can be postponed.

Then we can redirect most of this discussion, based on very little experience,  back to working with the actual tags, languages, and tools.  We'll all learn a lot.

Can we manage this within our API constraints?  I’d be happy to keep a new set of standard tags out of the published API during this experimental period, if that’s workable; otherwise we just have to take some stability hit so we can work out issues between language implementations and tools.

chumer:
@mlvdv don't my latest changes exactly do what you want? Introducing a minimal StandardTags class? I want to add the tags to the published API. No problem to deprecate and remove them later. Could you comment on the proposed set of StandardTags (with javadoc)?

Also another small correction: we don't move to objects as tags but classes as tags which is somewhat different.

jtulach:
I did a quick mapping of @mlvdv high-level goals to the actual code included in this pull request and as far as I can tell, the essentials are part of @chumer changes. The **flexibility** is there, **type-safety** has been increased, and now we even converged on a minimal set of **general** purpose **tags**! 

That is great and I am looking forward to offer these changes to public as part of Truffle 0.12 release.

chumer:
Will merge after the check is through.

PR #125, opened 2016-03-14
New truffle examples project.

cosminbasca:
Contains a simple graph analytics example on how the truffle API could be used in order to avoid the use of the Node class.

### DIFF
--- a/mx.truffle/suite.py
+++ b/mx.truffle/suite.py
@@ -364,6 +364,21 @@ suite = {
        "workingSets" : "Truffle,SimpleLanguage,Tools",
        "license" : "UPL",
      },
+
+    "com.oracle.truffle.examples" : {
+       "subDir" : "truffle",
+       "sourceDirs" : ["src"],
+       "dependencies" : [
+                        "TRUFFLE_API",
+                        ],
+       "annotationProcessors" : ["TRUFFLE_DSL_PROCESSOR"],
+       "checkstyle" : "com.oracle.truffle.sl",
+       "javaCompliance" : "1.8",
+       "workingSets" : "Truffle",
+       "license" : "UPL",
+     },
+
+
    },
 
   "licenses" : {
@@ -475,5 +490,6 @@ suite = {
          "TRUFFLE_DEBUG",
       ],
      },
+
   },
 }
--- a/truffle/com.oracle.truffle.examples/src/com/oracle/truffle/example/simplegraphanalitycs/TruffleExt.java
+++ b/truffle/com.oracle.truffle.examples/src/com/oracle/truffle/example/simplegraphanalitycs/TruffleExt.java
@@ -1,0 +1,77 @@
+package com.oracle.truffle.example.simplegraphanalitycs;
+
+import com.oracle.truffle.api.CallTarget;
+import com.oracle.truffle.api.RootCallTarget;
+import com.oracle.truffle.api.Truffle;
+import com.oracle.truffle.api.TruffleLanguage;
+import com.oracle.truffle.api.frame.MaterializedFrame;
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.instrument.Visualizer;
+import com.oracle.truffle.api.instrument.WrapperNode;
+import com.oracle.truffle.api.nodes.Node;
+import com.oracle.truffle.api.nodes.RootNode;
+import com.oracle.truffle.api.source.Source;
+import java.io.IOException;
+import java.util.function.Supplier;
+
+public class TruffleExt {
+    
+    public static RootCallTarget createCallTarget(Supplier r) {
+        RootCallTarget rootTarget = Truffle.getRuntime().createCallTarget(new RootNode(TruffleLanguage.class, null, null) {
+            @Override
+            public Object execute(VirtualFrame vf) {
+                return r.get();
+            }
+        });
+        return rootTarget;
+    }
+    
+    public class TestLanguage extends TruffleLanguage {
+
+        @Override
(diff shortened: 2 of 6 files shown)


truffle/com.oracle.truffle.examples/src/com/oracle/truffle/example/simplegraphanalitycs/TruffleExt.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+        });
+        return rootTarget;
+    }
+    
+    public class TestLanguage extends TruffleLanguage {
+
+        @Override
+        protected Object createContext(TruffleLanguage.Env env) {
+            throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
```
woess: I don't like the comment, but actually I think this class is unnecessary (is it even used). If you need a dummy `TruffleLanguage` class, I think you can use the existing `com.oracle.truffle.api.TestingLanguage`.

woess:
Package name: s/analitycs/analytics/. Also, file headers are missing.

truffle/com.oracle.truffle.examples/src/com/oracle/truffle/examples/simplegraphanalytics/TruffleExt.java
```diff
@@ -0,0 +1,111 @@
...
+            @Override
+            public Object execute(VirtualFrame vf) {
+                return r.get();
+            }
+        });
+        return rootTarget;
+    }
+    
+    public class TestLanguage extends TruffleLanguage<Object> {
+
```
jtulach: Where is the TestLanguage used?

mx.truffle/suite.py
```diff
@@ -364,6 +364,21 @@
        "workingSets" : "Truffle,SimpleLanguage,Tools",
        "license" : "UPL",
      },
+
+    "com.oracle.truffle.examples" : {
+       "subDir" : "truffle",
```
jtulach: I don't think we need new project/module. Rather than a static example, it is better to convert the sample into test that we execute regularly. I would put it into one of existing `*.test` projects. Creating new *test* project is also an option, but I am not sure we need that.

Please note that https://github.com/graalvm/truffle/pull/126/ tries to introduce micro benchmarks into Truffle and this example is a perfect candidate for a microbenchmark.

mx.truffle/suite.py
```diff
@@ -364,6 +364,21 @@
...
+       "sourceDirs" : ["src"],
+       "dependencies" : [
+                        "TRUFFLE_API",
+                        ],
+       "annotationProcessors" : ["TRUFFLE_DSL_PROCESSOR"],
+       "checkstyle" : "com.oracle.truffle.sl",
+       "javaCompliance" : "1.8",
```
jtulach: Currently Truffle default compliance is 1.7 - it shouldn't be that big problem - just replace `Suplier` with `java.util.concurrent.Callable`.

truffle/com.oracle.truffle.examples/src/com/oracle/truffle/examples/simplegraphanalytics/TruffleExt.java
```diff
@@ -0,0 +1,111 @@
...
+import com.oracle.truffle.api.nodes.Node;
+import com.oracle.truffle.api.nodes.RootNode;
+import com.oracle.truffle.api.source.Source;
+import java.io.IOException;
+import java.util.function.Supplier;
+
+public class TruffleExt {
+
+    public static RootCallTarget createCallTarget(Supplier<Integer> r) {
+        RootCallTarget rootTarget = Truffle.getRuntime().createCallTarget(new RootNode(TruffleLanguage.class, null, null) {
```
jtulach: If I understand correctly, the whole example is about intricate use of this method/`CallTarget`. We need to make sure the Javadoc of the method references prose sections describing what this sample does and showing selected [code snippets](https://github.com/graalvm/truffle/pull/95) illustrating the behavior. Without connection to the documentation the sample will be hard to discover.

Alternatively we can introduce new method somewhere that makes this kind of usage convenient. Then the documentation would belong to that method.

truffle/com.oracle.truffle.examples/src/com/oracle/truffle/examples/package-info.java
```diff
@@ -0,0 +1,42 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package com.oracle.truffle.examples;
```
jtulach: I don't think we want this sample to provide package-info file.

In Truffle repository package-info is reserved for API packages and needs additional "annotations". This example isn't part of the API as such. I would prefer to reference its [code snippets](https://github.com/graalvm/truffle/pull/95) at appropriate (probably `TruffleRuntime.createCallTarget` method) place in the API.

PR #126, opened 2016-03-14
Add JavaInteropSpeedBench.

zapster:
JMH micro benchmark derived from `JavaInteropSpeedTest` (cf #112). 

Run this from **graal-core**:
```
mx microbench -- com.oracle.truffle.api.interop.java.bench.JavaInteropSpeedBench
```

Note that this is the truffle alternative to graalvm/graal-core#126. I'm fine with versions.

/cc @woess


zapster:
Wow, both alternatives (#126 and graalvm/graal-core#126) have the same PR number.

woess:
I'm ok with both alternatives. Leaving the decision to @jtulach.

chumer:
I'd prefer having it in Truffle instead of Graal. I also want to add some DSL interpreter benchmarks.

zapster:
@chumer Note that currently the `mx microbench` command is part of **graal-core**. We would need to move it if you want to use it in truffle without graal.

woess:
I guess we can remove `JavaInteropSpeedTest` now, too.

zapster:
Closing as graalvm/graal-core#126 has been merged.

jtulach:
I've merged graalvm/graal-core#126 as it is smaller change - the jmh infrastructure is already there. On the other hand,  I would still like to have jmh in Truffle. We need benchmarks and having them in two repositories isn't comfortable - which is the most likely reason why we had just the one "speed" test.

PR #129, opened 2016-03-15
Fixing typo in the ValueProfile documentation sample

jtulach:
There was an error in `profieldValue`...


chumer:
The title of the PR is misleading. It also changes the example quite a bit. It does not actually represent how I intended this profile to be used anymore. Its very uncommon (I've never seen this), that you want to configure your profile in the constructor of a node. I'd prefer if we instead split it into three examples for you profile kind: identity, class and equality.

jtulach:
As you wish. I only wanted to report the bug. Feel free to make the code look the way you like.

PR #131, opened 2016-03-16
Byte overflow & failure message simplification 

jtulach:
@lukasstadler pointed out that the byte addition could in fact overflow in some languages. Narrowing the range to prevent that.

@lukasstadler also suggested to remove the failure message and keep just the formula.


lukasstadler:
looks good. we certainly cannot mandate overflow or non-overflow behavior.

PR #133, opened 2016-03-16
Add method SuspendedEvent.kill()

mlvdv:
Remove the need for debugger clients to explicitly use KillException.

@chumer please also review


jtulach:
Martin @entlicher has just given me an idea: Isn't the *kill* operation an alternative to *step over*, *step into* or *continue*? If so, then in [favor of consistency](http://lafo.ssw.uni-linz.ac.at/javadoc/truffle/latest/com/oracle/truffle/api/debug/SuspendedEvent.html) the method should be named
```java
public void prepareKill();
```
and not throw any exception until the execution is returned back to the debugger from the `SuspendedEvent` handler.

Issue #128, opened 2016-03-15
Truffle fails to compile with jdk7


dougxc:
On a clean Truffle repo:

```
dsimon@kurz ~/g/truffle> mx --strict build -n --no-javac-crosscompile --dependencies com.oracle.truffle.api.interop.java
Compiling com.oracle.truffle.api with javac... [/Users/dsimon/graal/truffle/mxbuild/truffle/com.oracle.truffle.api/bin/com/oracle/truffle/api/Assumption.class does not exist]
/Users/dsimon/graal/truffle/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/instrument/impl/DefaultVisualizer.java:91: warning: [deprecation] displayIdentifier(FrameSlot) in Visualizer has been deprecated
    public String displayIdentifier(FrameSlot slot) {
                  ^
/Users/dsimon/graal/truffle/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/instrument/impl/DefaultVisualizer.java:83: warning: [deprecation] displayValue(Object,int) in Visualizer has been deprecated
    public String displayValue(Object value, int trim) {
                  ^
/Users/dsimon/graal/truffle/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/instrument/impl/DefaultVisualizer.java:78: warning: [deprecation] displayCallTargetName(CallTarget) in Visualizer has been deprecated
    public String displayCallTargetName(CallTarget callTarget) {
                  ^
/Users/dsimon/graal/truffle/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/instrument/impl/DefaultVisualizer.java:66: warning: [deprecation] displayMethodName(Node) in Visualizer has been deprecated
    public String displayMethodName(Node node) {
                  ^
/Users/dsimon/graal/truffle/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/instrument/impl/DefaultVisualizer.java:49: warning: [deprecation] displaySourceLocation(Node) in Visualizer has been deprecated
    public String displaySourceLocation(Node node) {
                  ^
/Users/dsimon/graal/truffle/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/instrument/impl/DefaultVisualizer.java:44: warning: [deprecation] getASTPrinter() in Visualizer has been deprecated
    public ASTPrinter getASTPrinter() {
                      ^
6 warnings
[...]

PR #136, opened 2016-03-17
Truffle shall be compilable on JDK7

jtulach:
First step in addressing #128

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: java
 python:
   - "2.7"
 jdk:
-  - oraclejdk7
+  - oraclejdk8
 before_install:
   - sudo pip install astroid==1.1.0
   - sudo pip install pylint==1.1.0
@@ -25,5 +25,8 @@ env:
   - TEST_COMMAND='sh .travis.sigtest.sh'
   - TEST_COMMAND='mx/mx testdownstream'
 matrix:
-  allow_failures:
+    include:
+      - env: TEST_COMMAND='mx/mx -v build'
+        jdk: oraclejdk7
+    allow_failures:
     - env: TEST_COMMAND='mx/mx testdownstream'
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: java
 python:
   - "2.7"
 jdk:
-  - oraclejdk8
+  - oraclejdk7
 before_install:
   - sudo pip install astroid==1.1.0
   - sudo pip install pylint==1.1.0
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -2,7 +2,7 @@ language: java
 python:
   - "2.7"
 jdk:
-  - oraclejdk8
+  - oraclejdk7
```
smarr: Should these tests run on both JDKs?
If so, you could keep the `- oraclejdk8` line.

jtulach: The original gate used to run on JDK7 only, so I would not be afraid to switch back just to JDK7. In fact all of us are using JDK8 to develop (at least I do), so the JDK8 compatibility is naturally guaranteed. Should something sneak through, it would be discovered by graal-core gate when updating Truffle dependency, as they use oraclejdk8, I assume. These are the thoughts that are influencing my opinion, but in general I am ambivalent:

**Pros**: Increased coverage and better sanity check
**Cons**: Longer execution time

dougxc:
When this passes the `--strict` gate on jdk7, there's not really anything extra I need to verify ;-)

.travis.yml
dougxc: I think you can rely on graal-core and standard development environment for jdk8 testing.

jtulach:
The [last Travis run](https://travis-ci.org/graalvm/truffle/builds/116662266) compiled everything correctly on JDK7, but failed when generating Javadoc. Looks like we need JDK8 to properly generate Javadoc... maybe I will need the dual JDK config at the end.

Anyone has ideas about best way to convince Travis to compile on JDK7 and use JDK8 for the `mx javadoc` call?

smarr:
Perhaps call `jdk_switcher use oraclejdk8` explicitly in the script? That's what travis seems to do at the beginning if I read the log right.

.travis.yml
```diff
@@ -25,5 +25,8 @@ env:
   - TEST_COMMAND='sh .travis.sigtest.sh'
   - TEST_COMMAND='mx/mx testdownstream'
 matrix:
-  allow_failures:
+    include:
+      - env: TEST_COMMAND='mx/mx -v build'
+        jdk: oraclejdk7
```
jtulach: I decided to use the [config suggested](https://github.com/jruby/jruby/blob/master/.travis.yml#L74-L75) by @eregon.

jtulach:
As https://travis-ci.org/graalvm/truffle/jobs/116873622 indicates, we really run `mx build` on JDK7. That should be enough to address #128. @dougxc please merge or comment.

I see an `accept` label and since then only formatting and travis config changes were made. I am going to rebase and merge now.

.travis.yml
eregon: There seems to be one extra level of indentation for `include:` and below.

PR #137, opened 2016-03-17
Rename SuspendedEvent.kill() to prepareKill() and do kill lazily.

entlicher:
`SuspendedEvent.kill()` directly throw the `KillException`, which has unpleasant consequences:

- Any code that follows `SuspendedEvent.kill()` call will not be processed, which may not be expected.
- We can not prevent users from accidentally calling `SuspendedEvent.kill()` in a possible asynchronous code handling the events. In this case it unexpectedly blows up.

To achieve the same effect, which was intended by the original code, but to be more polite to the caller of this method, I suggest to have just a `kill` marker on the `SuspendedEvent` and throw the `KillException` in the infrastructure. As the method is not killing right away now, I'm changing the name from `kill()` to `prepareKill()`, which corresponds with other `prepare*()` methods.

### DIFF
--- a/truffle/com.oracle.truffle.tools.debug.shell/src/com/oracle/truffle/tools/debug/shell/client/REPLRemoteCommand.java
+++ b/truffle/com.oracle.truffle.tools.debug.shell/src/com/oracle/truffle/tools/debug/shell/client/REPLRemoteCommand.java
@@ -579,6 +579,7 @@ public abstract class REPLRemoteCommand extends REPLCommand {
             } else {
                 context.displayFailReply(replies[0].get(REPLMessage.DISPLAY_MSG));
             }
+            throw new REPLContinueException();
         }
     };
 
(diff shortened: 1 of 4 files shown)


jtulach:
Everybody seems to be OK with #140 (which includes Martin's commit), so let's focus on that one.

truffle/com.oracle.truffle.tools.debug.shell/src/com/oracle/truffle/tools/debug/shell/client/REPLRemoteCommand.java
```diff
@@ -579,6 +579,7 @@ void processReply(REPLClientContext context, REPLMessage[] replies) {
             } else {
                 context.displayFailReply(replies[0].get(REPLMessage.DISPLAY_MSG));
             }
+            throw new REPLContinueException();
```
woess: was this a bug in the REPL?

PR #138, opened 2016-03-17
OM-DSL: make the generated ObjectType public

chrisseaton:
The `ObjectType` subclass we generate as part of the OM DSL is currently hidden as an implementation detail. But now I've found out that it would be useful to have it for the foreign access DSL, which needs it in the required `receiverType` parameter of the  `AcceptMessage` annotation.

This doesn't change any signatures but I guess it counts as an API change on the as-yet unreleased OM-DSL as the generated code is different, but backwards compatible.

### DIFF
--- a/truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/ObjectTypePublicTest.java
+++ b/truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/ObjectTypePublicTest.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.truffle.api.object.dsl.test;
+
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+import com.oracle.truffle.api.object.DynamicObject;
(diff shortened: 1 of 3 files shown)


truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/ObjectTypePublicTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class ObjectTypePublicTest {
+
+    private static final ObjectTypePublicTestLayout LAYOUT = ObjectTypePublicTestLayoutImpl.INSTANCE;
+
+    @Test
+    public void testIsPublic() {
+        final DynamicObject instance = LAYOUT.createObjectTypePublicTest(14);
+        assertTrue(ObjectTypePublicTestLayoutImpl.ObjectTypePublicTestType.class.isAssignableFrom(instance.getShape().getObjectType().getClass()));
+    }
```
jtulach: At first sight, I was not sure what this test checks. Is `isAssignableFrom` anyhow influenced by public visibility? Probably not. But without the `public` change you wouldn't be able to reference `ObjectTypePublicTestLayoutImpl.ObjectTypePublicTestType`.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/LayoutGenerator.java
```diff
@@ -196,7 +196,7 @@ private void generateObjectType(final PrintStream stream) {
             typeSuperclass = layout.getSuperLayout().getName() + "LayoutImpl." + layout.getSuperLayout().getName() + "Type";
         }
 
-        stream.printf("    protected static class %sType extends %s {%n", layout.getName(), typeSuperclass);
+        stream.printf("    public static class %sType extends %s {%n", layout.getName(), typeSuperclass);
```
jtulach: The class was `protected`!? That still makes it visible. It should have been `private`.

jtulach:
My obsession with showing as little as possible is known since #40 where I suggested to use one shared `Layouts` class for all types. That didn't work out due to some problems with incremental compilation in Eclipse. Still my desire to expose only what is necessary hasn't gone away.

With this change we have two classes and one field generated per layout class. I'd like to reduce that a bit. Would it be enough to generate just:
```java
public final class MyLayoutType extends ObjectType {
  public static final Layout INSTANCE;
}
```
for an interface `MyLayout`? The actual implementation of the layout would be `private static` in the `MyLayoutType` class. Or is there something missing that needs to be exposed in addition to that?

I know this is an incompatible change, so it needs to be introduced in two steps, but if we agree we can do it, it is better to do it before Truffle @ 0.12 is out.

truffle/com.oracle.truffle.api.object.dsl.test/src/com/oracle/truffle/api/object/dsl/test/ObjectTypePublicTest.java
chrisseaton: The test is to just check that the layout is visible, but that's a compile-time thing, so then I thought I might as well do something at runtime as well to really check it's the object I think it is.

truffle/com.oracle.truffle.object.dsl.processor/src/com/oracle/truffle/object/dsl/processor/LayoutGenerator.java
chrisseaton: It's currently protected because other layouts need to be able to make subclass it. It is indeed visible from outside, but now that's what we want. It's restricted to the package though so it's not visible enough.

chrisseaton:
Yes I think making the object type the key generated class, and having the layout instance field within that could work.

I see that as unrelated to the need that motivates this particular PR though. How about we merge this PR, and then I commit to making your suggested change afterwards?

jtulach:
OK.

PR #139, opened 2016-03-18
Truffle services

jtulach:
Works in orchestration with https://github.com/graalvm/graal-core/pull/135: rather than letting the graal-core to access `Accessor` instance directly, it introduced new package and (currently) three interfaces to let Graal and Truffle internals to talk to each other.

The short term goal is to make `Accessor` private again. The ultimate goal is to use the new mechanism instead of` TruffleRuntime`.


truffle/com.oracle.truffle.api.boot/src/com/oracle/truffle/api/boot/TruffleInfo.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Finds class of a {@link com.oracle.truffle.api.TruffleLanguage language} this node belongs
+     * to.
+     *
+     * @param node instance of {@link com.oracle.truffle.api.nodes.Node}
+     * @return class of the language that created the node
+     * @since 0.12
+     */
+    public abstract Class<?> findLanguage(Object node);
```
chumer: The language is always the same per RootNode. I'd prefer the node argument beeing a RootNode.

truffle/com.oracle.truffle.api.boot/src/com/oracle/truffle/api/boot/TruffleInfo.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Notifies the Truffle runtime that a call target has just been initialized.
+     *
+     * @param callTarget instance of {@link com.oracle.truffle.api.CallTarget}
+     * @since 0.12
+     */
+    public abstract void initializeCallTarget(Object callTarget);
```
chumer: Same here i'd prefer typed APIs.

truffle/com.oracle.truffle.api.boot/src/com/oracle/truffle/api/boot/LoopCountSupport.java
```diff
@@ -0,0 +1,52 @@
...
+ * @since 0.12
+ */
+public abstract class LoopCountSupport<Node> {
+    /**
+     * Constructor for subclasses.
+     *
+     * @param nodeClazz reference to {@link com.oracle.truffle.api.nodes.Node}
+     * @since 0.12
+     */
+    protected LoopCountSupport(Class<Node> nodeClazz) {
```
chumer: I don't understand the use of this constructor.

truffle/com.oracle.truffle.api.boot/src/com/oracle/truffle/api/boot/LoopCountSupport.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package com.oracle.truffle.api.boot;
+
+/**
+ * Special support for {@link com.oracle.truffle.api.nodes.LoopNode loop counting nodes}.
+ *
+ * @param <Node> instance of {@link com.oracle.truffle.api.nodes.Node}
+ * @since 0.12
+ */
+public abstract class LoopCountSupport<Node> {
```
chumer: Its very confusing to use Node as a type variable. Please remove the type variable here.

chumer:
As far as I can see 
LoopCountSupport is used to implement special services for the Truffle API implemented by the runtime.
TruffleInfo is used to implement special services by the Truffle API to the runtime.

My suggestion for naming is LoopCountSupport -> RuntimeServices and TruffleInfo -> TruffleAPIServices. This should make it clearer and the name abstract enough for future extensions. Similar names with a similar abtraction level are fine with me as well.

jtulach:
`Node` as type variable is confusing. Right. Using `Object` instead of `RootNode` is not good either. True. But I cannot reference real `Node`, `RootNode` and `CallTarget` classes as this modules compiles sooner than `com.oracle.truffle.api` module compiles. Thus I resorted to use of type variables and/or `Object`.

chumer:
> Node as type variable is confusing. Right. Using Object instead of RootNode is not good either. True. But I cannot reference real Node, RootNode and CallTarget classes as this modules compiles sooner than com.oracle.truffle.api module compiles. Thus I resorted to use of type variables and/or Object.

Why does it have to be a separate module? Just make runtime (both-ways) support is part of the "Truffle-core" API?

> Btw. I would prefer to consistently use type variables as with singletonizer the dependencies between Graal runtime and Truffle API might be significantly reduced in the future.

I'd prefer to not do this prematurely. It also introduces a lot of casts everywhere. We can still evolve and delete APIs after releases.

woess:
Naming: is `truffle.api.boot` a good package name? how about `truffle.api.support`?
Same for `TruffleInfo`, the name is too generic imho.

jtulach:
Support is used in NetBeans as name for APIs that are generally helpful, but not essential for usage of the system. While the "non-essential" property is correct in this case, the "generally helpful" isn't true. These APIs are only for the minority of users. A cryptic name like `jvmci` would be the best.

truffle/com.oracle.truffle.api.boot/src/com/oracle/truffle/api/boot/TruffleServices.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Obtains programmatic name of this implementation.
+     *
+     * @return the name identifying this implementation
+     * @since 0.12
+     */
+    public final String getName() {
+        return name;
+    }
```
woess: why is this here? is there any use for it?

woess:
Since it's not really an API, you can drop the `.api` part of the package name.

jtulach:
Yes, having the package out of `.api` segment would be great. But if I put the code into `com.oracle.truffle.api` project as @chumer suggests, then `mx` will mandate the code to be under the `com.oracle.truffle.api` namespace.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
```diff
@@ -167,8 +172,91 @@ protected LoopNode() {
...
+        }
+    }
+
+    /**
+     * An interface between Truffle API and hosting virtual machine. Not interesting for regular
+     * Truffle API users. Acronym for Truffle Virtual Machine Compiler Interface.
+     * 
+     * @since 0.12
+     */
+    protected abstract class TVMCI {
```
jtulach: Using `TruffleRuntime.getCapability` allows me to expose just this one class in the API. In future there can be more of such classes, if needed.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
```diff
@@ -167,8 +172,91 @@ protected LoopNode() {
...
+
+            @Override
+            public RepeatingNode getRepeatingNode() {
+                throw new UnsupportedOperationException();
+            }
+        };
+
+        NoOp() {
+            NO_OP_NODE.super();
+        }
```
woess: what do you need `NO_OP_NODE` for?

jtulach: By having `TVMCI` a non-static member class of LoopNode, it disappears from Javadoc. As `TVMCI` is an implementation detail, having it hidden is a laudable goal. We are using simpler approach with Truffle `Builder` classes.

I agree the trick with `NO_OP_NODE` is nasty, but it is just an implementation detail.

jtulach:
The API changes report looks as expected:
```
Added Classes

com.oracle.truffle.api.nodes.LoopNode$TVMCI

Added Nested Classes

com.oracle.truffle.api.nodes.LoopNode:  
inner protected abstract com.oracle.truffle.api.nodes.LoopNode$TVMCI
```

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
woess: I'm not very fond of such trickery and I hope we can get rid of it later. If it's just for Javadoc, then maybe we can find another way to hide these classes from Javadoc. Anyway, it's out of scope for this PR, so +1.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
```diff
@@ -167,8 +172,95 @@ protected LoopNode() {
...
     public static void reportLoopCount(Node source, int iterations) {
         if (CompilerDirectives.inInterpreter()) {
-            ACCESSOR.onLoopCount(source, iterations);
+            if (SUPPORT == null) {
+                final TruffleRuntime runtime = Truffle.getRuntime();
+                SUPPORT = runtime.getCapability(TVMCI.class);
+                if (SUPPORT == null) {
+                    SUPPORT = new NoOp();
+                }
+            }
```
chumer: What would speak against lazy initialization in the static initializer and make SUPPORT final?
At the moment this SUPPORT field accesses are not thread-safe. We would need double checked locking and SUPPORT to become volatile. If possible we should avoid this.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
chumer: This is actually problematic in the sense that we now load multiple implementations of LoopNode in the VM. There should be only one implementation loaded. 

Can't we just exclude this one class from javadoc instead in a separate PR and make it static as it should be?

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
```diff
@@ -107,6 +110,8 @@
  * @since 0.8 or earlier
  */
 public abstract class LoopNode extends Node {
```
chumer: Why is TVMCI in LoopNode? Having TMVCI#initializeCallTarget  in LoopNode does not feel right. Couldn't we introduce a abstract base class for TruffleRuntime instead and do the same thing with TVMCI there?

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
woess: That would be good. LoopNode.reportLoopCount is a frequently called method.

jtulach:
OK, thanks. I prepare changes to address you comments tomorrow.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
jtulach: OK, make final in https://github.com/graalvm/truffle/pull/139/commits/d0ee89755dab0592ecc15769039a570daffdf4c8

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
jtulach: Good thing is that https://github.com/graalvm/truffle/pull/139/commits/d0ee89755dab0592ecc15769039a570daffdf4c8 removes this *object oriented fallback machinery* all together and replaces it with simple `if (SUPPORT == null)` check.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/LoopNode.java
jtulach: Right, `TVMCI.initializeCallTarget` isn't perfect. On the other hand, I don't want to introduce another TVMCI class right now. Here is possible solution: https://github.com/graalvm/truffle/pull/139/commits/f73c774218806694b54b8ad7925a53c2ce7de121 - e.g. let's rename the method to `installRootNode` and then it starts to be more appropriate. The method works on RootNode and thus the method is now in the right (e.g. `com.oracle.truffle.api.nodes`) package.

truffle/com.oracle.truffle.api.boot/src/com/oracle/truffle/api/boot/TruffleServices.java
jtulach: This was the implementation of `TruffleRuntime.getName()` in `GraalTruffleRuntime` that extended `TruffleServices` and implemented `TruffleRuntime`. Nothing to worry about anymore - this is all gone.

jtulach:
Accepting Christian's work.

truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/impl/TVMCI.java
```diff
@@ -0,0 +1,84 @@
...
+     * Only useful for virtual machine implementors.
+     *
+     * @since 0.12
+     */
+    protected TVMCI() {
+        // export only for com.oracle.graal.truffle and com.oracle.truffle.api.impl
+        assert getClass().getPackage().getName().equals("com.oracle.graal.truffle") || getClass().getPackage().getName().equals("com.oracle.truffle.api.impl");
+    }
+
+    // the class is a member class to not be (easily) visible in Javadoc
```
woess: Outdated comment. I'm also not very happy with hardcoding package names like that.

jtulach: Please merge #148 once verified.

Re. hardcoding: Yes, it may complicate refactorings. On the other hand, if there is a change in packaging, the unit tests are likely to discover that quite easily. I like that such check really creates "friend API" - e.g. enforces the list of "friends" that can implement the class.

PR #141, opened 2016-03-21
Update Javadoc of getExecutionContext()

entlicher:
The ExecutionContext does not contain getLanguageShortName() method, therefore it's reference should be removed from Javadoc.


jtulach:
Another example of samples that are not verified by compilation.

PR #144, opened 2016-03-22
REPL debugger now launches polyglot

mlvdv:
- Reconfigure REPL debugger to start from command line client in polyglot mode
- Remove language-specific instantiation of REPL
- Rename SL-specific launcher from "mx sldebug" to "mx repl"
- Reconfigure launch so all languages on the class path are offered as alternatives
- REPL client now offers user initial choice of default language if more than one available
- Remove project com.oracle.truffle.sl.tools


jtulach:
I like the change.

PR #145, opened 2016-03-22
REPL debugger:  Deprecate public classes for removal from Truffle API

mlvdv:
Add a warning to com.oracle.truffle.tools.debug.shell.package-info.java and deprecate the two public names; they will be be removed from the Truffle public API.


mlvdv:
Changes merged into #147

PR #146, opened 2016-03-22
Consistently dispose PolyglotEngine after being used in tests

jtulach:
Consistently dispose PolyglotEngine after being used in tests. Give TruffleTCK and its implementor a chance to co-operate on construction of new PolyglotEngine.

Yesterday Andreas @Woess noted that the change proposed in #85 is huge. I took a look and yes, it seems huge as it contains a lot of fixes in tests. These fixes are however independent of the #85 change - they just consistently dispose `PolyglotEngine` after it has been used. I believe this kind of robustness improving changes can go in without any delay. Thus I am starting this pull request.

In addition to that there is one change in `TruffleTCK` to allow the TCK and the tested language to co-operate on creation of a `PolyglotEngine` - such factory is needed not only for #85, but also to test import of symbols - the TCK needs to pre-register them into the builder which will be possible with this new method.

### DIFF
--- a/truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/utilities/InstrumentationTestMode.java
+++ b/truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/utilities/InstrumentationTestMode.java
@@ -33,8 +33,8 @@ public class InstrumentationTestMode {
 
     public static void set(boolean enable) {
 
+        final PolyglotEngine vm = PolyglotEngine.newBuilder().build();
         try {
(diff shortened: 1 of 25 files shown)


jtulach:
Thanks @chrisseaton, your Ruby check in Truffle Travis seems useful: https://travis-ci.org/graalvm/truffle/builds/117651536

jtulach:
With my latest changes the `RubyTckTest` succeeds with `Tests run: 71, Failures: 0, Errors: 0, Skipped: 0`, so I think Ruby is fine. I'll check other Tck implementations.

chumer:
Looks good to me.

PR #147, opened 2016-03-23
Deprecate old instrumentation and REPL classes

mlvdv:
- @Deprecate all classes in the old instrumentation framework, along with tests, utilities
- @Deprecate the two REPL classes that don't belong in the Truffle API


jtulach:
The gate checks are failing. Assigning back.

mlvdv:
On my machine this PR builds, runs, and passes units tests cleanly with no complaints from Eclipse.  However I have not been able to get it past the gate.  Running the gate on my machine produces the same results as the Travis build.

The gate reports `Unnecessary @SuppressWarnings("rawtypes")` at lines 508 and 559 in the _old_ Instrumenter class:  `com.oracle.truffle.api.instrument.Instrumenter`

However, when those warnings are removed:
- Eclipse complains that they are needed and
- the gate reports that fully qualified references to `com.oracle.truffle.api.instrument.Instrumenter` "cannot be resolved to a type"

woess:
what's the reason for fully qualifying all deprecated class names?

mlvdv:
There appears to be no way to suppress warnings (-> gate errors) on imports of deprecated classes.

PR #149, opened 2016-03-25
Patch:  Remove com.oracle.truffle.api.test.JavaRecognizer

mlvdv:
it doesn't seem to be needed for any tests and it triggers a problem running the new "mx repl" command. A more general fix for "mx repl" will be needed.


jtulach:
If the checks pass and the SourceTest is OK on Mac, feel free to proceed with integration.

PR #153, opened 2016-03-29
Added Debugger.pause() and Debugger.isExecuting() methods.

entlicher:
`ExecutionEvent` is made disposable. When outside of event handler process, all it's methods should throw `IllegalStateException`, according to javadoc. Also, I've deprecated `prepareContinue()` method, since continue is by default.
To add a "pause" functionality (which could be achieved by calling `ExecutionEvent.prepareStepInto()` any time during language execution, but this is not possible any more), `Debugger.pause()` method is introduced. Also, `Debugger.isExecuting()` is added so that we can test whether we can expect that `pause()` will actually pause anything.
Test is included.

### DIFF
--- a/truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/Debugger.java
+++ b/truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/Debugger.java
@@ -191,6 +191,7 @@ public final class Debugger {
      * Head of the stack of executions.
      */
     private DebugExecutionContext currentDebugContext;
+    private volatile boolean executing;
 
     /**
      * Sets a breakpoint to halt at a source line.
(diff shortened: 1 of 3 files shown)


truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/ExecutionEvent.java
```diff
@@ -87,9 +90,11 @@ public Debugger getDebugger() {
      * </ul>
      * 
      * @since 0.9
+     * @deprecated Continue is by default, this method is not necessary.
      */
```
jtulach: I don't think it is necessary to deprecated this method. I know continue is the default action, but having the method allows people to "change their mind". It keeps some symmetry and also advertises that debugger can "continue".

I suggest to integrate the change without this deprecation.

truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/Debugger.java
```diff
@@ -234,6 +235,29 @@ public Breakpoint setTagBreakpoint(int ignoreCount, SyntaxTag tag, boolean oneSh
...
+        return executing;
+    }
+
+    /**
+     * Request a pause.
+     * As soon as the execution arrives at a node holding a debugger tag,
+     * {@link SuspendedEvent} is emitted.
+     * @since 0.13
+     */
+    public void pause() {
```
jtulach: Description of a threading behavior of this method (which is different than for other methods) is necessary.

truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/ExecutionEvent.java
entlicher: Well, you have just a single event handler callback to "change your mind", thus I considered this method unnecessary.
But O.K. if you'd like to keep it, I'll change the code accordingly.

truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/Debugger.java
entlicher: I'll improve it.

entlicher:
I've updated the pull request and addressed @jtulach comments.
Also, the `SLDebugTest.testPause()` is made more robust.

jtulach:
I am OK with this change from an API and test perspective.

entlicher:
Debugger.pause() is there already, but is replaced with DebuggerSession.suspendNextExecution().

PR #164, opened 2016-04-01
A BreakpointTag introduced, debugger creates a default breakpoint on it

entlicher:
There is no support for `debugger` statement, known e.g. from JavaScript, yet.
This PR introduces a new tag for this purpose - `BreakpointTag` - to mark locations where the code should always halt on a breakpoint. Debugger submits a breakpoint for this tag by default so that it's always hit.
To test it, there is an implementation of `debugger` statement in SL. Please review if this implementation is correct, it just mimics implementation of other statements.
A test is included.


mlvdv:
Looks good.  The tag should be defined in the Debugger, however, since it is intended to affect only the behavior of that one tool. There will be more tool-specific tags needed, including for the Debugger.

mlvdv:
It looks like you modified the scanner and parser by hand.  The are actually generated by the Coco/R parser generator (not really supported any more) from grammar in `SimpleLanguage.atg`.

entlicher:
Thanks for the warning about generated parser. I've forgot to check it out.
I've change the PR into two commits - one introduces `DebuggerTags.BreakpointTag` with a default breakpoint submitted on it and the other adds `debugger` statement into SL.
Is the simple language described somewhere? `debugger` statement would have to be added into the description.

mlvdv:
Suggest name "HaltTag" instead of "BreakpointTag" in order to avoid confusion.  "AlwaysHaltTag" would be even more clear.

jtulach:
I'd like to point out the duplication of word tag in `DebuggerTag.AlwaysHaltTag`. One solution would be put the nested class `AlwaysHaltTag` to already existing `Debugger` class. Then the usage would read as `Debugger.AlwaysHaltTag` and I think it would be more straightforward.

entlicher:
StandardTags inner classes contain the "Tag" as well. But I agree that it's superfluous.
I'd stay with a special "DebuggerTags" class, because this is for wider, or possibly different, audience than Debugger class. We can have DebuggerTags.AlwaysHalt - that sounds good to me, what do you think?

jtulach:
OK.

entlicher:
BreakpointTag renamed to AlwaysHalt. Also rebased to master. Thanks.

entlicher:
The gate check has failed, but in bunch of files that I did not touch. Someone should take care of it...

smarr:
Truffle has not yet been formatted with the latest Eclipse formatter, which is used by Travis. It probably needs to be fixed by someone running the formatter on everything.

entlicher:
The AlwaysHalt tag is integrated already.

PR #165, opened 2016-04-01
Add support for incremental compilation to Instrument registration

smarr:
- load existing registration file
- check whether a tool with the same id was registered before, and then updates it's set of entries

This does not clean up the file by removing entries for changed or removed tool ids, but it is probably sufficient because a clean rebuild will do that. For incremental compilation this seems sufficient.

Signed-off-by: Stefan Marr <git@stefan-marr.de>

### DIFF
--- a/truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
+++ b/truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
@@ -57,24 +57,34 @@ public final class InstrumentRegistrationProcessor extends AbstractProcessor {
         return SourceVersion.latest();
     }
 
+    private static final int NUMBER_OF_PROPERTIES_PER_ENTRY = 4;
+
     private void generateFile(List<TypeElement> instruments) {
         String filename = "META-INF/truffle/instrument";
         Properties p = new Properties();
+        int numInstruments = loadIfFileAlreadyExists(filename, p);
 
-        int cnt = 0;
         for (TypeElement l : instruments) {
             Registration annotation = l.getAnnotation(Registration.class);
             if (annotation == null) {
                 continue;
             }
-            String prefix = "instrument" + ++cnt + ".";
+
+            int instNum = findInstrument(annotation.id(), p);
+            if (instNum == 0) { // not found
+                numInstruments += 1;
+                instNum = numInstruments;
+            }
+
+            String prefix = "instrument" + instNum + ".";
             String className = processingEnv.getElementUtils().getBinaryName(l).toString();
+
             p.setProperty(prefix + "id", annotation.id());
             p.setProperty(prefix + "name", annotation.name());
             p.setProperty(prefix + "version", annotation.version());
             p.setProperty(prefix + "className", className);
         }
-        if (cnt > 0) {
+        if (numInstruments > 0) {
             try {
                 FileObject file = processingEnv.getFiler().createResource(StandardLocation.CLASS_OUTPUT, "", filename);
                 try (OutputStream os = file.openOutputStream()) {
@@ -83,6 +93,31 @@ public final class InstrumentRegistrationProcessor extends AbstractProcessor {
             } catch (IOException e) {
                 processingEnv.getMessager().printMessage(Kind.ERROR, e.getMessage(), instruments.get(0));
             }
+        }
+    }
+
+    private static int findInstrument(String id, Properties p) {
+        int cnt = 1;
+        String val;
+        while ((val = p.getProperty("instrument" + cnt + ".id")) != null) {
+            if (id.equals(val)) {
+                return cnt;
+            }
+            cnt += 1;
+        }
+        return 0;
+    }
+
+    private int loadIfFileAlreadyExists(String filename, Properties p) {
+        try {
+            FileObject file = processingEnv.getFiler().getResource(
+                            StandardLocation.CLASS_OUTPUT, "", filename);
+            p.load(file.openInputStream());
+
+            return p.keySet().size() / NUMBER_OF_PROPERTIES_PER_ENTRY;
+        } catch (IOException e) {
+            // Ignore error. It is ok if the file does not exist
+            return 0;
         }
     }
 


truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
```diff
@@ -57,24 +57,34 @@ public SourceVersion getSupportedSourceVersion() {
...
+            String prefix = "instrument" + instNum + ".";
             String className = processingEnv.getElementUtils().getBinaryName(l).toString();
+
             p.setProperty(prefix + "id", annotation.id());
             p.setProperty(prefix + "name", annotation.name());
             p.setProperty(prefix + "version", annotation.version());
             p.setProperty(prefix + "className", className);
         }
-        if (cnt > 0) {
+        if (numInstruments > 0) {
```
chumer: If you just change an entry in @Registration it will not get updated anymore.

truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
```diff
@@ -57,24 +57,34 @@ public SourceVersion getSupportedSourceVersion() {
         return SourceVersion.latest();
     }
 
+    private static int NumPropertiesPerEntry = 4;
```
chumer: static final. naming convention. how did this go through the gate?

truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
```diff
@@ -86,6 +96,31 @@ private void generateFile(List<TypeElement> instruments) {
         }
     }
 
+    private static int findInstrument(String id, Properties p) {
+        int cnt = 1;
+        Object val;
+        while ((val = p.get("instrument" + cnt + ".id")) != null) {
```
chumer: use p.getProperty(String) instead. change local var from Object to String.

truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
```diff
@@ -86,6 +96,31 @@ private void generateFile(List<TypeElement> instruments) {
...
+            if (id.equals(val)) {
+                return cnt;
+            }
+            cnt += 1;
```
chumer: Nitpicking: This is Java we can use cnt++ ;-)

truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
smarr: Ehm, why?
My intent with this code was: 
 - search for a matching entry by `id`
 - if it is found, overwrite its current values

Why does that not work for changing a property for the annotation?

truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
smarr: I dislike `++`. Do you want me to change this?

smarr:
Fixed the missing `final` and changed to `String`/`getProperty()`

truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
chumer: You are right. Sorry.

truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
chumer: You optimize for the normal Java reader and not for yourself. i'd guess that cnt++ reads more common for them. I might be wrong. Feel free to decide yourself.

truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
```diff
@@ -57,24 +57,34 @@ public SourceVersion getSupportedSourceVersion() {
         return SourceVersion.latest();
     }
 
+    private static final int NumPropertiesPerEntry = 4;
```
chumer: NUMBER_OF_PROPERTIES_PER_ENTRY

chumer: Thats the Java convention for constants. They should also be defined on top of the class.

smarr: Yeah, I had that before, and thought is was a style violation because checkstyle complained. But the `final` was missing... will fix.

chumer:
Thanks for the fix! Did you test this in eclipse? Last time I tried something like this the meta-inf file was cleaned each time I changed a java class with a @Registration annotation.

So does the following test case now work?
1) Put a space in CoverageExample and save in Eclipse
2) Put a space in DebuggerExample and save in Eclipse
3) Run mx unittest directly on a shell.

smarr:
Ehm, that was the whole point of the exercise.
But to be sure, yes, I followed you steps and `mx unittest` passes.

In SOMns, I got multiple tools, and as soon as one is not registered, I would get errors on startup. That's why using a work around with cleaning or something was just not convenient.

chumer:
I just wondered what I did different last time. I had a very similar version some time ago, that I discarded...

smarr:
Hm, don't know. Works so far here. But that's also my first time messing with that kind of stuff.

truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/InstrumentRegistrationProcessor.java
```diff
@@ -57,24 +57,34 @@ public SourceVersion getSupportedSourceVersion() {
         return SourceVersion.latest();
     }
 
+    private static final int NUMBER_OF_PROPERTIES_PER_ENTRY = 4;
```
woess: can we rely on the number of properties per entry to stay the same?

smarr: Don't know. Using a named constant was the simplest solution I could think off to know how many instruments are in a file (line 117). Otherwise, I'd have to count id entries or something. Might be more robust, but it's also more code to maintain.

woess: Ok, assuming it's unlikely to change any time soon, we can change this code later if need be.

smarr:
@chumer I rebased the PR. Could you please take care of the merge?

woess:
Merged. Please close.

smarr:
Merged as ebe252dae099a1256caa0039a9df13d0e3a7e4b4

PR #166, opened 2016-04-04
EventContext#findParentEventNode and findDirectParentEventNode

smarr:
This PR is based on work of @chumer, and I created it to figure out what would need to be done to get this functionality in.

I care mostly for findParentEventNode and findDirectParentEventNode.
I use those two to hook up instrumentation for related AST nodes.

/cc @jtulach


chumer:
Lazy factory node creation is an implementation detail and is exposed using this feature. It needs to materialize all the child and parent event nodes when asked. This needs more work.

smarr:
Sorry, I don't know what you want to tell me, and I also don't know what you would expect me to change. Could you please elaborate?

chumer:
We have introduced the notion of input value instrumentation in 0.32 that allows to listen to child input values. In a future version we will introduce navigation between instrumented nodes. The idea is that we want to expose an instrumentation view on the AST. That exposes only instrumentable nodes.

Closing this, as its hopelessly outdated. But we did not forget about this requirement.

smarr:
In case anyone needs this, there is an updated copy maintained here: https://github.com/smarr/truffle/tree/debugger/master

Not sure what to understand under "input value instrumentation".
Will need to look into it some point.

PR #167, opened 2016-04-05
Feature Request: `run` command for REPL/debugger and `PolyglotEngine.start()` 

smarr:
This is more of a feature request than an actual pull request. So, please consider the attached code merely as a sketch.

I would like to discuss two features:
 - a `run` command for the REPL/debugger
 - `PolyglotEngine.start()` to initialize execution based on the given engine configuration

### run

For users of GDB this might be familiar as `gdb --args foo bar baz` and then using `run` in GDB.

The idea is to execute a program as if it was started normally from the command line.

My main reason for this is to avoid having to revamp my whole language implementation to be compatible with the repl/debugger. I just want to use the debugger features.

What's currently missing to make this work is a way to initialize execution from the REPL.

### start

The Polyglot engine only offers an `eval()` but no `start()` method, which would be very helpful to trigger standard execution based on the configuration provided to the engine.
This means, it is not the polyglot engine anymore that needs to determine an initial file for evaluation. Instead, it is up to the language implementation to decide how to run code.
I wonder whether other languages also currently work around polyglot to realize that (/cc @lukasstadler @chrisseaton).

The way I currently work around this is by having a magic "START" source.

Ping: @mlvdv @jtulach

### DIFF
--- a/truffle/com.oracle.truffle.tools.debug.shell/src/com/oracle/truffle/tools/debug/shell/server/REPLServer.java
+++ b/truffle/com.oracle.truffle.tools.debug.shell/src/com/oracle/truffle/tools/debug/shell/server/REPLServer.java
@@ -55,6 +55,7 @@ import com.oracle.truffle.api.source.Source;
 import com.oracle.truffle.api.source.SourceSection;
 import com.oracle.truffle.api.vm.EventConsumer;
 import com.oracle.truffle.api.vm.PolyglotEngine;
+import com.oracle.truffle.api.vm.PolyglotEngine.Builder;
 import com.oracle.truffle.api.vm.PolyglotEngine.Language;
 import com.oracle.truffle.api.vm.PolyglotEngine.Value;
 import com.oracle.truffle.tools.debug.shell.client.SimpleREPLClient;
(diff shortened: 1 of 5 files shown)


truffle/com.oracle.truffle.tools.debug.shell/src/com/oracle/truffle/tools/debug/shell/server/REPLServer.java
```diff
@@ -203,6 +209,17 @@ public LocationPrinter getLocationPrinter() {
         return locationPrinter;
     }
 
+    void run() {
+        try {
+            Source source = Source.fromNamedText("", "START");
+            source = source.withMimeType(getLanguages().iterator().next().getMimeTypes().iterator().next());
```
jtulach: Execute empty source! That is a cute trick.

smarr: Cute trick? I find that a rather ugly hack ;-P

chumer: I did the same in the multi-language REPL.

jtulach:
```java
Source source = Source.fromNamedText("", "START");
getLanguages().iterator().next().eval(source)
```
might "work" as well. But I write "work" in quotes as the code doesn't even know what language is being initialized! The whole approach is wrong. The API hasn't been designed to support this use-case. Alas, it hasn't be designed to make it impossible either.

mlvdv:
One bit of good news is that the REPL will soon be abandoned in favor of a multi-language shell implementation with debugging commands available. It will have the same functionality at first, but the implementation will much easier to read and modify.

chumer:
@smarr  is this question now answered for you? Please close if so.
As we have discussed elsewhere TruffleLanguage should be implemented in a way so it can be used in a language agnostic way.

smarr:
I'll continue using my hack, and the addition of the `run` command in my fork. In case someone wants to pick up the web-based debugger, this might need to reconsidered, at least with respect to the `run` command in the debugger.

PR #168, opened 2016-04-05
Feature Request: Support for on-demand instrumentation

smarr:
Note, this is more of a feature request than an actual pull request. Please consider the attached code merely as a sketch/hack.

Support for on-demand instrumentation is useful for dynamic languages where execution leads to node
replacement, which requires new instrumentation. Node classes can be tagged with all kind of tags. Currently, wrapping is only done when starting the execution of a method. So, this API here allows the language to notify the instrumentation framework about relevant new nodes that might need to be wrapped.

Again, this change is a _hack_, and not supposed to be merged directly.

Handled special cases:
- unwrap wrapper nodes when inserting new wrappers
- when there is already a wrapper node, just invalidate it

Ping: @chumer @mlvdv @jtulach

### DIFF
--- a/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentationHandler.java
+++ b/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentationHandler.java
@@ -56,10 +56,12 @@ import com.oracle.truffle.api.source.SourceSection;
  * Central coordinator class for the Truffle instrumentation framework. Allocated once per
  * {@linkplain com.oracle.truffle.api.vm.PolyglotEngine engine}.
  */
-final class InstrumentationHandler {
+public final class InstrumentationHandler {
 
     /* Enable trace output to stdout. */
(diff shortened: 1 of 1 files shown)


mlvdv:
This would be a very useful feature and would address one of the strongest constraints in place around instrumentation since the beginning.

chumer:
Yes notifying about changes (SourceSection and/or Tags) in the AST might be useful. 
Manual insertion of wrappers is not something that I think is a good idea because it breaks our abstractions.

arthurp:
Sorry if this is a necro-post, but:

@chumer What about a call which triggers re-wrapping of all nodes in a RootNode. This would mean that a wrapper could never be missed by failing to call a low-level insertWrapper operation, but it would still allow arbitrary changes to the AST. Obviously this would be much heavier weight operation, but in some cases I think it would definitely be worth it. For instance, I use instrumentation to insert profiling code on call nodes, but in my language call nodes are dynamically created when a future force is eliminated at runtime (the force is effectively optimized into a call).

chumer:
Yes notifyInserted was designed for this.

PR #169, opened 2016-04-05
SourceSectionFilter should also error on empty array of tags

smarr:
In the `SourceSectionFilter`, `verifyNotNull(arr)` is used to make sure that not illegal arguments are used for filtering.

Currently, it checks for `null` values only, but ignores empty arrays. My understanding is that empty arrays are problematic here, because the `Not` filter merely negates the matching filter, which does not have any matches for empty arrays. So, I'd assume that empty arrays here are usually bugs, similar to passing `null`.

Ping: @chumer

### DIFF
--- a/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/SourceSectionFilter.java
+++ b/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/SourceSectionFilter.java
@@ -362,7 +362,7 @@ public final class SourceSectionFilter {
         }
 
         private void verifyNotNull(Object[] values) {
-            if (values == null) {
+            if (values == null || values.length == 0) {
                 throw new IllegalArgumentException("Given arguments must not be null.");
             }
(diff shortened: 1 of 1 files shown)


truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/SourceSectionFilter.java
```diff
@@ -362,7 +362,7 @@ public SourceSectionFilter build() {
         }
 
         private void verifyNotNull(Object[] values) {
-            if (values == null) {
+            if (values == null || values.length == 0) {
                 throw new IllegalArgumentException("Given arguments must not be null.");
```
woess: Should be a different error message for `value.length == 0`. Also, the method name does not indicate that this check is done.

smarr: Yeah, and tests fail because of this change. So, for me the question is, is this here the intended behavior, or would there be a reason to allow for empty arrays?

PR #171, opened 2016-04-06
Feature Request: Added isTaggedWith(Node,Class<?>) to Instrumenter

smarr:
I would like to have `isTaggedWith()` on `Instrumenter`, similar to the already available `queryTags(Node)`.

For my dynamic metric tool, I am hooking up profiling objects on a group of related AST nodes.
I use `isTaggedWith()` during the setup to identify the child nodes tagged with the `Argument` tag so that I can hook them up with the profile object and record the values of the arguments separately from each other.

Note, I didn't add any tests/docs yet. First, I wanted to know whether this would be acceptable/useful.

Ping: @chumer @jtulach.

### DIFF
--- a/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentationHandler.java
+++ b/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentationHandler.java
@@ -616,6 +616,11 @@ final class InstrumentationHandler {
         }
 
         @Override
+        public boolean isTaggedWith(Node node, Class<?> tag) {
+            return ACCESSOR.isTaggedWith(node, tag);
+        }
+
+        @Override
         void verifyFilter(SourceSectionFilter filter) {
         }
 
@@ -713,6 +718,11 @@ final class InstrumentationHandler {
         @Override
         public Set<Class<?>> queryTags(Node node) {
             return queryTagsImpl(node, language.getClass());
+        }
+
+        @Override
+        public boolean isTaggedWith(Node node, Class<?> tag) {
+            return ACCESSOR.isTaggedWith(node, tag);
         }
 
         @Override
--- a/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/Instrumenter.java
+++ b/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/Instrumenter.java
@@ -75,4 +75,5 @@ public abstract class Instrumenter {
      */
     public abstract Set<Class<?>> queryTags(Node node);
 
+    public abstract boolean isTaggedWith(Node node, Class<?> tag);
 }


chumer:
Why does queryTags(Node) not work for that?

smarr:
Of course, `queryTags()` does work, but it is clunky and slow.
I use `isTaggedWith()` in the interpreter code, so, it would be great if it could be efficient. I still care about interpreter speed.

smarr:
Ping @chumer and @jtulach: this is still something I want.

While it does not introduce a new capability in addition to `queryTags()`, it has better performance properties, and is from my point of a few a natural complement to `queryTags()`.

What are the arguments against it?

mlvdv:
I would like to see this added.

jtulach:
The builder is failing. One error I see is:
```
stderr: /home/travis/build/graalvm/truffle/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/Instrumenter.java:78: warning - missing @since tag for com.oracle.truffle.api.instrumentation.Instrumenter.isTaggedWith(com.oracle.truffle.api.nodes.Node, java.lang.Class<?>)`
```

smarr:
Ok, so, generally this is acceptable? Then, I'll polish the PR. It also needs tests and documentation.

jtulach:
Providing simpler and more performant access to functionality that is already available is generally acceptable. However in this case I have to correct myself (a bit):

There is [Instrumenter.queryTags](http://lafo.ssw.uni-linz.ac.at/javadoc/truffle/latest/com/oracle/truffle/api/instrumentation/Instrumenter.html#queryTags-com.oracle.truffle.api.nodes.Node-) and it returns a `Set`. Rather than creating new method `isTaggedWith`, make sure the following code is fast:
```java
instrumenter.queryTags(node).contains(tagClass)
```
that can be done by lazy initialization of the `Set` and overriding `contains` to not initialize the whole `Set`, but query directly for the single class.

Such solution keeps the API simplicity of `Set` facade and just makes sure it behaves fast for the important use-case. An example to copy is [here](http://hg.netbeans.org/releases/file/71e2ea254889/openide.loaders/src/org/openide/loaders/DataNode.java#l1210) - it optimizes for `iterator()`, but the idea is the same.

mlvdv:
>Such solution keeps the API simplicity of Set facade and just makes sure it behaves fast for the important use-case.

I would be happy with this solution.

smarr:
Ok, so, this is again one of those moments, where I feel the proposed solution that is supposedly keeping the API simpler, is 1) not the simples possible solution from a user perspective, and 2) introduces an amount of complexity in the implementation that makes me wonder whether it is really worth it.

1) while the difference between `instrumenter.queryTags(node).contains(tagClass)` and `instrumenter.isTaggedWith(node, tagClass)` seems to be small, the naming makes for me quite a difference. The `isTaggedWith()` method is from my point of view easier to discover, especially from a language implementer's perspective.

2) a direct style implementation of `isTaggedWith()` seems to be orders of magnitudes simpler to understand and maintain than a custom lazy set implementation.

So, this is all just to say, could you please explain me a little more why the tradeoffs are here in favor for `instrumenter.queryTags(node).contains(tagClass)`?

jtulach:
Re. "makes me wonder whether it is really worth it."

Whether you want to polish your pull request, is more a question for you.

Re. "not the simples possible solution from a user perspective"

I believe >95% of Java users are well aware of the collection interfaces and know that the right way to check whether a tag is contained in a tag set is to call `tagSet.contains(tag)`. In the APIs the design of I oversight I try to reuse existing concepts of the underlying platforms - as that builds on already existing knowledge rather than requiring users to learn new and specific concepts. The same motivation leads me to prefer `Set.contains` over superfluous  `isTaggedWith` which just increases conceptual surface of what is already available via more standard API.

Re. "introduces an amount of complexity in the implementation"

Let me point out that now we are not talking about API users anymore. As there is just one API writer, and millions of API users, I believe it doesn't make sense to discuss optimizations visible only to API writers.

smarr:
Ok, let's revisit this once there is a tool that requires good performance in the interpreter. I don't see how your approach is going to win this one.

jtulach:
OK.

chumer:
The question of whether we use Set or another Method is not very important. 

The problem with this patch is that it ignores all the security checks that are necessary. See https://github.com/smarr/truffle/blob/2259bf5fcd72d194b6a776ee2e23ed18a81a93ba/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/InstrumentationHandler.java#L788 for all the checks that are ignored by this implementation.

Things to do before Node.isTaggedWith is called:
1) Check if a node is instrumentable .
2) Lookup if the instrumenter is allowed to query that node (using TruffleLanguage in the RootNode)
3) Check if the queried tag is actually a provided tag. (other tags are guaranteed to be not calling Node.isTaggedWith)

I think these checks dominate the performance of this method anyway.

PR #172, opened 2016-04-06
Fix race condition between Breakpoint construction and access from listeners

smarr:
This change fixes a race between the construction of a `BreakpointImpl` object and its access from another listener thread. This can happen easily if another thread related to a debugger UI creates a breakpoint while a program is executing.

Details:
- make BreakpointListener static class, because letting `this` escape to another thread during construction is just evil: I can't see that race when looking at the code
- use factory method for BreakpointImpl construction to register listener after object creation
- this relies on JVM semantics that ensures that all object fields are visible after object construction, in case at least one of the fields in final. Otherwise, we'd need to make the racy fields volatile.

- make assumption fields `@CompilationFinal` otherwise they are not properly optimized

Ping: @chumer @mlvdv

### DIFF
--- a/truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/BreakpointFactory.java
+++ b/truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/BreakpointFactory.java
@@ -157,7 +157,7 @@ final class BreakpointFactory {
         if (breakpoint == null) {
             final SourceSectionFilter query = SourceSectionFilter.newBuilder().sourceIs(lineLocation.getSource()).lineStartsIn(IndexRange.byLength(lineLocation.getLineNumber(), 1)).tagIs(
                             StandardTags.StatementTag.class).build();
-            breakpoint = new BreakpointImpl(lineLocation, query, ignoreCount, oneShot);
+            breakpoint = createBreakpoint(lineLocation, query, ignoreCount, oneShot);
             if (TRACE) {
                 trace("NEW " + breakpoint.getShortDescription());
             }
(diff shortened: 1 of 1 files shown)


smarr:
@entlicher this one might be also relevant for the NetBeans debugger.

mlvdv:
looks good.

smarr:
If this fix is acceptable, could please someone accept it and integrate it, @chumer @mlvdv @jtulach?
Thanks

truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/BreakpointFactory.java
```diff
@@ -253,6 +253,16 @@ private void forget(BreakpointImpl breakpoint) {
         breakpoints.remove(breakpoint.getKey());
     }
 
+    BreakpointImpl createBreakpoint(Object key, SourceSectionFilter query, int ignoreCount, boolean isOneShot) {
+        BreakpointImpl breakpoint = new BreakpointImpl(key, query, ignoreCount, isOneShot);
+        // Only register the listener after the breakpoint has been partially
+        // constructed. Otherwise, we can't be sure that the assumptions are visible.
```
eregon: Why `partially`? The BreakpointImpl will be "fully" constructed, isn't it?

smarr: will fix the comment

smarr:
Rephrased comment, and rebased on master.

entlicher:
The change looks good to me, thanks.

smarr:
Thanks.

PR #174, opened 2016-04-13
Fix handling of empty source sections

smarr:
This is work in progress. I'd like first to get feedback on the desired semantics.

Currently, `SourceSection.getEndLine()` does pass `-1` for an empty source section to `source.getLineNumber(offset)`, which causes an `IllegalArgumentException`. However, this does not seem to be intended.

Source.offsetToLine uses the following check:

`if (offset == 0 && textLength == 0) { return 1; }`

This seems to indicate that empty source sections should be supported.

If this is indeed the case, this should be further be tested. Other methods might not be correctly handling this either, for instance `getEndColumn()`.

/cc @chumer @jtulach @mlvdv

### DIFF
--- a/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/source/SourceSection.java
+++ b/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/source/SourceSection.java
@@ -172,6 +172,9 @@ public final class SourceSection {
      * @since 0.8 or earlier
      */
     public int getEndLine() {
+        if (source.getLength() == 0) {
+            return 1;
+        }
         return source.getLineNumber(charIndex + charLength - 1);
     }
 


jtulach:
Returning reasonable values is certainly preferred over throwing exceptions. However I don't feel qualified to decide whether 0 is reasonable value or not. In any case, associate the change with a unit test.

smarr:
@mlvdv perhaps that's more a question for you Michael? Are empty source sections supposed to be supported? There are some indications that they should be.

smarr:
@jtulach yes, there are more changes necessary here. If we know what's desirable, I'll add tests and make the rest consistent.

chumer:
I have no good answer either. But I could imagine that empty SourceSection's are required for some implicit language features. So I don't see a reason why we should not support it. The instrumentation should be fine with this.

smarr:
@mlvdv can we do something about this? This is one of the hacks I still use currently.

chumer:
current getEndLine implementation looks like this:
` public int getEndLine() {
        if (source == null) {
            return -1;
        }
        if (!isValid()) {
            return 1;
        }
        return source.getLineNumber(getCharIndex() + Math.max(0, getCharLength() - 1));
    }`

PR #176, opened 2016-04-15
Document guideline to design TruffleLanguage with language-agnostic initialization.

smarr:
A TruffleLanguage should provide a language-agnostic interface to be able to be usable in multi-language environments. Language-specific front-ends or launchers, for instance, emulating existing implementations, should be realized externally to the PolyglotEngine.

This change fixes also minor issues in existing comments.

### DIFF
--- a/truffle/com.oracle.truffle.api.vm/src/com/oracle/truffle/api/vm/PolyglotEngine.java
+++ b/truffle/com.oracle.truffle.api.vm/src/com/oracle/truffle/api/vm/PolyglotEngine.java
@@ -65,8 +65,8 @@ import com.oracle.truffle.api.source.Source;
 
 /**
  * Gate way into the world of {@link TruffleLanguage Truffle languages}. {@link #buildNew()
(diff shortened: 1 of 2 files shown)


jtulach:
Very nice. I want to accept the change.

jtulach:
Can you please re-format your changes, so they pass the gate?

smarr:
The travis gate is broken, because it is using the new formatter but Truffle isn't yet updated.
So, I am not really sure how to verify the formatting is correct.

jtulach:
I am not sure what you mean by "travis gate is broken"!? The only thing it does is to execute `mx gate` in truffle directory. I don't see a way for this to be broken. In any case I cannot accept your pull request until `mx gate` passes without errors. Closing until that gets fixed.

smarr:
It uses the Eclipse 4.5 formatter I think, while Truffle is still supposed to use the 4.4 formatter. Or am I mistaken?

PR #177, opened 2016-04-15
Feature Request: Enable debugging of multithreaded programs in a single PolyglotEngine

smarr:
To support debugging in scenarios with 1 polyglot engine with n threads, we need separate DebugExecutionContext for each thread. This is a first step however. The REPL would need further features to properly support this. Furthermore, the other elements of the debugger infrastructure are likely to have thread-safety issues, to. So, this is only a very first step and should provide a foundation for further work.

Furthermore, this should be considered a hack. Each additional thread beside the main thread needs to call `executionStarted()` and `executionEnded()` appropriately, manually. This might not be desirable in the long run. However, in my case, the relevant code is highly performance sensitive: I need to initialize the debugger properly for each and every asynchronous message processed. This should be considered when designing an API. Messages are processed in separation, which means they likely need separate `DebugExecutionContexts` making `executionStarted()` and `executionEnded()` highly performance critical.

This is likely related to changes in https://github.com/jtulach/truffle/tree/FasterExecute

Next steps?
This should probably be discussed to figure out all things that need to be considered to make it work for real.

Ping: @mlvdv @chumer @jtulach

### DIFF
--- a/truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/Debugger.java
+++ b/truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/Debugger.java
@@ -274,8 +274,8 @@ public final class Debugger {
      * </ul>
      */
     @TruffleBoundary
-    void prepareContinue(int depth) {
-        getCurrentDebugContext().setAction(depth, new Continue());
+    void prepareContinue(DebugExecutionContext context, int depth) {
+        context.setAction(depth, new Continue());
     }
 
     /**
@@ -295,11 +295,11 @@ public final class Debugger {
      * @throws IllegalArgumentException if the specified number is {@code <= 0}
      */
     @TruffleBoundary
-    void prepareStepInto(int stepCount) {
+    void prepareStepInto(DebugExecutionContext context, int stepCount) {
         if (stepCount <= 0) {
             throw new IllegalArgumentException();
         }
-        getCurrentDebugContext().setAction(new StepInto(stepCount));
+        context.setAction(new StepInto(stepCount));
     }
 
     /**
@@ -308,7 +308,8 @@ public final class Debugger {
      * <li>User breakpoints are enabled.</li>
      * <li>Execution will continue until either:
      * <ol>
-     * <li>execution arrives at the nearest enclosing call site on the stack, <strong>or</strong></li>
+     * <li>execution arrives at the nearest enclosing call site on the stack, <strong>or</strong>
+     * </li>
      * <li>execution completes.</li>
      * </ol>
      * <li>StepOut mode persists only through one resumption, and reverts by default to Continue
@@ -316,8 +317,8 @@ public final class Debugger {
      * </ul>
      */
     @TruffleBoundary
-    void prepareStepOut() {
-        getCurrentDebugContext().setAction(new StepOut());
+    void prepareStepOut(DebugExecutionContext context) {
+        context.setAction(new StepOut());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


smarr:
A second set of changes is added to always maintain execution context explicitly as part of the events. This is necessary to know which context is the relevant in a multithread/multi-context environment. And, as simple example, a UI thread can't access a thread-local context because it doesn't have one, and even if, it is not the correct one.

truffle/com.oracle.truffle.api.debug/src/com/oracle/truffle/api/debug/Debugger.java
```diff
@@ -274,8 +274,8 @@ public Breakpoint setTagBreakpoint(int ignoreCount, com.oracle.truffle.api.instr
      * </ul>
      */
     @TruffleBoundary
-    void prepareContinue(int depth) {
-        getCurrentDebugContext().setAction(depth, new Continue());
+    void prepareContinue(DebugExecutionContext context, int depth) {
```
jtulach: Why you need "DebugExecutionContext public" as your commit message says? It seems to me that all changes in this changeset are only modifying classes in `com.oracle.truffle.api.debug` package - e.g. no API changes are necessary.

smarr: Indeed, I think package-private is sufficient here. The original `private` won't suffice anymore however. Will change that. Thanks.

mlvdv:
This looks like a good start.  Yes, the `DebugExecutionContext` is the one piece of the current design that would thread-specific: its main job is to track segments of nested executions on a single thread (Truffle execution stack).  And yes, having some way to be notified when executions on that stack begin and end is needed to keep track of nested segments.
Global state includes both the current set of active breakpoints and the current stepping mode (e.g. StepIn, StepOver, etc).  Making those thread-specific would be a different issue.

jtulach:
The current thinking (based on the work in https://github.com/jtulach/truffle/tree/FasterExecute) is that one can *fork* an engine. E.g. create a clone that could be used in parallel on a different thread - the state might be isolated, yet the code would be shared. As a result of that:
```java
PolyglotEngine one = ...;
PolyglotEngine two = one.fork();
assert Debugger.find(one) == Debugger.find(two);
```
E.g. breakpoints would be shared between both engines (because they share the code).  The per thread state `DebugExecutionContext` would then have to be local per each `PolyglotEngine` / thread.

smarr:
Hm, forking a `PolyglotEngine`. What's the motivation behind that? Do you by chance have perhaps a list of things that are shared/not-shared between engines?

You say that state 'might' be isolated. Can I as a language implementer influence that somehow?

Thanks for the update.

ldaynes:
Motivation for forking with isolation includes be startup-time and fast resource reclamation. Consider this scenario: a server clones the request handler which already has all the scripts needed evaluated for responding to the request; once handled the new polyglot engine can be disposed off and it's resource could be reclaimed at once. The latter, to be timely, would require a mechanism like V8's isolate. 

I do not like the "might be isolated".  I'd rather have them completely isolated. If sharing is needed, then a multi-threading mechanism within the language is more appropriate.

smarr:
Ok, this means I probably want something else than `.fork()`. I want really shared memory multithreads, and likely with a single `PolyglotEngine`. At least I don't see at the moment, why I would want more than one.

smarr:
@chumer @jtulach @mlvdv any update in this direction?
What would be needed to make this change acceptable?

smarr:
As example for my concrete use case, I am doing rougly this:

```Java
// executed by threads of a thread pool
void processActorMessages(Debugger dbg) {
  for (EventualMessage msg : messageBuffer) {
    dbg.executionStarted(-1, msg.getTargetSourceSection().getSource());
    msg.execute();
    dbg.executionEnded();
  }
}
```

This means, a thread from a thread pool picks up a `messageBuffer`, and processes the messages one by one. To enable debugging, I need to initialize the debugger for each message separately to point it at the correct source.

smarr:
I think this is now supported.

PR #179, opened 2016-04-18
Update Truffle formatting to Eclipse 4.5.2

smarr:
The Travis gate uses already the new formatter and fails because of that.

The new formatter changes comment formatting slightly, as well as whitespace handling in some corner cases.

This patch is the result of running `mx gate` with ECLIPSE_EXE pointing to an Eclipse 4.5.2

/cc @chumer @woess @jtulach


chumer:
Merged: https://github.com/graalvm/truffle/commit/ece7af26966ccc12916e6aeab7c97a624ba95412

PR #187, opened 2016-06-22
Document that @Child and @Children imply @CompilationFinal

smarr:
This is something that came up as a question on the gitter channel. Seems worthwhile documenting.

/cc @jtulach


truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/nodes/Node.java
```diff
@@ -58,6 +58,8 @@
     /**
      * Marks array fields that are children of this node.
      *
+     * This annotation implies the semantics of {@link CompilationFinal}.
```
woess: more specifically, the semantics of `@CompilationFinal(dimensions = 1)`

smarr:
I added the `dimensions` note to the `@Children` javadoc. @woess, is it now good to go?

PR #190, opened 2016-07-01
Added source-section-based breakpoints

smarr:
This complements line-based breakpoints by allowing a more precise selection of where to interrupt execution.

The implementation uses source section equality (with the unchanged filter mechanism) to identify where the breakpoint applies.

TODO: update `@since` for corresponding release

NOTE: I'm also thinking about adding another filter option by tags so that the source section and the tag needs to match.

/cc @jtulach @mlvdv


mlvdv:
Yes, it will be necessary in general to be able to specify filters both SourceSection and tag in some situations.  I ran into this in my prototyping.

chumer:
In my latest changes for the Debugger, I changed the breakpoint creation to builder style. I also included support for creating SourceSection based breakpoints there. ETA: next 1-3 weeks

chumer:
You can use

smarr:
Ok, thanks, I'll probably not be able to try it earlier than in two weeks.

PR #192, opened 2016-07-01
Make generated Interop classes public

smarr:
The classes generated by the other DSL processors are public.
This change makes the Interop DSL consistent with that and avoids requiring strange workarounds or putting undesirable restrictions on how to divide code into packages.

### DIFF
--- a/truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/interop/ForeignAccessFactoryGenerator.java
+++ b/truffle/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/interop/ForeignAccessFactoryGenerator.java
@@ -70,7 +70,7 @@ public final class ForeignAccessFactoryGenerator {
         Writer w = factoryFile.openWriter();
         w.append("package ").append(packageName).append(";\n");
         appendImports(w);
-        w.append("final class ").append(simpleClassName);
+        w.append("public final class ").append(simpleClassName);
         w.append(" implements Factory10, Factory {\n");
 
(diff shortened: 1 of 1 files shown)


jtulach:
I do rely on the classes being non-public in my [HTML/Java codebase](http://source.apidesign.org/hg/html~html4j/diff/47cd2110ed8d/boot-truffle/src/main/java/net/java/html/boot/truffle/JavaObject.java#l1.55). All the implementation classes are non-public and I certainly don't want the annotation processor to generate anything public in this case.

Some kind of hint or configuration is needed to support both scenarios, in my opinion.

smarr:
The current behavior is inconsistent compared to the other annotation processors, so Truffle itself already sets another expectation. (Note also that I only made the classes public that need to be accessible, there are true implementation classes that are not made public.)

eregon:
@jtulach This is unreasonable. It's generated code, there is no point to see that in documentation or ever be checked-in in a repository. It should be made convenient, like the DSL generated nodes.
`JavaObject` can remain package-private of course.

eregon:
One compromise could be to generate a package-private class if the source class is package-private.
But it would neither be very intuitive, nor particularly simple for the user.
I value (as probably most users) those much more than exposed classes.
Simplicity is also important in the implementation (so introducing e.g. a new annotation to specify visibility would not be worth it IMO).

woess:
> One compromise could be to generate a package-private class if the source class is package-private.

sounds reasonable to me

smarr:
@jtulach why do you generate JavaDoc for generated code? There aren't any comments anyway?

smarr:
I think this one was resolved.

PR #193, opened 2016-07-01
Fix checkstyle issues in generated code

smarr:
Changes the generated code to use “static final” as it is the recommended order by JLS.
Also fix two minor whitespace issues.


smarr:
Thanks

PR #196, opened 2016-08-03
Make ProfilerInstrument.ID public

smarr:
Since we need to use strings to get instruments from the polyglot engine, this ID should be public to be usable and avoid stuff breaking inadvertently on change.

/cc @chumer @jtulach @mlvdv

### DIFF
--- a/truffle/com.oracle.truffle.tools/src/com/oracle/truffle/tools/ProfilerInstrument.java
+++ b/truffle/com.oracle.truffle.tools/src/com/oracle/truffle/tools/ProfilerInstrument.java
@@ -30,7 +30,7 @@ import com.oracle.truffle.api.instrumentation.TruffleInstrument.Registration;
 
 @Registration(id = ProfilerInstrument.ID)
 public class ProfilerInstrument extends TruffleInstrument {
-    static final String ID = "profiler";
+    public static final String ID = "profiler";
 
     private Profiler profiler;
     private Instrumenter instrumenter;


chumer:
Where and how do you use it? I don't think it should be necessary to have this public.

smarr:
This is how I setup the polyglot engine:  https://github.com/smarr/SOMns/blob/master/src/som/VM.java#L322-L358 

So, I use the IDs.
But, the new Profiler.find() might actually be a replacement: https://github.com/graalvm/truffle/blob/master/truffle/com.oracle.truffle.tools/src/com/oracle/truffle/tools/Profiler.java#L86

smarr:
Actually, no, my last comment does not fully apply.
The problem is that I need to be able to lookup up the instrument via `engine.getInstruments().get(instrumentNameStr)` in order to be able to enable/disable it based on user settings.

So, if the ID is not public, and should not be public there, perhaps because it is undesirable to expose the class, then what about another solution?

Perhaps it would be useful to have a Tools class, which exposes the IDs for all tools know in the Truffle framework that are supposed to be useable by external users. It would also be a useful entry point to discover which tools are actually available.

chumer:
> Perhaps it would be useful to have a Tools class, which exposes the IDs for all tools know in the Truffle framework that are supposed to be useable by external users. It would also be a useful entry point to discover which tools are actually available.

An interesting idea. But I think it would counteract the idea of evolving tools independently. I still don't quite understand why you need to enumerate all the Truffle tools? Could you provide an example where Profiler#find is not flexible enough? I mean the Profiler class allows to enable disable profiling on its own. The new API will even support sessions like: Profiler.find(engine).startSampling() -> returns SamplingSession that can be closed.

smarr:
Perhaps this has changed, but quite a while ago, we discussed whether tools should be enabled or disabled per default. My memory tells me that we switched to a disabled-per-default model that requires strict opt-in per polyglot engine.

My main reason for that is that I don't want to know each of the tools existing, or on the class path for possibly unexpected reasons. I want to be sure that none of them interfere with my execution without me knowing about it explicitly.

Is this still true?

chumer:
Yes all instrumentations are disabled by default. If you use Debugger#find or Profiler#find you are going to enable its instrumentations implicitely.

chumer:
The preferred way to lookup an instrument via API is using its static find method. Then there is no need to get to its id.

PR #197, opened 2016-08-05
Add check to error when jvmci.Compiler is set but DefaultTruffleRuntime is used

smarr:
This is to avoid errors when using custom Graal/JVMCI builds and Graal is configured but not properly on the classpath.

### DIFF
--- a/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/Truffle.java
+++ b/truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/Truffle.java
@@ -126,6 +126,16 @@ public class Truffle {
                 if (access != null) {
                     return access.getRuntime();
                 }
+
+                // before falling back to the standard runtime, let's briefly
+                // check consistency
+                if (System.getProperties().containsKey("jvmci.Compiler")) {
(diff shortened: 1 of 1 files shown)


truffle/com.oracle.truffle.api/src/com/oracle/truffle/api/Truffle.java
```diff
@@ -126,6 +126,16 @@ public TruffleRuntime run() {
                 if (access != null) {
                     return access.getRuntime();
                 }
+
+                // before falling back to the standard runtime, let's briefly
+                // check consistency
+                if (System.getProperties().containsKey("jvmci.Compiler")) {
+                    // We assume that all JVMCI compiler are going to come with
+                    // a Truffle runtime.
```
dougxc: I think that's a dangerous assumption. If Graal is integrated to a future JDK, there's no guarantee the Truffle parts of it will be included. You have to simply trust that the VM was launched with the right options and paths for enabling JVMCI.

chumer: I agree.

PR #201, opened 2016-09-08
Added SourceSectionFilter support for source section based on column and length

smarr:
This complements the support for line breakpoints on URIs for which the code is not yet loaded.
Currently, only line breakpoints can be registered for URIs. With this addition, we can also request breakpoints for source sections based on line, column, and length.

/cc @mlvdv


chumer:
Wouldn't be a charIndex + length breakpoint more powerful? Do you really need to specify a column?

smarr:
This is an artifact of what I get from the editor, a pair of [line and column](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#position), and for section breakpoints, the token length in addition.

And, without loading the code, I can't calculate the index, because I don't know the newlines in the file. I suppose other editors will do it the other way around, and the index in range feature can be used.

chumer:
We have added column support in 0.33:
http://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/instrumentation/SourceSectionFilter.Builder.html#columnIn-int-int-

PR #202, opened 2016-09-08
Remove JRuby downstream test from Travis checks

smarr:
The downstream tests do not seem to be maintained, and are likely not very helpful.

To avoid extra maintenance overhead, we might want to remove it from the Travis tests.

/cc @eregon @chrisseaton


eregon:
Tests in allowed_failures are quite useless as they are not checked so I guess it's fine.

smarr:
ok. Could one of the two of you push it through then?

smarr:
Thanks!

PR #203, opened 2016-09-08
Document breaking change to `@Fallback`

smarr:
- also add some structure to changelog, because it is so long
- fix two minor typos


chumer:
Thanks. PR started.

chumer:
Got merged.

Issue #205, opened 2016-09-09
DSL creates incorrect createNode(...) if constructor has `Object` arguments


smarr:
The following example leads to a wrong `createNode(..)` method:

``` java
@GenerateNodeFactory
@NodeChildren({
  @NodeChild(value = "foo",  type = FooBarNode.class),
  @NodeChild(value = "bar",  type = FooBarNode.class)})
public abstract class FooBarNode extends Node {
  public FooBarNode(final SourceSection source, final Object a, final Object b) { }

  public abstract Object executeGeneric(final VirtualFrame frame);

  @Specialization
  public Object doSomething(final VirtualFrame frame, final Object foo, final Object bar) {
    return null;
  }
}
```

The result looks like this:

``` java
public FooBarNode createNode(Object... arguments) {
    if (arguments.length == 5 && (arguments[0] == null || arguments[0] instanceof SourceSection) && (arguments[1] == null || arguments[1] instanceof FooBarNode) && (arguments[2] == null || arguments[2] instanceof FooBarNode)) {
        return create((SourceSection) arguments[0], arguments[1], arguments[2], (FooBarNode) arguments[3], (FooBarNode) arguments[4]);
    } else {
        throw new IllegalArgumentException("Invalid create signature.");
    }
}
```

The important bit is that the type checks miss the `Object` arguments, and thus, the indexes used to check for the two `FooBarNode` are off by the number of `Object` arguments in the `FooBarNode` constructor.

smarr:
Fixed with https://github.com/graalvm/truffle/commit/8caa7bae2ba6eb518ee2b82efa858473fb091799

PR #206, opened 2016-09-09
Fixes incorrect argument index in createNode(..)

smarr:
This fixes issue #205.

The problem was that when there is an `Object` argument, the `index` would not be properly increased.

I didn't see any tests related to the annotation processor, so, not sure how that is supposed to be verified.


smarr:
@eregon I believe this one is still a real issue, could you perhaps pick up the patch?

smarr:
merged with https://github.com/graalvm/truffle/commit/8caa7bae2ba6eb518ee2b82efa858473fb091799
Thanks!

PR #211, opened 2016-11-30
Clarify StatementTag semantics

smarr:
Elaborate the intuition required for tagging nodes and give an example for the debugger.

### DIFF
--- a/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/StandardTags.java
+++ b/truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/StandardTags.java
@@ -43,7 +43,12 @@ public final class StandardTags {
      * <ul>
      * <li><b>Debugger:</b> Marks program locations where ordinary stepping should halt. The
      * debugger will halt just <em>before</em> a code location is executed that is marked with this
-     * tag.</li>
+     * tag.
+     * <p>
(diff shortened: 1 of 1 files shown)


truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/StandardTags.java
```diff
@@ -43,7 +43,12 @@ private StandardTags() {
      * <ul>
      * <li><b>Debugger:</b> Marks program locations where ordinary stepping should halt. The
      * debugger will halt just <em>before</em> a code location is executed that is marked with this
-     * tag.</li>
+     * tag.
+     * <p>
+     * In most languages, this means statements are distinct from expressions and only one node
```
jtulach: Using <p> inside of <li> is not valid (X)HTML as far as I know. But of course, thanks to high fault tolerance of browsers it will render something. Anyway, I'd restructure the text somehow.

jtulach:
Otherwise good - thanks for updating our documentation.

truffle/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/StandardTags.java
smarr: ok, are you doing that, or you want me to do something?
What change in structure are you thinking of?

PR #215, opened 2017-01-17
Remove incorrect cast on method calls (fixes #210)

### LINKED ISSUE #210: DSL flat layout issue with custom abstract execute method: Incorrect cast
The following class leads to incorrectly generated code:

```Java
@DSLOptions(defaultGenerator = DSLGenerator.FLAT)
public abstract class FooBar extends Node {
  @Children protected final FooBar[] args;

  protected FooBar() { args = new FooBar[0]; }

  public abstract Object execute(VirtualFrame frame, Object[] args);

  @Specialization
  public Object test(final Object[] args) {
    return null;
  }
}
```

It results in the following `execute(.)` method. The last line has the incorrect `(Object)` cast.

```Java
    @Override
    public Object execute(VirtualFrame frameValue, Object[] arg0Value) {
        int state = state_;
        if ((state & 0b10) != 0 /* is-active test(Object[]) */) {
            return test(arg0Value);
        }
        CompilerDirectives.transferToInterpreterAndInvalidate();
        return executeAndSpecialize((Object) arg0Value);
    }
```

smarr:
This adds a test and fix for a problem with having `Object[]` arguments to execute methods reported in #210.

The cast to `Object` that was inserted does not seem necessary anymore.

### DIFF
--- a/truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/ExecuteMethodTest.java
+++ b/truffle/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/ExecuteMethodTest.java
@@ -436,4 +436,24 @@ public class ExecuteMethodTest {
         }
     }
 
+    @TypeSystem
+    @DSLOptions(defaultGenerator = DSLOptions.DSLGenerator.FLAT)
+    static class ObjectArrayTypes {
+    }
+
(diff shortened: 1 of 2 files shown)


smarr:
was fixed in https://github.com/graalvm/truffle/commit/ff75d5b2e28de281a10c595960c788f84c4dd66c.
But unclear why the provided fix was ignored without discussion :(

PR #225, opened 2017-06-16
Aarch64 tbz patch

adinn:
This patch fixes a problem introduced into AArch64 by a previous truffle compiler fix.

Testing:
The patch fixes all the unit tests which were failing because of a broken tbz implementation.

n.b. 12 CountedLoop tests are still failing but that is due to an unrelated error to do with missing frame state for a SafeSignedDiv node

### DIFF
--- a/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
+++ b/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
@@ -95,6 +95,7 @@ import static org.graalvm.compiler.asm.aarch64.AArch64Assembler.Instruction.STXR
 import static org.graalvm.compiler.asm.aarch64.AArch64Assembler.Instruction.SUB;
 import static org.graalvm.compiler.asm.aarch64.AArch64Assembler.Instruction.SUBS;
 import static org.graalvm.compiler.asm.aarch64.AArch64Assembler.Instruction.TBZ;
+import static org.graalvm.compiler.asm.aarch64.AArch64Assembler.Instruction.TBNZ;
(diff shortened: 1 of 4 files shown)


compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
```diff
@@ -811,17 +813,74 @@ protected void cbz(int size, Register reg, int imm21, int pos) {
...
+
+    /**
+     * Test a single bit and branch if the bit is nonzero.
+     *
+     * @param reg general purpose register. May not be null, zero-register or stackpointer.
+     * @param uimm6 Unsigned 6-bit bit index.
+     * @param imm16 signed 16 bit offset
+     * @param pos Position at which instruction is inserted into buffer. -1 means insert at end.
+     */
+    protected void tbnz(Register reg, int uimm6, int imm16, int pos) {
```
axel22: It seems like you could optionally factor out `tbz` and `tbnz` into a common function called e.g. `testBitAndBranch`, and pass only the encoding to it.

I have no strong feelings, however - as you prefer.

axel22:
LGTM, with an optional comment.

compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
adinn: Yes, I suppose I could but I don't see it as anything to worry much about.

axel22: Sounds great, should land soon then!

PR #226, opened 2017-06-16
avoid asserts over missing before state when generating safe Div/Rem nodes on AArch64

adinn:
The Problem:
This patch fixes the 12 CountedLoop unit tests that currently fail with an assert during code generation.

Diagnosis:
On AArch64 DIv/Rem nodes are replaced by a snippet which includes AArch64-specific SafeDiv/Rem nodes. The latter are cloned from the original DIv/Rem nodes with a null beforeState. Unfortunately, they inherit their generate method from the generic Div/Rem nodes. The inherited methods call gen.state(this) to retrieve the beforeState and construct a FrameState passed in the call to the generators emitDiv/Rem method. This blows up because the beforeState is null.

Fix:
The AArch64 generator's emitDiv/Rem methods ignore the state argument because they will never encounter a div by zero. So, there is no need to compute a frame state. The fix works by overriding the inherited methods and passing null for the frame state.

Testing:
After this patch the 12 failing CountedLoop unit tests all succeed.

### DIFF
--- a/compiler/src/org.graalvm.compiler.replacements.aarch64/src/org/graalvm/compiler/replacements/aarch64/AArch64IntegerArithmeticSnippets.java
+++ b/compiler/src/org.graalvm.compiler.replacements.aarch64/src/org/graalvm/compiler/replacements/aarch64/AArch64IntegerArithmeticSnippets.java
@@ -206,6 +206,7 @@ public class AArch64IntegerArithmeticSnippets extends AbstractTemplates implemen
         protected SafeSignedDivNode(ValueNode x, ValueNode y) {
             super(TYPE, x, y);
         }
+
         @Override
         public void generate(NodeLIRBuilderTool gen) {
             // override to ensure we always pass a null frame state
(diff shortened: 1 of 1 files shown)


compiler/src/org.graalvm.compiler.replacements.aarch64/src/org/graalvm/compiler/replacements/aarch64/AArch64IntegerArithmeticSnippets.java
```diff
@@ -214,6 +221,12 @@ protected SafeSignedDivNode(ValueNode x, ValueNode y) {
         protected SafeSignedRemNode(ValueNode x, ValueNode y) {
             super(TYPE, x, y);
         }
+        @Override
```
axel22: Eclipseformat is complaining about a missing empty line between the methods.

adinn:
Restyled and pushed :-) Good to go?

axel22:
@adinn Should be great - the merge should land soon! Thanks a lot!

PR #237, opened 2017-07-11
fix AArch64 stack saves and copies

adinn:
This patch fixes the problem reported here: https://github.com/graalvm/graal/issues/236

The patch includes a unit test which manifest the problem and confirms the fix as described in the issyue mentioned above.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
   global:
     - JVMCI_VERSION="jvmci-0.30"
     - JVMCI_BASE_JDK8="jdk1.8.0_121"
-    - JDK9_EA_BUILD="176"
+    - JDK9_EA_BUILD="177"
   matrix:
 # Cannot build OpenJDK8 based JVMCI JDK until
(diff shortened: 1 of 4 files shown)


dougxc:
lgtm

dougxc:
I've started the process for integrating this. I reverted the change in `JVMCIVersionCheck` as it forces us to prepare a 177 binary internally. The only change really needed is in `.travis.yml`.

adinn:
Ok, great. Thanks very much.

adinn:
I have deleted the unused var stackCopy. I did not detect this error when I ran the eclipseformat command. Is this the right command? Do I need to provide some other arguments to mx?

dougxc:
This should do it: `mx clean; mx build --jdt /path/to/ecj-4.5.1.jar --warning-as-error`
You can get ecj-4.5.1.jar [here](http://archive.eclipse.org/eclipse/downloads/drops4/R-4.5.1-201509040015/).

PR #251, opened 2017-08-23
[AMD64] implement String.compareTo substitutions

christhalinger:
This change ports the intrinsics from C2.  The substitutions are only enabled on JDK 9.  It might be possible to use it for 8 too but I didn't bother.

`compiler/intrinsics/string` all pass with:

- UseSSE=0 UseAVX=0
- UseSSE=4 UseAVX=0
- UseSSE=4 UseAVX=2

which enables the different code paths.  I am not able to test the `AVX512VLBW` code path as I don't have access to a machine that supports these instructions.

Here are the benchmark numbers:

http://cr.openjdk.java.net/~shade/density/string-density-bench.jar

Graal:

```
Benchmark                 (size)  Mode  Cnt     Score    Error  Units
CompareToBench.cmp1_cmp1       1  avgt   25     5.480 ±  0.147  ns/op
CompareToBench.cmp1_cmp1      64  avgt   25    61.178 ±  1.768  ns/op
CompareToBench.cmp1_cmp1    4096  avgt   25  2621.249 ± 63.138  ns/op
CompareToBench.cmp1_cmp2       1  avgt   25     5.914 ±  0.186  ns/op
CompareToBench.cmp1_cmp2      64  avgt   25    93.736 ±  1.029  ns/op
CompareToBench.cmp1_cmp2    4096  avgt   25  4975.549 ± 45.001  ns/op
CompareToBench.cmp2_cmp1       1  avgt   25     5.841 ±  0.049  ns/op
CompareToBench.cmp2_cmp1      64  avgt   25    94.098 ±  0.921  ns/op
CompareToBench.cmp2_cmp1    4096  avgt   25  5052.871 ± 80.579  ns/op
CompareToBench.cmp2_cmp2       1  avgt   25     5.924 ±  0.062  ns/op
CompareToBench.cmp2_cmp2      64  avgt   25   119.063 ±  1.316  ns/op
CompareToBench.cmp2_cmp2    4096  avgt   25  6674.597 ± 63.872  ns/op
```

Graal with intrinsic:

```
Benchmark                 (size)  Mode  Cnt    Score   Error  Units
CompareToBench.cmp1_cmp1       1  avgt   25    5.898 ± 0.059  ns/op
CompareToBench.cmp1_cmp1      64  avgt   25   14.540 ± 0.160  ns/op
CompareToBench.cmp1_cmp1    4096  avgt   25  126.307 ± 6.868  ns/op
CompareToBench.cmp1_cmp2       1  avgt   25    5.736 ± 0.082  ns/op
CompareToBench.cmp1_cmp2      64  avgt   25   15.717 ± 0.219  ns/op
CompareToBench.cmp1_cmp2    4096  avgt   25  210.045 ± 5.888  ns/op
CompareToBench.cmp2_cmp1       1  avgt   25    6.591 ± 0.863  ns/op
CompareToBench.cmp2_cmp1      64  avgt   25   16.949 ± 1.856  ns/op
CompareToBench.cmp2_cmp1    4096  avgt   25  211.254 ± 8.404  ns/op
CompareToBench.cmp2_cmp2       1  avgt   25    5.928 ± 0.091  ns/op
CompareToBench.cmp2_cmp2      64  avgt   25   15.976 ± 0.188  ns/op
CompareToBench.cmp2_cmp2    4096  avgt   25  206.201 ± 7.247  ns/op
```

And for comparison C2:

```
Benchmark                 (size)  Mode  Cnt    Score   Error  Units
CompareToBench.cmp1_cmp1       1  avgt   25    5.759 ± 0.160  ns/op
CompareToBench.cmp1_cmp1      64  avgt   25   14.152 ± 0.302  ns/op
CompareToBench.cmp1_cmp1    4096  avgt   25  115.376 ± 5.426  ns/op
CompareToBench.cmp1_cmp2       1  avgt   25    5.768 ± 0.095  ns/op
CompareToBench.cmp1_cmp2      64  avgt   25   15.646 ± 0.218  ns/op
CompareToBench.cmp1_cmp2    4096  avgt   25  211.439 ± 2.914  ns/op
CompareToBench.cmp2_cmp1       1  avgt   25    5.884 ± 0.072  ns/op
CompareToBench.cmp2_cmp1      64  avgt   25   15.139 ± 0.221  ns/op
CompareToBench.cmp2_cmp1    4096  avgt   25  214.635 ± 5.918  ns/op
CompareToBench.cmp2_cmp2       1  avgt   25    5.573 ± 0.076  ns/op
CompareToBench.cmp2_cmp2      64  avgt   25   15.721 ± 0.239  ns/op
CompareToBench.cmp2_cmp2    4096  avgt   25  203.789 ± 7.579  ns/op
```


compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -1825,6 +1860,26 @@ public final void divsd(Register dst, Register src) {
         emitByte(0xC0 | encode);
     }
 
+    public final void evmovdquq(Register dst, AMD64Address src, int vectorLen) {
+        assert supports(CPUFeature.AVX512F);
+        AMD64InstructionAttr attributes = new AMD64InstructionAttr(vectorLen, /* vex_w */ true, /* legacy_mode */ false, /* no_mask_reg */ false, /* uses_vl */ true, target);
+        attributes.setAddressAttributes(/* tuple_type */ EvexTupleType.EVEX_FVM, /* input_size_in_bits */ EvexInputSizeInBits.EVEX_NObit);
```
christhalinger: Maybe we should remove the comments since the class name already is enough explanation.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/nodes/ArrayCompareToNode.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Override
+    public Node canonical(CanonicalizerTool tool) {
+        if (tool.allUsagesAvailable() && hasNoUsages()) {
+            return null;
+        }
+        ValueNode a1 = GraphUtil.unproxify(array1);
+        ValueNode a2 = GraphUtil.unproxify(array2);
+        if (a1 == a2) {
+            return ConstantNode.forInt(0);
```
christhalinger: There might be more opportunities for optimization (like `ArrayEqualsNode`) but I only did identity.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/nodes/ArrayCompareToNode.java
```diff
@@ -0,0 +1,139 @@
...
+        return this;
+    }
+
+    @Override
+    public void virtualize(VirtualizerTool tool) {
+        ValueNode alias1 = tool.getAlias(array1);
+        ValueNode alias2 = tool.getAlias(array2);
+        if (alias1 == alias2) {
+            // the same virtual objects will always have the same contents
+            tool.replaceWithValue(ConstantNode.forInt(0, graph()));
```
christhalinger: Same.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -1950,6 +2005,32 @@ public final void jmpb(Label l) {
...
+        assert supports(CPUFeature.AVX512BW);
+        AMD64InstructionAttr attributes = new AMD64InstructionAttr(AVX_128bit, /* rex_w */ true, /* legacy_mode */ true, /* no_mask_reg */ true, /* uses_vl */ false, target);
+        int encode = vexPrefixAndEncode(src1, Register.None, src2, VexSimdPrefix.VEX_SIMD_NONE, VexOpcode.VEX_OPCODE_0F, attributes);
+        emitByte(0x98);
+        emitByte(0xC0 | encode);
+    }
+
+    public final void kmovql(Register dst, Register src) {
+        assert supports(CPUFeature.AVX512BW);
+        if (src.getRegisterCategory().equals(AMD64.MASK)) {
```
christhalinger: We have to do this because Graal doesn't distinguish between `Register` and `KRegister` like C2 does.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -2512,6 +2597,26 @@ void pcmpestri(Register dst, Register src, int imm8) {
         emitByte(imm8);
     }
 
+    public final void pmovzxbw(Register dst, AMD64Address src) {
+        assert supports(CPUFeature.SSE4_2);
+        // XXX legacy_mode should be: _legacy_mode_bw
```
christhalinger: This is the only thing I'm worried about.  Not sure what `_legacy_mode_bw` really is.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -2512,6 +2597,26 @@ void pcmpestri(Register dst, Register src, int imm8) {
...
+        AMD64InstructionAttr attributes = new AMD64InstructionAttr(AvxVectorLen.AVX_128bit, /* rex_w */ false, /* legacy_mode */ false, /* no_mask_reg */ true, /* uses_vl */ false, target);
+        attributes.setAddressAttributes(/* tuple_type */ EvexTupleType.EVEX_HVM, /* input_size_in_bits */ EvexInputSizeInBits.EVEX_NObit);
+        simdPrefix(dst, Register.None, src, VexSimdPrefix.VEX_SIMD_66, VexOpcode.VEX_OPCODE_0F_38, attributes);
+        emitByte(0x30);
+        emitOperandHelper(dst, src, 0);
+    }
+
+    public final void vpmovzxbw(Register dst, AMD64Address src, int vectorLen) {
+        assert supports(CPUFeature.AVX);
+        // XXX legacy_mode should be: _legacy_mode_bw
```
christhalinger: Same.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
dougxc: Yes, the redundant comments should be removed.

dougxc:
This needs some tests added to `org.graalvm.compiler.replacements.test.ArraysSubstitutionsTest` for the new intrinsic.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
tkrodriguez: It looks like part of the magic autoencoding for supporting AVX512.  We don't really support AVX512 right now I think.  Maybe @rschatz has some thoughts on how we should deal with?  I suspect we leave it as false but maybe we also need to explicitly disable the AVX512 feature tests?

tkrodriguez:
Where is the CompareToBench benchmark?

christhalinger:
https://bugs.openjdk.java.net/browse/JDK-8136757?focusedCommentId=14110566&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14110566

compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64LIRGenerator.java
```diff
@@ -441,6 +442,20 @@ public Variable emitByteSwap(Value input) {
...
 
     @Override
+    public Variable emitArrayCompareTo(JavaKind kind1, JavaKind kind2, Value array1, Value array2, Value length1, Value length2) {
+        LIRKind resultKind = LIRKind.value(AMD64Kind.DWORD);
+        RegisterValue raxRes = AMD64.rax.asValue(resultKind);
+        RegisterValue cnt1 = AMD64.rcx.asValue(length1.getValueKind());
+        RegisterValue cnt2 = AMD64.rdx.asValue(length2.getValueKind());
+        emitMove(cnt1, length1);
+        emitMove(cnt2, length2);
+        append(new AMD64ArrayCompareToOp(this, kind1, kind2, raxRes, array1, array2, cnt1, cnt2));
```
tkrodriguez: I think the register selection logic here is broken and also in AMD64ArrayCompareToOp .  C2 uses bound registers and USE_KILL for all the inputs so I think this needs to work differently.  The assembly itself doesn't seem to care about register selection unless there are hidden assumptions in it so I think you could drop the fixed register usage.  If it requires fixed register usage then the assembly should be asserting them.

christhalinger: No, it cares; `pcmpestri` does.  And there are asserts.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64ArrayCompareToOp.java
```diff
@@ -0,0 +1,595 @@
...
+public final class AMD64ArrayCompareToOp extends AMD64LIRInstruction {
+    public static final LIRInstructionClass<AMD64ArrayCompareToOp> TYPE = LIRInstructionClass.create(AMD64ArrayCompareToOp.class);
+
+    private final JavaKind kind1;
+    private final JavaKind kind2;
+    private final int array1BaseOffset;
+    private final int array2BaseOffset;
+
+    @Def({REG}) protected Value resultValue;
+    @Alive({REG}) protected Value array1Value;
```
tkrodriguez: To get something similar to C2s USE_KILL you need pairs of Use and Temp which both point to same physical.  Look at AMD64StringIndexOfOp for instance.  I think the current definition will allow reuse of the values after it's been destroyed by the assembly.

tkrodriguez:
I think we'd like to use this in 8 if possible since it will get more testing coverage if we're really going to use it.  The only difference should be that since we're using a real char[] the length will be scaled differently, right?  Maybe we should be using an enumeration like C2 does but with an extra CC case for a true char[] compare.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64ArrayCompareToOp.java
christhalinger: I can certainly change it to work like `AMD64StringIndexOfOp` but I don't see why that's required.  The `LIRGenerator` grabs the registers and they are alive throughout the op.

compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64LIRGenerator.java
tkrodriguez: Oh I see them now.  They are about 150 lines into emitCode which seems a bit late.  :)  I would expect them at the beginning and/or in the constructor.  Can you add them in the constructor?

Sorry I edited my comment in between you reply.  We have to use fixed registers like C2 does to get the kill semantics we need anyway.

christhalinger:
Using:

```
ArrayCompareToNode.compareTo(value, other, value.length, other.length, JavaKind.Char, JavaKind.Char);
```

should just work because the assembly already handles that case.

tkrodriguez:
We should at least have a basic unit test for this as well.

tkrodriguez:
I thought JavaKind.Char doesn't mean char[] since String.value is always a byte[].  It means a byte[] encoded in UTF16 so the length value is different.  I guess you could do value.length * 2.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64ArrayCompareToOp.java
```diff
@@ -0,0 +1,595 @@
...
+
+    @Temp({REG, ILLEGAL}) protected Value vectorTemp1;
+
+    public AMD64ArrayCompareToOp(LIRGeneratorTool tool, JavaKind kind1, JavaKind kind2, Value result, Value array1, Value array2, Value length1, Value length2) {
+        super(TYPE);
+        this.kind1 = kind1;
+        this.kind2 = kind2;
+
+        // Both offsets should be the same but better be safe than sorry.
+        Class<?> array1Class = Array.newInstance(kind1.toJavaClass(), 0).getClass();
```
tkrodriguez: This isn't really right is it?  In 9 String.value is always a byte[] and in 8 it's always a char[].

christhalinger: Absolutely correct. Let me fix that.

christhalinger:
Yes, this is the issue:
```
        if (kind1 == JavaKind.Char && kind2 == JavaKind.Char) {
            // Divide length by 2 to get number of chars
            masm.shrl(cnt2, 1);
        }
```
but as you say, value.length * 2 should work.

compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64LIRGenerator.java
christhalinger: I was thinking about putting them into the constructor but then didn't do because it would be redundant.  And I wanted to keep the code as close to the native one as possible.  If you want I can add an additional assert.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
rschatz: I can only speculate, I don't really know what that means either.

It seems the `legacy_mode` attribute is never actually used in our code.
From the name, I suspect it has something to do with selecting EVEX vs VEX prefix. Inside the constructor of `AMD64InstructionAttr` the `legacy_mode` flag is filtered by the `AVX512F` feature flag, which tells us whether `EVEX` is available. And `_legacy_mode_bw` sounds like it has something to do with the `AVX512BW` feature flag. The `VPMOVZXBW` needs the `AVX512BW` feature flag to be usable with `EVEX`.

dougxc:
Still needs some tests added to `ArraysSubstitutionsTest`.

christhalinger:
Had to do something else.  I will get back to this soon.

dougxc:
@christhalinger any chance you'll get back to this soon? Or should we open an OpenJDK to find other resources to take it over?

christhalinger:
Right.  Oleg Chirukhin offered to help out with the tests but I guess he got sidetracked too.  Maybe we should open a JBS ticket.  The code is done; we just need some tests.

christhalinger:
Talked to @dsamersoff today and he said he could pick this one up.

dougxc:
Excellent - thanks @dsamersoff !

dougxc:
Resolved by https://github.com/oracle/graal/pull/299.

PR #260, opened 2017-11-04
Fixing bugs found by static analysis

amaembo:
Hello!

There are fixes to some problems found by static analysis which looks as clear bugs/misprints.

OCA signed ( http://www.oracle.com/technetwork/community/oca-486395.html#v -- Tagir Valeev )

### DIFF
--- a/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/LIRIntrospection.java
+++ b/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/LIRIntrospection.java
@@ -340,7 +340,7 @@ abstract class LIRIntrospection<T> extends FieldIntrospection<T> {
     private static boolean isPrintableAsciiString(byte[] array) {
         for (byte b : array) {
             char c = (char) b;
-            if (c != 0 && c < 0x20 && c > 0x7F) {
+            if (c != 0 && (c < 0x20 || c > 0x7F)) {
                 return false;
             }
         }
--- a/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/constopt/ConstantTree.java
+++ b/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/constopt/ConstantTree.java
@@ -76,7 +76,7 @@ public class ConstantTree extends PrintableDominatorOptimizationProblem<Constant
 
         public List<UseEntry> getUsages() {
             if (usages == null) {
-                Collections.emptyList();
+                return Collections.emptyList();
             }
             return usages;
         }
--- a/compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ProfileCompiledMethodsPhase.java
+++ b/compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ProfileCompiledMethodsPhase.java
@@ -160,7 +160,7 @@ public class ProfileCompiledMethodsPhase extends Phase {
             return 10;
         } else if (node instanceof Access) {
             return 2;
-        } else if (node instanceof LogicNode || node instanceof ConvertNode || node instanceof BinaryNode || node instanceof NotNode) {
+        } else if (node instanceof LogicNode || node instanceof ConvertNode || node instanceof NotNode) {
             return 1;
         } else if (node instanceof IntegerDivRemNode || node instanceof FloatDivNode || node instanceof RemNode) {
             return 10;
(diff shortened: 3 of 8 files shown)


chrisseaton:
Tagir has signed the OCA for the OpenJDK project.

thomaswue:
Thanks for the contribution @amaembo! We will add you to the whitelist. /CC @dougxc

dougxc:
Hi @amaembo , what static analysis tool (if any) did you use to find these bugs?

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/LIRIntrospection.java
```diff
@@ -340,7 +340,7 @@ protected static String getFieldString(Object obj, int index, Fields fields) {
     private static boolean isPrintableAsciiString(byte[] array) {
         for (byte b : array) {
             char c = (char) b;
-            if (c != 0 && c < 0x20 && c > 0x7F) {
+            if (c != 0 && c >= 0x20 && c <= 0x7F) {
```
dougxc: Shouldn't this be:
```
if (c != 0 && (c < 0x20 || c > 0x7F)) {
```
?

amaembo:
@dougxc all bugs were found by IntelliJ IDEA static analysis, namely "Constant conditions & exceptions" and "Result of method call ignored" inspections. There are more suspicious places, I fixed only those which seem clear bugs/typos to me and the fix looks obvious.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/LIRIntrospection.java
chrisseaton: The first `c != 0` is now redundant, isn't it?

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/constopt/ConstantTree.java
```diff
@@ -76,7 +76,7 @@ public void addUsage(UseEntry usage) {
 
         public List<UseEntry> getUsages() {
             if (usages == null) {
-                Collections.emptyList();
+                return Collections.emptyList();
```
dougxc: @davleopo , looking at the callers of `getUsages()`, I'm very surprised this hasn't caused NPEs in the past - any idea why?

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ProfileCompiledMethodsPhase.java
```diff
@@ -160,15 +160,15 @@ private static double getNodeWeight(Node node) {
...
+        } else if (node instanceof LogicNode || node instanceof ConvertNode || node instanceof NotNode) {
             return 1;
         } else if (node instanceof IntegerDivRemNode || node instanceof FloatDivNode || node instanceof RemNode) {
             return 10;
         } else if (node instanceof MulNode) {
             return 3;
         } else if (node instanceof Invoke) {
             return 5;
-        } else if (node instanceof IfNode || node instanceof SafepointNode) {
+        } else if (node instanceof IfNode || node instanceof SafepointNode || node instanceof BinaryNode) {
```
dougxc: This is where seeing the tool output would be helpful - it took me a while to figure out why this change makes sense (i.e., `MulNode` should be tested before `BinaryNode`).

amaembo:
@chrisseaton oops, shame on me, that particular line fix was incorrect. Updated, thanks.

chrisseaton:
Yeah if you'd run the static analysis again on your fix for the static analysis you probably would have found it as well!

truffle/src/com.oracle.truffle.api.dsl/src/com/oracle/truffle/api/dsl/Introspection.java
```diff
@@ -241,8 +241,8 @@ SpecializationInfo getSpecialization(String methodName) {
 
     private void checkVersion() {
         int version = -1;
-        Object objectVersion = data[0];
-        if (data.length > 0 && objectVersion instanceof Integer) {
+        if (data.length > 0 && data[0] instanceof Integer) {
```
dougxc: @chumer seems like `data.length` is never 0 in practice otherwise we would have seen `ArrayIndexOutOfBoundsException`s here.

chumer: No it should not be. But the change makes it more robust. Thanks!

dougxc:
Thanks a lot for the changes! We need to see if/how we can run this analysis in our gate.

amaembo:
> This is where seeing the tool output would be helpful

The tool output here is not very verbose. It just says: `node instanceof MulNode` is always false, `node instanceof FloatDivNode` is always false and `node instanceof RemNode` is always false. After that I had to check the type hierarchy and inspect the surrounding code to understand why it's so.

PR #265, opened 2017-11-29
AArch64 support for HotSpot thread-local poll

dean-long:


dougxc:
Looks good to me but it would still be good if someone with access to an AArch64 system could test the change.

dean-long:
Andrew Haley has tested this and provided a fix.  It should be good to go now.

dougxc:
Resolved by https://github.com/graalvm/graal/commit/726364735fe24de21f49f2043fed7471291fda06

PR #266, opened 2017-12-01
Divide can deopt if y is -1

jcdavis:
(This is broken as-is due to breaking CountedLoopTest)


rschatz:
The broken `CountedLoopTest` is a testbug. The test tries to inspect the result of the counted loop detection, but it's doing it too late in the compilation. I'll fix that.

Looks good otherwise. Good catch ;)

PR #268, opened 2017-12-07
Implement CRC32C substitutions

dchuyko:
CRC32C is available since Java 9 and there are StubRoutines::updateBytesCRC32C() stubs for x86_64, aarch64 and sparcv9.

GraalHotSpotVMConfig got useCRC32CIntrinsics, it reads the flag for 9 and it is always false for 8.

New CRC32CSubstitutions class provides substitutions to update() methdods using the stub. 3 foreign calls providers were updated to use it and the substitutions are registered in HotSpotGraphBuilderPlugins.

New CRC32CSubstitutionsTest compares checksums for arrays with zero and non-zero offset and for direct and wrapped ByteBuffer. It passes in my environment.

On the whole CRC32C support is similar to CRC32 taking into account that the class is not present in Java 8.

Some x86 performance results on JDK9.
benchmark: http://cr.openjdk.java.net/~dchuyko/8189177/crc32c/CRC32CBench.java
size: 512 bytes
before:     351 ± 16  ns/op
after:     121 ± 2  ns/op 3x


dchuyko:
I act as an employee of BellSoft, the company has signed the OCA.

dougxc:
Looks good to me - thanks!

PR #274, opened 2018-01-05
Push down constants in AddNode

jcdavis:
The current `reassociate` call only handles cases such as `(A+C1) + C2`. This adds cases for `(X+C1)+Y` and `X+(Y+C2)`, which allows simplifying
```
      (A+1)+(B+2)
    =>(A+(B+2))+1
    =>((A+B)+2)+1
```` 
from which the existing `reassociate` will pull out the 2 and we end up with `(A+B)+3`


tkrodriguez:
As @gilles-duboscq suggested this could interact with loop invariant reassociation but that I think that's true of any reassociation so I don't think it matters for this change.  I"ll push once this passes the gate.

jcdavis:
What's the correct way to avoid the eclipse comment formatter rewriting here?
`@formatter:on` and `@formatter:off` ?

dougxc:
Yes, for example https://github.com/oracle/graal/blob/73196584fec60b58ff8d70fd1c3d29cc2634d557/truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/vm/ImplicitExplicitExportTest.java#L103-L111

PR #281, opened 2018-01-13
Skip jdk11 CONSTANT_Dynamic tag

dsamersoff:
Small addition to the patch that bumps supported classfile version to JDK11(55). Skip new CONSTANT_Dynamic tag.


dsamersoff:
I signed OCA as:
Dmitry Samersov - OpenJDK - dsamersoff

PR #282, opened 2018-01-13
Skip jdk11 CONSTANT_Dynamic tag

dsamersoff:
Small addition to the patch that bumps supported classfile version to JDK11(55). Skip new CONSTANT_Dynamic tag.

I signed OCA as:
Dmitry Samersov - OpenJDK - dsamersoff


dougxc:
Looks good. I'll integrate it.

PR #283, opened 2018-01-14
Replace javah with javac

dsamersoff:
Warning! These changes has an impact to the process.

1. It depends to mx changes (https://github.com/graalvm/mx/pull/142).
2. It's not clear to me how we should handle jdk8 that stay with javah

### DIFF
--- a/truffle/src/com.oracle.truffle.nfi.native/Makefile
+++ b/truffle/src/com.oracle.truffle.nfi.native/Makefile
@@ -29,31 +29,22 @@ QUIETLY$(MX_VERBOSE)  = @
 
 .PHONY: default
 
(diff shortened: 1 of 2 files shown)


dougxc:
@rschatz , does this look like the right way to accomodate the removal of `javah` from recent JDKs?

rschatz:
> It's not clear to me how we should handle jdk8 that stay with javah

That's not really a problem, `javac` on jdk8 does support the `-h` option. The question is how to support building with ecj/eclipse/netbeans/... I'll look into that.

rschatz:
I made a new PR for mx (graalvm/mx#143) that works around the ecj build issue by still using javah there. With that, this PR should work. We have to wait until the mx change is merged, then I can merge this PR.

Thanks for the contribution.

PR #292, opened 2018-01-30
AArch64: fix branch target overflows and return addresses.

theRealAph:
This fixes a few overflows discovered when testing with a large (i.e. more than 128Mb) code buffer.


tkrodriguez:
This looks fine apart from the formatting issue found by the gate.

sanzinger:
I am working to integrate this PR.

tkrodriguez:
Thanks!

adinn:
stefan,

you really need to just pull this into the main tree. Andrew Haley's fix doesn't only correct the problem with offsets > 26 bits. It also fixes another serious issue in method deopt which causes several of the unit tests to fail or crash the JVM (I don't know how many will fail because of the crash).

sanzinger:
I just struggle with the gate. But instead of bugging around I will now just merge it, despite the gate has errors. We're going to fix them in the long term.

sanzinger:
Change is scheduled for merge and will be available soon.

adinn:
BTW, there are still problems running on jdk10 even with this patch because the Graal fixes also require fixes to the jvmci code in hotspot. In order to get this to work it has to be built using the latest jdk11 tree.

sanzinger:
Trampolines come with https://bugs.openjdk.java.net/browse/JDK-8193260 on JDK11.

Issue #298, opened 2018-02-16
revive support for HotSpot Truffle trampoline in AArch64 under CompressedOops


dougxc:
In https://github.com/oracle/graal/commit/6afc8f0ba83c7c73fa5b455d2bfa9bf86efe7d11, the instrumentation added to methods annotated by `@TruffleCallBoundary` had to be updated to account for the [extra load](https://github.com/oracle/graal/commit/6afc8f0ba83c7c73fa5b455d2bfa9bf86efe7d11#diff-c75a35e6f1e812d8592951c060d634c8L38) required after making `OptimizedCallTarget` contain an `InstalledCode` instead of directly subclassing it.

This has already been for [AMD64](https://github.com/oracle/graal/blob/9eac6585eb4aa211c030362eceda5edcbee7b4a6/compiler/src/org.graalvm.compiler.truffle.compiler.hotspot.amd64/src/org/graalvm/compiler/truffle/compiler/hotspot/amd64/AMD64TruffleCallBoundaryInstrumentationFactory.java#L68-L73) and [SPARC](https://github.com/oracle/graal/blob/9eac6585eb4aa211c030362eceda5edcbee7b4a6/compiler/src/org.graalvm.compiler.truffle.compiler.hotspot.sparc/src/org/graalvm/compiler/truffle/compiler/hotspot/sparc/SPARCTruffleCallBoundaryInstumentationFactory.java#L69-L72). However, I wasn't sure of the necessary changes for [AArch64](https://github.com/oracle/graal/blob/9eac6585eb4aa211c030362eceda5edcbee7b4a6/compiler/src/org.graalvm.compiler.truffle.compiler.hotspot.aarch64/src/org/graalvm/compiler/truffle/compiler/hotspot/aarch64/AArch64TruffleCallBoundaryInstumentationFactory.java#L64).

dougxc:
On a side node, the spelling of `AArch64TruffleCallBoundaryInstumentationFactory` should be fixed to `AArch64TruffleCallBoundaryInstrumentationFactory` (add the missing "r").

dougxc:
@adinn , would you be able to look into this issue?

adinn:
sure I will take a look and see what I can do

n.b. expect possible slight delay and maybe some noob questions as this will be a cold start -- I have not built or debugged Graal for a few months :-)

dougxc:
No problem. Once you have something that compiles, this is a simple way to test it:
```
[...]

PR #302, opened 2018-02-19
Support compressed HotSpot Truffle trampoline on AArch64

sanzinger:
This PR adds support for compressed HotSpot Truffle trampoline on AArch64. Solves #298


sanzinger:
@adinn please review this change.

PR #312, opened 2018-03-09
JDK-8193241: Part 1. Added intrinsic support for (Math) dec/incrementExact

phedlin:
Minimal intrinsic support for Math./dec/inc/rementExact.
(Primarily testing account setup and work-flow.)

### DIFF
--- a/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
+++ b/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
@@ -385,6 +385,23 @@ public class StandardGraphBuilderPlugins {
         if (allowDeoptimization) {
             for (JavaKind kind : new JavaKind[]{JavaKind.Int, JavaKind.Long}) {
                 Class<?> type = kind.toJavaClass();
+
+                r.register1("decrementExact", type, new InvocationPlugin() {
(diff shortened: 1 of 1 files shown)


eregon:
Those intrinsics in the pull request might actually decrease performance unless IntegerAddExactNode is smart enough. The pure-Java implementation seems better because it generates a normal floating AddNode and a check against `Integer.MAX_VALUE`.
IntegerAddExactNode on the other hand behaves like a fixed node IIRC, which restricts optimizations.
cc @aardvark179

thomaswue:
@eregon IntegerAddExactNode is a floating node that is only at later stages during compilation lowered into a fixed IntegerAddExactSplitNode.

aardvark179:
I'm not sure if this behaviour has changed since July, but we originally added special cases in TruffleRuby for incrementing fixnums because in tight loops it ultimately generated code like this
```
cmp    r10d,0x7fffffff
je     handler
cmp    r10d,0x7ffffffe
je     handler
...
add    r10d, 0x13
```
rather than
```
add    r13d, 0x1
jo     handler
add    r13d, 0x1
jo     handler
add    r13d, 0x1
jo     handler
...
```
Which can have a large (upto 20x on my machine) performance difference, though ideally we'd see this series of incrementExacts compiled down to something like
```
add    r13d, 0x13
jo     handler
```

dougxc:
@aardvark179 , once this PR is merged, please double check the impact on the code pattern you show above.

PR #315, opened 2018-03-16
[GR-8748] Use @Use/@Temp instead of @Alive for modifiable variables.

mur47x111:


mur47x111:
Splitting @Alive into @Use/@Temp indicates that the contents could be modified, in this specific intrinsic, `cnt1` and `cnt2`. The test case might or might not be effective depending on the machine (mine without AVX2). The goal is to make register allocator store `length1 << 1` in `ECX` before the intrinsic, and the intrinsic has to be wrapped in a loop, see `TreeMap.getEntry`.

PR #316, opened 2018-03-16
Update reference to Graal JVMCI 8 repo

smarr:
This change simply updates the URL and repo type for Graal JVMCI 8, [since it recently moved](https://markmail.org/message/kdnj2kmhodmx47rs).


dougxc:
To mitigate integration overhead, I'll include this change as part of a PR I'm current pushing through.

dougxc:
Integrated in 9bc81e97b2c.

PR #319, opened 2018-03-18
String.compareTo for AArch64

dsamersoff:
String compareTo intrinsic for  AArch64

### DIFF
--- a/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayCompareToOp.java
+++ b/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayCompareToOp.java
@@ -1,0 +1,286 @@
+/*
+ * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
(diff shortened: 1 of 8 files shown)


dsamersoff:
Style check failed with strange warning from findbugs that I can't reproduce locally. So leaving it as is and address it later.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayCompareToOp.java
```diff
@@ -0,0 +1,277 @@
...
+
+    private final int array1BaseOffset;
+    private final int array2BaseOffset;
+
+    @Def({REG}) protected Value resultValue;
+
+    @Alive({REG}) protected Value array1Value;
+    @Alive({REG}) protected Value array2Value;
+    @Alive({REG}) protected Value length1Value;
+    @Alive({REG}) protected Value length2Value;
```
dougxc: Please check with @mur47x111 whether the [fix](https://github.com/oracle/graal/commit/c5e87525231707e6c0804b97a72cf90836910c4f#diff-28a415c4cd88eb4a9c38354ecb216c17) he made for GR-8748 is also needed here.

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/StringCompareToTest.java
```diff
@@ -100,15 +102,29 @@ public void testEqualString() {
     @Test
     @Ignore("GR-8748")
```
dougxc: I don't think we need this `@Ignore` any more right @mur47x111 ?

mur47x111: My commit has removed this annotation and reenable the test.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayCompareToOp.java
mur47x111: Please apply the same pattern as [here](https://github.com/oracle/graal/commit/a188baff1435dd34a96f76d21275e269cfa546b3#diff-28a415c4cd88eb4a9c38354ecb216c17) to ensure modifiable variables are killed after the intrinsic.

dsamersoff:
Adopted all comments. The fix is ready to push.

PR #320, opened 2018-03-19
Add AArch64 support() function for determining CPUFeature support.

daniel-stewart-qdt:
Adding a function for determining the support for AArch64 CPUFeatures. This is a small function to prepare the way for implementing certain intrinsics which require atomic instructions only available on AArch64 architectures that have implemented certain 8.X extensions.

This function is modeled after the same function in AMD64.


sanzinger:
Thank you for your contribution. I am going to integrate this change.

daniel-stewart-qdt:
Thanks!

PR #324, opened 2018-03-20
Framework for Benchmarking Partial Evaluation

flortsch:
This PR is part of work which adds a framework for benchmarking partial evaluation in the Truffle compiler. The framework provides a parameterizable base class which benchmarks methods of the class PartialEvaluator. Sub classes implement a method to define call targets which are used during partial evaluation. It is intended that future benchmarks will use a specific builder class to create various kinds of call targets in a specific Truffle language implementation in order to benchmark corner cases of partial evaluation. The language implementation is also provided in this PR and has the following simple semantics:
- one single kind of arithmetic (addition)
- long and string literals
- if and while constructs
- read and write of local variables


flortsch:
PELang, the language built in this PR for benchmarking partial evaluation, was extended to also support basic blocks and block dispatches.

chumer:
Did you deliberately decide to not include support calls for these benchmarks? Could be interesting to see as inlining is done during PE. (depending on the strategy you use for inlining)

compiler/src/org.graalvm.compiler.truffle.compiler.benchmark/src/org/graalvm/compiler/truffle/compiler/benchmark/pelang/PELangIfNode.java
```diff
@@ -0,0 +1,50 @@
...
+        this.conditionNode = conditionNode;
+        this.thenNode = thenNode;
+        this.elseNode = elseNode;
+    }
+
+    @Override
+    public Object executeGeneric(VirtualFrame frame) {
+        long conditionResult = conditionNode.evaluateCondition(frame);
+
+        if (conditionResult == 0L) {
```
chumer: Normal languages would have a counting condition profile here. Not sure if you want to be close to languages for this.

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/PartialEvaluator.java
```diff
@@ -231,6 +231,47 @@ public StructuredGraph createGraph(DebugContext debug, final CompilableTruffleAS
         return graph;
     }
 
+    @SuppressWarnings({"try", "unused"})
```
chumer: Doesn't look it was necessary to move this method. Please put it back if possible.

compiler/src/org.graalvm.compiler.truffle.compiler.benchmark/src/org/graalvm/compiler/truffle/compiler/benchmark/pelang/PELangExpressionBlockNode.java
```diff
@@ -0,0 +1,48 @@
...
+public final class PELangExpressionBlockNode extends PELangExpressionNode {
+
+    @Children private final PELangExpressionNode[] bodyNodes;
+
+    public PELangExpressionBlockNode(PELangExpressionNode[] bodyNodes) {
+        this.bodyNodes = bodyNodes;
+    }
+
+    @Override
+    public Object executeGeneric(VirtualFrame frame) {
```
chumer: `@ExplodeLoop` is missing here.

compiler/src/org.graalvm.compiler.truffle.compiler.benchmark/src/org/graalvm/compiler/truffle/compiler/benchmark/pelang/PELangExpressionNode.java
```diff
@@ -0,0 +1,51 @@
...
+import com.oracle.truffle.api.nodes.Node;
+import com.oracle.truffle.api.nodes.UnexpectedResultException;
+
+@TypeSystemReference(PELangTypes.class)
+public abstract class PELangExpressionNode extends Node {
+
+    public abstract Object executeGeneric(VirtualFrame frame);
+
+    public long executeLong(VirtualFrame frame) throws UnexpectedResultException {
+        return PELangTypesGen.expectLong(executeGeneric(frame));
```
chumer: Doesn't look its worth using a typesystem to generate these to expectLong methods. I'd just remove the typesystem and inline its implementation here.

flortsch:
Thanks @chumer for your feedback. It's highly appreciated. Could you elaborate on what you mean by support calls? The Graal/Truffle platform is still very new to me as I am doing this in a project for my master's. And you are right, the framework and the benchmarks will be further developed.

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/PartialEvaluator.java
flortsch: I thought about moving it to the other public methods, as I had to change the modified in order to access the method for benchmarking purposes. But I will move it back if required.

compiler/src/org.graalvm.compiler.truffle.compiler.benchmark/src/org/graalvm/compiler/truffle/compiler/benchmark/pelang/PELangExpressionBlockNode.java
flortsch: Thanks for spotting this.

compiler/src/org.graalvm.compiler.truffle.compiler.benchmark/src/org/graalvm/compiler/truffle/compiler/benchmark/pelang/PELangExpressionNode.java
flortsch: Yeah you are right, I will definitively do this.

boris-spas:
Is there still interest in merging this PR?
If the conflicts are resolved I'd try to shepherd it.

flortsch:
I think this PR can be closed for now. We did some bigger changes in the language itself and the benchmarking classes. We are also working on some partial evaluation improvements in general. So I think it's better when we come up with a new PR in the future.

PR #330, opened 2018-04-09
Add AArch64 support for the CAS instruction

daniel-stewart-qdt:
This adds the basic support for the CAS instruction in AArch64. It supports the various acquire/release variants, though it appears that only the acquire/release variant would currently be used. It also allows the CASAL instruction to be used for the lowering of a CompareAndSwapOp if the corresponding support is there in the hardware. 

This takes the same approach as HotSpot in dealing with the UseLSE flag. That is, if either the flag is thrown OR the hardware support is there, the CAS instruction will be used.

### DIFF
--- a/compiler/mx.compiler/suite.py
+++ b/compiler/mx.compiler/suite.py
@@ -674,6 +674,19 @@ suite = {
       "workingSets" : "Graal,LIR,AArch64",
     },
 
+    "org.graalvm.compiler.lir.aarch64.jdk11" : {
+      "subDir" : "src",
+      "sourceDirs" : ["src"],
+      "dependencies" : [
+        "org.graalvm.compiler.lir.aarch64",
+      ],
+      "annotationProcessors" : ["GRAAL_OPTIONS_PROCESSOR"],
+      "checkstyle" : "org.graalvm.compiler.graph",
+      "javaCompliance" : "11",
+      "workingSets" : "Graal,LIR,AArch64",
+      "multiReleaseJarVersion" : "11",
+    },
+
     "org.graalvm.compiler.lir.amd64" : {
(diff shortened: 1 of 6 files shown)


daniel-stewart-qdt:
I see that this is checked against version 9 of OpenJDK. I have added the CPUFeature flag support to the tip of OpenJDK. So the checks will fail because the CPUFeature flags don't exist in OpenJDK9. How is this handled when two projects are moving together?

dougxc:
The solution is to use [multi-release jars](http://openjdk.java.net/jeps/238). An example is in https://github.com/oracle/graal/commit/4ef5a08e45a40c3fc34c16bda909187333b53c70. You'll need to create a `org.graalvm.compiler.lir.aarch64.jdk11` project in `compiler/mx_compiler/suite.py` and set its `multiReleaseJarVersion` attribute to `11`. You then need to factor out the version dependent logic into a base version and an overlay for 11.

daniel-stewart-qdt:
I am splitting the commit into two. The first one is just the definition of the CAS instruction. It should not require any particular version of the JDK.

daniel-stewart-qdt:
I believe this is ready for review. 

The first commit just defines the CAS instruction itself for AArch64. It does not require any special version of OpenJDK.

The second commit uses the CAS instruction for implementing the CompareAndSwapOp if the hardware supports it OR if the UseLSE flag was set in OpenJDK. While this does allow a CAS instruction to be generated even if the hardware doesn't support it, this is how HotSpot implemented it and I simply copied their approach to this.

sanzinger:
Looks good to me. I have made a checkstyle fix. I'll integrate the change.

daniel-stewart-qdt:
Thanks!

PR #332, opened 2018-04-12
Fix typo

fniephaus:


dougxc:
I'll include this typo fix (as well as the `a an` -> `an` correction earlier on the same line) in one of my upcoming changes. Thanks for reporting!

PR #334, opened 2018-04-13
Aarch64 aot

theRealAph:
Changes in Graal needed to support AOT compilation on AArch64.


dougxc:
I assume these change has been tested locally with necessary hotspot and jvmci changes? There's not much extra testing we can do at this point in time given limited ARM64 hardware access.

theRealAph:
The OpenJDK changes are at http://cr.openjdk.java.net/~aph/8185505-1/

I have a bunch of testers looking at this.

dougxc:
Ok, once it passes the Travis gate and your testers give it the ok, I'll integrate this change.

nsjian:
We've also tested this changeset locally with JDK patch, and it worked fine.

theRealAph:
This should be good to merge now, Doug. Thanks for the help.

dougxc:
We will try and merge this as soon as possible.

PR #335, opened 2018-04-16
Add modified DaCapo benchmark collecting hardware counters

vhotspur:
This PR adds a new benchmark `dacapo-d3s` that uses a modified version of the DaCapo benchmark that can query hardware performance counters (Linux only).

Hardware counters are accessed via [PAPI](http://icl.cs.utk.edu/papi/) through a simple C agent. The agent is compiled automatically when running the suite for the first time.

The [C agent is available](https://github.com/D-iii-S/java-ubench-agent) at GitHub, the JAR with the modified DaCapo is hosted by [Department of Distributed and Dependable Systems at Charles University](http://d3s.mff.cuni.cz/software/benchmarking/).

As an extra feature, the results can be also stored in a CSV file.

Example run to collect total number of instructions and L1 cache misses with results stored in `avrora.csv`:

```shell
mx --primary-suite compiler \
  benchmark dacapo-d3s:avrora \
  -- -- \
  -e PAPI_TOT_INS,PAPI_L1_DCM -o avrora.csv
```

### DIFF
--- a/compiler/mx.compiler/mx_graal_benchmark.py
+++ b/compiler/mx.compiler/mx_graal_benchmark.py
@@ -800,7 +800,7 @@ class DaCapoD3SBenchmarkSuite(DaCapoBenchmarkSuite): # pylint: disable=too-many-
         return "DACAPO_D3S"
 
     def successPatterns(self):
-        return ['']
+        return []
 
     def resultFilter(self, values, iteration, endOfWarmupIndex):
         """Count iterations, convert iteration time to milliseconds."""
--- a/compiler/mx.compiler/mx_graal_benchmark.py
+++ b/compiler/mx.compiler/mx_graal_benchmark.py
@@ -784,6 +784,163 @@ class DaCapoMoveProfilingBenchmarkSuite(DaCapoMoveProfilingBenchmarkMixin, DaCap
 
 mx_benchmark.add_bm_suite(DaCapoMoveProfilingBenchmarkSuite())
 
+class DaCapoD3SBenchmarkSuite(DaCapoBenchmarkSuite): # pylint: disable=too-many-ancestors
+    """DaCapo 9.12 Bach benchmark suite implementation with D3S modifications."""
+
+    def name(self):
+        return "dacapo-d3s"
+
+    def daCapoSuiteTitle(self):
+        return "DaCapo 9.12-D3S-20180206"
+
+    def daCapoClasspathEnvVarName(self):
+        return "DACAPO_D3S_CP"
+
+    def daCapoLibraryName(self):
+        return "DACAPO_D3S"
+
+    def successPatterns(self):
+        return ['']
+
+    def resultFilter(self, values, iteration, endOfWarmupIndex):
+        """Count iterations, convert iteration time to milliseconds."""
+        # Called from lambda, increment call counter
+        iteration['value'] = iteration['value'] + 1
+        # Skip warm-up?
+        if iteration['value'] < endOfWarmupIndex:
+            return None
+
+        values['iteration_time_ms'] = str(int(values['iteration_time_ns']) / 1000 / 1000)
+        return values
+
+    def rules(self, out, benchmarks, bmSuiteArgs):
+        runArgs = self.postprocessRunArgs(benchmarks[0], self.runArgs(bmSuiteArgs))
+        if runArgs is None:
+            return []
+        totalIterations = int(runArgs[runArgs.index("-n") + 1])
+        out = [
+          mx_benchmark.CSVFixedFileRule(
+            self.resultCsvFile,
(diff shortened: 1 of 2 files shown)


compiler/mx.compiler/mx_graal_benchmark.py
```diff
@@ -784,6 +784,163 @@ def benchSuiteName(self):
...
+        return "DaCapo 9.12-D3S-20180206"
+
+    def daCapoClasspathEnvVarName(self):
+        return "DACAPO_D3S_CP"
+
+    def daCapoLibraryName(self):
+        return "DACAPO_D3S"
+
+    def successPatterns(self):
+        return ['']
```
axel22: I'm not sure that you should to override this. Does the standard output change in this suite?

If you do want to change the success patterns, then you would need to return a list of regex patterns here (see other `BaseDaCapoBenchmarkSuite` as an example).

Alternatively, if there is no success pattern, it should just be an empty list.

compiler/mx.compiler/suite.py
```diff
@@ -74,6 +74,13 @@
       "sha1" : "59b64c974662b5cf9dbd3cf9045d293853dd7a51",
     },
 
+    "DACAPO_D3S" : {
+      "urls" : [
+        "http://d3s.mff.cuni.cz/software/benchmarking/files/dacapo-9.12-d3s.jar",
```
axel22: cc @gilles-duboscq Should this be moved to lafo, as e.g. `JAVA_ALLOCATION_INSTRUMENTER`?

compiler/mx.compiler/mx_graal_benchmark.py
```diff
@@ -784,6 +784,163 @@ def benchSuiteName(self):
...
+              "config.vm-flags": self.shorten_vm_flags(self.vmArgs(bmSuiteArgs)),
+              "metric.name": "warmup",
+              "metric.value": ("<iteration_time_ms>", int),
+              "metric.unit": "ms",
+              "metric.type": "numeric",
+              "metric.score-function": "id",
+              "metric.better": "lower",
+              "metric.iteration": ("$iteration", int)
+            },
+            filter_fn=lambda x, counter={'value': 0}: self.resultFilter(x, counter, 0)
```
axel22: If I understood correctly, the default parameter is initialized only once per lambda, so this is effectively creating some state associated with that lambda. My python-fu might be lacking, but are you sure this is a standard idiom? If not, maybe a better idea is to initialize counter as a local variable inside `def rules`.

axel22: One other thing I'm slightly worried about here is that `filter_fn` might be called with the assumption that it is pure function. Will that always be the case?

https://github.com/graalvm/mx/blob/master/mx_benchmark.py#L525

compiler/mx.compiler/suite.py
gilles-duboscq: We can mirror it on lafo but the d3s url can stay

gilles-duboscq: You can now add this URL: https://lafo.ssw.uni-linz.ac.at/pub/graal-external-deps/dacapo-9.12-d3s.jar
It would be more convenient for us if you put the lafo URL first in the list and then the d3s.mff.cuni.cz one.

compiler/mx.compiler/mx_graal_benchmark.py
vhotspur: The standard output changed but it makes sense here to check for failures only. Should be fixed now.

compiler/mx.compiler/suite.py
vhotspur: Thanks, I have added lafo as the primary server.

compiler/mx.compiler/mx_graal_benchmark.py
axel22: So the CSV file does not have any information about whether the row is a part of the warmup or the final measurement? In that case, let's keep it as-is.

axel22:
LGTM

compiler/mx.compiler/mx_graal_benchmark.py
vhotspur: Exactly, the rows are not numbered or marked in any way. Perhaps that is something we should eventually add.

axel22: Before we merge, could you please add a comment above this lambda, explaining that assumes that it's invoked in only one batch of calls, because it has internal state?

farquet:
Also please sign the OCA, such that we can merge your PR.

vhotspur:
Hello, it seems that OCA has been processed. Is it somehow possible to reschedule the `oca-check`?

PR #340, opened 2018-04-18
README.md, License: Fix component list format

fweimer:


dougxc:
The readme has now been simplified, making this PR obsolete. Thanks for suggesting the improvement though!

eregon:
The updated README still has the same problem:
https://github.com/oracle/graal#license
A list as in this PR would be much clearer.

dougxc:
Will will fold this format changes into an upcoming PR that further improves the README.

PR #350, opened 2018-04-18
Fix broken links in README

kcsongor:
A couple of links were dead (possibly after moving some files around?)


thomaswue:
Thanks for the report! We will fix the links asap.

kcsongor:
It's probably easier if someone just fixes them, as filling out the OCA seems quite involved.

PR #383, opened 2018-04-28
Update maximum file name length to support AUFS

zhongweiy:
AUFS is the default file system for Docker container on some Linux
distribution. And the maximum filename length on AUFS is 242.


zhongweiy:
Hi, all, 
Could you help take a review?

zhongweiy:
Thanks for your review.

PR #391, opened 2018-05-01
Add support for binary sources

sudosalim:
Provide support for languages that need to maintain the original bytes of the source file.

For instance, Sulong currently ignores the file read by the Source class and re-read it again at the Runner class as the original bytes are not preserved. This trick is not possible when reading from standard input.

### DIFF
--- a/sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Engine.java
+++ b/sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Engine.java
@@ -587,6 +587,11 @@ public final class Engine implements AutoCloseable {
             }
 
             @Override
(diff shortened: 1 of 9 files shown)


sudosalim:
OCA name: Salim Shaaban Salim

sudosalim:
Hi everyone, 
does anybody know why clientManagedSourceChange and other two tests uses test files that are not there? Should these files be present? (test.input, some.js)

Because currently gate fails on NoSuchFileException

Thanks

boris-spas:
Closing becuse the PR is from an unknown repository. 
Please reopen if still relevant.

sudosalim:
I think this support is already added from Truffle 19.0 - so not relevant anymore. Thanks

PR #399, opened 2018-05-03
Add support for getAndAddInt() and getAndAddLong() intrinsics to AArch64

daniel-stewart-qdt:
This is an implementation of the getAndAddInt() and getAndAddLong() intrinsics for AArch64. This should support both AArch64 systems with 8.1 extensions and those without. It uses acquire/release semantics as I don't currently see a way to distinguish if we need them or if that information can be passed in. The registration of the unsafe plugin code was taken from AMD and therefore copies that structure, even though this patch doesn't currently support Objects. In the future, additions to the registration code could, and therefore that part of the code which includes Java Objects in an enumeration was left in, even though the code immediately looks for that and passes over it.


daniel-stewart-qdt:
Just a ping to see if I can get this reviewed.

sanzinger:
I'll take care of it right now.

axel22:
Recent merge of Amd64 JDK9-related intrinsics for `getAndSet` and `getAndAdd` for sub-32-bit data types might affect this PR:

https://github.com/oracle/graal/commit/acd7e77ec781b12c5616333df7192b6c606a6c1e

sanzinger:
I have created a fix for this.

sanzinger:
FYI, I have consolidated the versioned classes in `AArch64AtomicMove` class to only have the actual `useLSE` check versioned.

daniel-stewart-qdt:
Thank you very much!

PR #429, opened 2018-05-22
Add RoundNode for intrinsic optimization on AArch64

zhongweiy:


zhongweiy:
Hi, there,
I've signed OCA already. Could you help add my mail to the OCA list? Thanks!

zhongweiy:
Hi, all, could you please take a review?

axel22:
It might make sense to add a unit test here (by subclassing `GraalCompilerTest`), to ensure that the intrinsic behaves in the same way as `Math` in special cases.

zhongweiy:
@axel22 Because there is already a test: `Math_round` [link](https://github.com/oracle/graal/blob/3b1d7f2e9c3bb6a57600dbbcad20e5ecea36a20f/compiler/src/org.graalvm.compiler.jtt/src/org/graalvm/compiler/jtt/lang/Math_round.java), do you mean add more test case than what `Math_round` does.

axel22:
I missed that one - I think there's no need to add a test then - looks good to me.

sanzinger:
LGTM, I will integrate this change later today.

PR #446, opened 2018-06-04
Fix getAndAddInt() to take an Allocatable delta

daniel-stewart-qdt:
Fix for the getAndAddInt() function to take an AllocatableValue.


daniel-stewart-qdt:
Not sure I understand the failures. It seems like maybe the JDK11 requirement is not getting met? Is there something I need to change in order to find the right JDK?

sanzinger:
I am currently integrating this fix.

dougxc:
Did you see my comment above @sanzinger on the fix needed for the Travis failure?

daniel-stewart-qdt:
I did. I assume that is still needed even though @sanzinger is integrating?

dougxc:
I've canceled the integration so that you can push the change I suggested to ensure it fixes the Travis issue.

sanzinger:
Thank you for reminding me on this. I'll add it.

sanzinger:
Ok, now it seems to be clear.

sanzinger:
Wow, I was even able to update the branch in your repository fork.

PR #462, opened 2018-06-11
[doc] Update REFLECTIONS.md, change invalid urls.

cqjjjzr:
Already signed OCA, but only OpenJDK, is it applicable?


cqjjjzr:
There must be somewhere wrong-- I signed using my current email address, on Aug.5 2017, please check it. Also my name(CharlieJiang) is on http://www.oracle.com/technetwork/community/oca-486395.html#j

peter-hofer:
Thank you for your patch. However, the paragraph in question has been removed because the limitation it describes no longer exists.

PR #487, opened 2018-06-21
Update CheckGraalIntrinsics test case for AArch64.

zhongweiy:


zhongweiy:
Hi, could you please take a review? Thanks.

PR #489, opened 2018-06-21
JDK-8202533: [Graal] Implement StringUTF16.compress & StringLatin1.inflate

phedlin:
Preliminary version for feedback on adopting C2 intrinsics "compress" and "inflate" to Graal.

This is a sub-task to JDK-8193241 [Graal] implement missing intrinsics.


compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64MacroAssembler.java
```diff
@@ -773,4 +777,412 @@ public void stringIndexOf(Register str1, Register str2,
...
+     *   dst   (rdi) the start address of destination byte[] vector
+     *   len   (rdx) the length
+     *   vtmp1 (xmm)
+     *   vtmp2 (xmm)
+     *   vtmp3 (xmm)
+     *   vtmp4 (xmm)
+     *   tmp   (gpr)
+     *   res   (rax) the result code (length on success, zero otherwise)
+     */
+    public void char_array_compress(Register src, Register dst, Register len,
```
christhalinger: Maybe I'm alone with my opinion but I don't see a reason why having a method for these in MacroAssembler is better than just having the code in the Op. All the ops I've added in the past just have the assembler code in them. Makes it somewhat easier to read and understand.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64MacroAssembler.java
```diff
@@ -773,4 +777,412 @@ public void stringIndexOf(Register str1, Register str2,
...
+    /* Copy (inflate) a Latin1 string using a byte[] array representation into
+     * a UTF16 string using a char[] array representation.
+     *
+     *   src    (rsi) the start address of source byte[] to be inflated
+     *   dst    (rdi) the start address of destination char[] array
+     *   len    (rdx) the length
+     *   vtmp   (xmm)
+     *   tmp    (gpr)
+     */
+    public void byte_array_inflate(Register src, Register dst, Register len,
```
christhalinger: Same here.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGeneratorTool.java
```diff
@@ -268,6 +268,26 @@ default Variable emitStringIndexOf(Value sourcePointer, Value sourceCount, Value
         throw GraalError.unimplemented("String.indexOf substitution is not implemented on this architecture");
     }
 
+    // The routines emitStringLatin1Inflate/3 and emitStringUTF16Compress/3 models a simplified
```
dougxc: Please use `/* ... */` style comments for multiline comments. If you want to prevent `mx eclipseformat` from re-formatting the comment, use `// @formatter:off`  and `// @formatter:on`. For example:
```
    // @formatter:off
    /*
     * The routines emitStringLatin1Inflate/3 and emitStringUTF16Compress/3
     * models a simplified version of:
     *
     *    emitStringLatin1Inflate(Value src, Value src_ndx, Value dst, Value dst_ndx, Value len)
     *
     * and
     *
     *    emitStringUTF16Cpmpress(Value src, Value src_ndx, Value dst, Value dst_ndx, Value len)
     *
     * respectively, where we have hoisted the offset address computations
     * in a method replacement snippet.
     */
    // @formatter:on
```

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/StringCompressInflateTest.java
```diff
@@ -0,0 +1,237 @@
...
+@AddExports({"java.base/java.lang"})
+public final class StringCompressInflateTest extends MethodSubstitutionTest {
+
+    final static int N = 1000;
+
+    @Test
+    public void testStringLatin1Inflate() throws ClassNotFoundException,
+                                                 UnsupportedEncodingException
+    {
+        if (Java8OrEarlier)
```
dougxc: Consider using `org.junit.Assume.assumeFalse(Java8OrEarlier)` instead for this kind of thing.

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/StringCompressInflateTest.java
```diff
@@ -0,0 +1,237 @@
...
+            return;   // StringLatin1.inflate introduced in Java 9.
+        if (!(getTarget().arch instanceof AMD64))
```
dougxc: Same here.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64MacroAssembler.java
phedlin: Essentially I'm just following suite with *AMD64StringIndexOfOp*. However, I would argue that providing the implementation of two rather large and similar/related assembly routines are best located together instead of spread out in two not obviously related files, interleaved with what is boilerplate only. This way, the assembly routines read very much like regular assembly code (a suitable idiom) and the boilerplate is kept in the "Op" classes. I think it makes for considerably easier reading of the assembly while keeping the boilerplate code at a "follow-the-template" level.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64MacroAssembler.java
```diff
@@ -773,4 +777,412 @@ public void stringIndexOf(Register str1, Register str2,
...
+        assert vtmp.getRegisterCategory().equals(AMD64.XMM);
+
+        Label L_done = new Label();
+        Label L_below_threshold = new Label();
+
+        assert src.number != dst.number && src.number != len.number && src.number != tmp.number;
+        assert dst.number != len.number && dst.number != tmp.number;
+        assert len.number != tmp.number;
+
+        if (UseAVX > 2 && (supports(CPUFeature.AVX512BW) &&
```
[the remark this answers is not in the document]
phedlin: I don't see that "config.useAVX" (from HS: "UseAVX") is actually propagated to the feature setup (not in the code I have). The "config.useSSE" is indeed but the interpretation has been changed such that it is no longer possible to turn SSE code generation off with "UseSSE=0". Moreover, the Graal interpretation of "config.useSSE" has been made AMD64 exclusive which is not the case in HS. In short, I just bailed out to take an orthogonal approach for now.

compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/CheckGraalIntrinsics.java
```diff
@@ -376,9 +376,9 @@ public CheckGraalIntrinsics() {
                             "java/lang/StringLatin1.indexOf([BI[BII)I",
                             "java/lang/StringLatin1.indexOf([B[B)I",
                             "java/lang/StringLatin1.inflate([BI[BII)V",
-                            "java/lang/StringLatin1.inflate([BI[CII)V",
+                            //"java/lang/StringLatin1.inflate([BI[CII)V",
```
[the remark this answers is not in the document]
phedlin: Sure. There is some cleanup lacking.

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/StringCompressInflateTest.java
phedlin: I found the use in the code to be inconclusive, thus adopting the one more clear to me, with obvious operational interpretation (as opposed to the quasi-declarative version in *junit* using an implicit operational realisation, hiding the control-flow -- which I consider a rather poor approach). I don't think the "junit.Assume" support adds anything in this case.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64MacroAssembler.java
christhalinger: Look at `AMD64ArrayEqualsOp` instead.

You could argue that having similar assembly code in two different files makes it easier to look at/compare because you could open the two files side-by-side :-)  Weak argument, I know...

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/StringCompressInflateTest.java
tkrodriguez: The main difference is that the junit output includes a message indicating the test passed because of an assumption.  assumeTrue is broadly how we disable tests which only run on certain configurations.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64MacroAssembler.java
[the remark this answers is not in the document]
tkrodriguez: Well we don't expose UseAVX as value since everything is expressed in terms of CPUFeatures.  If there are levels above AVX2 that we need to support then they need to be added to CPUFeatures in JVMCI and propagated that way.  I don't think our assembly has working support for EVEX as the point either.

compiler/src/org.graalvm.compiler.replacements.amd64/src/org/graalvm/compiler/replacements/amd64/AMD64StringLatin1InflateNode.java
```diff
@@ -0,0 +1,106 @@
...
+        this.len = len;
+    }
+
+    @Override
+    public LocationIdentity getLocationIdentity() {
+        // Model read access via 'src' using:
+        return NamedLocationIdentity.getArrayLocation(JavaKind.Byte);
+    }
+
+    @Override
```
tkrodriguez: You can implement `MemoryCheckpoint.Single` instead and then you won't need this override.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64MacroAssembler.java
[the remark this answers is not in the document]
tkrodriguez: The path by which UseAVX propagates into CPUFeatures is indeed complex but it goes through VM_Version::Feature_Flags which a driven by UseAVX.  For levels above 2 I think `jdk.vm.ci.hotspot.amd64.AMD64HotSpotJVMCIBackendFactory#computeFeatures` will have to directly test against UseAVX since there are no FeatureFlags for them.

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/StringCompressInflateTest.java
phedlin: Fair point.

compiler/src/org.graalvm.compiler.replacements.amd64/src/org/graalvm/compiler/replacements/amd64/AMD64StringLatin1InflateNode.java
phedlin: Hmm, I actually did start off trying to use "Single" but could not make it produce the proper graph... perhaps I didn't get the annotation/properties right to begin with...

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -72,6 +72,10 @@
 import jdk.vm.ci.code.TargetDescription;
 import jdk.vm.ci.meta.PlatformKind;
 
+// @formatter:off
```
dougxc: This effectively turns off formatting for the whole file. Please remove it. The pattern for disabling/re-enabling formatting is:
```
// @formatter:off
<code that you don't want reformatted>
// @formatter:on
```

As I mentioned offline, the Eclipse re-formatting of commented boolean constant arguments in calls to the `AMD64InstructionAttr` is unfortunate. @mur47x111 is converting these booleans to a set of enum values.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -72,6 +72,10 @@
...
+
+// Checkstyle: stop
```
dougxc: Please don't disable Checkstyle for the whole file.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -825,7 +863,7 @@ protected AMD64RRROp(String opcode, int prefix1, int prefix2, int op, boolean ds
         // TEST is documented as MR operation, but it's symmetric, and using it as RM operation is more convenient.
         public static final AMD64RMOp TESTB  = new AMD64RMOp("TEST",               0x84, OpAssertion.ByteAssertion);
         public static final AMD64RMOp TEST   = new AMD64RMOp("TEST",               0x85);
-        // @formatter:on
+        // @formatter:off
```
dougxc: All these `// @formatter:off` comments need to be paired with a `// @formatter:on` comment.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64InstructionAttr.java
```diff
@@ -164,6 +175,11 @@ public void setEvexEncoding(int value) {
         evexEncoding = value;
     }
 
+    // XXX: Tentative.
```
dougxc: This comment could be more helpful.

compiler/src/org.graalvm.compiler.replacements.amd64/src/org/graalvm/compiler/replacements/amd64/AMD64StringLatin1InflateNode.java
dougxc: @tkrodriguez I think the issue is that `AMD64StringLatin1InflateNode` reads from a `byte[]` and writes to a `char[]`. `getLocationIdentity` is modelling the read (by implementing `MemoryAccess`) and `getLocationIdentities()` is modelling the write (by implementing `MemoryCheckpoint.Multi`). It's an interesting trick to disambiguate modelling the read and write in the memory graph.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64InstructionAttr.java
phedlin: Indeed.

compiler/src/org.graalvm.compiler.replacements.amd64/src/org/graalvm/compiler/replacements/amd64/AMD64StringLatin1InflateNode.java
phedlin: Yes, but perhaps not necessary?

gilles-duboscq: I think the current code with `Multi` is correct. While we *might* get lucky and not break anything if we use `Single` because this node is fixed, i think it's better to model things precisely so that the rest of the compiler can make correct decisions.

tkrodriguez: Yes I missed that it was reading and writing different types.  I think a comment explaining this slightly odd usage would be a good idea.

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/StringCompressInflateTest.java
```diff
@@ -0,0 +1,244 @@
...
+        // Test case is (currently) AMD64 only.
+        org.junit.Assume.assumeTrue(getTarget().arch instanceof AMD64);
+
+        Class<?> javaclass = Class.forName("java.lang.StringLatin1");
+        Class<?> testclass = AMD64StringLatin1InflateNode.class;
+
+        TestMethods tms = new TestMethods("testInflate", javaclass,
+                                              "inflate", byte[].class, int.class,
+                                                         char[].class, int.class, int.class);
+
```
tkrodriguez: Why is there so much vertical whitespace in this file?

PR #492, opened 2018-06-22
Moving jdk.aot (jaotc) module to Graal repo.

mukel:


compiler/mx.compiler/mx_jaotc.py
```diff
@@ -0,0 +1,127 @@
...
+from argparse import ArgumentParser
+
+import mx
+import mx_compiler
+from mx_gate import Task
+
+jdk = mx.get_jdk(tag='default')
+
+
+def run_jaotc(args, cwd=None):
```
dougxc: please add a pydoc comment here that will also provides the text for `mx help` and `mx help jaotc`. For example:
```+def run_jaotc(args, cwd=None):
    """run AOT compiler with classes in this repo instead of those in the JDK"""
```

compiler/mx.compiler/mx_jaotc.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def jaotc_gate_runner(tasks):
+    with Task('jaotc', tasks, tags=['jaotc', 'fulltest']) as t:
+        if t: jaotc_tests([])
+
+
+def jaotc_tests(args):
+    parser = ArgumentParser(prog='mx jaotc-tests')
+    parser.add_argument('test', help='test to run', nargs='*')
```
dougxc: Should the help text be `tests to run (omit to run all tests)`?
Also, since the set of tests is closed, you could add `choices=['HelloWorld']`.

dougxc:
Once https://bugs.openjdk.java.net/browse/JDK-8205025 shows up in a JDK 11 early access build, we should add JDK 11 testing to Travis and run the AOT tests. This can be done in a subsequent PR.

mukel:
Add `mx jaotc` and `mx jaotc-test` to run/test the AOT compiler.
Besides compiling and running HelloWorld I added a test to aot-compile `java.base` since I ran into an issue with some methods in `java-base` and class fingerprints (missing `-Djvmci.UseProfilingInformation=false` int the `mx jaotc` launcher).
A similar issue still occurs when running `jaotc` with `graal-enterprise` (even with the proper flags):
```
..Error: Failed compilation: java.util.stream.IntStream.rangeClosed(II)Ljava/util/stream/IntStream;: org.graalvm.compiler.debug.GraalError: should not reach here: Metaspace constant load should not be happening directly
        at lir instruction: B19@206 org.graalvm.compiler.hotspot.amd64.AMD64HotSpotMove$HotSpotLoadMetaspaceConstantOp rsi|QWORD = HOTSPOTLOADMETASPACECONSTANT input: meta{HotSpotType<Ljava/util/stream/Streams$StreamBuilderImpl;, resolved>}
[B0, null, B3, B28, null, B31, null, B34, null, B37, null, B40, B42, B43, B44, B46, B47, B4, null, B7, null, B10, null, B13, null, B16, null, B19, B21, B22, B23, B25, B26, B20, B41, B45, B24, B1, B29, B32, B14, B17, B8, B38, B35, B11, B5, B48, B27]
...
52072 methods compiled, 63 methods failed (126572 ms)
...
```
 * `-Djvmci.UseProfilingInformation=false` is not needed anymore (see 37c85c42b4a9fce819e561ff9b285c740ed9cc47) I also removed `-Dgraal.UseExceptionProbability=false` from the [mx jaotc](https://github.com/mukel/graal/blob/bd55d1c455c9bb61eb54e509541cc972f62d203b/compiler/mx.compiler/mx_jaotc.py#L57-L73) launcher.
* `mx gate -t jaotc` takes ~10 minutes. @dougxc 
* `jaotc` class search unittests are in. i.e `mx unittest jdk.tools.jaotc`
* Tested `mx updategraalinopenjdk` manually (build a fresh JDK and ran `build-bootmodules.sh` and the `test-*.sh` scripts).
* I couldn't test on ARM, but a weekly build was added. @theRealAph 
* `jtreg` tests are pending (not in the gate).
* CDS is not compatible with `--upgrade-module-path` used to inject `jdk.aot` via `mx jaotc`.

PR #504, opened 2018-06-27
Add reserved stack access support to hotspot

dean-long:


tkrodriguez:
look good to me.

dean-long:
@theRealAph These aarch64 changes seem to work.  Please take a look.

theRealAph:
AArch64 looks OK, but there is some non-idiomatic code:

movx( should be mov(64, . I don't know why movx exists at all, and it's only used in a couple of places. It shouldn't be used in new code.

subs should be cmp.

Thanks.

dean-long:
@theRealAph I tried using "cmp" before, but it mapped "sp" to "zr" and generated bad code.  I could only get it to work using "subs".  I'll change the movx.

PR #506, opened 2018-07-01
String equals jdk9

dsamersoff:
Enable string equals substitution for jdk9 and later


tkrodriguez:
Looking at this more closely I think this change was obviated by 4cfc941ef83 which added a more generic version of this intrinsic.  So I'm not sure it's required to have intrinsics for StringLatin1.equals and StringUTF16.equals.  As far as I can tell the new code paths are never executed in the current graal.

tkrodriguez:
I pushed your test changes from this PR in 093fca60e7b23f72c9558b4fb1fe3a2281b64b8d along with some changes to CheckGraalIntrinsics to mark those intrinsics as handled.

Issue #539, opened 2018-07-12
java.lang.reflect.Parameter usage prevents native image generation


stuartwdouglas:
I got the following error when attempting to compile a jar that included eclipse Yasson (the JSON-B RI):

`
Error: Unsupported method java.lang.reflect.Executable.getParameters0() is reachable: Native method. If you intend to use the Java Native Interface (JNI), specify -H:+JNI and see also -H:JNIConfigurationFiles=<path> (use -H:+PrintFlags for details)
To diagnose the issue, you can add the option -H:+ReportUnsupportedElementsAtRuntime. The unsupported element is then reported at run time when it is accessed the first time.
Call path from entry point to java.lang.reflect.Executable.getParameters0():
at java.lang.reflect.Executable.getParameters0(Executable.java)
at java.lang.reflect.Executable.privateGetParameters(Executable.java:404)
at java.lang.reflect.Executable.getParameters(Executable.java:357)
at java.lang.reflect.Executable.getAllGenericParameterTypes(Executable.java:305)
at java.lang.reflect.Parameter.getParameterizedType(Parameter.java:201)
at java.lang.reflect.Parameter.toString(Parameter.java:125)
at com.oracle.svm.reflect.proxies.Proxy_1042_Object_toString.invoke(Unknown Source)
at java.lang.reflect.Method.invoke(Method.java:498)
at com.oracle.svm.core.JavaMainWrapper.run(JavaMainWrapper.java:173)
at com.oracle.svm.core.code.CEntryPointCallStubs.com_002eoracle_002esvm_002ecore_002eJavaMainWrapper_002erun_0028int_002corg_002egraalvm_002enativeimage_002ec_002etype_002eCCharPointerPointer_0029(generated:0)
`

It appears to be caused by the usage of java.lang.reflect.Parameter

cstancu:
The issue has been fixed.

PR #547, opened 2018-07-13
Compute parameters for Executable objects

dmlloyd:


dmlloyd:
I am already covered by the OCA.

dmlloyd:
Fixes #539

cstancu:
Thanks for your contribution. I'll take a close look early next week. There is an alternative solution that may be simpler.

panga:
I hit in the same error. Merging the @dmlloyd fix worked for me.

cstancu:
Yes, the changes are valid and fix the problem. I'll bring them in as part of a bigger refactoring of the reflection meta-data caching.

cstancu:
Thanks for your contribution. The PR has been merged. As I was hinting at above I additionally refactored the reflection metadata computation. We don't need a custom field recomputation for lazily computed field values. It is enough to invoke the method that leads to the field initialization in an object replacer, i.e., at image build time, to add that value to the native image heap. See [here](https://github.com/oracle/graal/commit/76595b624175650d143e7775dce1a1ae9d060368#diff-8d00382cbe8559ce3a72dc0b4c36f093R48).

PR #553, opened 2018-07-19
Handle special case where offendingSymbol is EOF

numberpi:
Fixes https://github.com/graalvm/simplelanguage/issues/40


numberpi:
Oh, just realized that I fixed the generated parser, not the grammar. Will create a new PR.

PR #554, opened 2018-07-19
Reinstate JarURLConnection

dmlloyd:
Since JarFile is back, we can now use JarURLConnection again.


cstancu:

PR #555, opened 2018-07-19
Handle special case where offendingSymbol is EOF

numberpi:
Fixes graalvm/simplelanguage#40

### DIFF
--- a/truffle/src/com.oracle.truffle.sl.test/src/com/oracle/truffle/sl/test/SLParseErrorTest.java
+++ b/truffle/src/com.oracle.truffle.sl.test/src/com/oracle/truffle/sl/test/SLParseErrorTest.java
@@ -74,4 +74,16 @@ public class SLParseErrorTest {
             Assert.assertNotNull("Should have source section.", e.getSourceLocation());
         }
     }
+
+    @Test
+    public void testParseErrorEOF() throws IOException {
+        try {
+            final Source src = Source.newBuilder("sl", "function", "testSyntaxErrorEOF.sl").build();
+            context.eval(src);
+            Assert.assertTrue("Should not reach here.", false);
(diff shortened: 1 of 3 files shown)


numberpi:
There is still another bug in there.
When adding another test for parsing an empty sl code snippet:
```java
    @Test
    public void testParseErrorEOFEmpty() throws IOException {
        try {
            final Source src = Source.newBuilder("sl", "", "testSyntaxErrorEOFEmpty.sl").build();
            context.eval(src);
            Assert.assertTrue("Should not reach here.", false);
        } catch (PolyglotException e) {
            Assert.assertTrue("Should be a syntax error.", e.isSyntaxError());
            Assert.assertNotNull("Should have source section.", e.getSourceLocation());
        }
    }
```
It throws:
```
java.lang.IllegalArgumentException: startColumn < 1
	at com.oracle.truffle.api.source.Source.createSection(Source.java:585)
	at com.oracle.truffle.sl.parser.SLParseError.getSourceLocation(SLParseError.java:67)
	at com.oracle.truffle.api.vm.PolyglotExceptionImpl.<init>(PolyglotExceptionImpl.java:105)
	at com.oracle.truffle.api.vm.PolyglotExceptionImpl.<init>(PolyglotExceptionImpl.java:80)
	at com.oracle.truffle.api.vm.PolyglotImpl.wrapGuestException(PolyglotImpl.java:279)
	at com.oracle.truffle.api.vm.PolyglotContextImpl.eval(PolyglotContextImpl.java:791)
	at org.graalvm.polyglot.Context.eval(Context.java:313)
        ...
```

So it is a general question: Should it be allowed to create a SourceSection for an empty code snippet? If not, should `SLParseError.getSourceLocation()` return `null` in this case?

numberpi:
Outdated

PR #562, opened 2018-07-20
Inflater/Deflater implementation for Java 11

dmlloyd:


dmlloyd:
This brings in a JDK10/11 check.  This implementation does coexistent substitution for both 10-and-earlier ad 11-and-later.  It was build tested but I don't think 11 is testable just yet, at least not by mere mortals.

An alternative approach would be to introduce more aggressive deletions and substitute the entire API, providing a single common implementation between <=10 and >=11.  This might or might not end up being somewhat more complex of a solution.  On the one hand, more code would need replacement, but on the other hand, a lot of source optimization could be done to make the code compact that is not possible with a Java/JNI split such as is found in OpenJDK.

compiler/src/org.graalvm.compiler.serviceprovider/src/org/graalvm/compiler/serviceprovider/GraalServices.java
```diff
@@ -71,6 +71,11 @@ private static int getJavaSpecificationVersion() {
      */
     public static final boolean Java8OrEarlier = JAVA_SPECIFICATION_VERSION <= 8;
 
+    /**
+     * Determines if the Java runtime is version 8 or earlier.
+     */
+    public static final boolean Java11OrLater = JAVA_SPECIFICATION_VERSION >= 11;
```
Peter-B-Kessler: Do you really need to know if you are on 11 or later?  I don't think SubstrateVM ever plans to (or can) run on 9 or 10 because of JVMCI changes; I think we are going straight to 11.

I am also not sure if I can make changes to the public API of org.graalvm.compiler.serviceprovider.GraalServices.  But I will check.

dmlloyd: This API was added in Java 11 and is not present in 9 or 10, so saying 9-or-later might be confusing to say the least.  That said, it doesn't actually matter too much to me.  I can change it in whatever manner you think is best.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -154,26 +154,34 @@ private static long init(int level, int strategy, boolean nowrap) {
         }
     }
 
+    private static void doSetDictionary(long addr, CCharPointer buf, int len) {
```
Peter-B-Kessler: I usually like to put in a link to where the original source code is.  Like the minimalist comment at line 132.  The idea is to distinguish translations from new code (e.g., utility methods) that you write.

I would expand that to `src/java.base/share/native/libzip/Deflater.c` and hope that they don't rearrange the code base (again :-).

dmlloyd: All fixed.  I noted what JDK version the copied code came from as well.

compiler/src/org.graalvm.compiler.serviceprovider/src/org/graalvm/compiler/serviceprovider/GraalServices.java
Peter-B-Kessler: It turns out this is internal API, so adding `Java11OrLater` is fine.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -39,13 +41,12 @@
...
 import com.oracle.svm.core.annotate.TargetElement;
-import com.oracle.svm.core.jdk.JDK8OrEarlier;
-import com.oracle.svm.core.jdk.JDK9OrLater;
 import com.oracle.svm.core.posix.headers.LibC;
 import com.oracle.svm.core.posix.headers.ZLib;
 import com.oracle.svm.core.posix.headers.ZLib.z_stream;
 import com.oracle.svm.core.snippets.KnownIntrinsics;
-import com.oracle.svm.core.util.VMError;
+
+// See notes at the bottom of this file.
```
Peter-B-Kessler: This is a funny forward reference.  I think the comments belong at the top of the file.  If the importing of `JDK10OrEarlier` and `JDK11OrLater` are not enough of a tip-off that there is going to be version-specific interfaces and implementations in here.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -108,28 +109,29 @@ private static int updateByteBuffer(int crc, long addr, int off, int len) {
...
     @Alias //
-    @TargetElement(onlyWith = JDK8OrEarlier.class) //
+    @TargetElement(onlyWith = JDK10OrEarlier.class) //
     int off;
-
     @Alias //
-    @TargetElement(onlyWith = JDK8OrEarlier.class) //
+    @TargetElement(onlyWith = JDK10OrEarlier.class) //
     int len;
+    @Alias //
```
Peter-B-Kessler: I like the consistent formatting separating annotations from declarations.  I think there should be consistent use of whitespace: each (now) multi-line declaration should have whitespace around it.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -154,26 +156,43 @@ private static long init(int level, int strategy, boolean nowrap) {
...
+    //                       jbyte *output, jint outputLen,
+    //                       jint flush, jint params)
+    private static void doSetDictionary(long addr, CCharPointer buf, int len) {
+        int res = ZLib.deflateSetDictionary(WordFactory.pointer(addr), buf, len);
+        if (res == ZLib.Z_OK()) {
+            return;
+        } else if (res == ZLib.Z_STREAM_ERROR()) {
+            throw new IllegalArgumentException();
+        } else {
+            throw new InternalError();
```
Peter-B-Kessler: Is there some point in preserving the error message `((z_stream *)jlong_to_ptr(addr))->msg`?

Peter-B-Kessler: I know that was not done in `init` (above), but that is not an excuse to be sloppy.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -336,26 +418,42 @@ private static long init(boolean nowrap) {
         }
     }
 
-    // Inflater.c-Java_java_util_zip_Inflater_setDictionary(JNIEnv *env, jclass cls, jlong addr,
+    // (JDK 11)src/java.base/share/native/libzip/Inflater.c-static void doSetDictionary(JNIEnv *env, jlong addr, jbyte *buf, jint len)
+    private static void doSetDictionary(final long addr, final CCharPointer bytes, final int len) {
```
Peter-B-Kessler: We have checkstyle rules that enforce that method arguments are `final`.  (I had to be weaned from declaring them `finel` too.)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -209,62 +228,111 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
         }
     }
 
+    // (JDK 11)src/java.base/share/native/libzip/Deflater.c-static jlong doDeflate(JNIEnv *env, jobject this, jlong addr,
+    //                       jbyte *input, jint inputLen,
+    //                       jbyte *output, jint outputLen,
+    //                       jint flush, jint params)
+    private static long doDeflate(long addr, CCharPointer input, int inputLen, CCharPointer output, int outputLen, int flush, int params) {
+        z_stream strm = WordFactory.pointer(addr);
+        int inputUsed, outputUsed;
```
Peter-B-Kessler: One declaration per line, please.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -209,62 +228,111 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
...
+        int finished = 0;
+        int setParams = params & 1;
+
+        strm.set_next_in(input);
+        strm.set_next_out(output);
+        strm.set_avail_in(inputLen);
+        strm.set_avail_out(outputLen);
+
+        if (setParams != 0) {
```
Peter-B-Kessler: For converting a C `int` to a Java `boolean`, please use `CTypeConversion.toBoolean(int)`.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -209,62 +228,111 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
...
+            int strategy = (params >> 1) & 3;
+            int level = params >> 3;
+            int res = ZLib.deflateParams(strm, level, strategy);
+            if (res == ZLib.Z_OK() || res == ZLib.Z_BUF_ERROR()) {
+                if (res == ZLib.Z_OK()) {
+                    setParams = 0;
+                }
+            } else {
+                throw new InternalError();
```
Peter-B-Kessler: What happened to `strm->msg`?  If the message is not useful, then add a comment saying you are deliberately ignoring it.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
dmlloyd: OK I'll fix that.  Also I just figured out that mx has a checkstyle action, so I'll run that locally to verify.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
dmlloyd: It's a C int intentionally because of its role in the bit shift at the bottom of the method (this is mirroring the implementation code).  So I think it's better to keep it as-is: a value of 0 explicitly means do not set params, whereas a value of 1 means set params.  In this case I could use that conversion and it would happen to work because the domain of values is 0 or 1, but it's not really a boolean, it's a one-bit integer.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -209,62 +228,111 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
...
+        strm.set_avail_in(inputLen);
+        strm.set_avail_out(outputLen);
+
+        if (setParams != 0) {
+            int strategy = (params >> 1) & 3;
+            int level = params >> 3;
+            int res = ZLib.deflateParams(strm, level, strategy);
+            if (res == ZLib.Z_OK() || res == ZLib.Z_BUF_ERROR()) {
+                if (res == ZLib.Z_OK()) {
+                    setParams = 0;
```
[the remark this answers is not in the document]
dmlloyd: I can look into it.  Mainly I was trying to keep it as similar as possible to the C code, which uses switch.  But I can go for optimal style instead.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -209,62 +228,111 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
...
+                if (res == ZLib.Z_STREAM_END()) {
+                    finished = 1;
+                }
+            } else {
+                throw new InternalError();
+            }
+        }
+        inputUsed = inputLen - strm.avail_in();
+        outputUsed = outputLen - strm.avail_out();
+        return ((long) inputUsed) | (((long)outputUsed) << 31) | (((long)finished) << 62) | (((long)setParams) << 63);
```
Peter-B-Kessler: All the style guidelines I can find say there should be a space after the cast.  And having a space after the first is just weird.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -209,62 +228,111 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
...
     @Substitute
-    @TargetElement(onlyWith = JDK9OrLater.class)
-    @SuppressWarnings({"unused", "static-method"})
-    long deflateBytesBytes(long addr,
-                    byte[] inputArray, int inputOff, int inputLen,
-                    byte[] outputArray, int outputOff, int outputLen,
-                    int flush, int params) {
-        throw VMError.unsupportedFeature("JDK9OrLater: Target_java_util_zip_Deflater.deflateBytesBytes(long, byte[], int, int, byte[], int, int, int, int)");
+    @TargetElement(onlyWith = JDK11OrLater.class)
+    private long deflateBytesBytes(long addr, byte[] inputArray, int inputOff, int inputLen, byte[] outputArray, int outputOff, int outputLen, int flush, int params) {
```
Peter-B-Kessler: The C code nicely separates the `input*` and 'output*` formals on to their own lines.  I find that more readable.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -209,62 +228,111 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
...
-                    long outputAddress, int outputLen,
-                    int flush, int params) {
-        throw VMError.unsupportedFeature("JDK9OrLater: Target_java_util_zip_Deflater.deflateBufferBuffer(long, long, int, long, int, int, int)");
+    @TargetElement(onlyWith = JDK11OrLater.class)
+    private long deflateBufferBuffer(long addr, long inputBuffer, int inputLen, long outputBuffer, int outputLen, int flush, int params) {
+        return doDeflate(addr, WordFactory.pointer(inputBuffer), inputLen, WordFactory.pointer(outputBuffer), outputLen, flush, params);
     }
 
-    // Deflater.c-Java_java_util_zip_Deflater_getAdler(JNIEnv *env, jclass cls, jlong addr)
+    // (JDK 11)src/java.base/share/native/libzip/Deflater.c-ava_java_util_zip_Deflater_getAdler(JNIEnv *env, jclass cls, jlong addr)
```
Peter-B-Kessler: This method seems unchanged from JDK-8.  Is the `(JDK 11)` on the comment just  a notation about where code is now, as opposed to `jdk8u/jdk/src/share/native/java/util/zip/Deflater.c`?  If you are going to cite anything I would cite the JDK-8 location rather than making it look like this was a JDK-11 method.  (Even though it does not say `JDK11OrLater`.)

Peter-B-Kessler: Missing the initial `J` on `Java_java_util_zip_Deflater_getAdler`.

dmlloyd: Since the comments are for reference, I went with Java 11 because it will generally have the more up-to-date reference code and that's the version I looked at when I checked the implementation over.  I assume that you wanted these comments for reference/matching against the originals, not for marking what API methods come from what JDK versions.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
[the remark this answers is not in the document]
Peter-B-Kessler: Part of the confusion may be that in JNI C code, "throwing" an exception just posted it for discovery when crossing back into Java code.  That is, throwing an error did not affect the control flow.  In SubstrateVM substitutions, throwing an error throws an error, interrupting the control flow.

(This comment occurred to me when looking at `end(long)`, where it is more obvious that the two branches are not *alternatives*: if the exception is thrown the call to `LibC.free` will not happen.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -209,62 +228,111 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
...
     @Substitute
     private static int getAdler(long addr) {
         z_stream strm = WordFactory.pointer(addr);
         return (int) strm.adler().rawValue();
     }
 
-    // Deflater.c-Java_java_util_zip_Deflater_reset(JNIEnv *env, jclass cls, jlong addr)
+    // (JDK 11)src/java.base/share/native/libzip/Deflater.c-Java_java_util_zip_Deflater_reset(JNIEnv *env, jclass cls, jlong addr)
```
Peter-B-Kessler: Unchanged from JDK-8.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -209,62 +228,111 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
...
     @Substitute
     private static void reset(long addr) {
         if (ZLib.deflateReset(WordFactory.pointer(addr)) != ZLib.Z_OK()) {
             throw new InternalError();
         }
     }
 
-    // Deflater.c-Java_java_util_zip_Deflater_end(JNIEnv *env, jclass cls, jlong addr)
+    // (JDK 11)src/java.base/share/native/libzip/Deflater.c-Java_java_util_zip_Deflater_end(JNIEnv *env, jclass cls, jlong addr)
```
Peter-B-Kessler: Unchanged from JDK-8

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -108,28 +109,29 @@ private static int updateByteBuffer(int crc, long addr, int off, int len) {
...
-    @Alias private int strategy;
-    @Alias private boolean setParams;
-    @Alias private boolean finish;
-    @Alias private boolean finished;
-
-    // Deflater.c-Java_java_util_zip_Deflater_init(JNIEnv *env, jclass cls, jint level,
+    // (JDK 11)src/java.base/share/native/libzip/Deflater.c-Java_java_util_zip_Deflater_init(JNIEnv *env, jclass cls, jint level, jboolean nowrap)
     @Substitute
     private static long init(int level, int strategy, boolean nowrap) {
         z_stream strm = LibC.calloc(WordFactory.unsigned(1), SizeOf.unsigned(z_stream.class));
```
Peter-B-Kessler: [An informational note: If you were not in Posix code, you could use `UnmanagedMemory.calloc(UnsignedWord)` to do the same thing in a platform-independent way.]

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
Peter-B-Kessler: You are right.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -330,32 +435,48 @@ private static long init(boolean nowrap) {
...
             }
         }
     }
 
-    // Inflater.c-Java_java_util_zip_Inflater_setDictionary(JNIEnv *env, jclass cls, jlong addr,
+    // (JDK 11)src/java.base/share/native/libzip/Inflater.c-static void doSetDictionary(JNIEnv *env, jlong addr, jbyte *buf, jint len)
+    private static void doSetDictionary(long addr, CCharPointer bytes, int len) {
+        z_stream strm = WordFactory.pointer(addr);
+        int res = ZLib.inflateSetDictionary(strm, bytes, len);
+        if (res == ZLib.Z_OK()) {
```
Peter-B-Kessler: Here is another place where `return` and `throw` are not if-then-else alternatives: If you return, you have returned; if you throw you have thrown an exception.

The generated code won't be any different, but the `else`s don't have to be in the source code.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -383,75 +504,145 @@ private int inflateBytes(long addr, byte[] b, int off, int len) throws DataForma
...
+            outputUsed = outputLen - strm.avail_out();
+        } else if (ret == ZLib.Z_DATA_ERROR()) {
+            // these fields act as "out" parameters
+            inputConsumed = inputLen - strm.avail_in();
+            outputConsumed = outputLen - strm.avail_out();
+            throw new DataFormatException(CTypeConversion.toJavaString(strm.msg()));
+        } else if (ret == ZLib.Z_MEM_ERROR()) {
+            throw new OutOfMemoryError();
+        } else if (ret != ZLib.Z_BUF_ERROR()) {
+            throw new InternalError(CTypeConversion.toJavaString(strm.msg()));
```
Peter-B-Kessler: The C code is
```
    case Z_BUF_ERROR:
        break;
```
and then a return of OR of the status bits.  Rather than throwing an `InternalError`.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -383,75 +504,145 @@ private int inflateBytes(long addr, byte[] b, int off, int len) throws DataForma
...
 
-    // Inflater.c-Java_java_util_zip_Inflater_reset(JNIEnv *env, jclass cls, jlong addr)
+    // (JDK 11)src/java.base/share/native/libzip/Inflater.c-Java_java_util_zip_Inflater_reset(JNIEnv *env, jclass cls, jlong addr)
     @Substitute
     private static void reset(long addr) {
-        if (ZLib.inflateReset(WordFactory.pointer(addr)) != ZLib.Z_OK()) {
-            throw new InternalError();
+        z_stream strm = WordFactory.pointer(addr);
+        if (ZLib.inflateReset(strm) != ZLib.Z_OK()) {
+            throw new InternalError(CTypeConversion.toJavaString(strm.msg()));
```
Peter-B-Kessler: The C code does not try to get `stem.msg()` if things are not okay.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -383,75 +504,145 @@ private int inflateBytes(long addr, byte[] b, int off, int len) throws DataForma
...
     }
 
-    // Inflater.c-Java_java_util_zip_Inflater_end(JNIEnv *env, jclass cls, jlong addr)
+    // (JDK 11)src/java.base/share/native/libzip/Inflater.c-Java_java_util_zip_Inflater_end(JNIEnv *env, jclass cls, jlong addr)
     @Substitute
     private static void end(long addr) {
         z_stream strm = WordFactory.pointer(addr);
         if (ZLib.inflateEnd(strm) == ZLib.Z_STREAM_ERROR()) {
-            throw new InternalError();
+            throw new InternalError(CTypeConversion.toJavaString(strm.msg()));
```
Peter-B-Kessler: The C code does not try to get `strm.msg()` if there is an error.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -461,6 +652,12 @@ static int update(Object obj, int len, z_stream strm) {
 
 }
 
-/** Dummy class to have a class with the file's name. */
+/**
+ * Dummy class to have a class with the file's name.  This class contains substitutions for {@code java.util.zip}'s
+ * {@code Inflater} and {@code Deflater} classes.  Because both Java 8 and 11 are supported, and these implementations
+ * have differing API, the substitutions contain a mix of things which are shared and which are unique to a specific
+ * version.  Where possible, the origin of the ported code is noted before each substituted method.  The OpenJDK 11
+ * version of these methods are preferred whenever possible.
```
Peter-B-Kessler: All but the first line of this comment belongs on the declaration of `Target_java_util_zip_Inflater`.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
Peter-B-Kessler: There is also the use of the comments for code archeologists: our implementation of the method was based on JDK-8.  The previous comments were silent on that because we never thought we would make it past JDK-8.  (And we still haven't \:-)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
dmlloyd: But this is ``!=``, not ``==``.

dmlloyd:
All fixed.

dmlloyd:
And hit the wrong button ;)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
Peter-B-Kessler: You are correct.

Peter-B-Kessler:
Did you do a "force push" (around 1530 PDT) while I was reviewing this pull request?  Suddenly I could not add comments and had to "show outdated" on all the comments I had made.  This is my first github pull request review so I am just learning how the UI works, but I was surprised.

dmlloyd:
Yeah I did a couple now with fixes.

dmlloyd:
I also inadvertently closed the PR and reopened it.  That might have caused a problem also.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -187,89 +209,158 @@ private int deflateBytes(long addr, byte[] b, int off, int len, int flush) {
...
                 }
             }
         }
     }
 
+    // (JDK 11)src/java.base/share/native/libzip/Deflater.c-static jlong doDeflate(JNIEnv *env, jobject this, jlong addr,
+    //                       jbyte *input, jint inputLen,
+    //                       jbyte *output, jint outputLen,
+    //                       jint flush, jint params)
+    private static long doDeflate(long addr,
```
Peter-B-Kessler: (Now we are getting into things that are checked by the build process.)
This method needs an `@Substitute` annotation.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -146,34 +153,49 @@ private static long init(int level, int strategy, boolean nowrap) {
...
-                    throw new InternalError();
+                    throw new InternalError(CTypeConversion.toJavaString(strm.msg()));
                 }
             }
         }
     }
 
-    // Deflater.c-Java_java_util_zip_Deflater_setDictionary(JNIEnv *env, jclass cls, jlong addr,
+    // (JDK 11)src/java.base/share/native/libzip/Deflater.c-static void doSetDictionary(JNIEnv *env, jlong addr, jbyte *buf, jint len)
+    private static void doSetDictionary(long addr, CCharPointer buf, int len) {
```
Peter-B-Kessler: This method needs an `@Substitute` annotation.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -330,32 +435,48 @@ private static long init(boolean nowrap) {
...
-                    throw new InternalError();
+                    throw new InternalError(CTypeConversion.toJavaString(strm.msg()));
                 }
             }
         }
     }
 
-    // Inflater.c-Java_java_util_zip_Inflater_setDictionary(JNIEnv *env, jclass cls, jlong addr,
+    // (JDK 11)src/java.base/share/native/libzip/Inflater.c-static void doSetDictionary(JNIEnv *env, jlong addr, jbyte *buf, jint len)
+    private static void doSetDictionary(long addr, CCharPointer bytes, int len) {
```
Peter-B-Kessler: This method needs an `@Substitute` annotation.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
Peter-B-Kessler: See my note on `Inflater.doSetDictionary`.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
Peter-B-Kessler: My bad: Believing the error message from the build.  Since the original of this method is a C file-static method, this method needs to move to being a static method in a `Util_java_util_zip_Inflater` class.

Similarly for `Deflater.doSetDictionary`.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -383,75 +504,146 @@ private int inflateBytes(long addr, byte[] b, int off, int len) throws DataForma
...
-                throw new InternalError();
+                throw new InternalError(CTypeConversion.toJavaString(strm.msg()));
+            }
+        }
+    }
+
+    // (JDK 11)src/java.base/share/native/libzip/Inflater.c-static jlong doInflate(JNIEnv *env, jobject this, jlong addr,
+    //                       jbyte *input, jint inputLen,
+    //                       jbyte *output, jint outputLen)
+    private long doInflate(long addr, CCharPointer input, int inputLen, CCharPointer output, int outputLen) throws DataFormatException {
```
Peter-B-Kessler: This method needs to move to `Util_java_util_zip_Inflate`, since it is not a method in the target class for which you are writing substitutions.

(Did you try building with both JDK-8 and JDK-11?  You should have gotten errors from the build on JDK-8.  I haven't tried the JDK-11 build yet.)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
Peter-B-Kessler: Eclipse compiling the JDK-8 view of this code shows me 9 warnings.  I have not yet tried Eclipse with the JDK-11 view of this code.  Do I have to copy them to comments on this pull request, or can you reproduce (and fix \:-) them on your side?  The JDK-8 warnings are for field hidings, methods that could be declared static, unused parameters, and unnecessary casts.  Probably all those warnings should be suppressed unless you want to rename the names of locals from translated C functions so as not to hide the fields of the Java classes.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
```diff
@@ -330,32 +435,48 @@ private static long init(boolean nowrap) {
...
         }
     }
 
-    // Inflater.c-Java_java_util_zip_Inflater_inflateBytes(JNIEnv *env, jobject this, jlong addr,
+    // (JDK 11)src/java.base/share/native/libzip/Inflater.c-Java_java_util_zip_Inflater_setDictionaryBuffer(JNIEnv *env, jclass cls, jlong addr,
+    //                                          jlong bufferAddr, jint len)
+    @Substitute
+    @TargetElement(onlyWith = JDK11OrLater.class)
+    private static void setDictionaryBuffer(long addr, long bufAddress, int len) {
+        doSetDictionary(addr, WordFactory.pointer(addr), len);
```
Peter-B-Kessler: The C code is
```
    jbyte *buf = jlong_to_ptr(bufferAddr);
    doSetDictionary(env, addr, buf, len);
```
but you seem not to be using `bufAddress`.  Should this line be
```
    doSetDictionary(addr, WordFactory(bufAddress), len)
```
?  (I admit to not knowing enough about the zip API to know.)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
dmlloyd: I don't use Eclipse.  And, I'm not really interested in conforming to every potential reviewer's personal IDE settings in any event.  The main build seems to proceed with no warnings and no checkstyle problems, and my IDE (IntelliJ) is happy with the code.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
Peter-B-Kessler: Do you want me to make these re-factorings to get the code to build for JDK-8?  Or would you like that challenge?

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaUtilZipSubstitutions.java
dmlloyd: I don't see why, but OK.

dmlloyd:
All fixed.

Peter-B-Kessler:
Okay. I will take it from here.  Thanks for your contribution.

Issue #563, opened 2018-07-23
Automatically register classes in META-INF/services for reflection.


cstancu:
Currently all service providers declared in `META-INF/services` resources need to be manually registered for [reflection](https://github.com/oracle/graal/blob/master/substratevm/LIMITATIONS.md#reflection). To improve usability this should be automated.

Registration for reflection is needed because the service providers are loaded with `Class.forName()` and are instantiated with `Class.newInstance()`. A typical reflection configuration for service providers looks like this:

```
[
  {
    "name" : "service.ServiceImplementation0",
    "methods": [
      { "name": "<init>", "parameterTypes": [] }
    ]
]
```
,i.e., it specifies the class and the no-parameter constructor. A complete example of registering service providers can be found in this [demo project](https://github.com/cstancu/native-image-service-loader-demo).

The automatic registration can be achieved by parsing `META-INF/services` at native image build time and using the SubstrateVM API to register the classes.

Automatic registration should be limited to the `META-INF/services` resources explicitely registered using the [`-H:IncludeResources=<resource-path-regexp>`](https://github.com/oracle/graal/blob/master/substratevm/RESOURCES.md#accessing-resources-in-substrate-vm-images) option to avoid polution of the image with all resources on the classpath.

SubstrateVM provides the [`Feature`](https://github.com/oracle/graal/blob/master/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Feature.java) mechanism to extend its internals with plugable components. For reflection registration the [`RuntimeReflection`](https://github.com/oracle/graal/blob/master/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/RuntimeReflection.java) API can be invoked from a `Feature`. The [reflection documentation](https://github.com/oracle/graal/blob/master/substratevm/REFLECTION.md) provides details about the reflection API. An example of registering elements for reflection from a `Feature` can be found in [`GsonFeature`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.thirdparty/src/com/oracle/svm/thirdparty/gson/GsonFeature.java). 

(Make sure to read the [contributing guidelines](https://github.com/oracle/graal/blob/master/CONTRIBUTING.md) if you would like to take on this issue.)

dmlloyd:
[...]

PR #566, opened 2018-07-25
Automatically register classes used with ServiceLoader

vhiairrassary:
Here is my attempt to fix https://github.com/oracle/graal/issues/563. The PR is still a WIP as it does not contain any test and as it relies on hacks. Would love to get feedbacks on it.

Here are my main concerns:
1. Is it a good idea to force the user to use something like the `-H:IncludeResources=META-INF/services/service.ServiceBase` to find the concrete classes? I am not sure we want to include these files in the final image. We can just parse all `META-INF/*/*` files for example.
2. If you want to use the above option, how can I access the resources from the feature? Currently I rely on a public variable in `Resources.java`.
3. Should we abort if no concrete class is found (currently a log is emitted is the tracing is enabled)?
4. Is there a better way than my `ServiceRegistry.java` to share data between the feature and the plugin?
5. Is the feature in the correct package?

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
@@ -24,19 +24,24 @@
  */
 package com.oracle.svm.hosted.snippets;
 
+import java.io.InputStream;
 import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Proxy;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
+import java.util.ServiceLoader;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.stream.Stream;
 
+import com.oracle.svm.core.jdk.Resources;
+import com.oracle.svm.hosted.ImageClassLoader;
 import org.graalvm.compiler.api.replacements.SnippetReflectionProvider;
 import org.graalvm.compiler.bytecode.BytecodeProvider;
 import org.graalvm.compiler.core.common.CompressEncoding;
@@ -142,17 +147,22 @@ import sun.misc.Unsafe;
 class Options {
     @Option(help = "Enable trace logging for dynamic proxy.")//
     static final HostedOptionKey<Boolean> DynamicProxyTracing = new HostedOptionKey<>(false);
+
+    @Option(help = "Enable trace logging for service loader.")//
+    static final HostedOptionKey<Boolean> ServiceLoaderTracing = new HostedOptionKey<>(false);
 }
 
 public class SubstrateGraphBuilderPlugins {
     public static void registerInvocationPlugins(MetaAccessProvider metaAccess, ConstantReflectionProvider constantReflection,
-                    SnippetReflectionProvider snippetReflection, InvocationPlugins plugins, BytecodeProvider bytecodeProvider, boolean analysis) {
+                                                 SnippetReflectionProvider snippetReflection, InvocationPlugins plugins, BytecodeProvider bytecodeProvider,
+                                                 ImageClassLoader loader, boolean analysis) {
 
         // register the substratevm plugins
         registerSystemPlugins(metaAccess, plugins);
         registerImageInfoPlugins(metaAccess, plugins);
         registerProxyPlugins(snippetReflection, plugins, analysis);
         registerAtomicUpdaterPlugins(metaAccess, snippetReflection, plugins, analysis);
+        registerServiceLoaderPlugins(snippetReflection, plugins, loader, analysis);
         registerObjectPlugins(plugins);
         registerUnsafePlugins(plugins);
         registerKnownIntrinsicsPlugins(plugins, analysis);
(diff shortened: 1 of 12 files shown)


vhiairrassary:
@cstancu Thank you for your review and these precious comments.

1. I have changed my code to not use a feature anymore. Now it automatically finds & registers the resource files if they are not already registered when we determine statically the concrete value of the argument to `ServiceLoader.load(Class)`.

2. Do you want me to address the case of a non-constant argument? If this PR covers most of the use cases then this can be done later (this PR will also be simpler).

3. Once you have reviewed the new code then I will:
    - Write the tests;
    - Write the documentation;
    - Squash the commits & prepare the PR for a final review.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
```diff
@@ -399,6 +409,65 @@ private static void interceptUpdaterInvoke(MetaAccessProvider metaAccess, Snippe
...
+
+                            byte[] resource = resources.get(0);
+
+                            String[] lines = new String(resource, StandardCharsets.UTF_8).split("\\r?\\n");
+
+                            for (String line : lines) {
+                                String concreteClassName = line.replaceAll("[ \t]|#.*$", "");
+
+                                if (!concreteClassName.isEmpty()) {
+                                    Class<?> concreteClazz = loader.findClassByName(concreteClassName, false);
```
cstancu: `ImageClassLoader.findClassByName(className, false)` can return `null`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
```diff
@@ -399,6 +409,65 @@ private static void interceptUpdaterInvoke(MetaAccessProvider metaAccess, Snippe
         }
     }
 
+    private static void registerServiceLoaderPlugins(SnippetReflectionProvider snippetReflection, InvocationPlugins plugins, ImageClassLoader loader, boolean analysis) {
+        Registration serviceLoaderRegistration = new Registration(plugins, ServiceLoader.class);
+        serviceLoaderRegistration.register1("load", Class.class, new InvocationPlugin() {
```
cstancu: You also need to register a plugin for `public static <S> ServiceLoader<S> load(Class<S> service, ClassLoader loader)`. You can ignore the `loader` argument, just like we to for [`ReflectionPlugins`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java#L118) since the `ImageClassLoader` is used to load the resources anyway.

cstancu:
1. It looks good, this is the approach that I had in mind. (See also my embeded comments.)

2. I agree, we can leave the `non-constant argument` case for later. We have a solution that works for now, i.e., registering the serviceproviders for reflection manually.

3. You can add unit tests in the `com.oracle.svm.test` project. Then you can run them with `mx native-unittest YourTest`. As an example you can loook at [`ImageInfoTest`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/ImageInfoTest.java) which you can run with `mx native-unittest ImageInfoTest`. These tests are not added to the gate yet but they will be.

vhiairrassary:
Amesome! Should be done by sunday.

vhiairrassary:
@cstancu I have added tests to validate `load(Class<S>)` & `load(Class<S>, ClassLoader)` but few times it failed with:
```
Building junit image for matching: com.oracle.svm.test.ServiceLoaderTest
   classlist:   1,140.46 ms
       (cap):     905.39 ms
       setup:   2,112.10 ms
    analysis:   5,581.91 ms
error: Error encountered while parsing com.oracle.svm.test.ServiceLoaderTest.testServiceLoaderLoad2()
Parsing context:
	parsing com.oracle.svm.reflect.proxies.Proxy_2_ServiceLoaderTest_testServiceLoaderLoad2.invoke(Unknown Source)
	parsing java.lang.reflect.Method.invoke(Method.java:498)
        [...]
	parsing com.oracle.svm.core.JavaMainWrapper.run(JavaMainWrapper.java:163)
	parsing com.oracle.svm.core.code.CEntryPointCallStubs.com_002eoracle_002esvm_002ecore_002eJavaMainWrapper_002erun_0028int_002corg_002egraalvm_002enativeimage_002ec_002etype_002eCCharPointerPointer_0029(generated:0)
Original error: com.oracle.svm.core.util.UserError$UserException: ImageSingletons.add must not overwrite existing key com.oracle.svm.core.jdk.Resources$ResourcesSupport
Existing value: com.oracle.svm.core.jdk.Resources$ResourcesSupport@5983e751
New value: com.oracle.svm.core.jdk.Resources$ResourcesSupport@f54984c
	at com.oracle.svm.core.util.UserError.abort(UserError.java:67)
	at com.oracle.svm.hosted.ImageSingletonsSupportImpl$HostedManagement.doAdd(ImageSingletonsSupportImpl.java:102)
	at com.oracle.svm.hosted.ImageSingletonsSupportImpl.add(ImageSingletonsSupportImpl.java:39)
	at org.graalvm.nativeimage.ImageSingletons.add(ImageSingletons.java:56)
	at com.oracle.svm.core.jdk.Resources.registerResource(Resources.java:69)
	at com.oracle.svm.hosted.snippets.SubstrateGraphBuilderPlugins.interceptServiceLoaderInvoke(SubstrateGraphBuilderPlugins.java:453)
	at com.oracle.svm.hosted.snippets.SubstrateGraphBuilderPlugins.access$200(SubstrateGraphBuilderPlugins.java:155)
	at com.oracle.svm.hosted.snippets.SubstrateGraphBuilderPlugins$11.apply(SubstrateGraphBuilderPlugins.java:427)
```

Is it possible that `com.oracle.svm.core.jdk.Resources.registerResource(String, InputStream)` is not thread-safe AND that invocation plugins are applied in parallel?

Code for registering a resource is:
```java
@Platforms(Platform.HOSTED_ONLY.class)
public static void registerResource(String name, InputStream is) {
    ResourcesSupport support;
    if (ImageSingletons.contains(ResourcesSupport.class)) {
        support = ImageSingletons.lookup(ResourcesSupport.class);
    } else {
        support = new ResourcesSupport();
        ImageSingletons.add(ResourcesSupport.class, support);
    }
    // ...
}
```

jhult:
Any news on merging this?

graemerocher:
Yeah it would be a really great if this could be merged, it would eliminate a whole bunch of problems and work that has to be done to build the reflect JSON file

cstancu:
https://github.com/oracle/graal/pull/692, that was just merged, supersedes the current PR. Thinking more about this we have arrived at a different, more flexible approach: [we add all service loader files and service implementation classes of the service interfaces that are seen as reachable by the static analysis](https://github.com/oracle/graal/commit/84dcb4e9ad49ae749c88ee612ba5dc95ab171bd8#diff-94f52964cecab0a25ac57a96aa80b97aR55).

@vhiairrassary thank you for your work!

PR #568, opened 2018-07-25
[#557] Add system property 'os.version' to the list of eager substitutions

### LINKED ISSUE #557: Support "os.version" system property in SubstrateVM
Like the already supported "os.name" system property, "os.version" is often used in informational and diagnostic messages, which makes it worthwhile to support.

olegchir:
Fixes #557.

Soft API change: new system property

oca-signed: Oleg Chirukhin
(http://www.oracle.com/technetwork/community/oca-486395.html)


cstancu:
I think `os.version` should be a runtime computed value. By adding it to `HOSTED_PROPERTIES` you compute it at image build time, i.e., it is the `os.version` of the OS on which the image was built. While we can get away with it for `os.arch` and `os.name` for now since we don't have cross-compilation support the same doesn't apply for `os.version`.

olegchir:
@cstancu ok, I'll investigate it further, thank you

olpaw:
@olegchir I have a full implementation of `os.version` almost done that gets you the correct **runtime value** of `os.version` for linux and darwin. Stay tuned. It will land on master today or tomorrow.

PR #512, opened 2018-07-03
Fix JSON snippet in documentation

vhiairrassary:
The JSON snippet in `substratevm/REFLECTION.md` is not valid, let's fix it

PR #585, opened 2018-08-02
Add missing comma

joprice:
Fixes an issue in the docs that causes this error when run:

`line 6 column 4 Expected , or } but found "`


vhiairrassary:
@joprice Already fixed in https://github.com/oracle/graal/pull/512 which contain other fixes for the same snippet

joprice:
Oh great! Closing

PR #601, opened 2018-08-07
fix some warnings found by error prone

iignatev:
experimenting w/ using [error-prone](http://errorprone.info) for `jdk/jdk` repo, I got a number of reports for `jdk.compiler` module. this PR fixes some (most straightforward and less invasive) of them.


dougxc:
Thanks for the PR @iignatev !

Issue #596, opened 2018-08-06
DatagramSocket Unsupported


bobmcwhirter:
Trying to use Jaeger with UDP instead of HTTP and tripping over the lack of support for DatagramSockets.

Happy to try to contribute support, if someone could point me in the general direction.

bobmcwhirter:
com.oracle.svm.core.jdk.UnsupportedFeatureError: Unimplemented: java.net.DatagramSocket.<init>(SocketAddress)

And the JavaNetSubsitutions seem to explicit throw that.

Peter-B-Kessler:
I am pretty sure the problem is that we do not (yet) have translations of the native methods in `sun.nio.ch.DatagramChannelImpl`.  E.g., http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/6bfaecb8ff77/src/solaris/native/sun/nio/ch/DatagramChannelImpl.c.  Let me see if I can carve out some time for it.

PR #604, opened 2018-08-09
Hopefully actually pretty good attempt at getting Datagram sockets to work.

bobmcwhirter:
Relates to https://github.com/oracle/graal/issues/596

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaNetNetUtil.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaNetNetUtil.java
@@ -223,6 +223,111 @@ class JavaNetNetUtil {
         return iaObj;
     }
 
+    // 255 JNIEXPORT jint JNICALL
+    // 256 NET_SockaddrEqualsInetAddress(JNIEnv *env, struct sockaddr *him, jobject iaObj)
+    static boolean NET_SockaddrEqualsInetAddress(Socket.sockaddr him, InetAddress iaObj) {
+        // 258 jint family = AF_INET;
+        int family = Socket.AF_INET();
+
+        /* Restructured due to #ifdef mixing with either an else-branch or just a curlied block */
+
+        // 260 #ifdef AF_INET6
+        if (IsDefined.socket_AF_INET6()) {
+            // 261 family = getInetAddress_family(env, iaObj) == IPv4? AF_INET : AF_INET6;
+            family = getInetAddress_family(iaObj) == Target_java_net_InetAddress.IPv4 ? Socket.AF_INET() : Socket.AF_INET6();
+            // 262 JNU_CHECK_EXCEPTION_RETURN(env, JNI_FALSE);
+            // 263 if (him->sa_family == AF_INET6) {
+        }
+
+        /* This only occurs if AF_INET6 above && him is IPv6 */
+        if (him.sa_family() == Socket.AF_INET6()) {
+            // 264 #ifdef WIN32
+            // 265 struct SOCKADDR_IN6 *him6 = (struct SOCKADDR_IN6 *)him;
+            // 266 #else
+            // 267 struct sockaddr_in6 *him6 = (struct sockaddr_in6 *)him;
+            NetinetIn.sockaddr_in6 him6 = (NetinetIn.sockaddr_in6) him;
+            // 268 #endif
+            // 269 jbyte *caddrNew = (jbyte *)&(him6->sin6_addr);
+            CCharPointer caddrNew = him6.sin6_addr().s6_addr();
+            // 270 if (NET_IsIPv4Mapped(caddrNew)) {
+            if (isIPv4Mapped(caddrNew)) {
+                // 271 int addrNew;
+                int addrNew;
+                // 272 int addrCur;
+                int addrCur;
+                // 273 if (family == AF_INET6) {
+                if (family == Socket.AF_INET6()) {
+                    // 274 return JNI_FALSE;
+                    return false;
+                }
+                // 276 addrNew = NET_IPv4MappedToIPv4(caddrNew);
+                addrNew = JavaNetNetUtilMD.NET_IPv4MappedToIPv4(caddrNew);
+                // 277 addrCur = getInetAddress_addr(env, iaObj);
+                addrCur = JavaNetNetUtilMD.getInetAddress_addr(iaObj);
+                // 278 JNU_CHECK_EXCEPTION_RETURN(env, JNI_FALSE);
+                // 279 if (addrNew == addrCur) {
+                if (addrNew == addrCur) {
+                    // 280 return JNI_TRUE;
+                    return true;
+                } else {
+                    // 282 return JNI_FALSE;
+                    return false;
+                }
+                // 284 } else {
+            } else {
+                // 285 jbyteArray ipaddress;
+                /* Unused. */
+                // 286 jbyte caddrCur[16];
+                CCharPointer caddrCur = StackValue.get(16, CCharPointer.class);
+                // 287 int scope;
+                int scope;
+                // 289 if (family == AF_INET) {
+                if (family == Socket.AF_INET()) {
+                    // 290 return JNI_FALSE;
+                    return false;
+                }
+                // 292 scope = getInet6Address_scopeid(env, iaObj);
+                scope = getInet6Address_scopeid((Inet6Address) iaObj);
+                // 293 getInet6Address_ipaddress(env, iaObj, (char *)caddrCur);
+                getInet6Address_ipAddress((Inet6Address) iaObj, caddrCur);
+                // 294 if (NET_IsEqual(caddrNew, caddrCur) && cmpScopeID(scope, him)) {
+                if (JavaNetNetUtilMD.NET_IsEqual(caddrCur, caddrCur) && JavaNetNetUtilMD.cmpScopeID(scope, him)) {
+                    // 295 return JNI_TRUE;
+                    return true;
+                } else {
+                    // 297 return JNI_FALSE;
+                    return false;
+                }
+            }
+            // 301 #endif /* AF_INET6 */
+        }
+
+        // 303 struct sockaddr_in *him4 = (struct sockaddr_in *)him;
+        NetinetIn.sockaddr_in him4 = (NetinetIn.sockaddr_in) him;
+        // 304 int addrNew, addrCur;
+        int addrNew, addrCur;
+
+        // 305 if (family != AF_INET) {
+        if (family != Socket.AF_INET()) {
+            // 306 return JNI_FALSE;
+            return false;
+        }
+
+        // 308 addrNew = ntohl(him4->sin_addr.s_addr);
+        addrNew = NetinetIn.ntohl(him4.sin_addr().s_addr());
+        // 309 addrCur = getInetAddress_addr(env, iaObj);
+        addrCur = JavaNetNetUtilMD.getInetAddress_addr(iaObj);
+        // 310 JNU_CHECK_EXCEPTION_RETURN(env, JNI_FALSE);
+        // 311 if (addrNew == addrCur) {
+        if (addrNew == addrCur) {
+            // 312 return JNI_TRUE;
+            return true;
+        } else {
+            // 314 return JNI_FALSE;
+            return false;
+        }
+    }
+
     static boolean isIPv4Mapped(CCharPointer caddr) {
         int i;
         for (i = 0; i < 10; i++) {
@@ -413,6 +518,13 @@ class JavaNetNetUtil {
         // 216 return (*env)->GetIntField(env, holder, iac_familyID);
         return holder.family;
     }
+
+    @Fold
+    static int MAX_PACKET_LEN() {
+        // from {jdk8}/share/native/java/net/net_util.h
+        // 37 #define MAX_PACKET_LEN 65536
+        return 65536;
+    }
 }
 
 /** Native methods from jdk/src/solaris/native/java/net/net_util_md.c translated to Java. */
@@ -462,6 +574,31 @@ class JavaNetNetUtilMD {
         return ((caddr.read(12) & 0xff) << 24) | ((caddr.read(13) & 0xff) << 16) | ((caddr.read(14) & 0xff) << 8) | (caddr.read(15) & 0xff);
     }
     /* @formatter:on */
+
+    // 963 int
+    // 964 NET_IsEqual(jbyte* caddr1, jbyte* caddr2) {
+    static boolean NET_IsEqual(CCharPointer caddr1, CCharPointer caddr2) {
+        // 965 int i;
+        int i;
+        // 966 for (i = 0; i < 16; i++) {
+        for (i = 0; i < 16; i++) {
+            // 967 if (caddr1[i] != caddr2[i]) {
+            if (caddr1.read(i) != caddr2.read(i)) {
+                // 968 return 0; /* false */
+                return false;
+            }
+        }
+        // 971 return 1;
+        return true;
+    }
+
+    // 251 int cmpScopeID (unsigned int scope, struct sockaddr *him) {
+    static boolean cmpScopeID(int scope, Socket.sockaddr him) {
+        // 252 struct sockaddr_in6 *him6 = (struct sockaddr_in6 *)him;
+        NetinetIn.sockaddr_in6 him6 = (NetinetIn.sockaddr_in6) him;
+        // 253 return him6->sin6_scope_id == scope;
+        return him6.sin6_scope_id() == scope;
+    }
 
     // 237 int getScopeID (struct sockaddr *him) {
     static int getScopeID(Socket.sockaddr him) {
@@ -1211,6 +1348,14 @@ class JavaNetNetUtilMD {
         return VmPrimsJVM.JVM_Send(fd, buf, nBytes, flags);
     }
     /* @formatter:on */
+
+    static int NET_SendTo(int fd, CCharPointer buf, int n, int flags, Socket.sockaddr addr, int addr_len) {
+        return VmPrimsJVM.JVM_SendTo(fd, buf, n, flags, addr, addr_len);
+    }
+
+    static int NET_RecvFrom(int fd, CCharPointer buf, int n, int flags, Socket.sockaddr addr, CIntPointer addr_len) {
+        return VmPrimsJVM.JVM_RecvFrom(fd, buf, n, flags, addr, addr_len);
+    }
 
     /* Do not re-wrap commented-out code.  @formatter:off */
     // 926 JNIEXPORT jint JNICALL
@@ -2060,6 +2205,14 @@ class VmPrimsJVM {
         // 3750 JVM_END
     }
 
+    static int JVM_SendTo(int fd, CCharPointer buf, int n, int flags, Socket.sockaddr addr, int addr_len) {
+        return Target_os.sendto(fd, buf, n, flags, addr, addr_len);
+    }
+
+    static int JVM_RecvFrom(int fd, CCharPointer buf, int n, int flags, Socket.sockaddr addr, CIntPointer addr_len) {
+        return Target_os.recvfrom(fd, buf, n, flags, addr, addr_len);
+    }
+
     // 3725 JVM_LEAF(jint, JVM_SocketClose(jint fd))
     static int JVM_SocketClose(int fd) {
         // 3726 JVMWrapper2("JVM_SocketClose (0x%x)", fd);
@@ -2302,6 +2455,30 @@ class Target_os {
             int _result;
             do {
                 _result = (int) Socket.send(fd, buf, WordFactory.unsigned(nBytes), flags).rawValue();
+            } while ((_result == VmRuntimeOS.OSReturn.OS_ERR()) || (Errno.errno() == Errno.EINTR()));
+            return _result;
+        } while (false);
+    }
+
+    // 248 inline int os::sendto(int fd, char* buf, size_t len, uint flags, struct sockaddr *to, socklen_t tolen) {
+    static int sendto(int fd, CCharPointer buf, int n, int flags, Socket.sockaddr addr, int addr_len) {
+        // 250   RESTARTABLE_RETURN_INT((int)::sendto(fd, buf, len, flags, to, tolen));
+        do {
+            int _result;
+            do {
+                _result = (int) Socket.sendto(fd, buf, WordFactory.unsigned(n), flags, addr, addr_len).rawValue();
+            } while ((_result == VmRuntimeOS.OSReturn.OS_ERR()) || (Errno.errno() == Errno.EINTR()));
+            return _result;
+        } while (false);
+    }
+
+    // 243 inline int os::recvfrom(int fd, char* buf, size_t nBytes, uint flags, sockaddr* from, socklen_t* fromlen) {
+    static int recvfrom(int fd, CCharPointer buf, int n, int flags, Socket.sockaddr addr, CIntPointer addr_len) {
+        // 245   RESTARTABLE_RETURN_INT((int)::recvfrom(fd, buf, nBytes, flags, from, fromlen));
+        do {
+            int _result;
+            do {
+                _result = (int) Socket.recvfrom(fd, buf, WordFactory.unsigned(n), flags, addr, addr_len).rawValue();
             } while ((_result == VmRuntimeOS.OSReturn.OS_ERR()) || (Errno.errno() == Errno.EINTR()));
             return _result;
         } while (false);
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
@@ -36,10 +36,10 @@ import java.net.InetAddress;
 import java.net.InterfaceAddress;
 import java.net.NetworkInterface;
 import java.net.NoRouteToHostException;
+import java.net.PortUnreachableException;
 import java.net.ProtocolException;
 import java.net.Proxy;
 import java.net.ServerSocket;
-import java.net.SocketAddress;
 import java.net.SocketException;
 import java.net.SocketImpl;
 import java.net.SocketOptions;
@@ -48,6 +48,7 @@ import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 
+import com.oracle.svm.core.posix.headers.Ioctl;
 import org.graalvm.compiler.serviceprovider.GraalServices;
 import org.graalvm.nativeimage.PinnedObject;
 import org.graalvm.nativeimage.Platform;
@@ -59,6 +60,7 @@ import org.graalvm.nativeimage.c.type.CIntPointer;
 import org.graalvm.nativeimage.c.type.CTypeConversion;
 import org.graalvm.nativeimage.c.type.CTypeConversion.CCharPointerHolder;
 import org.graalvm.nativeimage.c.type.WordPointer;
+import org.graalvm.word.PointerBase;
 import org.graalvm.word.WordFactory;
 
 import com.oracle.svm.core.annotate.Alias;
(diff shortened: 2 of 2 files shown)


bobmcwhirter:
I'm happy to sign the OCA personally, but I'm also doing this as sanctioned work at Red Hat, who is already a signatory.

Peter-B-Kessler:
Red Hat is a signatory of the OCA.  JBoss is a division of Red Hat, so I think you are covered.

Peter-B-Kessler:
https://github.com/oracle/graal/pull/604#issue-207421675: Re: Line numbers: I have a code viewer that allows me to copy the "Line number" column.  But I guess one could use `cat -n` on the source file.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
 @TargetClass(className = "java.net.PlainDatagramSocketImpl")
 @Platforms({Platform.LINUX.class, Platform.DARWIN.class})
 final class Target_java_net_PlainDatagramSocketImpl {
+    @Alias boolean connected;
 
     @Substitute
```
Peter-B-Kessler: It would be good to tell the reader where this code comes from.  Something like
```
/* Translated from: {jdk8u}/src/solaris/native/java/net/PlainDatagramSocketImpl.c */
// JNIEXPORT void JNICALL
// Java_java_net_PlainDatagramSocketImpl_bind0(JNIEnv *env, jobject this,
//                                            jint localport, jobject iaObj) {
```
If that is in fact where it comes from.  That at least gets me to the correct file and method.  Then I can start following lines even if they do not have line numbers.  Extra credit for line numbers.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
     @SuppressWarnings({"static-method", "unused"})
     protected void bind0(int lport, InetAddress laddr) throws SocketException {
-        throw VMError.unsupportedFeature("Unimplemented: java.net.PlainDatagramSocketImpl.bind0(int, InetAddress)");
+        //        jobject fdObj = (*env)->GetObjectField(env, this, pdsi_fdID);
```
Peter-B-Kessler: You are starting to diverge from the C code.  That makes me have to think!  In code like this I would rather not have to think.  You have the code at line 109, just move it up here and I do not have to wonder what happened to it.  Or wonder how it got to line 109 when I get there.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        //        int fd;
+        //        int len = 0;
+        CIntPointer len_Pointer = StackValue.get(CIntPointer.class);
```
Peter-B-Kessler: You forgot the initialization of `len = 0`.  (That's a common mistake!)  You need
```
len_Pointer.write(0);
```

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        FileDescriptor fdObj = Util_java_net_DatagramSocketImpl.as_Target_java_net_DatagramSocketImpl(this).fd;
+        //        if (IS_NULL(fdObj)) {
+        if (fdObj == null) {
+            throw new SocketException("Socket closed");
+        }
+        //            JNU_ThrowByName(env, JNU_JAVANETPKG "SocketException",
+        //                            "Socket closed");
+        //            return;
+        //        } else {
+        int fd = PosixUtils.getFD(fdObj);
```
Peter-B-Kessler: Tempting as it is to simplify the control flow (changing `if () { throw } else ....` to `if () { throw }; ....` that is one more thing I have to think about when reviewing the code.

To be fair, I usually reuse `} else {` and `}` lines from the original C code.  But I try not to change the control flow unless I have to.  Among the reasons for changing the control flow: In C code one can "throw" and exception and then continue execution, knowing that the exception will not be noticed until one crosses back from C to Java.  In SubstrateVM it is all Java code, so if you throw and exception, execute terminate abruptly right there.  That's a rare case.  Sometimes I rewrite the code, sometimes I wrap `try ... finally ...` blocks around things.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        //            fd = (*env)->GetIntField(env, fdObj, IO_fd_fdID);
+        //        }
+        //
+        //        if (IS_NULL(iaObj)) {
+        if (laddr == null) {
```
Peter-B-Kessler: Tempting as it is to change the names of the names of the formal parameters, if you do that I have to check the correspondence between the `iaObj` in the C code and the `laddr` in your Java code.  On the other hand, I do allow myself to change the declaration of `jobject iaObj` to `InetAddress iaObj`, since we are in (typed) Java code, and that's the type `iaObj` has to have for this code to work.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+            throw new NullPointerException("iaObj is null.");
```
Peter-B-Kessler: This line belongs below the corresponding C code, e.g., after 126 or maybe 127.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        }
+        //            JNU_ThrowNullPointerException(env, "iaObj is null.");
+        //            return;
+        //        }
+        //
+        //    /* bind */
+        //        if (NET_InetAddressToSockaddr(env, iaObj, localport, (struct sockaddr *)&him, &len, JNI_TRUE) != 0) {
+        if (JavaNetNetUtilMD.NET_InetAddressToSockaddr(laddr, lport, him, len_Pointer, Util_jni.JNI_TRUE()) != 0) {
```
Peter-B-Kessler: s/laddr/localport/

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
Peter-B-Kessler: Though, in this case, by reading much further, it looks like `len` is written before it is read.  But I don't want to have to do that analysis.  And I don't want to have to debug that if the code evolves to read `len` before it is written.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        //            (*env)->SetIntField(env, this, pdsi_localPortID, localport);
+        //        }
+        if (lport == 0) {
+            if (VmPrimsJVM.JVM_GetSockName(fd, him, len_Pointer) == -1) {
+                throw new SocketException(PosixUtils.lastErrorString("Error getting socket name"));
+            }
+            // 602         localport = NET_GetPortFromSockaddr((struct sockaddr *)&him);
+            lport = JavaNetNetUtilMD.NET_GetPortFromSockaddr(him);
+            // 603         (*env)->SetIntField(env, this, psi_localportID, localport);
+            Util_java_net_PlainDatagramSocketImpl.as_Target_java_net_DatagramSocketImpl(this).localPort = lport;
```
Peter-B-Kessler: Would it be worth having a `Util_java_net_PlainDatagramSocketImpl.setLocalPort(...., int)` to mimic this `SetIntField`, including mimicking the Java class hierarchy where it is missing from the `Target` class hierarchy?  I almost suggested up where you get the `fd` field.  It has to be a static method taking `this` as a parameter, but that just makes the declaration ugly, not so much the use.

(Wow, I hope the C and Java compilers do "common tail" optimizations to pull that assignment out of both the `then` and `else` branches.  Not my problem, this week. :-)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        }
     }
 
     @Substitute
     @SuppressWarnings({"static-method", "unused"})
-    protected void send(DatagramPacket p) throws IOException {
-        throw VMError.unsupportedFeature("Unimplemented: java.net.PlainDatagramSocketImpl.send(DatagramPacket)");
+    protected void send(DatagramPacket packet) throws IOException {
+        //    char BUF[MAX_BUFFER_LEN];
```
Peter-B-Kessler: Argh!  What happened to `BUF`?  Now I have to trace through the control flow to see if the uses are really replaced.

Oh, and what is worse: You have changed what might a (essentially free) stack allocation, probably in the common case, into (not free) calls to `malloc` and `free`.  At least it looks like you remembered to call `free` on it. :-)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        //
+        //        if (IS_NULL(packet)) {
+        if (packet == null) {
+            //            JNU_ThrowNullPointerException(env, "packet");
+            throw new NullPointerException("packet");
+        }
+        //        packetBuffer = (*env)->GetObjectField(env, packet, dp_bufID);
+        //        packetAddress = (*env)->GetObjectField(env, packet, dp_addressID);
+        //        if (IS_NULL(packetBuffer) || IS_NULL(packetAddress)) {
+        if (packet.getData() == null || packet.getAddress() == null) {
```
Peter-B-Kessler: Whose mistake (!) is it that this field is accessed with a method called `getData()`, when `DatagramPacket` does not seem to have a field named `data`?  It has a field called `but`, but that should be accessed by turning `packet` into a `Target_java_net_DatagramPacket` and looking that the `buf` field.  Oh, I see, you are just using the Java access methods.  I guess that's allowed.  Might be worth a comment in this rare case where there are public accessors for the fields.

Hmm, the Java access methods are all `synchronized`.  Note that the C code doesn't obey that protocol.  In particular, if the C code is reading the length and the data while someone else (even in synchronized Java code) is calling `setData` which updates a bunch of the fields, supposedly atomically.  Of course, Java code (or this code) that separately calls `getData` and `getLength` could get mis-matched values, too.  Turn over a rock, find a bug.  Maybe the worst that can happen is that the Java code gets an `ArrayIndexOutOfBoundsException`.  Maybe the C code could get itself into more trouble.

What happens if you call a synchronized access method and block?  Might you have just introduced a deadlock, since no one is expecting this method to block on the field accesses? I *think* not, because the accesses should all complete without any long-lasting contention.  But I am sure I do not want to debug that.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        //    /*
+        //     * Send the datagram.
+        //     *
+        //     * If we are connected it's possible that sendto will return
+        //     * ECONNREFUSED indicating that an ICMP port unreachable has
+        //     * received.
+        //     */
+        //                ret = NET_SendTo(fd, fullPacket, packetBufferLen, 0,
+        //                                 (struct sockaddr *)rmtaddrP, len);
+        //CCharPointer bufP = WordFactory.nullPointer();
```
Peter-B-Kessler: Whose line of code is that!  It is not in the original C.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        VmPrimsJNI.GetByteArrayRegion(packet.getData(), 0, packetBufferLen, fullPacket);
```
Peter-B-Kessler: Wow!  This is from line 292 above.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -94,17 +94,246 @@ private PosixJavaNetSubstitutions() {
...
+        //                    JNU_ThrowByName(env, "java/io/InterruptedIOException",
+        //                                    "operation interrupted");
+        //                    break;
+        //            }
+        //        }
+        //
+        //        if (mallocedPacket) {
+        //            free(fullPacket);
+        //        }
+        //        return;
```
Peter-B-Kessler: What happened to all this error checking?

(Ah, see, here is a case where the C code "throws" and exception and then runs some more code (`free(fullPacket)`).  I would translate that as a `try { .... } finally { .... }`.

Peter-B-Kessler:
Your contribution (https://github.com/oracle/graal/commit/3d9bb2886cbe655326f04f9afdf3d04c0e5f004b.patch) is from <bob@mcwhirter.org>.  I think you need to send the contribution from an "@redhat.com" (or "@jboss.com") address to be covered by Red Hat's signing of the Oracle Contributor Agreement.  Or you could sign as an individual and contribute from your individual account.

bobmcwhirter:
FWIW now committing as a Red Hatter.

bobmcwhirter:
@Peter-B-Kessler Okay, I'm happy (well... only if you're happy) with this PR now. I'll stop `push -f`'ing against this branch until your further review (or merge, because you've had a beer and are feeling generous).

Thanks for your guidance with this. Hopefully it's not terribly terrible.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaNetNetUtil.java
```diff
@@ -223,6 +223,114 @@ static InetAddress NET_SockaddrToInetAddress(Socket.sockaddr him, CIntPointer po
         return iaObj;
     }
 
+    // 255 JNIEXPORT jint JNICALL
+    // 256 NET_SockaddrEqualsInetAddress(JNIEnv *env, struct sockaddr *him, jobject iaObj)
+    static boolean NET_SockaddrEqualsInetAddress(Socket.sockaddr him, InetAddress iaObj) {
+        // 258 jint family = AF_INET;
+        int family = Socket.AF_INET();
+
+        int sa_family = Socket.AF_INET();
```
Peter-B-Kessler: This variable appears to be unused.  Probably you thought you would need it to untangle the mess below.

bobmcwhirter:
Right on. Will remedy.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaNetNetUtil.java
```diff
@@ -223,6 +223,114 @@ static InetAddress NET_SockaddrToInetAddress(Socket.sockaddr him, CIntPointer po
...
+        if (addrNew == addrCur) {
+            // 312 return JNI_TRUE;
+            return true;
+        } else {
+            // 314 return JNI_FALSE;
+            return false;
+        }
+    }
+
     static boolean isIPv4Mapped(CCharPointer caddr) {
```
Peter-B-Kessler: Not your fault: This method should almost certainly move to `JavaNetNetUtilMD.NET_IsIPv4Mapped`.  I will file an issue to move it, either as a separate commit when I merge your changes, or later.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaNetNetUtil.java
```diff
@@ -223,6 +223,114 @@ static InetAddress NET_SockaddrToInetAddress(Socket.sockaddr him, CIntPointer po
...
+                // 273 if (family == AF_INET6) {
+                if (family == Socket.AF_INET6()) {
+                    // 274 return JNI_FALSE;
+                    return false;
+                }
+                // 276 addrNew = NET_IPv4MappedToIPv4(caddrNew);
+                addrNew = JavaNetNetUtilMD.NET_IPv4MappedToIPv4(caddrNew);
+                // 277 addrCur = getInetAddress_addr(env, iaObj);
+                addrCur = JavaNetNetUtilMD.getInetAddress_addr(iaObj);
+                // 278 JNU_CHECK_EXCEPTION_RETURN(env, JNI_FALSE);
```
Peter-B-Kessler: I wonder who they were expecting to post an exception?  Oh well, if an exception is thrown we do not need this check!

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaNetNetUtil.java
```diff
@@ -223,6 +223,114 @@ static InetAddress NET_SockaddrToInetAddress(Socket.sockaddr him, CIntPointer po
...
+                    // 280 return JNI_TRUE;
+                    return true;
+                } else {
+                    // 282 return JNI_FALSE;
+                    return false;
+                }
+                // 284 } else {
+            } else {
+                // 285 jbyteArray ipaddress;
+                byte[] ipaddr;
```
Peter-B-Kessler: s/ipaddr/ipaddress/

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaNetNetUtil.java
```diff
@@ -223,6 +223,114 @@ static InetAddress NET_SockaddrToInetAddress(Socket.sockaddr him, CIntPointer po
...
+                // 286 jbyte caddrCur[16];
+                CCharPointer caddrCur = StackValue.get(16);
```
Peter-B-Kessler: Technically this should be `StackValue.get(16, CCharPointer.class)`.  Not worth a whole new pull request.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaNetNetUtil.java
Peter-B-Kessler: On the other hand, it appears to be unused.  Maybe change the translation to `/* Unused. */`.  Or leave it as a faithful translation.  (I just looked in the JDK-11 sources, and it is gone from there, so there is no point in filing a bug against the JDK.)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/JavaNetNetUtil.java
```diff
@@ -463,6 +578,30 @@ static int NET_IPv4MappedToIPv4(CCharPointer caddr) {
...
+        int i;
+        // 966 for (i = 0; i < 16; i++) {
+        for (i = 0; i < 16; i++) {
+            // 967 if (caddr1[i] != caddr2[i]) {
+            if (caddr1.read(i) != caddr2.read(i)) {
+                // 968 return 0; /* false */
+                return false;
+            }
+        }
+        return true;
```
Peter-B-Kessler: Not worth another pull request:
```
// 971 return 1;
```

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -95,16 +99,319 @@ private PosixJavaNetSubstitutions() {
...
+        Socket.sockaddr him = StackValue.get(JavaNetNetUtilMD.SOCKADDR_LEN());
+
+        //   198      if (IS_NULL(fdObj)) {
+        if (fdObj == null) {
+            //   199          JNU_ThrowByName(env, JNU_JAVANETPKG "SocketException",
+            //   200                          "Socket closed");
+            //   201          return;
+            throw new SocketException("Socket closed");
+        } else {
+            fd = PosixUtils.getFD(fdObj);
```
Peter-B-Kessler: The translation goes below the original.

bobmcwhirter:
Aw shucks. I’ll fix all of these tomorrow.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -95,16 +99,319 @@ private PosixJavaNetSubstitutions() {
...
+        //   206      if (IS_NULL(iaObj)) {
+        if (iaObj == null) {
+            //   207          JNU_ThrowNullPointerException(env, "iaObj is null.");
+            //   208          return;
+            throw new NullPointerException("iaObj is null.");
+        }
+
+        //   211      /* bind */
+        //   212      if (NET_InetAddressToSockaddr(env, iaObj, localport, (struct sockaddr *)&him, &len, JNI_TRUE) != 0) {
+        len_Pointer.write(0);
```
Peter-B-Kessler: I think this is now redundant with the same line up by the declaration.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -95,16 +99,319 @@ private PosixJavaNetSubstitutions() {
...
+            if (VmPrimsJVM.JVM_GetSockName(fd, him, len_Pointer) == -1) {
+                //   235              NET_ThrowByNameWithLastError(env, JNU_JAVANETPKG "SocketException",
+                //   236                              "Error getting socket name");
+                try {
+                    JavaNetNetUtilMD.NET_ThrowNew(Errno.errno(), "Error getting socket name");
+                    //   237              return;
+                    return;
+                } finally {
+                    return;
+                }
```
Peter-B-Kessler: You do not need the `try .... finally`.
You could add a comment that says that the C line 237 is unreachable in the Java code.

Peter-B-Kessler:
I am not quite done with the review, but it is looking really good.  You did a really thorough job.  There are some style things that Eclipse (and our style rules) point out, but I can fix them when I merge your changes.

Peter-B-Kessler:
And, of course, the crucial question: Does this run whatever it was you wanted to push through DatagramSockets?  Alternatively: do you have a simple test you could contribute?  (Our tests are not open.  Not my decision.)  If you have a test I could see if I can add it so we keep DatagramSockets working.

Peter-B-Kessler:
There will be a way to contribute open tests.  Real soon now.

bobmcwhirter:
So it does work. But my local monkeytest involves a native-image’d app
using jaeger with a jaeger server running in a Docker container. Not
particularly ... automated.

I can try to write up a simple main() with some threads that talk amongst
themselves. Or a JUnit if I can figure out how and where to wire it up at
least temporarily.

Does my RHT/OpenJDK/OCA-signer status help with access to the
aforementioned non-open tests?

bobmcwhirter:
Alrighty, I think I've addressed everything you've pointed out.

bobmcwhirter:
That CI failure doesn't look like me...

    There was 1 failure:
    1) testSourceFreeContextStrong(com.oracle.truffle.api.test.polyglot.PolyglotCachingTest)

Peter-B-Kessler:
I see the error on `testSourceFreeContextStrong` very occasionally in our internal testing (without your changes). Not your fault.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
Peter-B-Kessler: But the original C code is, as you quoted,
```
//   235              NET_ThrowByNameWithLastError(env, JNU_JAVANETPKG "SocketException",
//   236                              "Error getting socket name");
```
which can only throw a `SocketException`.  The translation should be
```
throw new SocketException(PosixUtils.lastErrorString("Error getting socket name"));
```
That matches the declaration of the Java method `PlainDatagramSocketImpl.bind0`, which says it only throws `SocketException`.

bobmcwhirter: Yah, I finally realized that. Some days I can be a tad slow on the uptake.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+            //   785      jboolean connected = JNI_FALSE;
+            boolean connected = false;
+            //   786      jobject connectedAddress = NULL;
+            InetAddress connectedAddress = null;
+            //   787      jint connectedPort = 0;
+            int connectedPort = 0;
+            //   788      jlong prevTime = 0;
+            long prevTime = 0;
+            //   789  #endif
+        }
```
Peter-B-Kessler: Well, this is not going to do what you want.  You have declared 4 locals that go out of scope when you leave the `then` branch.  On the other hand, I do not see where these variables are used, anywhere in this method.  So maybe the best thing to do is leave the C code as `//` comments and add a Java comment `/* Unused. */`.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+                            //   870                  if (mallocedPacket) {
+                            if (mallocedPacket) {
+                                //   871                      free(fullPacket);
+                                LibC.free(fullPacket);
+                                /* so that the outter try/finally does not attempt to double-free() */
+                                mallocedPacket = false;
+                            }
+                            //   874                  return;
+                            return;
+                        }
```
Peter-B-Kessler: `javac` complains on this line
```
graal/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java:609: warning: [finally] finally clause cannot complete normally
                        }
                        ^
```
which is not much help until you look up what the error message means: that you have a `return` in a `finally` block.

I think you have to either put the `return` just after of this `try/finally`, or trust the outer `try/finally` to free `fullPacket` when (not if!) an exception gets thrown from this code.  (Your translation is already aware of the "outter" (sic) `try/finally`.)  I would be in favor of one `try/finally` starting right after the potential `malloc` and enforcing the potential `free`.

Peter-B-Kessler: I get that same warning on lines 609, 996, 1024, 1040, 1060, 1086, and 1117.  I am surprised you don't get them.  We do not allow code that generates warnings.

bobmcwhirter:
Sorry I’m good at ignoring warnings. Will rectify. One day, I’ll get it in shape. One day. But not today. Apparently.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+                    n = packetBufferLen;
+                }
+
+                //   885          if (n == JVM_IO_ERR) {
+                if (n == Target_jvm.JVM_IO_ERR()) {
+                    //   886              (*env)->SetIntField(env, packet, dp_offsetID, 0);
+                    Util_java_net_DatagramPacket.as_Target_java_net_DatagramPacket(packet).offset = 0;
+                    //   887              (*env)->SetIntField(env, packet, dp_lengthID, 0);
+                    Util_java_net_DatagramPacket.as_Target_java_net_DatagramPacket(packet).length = 0;
+                    if (Errno.errno() == Errno.ECONNREFUSED()) {
```
Peter-B-Kessler: How much better would we all sleep at night if you saved a copy of `Errno.errno()` after the call to `JavaNetNetUtilMD.NET_Recv` and used that for these tests, rather than assuming that none of the intervening calls did anything that might change that variable?  I don't *think* anything does, here, but it is easy enough to be safe.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+                        //   890                                  "ICMP Port Unreachable");
+                        //   891              } else {
+                    } else {
+                        //   892                  if (errno == EBADF) {
+                        if (Errno.errno() == Errno.EBADF()) {
+                            //   893                       JNU_ThrowByName(env, JNU_JAVANETPKG "SocketException", "Socket closed");
+                            throw new SocketException("Socket closed");
+                        } else {
+                            //   895                       NET_ThrowByNameWithLastError(env, JNU_JAVANETPKG "SocketException", "Receive failed");
+                            throw new SocketException("Receive failed");
```
Peter-B-Kessler: Missing the `PosixUtils.lastErrorString` part of the translation.  Hmm, and now that I look, I missed commenting on several others that are missing.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+            InetAddress iaObj;
+            //  1760
+            //  1761          len = SOCKADDR_LEN;
+            len_Pointer.write(JavaNetNetUtilMD.SOCKADDR_LEN());
+            //  1762
+            //  1763          if (getsockname(fd, (struct sockaddr *)&him, &len) == -1) {
+            if (Target_os.get_sock_name(fd, him, len_Pointer) == -1) {
+                //  1764              NET_ThrowByNameWithLastError(env, JNU_JAVANETPKG "SocketException",
+                //  1765                             "Error getting socket name");
+                throw new SocketException("Error getting socket name");
```
Peter-B-Kessler: Missing the "with last error" part.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+            optlen_Pointer.write(SizeOf.get(CIntPointer.class));
+            optval_Pointer = optval_i_Pointer;
+        }
+        //  1788
+        //  1789      if (NET_GetSockOpt(fd, level, optname, (void *)&optval, &optlen) < 0) {
+        if (JavaNetNetUtilMD.NET_GetSockOpt(fd, level_Pointer.read(), optname_Pointer.read(), optval_Pointer, optlen_Pointer) < 0) {
+            //  1790          NET_ThrowByNameWithLastError(env, JNU_JAVANETPKG "SocketException",
+            //  1791                           "Error getting socket option");
+            //  1792          return NULL;
+            throw new SocketException("Error getting socket option");
```
Peter-B-Kessler: With last error.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+        }
+
+        //  1795      switch (opt) {
```
Peter-B-Kessler: I think this could be a Java `switch` since the arm labels are all constants.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+        //  1796          case java_net_SocketOptions_IP_MULTICAST_LOOP:
+        if (opt == SocketOptions.IP_MULTICAST_LOOP) {
+            //  1798              if (level == IPPROTO_IP) {
+            if (level_Pointer.read() == NetinetIn.IPPROTO_IP()) {
+                //  1799                  return createBoolean(env, (int)!optval.c);
```
Peter-B-Kessler: Since you need these more than once, I think you should leave them as calls to methods.  

That also separates the computing of the argument to the method (an `int`) from the implementation of the method.  I do not want to track the `!` and `== 0` here through the `!= 0` in `createBoolean`.

I would have a `createInteger` method, even though it is trivial.  Our compilers are good at inlining trivial methods.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+                    return Boolean.TRUE;
+                }
+                return Boolean.FALSE;
+            }
+        }
+        //  1804          case java_net_SocketOptions_SO_BROADCAST:
+        //  1805          case java_net_SocketOptions_SO_REUSEADDR:
+        else if (opt == SocketOptions.SO_BROADCAST || opt == SocketOptions.SO_REUSEADDR) {
+            //  1806              return createBoolean(env, optval.i);
+            if (optval_i_Pointer.read() == 0) {
```
Peter-B-Kessler: This is `optval.i`, not `!optval.i`.  See comment above about arguments to even (especially?) trivial methods.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
-    protected void datagramSocketCreate() {
-        throw VMError.unsupportedFeature("Unimplemented: java.net.PlainDatagramSocketImpl.datagramSocketCreate()");
+    @SuppressWarnings({"static-method", "unused"})
+    protected void datagramSocketCreate() throws SocketException {
+
+
+        //   957      jobject fdObj = (*env)->GetObjectField(env, this, pdsi_fdID);
+        FileDescriptor fdObj = Util_java_net_DatagramSocketImpl.as_Target_java_net_DatagramSocketImpl(this).fd;
+        //   958      int arg, fd, t = 1;
+        int arg, fd, t = 1;
```
Peter-B-Kessler: Just a heads-up: Our style guidelines only allow one declaration per line.  I will fix all our style constraints in a separate commit in my pull request to master.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+            throw new SocketException("Socket closed");
+            //   970      }
+        }
+
+        //   972      if ((fd = JVM_Socket(domain, SOCK_DGRAM, 0)) == JVM_IO_ERR) {
+        if ((fd = Socket.socket(domain, Socket.SOCK_DGRAM(), 0)) == Target_jvm.JVM_IO_ERR()) {
+            //   973          NET_ThrowByNameWithLastError(env, JNU_JAVANETPKG "SocketException",
+            //   974                         "Error creating socket");
+            throw new SocketException(PosixUtils.lastErrorString("Error creating socket") + " (Error creating socket)");
+            //   975          return;
```
Peter-B-Kessler: A string concatenation from the Department of Redundancy Department.

bobmcwhirter:
Right on. Thanks for your patience at this. I know it’s probably pretty
tedious. But I’ll improve.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+                //   983                         sizeof(int)) < 0) {
+                if (Socket.setsockopt(fd, NetinetIn.IPPROTO_IPV6(), NetinetIn.IPV6_V6ONLY(), arg_Pointer, SizeOf.get(CIntPointer.class)) < 0) {
+                    //   984              NET_ThrowNew(env, errno, "cannot set IPPROTO_IPV6");
+                    try {
+                        JavaNetNetUtilMD.NET_ThrowNew(Errno.errno(), "cannot set IPPROTO_IPV6");
+                    } finally {
+                        //   985              close(fd);
+                        //   986              return;
+                        Unistd.close(fd);
+                        /* return required to prevent compiler from complaining about unhandled InterruptedIOException */
```
Peter-B-Kessler: (You have `JavaNetNetUtilMD.NET_ThrowNew` throwing an `InterruptedException` instead of an `InterruptedIOException`.  I will try to remember to comment on that when I get to that method.)

Since `PlainDatagramSocketImpl.datagramSocketCreate` is declared to only throw `SocketException`, but we are the only ones who have to live by that rule, I suggest you turn an `InterruptedIOException` plowing through here by throwing a new `SocketException` whose message string is "InterruptedIOException" plus the message string from the `InterruptedIOException`, if that is non-null.

On the other hand, `setsockopt(2)` says it can not return with `errno` set to `EINTR`, so maybe we should not spend too much time agonizing over this.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
-        throw VMError.unsupportedFeature("Unimplemented: java.net.PlainDatagramSocketImpl.datagramSocketCreate()");
+    @SuppressWarnings({"static-method", "unused"})
+    protected void datagramSocketCreate() throws SocketException {
+
+
+        //   957      jobject fdObj = (*env)->GetObjectField(env, this, pdsi_fdID);
+        FileDescriptor fdObj = Util_java_net_DatagramSocketImpl.as_Target_java_net_DatagramSocketImpl(this).fd;
+        //   958      int arg, fd, t = 1;
+        int arg, fd, t = 1;
+
```
Peter-B-Kessler: You omitted 
```
959    char tmpbuf[1024];
```
You are not using it, but it would be good to leave the line in (commented out) with an explanation that you find Java Strings easier to deal with.  Otherwise readers get confused when they get to code that seems to use it.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+            arg = 65507;
+            CIntPointer arg_Pointer = StackValue.get(CIntPointer.class);
+            arg_Pointer.write(arg);
+            //   993      if (JVM_SetSockOpt(fd, SOL_SOCKET, SO_SNDBUF,
+            //   994                         (char *)&arg, sizeof(arg)) < 0) {
+            if (Socket.setsockopt(fd, Socket.SOL_SOCKET(), Socket.SO_SNDBUF(), arg_Pointer, SizeOf.get(CIntPointer.class)) < 0) {
+                try {
+                    //   995          getErrorString(errno, tmpbuf, sizeof(tmpbuf));
+                    //   996          JNU_ThrowByName(env, JNU_JAVANETPKG "SocketException", tmpbuf);
+                    JavaNetNetUtilMD.NET_ThrowNew(Errno.errno(), PosixUtils.lastErrorString("cannot set socket option"));
```
Peter-B-Kessler: Why aren't these (all) `throw new SocketException(PosixUtils.lastErrorString(PosixUtils.lastErrorString("cannot set socket option")))` like the C code says?  With the default string changed appropriately on the other instances?

Or be completely pedantic and after declaring `String tmpbuf` at the top of the method, say 
```
tmpbuf = PosixUtils.lastErrorString("cannot set socket option");
throw new SocketException(tmpbuf);
```

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+        //  1918      jint fd = -1;
+        int fd = -1;
+
+        //  1920      if (IS_NULL(fdObj)) {
+        if (fdObj == null) {
+            //  1921          JNU_ThrowByName(env, JNU_JAVANETPKG "SocketException",
+            //  1922                          "Socket closed");
+            //  1923          return -1;
+            //throw new SocketException("Socket closed");
+            return -1;
```
Peter-B-Kessler: This is a weird one, because you have to choose between throwing the exception or returning a value.  Java code that calls this through JNI will get the exception before (instead of) the return value.  I think you should throw the exception.  Unless you have some local Java code that wants the return value and will do the right thing with it.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+        } else {
+            //  1925          fd = (*env)->GetIntField(env, fdObj, IO_fd_fdID);
+            fd = PosixUtils.getFD(fdObj);
+        }
+
+        //  1927      /* getsockopt of TTL */
+        //  1928  #ifdef AF_INET6
+        //  1929      if (ipv6_available()) {
+        if (IsDefined.socket_AF_INET6() && JavaNetNetUtil.ipv6_available()) {
```
Peter-B-Kessler: Grr.  People who use macros to add control flow.  Not your fault.  I'll get over it.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+        FileDescriptor fdObj = Util_java_net_DatagramSocketImpl.as_Target_java_net_DatagramSocketImpl(this).fd;
+        //   258      /* The fdObj'fd */
+        //   259      jint fd;
+        int fd;
+        //   260      /* The packetAddress address, family and port */
+        //   261      SOCKADDR rmtaddr;
+        Socket.sockaddr rmtaddr = StackValue.get(Socket.sockaddr.class);
+
+        //   262      int len = 0;
+        CIntPointer len_Pointer = StackValue.get(CIntPointer.class);
```
Peter-B-Kessler: `len_Pointer.write(0)`.  Maybe unnecessary, but I want to read this line of translation and know that it is correct in and of itself.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+    @Alias InetAddress address;
+    @Alias int offset;
+    @Alias int length;
+    @Alias int bufLength;
+    @Alias int port;
+}
+
+class Util_java_net_DatagramPacket {
+    static Target_java_net_DatagramPacket as_Target_java_net_DatagramPacket(DatagramPacket packet) {
+        return KnownIntrinsics.unsafeCast(packet, Target_java_net_DatagramPacket.class);
```
Peter-B-Kessler: A little late, and not in the style of the rest of the code in this file, but casts between "real" Java types and substitute "Target" types can be done with `.class.cast`.  See `JavaThreads.fromTarget(Target_java_lang_Thread)` and `JavaThreads.toTarget(Thread)`.  The types really are the same at runtime, otherwise the `unsafeCast` would not work.  But you have to put the `@SuppressFBWarnings(value = "BC", justification = "Cast for @TargetClass") annotation on the casting methods or FindBugs objects that the cast can never work.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+    @Alias int localPort;
+}
+
+class Util_java_net_DatagramSocketImpl {
+    static Target_java_net_DatagramSocketImpl as_Target_java_net_DatagramSocketImpl(Target_java_net_PlainDatagramSocketImpl tjnsi) {
+        return KnownIntrinsics.unsafeCast(tjnsi, Target_java_net_DatagramSocketImpl.class);
+    }
+}
+
+class Util_java_net_PlainDatagramSocketImpl {
```
Peter-B-Kessler: This is interesting.  We conventionally use "Util_" classes as collections of new Java methods that are useful to "Target_" implementations.  Here you are using it also as a collection of C "file static" methods from PlainDatagramSocketImpl.c that do not otherwise have a place to live.  I guess that's cool.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+            //  1516          int len = sizeof(struct in_addr);
+            CIntPointer len_Pointer = StackValue.get(CIntPointer.class);
+            //  1517
+            //  1518          if (JVM_GetSockOpt(fd, IPPROTO_IP, IP_MULTICAST_IF,
+            //  1519                             (char *)inP, &len) < 0) {
+            if (VmPrimsJVM.JVM_GetSockOpt(fd, NetinetIn.IPPROTO_IP(), NetinetIn.IP_MULTICAST_IF(), (CCharPointer) inP, len_Pointer) < 0) {
+                //  1520              NET_ThrowByNameWithLastError(env, JNU_JAVANETPKG "SocketException",
+                //  1521                               "Error getting socket option");
+                //  1522              return NULL;
+                throw new SocketException("Error getting socket option");
```
Peter-B-Kessler: Missing "with last error".

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+            }
+
+            //  1573          /*
+            //  1574           * The address doesn't appear to be bound at any known
+            //  1575           * NetworkInterface. Therefore we construct a NetworkInterface
+            //  1576           * with this address.
+            //  1577           */
+            //  1578          ni = (*env)->NewObject(env, ni_class, ni_ctrID, 0);
+            ni = Util_java_net_NetworkInterface.newNetworkInterface();
+            //  1579          CHECK_NULL_RETURN(ni, NULL);
```
Peter-B-Kessler: You should add a comment saying that `Util_java_net_NetworkInterface.newNetworkInterface()` will never return null.

Similarly below after `new InetAddress[1]`.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -119,84 +419,1339 @@ protected int peekData(DatagramPacket p) throws IOException {
...
+                //  1601          static jclass ni_class;
+                //  1602          static jmethodID ni_ctrID;
+                //  1603          static jfieldID ni_indexID;
+                //  1604          static jfieldID ni_addrsID;
+                //  1605          static jclass ia_class;
+                //  1606          static jfieldID ni_nameID;
+                //  1607          static jmethodID ia_anyLocalAddressID;
+                //  1608
+                //  1609          int index = 0;
+                CCharPointer index_Pointer = StackValue.get(CCharPointer.class);
```
Peter-B-Kessler: index_Pointer.write(0);

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
Peter-B-Kessler: Nah, never mind.  No one else does that.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
Peter-B-Kessler: I don't know how else to find these things except by code reviews.

Peter-B-Kessler:
I am through one complete pass.  A tremendous amount of work, really well done, and definitely the deep end of the "substitutions" pool.
If it runs your sample code, and passes our internal gate tests, I am willing to accept it (after you fix the things I have commented on, which you seem really good about doing).
I will fix the things that violate our style rules. (Otherwise it will not pass our internal gate tests.)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
bobmcwhirter: In this case, `NET_ThrowNew(...)` existed prior to my work. I'm just using it.

bobmcwhirter:
Added a commit (instead of squash and `push -f`) to reduce surface area for review.

I think I've nailed everything, did more grepping around for a few more _with last error_ exceptions to fix.

Left the `InterruptedException` from `NET_ThrowNew` as it's not code I wrote, and other code I'm unfamiliar with doesn't expect a change from `InterruptedException` to `InterruptedIOException` and changing those callsites seem out of scope for this PR.

Only warning I see during the build now is (not mine):

    Warning: Detected unnecessary RecomputeFieldValue.ArrayBaseOffset com.oracle.svm.core.jdk.Target_java_nio_DirectByteBuffer.arrayBaseOffset substitution field for java.nio.DirectByteBuffer.arrayBaseOffset. The annotated field can be removed. This ArrayBaseOffset computation can be detected automatically. Use option -H:+UnsafeAutomaticSubstitutionsLogLevel=2 to print all automatically detected substitutions.

If I ever make it to JavaOne^H^H^H^H^H^H^HCodeOne again, I'll buy you a beverage of your choosing.

Peter-B-Kessler:
Progress report: I pulled your branch, merged it with our master, fixed a few style things (one declaration per line, complaints about names with underscores, etc.), turned off the Eclipse formatter around the C code you quoted (so it didn't all get folded and reinvented), and created in internal pull request.  It passed all our tests (so at least we didn't break anything crucial).  Now I am just waiting for our internal review process to grind through.

bobmcwhirter:
You rock! Thank you!

Peter-B-Kessler:
Merged.  The internal review wanted to request a test of the new functionality, but admitted that there is no process for external people committing tests.  When that process opens, you could consider submitting a test.

bobmcwhirter:
For future ref.. should do lenPointer instead of len_Pointer?

bobmcwhirter:
I’ll also fire up eclipse now so I can live the same lifestyle. Are vi keybindings baked in or do I still need to buy the viPlugin?

Peter-B-Kessler:
You can also run `mx checkstyle` and it should run all the style checks that our gates run.  No need for eclipse, certainly not for the vi keybinding plugin.

Peter-B-Kessler:
Re: `lenPointer` instead of `len_Pointer`.  I got myself into trouble when renaming locals that were hiding fields of the instance.  I was trying out a convention of, for example, changing the local `connect` to `localConnect`, and then I saw the variable `localport` and thought that would be confusing, so I went with `connectLocal`.  Probably I should have used `connect_Local`.  I really want to distinguish the rare Java variable whose usage is different from the corresponding C variable.  (Though, turning off checkstyle to allow underscores in names also masks other checkstyle violations, like locals hiding fields of the instance.)

PR #628, opened 2018-08-21
Add examples of -H:IncludeResources command line options

### LINKED ISSUE #626: Support for multiple values in -H:IncludeResources
According to the docs `-H:IncludeResources` is a single regex, and that seems to be the way it works. It's just not very convenient, if you want to include specific sets of resources, from more than one location - the only pattern I found that worked is `.*` so the binary image is enormous when it is finished building. Maybe multiple values could be supported through a comma-separated list (like `-H:IncludeResourceBundles` already supports)?

dsyer:
Fixes #626

### DIFF
--- a/substratevm/RESOURCES.md
+++ b/substratevm/RESOURCES.md
@@ -9,6 +9,8 @@ specified with the following option
 ```bash
 -H:IncludeResources=<Java regexp that matches resources to be included in the image>
 ```
+For example: `-H:IncludeResources=/META-INF/.*`. Note that multiple patterns can be 
+joined together with `|`, e.g. `-H:IncludeResources=/META-INF/.*|application.properties`.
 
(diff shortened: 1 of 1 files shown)


dsyer:
I sent in the OCA. Please check.

olpaw:
@dsyer I added you the the contributors db. Can you rebase and push again so that I see if the  
oca-check succeeds now.

olpaw:
@dsyer so far you are not showing up in http://www.oracle.com/technetwork/community/oca-486395.html
I sent an email to get verification you signed the OCA. Sorry for the inconvenience.

olpaw:
We have provided more sensible docu for `-H:IncludeResources` on our own.
See https://github.com/oracle/graal/blob/master/substratevm/RESOURCES.md

PR #636, opened 2018-08-24
virtualize unsafe compare and swap calls

fwbrasil:
## Problem

Compare and swap calls using `Unsafe` are currently always lowered to an atomic operation. That's unnecessary when the CAS mutates a field of an instance that is virtual. It introduces some performance overhead and prevents other optimizations.

Let's take this JMH benchmark class as an example:

```java
public class VirtualCASBench {

  public static final long valueOffset = UnsafeUtil.fieldOffset(TestClass.class, "value");

  private static class TestClass {
    public volatile int value;

    public TestClass(int value) {
      this.value = value;
    }
  }

  @Benchmark
  public boolean testUnsafe() {
    TestClass t = new TestClass(0);
    return UnsafeUtil.unsafe.compareAndSwapInt(t, valueOffset, 0, 1);
  }

  @Benchmark
  public boolean testIfElse() {
    TestClass t = new TestClass(0);
    synchronized (t) {
      if (t.value != 0)
        return false;
      else {
        t.value = 1;
        return true;
      }
    }
  }
}
```

The if/else version has much better performance than the version using the unsafe compare and swap:

image

It is optimized before lowering to a graph that returns a constant:

image

That's expected since the instance doesn't escape and constant folding can determine that the CAS will always return true. The same doesn't happen with the unsafe version since the compare and swap node is opaque to constant folding:

image

## Solution

When an object is virtualized, compare and swap can be replaced by a guard and the new value can be set using the `VirtualizerTool`. I've implemented such optimization and the benchmark results are promising:

image

With the CAS virtualization, the unsafe compare and swap benchmark is also optimized to a constant:

image

## Notes and questions

1. I've used a guard that will trigger deoptimization in case the current value doesn't match the expected value. It assumes that users won't write code that will makes a CAS operation fail since the object doesn't escape and can't be used concurrently.

2. I couldn't link the guard node to the predecessor using the `VirtualizerTool`, so I set it in `UnsafeCompareAndSwapNode.virtualize`. I'm not sure if that could be problematic since it seems that all effects should be done through the tool during virtualization.

3. The optimization is applied only when the field can be resolved. If the user uses a dynamic or invalid offset, it won't be applied.

4. I've added an option to enable the optimization. Should it be enabled by default?

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ea/UnsafeCompareAndSwapVirtualizationTest.java
+++ b/compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ea/UnsafeCompareAndSwapVirtualizationTest.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2011, 2016, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.graalvm.compiler.core.test.ea;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.graalvm.compiler.nodes.java.LogicCompareAndSwapNode;
+import org.junit.Test;
+
+import jdk.vm.ci.meta.JavaConstant;
+
+public class UnsafeCompareAndSwapVirtualizationTest extends EATestBase {
+
+    private static Object obj1 = new Object();
+    private static Object obj2 = new Object();
+
+    public static boolean bothVirtualNoMatch() {
+        AtomicReference<Object> a = new AtomicReference<>();
+        return a.compareAndSet(new Object(), new Object());
+    }
+
+    @Test
+    public void bothVirtualNoMatchTest() {
+        testEscapeAnalysis("bothVirtualNoMatch", JavaConstant.INT_0, true);
+        assertTrue(graph.getNodes(LogicCompareAndSwapNode.TYPE).isEmpty());
+    }
+
+    public static boolean bothVirtualMatch() {
+        Object expect = new Object();
+        AtomicReference<Object> a = new AtomicReference<>(expect);
+        return a.compareAndSet(expect, new Object());
+    }
+
+    @Test
+    public void bothVirtualMatchTest() {
+        testEscapeAnalysis("bothVirtualMatch", JavaConstant.INT_1, true);
+        assertTrue(graph.getNodes(LogicCompareAndSwapNode.TYPE).isEmpty());
+    }
+
+    public static boolean expectedVirtualMatch() {
+        Object o = new Object();
+        AtomicReference<Object> a = new AtomicReference<>(o);
+        return a.compareAndSet(o, obj1);
+    }
+
+    @Test
+    public void expectedVirtualMatchTest() {
+        testEscapeAnalysis("expectedVirtualMatch", null, true);
+        assertTrue(graph.getNodes(LogicCompareAndSwapNode.TYPE).isEmpty());
+    }
+
+    public static boolean expectedVirtualNoMatch() {
+        Object o = new Object();
+        AtomicReference<Object> a = new AtomicReference<>();
+        return a.compareAndSet(o, obj1);
+    }
+
+    @Test
+    public void expectedVirtualNoMatchTest() {
+        testEscapeAnalysis("expectedVirtualNoMatch", null, true);
+        assertTrue(graph.getNodes(LogicCompareAndSwapNode.TYPE).isEmpty());
+    }
+
+    public static boolean bothNonVirtualNoMatch() {
+        AtomicReference<Object> a = new AtomicReference<>();
+        return a.compareAndSet(obj1, obj2);
+    }
+
+    @Test
+    public void bothNonVirtualNoMatchTest() {
+        testEscapeAnalysis("bothNonVirtualNoMatch", null, true);
+        assertTrue(graph.getNodes(LogicCompareAndSwapNode.TYPE).isEmpty());
+    }
+
+    public static boolean bothNonVirtualMatch() {
+        AtomicReference<Object> a = new AtomicReference<>(obj1);
+        return a.compareAndSet(obj1, obj2);
+    }
+
+    @Test
+    public void bothNonVirtualMatchTest() {
+        testEscapeAnalysis("bothNonVirtualMatch", null, true);
+        assertTrue(graph.getNodes(LogicCompareAndSwapNode.TYPE).isEmpty());
+    }
+
+    public static boolean onlyInitialValueVirtualMatch() {
+        AtomicReference<Object> a = new AtomicReference<>(new Object());
+        return a.compareAndSet(obj1, obj2);
+    }
+}
--- a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/UnsafeCompareAndSwapNode.java
+++ b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/UnsafeCompareAndSwapNode.java
@@ -24,6 +24,7 @@
  */
 package org.graalvm.compiler.nodes.java;
 
+import static org.graalvm.compiler.core.common.calc.CanonicalCondition.EQ;
 import static org.graalvm.compiler.nodeinfo.InputType.Memory;
 import static org.graalvm.compiler.nodeinfo.InputType.Value;
 import static org.graalvm.compiler.nodeinfo.NodeCycles.CYCLES_8;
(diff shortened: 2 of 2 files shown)


fwbrasil:
Twitter has already signed the OCA, but I've also submitted the form by email.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/UnsafeCompareAndSwapNode.java
```diff
@@ -98,4 +120,41 @@ public LocationIdentity getLocationIdentity() {
...
+
+            int index = resolveFieldIndex(obj);
+
+            if (index >= 0) {
+
+                ValueNode read = tool.getEntry(obj, index);
+
+                LogicNode equalsNode = CompareNode.createCompareNode(EQ, expected, read, tool.getConstantReflectionProvider(), NodeView.DEFAULT);
+
+                FixedGuardNode guardNode = new FixedGuardNode(equalsNode, DeoptimizationReason.Aliasing, DeoptimizationAction.InvalidateRecompile);
```
gilles-duboscq: A deopt without any alternative is a bit dangerous performance-wise: it might lead to deopt-loops. In particular here the "action" is `InvalidateRecompile` which is quite costly considering that the recompilation will lead to exactly the same code.

What about just using `ConditionalNodes`?
```java
ConditionalNode fieldValue = ConditionalNode.create(equalsNode, newValue, read);
ConditionalNode result = ConditionalNode.create(equalsNode);
//...
tool.setVirtualEntry(obj, index, fieldValue);
//...
tool.replaceWith(result);
```
In the "simple" cases it will immediately fold away and in more complex cases where equality can't be statically proven, performance should still be OK.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/UnsafeCompareAndSwapNode.java
```diff
@@ -32,22 +33,43 @@
...
  */
 @NodeInfo(allowedUsageTypes = {Value, Memory}, cycles = CYCLES_8, size = SIZE_8)
-public final class UnsafeCompareAndSwapNode extends AbstractMemoryCheckpoint implements Lowerable, MemoryCheckpoint.Single {
+public final class UnsafeCompareAndSwapNode extends AbstractMemoryCheckpoint implements Lowerable, MemoryCheckpoint.Single, Virtualizable {
+
+    public static class Options {
+
+        // @formatter:off
+        @Option(help = "Virtualize unsafe CAS calls", type = OptionType.Expert)
+        public static final OptionKey<Boolean> VirtualCAS = new OptionKey<>(false);
```
gilles-duboscq: This option is fine while you're experimenting but ultimately i don't think we need an option for this.

gilles-duboscq:
Thank you for the PR, that's a good idea.

I've answered some of your questions in comments.

Regarding the `predecessor` issue, the tool's `addNode` method weaves fixed nodes as needed: when the `addNode` is applied it adds the fixed node before the node that is being virtualized using `StructuredGraph#addBeforeFixed`.

fwbrasil:
@gilles-duboscq thank you for the review! I've changed the impl to use `ConditionalNodes` as suggested. I had to add a condition for when the logic node becomes a constant since the graph becomes invalid if I create the other nodes based on the logic constant.

gilles-duboscq:
> I had to add a condition for when the logic node becomes a constant since the graph becomes invalid if I create the other nodes based on the logic constant.

Right. I guess this is due to `tool.addNode` choking on nodes that are already in the graph. That's a bit inconvenient but we can probably work around that:
* `equalsNode` is always a new node (a constant or a comparison)
* `fieldValue` might be a new node, `newValue`, or `currentValue`
* `result` is always a new node

Based on this it might be enough to just gate the `tool.addNode(fieldValue)` on `if (!fieldValue.isAlive())`. If that works it would simplify this code a bit.

In any case, i think we can remove the option and we should be good to go.

fwbrasil:
@gilles-duboscq it seems more complicate than that. For some reason the partial escape analysis tries to remove twice the constant node representing true.

I'm testing this new version with a service and it seems that there's a bug. The constant folding somehow infers the comparison as false in some cases where they should be true. I'm investigating if it could be related to the stamps of the values.

gilles-duboscq:
I tried to write a few unit tests on top of your changes but i couldn't reproduce either error.
Do you have some example code that could reproduce?

fwbrasil:
@gilles-duboscq thank you for looking into it. I haven't been able to reproduce the issue with the logic constant being false in isolation yet. For the invalid graph issue, this always fail if I remove the if condition for logic constants:

```java
    while (true) {
      test();
    }
  }

  private static boolean test() {
    AtomicInteger a = new AtomicInteger(0);
    return a.compareAndSet(0, 1);
  }
```

```
[thread:5] scope: JVMCI CompilerThread0
    [thread:5] scope: JVMCI CompilerThread0.Compiling.GraalCompiler
    Context: StructuredGraph:139{HotSpotMethod<Bugs.main(String[])>}
                [thread:5] scope: JVMCI CompilerThread0.Compiling.GraalCompiler.FrontEnd.HighTier.PartialEscapePhase.iteration 0.EffectsPhaseWithSchedule.DeadCodeEliminationPhase
                Exception raised in scope JVMCI CompilerThread0.Compiling.GraalCompiler.FrontEnd.HighTier.PartialEscapePhase.iteration 0.EffectsPhaseWithSchedule.DeadCodeEliminationPhase: java.lang.ArrayIndexOutOfBoundsException: -15625002
                	at org.graalvm.compiler.graph.NodeBitMap.isMarked(NodeBitMap.java:85)
                	at org.graalvm.compiler.graph.NodeBitMap.isMarked(NodeBitMap.java:71)
                	at org.graalvm.compiler.graph.NodeFlood.isMarked(NodeFlood.java:65)
                	at org.graalvm.compiler.phases.common.DeadCodeEliminationPhase.deleteNodes(DeadCodeEliminationPhase.java:140)
                	at org.graalvm.compiler.phases.common.DeadCodeEliminationPhase.run(DeadCodeEliminationPhase.java:103)
                	at org.graalvm.compiler.phases.Phase.run(Phase.java:49)
                	at org.graalvm.compiler.phases.BasePhase.apply(BasePhase.java:197)
                	at org.graalvm.compiler.phases.Phase.apply(Phase.java:42)
                	at org.graalvm.compiler.phases.Phase.apply(Phase.java:38)
                	at org.graalvm.compiler.virtual.phases.ea.EffectsPhase.runAnalysis(EffectsPhase.java:108)
                	at org.graalvm.compiler.virtual.phases.ea.PartialEscapePhase.run(PartialEscapePhase.java:82)
                	at org.graalvm.compiler.virtual.phases.ea.EffectsPhase.run(EffectsPhase.java:1)
                	at org.graalvm.compiler.phases.BasePhase.apply(BasePhase.java:197)
                	at org.graalvm.compiler.phases.BasePhase.apply(BasePhase.java:139)
                	at org.graalvm.compiler.phases.PhaseSuite.run(PhaseSuite.java:212)
                	at org.graalvm.compiler.phases.BasePhase.apply(BasePhase.java:197)
                	at org.graalvm.compiler.phases.BasePhase.apply(BasePhase.java:139)
                	at org.graalvm.compiler.core.GraalCompiler.emitFrontEnd(GraalCompiler.java:256)
                	at org.graalvm.compiler.core.GraalCompiler.compile(GraalCompiler.java:180)
                	at org.graalvm.compiler.core.GraalCompiler.compileGraph(GraalCompiler.java:165)
                	at org.graalvm.compiler.hotspot.HotSpotGraalCompiler.compileHelper(HotSpotGraalCompiler.java:191)
                	at org.graalvm.compiler.hotspot.HotSpotGraalCompiler.compile(HotSpotGraalCompiler.java:204)
                	at org.graalvm.compiler.hotspot.CompilationTask$HotSpotCompilationWrapper.performCompilation(CompilationTask.java:181)
                	at org.graalvm.compiler.hotspot.CompilationTask$HotSpotCompilationWrapper.performCompilation(CompilationTask.java:1)
                	at org.graalvm.compiler.core.CompilationWrapper.run(CompilationWrapper.java:171)
                	at org.graalvm.compiler.hotspot.CompilationTask.runCompilation(CompilationTask.java:330)
                	at org.graalvm.compiler.hotspot.HotSpotGraalCompiler.compileMethod(HotSpotGraalCompiler.java:144)
                	at org.graalvm.compiler.hotspot.HotSpotGraalCompiler.compileMethod(HotSpotGraalCompiler.java:111)
                	at jdk.vm.ci.hotspot.HotSpotJVMCIRuntime.compileMethod(HotSpotJVMCIRuntime.java:525)
```

fwbrasil:
it's a concurrency bug, not related to the stamps. I'm analyzing it

fwbrasil:
@gilles-duboscq should we add a memory barrier somewhere? It seems that other threads see class instances set through a CAS operation with some of their fields null

gilles-duboscq:
I had a more thorough look at this.
Regarding the NPEs: The issue is that when creating the `Compare` and `Conditional` nodes, care must be taken not to mix virtual and non-virtual nodes: you can not have one side virtual and the other one not.

You should use `tool.getAlias` to get the current node for `expected` & `newValue`. Then you have to check the status of the inputs before creating binary nodes. In general, you can only create them if the inputs are not virtual (i.e., not instances of `VirtualObjectNode`). If both are virtual you can sometimes come to a conclusion. e.g., if 2 `VirtualObjectNode` are `!=` in the IR then they can also not represent `==` objects at runtime if they have identity (`VirtualObjectNode#hasIdentity`).

Regarding the other issue, i can not reproduce it when using
```
if (!fieldValue.isAlive()) {
    tool.addNode(fieldValue);
}
```
Also node that this kind of problems is usually detected earlier if you use `-esa`.

fwbrasil:
the build failure doesn't seem related to the change: https://travis-ci.org/oracle/graal/jobs/424908424#L1693

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/UnsafeCompareAndSwapNode.java
```diff
@@ -98,4 +109,48 @@ public LocationIdentity getLocationIdentity() {
...
+                if (equalsNode instanceof LogicConstantNode) {
+
+                    boolean equals = ((LogicConstantNode) equalsNode).getValue();
+                    if (equals) {
+                        tool.setVirtualEntry(obj, index, newValueAlias);
+                    }
+                    tool.replaceWith(ConstantNode.forBoolean(equals));
+
+                } else if (!(expectedAlias instanceof VirtualObjectNode) && !(newValueAlias instanceof VirtualObjectNode)) {
+                    ValueNode fieldValue = ConditionalNode.create(equalsNode, newValueAlias, currentValue, NodeView.DEFAULT);
```
gilles-duboscq: Did you try
```
Object o = new Object();
AtomicReference<Object> a = new AtomicReference<>(o);
return a.compareAndSet(obj1, obj2);
```
?
If `currentValue` is virtual but expected or new values are not? The created conditional node should lead to issues (virtual only on one side).
You should either disable optimization in this case or look at how `ObjectEqualsNode#virtualize` resolves `==` in various cases (note that you might also be able to resolve the case where both `currentValue` and `expectedAlias` are virtual if you want).

fwbrasil: Just added a test for it: https://github.com/oracle/graal/pull/636/files#diff-4039dbafeadfd69113c08dcedbdd913cR114.

It takes the branch where both values are non-virtual and virtualizes the CAS. I'm also running this version in a large codebase without problems.

Looking at the `ObjectEquals` virtualization, it'll infer that the the objects are not equal if one of them is virtual and the other is not, so it should be ok.

gilles-duboscq: If you return `a.get()` rather than the result of the CAS, you'll see in the graph that it returns a `ConditionalNode` with a `Virtual` node on one side an a `LoadField` on the other, although this test does not fail, we can not generate valid code for that, it would just crash later in the compilation process. See for example this test:
```java
public static Object onlyInitialValueVirtualMatch2() {
    AtomicReference<Object> a = new AtomicReference<>(new Object());
    a.compareAndSet(obj1, obj2);
    return a.get();
}

@Test
public void onlyInitialValueVirtualMatchTest2() {
    testEscapeAnalysis("onlyInitialValueVirtualMatch2", null, true);
    assertTrue(graph.getNodes(LogicCompareAndSwapNode.TYPE).isEmpty());
    test("onlyInitialValueVirtualMatch2");
}
```
This last line with `test` will try to actually compile and use the resulting code which fails because we don't generate code for virtual nodes.

fwbrasil: Thanks for the example. I've added a condition to check if the current value is not virtual as well.

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ea/UnsafeCompareAndSwapVirtualizationTest.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public static boolean onlyInitialValueVirtualMatch() {
+        AtomicReference<Object> a = new AtomicReference<>(new Object());
+        return a.compareAndSet(obj1, obj2);
+    }
+
+    @Test
+    public void onlyInitialValueVirtualMatchTest() {
+        testEscapeAnalysis("onlyInitialValueVirtualMatch", null, true);
```
gilles-duboscq: Unfortunately, this test now fails, (`testEscapeAnalysis` makes sure there is no more allocation in the snippet after escape analysis).
You can:
* just remove it
* handle more virtual vs. non-virtual cases by resolving the `==` like `ObjectEqualsNode#virtualize` does (this case for example is virtual vs. non-virtual where both types have identity -> `false`)

fwbrasil: I've just removed the test

gilles-duboscq:
Thank you @fwbrasil for bearing with all my comments and making the adjustments!
I'll integrate this.

dougxc:
Hi @fwbrasil , have you had a chance to measure the performance impact of this change?

PR #645, opened 2018-08-29
Test the sulong failure from the SVM repo

vjovanov:


vjovanov:
This PR was used only for testing purposes.

PR #647, opened 2018-08-30
Include related tooling in readme.md

samrogerson:
Specifically, link to the [Gradle-Graal](Link to Gradle-Graal project) project

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ GraalVM allows running of following languages which are being developed and test
 
 ## Related Tooling
 
-* [Gradle-Graal](http://github.com/palantir/gradle-graal) - A simple Gradle plugin for interacting with Graal tooling.
+* [gradle-graal](http://github.com/palantir/gradle-graal) - A simple Gradle plugin for interacting with Graal tooling.
 
 
 ## License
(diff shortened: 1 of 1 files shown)


samrogerson:
I have now signed the OCA

shelajev:
Thank you, @samrogerson, for the contribution! I'm trying to learn more about this palantir/gradle-graal plugin and the process of including links from the official documentation to the third party software. 

It might seem like a fairly straightforward change it might take a bit of time on my end to figure it out. 

While we're at it, can you please elaborate on what exactly does this plugin do and why particularly this plugin should be referenced?

samrogerson:
Totally understand it might take some time!

This plugin allows projects using Gradle to create native-images using Graal.  You can see an example [here](https://github.com/palantir/gradle-graal/blob/develop/src/test/groovy/com/palantir/gradle/graal/GradleGraalPluginIntegrationSpec.groovy#L22).  

To my knowledge this plugin is currently the only Gradle plugin that adds support for Graal native-images.  Given how common the use Gradle build tooling is, I felt that linking here would be useful to lower the barrier to entry for starting new projects using Graal, or just experimenting.

PR #666, opened 2018-09-10
[JDK-8198574] Add support for andn, blsi, blsr and blsmsk x86 instructions

jphalimi-intel:
This pull request adds support for the andn, blsi, blsr and blsmsk pattern matchings for amd64 architectures.

### DIFF
--- a/compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
+++ b/compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
@@ -1333,7 +1333,7 @@ public class AMD64Assembler extends AMD64BaseAssembler {
         }
     }
 
-    public static final class VexGeneralPurposeRVMOp extends VexOp {
+    public static final class VexGeneralPurposeRVMOp extends VexRVMOp {
         // @formatter:off
         public static final VexGeneralPurposeRVMOp ANDN   = new VexGeneralPurposeRVMOp("ANDN",   P_,   M_0F38, WIG, 0xF2, VEXOpAssertion.BMI1);
         public static final VexGeneralPurposeRVMOp MULX   = new VexGeneralPurposeRVMOp("MULX",   P_F2, M_0F38, WIG, 0xF6, VEXOpAssertion.BMI2);
@@ -1345,6 +1345,7 @@ public class AMD64Assembler extends AMD64BaseAssembler {
             super(opcode, pp, mmmmm, w, op, assertion);
         }
 
+        @Override
         public void emit(AMD64Assembler asm, AVXSize size, Register dst, Register src1, Register src2) {
             assert assertion.check((AMD64) asm.target.arch, LZ, dst, src1, src2, null);
             assert size == AVXSize.DWORD || size == AVXSize.QWORD;
@@ -1353,6 +1354,7 @@ public class AMD64Assembler extends AMD64BaseAssembler {
             asm.emitModRM(dst, src2);
         }
 
+        @Override
         public void emit(AMD64Assembler asm, AVXSize size, Register dst, Register src1, AMD64Address src2) {
             assert assertion.check((AMD64) asm.target.arch, LZ, dst, src1, null, null);
             assert size == AVXSize.DWORD || size == AVXSize.QWORD;
@@ -1389,6 +1391,36 @@ public class AMD64Assembler extends AMD64BaseAssembler {
             asm.vexPrefix(dst, src2, src1, size, pp, mmmmm, size == AVXSize.DWORD ? W0 : W1);
             asm.emitByte(op);
             asm.emitOperandHelper(dst, src1, 0);
+        }
+    }
+
+    public static final class VexGeneralPurposeRMOp extends VexRMOp {
+        // @formatter:off
+        public static final VexGeneralPurposeRMOp BLSI    = new VexGeneralPurposeRMOp("BLSI",   P_,    M_0F38, WIG, 0xF3, 3, VEXOpAssertion.BMI1);
+        public static final VexGeneralPurposeRMOp BLSMSK  = new VexGeneralPurposeRMOp("BLSMSK", P_,    M_0F38, WIG, 0xF3, 2, VEXOpAssertion.BMI1);
+        public static final VexGeneralPurposeRMOp BLSR    = new VexGeneralPurposeRMOp("BLSR",   P_,    M_0F38, WIG, 0xF3, 1, VEXOpAssertion.BMI1);
+        // @formatter:on
+        private final int ext;
+
+        private VexGeneralPurposeRMOp(String opcode, int pp, int mmmmm, int w, int op, int ext, VEXOpAssertion assertion) {
+            super(opcode, pp, mmmmm, w, op, assertion);
+            this.ext = ext;
+        }
+
+        @Override
+        public void emit(AMD64Assembler asm, AVXSize size, Register dst, Register src) {
+            assert assertion.check((AMD64) asm.target.arch, size, dst, null, null);
+            asm.vexPrefix(AMD64.cpuRegisters[ext], dst, src, size, pp, mmmmm, size == AVXSize.DWORD ? W0 : W1);
+            asm.emitByte(op);
+            asm.emitModRM(ext, src);
+        }
+
+        @Override
+        public void emit(AMD64Assembler asm, AVXSize size, Register dst, AMD64Address src) {
+            assert assertion.check((AMD64) asm.target.arch, size, dst, null, null);
+            asm.vexPrefix(AMD64.cpuRegisters[ext], dst, src, size, pp, mmmmm, size == AVXSize.DWORD ? W0 : W1);
+            asm.emitByte(op);
+            asm.emitOperandHelper(ext, src, 0);
         }
     }
 
--- a/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
+++ b/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
@@ -84,7 +84,10 @@ import org.graalvm.compiler.asm.amd64.AMD64Assembler.AMD64RMIOp;
 import org.graalvm.compiler.asm.amd64.AMD64Assembler.AMD64RMOp;
 import org.graalvm.compiler.asm.amd64.AMD64Assembler.AMD64Shift;
 import org.graalvm.compiler.asm.amd64.AMD64Assembler.SSEOp;
+import org.graalvm.compiler.asm.amd64.AMD64Assembler.VexGeneralPurposeRVMOp;
+import org.graalvm.compiler.asm.amd64.AMD64Assembler.VexGeneralPurposeRMOp;
 import org.graalvm.compiler.asm.amd64.AMD64BaseAssembler.OperandSize;
+import org.graalvm.compiler.asm.amd64.AVXKind.AVXSize;
 import org.graalvm.compiler.core.common.LIRKind;
 import org.graalvm.compiler.core.common.NumUtil;
 import org.graalvm.compiler.core.common.calc.FloatConvert;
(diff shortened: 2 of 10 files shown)


jphalimi-intel:
Hi,

My name is listed on the provided website under "Intel Corp" (look for Jean-Philippe Halimi). I can fill and send the document again if need be. Can you please let me know whether it is necessary?

Jp

dougxc:
Hi @jphalimi-intel , no need to take any further action regarding OCA - I will resolve this.

jphalimi-intel:
Thanks Douglas! :-)

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/vector/AMD64VexBinary.java
```diff
@@ -0,0 +1,73 @@
...
+import org.graalvm.compiler.asm.amd64.AMD64MacroAssembler;
+import org.graalvm.compiler.asm.amd64.AVXKind;
+import org.graalvm.compiler.lir.LIRInstructionClass;
+import org.graalvm.compiler.lir.Opcode;
+import org.graalvm.compiler.lir.amd64.AMD64LIRInstruction;
+import org.graalvm.compiler.lir.asm.CompilationResultBuilder;
+
+import jdk.vm.ci.meta.AllocatableValue;
+
+public class AMD64VexBinary {
```
mur47x111: Looks identical to `AMD64VectorBinary.AVXBinaryOp`. How about renaming the original?

jphalimi-intel: It looks almost identical, although I had to use a VexGeneralPurposeRVMOp (extends VexOp) opcode because ANDN is an instance of VexGeneralPurposeRVMOp, and AMD64VectorBinary.AVXBinaryOp uses VexRVMOp (extends VexOp). Both extend the same base class, but they are incompatible types.

From what I understand, two solutions: either modify AMD64VectorBinary.AVXBinaryOp to handle a generic VexOp, or create this new class for General Purpose RVM Ops. What do you think?

jphalimi-intel:
Following our conversation I have fixed the commit:

- It is now complete (it lacked other modifications)
- it includes blsi pattern matching. I had to add it in the Assembler as well.
- Andn emiter uses AMD64VectorBinary.AVXBinaryOp, it required VexGeneralPurposeRVMOp to extend VexRVMOp instead of VexOp for types to be compatible. Consequently I removed inherited code from VexRVMOp class.

Please let me know your thoughts.

mur47x111:
Having VexGeneralPurposeRVMOp extending VexRVMOp looks good. But the code in `VexGeneralPurposeRVMOp` is slightly different from the one in  `VexRVMOp`, e.g., assertion, REX.W bit. We still want to keep that.

mur47x111:
@jphalimi-intel Finally I got some time to look into the blsi issue. This is because BLSI has a special encoding `VEX.NDD.LZ.0F38.W0 F3 /3`. We cannot simply reuse `VexRMOp.emit`, because it has no idea about either `/3` or dst operand being stored in `VEX.vvvv`.

I would recommend either creating the special LIRInstruction at `AMD64Unary`, then calls into `AMD64Assembler.blsi()`, or a special `VexRMOp`.

jphalimi-intel:
My apologies for the delayed answer. I have updated the patch to add blsr and blsmsk. Funny story, blsr/blsmsk pattern had to match Add -1 instead of Sub 1.

Aside from that, I have overriden the emit methods for the RMOp to fix the ModRM. It looks like the code does not crash anymore, but the result is still incorrect. I am unsure why exactly.

@Yudi: what are your thoughts?

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -1392,6 +1394,36 @@ public void emit(AMD64Assembler asm, AVXSize size, Register dst, AMD64Address sr
...
+
+        private VexGeneralPurposeRMOp(String opcode, int pp, int mmmmm, int w, int op, int ext, VEXOpAssertion assertion) {
+            super(opcode, pp, mmmmm, w, op, assertion);
+            this.ext = ext;
+        }
+
+        @Override
+        public void emit(AMD64Assembler asm, AVXSize size, Register dst, Register src) {
+            assert assertion.check((AMD64) asm.target.arch, size, dst, null, null);
+            asm.vexPrefix(dst, Register.None, src, size, pp, mmmmm, size == AVXSize.DWORD ? W0 : W1);
```
mur47x111: We should pass `dst` as the second parameter, i.e.,
```
asm.vexPrefix(AMD64.rax, dst, src, size, pp, mmmmm, size == AVXSize.DWORD ? W0 : W1); 
```
Here `AMD64.rax` is a bit misleading, we should create a special place holder similar to `Register.None`, but has the encoding of 0 instead of -1. (I would suggest `Register.Ignore`.)

jphalimi-intel:
Good catch! I have updated the commit. The incorrect calculations are gone. I will now look into the testing part. Please let me know if there is anything else you'd like to edit.

jphalimi-intel:
I looked over the BMI tests you referred to (JDK-8198574). It looks like they check native code generation after compiling. From what I can see in Graal, there is no logging done nor infrastructure to check LIR transformations done in LIRGenPhase. Thus we would need to port the native code checker from OpenJDK. Can you confirm that this work is necessary before merging?

jphalimi-intel:
I have updated the patch with the testing part as I managed to easily port the tests from OpenJDK. I added a base class for BMI since they all share a native code checker. The tests goals are twofold:
1. Check a basic result correctness
2. Ensure that the compiled binary matches with the tested instruction (to make sure that the LIR generator matched the desired pattern)

The testing highlighted a missing check in the Matching pattern, namely for blsr and blsmsk, we must ensure that we substract 1 to the input, but we did not check the type of the constant, which caused an assert to fail. I fixed that as well.

I think the code is ready for final review.

mur47x111:
Thank you @jphalimi-intel ! I will integrate this.

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/BmiAndn.java
```diff
@@ -0,0 +1,105 @@
...
+
+    // Pattern matching check for andnl
+    @Test
+    public void test2() {
+        ResolvedJavaMethod method = getResolvedJavaMethod("andnl");
+        StructuredGraph graph = parseForCompile(method);
+
+        CompilationResult c = compile(method, graph);
+        byte[] targetCode = c.getTargetCode();
+        int cnt = (getTarget().arch instanceof AMD64) ? 1 : 2;
```
mur47x111: What is the intention of this line? I think the test case should be AMD64 specific, right?

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -1353,6 +1354,7 @@ public void emit(AMD64Assembler asm, AVXSize size, Register dst, Register src1,
             asm.emitModRM(dst, src2);
```
mur47x111: Regarding the caller of this method, we have to make sure the ordering of `src1` and `src2`. I think at `AMD64NodeMatchRules.logicalAndNot` we swap the operand ordering and result in the wrong output.

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/BmiAndn.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.graalvm.compiler.replacements.test;
```
mur47x111: We should move these tests to `org.graalvm.compiler.hotspot.amd64.test`

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/BmiAndn.java
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
```
mur47x111: 2018 please ;)

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/BmiAndn.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    // Result correctness check
+    @Test
+    public void test3() {
+        int n1 = 42;
+        int n2 = 100;
+        int result = andni(n1, n2);
+
+        Assert.assertEquals(result, 10);
```
mur47x111: Here we are testing the interpreter output. We could use
```
test("andni", 42, 100);
```
And you will detect the operand ordering issue.

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/BmiBlsmsk.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.Test;
+
+import jdk.vm.ci.meta.ResolvedJavaMethod;
+import org.junit.Assert;
+
+/**
+ * Tests Bit Manipulation Instruction blsmsk pattern matching and result.
+ */
+public class BmiBlsmsk extends BmiCompilerTest {
+    // from Intel manual VEX.NDD.LZ.0F38.W0 F3 /3, example c4e260f2c2
```
mur47x111: It should be /2

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/BmiBlsmsk.java
```diff
@@ -0,0 +1,105 @@
...
+                    (byte) 0x1F,
+                    (byte) 0x00,
+                    (byte) 0xFF,
+                    (byte) 0b0011_1000};
+    private final byte[] instrPattern = new byte[]{
+                    (byte) 0xC4, // prefix for 3-byte VEX instruction
+                    (byte) 0x02, // 00010 implied 0F 38 leading opcode bytes
+                    (byte) 0x00,
+                    (byte) 0xF3,
+                    (byte) 0b0001_0000}; // bits 543 == 011 (3)
```
mur47x111: It should be 010 (/2)

compiler/src/org.graalvm.compiler.replacements.test/src/org/graalvm/compiler/replacements/test/BmiBlsr.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.graalvm.compiler.replacements.test;
+
+import org.graalvm.compiler.code.CompilationResult;
+import org.graalvm.compiler.nodes.StructuredGraph;
+import org.graalvm.compiler.replacements.test.BmiCompilerTest;
```
mur47x111: Please remove the unused/unnecessary import.

jphalimi-intel:
Thanks for the review. I resolved all mentioned comments. I also pacified CheckStyle as some of my changes were not following the code style. It should be all good by now.

mur47x111:
Thanks @jphalimi-intel ! I will run the internal gates again.

mur47x111:
@jphalimi-intel I limited the optimizations when BMI1 is supported (see 73d2560300005334f22537aa44593c6288f30945). Also I learned that our node matching machinery understands commutativity, so I removed the redundant rules (see 90518ab7411f8ed73894108b3cb120935e963a5b). Thanks again for the contribution!

jphalimi-intel:
Great job, thanks for finalizing.

dougxc:
Hi @jphalimi-intel , have you had a chance to measure the performance impact of this change on any benchmarks? Are there any results you can share here?

jphalimi-intel:
Hi @dougxc,

I have ran the crypto SPECjvm08 benchmarks on one of our servers, I can see **200** andn applications throughout the JITted code. Performance is slightly increased as follows:

crypto.aes **+2.28%**
crypto.signverify **+2.23%**

I know that you are the original owner of JDK-8198574, so if you had other specific benchmarks in mind for us to try out, let me know.

dougxc:
Thanks for sharing the positive results! That's all I need for now.

PR #681, opened 2018-09-19
Update README with license move from GPL to UPL

jhult:
As part of [[GR-11502] Relicense Truffle and the SDK to UPL](/oracle/graal/commit/0f8c00b5241ec89154a79ac6e7627dbfbaf4dce6).


jhult:
I have already signed the OCA and my name is already listed on that page.

PR #692, opened 2018-09-25
Service loader and logging

tomas-langer:
Removed some logging substitutions as reflection is now supported.
Added ServiceLoader automatic feature to register service loader resource files and to register service implementation files for reflection.


christianwimmer:
Merged in https://github.com/oracle/graal/commit/84dcb4e9ad49ae749c88ee612ba5dc95ab171bd8

PR #707, opened 2018-10-03
Add support for LogRecord caller inference

dmlloyd:
Use the existing stack trace support in Throwable to infer callers for JUL.


dmlloyd:
Note that this could use a special stack walker instead.  But, this is probably good enough, as it is what logging frameworks have done for years.

cstancu:
@dmlloyd https://github.com/oracle/graal/commit/84dcb4e9ad49ae749c88ee612ba5dc95ab171bd8  [completely removes the `Target_java_util_logging_LogRecord.inferCaller()` substitution](https://github.com/oracle/graal/commit/84dcb4e9ad49ae749c88ee612ba5dc95ab171bd8#diff-eadc8ead3c22189829c4759e5067c8c4L133). Many things have been fixed/improved since we introduced that substitution and it seems that it is not needed anymore. Can you check if that works for you? Thanks!

dmlloyd:
Great!  It's a little odd that this seemingly unrelated change was made in a commit that relates to service loader, but that's a different conversation :)

I'll close this PR.  Thanks!

PR #724, opened 2018-10-07
Fix typos in class initialization messages

taylorwood:
I noticed some misspellings of initialization and reinitialized while debugging.


taylorwood:
@graalvmbot I signed/submitted OCA after creating this PR.

PR #734, opened 2018-10-12
Java 9+ StackWalker initial implementation

dmlloyd:


dmlloyd:
Grain of salt: I haven't been able to successfully test this!  But, I wanted to put it here for comment.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
```diff
@@ -722,6 +745,348 @@ static Package getPackage(Class<?> c) {
...
+        this.sp = sp;
+        this.ip = ip;
+        this.anchor = anchor;
+    }
+
+    public boolean hasNext() {
+        checkThread();
+        while (next == null) {
+            while (nested == null) {
+                while (anchor.isNonNull() && anchor.getLastJavaSP().belowOrEqual(sp)) {
```
christianwimmer: Why do you need to duplicate the code from `JavaStackWalker` here? `JavaStackWalker` is already a generic mechanism where you can plug in any stack frame visitor you want to write.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
```diff
@@ -722,6 +745,348 @@ static Package getPackage(Class<?> c) {
...
+
+        int getLineNumber();
+
+        boolean isNativeMethod();
+
+        StackTraceElement toStackTraceElement();
+    }
+
+    @TargetClass(className = "java.lang.StackWalker$Option", onlyWith = JDK9OrLater.class)
+    enum Option {
```
christianwimmer: Use a regular class `final class Target_java_lang_StackWalker_Option` here. The `@Alias` fields are really just an alias to the `static` fields defined in the original class. You cannot use the `enum` methods or the fact that the class is an enum anyway without bad things happening.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
dmlloyd: Because the access pattern of the JDK `StackWalker` is fundamentally incompatible with this model from a performance standpoint.  It does not call back to user code for each frame; rather, the user can draw out each frame via a Stream or Iterator.  If I used the `JavaStackWalker` approach, I would have to iterate the entire stack into an array or list, and then stream that to the user instead, or else do something that would result in a similar performance profile.  This would be unfortunate, because this stack walk otherwise has potential to be quite a bit more performant than HotSpot's one.

If there was a clean way to reuse code between the two, I'd be willing to do it, but I don't see such a way.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
```diff
@@ -722,6 +745,348 @@ static Package getPackage(Class<?> c) {
     }
 }
 
+@TargetClass(className = "java.lang.StackWalker", onlyWith = JDK9OrLater.class)
```
christianwimmer: All JDK 9 specific substitutions must be in the `com.oracle.svm.core.jdk9` project. Then you can also use, e.g., `StackWalker.StackFrame` directly without a target class - the substitutions are then compiled with JDK 9 (or actually 11)

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
christianwimmer: OK, then we need a bigger refactoring of our stack walking code to avoid code duplication.

dmlloyd: OK.  I have... maybe... an idea.  I'll give it a try and see how it goes.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
dmlloyd: Will do.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
dmlloyd: OK.

dmlloyd:
OK.  I've removed the duplication by creating a new JavaStackWalk API which uses stack value allocation to have (hopefully) a similar overhead profile to the original code.  The new API is used by both JavaStackWalker and the j.l.StackWalker implementation.

dmlloyd:
Resolved conflicts.

dmlloyd:
One thing that I think could possibly be interesting to explore would be the effect of inlining in combination with `getCallerClass()`.  It is interesting because in this case the caller is always known, thus it could possibly be constant-folded.

substratevm/src/com.oracle.svm.core.jdk9/src/com/oracle/svm/core/jdk/Target_java_lang_StackWalker.java
```diff
@@ -25,19 +25,340 @@
...
+        return frameInfo.isNativeMethod();
+    }
+
+    public StackTraceElement toStackTraceElement() {
+        StackTraceElement ste = this.ste;
+        if (ste == null) {
+            this.ste = ste = frameInfo.getSourceReference();
+        }
+        return ste;
+    }
```
Peter-B-Kessler: Is there an explanation of this dance with a local `ste` versus the instance `this.ste`?

Is there more here than caching the result of `frameInfo.getSourceReference()`?  Why not just cache it in the instance field?  Atomicity does not seem to be an issue.

dmlloyd: It's a very common pattern for lazy caching.

Peter-B-Kessler: Okay.  It only came to my attention because our style checker complained about the local variable having the same name as the instance field.  I can deal with that.  I just wanted to know if there was something other than lazy caching going on.

Peter-B-Kessler:
@dmlloyd [JavaStackWalk.java](https://github.com/oracle/graal/pull/734/files#diff-eac1d767d93a32ecb193ebb4332c6d7a) does not have any copyright statement on it.  Is it okay with you if I add the standard Oracle copyright and GPLv2 license to it?

dmlloyd:
Yep that's okay with me.

dmlloyd:
This PR has been rebased against upstream master.  I also have two commits on the side that incorporate the additional changes @johanvos authored that could be pushed up to #1213 to replace that PR.

Changes from the last revision include supporting calling class acquisition, and adapting to more recent API changes around frame info queries.

/cc @christianwimmer and @dougxc

dmlloyd:
Looks like it's failing on formatting.  I'll get into that today and push up the fix.

dmlloyd:
Formatting should be fixed now.

johanvos:
Do you plan to add support for RETAIN_CLASS_REFERENCE?

dmlloyd:
It should already be there in the latest version (basically the flag is ignored and it's always supported, since there's no security manager anyway).  Though I see CI is unhappy again so I'll have a look there.

dmlloyd:
Looks like more formatting difficulties.  Fixed now.  Really need to check that IDEA is using the correct formatter configuration...

christianwimmer:
Thanks for updating the branch! I'm running our internal tests on it now. Please don't amend or rebase your commits anymore, instead add new commits if you have changes. That avoids conflicts on my side.

dmlloyd:
Okay will do. CI is griping about another problem which I can try to reproduce tomorrow.

christianwimmer:
The refactoring of `JavaStackWalker` was obviously broken (even gathering normal exception stack traces did not work). I fixed that.

Need to have a look through the `StackWalker` substitutions tomorrow, then we'll merge the changes.

No need for you to push changes / fixes to this PR. I'll get back to you if I have questions about the `StackWalker` substitutions.

dmlloyd:
OK feel free to push additional commits on to this branch.  I'd be interested to see what went wrong as I'm fairly sure the code did work at some point (but was rebased numerous times since then, and I didn't have direct access to my one machine capable of actually building this for a few days).

johanvos:
When running this:
```
StackWalker sw = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE);
System.err.println("CC = "+sw.getCallerClass());
```
I get the exception:
`java.lang.IllegalStateException: No calling frame`

But maybe this is due to the refactory of JavaStackWalker which @christianwimmer mentioned?

dmlloyd:
Yeah I think I rebased wrongly and mixed something up.

bobmcwhirter:
"I'm running our internal tests on it now."

@christianwimmer -- Any chance those tests could be integrated into the public community repository, so that all of us can help share the load of validating PRs before trying up Oracle's time and efforts?

substratevm/src/com.oracle.svm.core.jdk9/src/com/oracle/svm/core/jdk/Target_java_lang_StackWalker.java
```diff
@@ -25,19 +25,367 @@
...
+        // TODO: ConstructorAccessor.class.isAssignableFrom(c)
+        // TODO: c.getName().startsWith("java.lang.invoke.LambdaForm")
+    }
+
+    static boolean isHidden(final String nextClassName) {
+        return ImplicitExceptions.class.getName().equals(nextClassName);
+    }
+}
+
+final class StackIterator implements Iterator<StackFrame> {
```
christianwimmer: Implementing this as an `Iterator` is overly complicated, because of the `hasNext` - `next` separation. I converted it to a `Spliterator` and now the whole implementation is in a single method.

dmlloyd: Sounds good.  I only went with `Iterator` as it was something I'm very familiar with.

johanvos:
My sample (StackWalker.getCallerClass()) still doesn't work, as 
`JavaStackWalker.findNextFrame` returns false when finding the first frame.

dmlloyd:
> My sample (StackWalker.getCallerClass()) still doesn't work, as
> `JavaStackWalker.findNextFrame` returns false when finding the first frame.

It looks like @christianwimmer didn't push up his version.  Christian, could you (at some point) please push your changes into this branch?  (You should be able to push it directly to https://github.com/dmlloyd/graal/tree/stack-walker by virtue of this pull request).

christianwimmer:
Pushing to the GitHub PR didn't work for me. But the changes were ready to be merged anyway, so I just initiated the merge to master. Note that I did some refactorings to avoid code duplication too, e.g., `getCallerClass` is the same code as `Reflection.getCallerClass` that already existed.

Regarding the tests: Many tests that we have check the integration with Oracle products, e.g., the Oracle MLE effort to integrate with the Oracle database. Therefore, most of our existing tests cannot be opened. We certainly want to expand our open testing, e.g., with Quarkus or Micronaut. If you can contribute in that area, please coordinate with @cstancu.

PR #738, opened 2018-10-15
Add "addSubShift" rule for AArch64.

XiaohongGong:


XiaohongGong:
Hi, I'm an employee from Arm and I have signed the OCA. Can you please check and add my email to the oca list? Thank you!

XiaohongGong:
Hi all,
Could you please take a review to my patch? Thank you!

XiaohongGong:
@sanzinger Could you please take a look at this patch? Thanks!

XiaohongGong:
@sanzinger could you please take a look at this patch? Thank you!

XiaohongGong:
@sanzinger could you please take a look at this patch? Thank you!

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
```diff
@@ -387,6 +387,8 @@ public void doBlock(Block block, StructuredGraph graph, BlockMap<List<Node>> blo
                         debug.log("interior match for %s", valueNode);
                     } else if (operand instanceof ComplexMatchValue) {
                         debug.log("complex match for %s", valueNode);
+                        // Set current position to the position of the root matched node.
+                        setSourcePosition(node.getNodeSourcePosition());
```
sanzinger: Please move this up to `if (node instanceof ValueNode) {` and remove the call in `emitNode` so it is used commonly for all cases.

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64AddSubShiftTest.java
```diff
@@ -0,0 +1,187 @@
+/*
+ * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, Arm Limited and affiliates. All rights reserved.
```
sanzinger: Our internal verification points out, that this copyright header is incorrect. At the moment I am waiting for input what we should do about it. I'll follow up on this later.

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64MatchRuleTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.graalvm.compiler.lir.phases.LIRPhase;
+import org.graalvm.compiler.lir.phases.LIRSuites;
+import org.graalvm.compiler.lir.phases.PreAllocationOptimizationPhase;
+import org.graalvm.compiler.options.OptionValues;
+import org.junit.Assert;
+import org.junit.Before;
+
+import static org.junit.Assume.assumeTrue;
+
+public abstract class AArch64MatchRuleTest extends GraalCompilerTest {
```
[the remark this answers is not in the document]
XiaohongGong: If making `MatchRuleTest` abstract, I think it shoud be removed to a general package. Is the "org.graalvm.compiler.core.test" all right?

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
XiaohongGong: Thanks for your suggestion! But `emitNode ` will not be called when "if (operand instanceof ComplexMatchValue)" is tue, and it is just called in `doRoot`, which is in the general branch. Also I have checked that there is already a "setSourcePosition(node.getNodeSourcePosition());" calling in `emitNode `. So is there any other better place to remove the function? Thanks!

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64AddSubShiftTest.java
XiaohongGong: Thanks for your reviewing! And waiting for your correct copyright!

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
XiaohongGong: Sorry, I just misunderstand your advise! So I will move this function call to `if (node instanceof ValueNode) {` and remove the call in `emitNode`. Thank you!

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64AddSubShiftTest.java
XiaohongGong: Hi sanzinger,
Is the correct copyright ok? Thanks！

XiaohongGong:
@sanzinger I have resolved the comments except for the copyright, and the new patch is updated. Can you please take a look at it? Thank you!

XiaohongGong:
@sanzinger Is there any conclusion about the correct copyright? I'm a little worried because there will be many optimizing PRs afterwards which all depends on this patch. This patch contains the basic match rule tests for AArch64. Waiting for your answer. Thank you!

dougxc:
Hi @XiaohongGong , sorry for the delay. We're trying to reach a conclusion as quickly as possible.

XiaohongGong:
@dougxc Thank you so much!

sanzinger:
We can keep the copyright header. I'm going to integrate this change.

XiaohongGong:
@sanzinger thank you so much!

XiaohongGong:
Hi @sanzinger, thanks for your reviewing. But is there any problems exist when integrating this change? I mean whether the merging job is conflict or any other else problems exist that need my help. Thank you!

XiaohongGong:
Hi @sanzinger , I have checked the latest codes in the master branch, but I have found that it is not the latest for this patch. The patch merged is the one I first committed. The newest version includes some fixed codes according to your comment. And I think they are all valuable. Is there any method to revert this merging and merge the right version? If not, I will commit a new patch which will include all these changes. Thank you!

XiaohongGong:
Hi @sanzinger , I have committed another path https://github.com/oracle/graal/pull/781 which is a part of this patch. Could you please take a look at it? Thank you!

sanzinger:
Thank you for your contributions. It seems that merged PR #781 contains all changes from  this PR. Hence closing.

PR #743, opened 2018-10-17
Consult sched_getaffinity in Linux-specific substitution for more accurate CPU count

dmlloyd:


substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
```diff
@@ -66,6 +72,48 @@ public static String mapLibraryName(String libname) {
...
 }
 
+@Platforms(Platform.LINUX.class)
+@TargetClass(java.lang.Runtime.class)
+final class Target_java_lang_Runtime {
+
+    @Substitute
+    public int availableProcessors() {
+        if (SubstrateOptions.MultiThreaded.getValue()) {
+            int res = (int) Unistd.sysconf(Unistd._SC_NPROCESSORS_CONF());
```
fweimer: This is not the right CPU count to check. You need to obtain the number of *possible* CPUs, which is the upper limit the system can support after CPU hotplug. The easiest way to do this is to call `sched_getaffinity` repeatedly with growing buffer sizes, unfortunately.

dmlloyd: Fair enough.  Fix forthcoming...

dmlloyd: Updated.

dmlloyd:
Added check of result of calloc.

dmlloyd:
Fix copypasta error

dmlloyd:
After this PR, I intend to look into using cgroups information for a more accurate picture when running inside a container; I will most likely replicate the OpenJDK logic in this area.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
fweimer: By the way, I think you should be able to test this with `possible_cpus=5000` or similar on the kernel command line, and a Red Hat Enterprise Linux or CentOS kernel (Debian won't work because the hard limit is 512 CPUs). No need to find a system with a broken firmware that says it can be hot-plugged to these many CPUs.

dmlloyd:
@fweimer does the new change look OK?  You've left a big X on this PR :)

fweimer:
@dmlloyd I don't see any X?

Is this `availableProcessors` function used for disabling SMP support on uniprocessors?

dmlloyd:
@fweimer Under "reviewers" it still says "fweimer requested changes".

I'm not sure if SubstrateVM uses it for that purpose.  It doesn't seem to do so.  I think most uniprocessor optimizations would happen at compile time, but the CPU count is never known until run time.

fweimer:
@dmlloyd That's very odd, it doesn't show this for me anywhere I can see. All I see in a hover text box is “Left a review”.

If it's not used for optimizations, than it's okay. We actually had to change the return value of our processor count implementation from 1 to 2 if `/proc` is not mounted because it broke Hotspot (it incorrectly assumed a UP system). Using the affinity mask has a similar issue because it can changed at run time, even from outside of the process.  (And of course there's also the matter of process migration.)

dmlloyd:
It really sounds like relying on availableProcessors for that kind of decision is probably erroneous.  Maybe there's a need for a maximumPossibleProcessors thing as well, or something.

Peter-B-Kessler:
@dmlloyd > After this PR, I intend to look into using cgroups information for a more accurate picture when running inside a container; I will most likely replicate the OpenJDK logic in this area.

I am about to push changes to look at the cgroup information for physical memory.  Using the hacky JDK-8 logic.  Having the more complete OpenJDK logic would be nice, but not if it has a significant impact on startup instruction count or time for things like "HelloWorld".  (I used `perf stat -e instructions` and `perf stat -e instructions:u` but maybe that isn't the best tool for the job.)

Among the problems I ran into with the memory sizing was that some code that wanted the memory size were in places that could not allocate, which made opening files and parsing numbers ... awkward.  I don't think you will run into those with the other cgroup parameters, but forewarned is four-armed.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/linux/LinuxSched.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (c) 2013, 2017, Oracle and/or its affiliates. All rights reserved.
```
Peter-B-Kessler: It's 2018.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/linux/LinuxSched.java
```diff
@@ -0,0 +1,59 @@
...
+import org.graalvm.nativeimage.c.CContext;
+import org.graalvm.nativeimage.c.function.CFunction;
+import org.graalvm.nativeimage.c.function.CMacroInfo;
+import org.graalvm.nativeimage.c.struct.CStruct;
+import org.graalvm.word.PointerBase;
+
+/**
+ */
+@CContext(PosixDirectives.class)
+@Platforms({ Platform.LINUX.class})
```
Peter-B-Kessler: s/{ P/{P/

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/linux/LinuxSched.java
```diff
@@ -0,0 +1,59 @@
...
+public class LinuxSched {
+
+    @CFunction
+    public static native int sched_setaffinity(int pid, int cpu_set_size, cpu_set_t set_ptr);
+
+    @CFunction
+    public static native int sched_getaffinity(int pid, int cpu_set_size, cpu_set_t set_ptr);
+
+    @CFunction
+    static native int __sched_cpucount(int cpu_set_size, cpu_set_t set_ptr);
```
Peter-B-Kessler: I guess this function is declared in <sched.h>, but it sure looks weird to be using a function that doesn't have a manual page entry.  I think you only need it for the implementation of the {{CPU_COUNT_S}} macro below.  But it seems worth a comment.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
```diff
@@ -66,6 +72,46 @@ public static String mapLibraryName(String libname) {
...
+@TargetClass(java.lang.Runtime.class)
+final class Target_java_lang_Runtime {
+
+    @Substitute
+    public int availableProcessors() {
+        if (SubstrateOptions.MultiThreaded.getValue()) {
+            LinuxSched.cpu_set_t cpuSet;
+            int size, res;
+            final int pid = Unistd.getpid();
+            // try easy path
```
Peter-B-Kessler: What makes this path "easy"?  That you get the answer in one `cpu_set_t`?

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
```diff
@@ -66,6 +72,46 @@ public static String mapLibraryName(String libname) {
...
 
+@Platforms(Platform.LINUX.class)
+@TargetClass(java.lang.Runtime.class)
+final class Target_java_lang_Runtime {
+
+    @Substitute
+    public int availableProcessors() {
+        if (SubstrateOptions.MultiThreaded.getValue()) {
+            LinuxSched.cpu_set_t cpuSet;
+            int size, res;
```
Peter-B-Kessler: Are you copying some C code?  Is there a reason not to declare `cpuSet`, `size`, and `res` where they are first needed?  (That would also declare them one per line. :-)

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
```diff
@@ -66,6 +72,46 @@ public static String mapLibraryName(String libname) {
...
+            final int pid = Unistd.getpid();
```
Peter-B-Kessler: The manual page for `sched_getaffinity` says "If pid is zero, then the mask of the calling process is returned.", which would save you explicitly getting your pid.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
```diff
@@ -66,6 +72,46 @@ public static String mapLibraryName(String libname) {
...
+            cpuSet = StackValue.get(LinuxSched.cpu_set_t.class);
+            size = SizeOf.get(LinuxSched.cpu_set_t.class);
+            res = LinuxSched.sched_getaffinity(pid, size, cpuSet);
+            if (res == 0) return LinuxSched.CPU_COUNT_S(size, cpuSet);
+            // try with more CPUs in a loop
+            size = Integer.highestOneBit(size) << 1;
+            while (Integer.numberOfTrailingZeros(size) < 16) { // we have to give up at *some* point
+                assert Integer.bitCount(size) == 1;
+                cpuSet = calloc(size); // to be safe
+                if (cpuSet.isNull()) throw new InternalError("Cannot determine CPU count");
```
Peter-B-Kessler: Missing `{}`s around the then-clause.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
```diff
@@ -66,6 +72,46 @@ public static String mapLibraryName(String libname) {
...
+                try {
+                    res = LinuxSched.sched_getaffinity(pid, size, cpuSet);
+                    if (res == 0) return LinuxSched.CPU_COUNT_S(size, cpuSet);
```
Peter-B-Kessler: Missing `{}`s around the then-clause.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
```diff
@@ -27,10 +27,16 @@
 import static com.oracle.svm.core.posix.headers.Time.gettimeofday;
 import static com.oracle.svm.core.posix.headers.linux.LinuxTime.CLOCK_MONOTONIC;
 import static com.oracle.svm.core.posix.headers.linux.LinuxTime.clock_gettime;
+import static org.graalvm.nativeimage.UnmanagedMemory.calloc;
+import static org.graalvm.nativeimage.UnmanagedMemory.free;
```
Peter-B-Kessler: We have changed styles and are now trying not to add more `import static` statements.  Also, since this is in POSIX code, you can just use `Libc.calloc` and `Libc.free`.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
```diff
@@ -66,6 +72,46 @@ public static String mapLibraryName(String libname) {
...
+            final int pid = Unistd.getpid();
+            // try easy path
+            cpuSet = StackValue.get(LinuxSched.cpu_set_t.class);
+            size = SizeOf.get(LinuxSched.cpu_set_t.class);
+            res = LinuxSched.sched_getaffinity(pid, size, cpuSet);
+            if (res == 0) return LinuxSched.CPU_COUNT_S(size, cpuSet);
+            // try with more CPUs in a loop
+            size = Integer.highestOneBit(size) << 1;
+            while (Integer.numberOfTrailingZeros(size) < 16) { // we have to give up at *some* point
+                assert Integer.bitCount(size) == 1;
```
Peter-B-Kessler: Why do you need this assert?  Line 91 sets `size` to have exactly 1 bit set, and line 102 shifts that bit but does not add any new bits.  And you give up if there are 16 trailing zeroes, so I don't think line 102 can shift the bit off into the sign bit, for reasonable values of `__CPU_SETSIZE` and `__NCPUBITS`.  And if the bit does get shifted into the sign bit, the assert is not going to help.  Maybe the while-loop wants to give up when `size` gets to be more than you are willing to `calloc`.  That way you can have a nicely-named `final int` as the limit instead of a `16`.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java
```diff
@@ -66,6 +72,46 @@ public static String mapLibraryName(String libname) {
...
+                try {
+                    res = LinuxSched.sched_getaffinity(pid, size, cpuSet);
+                    if (res == 0) return LinuxSched.CPU_COUNT_S(size, cpuSet);
+                } finally {
+                    free(cpuSet);
+                }
+                size <<= 1;
+            }
+            // give up
+            res = (int) Unistd.sysconf(Unistd._SC_NPROCESSORS_ONLN());
```
Peter-B-Kessler: Maybe break this into 2 methods: one to get the number of processors from `sched_getaffinity` and one to get the number of processors from `Unistd.sysconf` (and then one that gets the number of processors from the cgroup :-), with a method to call them in some appropriate order.

dmlloyd:
I don't have any further time to spend on this until probably after the holidays or later.  Feel free to adjust the formatting to your liking.  I think you should be able to push to/amend my branch however you want.

dmlloyd:
This has been rebased and conflicts resolved.  Could I please get a new review?

cstancu:
@dmlloyd this will be merged soon. Thank you!

PR #781, opened 2018-11-07
Refactor match rule tests for AArch64.

XiaohongGong:

PR #788, opened 2018-11-08
Add MNeg rule for AArch64

armLeiJin:
The patch replies on another one under review https://github.com/oracle/graal/pull/781

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64MNegTest.java
+++ b/compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64MNegTest.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, Arm Limited and affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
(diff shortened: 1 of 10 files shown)


dougxc:
Hi @armLeiJin , thanks for the submission. Can you please use your company email address for the commits so that we can associate you with an OCA signatory.

armLeiJin:
> Hi @armLeiJin , thanks for the submission. Can you please use your company email address for the commits so that we can associate you with an OCA signatory.

Thanks for replying to me. I am now using the  email address "Lei.Jin@arm.com", which is my company email address. Could you see it? Or what you see is something else?

armLeiJin:
@sanzinger Could you help to review the patch? Thanks

sanzinger:
@armLeiJin: I am going to resolve the merge issues.

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64MNegTest.java
```diff
@@ -0,0 +1,143 @@
...
+    private static long mulSubZeroLong(long x, long y) {
+        return x * (y - 0);
+    }
+
+    public static long multiplyNegateLong(long x, long y) {
+        return mulNegLong(x, y) + mulZeroSubLong(y, x) + mulSubZeroLong(x, y);
+    }
+
+    @Test
+    public void testMultiplyNegateLong() {
```
sanzinger: After merging 77d0221b486eac09756a26144632ee04fc3afc67, this test fails because the second MNEG is concurring with the MADD rule. I am updating the test accordingly.

sanzinger: I would even say, just to test the single case with `mulNegLong`, since we assume that the compiler can optimize `(0-y)` to `-y`.

sanzinger:
I have fixed the failing test by using `|` operator instead of `+`.

armLeiJin:
> I have fixed the failing test by using `|` operator instead of `+`.
> 
> I am going to integrate this change now.

Thank you very much for doing so!

PR #791, opened 2018-11-09
Add "logicShift" match rule for AArch64.

XiaohongGong:
This patch depends on PR https://github.com/oracle/graal/pull/781.

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
+++ b/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
@@ -25,12 +25,19 @@
 
 package org.graalvm.compiler.core.aarch64;
 
+import jdk.vm.ci.meta.AllocatableValue;
 import jdk.vm.ci.meta.JavaKind;
+import jdk.vm.ci.meta.Value;
+import org.graalvm.collections.EconomicMap;
+import org.graalvm.collections.Equivalence;
 import org.graalvm.compiler.asm.aarch64.AArch64MacroAssembler;
+import org.graalvm.compiler.core.common.LIRKind;
 import org.graalvm.compiler.core.gen.NodeMatchRules;
 import org.graalvm.compiler.core.match.ComplexMatchResult;
 import org.graalvm.compiler.core.match.MatchRule;
+import org.graalvm.compiler.graph.Node;
 import org.graalvm.compiler.lir.LIRFrameState;
+import org.graalvm.compiler.lir.Variable;
 import org.graalvm.compiler.lir.aarch64.AArch64ArithmeticOp;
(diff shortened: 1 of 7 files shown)


XiaohongGong:
@dougxc , could you please take a look at this path? Thank you!

XiaohongGong:
@tkrodriguez so, how is the reviewing going now. Is there anything update? Thanks!

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
```diff
@@ -25,28 +25,58 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
 public class AArch64NodeMatchRules extends NodeMatchRules {
+    private static final Map<Class<? extends Node>, AArch64ArithmeticOp> nodeOpMap;
+
+    private static final Map<Class<? extends BinaryNode>, AArch64MacroAssembler.ShiftType> shiftTypeMap;
+
+    static {
+        nodeOpMap = new HashMap<>();
```
sanzinger: As a general rule of thumb, we want to avoid usages of `java.util.HashMap` in Graal. Instead we have our own Map implementation which is created with`org.graalvm.collections.EconomicMap.create(Equivalence.IDENTITY, 5)`. Since the element count is below 8, the EconomicMap implementation does a linear search over the entities, which is pretty similar to a `if/elseif/else` cascade.

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
```diff
@@ -87,10 +123,37 @@ private ComplexMatchResult emitAddSubShift(AArch64ArithmeticOp op, ValueNode val
...
+    @MatchRule("(Or=binary a (Not (UnsignedRightShift=shift b Constant)))")
+    @MatchRule("(Xor=binary a (Not (LeftShift=shift b Constant)))")
+    @MatchRule("(Xor=binary a (Not (RightShift=shift b Constant)))")
+    @MatchRule("(Xor=binary a (Not (UnsignedRightShift=shift b Constant)))")
+    public ComplexMatchResult logicShift(BinaryNode binary, ValueNode a, BinaryNode shift) {
+        AArch64ArithmeticOp op = nodeOpMap.get(binary.getClass());
+        assert op != null;
+        ValueNode operand = binary.getX() == a ? binary.getY() : binary.getX();
+        if (operand instanceof NotNode) {
+            return emitBinaryShift(op, a, shift, true);
```
sanzinger: Can we change this to:
```
boolean isShiftNot = operand instanceof NotNode;
return emitBinaryShift(op, a, shift, isShiftNot);
```

XiaohongGong:
Thank you so much!

PR #800, opened 2018-11-11
Add function to select the right lib folder

JornVernee:
When running ~`mx makegraaljdk build`~ `mx gate` while using JDK 11 in the compiler suite I was seeing an error about the file `/lib/jdwp.dll` not existing. That is indeed the case, because the file is under `/bin` on windows.

This PR adds a function to return the right lib folder based on the platform, and uses that instead of the hardcoded `lib` directory.

### DIFF
--- a/compiler/mx.compiler/mx_compiler.py
+++ b/compiler/mx.compiler/mx_compiler.py
@@ -973,9 +973,10 @@ def java_base_unittest(args):
         shutil.rmtree(basejdk_dir)
     mx.run([jlink, '--output', basejdk_dir, '--add-modules', basemodules, '--module-path', join(jdk.home, 'jmods')])
     jdwp = mx.add_lib_suffix(mx.add_lib_prefix('jdwp'))
-    shutil.copy(join(jdk.home, 'lib', jdwp), join(basejdk_dir, 'lib', jdwp))
+    lib_folder = 'bin' if mx.get_os() == 'windows' else 'lib'
+    shutil.copy(join(jdk.home, lib_folder, jdwp), join(basejdk_dir, lib_folder, jdwp))
     dt_socket = mx.add_lib_suffix(mx.add_lib_prefix('dt_socket'))
-    shutil.copy(join(jdk.home, 'lib', dt_socket), join(basejdk_dir, 'lib', dt_socket))
+    shutil.copy(join(jdk.home, lib_folder, dt_socket), join(basejdk_dir, lib_folder, dt_socket))
 
     if not args:
         args = []


compiler/mx.compiler/mx_compiler.py
```diff
@@ -962,6 +962,11 @@ def sl(args):
     mx.get_opts().jdk = 'jvmci'
     mx_truffle.sl(args)
 
+def lib_folder():
```
dougxc:
```suggestion
def jdk_lib_folder():
```

This is a slightly more precise name.

dougxc:
Looks good to me. I'll integrate and make the change I suggested while integrating.

JornVernee:
I think the last 2 calls are a mistake while taking another look.

I think I based those off of where the folders were in the build output instead of looking at the base JDK used. Those folders are (probably) supposed to go under `lib`.

I'll update and make the change you suggested as well.

dougxc:
Since this is now local to `makegraaljdk`, please remove the function and replace it with a local variable:
```
    lib = 'bin' if mx.get_os() == 'windows' else 'lib' 
```

compiler/mx.compiler/mx_compiler.py
```diff
@@ -973,9 +973,10 @@ def java_base_unittest(args):
         shutil.rmtree(basejdk_dir)
     mx.run([jlink, '--output', basejdk_dir, '--add-modules', basemodules, '--module-path', join(jdk.home, 'jmods')])
     jdwp = mx.add_lib_suffix(mx.add_lib_prefix('jdwp'))
-    shutil.copy(join(jdk.home, 'lib', jdwp), join(basejdk_dir, 'lib', jdwp))
+    lib_folder = 'bin' if mx.get_os() == 'windows' else 'lib'
+    shutil.copy(join(jdk.home, lib_folder, jdwp), join(basejdk_dir, lib_folder, jdwp))
     dt_socket = mx.add_lib_suffix(mx.add_lib_prefix('dt_socket'))
-    shutil.copy(join(jdk.home, 'lib', dt_socket), join(basejdk_dir, 'lib', dt_socket))
+    shutil.copy(join(jdk.home, jlib_folder, dt_socket), join(basejdk_dir, lib_folder, dt_socket))
```
dougxc: `jlib_folder` -> `lib_folder`

JornVernee: Yeah I saw it :). Strangely enough, when testing the change the copy seems to succeed :/

dougxc:
Can you please squash the PR to a single commit and then I'll integrate it.

JornVernee:
@dougxc Done. Sorry for the stumbles.

JornVernee:
Well, it should need the changes. The `jdwp.dll` and `dt_socket.dll` _are_ under `jre/bin` in the labs JVMCI JDK [1] on windows.

It seems I miss-remembered the command I used. I was working on several things at once, and made PR later.

I can't reproduce the problem when running `make graaljdk build`, but I can reproduce it when running `mx gate` when using JDK 11, which doesn't even have a `/jre` folder. It's firing during `verifylibraryurls`, but only when running `mx gate`:

bug

Kind of a niche problem after all... I don't know if JDK 11 is supposed to work at all yet? Some other commands like `makegraaljdk` block it's usage and ask for JDK 8.

[1] : https://www.oracle.com/technetwork/oracle-labs/program-languages/downloads/index.html

JornVernee:
@dougxc Yes. Running `mx makegraaljdk build` completes successfully without making any modifications to the repo.

Though the produced image doesn't seem to support usage of the polyglot APIs yet?

dougxc:
That's right - `makegraaljdk` only makes a JDK that includes the Graal compiler and uses it by default. It doesn't not include the rest of  the functionality of GraalVM (polyglot etc).

PR #805, opened 2018-11-14
Add spin-lock support via Thread.onSpinWait() intrinsic

dmlloyd:


dmlloyd:
This is probably not the right approach, but it *seems* to work.

Peter-B-Kessler:
Thank you for your contribution to SubstrateVM!

Such a substitution is not appropriate for SubstrateVM, but should instead be handled by the Graal compiler, so that the benefits of `Thread.onSpinWait()` are available to everyone that uses the Graal compiler (including all the SubstrateVM code), not just SubstrateVM developers.

dmlloyd:
Yeah I think this should be done with an intrinsic, or something like that, but I am not really clear on the "right" approach.  I'd be happy to do it "right" though, if only I could understand what that is :)

cstancu:
@dmlloyd this PR should be superseded by https://github.com/oracle/graal/commit/6b20d947094eea9a140ae834dd54e8d5d003427d. Can you please confirm and close this? Thanks!

dmlloyd:
Great! Thanks.

PR #821, opened 2018-11-25
Add note about macOS SDK headers for LLVM

fniephaus:
This PR extends Sulong's `README.md` with information on macOS SDK headers which are required to be installed on macOS Mojave.


fniephaus:
I've pointed quite a few people at this PR in the last few months because they ran into the same issue. Any way we can get this merged sometime soon?

rschatz:
I'm sorry, somehow I missed this PR even though it's assigned to me.

Looks good at a glance, but I'm not a Mac user, so I don't know if these instructions make sense. I'll ask around for a Mac user on our team to approve the PR, then I will start the process of integrating it.

dougxc:
Just to help people searching for a solution, here's a representative failure message if the headers are not installed:
```
Building com.oracle.truffle.llvm.libraries.bitcode with GNU Make... [rebuild needed by GNU Make]
/Users/dsimon/hs/graal/sulong/projects/com.oracle.truffle.llvm.libraries.bitcode/src/abort.c:30:10: fatal error: 'stdio.h' file not found
#include <stdio.h>
         ^~~~~~~~~
1 error generated.
make: *** [bin/abort.noopt.bc] Error 1

Building com.oracle.truffle.llvm.libraries.bitcode with GNU Make failed
1 build tasks failed
```

fniephaus:
No worries and thanks!

fniephaus:
@dougxc would you prefer to add this failure message as an example to the `README.md`?

dougxc:
Yes, that's a good idea. Even better would be to modify mx to detect this failure and show a message with the solution but maybe that's for another day ;-)

fniephaus:
@dougxc: done! Would be could to detect this in `mx`, but maybe someone working on Sulong should implement that.
@rschatz: I've updated the PR, ready to be merged :)

eregon:
For reference, here is the similar documentation in TruffleRuby:
https://github.com/oracle/truffleruby/blob/master/doc/user/installing-llvm.md#development-tools-and-system-headers

PR #830, opened 2018-11-29
Use cset to conditionally set 1 or 0 for AArch64.

XiaohongGong:


XiaohongGong:
Hi there, this patch is to use "cset" instruction to conditionally set 1 or 0 to the result. Could anyone please take a look at it? Thanks!

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64ConditionalSetTest.java
```diff
@@ -0,0 +1,200 @@
...
+    public static int conditionalSetEQZero(int m) {
+        if ((m & 2) == 0) {
+            return 1;
+        }
+        return 0;
+    }
+
+    @Test
+    public void testConditionalSetEQZero() {
+        test("conditionalSetEQZero", 0);
```
sanzinger: This testclass only checks if the result is correct but does not verify if the proper LIR instruction was selected.

Can we do the same as we do in the MatchRuleTests with `MatchRuleTest.checkLIR`, so that we can actually verify, that the LIR instruction was applied. The simplest way is to put this class as a subclass of `AArch64MatchRuleTest` even though it is not testing a match rule. This activates the guard to test only for AArch64 and provides the `checkLIR` method.

XiaohongGong: Yes, make sense! I just thought it was not testing a match rule. I will modify this like what you suggest. Thanks!

sanzinger:
Looks good, I am going to merge this change.

PR #834, opened 2018-11-30
Add "testBitAndBranch" rule for AArch64.

XiaohongGong:


XiaohongGong:
Hi there,
I have committed a new patch which could use "tbz/tbnz" to check whether an integer value is odd or even. There is a little problem about the branch range now, which is just +-/32KB for tbz/tbnz. There is not the related codes to do the branch differentiate right now in the graal framework. I did many tests by running the jtreg and benchmarks to test the range problem, and it seems that the patch is ok now.  I commented "TODO" in my patch that means my team and I may do this part in future. Could anyone please take a look at this patch? Thank you!

sanzinger:
The issue with the branch offsets only applies for forward branches. In SPARC we had the same issue with instructions only allow a restricted jump offset (`cbcond`). 

In SPARC we ran the code generation a first time and collected all the offsets. In a second pass we knew the offsets and we could generate the appropriate instruction. However, it costs a lot of CPU time, to run the code generation twice.

I started with an implementation estimating the upper bound of instruction count and constants count in LIR. I created the `SPARCLIRInstructionMixin`, where every SPARC LIR instruction provides the maximum size to be emitted (for code and for constants). Unfortunately the implementation is only used for constant table size but not for instruction estimates, instead the two-pass approach is still used.

Since we face this issue a second time in our backend, we should think to add a feature which addresses this issue to the backend. 

The way to share the code is to implement this:
* move the code from `SPARCLIRInstructionMixin` into the `LIRInstruction` class
* define an estimation for each LIR instruction used in AArch64
* before generating the code, build an array the maximum offsets for each LIR instruction, if a LIR instruction does not provide any estimation, throw an error
* provide a method `CompilationResultBuilder.estimateDisplacementUpperBound(Label)` where the generating code gets an estimation, what the upper bound of displacement is.
* add assertion in the code generator, to check that all LIR instructions generate less or equal bytes of code as they have estimated

This is just a suggestion for now, we need to discuss, if we want to go this way.

sanzinger:
Please skip over the ideas I mentioned above.  I had a quick discussion with @dougxc and we decided to extend the code generation in a way, that we can estimate the code size for forward jumps with a simple heuristic.
If the code generation fails because the heuristic was wrong, we rerun the code generation with forcing long jumps everywhere.

For this, I am thinking of a method `CompilationResultBuilder.isLabelWithinDisplacement(LIRInstruction instruction, Label label, int disp)` where the LIR can ask the backend for its estimation.

If the patching evenutally fails, the whole code generation part is rerun where the `isLabelWithinDisplacement` method always returns false.

I'll try to provide an implementation for this by thursday this week.

XiaohongGong:
@sanzinger thanks. I also checked the SPARC implementation for the branch checking. I tried to add the same codes for AArch64 and it worked. In a word, if you have a better way to implement it, I will stop and wait for your implementation. Thank you!

XiaohongGong:
@sanzinger is there any update for the branch checking implementation? Thanks!

sanzinger:
Hello, I got veered away last week. Today I started with the change. Give you an update tomorrow.

XiaohongGong:
@sanzinger  thanks!  I'll wait for your update.

sanzinger:
# Yes it will influence _all_ far jumps. Anyway, this should not happen too often when the estimation function is good
# No, it bails out permanently, since in the second run it is assumed that it just works. If not it is the safest way to just bail out gracefully.

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
```diff
@@ -178,6 +194,54 @@ public ComplexMatchResult multiplyAddSub(BinaryNode binary, ValueNode a, ValueNo
...
+        }
+        return null;
+    }
+
+    /**
+     * x % 2 == 0 -> x & 1 == 0.
+     */
+    @MatchRule("(If (IntegerTest value Constant=a))")
+    public ComplexMatchResult integerTestOddEvenBranch(IfNode root, ValueNode value, ConstantNode a) {
+        if (value.getStackKind().isNumericInteger()) {
```
sanzinger: I think we could extend this pattern to any bit set - this is what `tbz` and `tbnz` actually do. I think of:
```
long constant = a.asJavaConstant();
if(Long.bitCount(constant) == 1) {
  int bitToTest = Long.numberOfTrailingZeros(constant);
  emitBitTestAndBranch(root, value, bitToTest);
}
```

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
```diff
@@ -178,6 +194,54 @@ public ComplexMatchResult multiplyAddSub(BinaryNode binary, ValueNode a, ValueNo
...
+     * If x < 0:
+     *
+     * -> -((-x) & 1) == 0.
+     *
+     * else:
+     *
+     * -> ((((x >> (bitCount - 1)) >>> (bitCount - 1)) + x) & (-2)) == x.
+     */
+    @MatchRule("(If (IntegerEquals (Negate=cmpLeft (And (Negate value) Constant=a)) Constant=b))")
+    @MatchRule("(If (IntegerEquals (And=cmpLeft (Add (UnsignedRightShift (RightShift value Constant=a) Constant=a) value) Constant=b) value))")
```
sanzinger: I think this complex rules will hardly match any real world programs. Especially since simplifications may be done on a higher level in the compiler. I would vote for removing this special case.

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64OddEvenBranchTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+package org.graalvm.compiler.core.aarch64.test;
+
+import org.graalvm.compiler.lir.LIRInstruction;
+import org.graalvm.compiler.lir.aarch64.AArch64ControlFlow;
+import org.junit.Test;
+
+import java.util.function.Predicate;
+
+public class AArch64OddEvenBranchTest extends AArch64MatchRuleTest {
```
sanzinger: Generally speaking, the `tbz` and `tbnz` is a special case to test for even/odd. But `tbz` is more generic and can test any bit.
So the canonicalization if `x%2 == 0`  to `x&1==0` should be done on a higher level in the compiler. The code generation part should only focus on the bit testing.

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
XiaohongGong: Yes, make sense!

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
XiaohongGong: Actually, most of  the rem for power-of-2 computations will generate this pattern  `((((x >> (bitCount - 1)) >>> (bitCount - 1)) + x) & (-2)) == x` . So I think this rule is needed just for the odd/even checking.

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64OddEvenBranchTest.java
XiaohongGong: Actually and currently, the rem node for "power-of-2" condition will simplified to "x&n" just for the positive integer type.

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
sanzinger: Indeed. The issue here is, that it relies on the canonicalization happening `SignedRemNode.canonical`. But overall, this optimization is a generic optimization we should introduce, because other platforms can also benefit from this, when the result of the modulo is compared against zero.
I am thinking of `if(x%2==0)` could be transformed to `if(x&((1<<1)-1)==0)`. I'll go and implement this kind of optimization in the HIR, so we can focus on the simpler cases in the code generation.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/SignedRemNode.java
```diff
@@ -81,27 +85,55 @@ private static ValueNode canonical(SignedRemNode self, ValueNode forX, ValueNode
...
                 return ConstantNode.forIntegerStamp(stamp, 0);
-            } else if (CodeUtil.isPowerOf2(constY)) {
-                if (xStamp.isPositive()) {
-                    // x & (y - 1)
-                    return new AndNode(forX, ConstantNode.forIntegerStamp(stamp, constY - 1));
-                } else if (xStamp.isNegative()) {
-                    // -((-x) & (y - 1))
-                    return new NegateNode(new AndNode(new NegateNode(forX), ConstantNode.forIntegerStamp(stamp, constY - 1)));
+            } else if (CodeUtil.isPowerOf2(constY) && tool != null && tool.allUsagesAvailable()) {
+                if (allUsagesCompareAgainstZero(self)) {
```
sanzinger: @XiaohongGong when all usages just compare against zero, we can transform this to `(x&(y-1))==0`. Then we do not need the complex match rule because the case `x&1==0` covers it.

sanzinger: Here I created a [proof](https://rise4fun.com/Z3/XPZk) of `((x%n)==0)==((x&(((1<<n)-1))==0)`:
```
(set-logic QF_BV)
; Constants
(declare-const one (_ BitVec 32))
(assert (= one (_ bv1 32)))
(declare-const zero (_ BitVec 32))
(assert (= zero (_ bv0 32)))

; Inputs
(declare-const x (_ BitVec 32))
(declare-const y (_ BitVec 32))
(declare-const n (_ BitVec 32))              ; nth power
(assert (= y (bvshl one n)))                 ; y = 1 << n;

; reference implementation
(declare-const reference Bool)
(assert (= reference (= (bvsrem x y) zero)))     ; reference = (x%y)==0

; optimization
(declare-const opt Bool)
(assert (= opt (= (bvand x (bvsub y one)) zero))) ; opt = (x&(y-1))==0

; proof reference=opt
(assert (not (= reference opt))) ; check whether reference!=opt is satisfiable; if unsatisfiable the opposite must be true
(check-sat)
```

XiaohongGong: Wonderful!

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
XiaohongGong: Hi sanzinger, I extend this rule as you suggested. But I tested the  `AArch64TestBitAndBranchTest.testBit42` and it failed. The `bitToTest = Long.numberOfTrailingZeros(constant)` is 41, but the finally assembly code is `tbz #5`. It seems that there is something wrong with the tbz assembler function. I tried `(a & (1 << 7)) == 0` and the result is `tbz #3`.

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64OddEvenBranchTest.java
XiaohongGong: I have deleted this test file, because you have added a test "AArch64TestBitAndBranchTest".

XiaohongGong:
Hi @sanzinger , I'm so sorry that I disappeared your two patches when updating my tbz patch. To revert this status, I committed your two patches manually. Could you please check whether they are the right version? Thank you!

sanzinger:
The changes look good now.
I'll amend the author and rebase it on the latest master. Then I am going through the internal review/test process.

sanzinger:
Hmm, I took the offset from the [`AArch64Assembler.tbz`](https://github.com/oracle/graal/blob/master/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java#L910). It uses the `imm16` which suggests that the value is the offset in bytes. But the assertion is wrong. It should be `NumUtil.isSignedNbit(16, imm16);` in the assembler as well as in the `AArch64ControlFlow`.

sanzinger:
`isFarBranch = !crb.labelWithinRange(this, label, (1 << 12) - 1);` is more a heuristic, which assumes, that does not count assembler instructions but LIR instructions. We assume that each LIR instruction emits two AArch64 instructions, which means 2*4=8 bytes. So indeed, we may use (1<<13) LIR instructions. I'll fix that and then complete the integration.

I have some issues with the recently updated `StringIndexOfConstantTest` updated recently, which I have to fix as well ...

I'm doing that tomorrow, today I am on vacation.

PR #835, opened 2018-12-03
Enhance updategraalinopenjdk for development use (on Windows)

JornVernee:
The `updategraalinopenjdk` command in the compiler suite checks to see if the target jdk repo has any uncommitted changes using `hg status`, but this also recognizes untracked files as uncommitted changes.

This PR lets the command ignore untracked files.

---

Another Windows specific issue is; Python uses system native line endings when reading/writing from/to a file in text mode. When copying files from the graal repo into an openjdk repo using updategraalinopenjdk, this will change all the unix line endings to windows ones when someone is developing on windows.

This PR instead opens the files in binary mode which preserves the line endings.

### DIFF
--- a/compiler/mx.compiler/mx_updategraalinopenjdk.py
+++ b/compiler/mx.compiler/mx_updategraalinopenjdk.py
@@ -31,7 +31,6 @@ import shutil
 from collections import namedtuple
 from argparse import ArgumentParser
 from os.path import join, exists, dirname
-import re
 
 import mx
 import mx_compiler
@@ -113,6 +112,8 @@ def updategraalinopenjdk(args):
 
     jdkrepo = args.jdkrepo
 
+    ignored_files = []
+
     for m in graal_modules:
         m_src_dir = join(jdkrepo, 'src', m.name)
         if not exists(m_src_dir):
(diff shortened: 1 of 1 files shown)


compiler/mx.compiler/mx_updategraalinopenjdk.py
```diff
@@ -126,8 +126,12 @@ def run_output(args, cwd=None):
         m_src_dir = join('src', m.name)
         mx.log('Checking ' + m_src_dir)
         out = run_output(['hg', 'status', m_src_dir], cwd=jdkrepo)
```
dougxc: Couldn't this be done with a single regex? Something like: `out = re.sub(r'^\?.*\$', '', out, re.MULTILINE)` (probably not 100% correct).

JornVernee: Sure, that works. Updated

JornVernee: And now it doesn't work...

Let me look into it.

JornVernee: Oh, I figured it out. `re.MULTILINE` was being passed as `count` argument be accident.

I'll update again :-/

JornVernee:
Ok there it is, I slightly changed the regex since your's wasn't filtering out the newlines, so `out` was not empty after filtering.

JornVernee:
hmm. As a side effect it's `hg add`-ing all the untracked files in the diff as well if they're in the Graal directories, so simply ignoring the untracked files from `hg status` doesn't seem to be an option. I think we can remember them and then exclude them from the `add` later.

I will look into this

JornVernee:
That seems to do the trick :-)

I switched back to the for loop approach since we now have to remember the individual files.

JornVernee:
@dougxc Merged the other PR into this, want me to squash the commits as well?

dougxc:
Up to whether you want to squash or not.

JornVernee:
I'll keep it like this I think then, my git tool doesn't seem to allow me to squash the merge commits any ways. (and just in case if I need to pick the changes apart again)

JornVernee:
Closing old PR

PR #836, opened 2018-12-03
Copy files in binary mode to prevent line endings being changed (updategraalinopenjdk)

JornVernee:
Python uses system native line endings when reading/writing from/to a file in text mode. When copying files from the graal repo into an openjdk repo using `updategraalinopenjdk`, this will change all the unix line endings to windows ones when someone is developing on windows.

This PR instead opens the files in binary mode which preserves the line endings.


dougxc:
Hi @JornVernee , please combine this with #835 and make it a general "enhance updategraalinopenjdk for development use on Windows" issue.

PR #859, opened 2018-12-12
add at-file support to jaotc

iignatev:
from [JDK-8215322](https://bugs.openjdk.java.net/browse/JDK-8215322):

currently, the number of jaotc argument is limited by max command line length, in some cases, it isn't enough, for example compiling a number of selected classes. many java tools have support for '@file' so adding the same support to jaotc is natural and will solve the problem w/ limitation.


dougxc:
Looks good to me. @veresov can you please either integrate this or explain to @iignatev how to do it himself.

dougxc:
Fixed by 426facf5f8489c27518ce8d98e042a61d9d3192e.

PR #884, opened 2018-12-25
Canonicalize IfNode to ConditionalNode for variable assignment.

XiaohongGong:
This patch here wants to canonicalize IfNode to ConditionalNode. We want to use "csel" to implement the conditional value selecting operation for more conditions like Math.min/max, instead of using "cmp + b.cond".

The trueSuccessor and falseSuccessor blocks of IfNode must meet the following conditions:
1. The nodes in block only contain a start and an end one except for the constant and parameter node.
2. The merge node only contains a phi member.

All the checking jobs must begin after the final schedule phase is finished, so that the
nodes in each block will not be changed anymore.

Change-Id: I4aa0ae791e05f34e6af3dd1da035e42a97815b9b

### DIFF
--- a/compiler/src/org.graalvm.compiler.nodes.test/src/org/graalvm/compiler/nodes/test/IfNodeCanonicalizationTest2.java
+++ b/compiler/src/org.graalvm.compiler.nodes.test/src/org/graalvm/compiler/nodes/test/IfNodeCanonicalizationTest2.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, Arm Limited. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package org.graalvm.compiler.nodes.test;
+
+import org.graalvm.compiler.core.test.GraalCompilerTest;
+import org.graalvm.compiler.nodes.StructuredGraph;
+import org.graalvm.compiler.nodes.calc.ConditionalNode;
+import org.graalvm.compiler.options.OptionValues;
+import org.graalvm.compiler.phases.Phase;
+import org.graalvm.compiler.phases.tiers.Suites;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class IfNodeCanonicalizationTest2 extends GraalCompilerTest {
+    private StructuredGraph structuredGraph;
+
+    @Override
+    protected Suites createSuites(OptionValues opts) {
+        Suites suites = super.createSuites(opts);
+        suites.getLowTier().appendPhase(new Phase() {
+            @Override
+            protected void run(StructuredGraph graph) {
+                structuredGraph = graph;
+            }
+        });
+        return suites;
+    }
+
+    /**
+     * Math.min() tests for integer type, which can be simplified to conditional move.
+     */
+    public int minInt(int m, int n) {
+        return Math.min(m, n);
+    }
+
+    @Test
+    public void testMinInt() {
+        test("minInt", 1, 2);
+        checkConditionalNode("minInt", 1);
+    }
+
+    public long minLong(long m, long n) {
+        return Math.min(m, n);
+    }
+
+    @Test
+    public void testMinLong() {
+        test("minLong", 1L, 2L);
+        checkConditionalNode("minLong", 1);
+    }
+
+    /**
+     * Math.max() tests for integer type, which can be simplified to conditional move.
+     */
+    public int maxInt(int m, int n) {
+        return Math.max(m, n);
+    }
+
+    @Test
+    public void testMaxInt() {
+        test("maxInt", 1, 2);
+        checkConditionalNode("maxInt", 1);
+    }
+
+    public long maxLong(long m, long n) {
+        return Math.max(m, n);
+    }
+
+    @Test
+    public void testMaxLong() {
+        test("maxLong", 1L, 2L);
+        checkConditionalNode("maxLong", 1);
+    }
+
+    /**
+     * General min test for integer type, which can be simplified to conditional move.
+     */
+    public int generalMin(int m, int n) {
+        int value;
+        if (m <= n) {
+            value = m;
+        } else {
+            value = n;
+        }
+        return 2 * value;
+    }
+
+    @Test
+    public void testGeneralMin() {
+        test("generalMin", 1, 2);
+        checkConditionalNode("generalMin", 1);
+    }
+
+    /**
+     * General max test for integer type, which can be simplified to conditional move.
+     */
+    public int generalMax(int m, int n) {
+        if (m >= n) {
+            return m;
+        }
+        return n;
+    }
+
+    @Test
+    public void testGeneralMax() {
+        test("generalMax", 1, 2);
+        checkConditionalNode("generalMax", 1);
+    }
+
+    /**
+     * General conditional set tests for integer type, which can be simplified to conditional move.
+     */
+    public int integerEqualsCondMove(int m, int n, int a, int b) {
+        if (m == n) {
+            return a;
+        }
+        return b;
+    }
+
+    @Test
+    public void testIntegerEqualsCondMove() {
+        test("integerEqualsCondMove", 1, 2, 2, 4);
+        checkConditionalNode("integerEqualsCondMove", 1);
+    }
+
+    public int isNullCondMove(Object obj, int m, int n) {
+        if (obj == null) {
+            return m;
+        }
+        return n;
+    }
+
+    @Test
+    public void testIsNullCondMove() {
+        test("isNullCondMove", new Object(), 1, 2);
+        checkConditionalNode("isNullCondMove", 1);
+    }
+
+    private void checkConditionalNode(String methodName, int expected) {
+        compile(getResolvedJavaMethod(methodName), null);
+        int actual = structuredGraph.getNodes().filter(ConditionalNode.class).count();
+        Assert.assertEquals(expected, actual);
+    }
+}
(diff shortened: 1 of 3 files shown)


XiaohongGong:
Hi there,
Could anyone please take a look at this patch? Thanks!

tkrodriguez:
This seems like a somewhat intrusive way of accomplishing this.  Performing an extra canonicalization round is fairly expensive.  Also if I read the code in LowTier correctly the final round shouldn't be performing GVN because of the optimizations performed in FixReadsPhase.  If I understand this correctly, the use of a schedule to detect empty blocks is attempting to handle cases where the conditional is the only user of a complex true or false expression?  I think that's the reason we only perform the transformation for constants and parameters since otherwise the transformation can introduce extra unconditional work.  @gilles-duboscq maybe examining the number of users of the values would work just as well and could done all the time instead of only during the final canonicalizer phase.  It also suggests that sometimes it might make sense to convert a ConditionalNode into an explicit if for the same reason we don't perform the reverse transform.

tkrodriguez:
I agree that this kind of transform should be done relatively late so we can make sure we don't introduce extra unconditional work but it doesn't have to be done at the very end. We could do it in the canonicalize phase that runs after the FixReadsPhase in LowTier.  The schedule right before that isn't a FINAL_SCHEDULE but I don't think that really makes any difference for the cases we care about.  The interesting cases are diamonds so the only thing which should be in the true or false branches are values being produced by the conditional.  So we could add a flag on CanoniclizerPhase to control this, kind of like canonicalizeReads, and propagate it through CanonicalizerTool and use that to control this optimization.  Structurally I think that would be more acceptable.  Optimization of conditionals is a real rathole but I agree this specific pattern should always be a win.

tkrodriguez:
That seems like a reasonable way to go.

XiaohongGong:
@tkrodriguez I updated my patch according to the discussion above. Could you please take a look at it again? Thank you!

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/AbstractBeginNode.java
```diff
@@ -116,6 +118,23 @@ public void generate(NodeLIRBuilderTool gen) {
         };
     }
 
+    public List<Node> getBlockNodeList() {
+        StructuredGraph.ScheduleResult schedule = graph().getLastSchedule();
+        if (schedule == null) {
+            return null;
+        }
+
+        Block[] blocks = schedule.getCFG().getBlocks();
```
tkrodriguez: Any reason you aren't using schedule.getCFG().getNodeToBlock().get(this) instead of the linear search?

tkrodriguez:
I'm running it through out internal gate and may do some benchmark runs since using the conditional operators can have surprising performance effects, at least partly because it makes larger blocks where local scheduling can become more important.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/AbstractBeginNode.java
XiaohongGong: I didn't found this function before. I will use `schedule.getCFG().blockFor(this)` instead. Thank you!

XiaohongGong:
@tkrodriguez how is the test going now? Is there any update? Thank you!

tkrodriguez:
Sorry I went silent on you.  Can you rebase onto the latest master?  There's a conflict with some of your changes in IfNode that needs to be resolved and I have some style comments and minor bug fixes.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/AbstractBeginNode.java
```diff
@@ -116,6 +118,15 @@ public void generate(NodeLIRBuilderTool gen) {
         };
     }
 
+    public List<Node> getBlockNodeList() {
+        StructuredGraph.ScheduleResult schedule = graph().getLastSchedule();
+        if (schedule == null) {
+            return null;
+        }
+        Block block = schedule.getCFG().blockFor(this);
```
tkrodriguez: You need a couple more guards in there to be safe.  Something like this:
`        if (schedule.getCFG().getNodeToBlock().isNew(this)) {
            // This can occur when nodes were created after the last schedule
            return null;
        }
        Block block = schedule.getCFG().blockFor(this);
        if (block == null) {
            return null;
        }
        return schedule.nodesFor(block);
`
I think we should move this over to IfNode as well.  It's not really something we want to advertise for general use.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/IfNode.java
```diff
@@ -893,14 +893,28 @@ private ValueNode canonicalizeConditionalViaImplies(ValueNode trueValue, ValueNo
         return null;
     }
 
+    private boolean isEmptyBlock(ValueNode value, AbstractBeginNode successor) {
```
tkrodriguez: Can we call this isSafeConditionalInput?  And some comments explaining the thinking.  Basically Constants and Parameters are cheap to materialize and if we have a schedule then an empty block means the value is already materialized so we aren't adding unconditional work by converting to an conditional form.

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/GraalOptimizationTest.java
```diff
@@ -69,4 +87,9 @@ protected void checkLIR(String methodName, Predicate<LIRInstruction> predicate,
         Assert.assertEquals(expected, actualOpNum);
     }
 
+    protected void checkNode(String methodName, Class<? extends Node> expectedClass, int expected) {
```
tkrodriguez: checkNode -> checkNodeCount?

tkrodriguez:
When you merge with latest master the new LLVM based SVM backend is now in this repo and I think these changes have exposed a problem with conditional lowering there.  I have a fix for it and if you don't mind I'll just include that fix when I merge your changes.  Otherwise I think these changes look ok.

compiler/src/org.graalvm.compiler.nodes.test/src/org/graalvm/compiler/nodes/test/IfNodeCanonicalizationTest2.java
```diff
@@ -0,0 +1,182 @@
...
+     * IfNode with computations in true and false blocks can't be simplified to ConditionalNode.
+     */
+    public int complexBlock(int m, int n) {
+        if (m <= n) {
+            return m + n;
+        }
+        return m - n;
+    }
+
+    @Test
```
tkrodriguez: Can we remove this test and testMultiPhi since they are negative tests?  We mainly want to test that patterns we are about are optimized.

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/GraalOptimizationTest.java
XiaohongGong: This name has the same style of "checkLIR". And "checkLIR" has been used in several existing testing files. So can we keep this name here?

tkrodriguez: Well checkLIR isn't a great name either.  :) Actually looking at this again, if we remove the negative tests then the only possible argument for count is 1 so I think can can inline that argument.  I'd also inline the CLASS argument and emit it as ConditionalNode.class instead of having an alias for it.  It will use an ldc anyway so having a static final alias for it is just obfuscating I think.  In that case maybe checkConditional is a better name for it.

XiaohongGong:
> When you merge with latest master the new LLVM based SVM backend is now in this repo and I think these changes have exposed a problem with conditional lowering there. I have a fix for it and if you don't mind I'll just include that fix when I merge your changes. Otherwise I think these changes look ok.

Hi @tkrodriguez,
I met a compiling error when running "mx build" after I merge with latest master. Here is the error: 
"WARNING: Make `ninja` binary available via PATH to build native projects.
WARNING: No platform-specific definition is available for library LLVM_PLATFORM for your architecture (aarch64)
WARNING: No platform-specific definition is available for library LLVM_PLATFORM_SPECIFIC for your architecture (aarch64)
library LLVM_PLATFORM_SPECIFIC:
Library without "path" attribute must have a non-empty "urls" list attribute or "maven" attribute". It seems that there is not the env setting for LLVM_PLATFORM for the aarch64 platform. Do you know how to resolve it, though it doesn't relate to this patch?

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/GraalOptimizationTest.java
XiaohongGong: Make sense! Anyway I will change this name as your suggestion! Thank you!

XiaohongGong:
> When you merge with latest master the new LLVM based SVM backend is now in this repo and I think these changes have exposed a problem with conditional lowering there. I have a fix for it and if you don't mind I'll just include that fix when I merge your changes. Otherwise I think these changes look ok.

What will the problem be? What is the relate backend difference between SVM backend and the current one?

tkrodriguez:
Basically the LLVM backend in the context of SVM ends up mixing a pointer and a long in a conditional and needs to add a little type conversion to make LLVM happy.  It does this in other places but seems to be missing it for conditionals.

tkrodriguez:
Your comment about build problems is gone so I assume you got that resolved?

XiaohongGong:
> Your comment about build problems is gone so I assume you got that resolved?

I resolved it locally to make sure my patch is compiled ok. But I didn't commit the changes here. I wonder whether someone will fix it for aarch64.

tkrodriguez:
```

**Comment by XiaohongGong** [comment]
> I think something like this is probably what's required. Could you try that out?
> 
> ```
> diff --git a/compiler/mx.compiler/suite.py b/compiler/mx.compiler/suite.py
> index d76db6d3f65..26848f3ec2a 100644
> --- a/compiler/mx.compiler/suite.py
> +++ b/compiler/mx.compiler/suite.py
> @@ -240,6 +240,10 @@ suite = {
>                "classifier": "linux-x86_64"
>              },
>            },
> +          "aarch64": {
> +            "path": "",
> +            "optional": True
> +          }
>          },
>          "darwin": {
>            "amd64": {
> ```

Yes. Similar but I use "\<others\>" while not "aarch64".

**Comment by XiaohongGong** [comment]
Hi @tkrodriguez , how is this patch going now?

File: compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/GraalOptimizationTest.java [hunk 6]
```diff
@@ -69,4 +87,9 @@ protected void checkLIR(String methodName, Predicate<LIRInstruction> predicate,
         Assert.assertEquals(expected, actualOpNum);
     }
 
+    protected void checkNodeCount(String methodName, Class<? extends Node> expectedClass, int expected) {
+        compile(getResolvedJavaMethod(methodName), null);
+        int actual = structuredGraph.getNodes().filter(expectedClass).count();
+        Assert.assertEquals(expected, actual);
+    }
```
XiaohongGong: I will move this function and its related codes to "IfNodeCanonicalizationTest2.java" and revert all the modifications in this file, to keep this file used for backend optimizations alone.

dougxc:
@tkrodriguez can you please give this PR some attention.

XiaohongGong:
Hi @tkrodriguez , how is this PR going now? Thank you!

XiaohongGong:
Hi @tkrodriguez , how is this PR going now? If it's controversial for this mid-end optimization, so how do you think if we just limit the optimization to Math.min/max? If so, we can simply add the intrinsics for Math on AArch64. Thank you!

tkrodriguez:
I'll look at this today and see where it stands.  I can't remember what the performance looks like which is my main concern about this change.

XiaohongGong:
> I'll look at this today and see where it stands. I can't remember what the performance looks like which is my main concern about this change.

Thank you so much that you could look at it again!

tkrodriguez:
Could you rebase this branch?  I've done a local merge for testing but it would be better to have a clean rebase as there are conflicts.

XiaohongGong:
> Could you rebase this branch? I've done a local merge for testing but it would be better to have a clean rebase as there are conflicts.

Sure! I'v rebased and updated my patch. Thanks a lot!

PR #887, opened 2018-12-28
CounterMode::implCrypt Hotspot stub

jphalimi-intel:
This patch enables HotSpot's CounterMode::implCrypt stub in Graal. I would like a second opinion on this code as I was not able to see the stub being applied on AESGCMBench.decrypt. Perhaps I missed something somewhere?

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/HotSpotBackend.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/HotSpotBackend.java
@@ -46,6 +46,7 @@ import org.graalvm.compiler.hotspot.nodes.aot.ResolveConstantStubCall;
 import org.graalvm.compiler.hotspot.replacements.AESCryptSubstitutions;
 import org.graalvm.compiler.hotspot.replacements.BigIntegerSubstitutions;
 import org.graalvm.compiler.hotspot.replacements.CipherBlockChainingSubstitutions;
+import org.graalvm.compiler.hotspot.replacements.CounterModeSubstitutions;
 import org.graalvm.compiler.hotspot.replacements.SHA2Substitutions;
 import org.graalvm.compiler.hotspot.replacements.SHA5Substitutions;
 import org.graalvm.compiler.hotspot.replacements.SHASubstitutions;
@@ -265,13 +266,27 @@ public abstract class HotSpotBackend extends Backend implements FrameMap.Referen
         unsafeArraycopyStub(UNSAFE_ARRAYCOPY, srcAddr, dstAddr, size);
     }
 
+    @NodeIntrinsic(ForeignCallNode.class)
+    private static native void unsafeArraycopyStub(@ConstantNodeParameter ForeignCallDescriptor descriptor, Word srcAddr, Word dstAddr, Word size);
+
     /**
      * Descriptor for {@code StubRoutines::_ghash_processBlocks}.
      */
     public static final ForeignCallDescriptor GHASH_PROCESS_BLOCKS = new ForeignCallDescriptor("ghashProcessBlocks", void.class, Word.class, Word.class, Word.class, int.class);
 
+    /**
+     * @see CounterModeSubstitutions#implCrypt
+     */
+    public static final ForeignCallDescriptor COUNTERMODE_IMPL_CRYPT = new ForeignCallDescriptor("counterModeAESCrypt", int.class, Word.class, Word.class, Word.class, Word.class, int.class,
+                    Word.class, Word.class);
+
+    public static int counterModeAESCrypt(Word srcAddr, Word dstAddr, Word kPtr, Word cntPtr, int len, Word encCntPtr, Word used) {
+        return counterModeAESCrypt(HotSpotBackend.COUNTERMODE_IMPL_CRYPT, srcAddr, dstAddr, kPtr, cntPtr, len, encCntPtr, used);
+    }
+
     @NodeIntrinsic(ForeignCallNode.class)
-    private static native void unsafeArraycopyStub(@ConstantNodeParameter ForeignCallDescriptor descriptor, Word srcAddr, Word dstAddr, Word size);
+    private static native int counterModeAESCrypt(@ConstantNodeParameter ForeignCallDescriptor descriptor, Word srcAddr, Word dstAddr, Word kPtr, Word cntPtr, int len, Word encCntPtr,
+                    Word used);
 
     /**
      * @see VMErrorNode
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/CounterModeSubstitutions.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/CounterModeSubstitutions.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2016, 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.graalvm.compiler.hotspot.replacements;
+
+import static org.graalvm.compiler.hotspot.GraalHotSpotVMConfigBase.INJECTED_INTRINSIC_CONTEXT;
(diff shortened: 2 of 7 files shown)


mur47x111:
Please apply the following patch for the moment:
```
diff --git a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/HotSpotReplacementsUtil.java b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/HotSpotReplacementsUtil.java
index 55f3abe7959..4b5d3a2ebf4 100644
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/HotSpotReplacementsUtil.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/HotSpotReplacementsUtil.java
@@ -138,10 +138,12 @@ public class HotSpotReplacementsUtil {
         }
     }
 
+    @Fold
     public static ResolvedJavaType methodHolderClass(@Fold.InjectedParameter IntrinsicContext context) {
         return context.getOriginalMethod().getDeclaringClass();
     }
 
+    @Fold
     static ResolvedJavaType getType(@Fold.InjectedParameter IntrinsicContext context, String typeName) {
         try {
             UnresolvedJavaType unresolved = UnresolvedJavaType.create(typeName);
```

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/CounterModeSubstitutions.java
```diff
@@ -0,0 +1,84 @@
...
+        Object realReceiver = piCastNonNull(receiver, HotSpotReplacementsUtil.methodHolderClass(INJECTED_INTRINSIC_CONTEXT));
+        Object embeddedCipher = RawLoadNode.load(realReceiver, embeddedCipherOffset(INJECTED_INTRINSIC_CONTEXT), JavaKind.Object, LocationIdentity.any());
+        Object aesCipher = piCastNonNull(embeddedCipher, aesCryptType(INJECTED_INTRINSIC_CONTEXT));
+
+        Word srcAddr = WordFactory.unsigned(ComputeObjectAddressNode.get(in, ReplacementsUtil.getArrayBaseOffset(INJECTED_METAACCESS, JavaKind.Byte) + inOff));
+        Word dstAddr = WordFactory.unsigned(ComputeObjectAddressNode.get(out, ReplacementsUtil.getArrayBaseOffset(INJECTED_METAACCESS, JavaKind.Byte) + outOff));
+
+        int cntOffset = HotSpotReplacementsUtil.getFieldOffset(
+                        HotSpotReplacementsUtil.getType(INJECTED_INTRINSIC_CONTEXT, "Lcom/sun/crypto/provider/CounterMode;"),
+                        "counter");
```
mur47x111: Extra this into a separate method annotated by `@Fold`, e.g.,
```
    @Fold
    static int counterOffset(@Fold.InjectedParameter IntrinsicContext context) {
        return HotSpotReplacementsUtil.getFieldOffset(
                HotSpotReplacementsUtil.getType(context, "Lcom/sun/crypto/provider/CounterMode;"),
                "counter");
    }
```

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/CounterModeSubstitutions.java
```diff
@@ -0,0 +1,84 @@
...
+        int encCntOffset = HotSpotReplacementsUtil.getFieldOffset(
+                        HotSpotReplacementsUtil.getType(INJECTED_INTRINSIC_CONTEXT, "Lcom/sun/crypto/provider/CounterMode;"),
+                        "encryptedCounter");
+        int usedOffset = HotSpotReplacementsUtil.getFieldOffset(
+                        HotSpotReplacementsUtil.getType(INJECTED_INTRINSIC_CONTEXT, "Lcom/sun/crypto/provider/CounterMode;"),
+                        "used");
+        Object kObject = RawLoadNode.load(aesCipher, AESCryptSubstitutions.kOffset(INJECTED_INTRINSIC_CONTEXT), JavaKind.Object, LocationIdentity.any());
+        Object cntObj = RawLoadNode.load(realReceiver, cntOffset, JavaKind.Object, LocationIdentity.any());
+        Object encCntObj = RawLoadNode.load(realReceiver, encCntOffset, JavaKind.Object, LocationIdentity.any());
+        Object used = RawLoadNode.load(realReceiver, usedOffset, JavaKind.Int, LocationIdentity.any());
```
mur47x111: This is not an object.

jphalimi-intel: Thanks for the review Yudi. Let me fix these things. The above change could be used throughout the rest of the method substitutions as well I think.

jphalimi-intel: @mur47x111 Just looked the type for used. I could not find another load type of Integer. What should the type of used be in this case if not object?

mur47x111: You could create one:
```
    @NodeIntrinsic(value = RawLoadNode.class)
    private static native int loadIntFromObjectIntrinsic(Object object, long offset, @ConstantNodeParameter JavaKind wordKind, @ConstantNodeParameter LocationIdentity locationIdentity);
```

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/HotSpotBackend.java
```diff
@@ -265,13 +265,27 @@ public static void unsafeArraycopy(Word srcAddr, Word dstAddr, Word size) {
...
     /**
      * Descriptor for {@code StubRoutines::_ghash_processBlocks}.
      */
     public static final ForeignCallDescriptor GHASH_PROCESS_BLOCKS = new ForeignCallDescriptor("ghashProcessBlocks", void.class, Word.class, Word.class, Word.class, int.class);
 
+    /**
+     * @see CounterMode#implCrypt
+     */
+    public static final ForeignCallDescriptor COUNTERMODE_IMPL_CRYPT = new ForeignCallDescriptor("counterModeAESCrypt", int.class, Word.class, Word.class, Word.class, Word.class, int.class,
+                    Word.class, Word.class);
```
mur47x111: Looking at http://hg.openjdk.java.net/jdk/jdk/file/af7afdababd3/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp#l3881, the very last argument would be `int.class`, right?

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/CounterModeSubstitutions.java
jphalimi-intel: Sorry Yudi, but which native call does this correspond to? I cannot find the source for the native calls such as loadWordFromObjectIntrinsic.

mur47x111: There are no implementations to `@NodeIntrinsic`-annotated methods. Graal will create IR nodes (of the type specified by `value=`) when encountering invocations to these methods.

mur47x111:
> Please apply the following patch for the moment:

I have pushed the full patch https://github.com/oracle/graal/commit/470fb3507fc881e50afaafc5f974620626cce899

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/CounterModeSubstitutions.java
jphalimi-intel: But how to RawLoadNode can figure out how to return an int?

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/HotSpotBackend.java
mur47x111: Sorry I was wrong. It is expecting an int pointer.

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/CounterModeSubstitutions.java
mur47x111: For example, it depends on the `JavaKind` parameter that we feed to `loadWordFromObjectIntrinsic`. You could also take a look at the generated source code in Plugin_HotSpotReplacementsUtil.java. You might look for the class `Plugin_HotSpotReplacementsUtil_loadWordFromObjectIntrinsic`.

Anyway, I think we don't need to load the value here. Just calculate the address of `used` might be sufficient. I.e.,
``
        Word usedPtr = WordFactory.unsigned(ComputeObjectAddressNode.get(realReceiver, usedOffset(INJECTED_INTRINSIC_CONTEXT)));
```

mur47x111:
Looks good to me! I will integrate it.

jphalimi-intel:
I fixed the unused imports.

mur47x111:
@jphalimi-intel I did not know that you have force-pushed another commit. Could you please check what is missing in https://github.com/oracle/graal/commit/617bf08253e5e315ba761ddbaa385fb4d7a62afb?

jphalimi-intel:
Hi Yudi,

I force-pushed to make sure all the Travis build checks work. There were a couple of issues that I was not able to reproduce on my local system but that the automatic build reported. All the issues should be resolved by now. The patch should be ready for merging.

mur47x111:
Sorry I was not clear in my last comment. I have fixed these issues and merged at https://github.com/oracle/graal/commit/617bf08253e5e315ba761ddbaa385fb4d7a62afb on your earlier commit https://github.com/oracle/graal/commit/59ce8b87e34ae1424bcc33e9e867aab4a6f49f06 . Anyway, I double-checked your newest commit. There is no additional change to be done.

I will close this PR. It won't show up as merged but your earlier commit is already in. Please avoid using git rebase in your next contribution as it will be more Github-friendly and collaborator-friendly ;)

dougxc:
Just to clarify, rebasing is ok up until the point where someone says they will integrate the PR. Even then, rebasing can be still done but should be coordinated with the Oracle dev doing the integration.

Issue #893, opened 2019-01-03
[Proposal


fwbrasil:
I've been working on prototypes to optimize interface dispatches in Graal. I'm planning to start working on pull requests but would like to get your feedback on the general approach before proceeding.

## Problem

Scala codebases usually define computations using composable APIs that build a graph to be later executed by an interpreter. These APIs include abstractions like [Scala's](https://github.com/scala/scala/blob/2.13.x/src/library/scala/concurrent/Future.scala) and [Twitter's](https://github.com/twitter/util/blob/develop/util-core/src/main/scala/com/twitter/util/Future.scala) `Future`s for asynchronous computations, [`Stitch`](https://www.youtube.com/watch?v=VVpmMfT8aYw) and [`Clump`](https://github.com/getclump/clump) for batching of remote asynchronous calls, Monix's [`Task`](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/Task.scala) and ZIO's [`IO`](https://github.com/scalaz/scalaz-zio) for defining pure computations, and many other monad-like APIs.

These abstractions introduce an execution behavior usually not present in traditional Java applications. Given that the definition of the computation graph involves several method invocations, the interpreter typically doesn't get inlined with the definition of the computation, which produces a high number of `invokeinterface` calls. At interpretation time, there isn't much profiling information (often none) to determine the concrete implementations of the transformations, so they don't get optimized.

For instance, this is a screenshot of a VTune profiling session of an application that uses `Stitch`:

![](https://i.imgur.com/4UidB7K.png)

The `itable stub` is the mechanism that Hotspot uses to dispatch `invokeinterface` calls, and, as we can observe, most of the CPU usage goes into executing these stubs.

Better inlining might be able to reduce the number of `invokeinterface` calls but, given the nature of these codebases, they'll always produce a high number of interface dispatches.

## Solution

I've been exploring alternatives to optimize interface dispatches in Graal. The idea is to have a set of strategies that leverage profiling information and common patterns to inline `invokeinterface` stubs. Graal would evaluate these strategies and choose the cheapest one.

### Strategies

I've identified a few strategies so far:

Common offset

If the observed implementations of an interface resolve to the same method offset, it's possible to inline the stub at the common offset and introduce a guard to deoptimize if a new implementation is observed.

Common superclass

[...]

Issue #412, opened 2018-05-09
Performance regresion on long methods with big lookup switches


plokhotnyuk:
I have a benchmark which parses JSON data from UTF-8 bytes to a structure of nested classes with 20-40 fields.

Code for that is generated by Scala macros and compiled by scalac to long methods with big lookup switches (one for each field) like here: https://gist.github.com/plokhotnyuk/1e2ee19d3cc80c3644bc9e453c8aae77

With GraalVM CE/EE it shows ~2x slowdown comparing to running on Oracle JDK 8.

Steps to reproduce: 
1. Ensure that [sbt](https://www.scala-sbt.org/download.html) is installed
2. Clone project with the benchmark that reproduces the issue: 
```sh
git clone git@github.com:plokhotnyuk/jsoniter-scala.git
```
3. Switch to the branch:
```sh
git checkout graalvm-slowdown-on-big-method-2
```
4. Run the benchmark on Oracle JDK 8 using specified path to it like here:
```sh
sbt clean 'benchmark/jmh:run -jvm /usr/lib/jvm/java-8-oracle/bin/java .*TwitterAPIBenchmark.readJsoniterScala.*'
```
5. Run the benchmark on GraalVM EE using specified path to it like here:
```sh
sbt clean 'benchmark/jmh:run -jvm /usr/lib/jvm/graalvm-ee-1.0.0-rc1/bin/java .*TwitterAPIBenchmark.readJsoniterScala.*'
```
6. Compare results of throughput (in ops/sec). 

I have manadged to run it with Async profiler through sbt-jmh extras integration. Please see flame graph reports, for both JVM which was committed in the following directory: https://github.com/plokhotnyuk/jsoniter-scala/tree/graalvm-slowdown-on-big-method-2/profile-async

plokhotnyuk:
Screenshot of the reverse flame graph from GraalVM EE:
![image](https://user-images.githubusercontent.com/890289/39872589-c480a816-5468-11e8-803f-0c26f61ca886.png)

plokhotnyuk:
Screenshot of the reverse flame graph build using GraalVM EE:

![image](https://user-images.githubusercontent.com/890289/39872677-076d2ec4-5469-11e8-8bea-84e925fcbb26.png)

plokhotnyuk:
The same graph with JDK 8:
![image](https://user-images.githubusercontent.com/890289/39872756-4345c79e-5469-11e8-8d0d-149aca60d123.png)

[...]

PR #895, opened 2019-01-05
Introduce hash table switch strategy

fwbrasil:
This is the first change related to https://github.com/oracle/graal/issues/893.

## Problem

The current switch strategies don't work well when the keys are sparse values since the density of the table switch is too low and Graal has to choose less efficient strategies.

## Solution

Introduce a new switch strategy based on cheap hash functions. The `Hasher` evaluates the available functions and tries to find one that guarantees no conflicts for the provided keys. The approach uses imperfect hash functions based on the paper ["Improving Switch Statement Performance with Hashing Optimized at Compile Time"](http://programming.sirrida.de/hashsuper.pdf).

`HashTableSwitchOp` emits the assembly code for the hash table switch. If there's a default target, it also emits the original keys as part of the table and verifies the value to handle possible conflicts with unknown keys.

I've done some [benchmarking](https://github.com/fwbrasil/scala-graal/blob/ed07f1a249b99a3fa89397392bfa9f5c15d56125/src/main/scala/bench/HashSwitchBench.java#L122) and, in some cases, the hash table strategy has similar performance to a regular table switch but with better compaction:

```
BinaryStrategy[avgEffort=6.416389351081533] (the default)

Benchmark                    Mode  Cnt         Score        Error  Units
HashSwitchBench.httpStatus  thrpt   20  26146297.275 ± 345930.644  ops/s

Table switch (density 0.15048543689320387)

Benchmark                    Mode  Cnt         Score        Error  Units
HashSwitchBench.httpStatus  thrpt   20  32873184.212 ± 384754.780  ops/s

Hasher[function=x, effort=1, cardinality=224] (density 0.2767857142857143)

Benchmark                    Mode  Cnt         Score        Error  Units
HashSwitchBench.httpStatus  thrpt   20  32400323.086 ± 218257.904  ops/s
```

Even though the hash table strategy can achieve better compaction than the regular table switch for some cases, the current heuristic used to decide between table switches and other strategies isn't very robust and rarely chooses the hash switch. This issue will be addressed in another PR as described in https://github.com/oracle/graal/issues/893

## Notes

1. Is it ok to implement the new strategy only for one architecture?
2. I don't have much experience writing assembly. This first version doesn't seem very optimized since it uses more registers and uses two moves when `defaultTarget` is not null. I'm also not sure if the padding is ok. Suggestions are appreciated :)


compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGenerator.java
```diff
@@ -449,29 +455,58 @@ public Variable emitForeignCall(ForeignCallLinkage linkage, LIRFrameState frameS
...
     @Override
     public void emitStrategySwitch(JavaConstant[] keyConstants, double[] keyProbabilities, LabelRef[] keyTargets, LabelRef defaultTarget, Variable value) {
+
         int keyCount = keyConstants.length;
         SwitchStrategy strategy = SwitchStrategy.getBestStrategy(keyProbabilities, keyConstants, keyTargets);
+
+        double minDensity = 1 / Math.sqrt(strategy.getAverageEffort() * tableSwitchBonus);
+
+        Optional<Hasher> hasher = Optional.empty();
+        if (supportsHashTableSwitch()) {
```
christhalinger: Having a method `

christhalinger: Having a `supportsHashTableSwitch()` seems odd. Could we instead add a method that returns the hasher and the default implementation just returns `Optional.empty()`?

christhalinger:
> Is it ok to implement the new strategy only for one architecture?

Yes, that's ok. Some things are only implemented for AMD64. We just need to find a good way to abstract it (see my comment).

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGenerator.java
```diff
@@ -88,6 +90,8 @@
 
     public static class Options {
         // @formatter:off
+        @Option(help = "", type = OptionType.Expert)
```
dougxc: New options must not have an empty `help` value. That said, it would be best to avoid adding this option altogether. Can you not tune the heuristic to choose the hashing strategy?

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/hashing/HashFunction.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.function.Function;
+
+import org.graalvm.compiler.lir.gen.ArithmeticLIRGeneratorTool;
+
+import jdk.vm.ci.meta.JavaConstant;
+import jdk.vm.ci.meta.Value;
+
+abstract class HashFunction {
+
+    public abstract int apply(long value, long s);
```
dougxc: What does the parameter `s` mean?

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/hashing/Hasher.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
+
+import org.graalvm.compiler.lir.gen.ArithmeticLIRGeneratorTool;
+
+import jdk.vm.ci.meta.JavaConstant;
+import jdk.vm.ci.meta.Value;
+
+public final class Hasher {
```
dougxc: Please javadoc this new class and its methods.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64ControlFlow.java
```diff
@@ -314,6 +314,97 @@ public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
...
+
+            if (!indexReg.equals(idxScratchReg)) {
+                masm.movl(idxScratchReg, indexReg);
+            }
+
+            // Set scratch to address of jump table
+            masm.leaq(scratchReg, new AMD64Address(AMD64.rip, 0));
+            final int afterLea = masm.position();
+
+            // Jump to default target if index doesn't contain the expected value
```
tkrodriguez: Some comments explaining the table structure would be helpful and maybe merge the defaultTarget != null ifs.  I think you might be able merge the leaq/movslq into one instruction, either by patching the movslq instead or by precomputing the expected offset to the jump table.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64ControlFlow.java
```diff
@@ -314,6 +314,97 @@ public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
...
+            if (defaultTarget != null) {
+                masm.movslq(valueScratchReg, new AMD64Address(scratchReg, idxScratchReg, Scale.Times8, 0));
```
tkrodriguez: you might consider loading 8 bytes here instead of issuing two loads.  If you put key is in the low 32 bits you can still use cmpl and then just perform a shift instead of the second load.  You'd want to align the table at 8 bytes instead which might be a good idea in any case when you have both key and offset in the table.  Having a single load would also allow you to eliminate the move of indexReg to idxScratchReg since you'd only have a single use of it.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/hashing/Hasher.java
```diff
@@ -0,0 +1,125 @@
...
+            for (HashFunction f : HashFunction.instances()) {
+                for (int cardinality = keys.length; cardinality < maxCardinality; cardinality++) {
+                    if (isValid(keys, min, f, cardinality)) {
+                        candidates.add(new Hasher(f, cardinality, min));
+                        break;
+                    }
+                }
+            }
+            if (candidates.isEmpty()) {
+                return Optional.empty();
```
tkrodriguez: Any idea how often this happens in practice?

tkrodriguez:
In your benchmark numbers what kind of switch implementation is being used in the "without hash switch" case?  Presumably it's not a tableswitch.  how does that perform?

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/hashing/Hasher.java
fwbrasil: It happens often since `minDensity` is normally too low and the hasher stops the evaluation based on it. It's necessary to review the heuristic that makes the decision between the density of the table switch x the cost of other strategies. I'll remove the bonus parameter for now and follow up with a separate PR.

plokhotnyuk:
I hope it will fix the following performance regression: https://github.com/oracle/graal/issues/412

fwbrasil:
@plokhotnyuk I doesn't seem it will. The values are too sparse and the hasher can only find a function at arity 846, which would make the table too large.

```java
        int[] keys = {3355, -1193609971, 3373707, -43264386, 1901043637, -1724546052, 116079, -2102114367, -608539730, -460163995, -1666926107, -1974943731, 1369680106, 226316666, 1699658638,
                        36848094, -1085743021, -1994383672, -1824078800, 3314158, 647263482, 583435343, 1690715422, 1592986856, -1189803120, 594447228, 1021718665, -174080651, -782008927, -690137550,
                        1512257332, -1412739479, 486184128, -1774134745, -101767272, 284519336, -312724053, 545330631, 765915793, 585125142, 1272354024, -475640257};
        JavaConstant[] c = new JavaConstant[keys.length];
        for (int i = 0; i < keys.length; i++)
            c[i] = JavaConstant.forInt(keys[i]);
        Optional<Hasher> forKeys = forKeys(c, 0.00001);
        System.out.println(forKeys); // Optional[Hasher[function=rotateRight(x, p) ^ x, effort=5, cardinality=864]]
    }
```

fwbrasil:
@christhalinger @tkrodriguez @dougxc thank you for the reviews! I've just addressed your suggestions.

@tkrodriguez I changed the code to force specific strategies to collect the performance numbers:

```
BinaryStrategy[avgEffort=6.416389351081533] (the default)

Benchmark                    Mode  Cnt         Score        Error  Units
HashSwitchBench.httpStatus  thrpt   20  26146297.275 ± 345930.644  ops/s


Table switch (density 0.15048543689320387)

Benchmark                    Mode  Cnt         Score        Error  Units
HashSwitchBench.httpStatus  thrpt   20  32873184.212 ± 384754.780  ops/s


Hasher[function=x, effort=1, cardinality=224] (density 0.2767857142857143)

Benchmark                    Mode  Cnt         Score         Error  Units
HashSwitchBench.httpStatus  thrpt   20  65606834.026 ± 1701259.405  ops/s
```

I'm puzzled that the hash table switch performs much better than the table switch. I did a bunch of manual testing and it seems to function properly, but I wonder if there's something wrong.

fwbrasil:
okay, I just found the bug. The register annotations were wrong. The hash table switch has performance similar to the regular table switch after the fix:

```
Hasher[function=x, effort=1, cardinality=224] (density 0.2767857142857143)

Benchmark                    Mode  Cnt         Score        Error  Units
HashSwitchBench.httpStatus  thrpt   20  32400323.086 ± 218257.904  ops/s
```

fwbrasil:
I've introduced tests and addressed the initial review comments. The PR is ready for review.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/hashing/Hasher.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+
+import org.graalvm.compiler.lir.gen.ArithmeticLIRGeneratorTool;
+
+import jdk.vm.ci.meta.JavaConstant;
+import jdk.vm.ci.meta.Value;
+
+/**
+ * This class holds a hash function at a specific cardinality and min value (lowest key).
```
dougxc: Can you please explain cardinality a bit more in this doc.

compiler/src/org.graalvm.compiler.jtt/src/org/graalvm/compiler/jtt/optimize/Switch03TestGenerator.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Random;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.graalvm.compiler.lir.hashing.HashFunction;
+import org.graalvm.compiler.lir.hashing.Hasher;
+
+import jdk.vm.ci.meta.JavaConstant;
+
+class Switch03TestGenerator {
```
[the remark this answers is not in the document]
dougxc: It's fine to leave it in. I would suggest making it a static nested class named `Generator` at the bottom of `Switch03.java` to make the connection as clear as possible.

fwbrasil:
@dougxc I've just pushed changes to address your comments and added a new test that will fail if the set of hash functions changes and the test is not updated.

tkrodriguez:
Why did the performance drop so much after your changes?  Was it executing incorrectly so it seemed faster than it really was?

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGenerator.java
```diff
@@ -449,29 +451,55 @@ public Variable emitForeignCall(ForeignCallLinkage linkage, LIRFrameState frameS
...
+
+        Optional<Hasher> hasher = hasherFor(keyConstants, minDensity);
+
+        double hashTableSwitchDensity = hasher.map(h -> keyCount / (double) h.cardinality()).orElse(0d);
+
         long valueRange = keyConstants[keyCount - 1].asLong() - keyConstants[0].asLong() + 1;
         double tableSwitchDensity = keyCount / (double) valueRange;
+
         /*
          * This heuristic tries to find a compromise between the effort for the best switch strategy
```
tkrodriguez: Do we still have a policy problem here?  The existing policy is hard to make sense of but it seems like it's not really making comparable trade offs between table switch and the other forms.  The table forms are only billed for their space requirements but the other forms are billed for their dynamic work and not for their code size.  Binary search for instance has a log scale effort but doubles in size each time which doesn't seem quite fair.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGenerator.java
```diff
@@ -449,29 +451,55 @@ public Variable emitForeignCall(ForeignCallLinkage linkage, LIRFrameState frameS
...
          * and the density of a tableswitch. If the effort for the strategy is at least 4, then a
          * tableswitch is preferred if better than a certain value that starts at 0.5 and lowers
          * gradually with additional effort.
          */
-        if (strategy.getAverageEffort() < 4 || tableSwitchDensity < (1 / Math.sqrt(strategy.getAverageEffort()))) {
+        if (strategy.getAverageEffort() < 4d || (tableSwitchDensity < minDensity && hashTableSwitchDensity < minDensity)) {
```
tkrodriguez: Shouldn't this be || instead of &&?  If the hashTableSwitchDensity is good but tableSwitchDensity isn't wouldn't we still want to use the hash?

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGenerator.java
```diff
@@ -449,29 +451,55 @@ public Variable emitForeignCall(ForeignCallLinkage linkage, LIRFrameState frameS
 
     @Override
     public void emitStrategySwitch(JavaConstant[] keyConstants, double[] keyProbabilities, LabelRef[] keyTargets, LabelRef defaultTarget, Variable value) {
+
         int keyCount = keyConstants.length;
         SwitchStrategy strategy = SwitchStrategy.getBestStrategy(keyProbabilities, keyConstants, keyTargets);
+
```
tkrodriguez: too much vertical whitespace.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64ControlFlow.java
```diff
@@ -283,9 +283,7 @@ public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
             masm.jmp(scratchReg);
 
             // Inserting padding so that jump table address is 4-byte aligned
-            if ((masm.position() & 0x3) != 0) {
-                masm.nop(4 - (masm.position() & 0x3));
-            }
+            masm.align(4);
```
tkrodriguez: I think you might want to align to 8 bytes if defaultTarget != null since each entry is 8 bytes wide.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGenerator.java
fwbrasil: yes, this heuristic needs to be improved. I'm planning to create a new PR with a better one.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGenerator.java
fwbrasil: The condition seems correct: 
- if both tables have density lower than `minDensity`, use the strategy switch
- if one of them is higher than `minDensity` , use one of the table switches

tkrodriguez: Yes of course you're right.  I was just misreading the logic.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64ControlFlow.java
fwbrasil: That's the alignment for the regular table switch, it hasn't changed. Here's the alignment for the hash table switch: https://github.com/oracle/graal/pull/895/files/7341ab80f998290de7310f393b4191a9f509813c#diff-a8baf9fe5005ee604d024bb2e35c173aR371

dougxc:
All my comments have been sufficiently addressed - thanks!

fwbrasil:
@dougxc @tkrodriguez @christhalinger thank you for the reviews! I believe all review comments were addressed, please let me know if there's anything pending to merge this change.

fwbrasil:
any updates on this? I'd like to create the second PR with a better heuristic to chose the switch strategy

fwbrasil:
@dougxc @tkrodriguez since I've planned opening multiple PR for https://github.com/oracle/graal/issues/893 and github doesn't support dependent pull requests, I'm currently blocked. Could you clarify how the merge process work so I can plan accordingly? I imagine it's not as simple as just hitting the merge button here on github.

dougxc:
@tkrodriguez can you please integrate this PR.

tkrodriguez:
Sorry I hadn't responded earlier.  It has to be run through out internal gate which does a much more extensive set of tests.  I think there's also a question of whether this is a win overall even though it looks good on a microbenchmark.  I'll run our benchmarks and see how it looks as well.  So it requires a bit more attention from me and I've been juggling a few other things.  Sorry for not communicating that.

You can always create a new PR that includes these changes and then merge and rebase it away once it lands.  Git makes that very easy.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/ArithmeticLIRGeneratorTool.java
```diff
@@ -44,6 +44,8 @@
  */
 public interface ArithmeticLIRGeneratorTool {
 
+    LIRGenerator getLIRGen();
```
tkrodriguez: Can we remove this override?  It sort of defeats the purpose of the interface to expose this here.  I think replacing ArithmeticLIRGeneratorTool with ArithmeticLIRGenerator in Hasher and HashFunction resolve the problem without requiring this override.

fwbrasil: good point! it does work, I've removed the override

fwbrasil:
@tkrodriguez branch rebased

compiler/src/org.graalvm.compiler.jtt/src/org/graalvm/compiler/jtt/optimize/Switch03.java
```diff
@@ -0,0 +1,635 @@
...
+import org.graalvm.compiler.lir.hashing.Hasher;
+import org.junit.Test;
+
+import jdk.vm.ci.meta.JavaConstant;
+
+/*
+ * Tests optimization of hash table switches.
+ * Code generated by `Switch03.TestGenerator.main`
+ */
+public class Switch03 extends JTTTest {
```
tkrodriguez: Please use a more descriptive name since this is specifically about the hash switch, like HashSwitchTest.  Also is there any reason this is a JTTest?  These are special variants of regular compiler tests that are exercised with their parameters bounds as constants in addition to a normal compile to test high level transforms.  I don't think you care about that here.  I think this should probably be over in org.graalvm.compiler.core.test.  Or if you wanted it to be a more specific test of the LIR you could make it a LIRTest and check for the existence of a HashTableSwitchOp in the final LIR.  Anyway, renaming is enough for now.  You can make any improvements in a follow on PR.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGenerator.java
```diff
@@ -449,29 +451,51 @@ public Variable emitForeignCall(ForeignCallLinkage linkage, LIRFrameState frameS
...
+                for (int i = 0; i < cardinality; i++) {
+                    keys[i] = JavaConstant.INT_0;
+                    targets[i] = defaultTarget;
+                }
+                for (int i = 0; i < keyCount; i++) {
+                    int idx = h.hash(keyConstants[i].asLong());
+                    keys[idx] = keyConstants[i];
+                    targets[idx] = keyTargets[i];
+                }
+                emitHashTableSwitch(hasher.get(), keys, defaultTarget, targets, value);
```
tkrodriguez: hasher.get is just h isn't it?

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/hashing/Hasher.java
```diff
@@ -0,0 +1,163 @@
...
+        for (int i = 1; i < keys.length; i++) {
+            assert keys[i - 1].asLong() < keys[i].asLong();
+        }
+    }
+
+    private static boolean isValid(JavaConstant[] keys, long min, HashFunction function, int cardinality) {
+        Set<Integer> seen = new HashSet<>(keys.length);
+        for (JavaConstant key : keys) {
+            int hash = function.apply(key.asLong(), min) & (cardinality - 1);
+            if (seen.contains(hash)) {
```
tkrodriguez: Can't this be simplified with the else removed?
```suggestion
            if (!seen.add(hash)) {
```

tkrodriguez:
The benchmark results look ok.  It doesn't trigger very often from what I can see.  Should we add an option to force it's use?  That could be in a follow on PR if you like.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/hashing/HashFunction.java
```diff
@@ -0,0 +1,170 @@
...
+
+        addWithPrimes("rotateRight(val, prime) ^ val", 4,
+                      prime -> (val, min) -> Long.rotateRight(val, prime) ^ val,
+                      (gen, prime) -> (val, min) -> gen.emitXor(rotateRight(gen, val, prime), val));
+      //@formatter:on
+    }
+
+    private static Value rotateRight(ArithmeticLIRGenerator gen, Value i, Value distance) {
+        // (i >>> distance) | (i << -distance)
+        return gen.emitOr(gen.emitUShr(i, distance), gen.emitShl(i, gen.emitNegate(distance)));
```
tkrodriguez: AMD64ArithmeticLIRGenerator provides emitRor which would be a lot more efficient.  Maybe provide a platform independent definition like above and use emitRor for AMD64?

fwbrasil: @tkrodriguez maybe I could add `emitRor` to `ArithmeticLIRGenerator` with the slow implementation as the default and then `AMD64ArithmeticLIRGenerator` overrides it with a better impl?

fwbrasil:
@tkrodriguez I've just addressed your last comments.

> The benchmark results look ok. It doesn't trigger very often from what I can see. Should we add an option to force it's use? That could be in a follow on PR if you like.

Yes, I have a branch with a new heuristic that will make the hash table switch more competitive. I'll create the PR once this change is merged. The binary strategy has a large code footprint, so the hash table strategy will be chosen in some cases where the binary one is chosen at the moment.

tkrodriguez:
Ok.  I'm going to do some copyright data adjustments in a new commit and push it.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/hashing/HashFunction.java
```diff
@@ -0,0 +1,165 @@
...
+                return toString;
+            }
+
+            @Override
+            public Value gen(Value val, Value min, ArithmeticLIRGenerator t) {
+                return gen.apply(t).apply(toDWORD(val, t), toDWORD(min, t));
+            }
+
+            private Value toDWORD(Value v, ArithmeticLIRGenerator t) {
+                if (v.getPlatformKind() != AMD64Kind.DWORD) {
```
tkrodriguez: This use of AMD64Kind in a shared file should be fixed in the next PR.  I would in general expect the values to come into this function with the right sizing of either int or long with the caller ensuring that it was at least those types already.

dougxc: Why wait for the next PR? Best to fix this here.

fwbrasil: Let me fix it. Should I create a new PR with the fix or are you going to reopen this one?

tkrodriguez: Create a new PR.

PR #923, opened 2019-01-18
Improve memory usage

n1hility:
# Memory usage improvements 

## Changelog

- 1/21/19 - Added Darwin COW support
- 1/21/19 - Dropped commit that changed SpawnIsolates default to false for executables
- 1/23/19 - Updated Linux COW to utilize overlapped anon relocatable section instead of patching, removed UseCOWIsolateHeaps,  removed UseCompressedReferences & spawn isolates usage (will continue discussion around that change separately)

## Overview

This pull request includes a set of changes to improve the memory usage of native images. I have broken these changes into ~3~ 2 separate commits to make it easier to review, and also make it easier to apply in piecemeal if any are contentious. If preferred, I can break these into separate PRs, just let me know. Also, I am happy to address any other concerns raised. The changes in this PR (justifications further down) are:

 1. The introduction of a new copy-on-write Linux implementation for efficient isolate heap image construction, when SpawnIsolates is enabled.
 2. The introduction of a new Darwin implementation for COW isolate heap image construction.
 3. ~The resurrection of UseHeapBase as UseCompressedReferences (defaulted on) to enable more efficient heap usage when SpawnIsolates is disabled.~

## New COW LinuxImageHeapProvider

Currently, when enabling isolates (the default since RC7), the CopyImageHeap provider constructs a new anonymous memory segment, and performs a user-space copy of the heap image embedded within the native executable/library into the new segment. This has two main drawbacks: one, it defeats lazy loading provided by the memory mapped region, forcing the full heap area to load, and two, the new anonymous backed copy won't be freed until swapping occurs. Compared to isolates disabled, this leads to an increase from r to (t - r) + t usage where r is the subset in use, and t is the total heap image.  

The newly introduced LinuxImageHeapProvider in this patch addresses this problem by creating a duplicate memory mapped segment for the isolate heap portion of the executable. The kernel is then able to share the underlying physical pages in a copy-on-write fashion. The resulting memory usage is comparable to when SpawnIsolates is disabled (see demonstration section below).  Due to the lack of a general interface for this purpose, the implementation is OS specific and includes a small portion of machine architecture specific code that is intended to be easy to expand.  It also falls back in the case of any failure to the standard copy provider.

## New COW DarwinImageHeapProvider

The similar functionality as above, but achieved on Darwin. This utilizes mach's vm_copy kernel interface which maps COW pages instead of performing an actual copy. 

On the Windows side, I have not looked into whether a COW strategy is possible, but from glancing at the WIN API set, it appears to have the hooks required. I plan to contribute this in the future, but due to potential reporting tool perception issues and to squeeze out a few more CPU cycles, I am also advocating for the subsequent changes listed.

## ~~Resurrection of UseHeapBase as UseCompressedReferences~~

~~A simple solution to the additional memory consumption imposed by SpawnIsolates is to just disable it. However, disabling it also has the negative side-effect of disabling compressed pointers as well.  Previously this was two separate settings, which I assume were collapsed to simplify complexity and usability. Hopefully the information included in this synopsis is enough to justify restoring it.  Keeping to that assumption, this change also renames the setting to match a familiar setting on OpenJDK, and it defaults it to true, which I argue is the more desirable value that a user should rarely need to touch. This also has the additional benefit of keeping the separate code paths of SpawnIsolates in the off and on case a bit closer than they were before this change.~~

## Example Linux Reproducer

```java
public class Test {
    private static int[] array = new int[5000000];    

    static {
        for (int i = 0; i < 5000000; i++) {
            array[i]=i;
        }
    }

    public static void main(String[] args)  throws Exception {
        int x = 0;
        for (int i = 0; i < 100; i++) {
           x+=array[(i > 50) ? i : (4999949+i)];
        }

        new ProcessBuilder("/bin/sh",  "-c", " ps aux | grep test | grep -v grep | awk '{print \"RSS(K): \"$6}'")
            .redirectOutput(ProcessBuilder.Redirect.INHERIT)
            .start().waitFor();

        System.out.println();
        new ProcessBuilder("/bin/sh",  "-c", " pmap -x `ps aux | grep test | grep -v grep | awk '{print $2}'`")
            .redirectOutput(ProcessBuilder.Redirect.INHERIT)
            .start().waitFor();


        new java.io.FileOutputStream("/dev/null").write(x);
    }
}
```

### Running with todays defaults (SpawnIsolates enabled, no COW)

This example consumes ~ 40MB of heap. The output below is annotated for clarity.

```
$ mx native-image Test -H:-UseCOWIsolateHeaps -H:+SpawnIsolates && ./test
RSS(K): 43728
 
15550:   ./test
Address           Kbytes     RSS   Dirty Mode  Mapping
0000000000400000   21920   21856   21856 r-x-- test         <-- COPY OPERATION FORCED FULL LOADING
0000000001b67000       4       4       4 r---- test
0000000001b68000      12      12      12 rw--- test
00000000021ee000     132       4       4 rw---   [ anon ]
00007f96dc000000     132       4       4 rw---   [ anon ]
00007f96dc021000   65404       0       0 -----   [ anon ]
00007f96e1e00000    1024       8       8 rw---   [ anon ]
00007f96e2000000    1024       8       8 rw---   [ anon ]
00007f96e2200000    1024      88      88 rw---   [ anon ]
 
00007f96e2309000     872     872     872 r----   [ anon ]   <--- COPIED ISOLATE HEAP AREAS HERE
00007f96e23e3000   19824   19824   19824 rw---   [ anon ]   <--- AND HERE, FULLY LOADED
 
-snip-
```

### Running with COW enabled and SpawnIsolates enabled

This scenario results in ~ 3MB. Note, with this PR these are the default settings.

```
$ mx native-image Test -H:+UseCOWIsolateHeaps -H:+SpawnIsolates && ./test

RSS(K): 2832

15239:   ./test
Address           Kbytes     RSS   Dirty Mode  Mapping
0000000000400000   21924    1216    1216 r-x-- test     <---- ONLY A FEW PAGES LOADED
0000000001b68000       4       4       4 r---- test
0000000001b69000      12      12      12 rw--- test
0000000001ba0000     132       8       8 rw---   [ anon ]
00007f20ac000000     132       4       4 rw---   [ anon ]
00007f20ac021000   65404       0       0 -----   [ anon ]
00007f20b2200000    1024       8       8 rw---   [ anon ]
00007f20b2400000    1024       8       8 rw---   [ anon ]
00007f20b2600000    1024      88      88 rw---   [ anon ]

00007f20b278b000     872     404     404 r---- test     <---- MMAPPED COW AREAS HERE
00007f20b2865000   19824     324     324 rw--- test     <---- AND HERE, PARTIALLY LOADED

-snip-
```

### Running with SpawnIsolates disabled

This scenario results in similar numbers to the COW + Isolates scenario, ~ 3MB of usage. 

```
$ mx native-image Test -H:-SpawnIsolates && ./test
RSS(K): 2796                                           <---- COW HAS SIMILAR RESULTS WITH ISOLATES DISABLED

15878:   ./test
Address           Kbytes     RSS   Dirty Mode  Mapping
0000000000400000    1220    1152    1152 r-x-- test    <---- COW HAS SIMILAR RESULTS WITH ISOLATES DISABLED
0000000000730000       4       4       4 r---- test
0000000000731000   20700     736     736 rw--- test
00000000030f9000     132       4       4 rw---   [ anon ]
00007fef34000000     132       4       4 rw---   [ anon ]
```

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/darwin/DarwinImageHeapProvider.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/darwin/DarwinImageHeapProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -22,42 +22,52 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
-package com.oracle.svm.core.os;
+package com.oracle.svm.core.posix.darwin;
 
 import static com.oracle.svm.core.Isolates.IMAGE_HEAP_WRITABLE_BEGIN;
 import static com.oracle.svm.core.Isolates.IMAGE_HEAP_WRITABLE_END;
 import static com.oracle.svm.core.util.PointerUtils.roundUp;
 
+import com.oracle.svm.core.Isolates;
+import com.oracle.svm.core.MemoryUtil;
+import com.oracle.svm.core.SubstrateOptions;
 import com.oracle.svm.core.annotate.AutomaticFeature;
+import com.oracle.svm.core.annotate.Uninterruptible;
+import com.oracle.svm.core.c.function.CEntryPointErrors;
+import com.oracle.svm.core.os.CopyingImageHeapProvider;
+import com.oracle.svm.core.os.ImageHeapProvider;
+import com.oracle.svm.core.os.VirtualMemoryProvider;
+import com.oracle.svm.core.os.VirtualMemoryProvider.Access;
+import com.oracle.svm.core.posix.headers.darwin.DarwinVirtualMemory;
+import com.oracle.svm.core.util.UnsignedUtils;
 import org.graalvm.compiler.word.Word;
 import org.graalvm.nativeimage.Feature;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.Platform;
 import org.graalvm.nativeimage.Platforms;
+import org.graalvm.nativeimage.StackValue;
 import org.graalvm.nativeimage.c.type.WordPointer;
 import org.graalvm.word.Pointer;
 import org.graalvm.word.PointerBase;
 import org.graalvm.word.UnsignedWord;
 
-import com.oracle.svm.core.Isolates;
-import com.oracle.svm.core.MemoryUtil;
-import com.oracle.svm.core.annotate.Uninterruptible;
-import com.oracle.svm.core.c.function.CEntryPointErrors;
-import com.oracle.svm.core.os.VirtualMemoryProvider.Access;
-import com.oracle.svm.core.util.UnsignedUtils;
-
 @AutomaticFeature
-@Platforms({Platform.WINDOWS.class})
-class CopyingImageHeapProviderFeature implements Feature {
+@Platforms({Platform.DARWIN.class})
+class DarwinImageHeapProviderFeature implements Feature {
     @Override
     public void beforeAnalysis(BeforeAnalysisAccess access) {
         if (!ImageSingletons.contains(ImageHeapProvider.class)) {
-            ImageSingletons.add(ImageHeapProvider.class, new CopyingImageHeapProvider());
+            ImageSingletons.add(ImageHeapProvider.class, new DarwinImageHeapProvider());
         }
     }
 }
 
-public class CopyingImageHeapProvider implements ImageHeapProvider {
+
+/**
+ * An optimal image heap provider for Darwin which creates isolate image heaps
+ * that are a copy-on-write clone of the original image heap.
+ */
+public class DarwinImageHeapProvider implements ImageHeapProvider {
     @Override
     @Uninterruptible(reason = "Called during isolate initialization.")
     public int initialize(PointerBase begin, UnsignedWord reservedSize, WordPointer basePointer, WordPointer endPointer) {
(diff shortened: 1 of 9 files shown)


n1hility:
Hello @graalvmbot. There is Red Hat wide OCA already signed. I believe I also signed an individual one long ago.

bobmcwhirter:
Well dang.

christianwimmer:
Thanks for the contribution, we will have a detailed look at the memory mapping providers soon.

Note that isolates are an important part of our strategy for executables too, e.g., we are working to have Truffle JIT compilation in a separate isolate for the JavaScript executable. So turning off `SpawnIsolates` by default is not what we want.

Also, compressed references always require isolates because compressed references require the heap base in a register. So your statement "However, disabling [SpawnIsolates] also has the negative side-effect of disabling compressed pointers as well" is true but changing flags is not going to help.

n1hility:
> Thanks for the contribution, we will have a detailed look at the memory mapping providers soon.
> 
> Note that isolates are an important part of our strategy for executables too, e.g., we are working to have Truffle JIT compilation in a separate isolate for the JavaScript executable. So turning off `SpawnIsolates` by default is not what we want.
> 
> Also, compressed references always require isolates because compressed references require the heap base in a register. So your statement "However, disabling [SpawnIsolates] also has the negative side-effect of disabling compressed pointers as well" is true but changing flags is not going to help.

n1hility:
Hi @christianwimmer,

Thanks again for looking into this.  I removed the commit with the default change. Within my PR bde04f9 heap bp seems to be working as I would expect with SpawnIsolates off. I double checked the call paths from SubstrateOptions.SpawnIsolates that I did not already convert to the resurrected UseHeapBase, and nothing stands out. 

```text
0x100001217 <+23>: movabsq $0x13e94b8, %rax          ; imm = 0x13E94B8      
0x100001221 <+33>: leaq   (%r14,%rax), %rax
0x100001225 <+37>: movl   $0x4, 0x38(%rax)

r14 = 0x000000010012e000  test`__svm_heap
```

peter-hofer:
Thank you @n1hility for this contribution!

I am in the process of importing your commits and running them against our test suite. We have also discussed the topic of options internally. We want copy-on-write cloning to be enabled whenever available, so we will not introduce an additional option. We also prefer not to resurrect `UseHeapBaseRegister`, as it adds little value without `SpawnIsolates` (and vice versa), which is the reason why we recently removed it when tidying up options.

I also have the following two remarks, but will happily make the necessary changes myself:
- We put all objects that contain relocations in their own heap partition, so we can simply copy the entire partition with all relocated values from the loader-mapped heap and therefore avoid traversing ELF data structures and locating values individually.
- It should be possible to simplify the `/proc/self/maps` parser for the task it is used for.

n1hility:
> Thank you @n1hility for this contribution!

Hi @peter-hofer 

Your welcome. Thank you for reviewing this patch.

>  I am in the process of importing your commits and running them against our test suite. 

That sounds great.


> We have also discussed the topic of options internally. We want copy-on-write cloning to be enabled whenever available, so we will not introduce an additional option. 

Sure I agree that it's unlikely anyone would not want it. The only reason I added a flag was to be able to disable it in the case of a problem. However, the fallback mechanism makes it less important. On that topic Did you have an opinion on the fallback aspect? It could just hard fail since a failure is unlikely.

>We also prefer not to resurrect `UseHeapBaseRegister`, as it adds little value without `SpawnIsolates` (and vice versa), which is the reason why we recently removed it when tidying up options.

Speaking as a user the potential value of disabling SpawnIsolates was simply that we didn't need isolates and didn't want the extra memory overhead it was adding. However this option wasn't great either since it increases heap usage from it's lack of compressed pointers. Since you are accepting patches to introduce COW support, I admittedly have less of a need for the option, since both elements are addressed. Although I have to ask what's the value of SpawnIsolates as an option going forward at all? legacy? Just supporting gigantic heaps?


> 
> I also have the following two remarks, but will happily make the necessary changes myself:

Ok If I understand correctly you prefer to just make these updates yourself? I don't mind if you want me to update this since I took your time to a) review the patch and b) run your testsuite against it.

> * We put all objects that contain relocations in their own heap partition, so we can simply copy the entire partition with all relocated values from the loader-mapped heap and therefore avoid traversing ELF data structures and locating values individually.

Ah yes of course. I wasn't sure if I could rely on that layout.

> * It should be possible to simplify the `/proc/self/maps` parser for the task it is used for.

Ok. The reason it was implemented as a state machine was just to avoid allocations, avoid non-transitional method calls, and minimize sys calls + c library calls to avoid any impact on boot/setup time.

n1hility:
> No worries, I have already started on this and will get back to you with any questions. Thank you!

Ok FWIW I pushed up change for an overlapping mmap segment along with dropping the flags aspect a little while ago.

peter-hofer:
We have merged your contribution now, thanks again!

PR #910, opened 2019-01-14
SubstrateVM on AArch64

sanzinger:
### Current status:

The state of this PR is in progress. Native-image command cannot produce any working binary for now.

### Testing

For testing I use the simple hello world example from substrate/README.md quickstart:

```
cd substratevm
mx build

echo "public class HelloWorld { public static void main(String[] args) { System.out.println(\"Hello World\"); } }" > HelloWorld.java
$JAVA_HOME/bin/javac HelloWorld.java
mx native-image HelloWorld -H:-SpawnIsolates
./helloworld
```

The command `mx native-image HelloWorld` fails currently with following error:
```
fatal error: com.oracle.svm.core.util.VMError$HostedError: currently only implemented on AMD64
	at com.oracle.svm.core.util.VMError.guarantee(VMError.java:85)
	at com.oracle.svm.jni.hosted.JNICallTrampolineMethod.lambda$createCustomCompileFunction$1(JNICallTrampolineMethod.java:111)
	at com.oracle.svm.hosted.code.CompileQueue.doCompile(CompileQueue.java:883)
	at com.oracle.svm.hosted.code.CompileQueue$CompileTask.run(CompileQueue.java:259)
	at com.oracle.graal.pointsto.util.CompletionExecutor.lambda$execute$0(CompletionExecutor.java:174)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1426)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
```

PR #929, opened 2019-01-21
This commit begins support for AArch64 native-image support.

bobmcwhirter:
* cpuid.c no-op for non-AMD64 machine (via sanzinger; cherry-picked)
* ELFMachine expansion for AArch64 type and elf_machine magic.
* Provides direct relocation types for AArch64.
* Allows ELFMachine selection based on os.arch system property
* (unused) provides for ELFMachine to be specified in the ELFObjectFile constructor
  anticipating a future cross-compilation support, instead of determining
  the type based upon the host architecture.


sanzinger:
part of #910

PR #930, opened 2019-01-21
Bring in implementation of compressed pointers for AArch64.

bobmcwhirter:
* Prepares for an SVM AArch64 backend.
* Mostly transliterating the Hotspot variant of compressed-pointers.
* Stub in (unused) boolean annotateImmediate for masm.mov(..)
  in preparation for instruction-patching/relocations.


bobmcwhirter:
I’ll fix that unused param.

sanzinger:
This change will be part of the general SVM/AArch64 #910

PR #932, opened 2019-01-21
Support for JMH Benchmarks in Native Images

flortsch:
This PR allows SubstrateVM to build native images out of JAR files containing JMH benchmarks. The list of benchmark classes generated by JMH is parsed out of the JAR file and the classes are registered for reflection at runtime. Also some small substitutions are made in order to make JMH happy. This is all implemented in features which can be enabled at image build time using ```--tool:jmh``` as command line argument to ```mx native-image```. Example: ```mx native-image --tool:jmh --jar benchmarks.jar && ./benchmarks```, where ```benchmarks.jar``` is a JAR file created by JMH.

Note that JMH benchmarks are only executed in embedded mode, i.e., the mode where JMH forks and spawns new VMs is not supported. This is because JMH communicates with these VMs via sockets and uses the Java Serialization API which is not supported in SubstrateVM. A solution would be to patch JMH so that is uses a different serialization technique for it's internal Frame objects which are sent over sockets. Then, a second native image could be built for the benchmark runner, and the main runner could be substituted to run the native image instead of spawning a new VM.

Feedback is highly appreciated, especially from JMH experts, as the image building in SubstrateVM might have an influence on the magic that happens inside JMH (which was not designed for such environments I guess).


substratevm/src/com.oracle.svm.jmh/src/com/oracle/svm/jmh/JMHSubstitutions.java
```diff
@@ -0,0 +1,83 @@
...
+        return BenchmarkList.fromString(Benchmarks.string());
+    }
+
+}
+
+/**
+ * Substitutes {@link Utils#getRecordedSystemProperties()} to set up properties with non-null
+ * values, as Substrate currently returns null for some system properties.
+ */
+@TargetClass(Utils.class)
```
olpaw: Instead of providing this substitution for specific system properties we should make underlying properties available by adjusting com.oracle.svm.core.jdk.SystemPropertiesSupport

substratevm/src/com.oracle.svm.jmh/src/com/oracle/svm/jmh/JMHSubstitutions.java
```diff
@@ -0,0 +1,83 @@
...
+final class Target_org_openjdk_jmh_util_Utils {
+
+    @Substitute
+    public static Properties getRecordedSystemProperties() {
+        Properties properties = new Properties();
+        properties.put("java.vm.name", System.getProperty("java.vm.name"));
+
+        // set up custom values as Substrate returns null for these properties
+        properties.put("java.version", "unknown");
+        properties.put("java.vm.version", "unknown");
```
olpaw: We already have `java.vm.version` (expands to `"Substrate VM " + System.getProperty("java.vm.version")`)

substratevm/src/com.oracle.svm.jmh/src/com/oracle/svm/jmh/JMHSubstitutions.java
```diff
@@ -0,0 +1,83 @@
...
+ * Substitutes {@link Utils#getRecordedSystemProperties()} to set up properties with non-null
+ * values, as Substrate currently returns null for some system properties.
+ */
+@TargetClass(Utils.class)
+final class Target_org_openjdk_jmh_util_Utils {
+
+    @Substitute
+    public static Properties getRecordedSystemProperties() {
+        Properties properties = new Properties();
+        properties.put("java.vm.name", System.getProperty("java.vm.name"));
```
olpaw: `java.vm.name` automatically expands to `"Substrate VM"`

substratevm/src/com.oracle.svm.jmh/src/com/oracle/svm/jmh/JMHSubstitutions.java
```diff
@@ -0,0 +1,83 @@
...
+
+        // set up custom values as Substrate returns null for these properties
+        properties.put("java.version", "unknown");
```
olpaw: ... and java.version is taken over from the corresponding hosted value. I.e. the whole `Target_org_openjdk_jmh_util_Utils` can be removed.

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -327,6 +327,7 @@ def __init__(self, image_deps=None, builder_deps=None, native_deps=None):
     'native-image' : ToolDescriptor(),
     'junit' : ToolDescriptor(builder_deps=['mx:JUNIT_TOOL', 'JUNIT', 'HAMCREST']),
     'regex' : ToolDescriptor(image_deps=['regex:TREGEX']),
+    'jmh' : ToolDescriptor(),
```
olpaw: Ideally we would not need this and instead annotate the two features with `@AutomaticFeature` and use a shared `com.oracle.svm.hosted.ServiceLoaderFeature#isInConfiguration` implementation to decide if the features should be active or not. @christianwimmer do you agree?

substratevm/src/com.oracle.svm.jmh/src/com/oracle/svm/jmh/JMHSubstitutions.java
flortsch: Fixed with 01bd194

substratevm/mx.substratevm/mx_substratevm.py
olpaw: A good example for auto enabling is com.oracle.svm.polyglot.kotlin.KotlinFeature

tshull226:
Hi. Whatever happened with this PR? Would JMH work well enough with SVM for this to be useful? If so, I potentially would be willing to prepare this PR to be merged.

flortsch:
Sorry for the late reply. I think supporting JMH in SVM would be very useful, so we can use the same known benchmarking framework for different configurations/deployments of Graal. However, there are still some open issues I have pointed out in my first post, which should be discussed before doing further work in this direction.

christianwimmer:
What we have realized in the last year is that is not feasible to have support for individual libraries and frameworks in the main GraalVM / Native Image repository. Instead, we now have the mechanism that libraries themselves contain `native-image.properties` files and `Feature` code that enables Native Image support. Or if the support is not in the library itself, have a separate library that adds the Native Image support. This does not change how the support for something like JMH is implemented, only how it is distributed.
For JMH I don't think this suppport code has been published though.

cstancu:
As @christianwimmer detailed above this code doesn't belong in the GraalVM repo. I'm closing this PR.

PR #938, opened 2019-01-28
jdk.internal.util.ArraysSupport.vectorizedMismatch intrinsic (JDK-8217161)

rwestrel:
…7161)
From the openjdk bug report:
Currently the vectorizedMismatch intrinsic is in the toBeInvestigated list: 

./share/classes/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/CheckGraalIntrinsics.java 

There is a JMH micro for testing ArraysMismatch: open/test/micro/org/openjdk/bench/java/util/ArraysMismatch.java 

Implementing this in the longest case would be a ~150% speedup.

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/CheckGraalIntrinsics.java
+++ b/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/CheckGraalIntrinsics.java
@@ -277,8 +277,6 @@ public class CheckGraalIntrinsics extends GraalTest {
                             "java/lang/Math.fma(FFF)F",
                             // Just check if the argument is a compile time constant
                             "java/lang/invoke/MethodHandleImpl.isCompileConstant(Ljava/lang/Object;)Z",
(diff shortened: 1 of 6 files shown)


rwestrel:
Regarding the oca: I'm a Red Hat employee which is of one of the oca signatories.

dougxc:
I get this failure on SPARC:
```
test(org.graalvm.compiler.hotspot.test.CheckGraalIntrinsics)
java.lang.AssertionError: missing Graal intrinsics for:
    jdk/internal/util/ArraysSupport.vectorizedMismatch(Ljava/lang/Object;JLjava/lang/Object;JII)I

	at org.graalvm.compiler.hotspot.test.CheckGraalIntrinsics.test(CheckGraalIntrinsics.java:626)
```
For arch specific intrinsics, you need to test the type of `CheckGraalIntrinsics.arch` when modifying the various lists.

rwestrel:
Thanks for looking at this, Doug. New change should fix this.

dougxc:
Looks good. This PR is going through the internal merge process now.

Issue #914, opened 2019-01-15
[native-image] Concurrent calls InetAddress.getAllByName hang forever


qdeee:
I'm trying to make concurrent calls to InetAddress.getAllByName in my native image. However, some of them hang forever in WAITING state on java.net.InetAddress.checkLookupTable. 
OS: linux
version: RC10, RC11

I've written minimal example with distinct "StackTracePrinter" thread.
```

$ native-image InetAddressResolveTest --no-server -Djava.net.preferIPv6Addresses=true
[inetaddressresolvetest:686759]    classlist:   1,298.65 ms
[inetaddressresolvetest:686759]        (cap):   1,207.02 ms
[inetaddressresolvetest:686759]        setup:   2,811.73 ms
[inetaddressresolvetest:686759]   (typeflow):   9,518.98 ms
[inetaddressresolvetest:686759]    (objects):  13,274.99 ms
[inetaddressresolvetest:686759]   (features):     558.97 ms
[inetaddressresolvetest:686759]     analysis:  23,807.75 ms
[inetaddressresolvetest:686759]     universe:     457.21 ms
[inetaddressresolvetest:686759]      (parse):   1,210.57 ms
[inetaddressresolvetest:686759]     (inline):   1,275.90 ms
[inetaddressresolvetest:686759]    (compile):   7,272.05 ms
[inetaddressresolvetest:686759]      compile:  10,914.39 ms
[inetaddressresolvetest:686759]        image:   1,940.33 ms
[inetaddressresolvetest:686759]        write:     307.57 ms
[inetaddressresolvetest:686759]      [total]:  41,663.40 ms

# it hangs forever:
$ ./inetaddressresolvetest google.com 2 2

```
Now in another console:
```

$ nc localhost 8082
thread: pool-1-thread-1 (WAITING)
trace:
com.oracle.svm.core.posix.headers.Pthread.pthread_cond_wait(Pthread.java)
com.oracle.svm.core.posix.thread.PosixParkEvent.condWait(PosixJavaThreads.java:266)
com.oracle.svm.core.thread.UnsafeParkSupport.park(JavaThreads.java:1039)
sun.misc.Unsafe.park(Unsafe.java:1154)
java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
[...]

PR #950, opened 2019-01-31
Use custom WeakIdentityHashMap for locks and conditions in MonitorSupport

qdeee:
Current implementation of MonitorSupport class has a few bugs because of WeakHashMap usage for storing locks and conditions:
1. You can't `wait` on mutable objects (e.g. collections) because hash code of the object can change, and `notify` won't find corresponding condition in `additionalConditions` map.
2. You get unexpected behaviour with equals semantics. For example, next code will sleep forever in regular java but it completes normally in native-image:
```java

public static void main(String... args) throws Exception {
    new Thread(() -> {
      String s = new String("blabla");
      synchronized (s) {
        try {
          s.wait();
        } catch (Exception e) {
        }
      }
    }).start();

    Thread.sleep(3000);

    new Thread(() -> {
      String s = new String("blabla");
      synchronized (s) {
        try {
          s.notifyAll();
        } catch (Exception e) {
        }
      }
    }).start();
}

``` 
(there are two distinct objects "s" but they have the same hashCode and equals, so MonitorSupport will create the only condition for them).

"WeakHashMap" problem was discovered by @Peter-B-Kessler in the issue #914 (see for more details).

To address this problem I introduced new WeakIdentityHashMap class. It's mostly adapted from java.util.WeakHashMap but it uses `System.identityHashCode(Object)` for hash code calculation and reference-equality for comparing objects.

Just in case I uploaded the diff for my WeakIdentityHashMap and java.util.WeakHashMap: https://github.com/qdeee/graal/blob/WeakIdentityHashMap.diff/WeakIdentityHashMap.diff . I hope it will simplify the review process.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/MonitorSupport.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/MonitorSupport.java
@@ -25,7 +25,6 @@
 package com.oracle.svm.core;
 
 import java.util.Map;
-import java.util.WeakHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.AbstractOwnableSynchronizer;
 import java.util.concurrent.locks.Condition;
@@ -47,6 +46,7 @@ import com.oracle.svm.core.snippets.SubstrateForeignCallTarget;
 import com.oracle.svm.core.thread.ThreadingSupportImpl.PauseRecurringCallback;
 import com.oracle.svm.core.thread.VMOperation;
 import com.oracle.svm.core.util.VMError;
+import com.oracle.svm.core.util.WeakIdentityHashMap;
 
 /**
  * Implementation of synchronized-related operations.
@@ -75,7 +75,7 @@ public class MonitorSupport {
      *
      * Synchronized to prevent concurrent access and modification.
      */
-    private final Map<Object, ReentrantLock> additionalMonitors = new WeakHashMap<>();
+    private final Map<Object, ReentrantLock> additionalMonitors = new WeakIdentityHashMap<>();
     private final ReentrantLock additionalMonitorsLock = new ReentrantLock();
 
     /**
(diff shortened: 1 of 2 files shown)


Peter-B-Kessler:
We had the same idea.  I think I have essentially the same clone of `WeakHashMap` working as an `WeakIdentityHashMap`.  Wow that diff is hard to read.  Let me try looking at it a different way.  (Also, I may have to make some syntactic changes to get it to pass our style guidelines (missing braces, more than one declaration per line, etc), unless I can find a way to suppress them.

Peter-B-Kessler:
And, of course, I can not put your code in `package ru.yandex.qe.graal.utils`.

Peter-B-Kessler:
Re: https://github.com/oracle/graal/pull/950#issue-249423366 you will be delighted to hear that now the native image for your test hangs, just like HotSpot.

Seriously: Thanks for the test programs!

qdeee:
> Dare I ask how you started from a `WeakHashMap.java` whose first lines are
> 
> ```
>  <  * Copyright (c) 1998, 2013, Oracle and/or its affiliates. All rights reserved.
>  <  * ORACLE PROPRIETARY/CONFIDENTIAL. Use is subject to license terms.
> ```
> !? You changed the license to UPL/v1.0. Now I going to have to ask _lawyers_ about your code. Sigh.

I'm sorry, I'm not an expert in licenses. I took new header with UPL in neighbour SVM classes. I will replace it with required text as soon as you tell me how to do it right.
The source code of WeakHashMap I took from Intellij IDEA that extracted it for me from installed JDK.

qdeee:
> And, of course, I can not put your code in package ru.yandex.qe.graal.utils.

 Sorry, it was only in the diff. I fixed it in updated version.
Source code contains right package.

Peter-B-Kessler:
How about if I say "Thank you very much for your contributions", but push my version of the changes?  

In addition to the functional changes I added a bunch of comments about where WeakIdentityHashMap.java comes from and why it has the changes (all the changes!) it has.  I may also have to wrestle with checkstyle violations, which I can not ask you to do.  I also have to write unit tests for WeakIdentityHashMap, and you do not have our testing harness.

That way you won't have to spend any more time on this: now that we understand the problem.

qdeee:
> How about if I say "Thank you very much for your contributions", but push my version of the changes?

It's ok. I only hope that changes will appear in RC12 :)
Anyway, thanks you for comments!

Peter-B-Kessler:
I think RC12 has left the station.  But there will be an RC13.

PR #981, opened 2019-02-15
[WIP] Add substitutions for annotation constructors

johanvos:


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/annotation/AnnotationSupport.java
```diff
@@ -158,6 +158,11 @@ private synchronized AnnotationSubstitutionType getSubstitution(ResolvedJavaType
                 }
                 result.addSubstitutionMethod(originalMethod, substitutionMethod);
             }
+            for (ResolvedJavaMethod originalMethod : type.getDeclaredConstructors()) {
+                AnnotationSubstitutionMethod substitutionMethod = new AnnotationAccessorMethod(originalMethod);
+                result.addSubstitutionField(new AnnotationSubstitutionField(result, originalMethod, snippetReflection, metaAccess));
```
vjovanov: Why do we need the field here?

johanvos: We don't. I modified that substitution into a DeadNode substitution to see why this is called at runtime. However, this generates another exception during image generation:

`java.lang.reflect.InaccessibleObjectException: Unable to make public final java.lang.String com.sun.proxy.jdk.proxy1.$Proxy117.value() accessible: module jdk.proxy1 does not "exports com.sun.proxy.jdk.proxy1" to unnamed module @59c38d4`

In order to isolate this problem, it would be good to intercept the creation of the $Proxy implementations.

vjovanov: This seems like another issue related to JDK 11. I would suspect it will happen here `com.oracle.svm.reflect.proxy.DynamicProxySupport` and here `com.oracle.svm.core.jdk.proxy.Target_java_lang_reflect_Proxy#getProxyClass0`.

Now I see that it is done for the JDK 8 only. @cstancu @Peter-B-Kessler is there any work-in-progress on proxies for JDK 11?

johanvos:
A clean PR for this is #1001

Issue #942, opened 2019-01-28
MatchContext should distinguish between nodes which are a pure function and side effecting ones


tkrodriguez:
MatchContext.validate tries to ensure that there isn't any unsafe work in between the nodes that are consumed as part of a match.  The current definition is very conservative because we don't have a clear model of which nodes are unsafe to skip over.  Looking at it from the other direction, for the nodes we match, we know their semantics and whether it's possible for other nodes to interact with them.  Many of the nodes are pure functions so they can't be affected by intervening nodes.  If we restrict the checked range to only visit from the root node to the last non-pure node then we could handle many more cases. I think we'd do this by adding a pureFunction element to the MatchableNode annotation.

Taking this as an example:

>             1: (0) 81|Membar { barriers=0, stamp=void, location=ANY_LOCATION,  } pred={17|LoopBegin}
>             2: (1) 120|Read { barrierType=NONE, stamp=i8 [0 - 1] ⇈0000000000000001, location=InfraControlL2.isDone, forceFixed=false, nullCheck=false,  } pred={81|Membar} address={122|AMD64Address}
>             3: (0) 82|Membar { barriers=3, stamp=void, location=ANY_LOCATION,  } pred={120|Read}
>             4: (1) 80|ZeroExtend { resultBits=32, inputBits=8, stamp=i32 [0 - 1] ⇈0000000000000001, getOp=org.graalvm.compiler.nodes.calc.ZeroExtendNode$$Lambda$15.897016085@7f61398cd358, inputAlwaysPositive=false, getReverseOp=org.graalvm.compiler.nodes.calc.ZeroExtendNode$$Lambda$16.1315998403@7f61398cd368,  } value={120|Read}

We would like to recognize that the ZeroExtend is a pure function so we don't have to check what nodes are between it and the Read, which would mean this is safe to match. The problem is that currently a match is emitted at the position of the last node matched because there might be other inputs that are only available at that point. So the other piece is figuring out the earliest point where all required inputs are available. So basically you walk back from endIndex examining each match node; if it's a pure function and it has no inputs outside of the match then you don't have to check it. Whatever that final index is becomes your emission point and the end of the region to be scanned for safety.

Then we need to fix MatchContext.setResult to put the emission on the right node for the emission point instead of on the root of the match.

tkrodriguez:
[...]

PR #984, opened 2019-02-15
Match context 942

rwestrel:
patch for https://github.com/oracle/graal/issues/942


rwestrel:
The patch includes a copy of the old validation logic to verify it doesn't cause a regression. That copy needs to be removed before the change is integrated.

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/match/MatchContext.java
```diff
@@ -125,6 +225,173 @@ public Result validate() {
...
+            }
+        }
+        assert verifyInputsDifferentBlock(root) == Result.OK;
+        return Result.OK;
+    }
+
+    private Result verifyInputsDifferentBlock(Node node) {
+        // Is there an input that's not part of the match that's after the emit position?
+        for (Node in : node.inputs()) {
+            if (in instanceof ValuePhiNode) {
```
tkrodriguez: This should be PhiNode since there are other kinds of PhiNode, in particular GuardPhiNode.

rwestrel: Ok. PR updated.

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/LIRGenerationPhase.java
```diff
@@ -84,6 +88,10 @@ private static void emitBlock(NodeLIRBuilderTool nodeLirGen, LIRGenerationResult
         instructionCounter.add(debug, lir.getLIRforBlock(b).size());
     }
 
+    private static void matchBlock(NodeLIRBuilderTool nodeLirGen, Block b, StructuredGraph graph, ScheduleResult schedule) {
```
tkrodriguez: Inline this into the caller.

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
```diff
@@ -419,11 +415,23 @@ public void doBlock(Block block, StructuredGraph graph, BlockMap<List<Node>> blo
         }
     }
 
+    @Override
     @SuppressWarnings("try")
-    protected void matchComplexExpressions(List<Node> nodes) {
+    public void matchBlock(Block block, StructuredGraph graph, StructuredGraph.ScheduleResult schedule) {
+        try (BlockScope blockScope = gen.getBlockScopeMatch(block)) {
```
tkrodriguez: I don't think using BlockScope for this adds anything.  Seems like BlockScope could be called MatchScope and directly subclass AutoCloseable.  getBlockScopeMatch becomes getMatchScope.  I don't think MatchScope even need to be a visible type since it can just return AutoCloseable.

tkrodriguez:
I think this basically looks ok.  Can you rebase onto latest master?  There's a new LLVM backend that subclasses NodeLIRBuilderTool so it needs a empty matchBlock implementation.  We're not sure the LLVM backend will stay around so that's why it's reusing LIR interfaces instead of being refactored to separate out the LIRness.  The testing in our internal gate looks good.

tkrodriguez:
Are there any statistics being captured that would demonstrate the improvement?  Like a counter for multi block matches and/or a counter where ignoring side effects allows a match?  I guess the last one is the difference between oldValidate and newValidate?

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
rwestrel: If MatchScope is not visible and matchBlock() uses AutoCloseable then it expects AutoCloseable.close() to throw Exception and something needs to be done with the exception eventhough it's never actually thrown. Exposing MatchBlock seems cleaner then.

rwestrel:
Thanks for the review

rwestrel:
> Are there any statistics being captured that would demonstrate the improvement? Like a counter for multi block matches and/or a counter where ignoring side effects allows a match? I guess the last one is the difference between oldValidate and newValidate?

There's no counter for multi block matches but adding one is straightforward. The simplest implementation of a counter for cases where ignoring a side effects allows a match I can think of would be to keep the old logic. I expected the old logic to be removed before this is integrated. Do you think I should refactor the code, keep the old logic but only call it when collecting counters?

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
tkrodriguez: Sorry I meant to suggest DebugCloseable which is our AutoCloseable that doesn't throw.

tkrodriguez:
I just wanted to look at the counters for some benchmarks to see what the results look like.  Could you add it so I can play with it and then I can strip out oldValidate when I push it?

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
rwestrel: Ok. I'm making that change.

rwestrel:
> I just wanted to look at the counters for some benchmarks to see what the results look like. Could you add it so I can play with it and then I can strip out oldValidate when I push it?

Sure.

rwestrel:
I pushed a rebased change.

adinn:
Hurrah!

tkrodriguez:
Your changes look good.  I'd wanted to look more deeply into the matcher effectiveness but I just don't have the time now and it's not something you'd need to fix in the PR anyway, so I'm going to go ahead and push this.  I'm adding a changeset to remove the oldValidate logic before we push as we discussed previously.  Thanks for the fix.  I'll create an issue with some of the general matcher issues I've noticed and maybe if you're interested you could pick that up later.  Or we could step forward into the HIR level matching that had been discussed in the past.

adinn:
Hurrah for real this time! :-)

PR #1006, opened 2019-02-25
Fix #1003: Check type of artifact is 'jar' before adding it to classpath

Jurrie:

PR #1007, opened 2019-02-25
Remove properties that are not used

Jurrie:
The org.apache.maven.plugins:maven-compiler-plugin has its own configuration block. There, source and target are set to '8'. So the properties maven.compiler.source and maven.compiler.target are not used.


Jurrie:
@graalvmbot  https://github.com/oracle/graal/pull/1006#issuecomment-466895840

olpaw:
Pending until https://github.com/oracle/graal/pull/1006#issuecomment-467354994 is resolved.

PR #1010, opened 2019-02-25
First GC interface reshufflings to isolate G1 and CardTable barriers

rkennke:
I am working on improving the GC interface for the Graal compiler. The goal is to isolate GC specific implementations out into their own packages, and let the compiler only call GC agnostic hooks.
This first PR towards that goal breaks up WriteBarrierAdditionPhase into G1BarrierSet (used by G1) and CardTableBarrierSet (used by all other currently supported GCs). Both are implementations of BarrierSet which defines the basic interface between compiler and GC barrier implemenatations, plus a factory to select the correct implementation.
One remaining piece of GC knowledge is the 'precise' argument that's passed via the memory access nodes. This shall be resolved in a follow-up change.

This is a re-work of https://github.com/oracle/graal/pull/977, this time hopefully sane.


rkennke:
The CI failures look like infra problems, e.g. jdk8-jvmci-0.55 not available. I'll wait until it's fixed.

mur47x111:
Thanks @rkennke ! I will start integrating this.

rkennke:
@mur47x111 thanks!
The CI issue looks to be an infrastructure problem. Is there a way to fix this, and kick off another test run?

mur47x111:
@rkennke I triggered another test run and it seems passing now. Also I did notice that this PR originates #977  and @christianhaeubl has already approved it. I think he is taking of the integration.

rkennke:
@mur47x111 perfect! Thanks!

PR #1011, opened 2019-02-25
SubtrateVM/GC: Log byte values with proper unit

rkennke:
Log messages in SubstrateVM's GC are logging all memory sizes in bytes, which is hard to read. I propose to auto-convert it into sensible units. E.g. 123456 becomes 123K, 1234567890 becomes 1234M and so on. This makes it much easier for humans to make sense of the numbers (IMO).

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/CollectionPolicy.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/CollectionPolicy.java
@@ -247,13 +247,13 @@ public abstract class CollectionPolicy {
             final UnsignedWord expectedSize = youngSize.add(oldInUse).add(averagePromotion);
             final UnsignedWord maxHeapSize = HeapPolicy.getMaximumHeapSize();
             final boolean vote = maxHeapSize.belowThan(expectedSize);
-            trace.string("  youngSize: ").unsigned(youngSize)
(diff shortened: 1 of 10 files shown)


Peter-B-Kessler:
In my opinion, for most of these messages, I want to see the actual number of bytes, not some truncation to some larger unit.  Most of these messages are for GC developers, where the exact numbers are important; or for performance monitoring scripts, where exact numbers are not inconvenient, and the sizing suffixes would be inconvenient.

For the one message that I expect **users** to look at, in `printGCAfter`, I deliberately copied the HotSpot "12345K-1234K" style and scaling, even though it makes reading the logs via scripts more difficult.  (Because I already have those scripts. :-)

Maybe a more useful addition to the `Log` interface would be one that prints numbers with thousand separators.  So that 123456 becomes 123,456 and 1234567890 becomes 1,234,567,890.  (Extra credit for separators in digits to the right of the decimal point in `Log.rational`.)  Then we will have to fight off people who disagree about the separator character, but not the people who disagree about whether "K" means "2^10" or "10^3".  (And the people whose log-parsing scripts we break. :-)

rkennke:
Ok, fair enough. I am a GC dev, and I find it difficult to tell the order of magnitude of (e.g.)  12232223 vs 1232223 when scanning logs manually. But you're right, I could easily employ some post-processing to get the information I need, and that would be lossless. Thousand-separators would indeed be more useful. Not sure I'll get to do that now, I basically picked this simple task to get started with the GC, and substrate, graal and git in general ;-)

PR #1015, opened 2019-02-27
Generalize barrier info on nodes. Figure out GC specific properties i…

rkennke:
As discussed here:
 https://mail.openjdk.java.net/pipermail/graal-dev/2019-February/005634.html
I would like to generalize the GC barrier information that we carry around in nodes. Instead of the rather GC specific PRECISE vs IMPRECISE information, I'd prefer to record whether or not the access is to array element or a field, and whether or not the access is to primitive field or object field. We also need to know if it's the special access to a Reference.referent field.

I tried to determine the primitive vs. object field information in the backend from the JavaKind of the node (or in case of writes, its value), but it turns out this not always consistent with the information carried in BarrierType. I suspect that object writes can sometimes be generated that don't require a barrier?

Notice that the implementation covers the need for current barriers. This means it's equivalent to what was there before. It's not yet generating the correct information for loads, yet (except for the referent field access), because no GC currently needs it. I will get to this later, when adding support for Shenandoah and/or ZGC, which will require this sort of information. I'd rather not do this blindly. ;-)

### DIFF
--- a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/memory/HeapAccess.java
+++ b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/memory/HeapAccess.java
@@ -34,17 +34,25 @@ public interface HeapAccess {
      */
     enum BarrierType {
         /**
-         * Primitive stores which do not necessitate barriers.
+         * Primitive access which do not necessitate barriers.
          */
         NONE,
         /**
-         * Array object stores which necessitate precise barriers.
+         * Array object access.
          */
-        PRECISE,
+        ARRAY,
         /**
-         * Field object stores which necessitate imprecise barriers.
+         * Field object access.
          */
-        IMPRECISE
+        FIELD,
+        /**
+         * Unknown (aka field or array) object access.
+         */
+        UNKNOWN,
+        /**
+         * Weak field access (e.g. Hotspot's Reference.referent field).
+         */
+        WEAK_FIELD
     }
 
     /**
--- a/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/DefaultJavaLoweringProvider.java
+++ b/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/DefaultJavaLoweringProvider.java
@@ -822,10 +822,10 @@ public abstract class DefaultJavaLoweringProvider implements LoweringProvider {
                                 if (virtual instanceof VirtualInstanceNode) {
                                     VirtualInstanceNode virtualInstance = (VirtualInstanceNode) virtual;
                                     address = createFieldAddress(graph, newObject, virtualInstance.field(i));
-                                    barrierType = BarrierType.IMPRECISE;
+                                    barrierType = BarrierType.FIELD;
                                 } else {
                                     address = createArrayAddress(graph, newObject, virtual.entryKind(i), ConstantNode.forInt(i, graph));
-                                    barrierType = BarrierType.PRECISE;
+                                    barrierType = BarrierType.ARRAY;
                                 }
                                 if (address != null) {
                                     WriteNode write = new WriteNode(address, LocationIdentity.init(), implicitStoreConvert(graph, JavaKind.Object, allocValue), barrierType);
@@ -940,24 +940,24 @@ public abstract class DefaultJavaLoweringProvider implements LoweringProvider {
 
     protected BarrierType fieldStoreBarrierType(ResolvedJavaField field) {
         if (field.getJavaKind() == JavaKind.Object) {
-            return BarrierType.IMPRECISE;
+            return BarrierType.FIELD;
         }
         return BarrierType.NONE;
     }
 
     protected BarrierType arrayStoreBarrierType(JavaKind elementKind) {
         if (elementKind == JavaKind.Object) {
-            return BarrierType.PRECISE;
+            return BarrierType.ARRAY;
         }
         return BarrierType.NONE;
     }
 
     public BarrierType fieldInitializationBarrier(JavaKind entryKind) {
-        return entryKind == JavaKind.Object ? BarrierType.IMPRECISE : BarrierType.NONE;
+        return entryKind == JavaKind.Object ? BarrierType.FIELD : BarrierType.NONE;
     }
 
     public BarrierType arrayInitializationBarrier(JavaKind entryKind) {
-        return entryKind == JavaKind.Object ? BarrierType.PRECISE : BarrierType.NONE;
+        return entryKind == JavaKind.Object ? BarrierType.ARRAY : BarrierType.NONE;
     }
 
     private BarrierType unsafeStoreBarrierType(RawStoreNode store) {
@@ -972,10 +972,12 @@ public abstract class DefaultJavaLoweringProvider implements LoweringProvider {
             ResolvedJavaType type = StampTool.typeOrNull(object);
             // Array types must use a precise barrier, so if the type is unknown or is a supertype
             // of Object[] then treat it as an array.
-            if (type == null || type.isArray() || type.isAssignableFrom(objectArrayType)) {
-                return BarrierType.PRECISE;
+            if (type != null && type.isArray()) {
+                return BarrierType.ARRAY;
+            } else if (type == null || type.isAssignableFrom(objectArrayType)) {
+                return BarrierType.UNKNOWN;
             } else {
-                return BarrierType.IMPRECISE;
+                return BarrierType.FIELD;
             }
         }
         return BarrierType.NONE;
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/thread/VMThreadSTFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/thread/VMThreadSTFeature.java
@@ -174,14 +174,14 @@ public class VMThreadSTFeature implements GraalFeature {
     private boolean handleGet(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver) {
         VMThreadLocalInfo info = threadLocalCollector.findInfo(b, receiver.get());
         VMThreadLocalSTHolderNode holder = b.add(new VMThreadLocalSTHolderNode(info));
-        b.addPush(targetMethod.getSignature().getReturnKind(), new LoadVMThreadLocalNode(b.getMetaAccess(), info, holder, BarrierType.PRECISE));
+        b.addPush(targetMethod.getSignature().getReturnKind(), new LoadVMThreadLocalNode(b.getMetaAccess(), info, holder, BarrierType.ARRAY));
         return true;
     }
 
     private boolean handleSet(GraphBuilderContext b, Receiver receiver, ValueNode valueNode) {
         VMThreadLocalInfo info = threadLocalCollector.findInfo(b, receiver.get());
         VMThreadLocalSTHolderNode holder = b.add(new VMThreadLocalSTHolderNode(info));
-        b.add(new StoreVMThreadLocalNode(info, holder, valueNode, BarrierType.PRECISE));
+        b.add(new StoreVMThreadLocalNode(info, holder, valueNode, BarrierType.ARRAY));
         return true;
     }
 
--- a/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/WriteBarrierAdditionTest.java
+++ b/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/WriteBarrierAdditionTest.java
@@ -315,9 +315,10 @@ public class WriteBarrierAdditionTest extends HotSpotGraalCompilerTest {
                     JavaConstant constDisp = ((OffsetAddressNode) read.getAddress()).getOffset().asJavaConstant();
                     Assert.assertNotNull(constDisp);
                     Assert.assertEquals(referentOffset(getMetaAccess()), constDisp.asLong());
-                    Assert.assertTrue(config.useG1GC);
-                    Assert.assertEquals(BarrierType.PRECISE, read.getBarrierType());
-                    Assert.assertTrue(read.next() instanceof G1ReferentFieldReadBarrier);
+                    Assert.assertEquals(BarrierType.WEAK_FIELD, read.getBarrierType());
+                    if (config.useG1GC) {
+                        Assert.assertTrue(read.next() instanceof G1ReferentFieldReadBarrier);
+                    }
                 }
             }
         } catch (Throwable e) {
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/gc/g1/G1BarrierSet.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/gc/g1/G1BarrierSet.java
@@ -26,6 +26,7 @@
 package org.graalvm.compiler.hotspot.gc.g1;
 
 import org.graalvm.compiler.debug.GraalError;
+import org.graalvm.compiler.hotspot.GraalHotSpotVMConfig;
 import org.graalvm.compiler.hotspot.gc.shared.BarrierSet;
 import org.graalvm.compiler.nodes.StructuredGraph;
 import org.graalvm.compiler.nodes.ValueNode;
@@ -41,10 +42,13 @@ import org.graalvm.compiler.nodes.type.StampTool;
 
 public class G1BarrierSet extends BarrierSet {
 
+    public G1BarrierSet(GraalHotSpotVMConfig vmConfig) {
+        super(vmConfig);
+    }
+
     @Override
     public void addReadNodeBarriers(ReadNode node, StructuredGraph graph) {
-        if (node.getBarrierType() != HeapAccess.BarrierType.NONE) {
-            assert (node.getBarrierType() == HeapAccess.BarrierType.PRECISE);
+        if (node.getBarrierType() == HeapAccess.BarrierType.WEAK_FIELD) {
             G1ReferentFieldReadBarrier barrier = graph.add(new G1ReferentFieldReadBarrier(node.getAddress(), node, false));
             graph.addAfterFixed(node, barrier);
         }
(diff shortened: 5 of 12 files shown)


compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/gc/shared/GCUtils.java
```diff
@@ -0,0 +1,69 @@
...
+ * questions.
+ */
+package org.graalvm.compiler.hotspot.gc.shared;
+
+import jdk.vm.ci.meta.JavaKind;
+import org.graalvm.compiler.nodes.memory.AbstractWriteNode;
+import org.graalvm.compiler.nodes.memory.FixedAccessNode;
+import org.graalvm.compiler.nodes.memory.HeapAccess;
+
+/**
```
Peter-B-Kessler: Should this be named `BarrierUtils`?

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/gc/shared/GCUtils.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Given a node and its BarrierType, determine the required precision of barriers.
+     *
+     * @param node the node for which to determine barrier precision
+     * @param useDeferredInitBarriers whether or not to use deferred initialization barriers
+     * @return the required barrier precision
+     */
+    public static BarrierPrecision getBarrierPrecision(FixedAccessNode node, boolean useDeferredInitBarriers) {
```
Peter-B-Kessler: This could be a static method on enum `BarrierPrecision`, but it could just as easily be with the `BarrierType` code.

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/gc/shared/GCUtils.java
rkennke: Yes, that is a good point! I will change it.

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/gc/shared/GCUtils.java
rkennke: Putting it in the enum sounds good. I'd rather not put it in the BarrierType class, because this is exposed everywhere.
I also have an idea how to reliably determine object vs. primitive. Stay tuned.

rkennke:
I just pushed a change that is closer to the original design, but renames and adds BarrierType values: IMPRECISE becomes FIELD, PRECISE becomes ARRAY, UNKNOWN means we don't know (and GCs treat it as array/precise) and REFERENT denotes a Reference.referent access. Initializing stores and useDeferredInitBarriers are now handled in the GC backends, rather than in the compiler. This still doesn't provide the implementation for load barriers, but it can easily be used for that. I'll do that in a followup PR.

Peter-B-Kessler:
@rkennke Not to confuse your life, but SubstrateVM has an annotation `@Hybrid` that allocates an array in an instance that also has fields.  See https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/annotate/Hybrid.java.

The `@Hybrid` annotation is used on `DynamicHub` (the thing pointed to from every object header) for the vtable for the class.  (Also a bit set for type comparisons.)  `DynamicHub` instances are read-only, so the current collector does not need write-barriers to access them.  (And the current collector does not move them.)

I don't know if that changes the names you want to give to `FIELD` and `ARRAY`.  Hybrids are enough of a special case that you might want ignore them.  If they don't need barriers, then they should not affect the names given to barrier types.

rkennke:
@Peter-B-Kessler oook, good to know, thanks! It sounds like I can safely ignore it for now (and not sure if the previous names have been any more useful either) :-)

I'm about to do some SubstrateVM/GC work soon anyway (mostly unrelated to this here), and then I guess I will get a better understanding of the situation there.

Peter-B-Kessler:
The thing to remember about the current SubstrateVM garbage collector is the (lack of) motivation for it.  The idea of SubstrateVM was fast startup for small, short-running, services.  In the best of all possible worlds, the application runs to completion without needing any collection.  But we had to have some garbage collector, so a serial, stop-the-world, copying collector was "good enough".  For now.  If your application has gigabyte heaps, you don't need fast startup, and you are probably better off using HotSpot with your favorite (:-) collector.  

The other constraint was running in environments where large blocks of contiguous memory are not available, which is why the current SubstrateVM heap comes in "chunks".  That also means one does not have to specify a maximum heap size, and one can change the heap shape at runtime.  Also, pinning objects is ... easier.

If you have questions about the current SubstrateVM collector and heap: ask away.

rkennke:
My application doesn't have GBs of heaps. I'm looking at 100-200MB. I guess it's too big/too long-running to not fit the original assumption, but not big enough to warrant full hotspot+GCs. Let's take that discussion to the mailing lists though... ;-)

This PR looks good to integrate (at least, CI-wise). Do the latest changes resolve your concerns?

rkennke:
@Peter-B-Kessler BTW, if you expect that an application dies before even hitting a single GC, you might want an even simpler GC: a single-space, sliding GC that triggers on memory exhaustion. This way you'd make best use of available memory, and not pay the runtime overhead of barriers. ;-)

adinn:
I suspect Peter (or, more importantly, Graal users) probably wants an insurance policy as well as low GC overhead.

rkennke:
The adoption of Epsilon for those sorts of uses is an indication that this is a more feasible proposal than it sounds like. And indeed, under closed world assumption, when you know how much you're going to allocate, and lifetime is short and limited anyway, this might indeed be your best bet.

Peter-B-Kessler:
Fortunately/Unfortunately, the GraalVM runtime compiler itself is written in Java, and allocates a lot of memory, and leaves behind a lot of garbage. So we chose a two-generation scavenger as a compromise.  The epsilon garbage collector might be useful for when we can ahead-of-time compile the whole application.  Eliminating barriers would be one advantage.  If we could eliminate safepoint checks that would win, too.  Lots of opportunities to win the "most improved VM" award. :-)

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/thread/VMThreadSTFeature.java
```diff
@@ -173,14 +173,14 @@ public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Rec
     private boolean handleGet(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver) {
         VMThreadLocalInfo info = threadLocalCollector.findInfo(b, receiver.get());
         VMThreadLocalSTHolderNode holder = b.add(new VMThreadLocalSTHolderNode(info));
-        b.addPush(targetMethod.getSignature().getReturnKind(), new LoadVMThreadLocalNode(b.getMetaAccess(), info, holder, BarrierType.PRECISE));
+        b.addPush(targetMethod.getSignature().getReturnKind(), new LoadVMThreadLocalNode(b.getMetaAccess(), info, holder, BarrierType.UNKNOWN));
```
christianwimmer: The correct new barrier type to use in this class (both changes) is `ARRAY`: In Substrate VM's single threaded mode, the thread local variables are stored in a normal Java array that resides in the image heap.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/memory/HeapAccess.java
```diff
@@ -34,17 +34,25 @@
...
-        IMPRECISE
+        FIELD,
+        /**
+         * Unknown (aka field or array) object access.
+         */
+        UNKNOWN,
+        /**
+         * Referent field access.
+         */
+        REFERENT
```
christianwimmer: It's not ideal to have a barrier type that is only used by one specific VM, the HotSpot VM, in the VM-independent part of Graal. Maybe it is a bad idea in the first place to have barrier types as an `enum` here, and not just define an empty marker interface here and have the actual value definitions in VM-specific code.

tkrodriguez: I think a clear set in the core would better than multiple different definitions but I agree this naming is very HotSpot specific.  Maybe this would be better called WEAK_FIELD to generalize it?

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/DefaultJavaLoweringProvider.java
```diff
@@ -972,10 +972,12 @@ private BarrierType guessStoreBarrierType(ValueNode object, ValueNode value) {
             ResolvedJavaType type = StampTool.typeOrNull(object);
             // Array types must use a precise barrier, so if the type is unknown or is a supertype
             // of Object[] then treat it as an array.
-            if (type == null || type.isArray() || type.isAssignableFrom(objectArrayType)) {
-                return BarrierType.PRECISE;
+            if (type == null || type.isAssignableFrom(objectArrayType)) {
```
tkrodriguez: I think you need to rearrange these tests.  type.isArray() must be tested before type.isAssignableFrom(objectArrayType) to find the ARRAY case.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/memory/HeapAccess.java
christianwimmer: We could do a split: some definitions (definitely `UNKNOWN`, but maybe also `FIELD` and `ARRAY`) in the platform independent part, and others like `REFERENT` in HotSpot specific code.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/thread/VMThreadSTFeature.java
tkrodriguez: This should probably be an implementation detail of LoadVMThreadLocalNode instead of being a constructor parameter.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/memory/HeapAccess.java
dougxc: Why is REFERENT HotSpot specific since it's a field in a (non-HotSpot) JDK class?

christianwimmer: I guess the field is defined in the `Reference` class because it simplified the HotSpot implementation (unfortunately the JDK and the VM are not as decoupled as they should be). But the field has the interesting comment `/* Treated specially by GC */` ;-)
So your argument is that the bad JDK/VM separation means that Graal also does not need to separate it out into VM independent code. Which is a valid argument.

tkrodriguez: If SVM adds a G1 style collector it will need the exact same special handling.  The read from referent needs special handling because it's reading from a weak field.  Even if the field were hidden in some way the actual read would still need this work.

rkennke: Currently, only Shenandoah and G1, probably ZGC (not sure) need to know about it, they are all Hotspot-only. Current SVM GCs don't currently need to know about it, but it surely belongs into the GC interface anyway (even if ignored by specific implementations). I'm fine with renaming it to WEAK_FIELD, this should be generic enough?

tkrodriguez: I think so.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/thread/VMThreadSTFeature.java
rkennke: I don't think so. VMThreadMTFeature.java passes NONE there. I'll change it to ARRAY for the VMThreadSTFeature.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/DefaultJavaLoweringProvider.java
rkennke: Latest push should fix it.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/memory/HeapAccess.java
rkennke: Latest push fixes this.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/thread/VMThreadSTFeature.java
rkennke: Changed to ARRAY in latest push.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/DefaultJavaLoweringProvider.java
```diff
@@ -972,10 +972,12 @@ private BarrierType guessStoreBarrierType(ValueNode object, ValueNode value) {
             ResolvedJavaType type = StampTool.typeOrNull(object);
             // Array types must use a precise barrier, so if the type is unknown or is a supertype
             // of Object[] then treat it as an array.
-            if (type == null || type.isArray() || type.isAssignableFrom(objectArrayType)) {
-                return BarrierType.PRECISE;
+            if (type.isArray()) {
```
tkrodriguez: This needs a type != null guard.

tkrodriguez:
The travis gate failed because of a minor indenting problem.  There's also a problem our internal gate found that didn't show up here.  There are a bunch of failures when running with -Dgraal.VerifyPhases=true that look like this:

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/DefaultJavaLoweringProvider.java
```diff
@@ -972,10 +972,12 @@ private BarrierType guessStoreBarrierType(ValueNode object, ValueNode value) {
...
+                return BarrierType.ARRAY;
+            } else if (type == null || type.isAssignableFrom(objectArrayType)) {
+                return BarrierType.UNKNOWN;
             } else {
-                return BarrierType.IMPRECISE;
+               return BarrierType.FIELD;
```
tkrodriguez: Wrong indent.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/thread/VMThreadSTFeature.java
christianwimmer: The difference between the `MT` and `ST` feature is correct: in multi-threaded (MT) mode, the thread local variables are in the thread descriptor is in C memory, i.e., no barriers needed. In single-threaded (ST) mode, the thread local variables are in the Java heap in a regular array, i.e., array write barriers are needed.

tkrodriguez: So the barrier kind is really a property of VMThreadLocalInfo?

christianwimmer: No, the barrier kind is a property of how the thread local storage is implemented. Which is all encapsulated in this class, so defining the barrier kind in this class is correct.

tkrodriguez: Sounds good.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/DefaultJavaLoweringProvider.java
rkennke: Should be fixed.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/DefaultJavaLoweringProvider.java
rkennke: Should be fixed.

dougxc:
Hi @rkennke can you please change the commits in this PR to use your RedHat email address since that it what is registered in the OCA database consulted by @graalvmbot .

rkennke:
> Hi @rkennke can you please change the commits in this PR to use your RedHat email address since that it what is registered in the OCA database consulted by @graalvmbot .

Sorry, I don't know how to do that. git rebase doesn't seem to work on merge commits, which is the one with wrong email.
I could fold all the commits in this PR into one changeset, and open a new PR for that. Might be a good idea anyway, this became a bit messy.
Sorry, those are my first steps with git and github.
What do you think?

dougxc:
I don't know how to fix the author in merge commits either so I think squashing all the commits in the PR is probably the right way to go. However, please don't open a new PR - just force push your squashed commit to this PR.

rkennke:
> I don't know how to fix the author in merge commits either so I think squashing all the commits in the PR is probably the right way to go. However, please don't open a new PR - just force push your squashed commit to this PR (i.e., `git pushf origin HEAD`).

Are you sure? force-pushing to a branch with PR is discouraged by github:
https://help.github.com/en/articles/about-pull-requests

Also, I am not sure how to do that either. Interactively 'rebase' the branch, thereby deleting all the changesets, and then commit the squashed changeset? Seems easier to create a new branch+PR, link back to this here, and start with clean slate?

dougxc:
I've force pushed often to GitHub PRs with no negative consequences. As far as I know, the best reason to avoid force pushing is when you are actively collaborating with others on a PR (as in they are also pushing commits to the PR - not just reviewing it).

thomaswue:
As long as you don't force push master, everything is fine ;-).

The following should work for squashing (make a local backup of the changesets in case something gets messed up):
`git rebase -i master`
https://github.com/wprig/wprig/wiki/How-to-squash-commits

rkennke:
Worked like a charm! Thanks!

rkennke:
Ok, I think I messed it up. Rebased and re-committed the whole thing as single changeset. I hope that works now.

rkennke:
Hmm, it looks like I killed the travis checker somehow. How should I proceed? Sorry for messing things up so badly...

thomaswue:
That is fine. For this type of change, we will anyway have to run much more tests than just the Travis checker.

rkennke:
Ok, cool. Travis seems kaputt anyway, my other PR (#1117) travis job seems stuck since a while already... Let me know if you find anything that requires fixing. Thanks!

tkrodriguez:
I grabbed your latest bits and they pass our internal gates so it all looks good.  We're waiting to push it until the GraalVM freeze is over at the end of the month.

rkennke:
Perfect, thanks!

rkennke:
It looks like this has already been integrated:

https://github.com/oracle/graal/commit/bbb272a2d0c53404d09ccc4c15745f70c7e5bb7b

Closing the PR.

tkrodriguez:
Sorry, it usually detects the merge of you commit and automatically closes it.  Not sure why it didn't detect it in this case.

PR #1018, opened 2019-03-01
Implement bitCount intrinsic using SIMD instructions for AArch64

armLeiJin:
1. Implement 3 new Neon instructions for AArch64, including cnt, addv, umov
2. Use fmov, cnt, addv, umov to implement bitcount intrinsic for Aarch64
3. Fix nodeinfo for bitcount node in Aarch64 since it is different from
other platforms
4. Add instructions encoding test for newly added instructions and code
gen test for bitcount node instrinsic modification

Change-Id:I8437d12e8c9dfdf354a8480d4f4a697f75eb9ab9


armLeiJin:
@sanzinger,  could you take a review on the patch? Thanks!

compiler/src/org.graalvm.compiler.replacements.aarch64/src/org/graalvm/compiler/replacements/aarch64/AArch64GraphBuilderPlugins.java
```diff
@@ -64,8 +64,8 @@ public static void register(Plugins plugins, BytecodeProvider bytecodeProvider,
         invocationPlugins.defer(new Runnable() {
             @Override
             public void run() {
-                registerIntegerLongPlugins(invocationPlugins, AArch64IntegerSubstitutions.class, JavaKind.Int, bytecodeProvider);
```
sanzinger: This would make the `AArch64IntegerSubstitutions` and `AArch64LongSubstitutions` obsolete right?

compiler/src/org.graalvm.compiler.replacements.aarch64/src/org/graalvm/compiler/replacements/aarch64/AArch64GraphBuilderPlugins.java
```diff
@@ -64,8 +64,8 @@ public static void register(Plugins plugins, BytecodeProvider bytecodeProvider,
...
-                registerIntegerLongPlugins(invocationPlugins, AArch64LongSubstitutions.class, JavaKind.Long, bytecodeProvider);
+                registerIntegerLongPlugins(invocationPlugins, JavaKind.Int, bytecodeProvider);
```
sanzinger: This would make the `AArch64IntegerSubstitutions` and `AArch64LongSubstitutions` obsolete right? So we could delete both classes?

sanzinger:
FYI, I have rebased the PR to latest `master` and removed `AArch64IntegerSubstitutions` and AArch64LongSubstitutions`.

compiler/src/org.graalvm.compiler.replacements.aarch64/src/org/graalvm/compiler/replacements/aarch64/AArch64GraphBuilderPlugins.java
armLeiJin: Yes, you are right.

armLeiJin:
> FYI, I have rebased the PR to latest `master` and removed `AArch64IntegerSubstitutions` and AArch64LongSubstitutions`. Please give me an OK if this correct.

OK. Thanks!

Issue #1020, opened 2019-03-01
[native-image] Class.getSimpleName not yet implemented


johanvos:
DynamicHub doesn't have a proper implementation for Class.getSimpleName in Java 11

johanvos:
See PR #1021 for an implementation

PR #1021, opened 2019-03-01
Implementation for Class.getSimpleName

johanvos:
This PR provides the required code to get Class.getSimpleName() working for Java 11 (see #1020)

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
@@ -260,6 +260,7 @@ public final class DynamicHub implements JavaKind.FormatWithToString, AnnotatedE
     private GenericInfo genericInfo;
     private AnnotatedSuperInfo annotatedSuperInfo;
 
+
     /**
      * Final fields in subsituted classes are treated as implicitly RecomputeFieldValue even when
      * not annotated with @RecomputeFieldValue. Their name must not match a field in the original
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
```diff
@@ -598,14 +602,19 @@ private ClassLoader getClassLoader0() {
...
+    private native String getSimpleNameJDK8OrEarlier();
 
-    @Substitute //
-    @TargetElement(onlyWith = JDK9OrLater.class)
-    private String getSimpleName0() {
-        throw VMError.unsupportedFeature("JDK9OrLater: DynamicHub.getSimpleName0()");
+    @Substitute
+    @TargetElement(name="getSimpleName", onlyWith = JDK9OrLater.class)
+    private String getSimpleNameJDK9OrLater() {
+        return simpleNameReference.get();
```
cstancu: Why do we need a `LazyFinalReference` for the simple name? The original implementation is not thread safe. We can use a simple `if (value == null) {compute and cache value}` pattern. Looking at the previous PRs we didn't need a `LazyFinalReference` for the `packageName` and `allPermDomain` fields either -  I will update those in a future PR. `LazyFinalReference` was introduced for the `singleModule` field because the returned object must always be the same, and reference comparisons might be used (at least JDK code does reference comparisons on `Module` objects).

 Moreover, do we really need to cache the value? I see that before Java 9 it was not cached. The computation is fairly efficient and it will probably not be called in performance critical code. Caching it potentially just pollutes the heap with garbage.

vjovanov:
Minor: this change is simple enough to squash all the commits together.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
johanvos: I copied the pattern from allPermDomain. I can remove the LazyFinalReference though.
As for caching: we might call this method >1000 times/second (simplename is linked to css selectors), and in that case I think caching should be beneficial?

cstancu: Yes, please remove the `LazyFinalReference` and let's use simple caching of the value to be equivalent with the JDK9 implementation.

johanvos:
I couldn't squash, but I created a new PR (#1042) that has those changes.
Note that we don't cache the call to getSimpleName() now. Adding a simple `String simpleName` to DynamicHub and use that to store the value didn't work.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
vjovanov: We can not use a simple field here as the DynamicHub is immutable. It would be nice if we could document or test this so we don't fail with a completely undebuggable error. FYI @christianwimmer 

For caching, we would have to use the reference then. I have no strong opinion about whether it is better to cache or not.

Issue #1035, opened 2019-03-04
[native-image] createTempFile().toPath() fails


johanvos:
The following application will cause an Error (at least on Linux):
```
import java.io.File;
import java.nio.file.Files;

public class HelloWorld {

    public static void main(String[] args) throws Exception  {
        File f = Files.createTempFile("foo", ".bar").toFile();
    }
}

```
```Exception in thread "main" java.lang.UnsupportedOperationException: Path not associated with default file system.```

For Path instances created using `Paths.get()`, the `Path.getFileSystem()` always returns the same value as `FileSystems.getDefault()`. However, Paths created via `Files.createTempFile()` return a different value for `getFileSystem()`.

cstancu:
Fixed in https://github.com/oracle/graal/commit/f4452fe3a67871989d309e39f8a24dabffc613ac.

PR #1037, opened 2019-03-04
Fix for tempfile dir #1035

johanvos:
Initialize tmpdir at runtime, making sure the FileSystem on it is computed at runtime.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaNIOSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaNIOSubstitutions.java
@@ -24,6 +24,7 @@
  */
 package com.oracle.svm.core.jdk;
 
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaNIOSubstitutions.java
```diff
@@ -46,6 +50,17 @@ static SecureRandom get() {
             return random;
         }
     }
+
+    static final class TempFileHelperDir {
+        private static Path dir;
+
+        static Path get() {
+            if (dir == null) {
+                dir = Path.of(System.getProperty("java.io.tmpdir"));
```
vjovanov: I have just replaced this with `Paths.get` as it works on JDK 8 as well.

vjovanov:
This has been merged, thank you!

PR #1042, opened 2019-03-06
Fix #1020: implement Class.getSimpleName() and required calls

### LINKED ISSUE #1020: [native-image] Class.getSimpleName not yet implemented

johanvos:


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
```diff
@@ -1105,6 +1118,10 @@ private String computePackageName() {
         return pn;
     }
 
+    private String computeSimpleName() {
```
vjovanov: Where is this used?

johanvos: Not. That was a left-over, I'll remove it.

PR #1058, opened 2019-03-12
Fix compiling issue after LLVM backend patch merged.

XiaohongGong:
The graal master compiles fails on AArch64 when running "mx build",
due to the omissive settings of LLVM_PLATFORM_SPECIFIC. This patch
will add optional attribute to it for other platforms except amd64.

Change-Id: I57bb3c1b075483ff475d66ea6be6bf3a02c85614


XiaohongGong:
This issue has been fixed in the latest master.

PR #1067, opened 2019-03-15
Use AArch64 instruction tbz #31 to do negative check in Graal

armLeiJin:
As  we have "if !(n < 0) <=> tbz n, sizeOfBits(n) - 1, label" in AArch64, this patch add a special match rule to find the case where IfNode has the condition that the greater value of  IntergerLessThanNode is zero and then simplify the generated assembly code from cmp, b.lt to tbz.

Change-Id: I7e72e12a0efea65b4fdc513200d105220590babe


armLeiJin:
@sanzinger  Could you help to take a look at the patch? Thanks.

PR #1071, opened 2019-03-16
fix NullPointerException when previous instruction in updateLastInstruction() is null

cosmicexplorer:
### Problem

There is a NullPointerException in `IntrinsicGraphBuilder.java`.

### Solution

Check for null in `.updateLastInstruction()`.

### Repro

Running the script at https://gist.github.com/cosmicexplorer/b8b4128315dfb4a30f462c5e73e5cd83 on `master` should produce:
```
# ... (lots of graal stacktraces)
        at org.graalvm.compiler.replacements.IntrinsicGraphBuilder.updateLastInstruction(IntrinsicGraphBuilder.java:130)
        at org.graalvm.compiler.replacements.IntrinsicGraphBuilder.append(IntrinsicGraphBuilder.java:148)
        at org.graalvm.compiler.replacements.StandardGraphBuilderPlugins$UnsafeAccessPlugin.createUnsafeAccess(StandardGraphBuilderPlugins.java:1058)
        at org.graalvm.compiler.replacements.StandardGraphBuilderPlugins$UnsafePutPlugin.apply(StandardGraphBuilderPlugins.java:1154)
        at org.graalvm.compiler.nodes.graphbuilderconf.InvocationPlugin.execute(InvocationPlugin.java:197)
        at org.graalvm.compiler.replacements.IntrinsicGraphBuilder.buildGraph(IntrinsicGraphBuilder.java:267)
        at com.oracle.graal.pointsto.flow.MethodTypeFlowBuilder.parse(MethodTypeFlowBuilder.java:191)
        at com.oracle.graal.pointsto.flow.MethodTypeFlowBuilder.apply(MethodTypeFlowBuilder.java:330)
        at com.oracle.graal.pointsto.flow.MethodTypeFlow.doParse(MethodTypeFlow.java:310)
        ... 12 more
Error: Image building with exit status 1
```

### Background

Executing the script at https://gist.github.com/cosmicexplorer/b8b4128315dfb4a30f462c5e73e5cd83 will build the wrapper for zinc (https://github.com/sbt/zinc) in the pants build tool (https://github.com/pantsbuild/pants) using `native-image`.

After starting off with https://github.com/graalvm/graalvm-demos/tree/master/scala-days-2018/scalac-native for scalac, which zinc wraps, I had spent some time figuring out which classes specifically were starting threads in static initializers (using `-H:NumberOfThreads=1` made the stacktraces usable here), and after using the new java agent in `substratevm/CONFIGURE.md` (and just removing entries that caused errors), I was able to get pretty close to making a native-image of this highly nontrivial and somewhat dynamic piece of Java/Scala. I experienced multiple errors while figuring out which entries to remove in the configuration generated by `native-image-configure`, which can be seen in the commit history. Only the one fix is now necessary after figuring this out -- it is not clear whether any of the other fixes are actual bugs, so I have removed them from the final diff.

This 100% works to compile code too, that was just a bit much for a minimal repro. I'm not sure if this change is correct -- I got to it just by looking at stack traces.

### DIFF
--- a/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/IntrinsicGraphBuilder.java
+++ b/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/IntrinsicGraphBuilder.java
@@ -127,7 +127,9 @@ public class IntrinsicGraphBuilder implements GraphBuilderContext, Receiver {
     private <T extends ValueNode> void updateLastInstruction(T v) {
         if (v instanceof FixedNode) {
             FixedNode fixedNode = (FixedNode) v;
-            lastInstr.setNext(fixedNode);
+            if (lastInstr != null) {
+                lastInstr.setNext(fixedNode);
+            }
             if (fixedNode instanceof FixedWithNextNode) {
                 FixedWithNextNode fixedWithNextNode = (FixedWithNextNode) fixedNode;
                 assert fixedWithNextNode.next() == null : "cannot append instruction to instruction which isn't end";


compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/IntrinsicGraphBuilder.java
```diff
@@ -127,7 +127,9 @@ protected IntrinsicGraphBuilder(OptionValues options, DebugContext debug, CorePr
     private <T extends ValueNode> void updateLastInstruction(T v) {
         if (v instanceof FixedNode) {
             FixedNode fixedNode = (FixedNode) v;
-            lastInstr.setNext(fixedNode);
+            if (lastInstr != null) {
+              lastInstr.setNext(fixedNode);
+            }
```
cstancu: @dougxc I wonder how this NPE was never a problem in Graal in until now since `lastInstr` can sometimes be `null`.

cstancu:
@cosmicexplorer thank you for the PR . Regarding the `substratevm/CONFIGURE.md`: it's still in an experimental state, we are working on improving its usability. Yes, it can sometimes report reflectively accessed elements that are not actually on the classpath. We are also working on a more automated solution for figuring out the started threads in static initializers.

cosmicexplorer:
The difficulty with locating started threads was also because of my almost complete unfamiliarity with the zinc/sbt codebase at first -- those errors went away pretty quickly once I dived into the source of that codebase a bit.

I would be super excited about any of the future developments you've mentioned -- I was having a lot of trouble understanding where the reflection in sbt/zinc/pants was coming from before I saw [this tweet](https://twitter.com/cstancu/status/1103413247305166849), and once I ran the agent and `native-image-configure` once I was able to get a working result within hours, which was awesome! I saw segfaults using the agent at first, but I haven't been able to repro it since then -- will make an issue if I do.

Having `native-image-configure` produce classes that aren't actually on the classpath was actually an extremely easy problem, because I was just able to remove the entries from the json file which matched the symbol names in the stack traces until it compiled!

cosmicexplorer:
I just fixed the formatting, but it seems like the other testing has all passed. The script I provided at https://gist.github.com/cosmicexplorer/b8b4128315dfb4a30f462c5e73e5cd83 has a very large not-really-minimal repro of the issue, but I'm not immediately familiar enough with the code in `IntrinsicGraphBuilder` to investigate how to minimize it to something more easily testable. Please let me know if there is any investigation I can do to minimize the repro case, or if there is any test file I should probably look into extending in order to test this, as necessary.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/IntrinsicGraphBuilder.java
dougxc: This bug in `IntrinsicGraphBuilder.updateLastInstruction` was exposed when [adding support for compressed pointers](https://github.com/oracle/graal/commit/f478d4a5d92489250bc67bf8e73e5340d3796aa7#diff-ff41b7830c4d716fddac0041450f3a0eR825). Normally `IntrinsicGraphBuilder` has only been used to manually weave a graph that has straight line control flow. Now that it is being used to weave conditional control flow, it is indeed possible for the `lastInstr` to be null. As such, I think the fix in this PR is correct.

cosmicexplorer:
Given that the fix seems correct (https://github.com/oracle/graal/pull/1071#discussion_r266390529), I have been told to mention that I work for Twitter and this is done as part of my work there, so I should be covered license-wise.

cstancu:
@cosmicexplorer I think the reason for the NPE is clear (especially after @dougxc's input). No further investigation is necessary. Can you please squash your changes to a single commit. Right now I see a number of unrelated, reverted in the meantime commits in the history. Thanks!

cosmicexplorer:
Yes, I will squash it now! Sorry about that.

cstancu:
Thank you for squashing! I will merge it as soon as I can confirm that we accept the GitHub automatically generated email address that you used (`1305167+cosmicexplorer@users.noreply.github.com`) or you need to use your `@twitter.com` address. Sorry for all this, once we get it sorted out your next PR will be painless!

cosmicexplorer:
I can amend the commits using my `@twitter.com` email and leave it in the repo's git config to avoid any issues later!

dougxc:
@cosmicexplorer that would make things much easier and clearer on our end - please go ahead.

cosmicexplorer:
Done! While I still prefer to block my email address, github allows temporarily toggling the "keep my email address private" setting, so this was easy.

PR #1072, opened 2019-03-16
Related publications to Graal and Truffle added

jjfumero:
Publications related to Truffle and Graal from the University of Manchester and Edinburgh added to the  general list of publications


jjfumero:
I had already signed the OCA

dougxc:
Hi @jjfumero , the OCA checking process uses the email address in each commit to check against an OCA signatory. I've updated the database used by the process with the emails you've used for this PR so it should soon be resolved.

Issue #1084, opened 2019-03-19
Allow LLVM 7.0.0 to be used


johanvos:
At this moment, there is a hard check on llvm 6.0.0 or 6.0.1 to be used. Version 7.0.0 works fine as well though.

vjovanov:
We will ship the supported LLVM as of 20.0.

PR #1085, opened 2019-03-19
Allow llvm 7 to be used

johanvos:
This addresses  #1084 and allows the use of llvm 7.0.0.


vjovanov:
We are still not sure about what to do here. The problem is that we need to fork the x86 LLVM backend in order to support reserved registers for GC and isolates.

If we allow multiple versions of LLVM now, then we would have to ship a version of our LLVM fork with GraalVM for every LLVM version. 

Suggestions appreciated here. I would really like to support LLVM versions and on only the one that we have forked. 

/cc @loicottet what is the exact size of the `graal-x86` target?

loicottet:
> @loicottet what is the exact size of the `graal-x86` target?
The macOS lib is 10.4MB, the Linux one is about 11MB.

vjovanov:
This fix only works until we have custom backends. @loicottet has a proposal on how to make this work properly for new LLVM versions.

Issue #1087, opened 2019-03-19
ClassCastException on java.net.Socket#setSoLinger


Sanne:
There's a simple ClassCastException triggered by attempting to explicitly control the SO_LINGER flag on a `java.net.Socket`.

```
java.lang.ClassCastException: java.lang.Boolean cannot be cast to 
java.lang.Integer
        at java.net.PlainSocketImpl.socketSetOption(PlainSocketImpl.java:5640)
        at 
java.net.AbstractPlainSocketImpl.setOption(AbstractPlainSocketImpl.java:275)
        at java.net.Socket.setSoLinger(Socket.java:1018)
````

I've already looked at investigating the reason; since it looks simple enough I'll try to fix it myself rather than boring you with further details :) I'll send a patch shortly.

Thanks

Peter-B-Kessler:
Are you saying that `Object value` can be either an `Integer` or a `Boolean` in the case of `SO_LINGER`?  I would be in favor of not trying to interpret `value` until I knew what I wanted out of it.  That's what the C code does, e.g., at the C line 945 and C line 952.

The C code at C lines 934-940 are only _setting up_ to access `Integer.value`, they are not accessing that field.  So, yes, the "easy" way of accessing that field is wrong (and subject to a `ClassCastException`).  But I would rather follow the C code and not access `value` until it is needed.  At which point it looks like it should be an `Integer`.

While we are turning over rocks in that code: What do the calls to `LibC.memcpy(optval_Pointer, optval_Pointer, WordFactory.unsigned(optlen))` think they are doing?

Sanne:
hi @Peter-B-Kessler thanks for the quick check!

to confirm, yes one could have a `Boolean` instance there instead of an `Integer`. I can easily get a `ClassCastException`, with a one liner reproducer:

     socket.setSoLinger(false, i); //(with any value of i)

Good idea to reduce the scopes. I'll change that.


Regarding `LibC.memcpy(optval_Pointer, optval_Pointer, WordFactory.unsigned(optlen))` .. interesting, I had not looked further.
May I leave that second issue to you? I have no idea what the intention was. N.B. that mistake seems repeated 3 times.

Sanne:
pushed an updated patch, accessing it only when needed like you suggested.

Thanks! it's indeed looking better as the Boolean seems to be ignored.

cstancu:
Fixed by https://github.com/oracle/graal/commit/8d2c98ca607b53b6ab1557ff207e408da81bae47.

PR #1088, opened 2019-03-19
Allow disabling SO_LINGER on java.net.Socket

Sanne:
Fixes ##1087

I'm a Red Hat employee so I believe I won't need to sign an individual contributor agreement.

It's my first contribution. Please let me know if I should do anything else; I'm wondering also if I should add a regression test? I could use a pointer to an example of how you want such tests structured.

Thanks!


Sanne:
@graalvmbot I'm a Red Hat employee so I believe I won't need to sign an individual contributor agreement.

cstancu:
@Sanne I update the OCA signatory list with your email.

Sanne:
thanks @cstancu !

Sanne:
pushed an improved commit, after feedback on #1087

PR #1093, opened 2019-03-20
Base64::encodeBlock stub

jphalimi-intel:
Implements Base64::encodeBlock stub enablement as an InvocationPlugin. Currently does not apply.

Base64$Encoder is a subclass of Base64.Encoder. My assumption is that the class name string in the Registration is wrong. I have tried $ and # signs but I was out of luck there too.

Another question I have is whether the BASE64_ENCODE_BLOCK instance should have its last constructor argument as a boolean.class or int.class, since I think we are dealing with bytecode at this point.

Let me know your thoughts.

Jp

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/meta/HotSpotGraphBuilderPlugins.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/meta/HotSpotGraphBuilderPlugins.java
@@ -25,6 +25,7 @@
 package org.graalvm.compiler.hotspot.meta;
 
 import static org.graalvm.compiler.core.common.GraalOptions.GeneratePIC;
+import static org.graalvm.compiler.hotspot.HotSpotBackend.BASE64_ENCODE_BLOCK;
 import static org.graalvm.compiler.hotspot.HotSpotBackend.GHASH_PROCESS_BLOCKS;
 import static org.graalvm.compiler.hotspot.meta.HotSpotAOTProfilingPlugin.Options.TieredAOT;
 import static org.graalvm.compiler.hotspot.replacements.HotSpotReplacementsUtil.JAVA_THREAD_THREAD_OBJECT_LOCATION;
@@ -100,6 +101,7 @@ import org.graalvm.compiler.replacements.ReplacementsImpl;
 import org.graalvm.compiler.replacements.StandardGraphBuilderPlugins;
 import org.graalvm.compiler.replacements.arraycopy.ArrayCopyNode;
 import org.graalvm.compiler.serviceprovider.GraalServices;
+import org.graalvm.compiler.word.Word;
 import org.graalvm.compiler.word.WordOperationPlugin;
 import org.graalvm.compiler.word.WordTypes;
 import org.graalvm.word.LocationIdentity;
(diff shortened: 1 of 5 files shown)


compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/meta/HotSpotGraphBuilderPlugins.java
```diff
@@ -563,6 +566,37 @@ private static void registerCounterModePlugins(InvocationPlugins plugins, GraalH
         }
     }
 
+    private static void registerBase64Plugins(InvocationPlugins plugins, GraalHotSpotVMConfig config, MetaAccessProvider metaAccess, ForeignCallsProvider foreignCalls) {
+        if (config.useBase64Intrinsics()) {
+            Registration r = new Registration(plugins, "java.util.Base64.Encoder");
```
mur47x111: `java.util.Base64$Encoder` should be fine. It is using `Class.forName` for resolving the class.

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/meta/HotSpotGraphBuilderPlugins.java
```diff
@@ -563,6 +566,37 @@ private static void registerCounterModePlugins(InvocationPlugins plugins, GraalH
...
+            r.register6("encodeBlock",
+                            byte[].class,
```
mur47x111:
```
            r.register7("encodeBlock",
                            Receiver.class,
```

mur47x111:
Second question, `boolean.class` should be fine. (correct me if wrong @dougxc )

jphalimi-intel:
Yes, UseBASE64Intrinsics is enabled, It is an AVX 512 stub and C2 applies it successfully. I applied the changes you mentioned to the updated commit, but I still see no luck. I tried to disable the inliner with Dgraal.Inline=false but it does not seem to apply either. I think the Registration pattern might be wrong. I have not found any other instance of a subclass pattern in the compiler codebase.

I tried to manually disable the flag check in the registration, but it did not seem to work either. I am a bit clueless here I admit :-(

mur47x111:
Could you please try with this patch?
```
diff --git a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/graphbuilderconf/InvocationPlugin.java b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/graphbuilderconf/InvocationPlugin.java
index 42a68f3a525..9b248f1d379 100644
--- a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/graphbuilderconf/InvocationPlugin.java
+++ b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/graphbuilderconf/InvocationPlugin.java
@@ -197,6 +197,11 @@ public interface InvocationPlugin extends GraphBuilderPlugin {
                 return apply(b, targetMethod, receiver, argsIncludingReceiver[1], argsIncludingReceiver[2], argsIncludingReceiver[3]);
             } else if (argsIncludingReceiver.length == 5) {
                 return apply(b, targetMethod, receiver, argsIncludingReceiver[1], argsIncludingReceiver[2], argsIncludingReceiver[3], argsIncludingReceiver[4]);
+            } else if (argsIncludingReceiver.length == 6) {
+                return apply(b, targetMethod, receiver, argsIncludingReceiver[1], argsIncludingReceiver[2], argsIncludingReceiver[3], argsIncludingReceiver[4], argsIncludingReceiver[5]);
+            } else if (argsIncludingReceiver.length == 7) {
+                return apply(b, targetMethod, receiver, argsIncludingReceiver[1], argsIncludingReceiver[2], argsIncludingReceiver[3], argsIncludingReceiver[4], argsIncludingReceiver[5],
+                                argsIncludingReceiver[6]);
             } else {
                 return defaultHandler(b, targetMethod, receiver, argsIncludingReceiver);
             }
```

mur47x111:
Great! I will start integrating this.

PR #1094, opened 2019-03-21
AArch64: Compressed pointer decoding optimization when base is zero.

XiaohongGong:
Currently decoding a compressed pointer with zero base will generate
the following code:
`add x0, xzr, x0, lsl #3`

This patch will optimize it to:
`lsl x0, x0, #3`

Here is a jmh benchmark to decode a compressed pointer:
```
public class UncompressPointer {
    private static final int field0 = 1000000;
    private int field1;

    static class A {
        B b;

        A(B b) {
            this.b = b;
        }
    }

    static class B {
        int field;

        B(int i) {
            this.field = i;
        }
    }

    static A[] arr;

    static {
        arr = new A[field0];
        for (int i = 0; i < field0; i++)
           arr[i] = new A(new B(i));
    }

    public static int func() {
        int result = 0;
        for (int i = 0; i < field0; i++) {
            result += arr[i].b.field;
        }
        return result;
    }

    @Benchmark
    public void jmhUncompressPointer() {
        field1 = func();
    }
}
```
And the performance results:
```
                         Score      Error    Units
Without this patch     21598.534 ? 2188.242  us/op
With this patch        19322.791 ? 1219.022  us/op
```

Change-Id: I5af6799eb82470d5598f5991e9210f99874f458a

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot.aarch64.test/src/org/graalvm/compiler/hotspot/aarch64/test/AArch64UncompressPointerTest.java
+++ b/compiler/src/org.graalvm.compiler.hotspot.aarch64.test/src/org/graalvm/compiler/hotspot/aarch64/test/AArch64UncompressPointerTest.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, Arm Limited and affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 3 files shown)


XiaohongGong:
Hi there,
Can anyone please take a look at this patch? Thank you!

sanzinger:
Thank you for your contribution. I'll take a look as soon as possible. (And the other AArch64 related patches too)

XiaohongGong:
> Thank you for your contribution. I'll take a look as soon as possible. (And the other AArch64 related patches too)

Thanks so much!

compiler/src/org.graalvm.compiler.hotspot.aarch64.test/src/org/graalvm/compiler/hotspot/aarch64/test/AArch64UncompressPointerTest.java
```diff
@@ -0,0 +1,125 @@
...
+        result = AArch64.r11;
+    }
+
+    private void emitUncompressPointer(Register base, int shift) {
+        AArch64HotSpotMove.UncompressPointer.emitUncompressCode(masm2, input, result, base, shift, true);
+    }
+
+    private void compareAssembly() {
+        byte[] expected = masm1.close(false);
+        byte[] actual = masm2.close(false);
```
sanzinger: I like this kind of test. Nice work.

XiaohongGong: Thank you!

PR #1117, opened 2019-03-27
Introduce -XX:MaxHeapFreeRatio option to avoid overloading of -Xms

rkennke:
Currently, -Xms is overloaded to also be the threshold for shrinking the heap. This seems counter-intuitive, esp. when coming from JDK. This change proposes to introduce a separate option, very similar to the same option in JDK. This should be less surprising.


rkennke:
Apparently this is violating NO_ALLOCATION restriction somehow. I don't fully understand through which call path that happens, and how to avoid it. Will dig a little deeper tomorrow. Any hints are appreciated.

rkennke:
Looks like a beginners mistake: I used + to concat strings. Should be fixed.

rkennke:
Oops, sorry, I accidentally pushed from a box that had git not properly set-up.

rkennke:
Travis suggests that I do this:

-    @Option(help = "The maximum percentage of heap free after GC to avoid shrinking")
-    public static final RuntimeOptionKey<Integer> MaxHeapFreeRatio = new RuntimeOptionKey<>(70);
+    @Option(help = "The maximum percentage of heap free after GC to avoid shrinking") public static final RuntimeOptionKey<Integer> MaxHeapFreeRatio = new RuntimeOptionKey<>(70);

Do I have to? It seems very wrong.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapPolicyOptions.java
```diff
@@ -53,6 +53,9 @@
     @Option(help = "How many bytes is enough to allocate an unaligned chunk for an array?  0 implies (AlignedHeapChunkSize / 8).") //
     public static final HostedOptionKey<Long> LargeArrayThreshold = new HostedOptionKey<>(HeapPolicy.LARGE_ARRAY_THRESHOLD_SENTINEL_VALUE);
 
+    @Option(help = "The maximum percentage of heap free after GC to avoid shrinking")
+    public static final RuntimeOptionKey<Integer> MaxHeapFreeRatio = new RuntimeOptionKey<>(70);
```
dmlloyd: You can avoid the format issue by appending ` //` to the end of the annotation line, like the others do.

dougxc: Or [surround](https://github.com/oracle/graal/blob/6db99e70eb5158923e5041586d876deef1ac2be9/compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/GraalCompilerOptions.java#L38) the options with `// @formatter:off` and `// @formatter:on`.

rkennke: Thanks! Very good to know! Should be good now.

rkennke:
The remaining CI failure doesn't look like related to my change.

dougxc:
This PR should not include the `substratevm/native-image` binary file.

rkennke:
Oops. That slipped in by accident. Should be removed now. Thanks for pointing out!

rkennke:
Accidentally closed the branch. Reopening...

rkennke:
I merged in latest changes from master, and all checks passed. From here, the PR is ready to go.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapPolicy.java
```diff
@@ -318,6 +318,11 @@ public static UnsignedWord getLargeArrayThreshold() {
         }
     }
 
+    @Fold
```
christianwimmer: This `@Fold` does not make much sense. It means that the option value of runtime option is only read and then constant folded during image generation.

You can either declare `MaxHeapFreeRatio` as a `HostedOptionKey` - then you can specify the option only during image generation and it is automatically constant folded even without the `@Fold` annotation.

Or you can remove the `@Fold` and get a "real" option that you can set at run time.

rkennke: Thanks, Christias! I removed the Fold. Remaining test failure doesn't look related.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapPolicyOptions.java
```diff
@@ -53,6 +53,9 @@
     @Option(help = "How many bytes is enough to allocate an unaligned chunk for an array?  0 implies (AlignedHeapChunkSize / 8).") //
     public static final HostedOptionKey<Long> LargeArrayThreshold = new HostedOptionKey<>(HeapPolicy.LARGE_ARRAY_THRESHOLD_SENTINEL_VALUE);
 
+    @Option(help = "The maximum percentage of heap free after GC to avoid shrinking") //
+    public static final RuntimeOptionKey<Integer> MaxHeapFreeRatio = new RuntimeOptionKey<>(70);
```
christianwimmer: `70%` sound like a very high value. Considering that the default maximum heap size is also a large fraction of the physical memory size, this means that we are basically never returning chunks back to the operating system.

What workloads did you use to evaluate the change? How does it affect performance vs. memory footprint?

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapChunkProvider.java
```diff
@@ -154,17 +154,18 @@ void consumeAlignedChunk(AlignedHeader chunk) {
...
         final Log trace = Log.noopLog().string("[HeapChunkProvider.keepAlignedChunk:");
-        final UnsignedWord minimumHeapSize = HeapPolicy.getMinimumHeapSize();
+        final UnsignedWord maximumHeapSize = HeapPolicy.getMaximumHeapSize();
         final UnsignedWord heapChunkBytes = HeapImpl.getHeapImpl().getUsedChunkBytes();
         final UnsignedWord unusedChunkBytes = bytesInUnusedAlignedChunks.get();
         final UnsignedWord bytesInUse = heapChunkBytes.add(unusedChunkBytes);
-        /* If I am under the minimum heap size, then I can keep this chunk. */
-        final boolean result = bytesInUse.belowThan(minimumHeapSize);
+        /* If I am under the max heap free ratio, then I can keep this chunk. */
+        final boolean result = bytesInUse.multiply(100).unsignedDivide(maximumHeapSize).aboveThan(100 - HeapPolicy.getMaxHeapFreeRatio());
```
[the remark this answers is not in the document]
rkennke: I changed the formula to be overflow-free (slightly differently to what you suggested to preserve the semantics), should be good now. Thanks for pointing out.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapPolicyOptions.java
rkennke: 70% comes from the default of the same value in JDK. It seems a bit confusing compared to the previous logic: before we kept chunks when used < minheap, now we keep chunks when *free* < MaxHeapFree. Which means that with a value of 70%, we keep chunks when used is > 30%. Does that still sound bad?

rkennke:
I'm closing this due to lack of interest on my side.

PR #1128, opened 2019-04-02
Support non-literal/non-inlinable doubles & floats for native-image

bobmcwhirter:
While HotSpot supported `fldr` to load them from a relatively close-by location,
in a native-image binary, the address space addressable by `fldr` is not sufficient
to locate a complex double placed in the data section.

Replace with an adrp/add + fldr against a scratch register to accomodate both.

### DIFF
--- a/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64Move.java
+++ b/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64Move.java
@@ -497,7 +497,12 @@ public class AArch64Move {
                         masm.fmov(32, dst, scratch);
                     }
                 } else {
-                    masm.fldr(32, dst, (AArch64Address) crb.asFloatConstRef(input));
+                    try (ScratchRegister scr = masm.getScratchRegister()) {
+                        Register scratch = scr.getRegister();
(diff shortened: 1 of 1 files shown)


bobmcwhirter:
@theRealAph does this look reasonable?

theRealAph:
> @theRealAph does this look reasonable?

Sure. I was a bit confused to begin with, but then I remembered the rather odd way that the Graal Aarch64 assembler works.

I'd move the crb.asFloatConstRef(...) line just before the masm.addressOf. I think that'd be less confusing, but it's about style rather than correctness.

bobmcwhirter:
Yah, it took me a while to realize htf `adrp` worked, considered the is only a destination reg, and no other argument. The assumption is that it is also patched/relocated, so any argument would be... useless.

And `addressOf` is short-hand for `adrp+add` with the abovementioned expectation.

I'd feel happier if it took some argument like we pass to `mov` often of `0xDEADDEADDEADDEADl` or whatnot...  And didn't call it `addressOf`.

Oh well.

theRealAph:
> I'd feel happier if it took some argument like we pass to `mov` often of `0xDEADDEADDEADDEADl` or whatnot... And didn't call it `addressOf`.

Yes, me too. This is a divergence from what we do in HotSpot. Never mind, it's good for now, thanks.

sanzinger:
Looks good, I'll merge it tomorrow.

PR #1129, opened 2019-04-03
Fix typo

AzeemJiva:
lfence was mistyped as lf**a**nce


AzeemJiva:
I fixed my local config, but seems too late for this.  My work email:  azeem.jiva@oracle.com

dougxc:
You can still change the author of the commit: https://makandracards.com/makandra/1717-git-change-author-of-a-commit

PR #1131, opened 2019-04-03
[WIP] cross-compile for ARM64

johanvos:


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/macho/MachOCpuType.java
```diff
@@ -29,14 +29,20 @@
  */
 public enum MachOCpuType {
     X86_64,
-    POWERPC64;
+    POWERPC64,
+    ARM,
+    ARM64;
```
bobmcwhirter: Is this to support Darwin on aarch64 or is this some LLVM stuff? (Just curious; I've only touched ELF).

Thus far, we've always called it `aarch64` and confusingly, I'd presume the 32-bit variant would be `armv7` to stick with what `uname` reports.

johanvos: Both. This is for generating aarch64 code using the LLVM backend.

johanvos:
superseded by https://github.com/oracle/graal/pull/1157

Issue #438, opened 2018-05-30
[native-image] Can't load Shared Libraries


Anuken:
I am trying to compile a java libGDX .jar using native-image. No errors are produced during compilation, but when I try to run the resulting executable, I get [this error](https://hastebin.com/utocobukej.sql), despite the fact that all the relevant `.so` files are in the jar file. 

[Here](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java#L288) is the specific method in the libGDX source which tries to load the libraries.

_Graal Version_: Latest commit, built from source.
_OS_: Linux Mint 18

_Full command_: `mx native-image -Djava.library.path=<path to directory with all .so files> -H:+JNI -jar <path to jar file>`

[Download link to jar file.](https://mega.nz/#!Ol41FQQI!NUwFxuAsXO7INXet9SBC7eRdrQMsa3deiSrJtt1c8JU)

Removing `-Djava.library.path` and/or adding a separate `-classpath` argument that points to the unzipped contents of the jar doesn't help. I've also tried using both absolute and relative paths for `-Djava.library.path`.

I've read the README on JNI, but it doesn't seem very clear what I should do when `System.load()` is used in the code-- I've tried creating a [jniconfig.json](https://hastebin.com/duguvexisi.json) file in the folder with the shared libraries, listing the libraries there and passing it to `JniConfigurationFiles`, but that gives me the error `Class <name>.so not found`.

Jotschi:
I had to place the lib files in a `clibraries/linux-amd64` folder. I think you can also use the `-H:CLibraryPath` setting to set the path.

Anuken:
@Jotschi Just tried putting all the shared libraries in the `clibraries/linux-amd64` folder, and then specifying it with `H:CLibraryPath` with both absolute and relative directories. Neither seemed to work-- I think the problem may be with how libGDX loads libraries. It fails when trying to call [getResourceAsStream](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/SharedLibraryLoader.java#L132), which seems to make sense since the normal jar file structure isn't present.

Were you using libGDX specifically? Just curious if I'm doing something wrong or if it is indeed a libGDX issue.

Jotschi:
@Anuken No - sorry. I was referring to my own usecase. I'm also a bit lost but I think you may need to specify the native methods of your class in the JNIConfigurationFile. (Wild guess)

[...]

PR #1136, opened 2019-04-05
Need to try all paths listed by 'java.library.path' before throwing UnsatisfiedLinkError

Sanne:
I believe this could be the fix for #438 

The code in :
https://github.com/oracle/graal/blob/cbadcf6f7a1890562984e1c581f7c3b31809b34e/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java#L91-L95

is able to deal with the requirement to list multiple paths for the _java.library.path_ system property, but then the code attempting to load the libraries is expecting to have "false" returned on failure from one path, so to try the next path.

However the failed attempt to load the library from the first path immediately causes an `UnsatisfiedLinkError` to be thrown, so the additional paths are never tested.

see also:
 - https://github.com/quarkusio/quarkus/pull/1674

I'd be happy to add an integration test, but I'm not familiar with how you might prefer something like this to be structured. If you could point me to something similar I'd be glad to learn and try putting one together.

Thanks

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java
@@ -150,6 +150,9 @@ public final class NativeLibrarySupport {
                 if (libraryInitializer != null) {
                     libraryInitializer.initialize(lib);
                 }
+            } catch (UnsatisfiedLinkError e) {
+                // Allow caller to try the next path:
+                return false;
             } finally {
                 NativeLibrary top = currentLoadContext.pop();
                 assert top == lib;


AzeemJiva:
Did you run manual tests? Various permutations (single path exception thrown or not, multiple paths, etc)?  Looks reasonable but question what happens if the last path in the list doesn't satisfy the linker?

Sanne:
hi @AzeemJiva , yes I tested this with local builds and it fixes all our issues with loading libraries.

If no single paths satisfies the loading, an `UnsatisfiedLinkError` is thrown: same semantics as previously.

Sanne:
@AzeemJiva @cstancu could you please have a look?

This is a blocker to use SSL / https on Linux, we'd really hope to have it in a release soon.

AzeemJiva:
LGTM, but I don't have merge permissions.

peter-hofer:
Thank you for your submission.

Sanne:
thanks all!

PR #1171, opened 2019-04-16
Float, Double Scalar FMA intrinsics

jphalimi-intel:
Initial PR for FMA intrinsics support. It implements the design suggested on the graal mailinglist, namely:

1. Add FMA instructions in VexRVMOp in AMD64Assembler
     a. It requires to add the VexOpAssertion.FMA and use CPUFeature.FMA flags
2. Add UseFMA flag from HotSpot flags in GraalHotSpotVMConfig.java
3. Add a registerFMA method in AMD64GraphBuilderPlugins::registerMathPlugins
     a. This requires to add a specific FusedMultiplyNodeNode, which will emit the corresponding FMA instructions.
           --> Need for TernaryNode support and AMD64Ternary support

I have a few questions regarding this PR considering this is the first time I touch a couple of things. I will post comments to make it easier.

### DIFF
--- a/compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
+++ b/compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
@@ -906,7 +906,8 @@ public class AMD64Assembler extends AMD64BaseAssembler {
         CPU_XMM(CPUFeature.AVX, null, CPU, null, XMM, null),
         AVX1_2_CPU_XMM(CPUFeature.AVX, CPUFeature.AVX2, CPU, null, XMM, null),
         BMI1(CPUFeature.BMI1, null, CPU, CPU, CPU, null),
-        BMI2(CPUFeature.BMI2, null, CPU, CPU, CPU, null);
+        BMI2(CPUFeature.BMI2, null, CPU, CPU, CPU, null),
+        FMA(CPUFeature.FMA, null, XMM, XMM, XMM, null);
 
         private final CPUFeature l128feature;
         private final CPUFeature l256feature;
@@ -1308,6 +1309,8 @@ public class AMD64Assembler extends AMD64BaseAssembler {
         public static final VexRVMOp VPCMPGTW  = new VexRVMOp("VPCMPGTW",  P_66, M_0F,   WIG, 0x65, VEXOpAssertion.AVX1_2);
         public static final VexRVMOp VPCMPGTD  = new VexRVMOp("VPCMPGTD",  P_66, M_0F,   WIG, 0x66, VEXOpAssertion.AVX1_2);
         public static final VexRVMOp VPCMPGTQ  = new VexRVMOp("VPCMPGTQ",  P_66, M_0F38, WIG, 0x37, VEXOpAssertion.AVX1_2);
+        public static final VexRVMOp VFMADD231SS = new VexRVMOp("VFMADD231SS", P_66, M_0F38, W0, 0xB9, VEXOpAssertion.FMA);
+        public static final VexRVMOp VFMADD231SD = new VexRVMOp("VFMADD231SD", P_66, M_0F38, W1, 0xB9, VEXOpAssertion.FMA);
         // @formatter:on
 
         private VexRVMOp(String opcode, int pp, int mmmmm, int w, int op) {
--- a/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
+++ b/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
@@ -57,6 +57,8 @@ import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VADDSD;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VADDSS;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VDIVSD;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VDIVSS;
+import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VFMADD231SD;
+import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VFMADD231SS;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VMULSD;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VMULSS;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VORPD;
@@ -120,6 +122,7 @@ import org.graalvm.compiler.lir.amd64.AMD64Move;
 import org.graalvm.compiler.lir.amd64.AMD64MulDivOp;
 import org.graalvm.compiler.lir.amd64.AMD64ShiftOp;
 import org.graalvm.compiler.lir.amd64.AMD64SignExtendOp;
+import org.graalvm.compiler.lir.amd64.AMD64Ternary;
 import org.graalvm.compiler.lir.amd64.AMD64Unary;
 import org.graalvm.compiler.lir.amd64.vector.AMD64VectorBinary;
 import org.graalvm.compiler.lir.amd64.vector.AMD64VectorBinary.AVXBinaryOp;
@@ -957,6 +960,21 @@ public class AMD64ArithmeticLIRGenerator extends ArithmeticLIRGenerator implemen
             getLIRGen().append(new AMD64Unary.RMOp(BSR, QWORD, result, asAllocatable(value)));
         } else {
             getLIRGen().append(new AMD64Unary.RMOp(BSR, DWORD, result, asAllocatable(value)));
+        }
+        return result;
+    }
+
+    @Override
+    public Variable emitFusedMultiplyAdd(Value a, Value b, Value c) {
+        Variable result = getLIRGen().newVariable(LIRKind.combine(a, b, c));
+        assert a.getPlatformKind().equals(b.getPlatformKind());
+        assert b.getPlatformKind().equals(c.getPlatformKind());
+
+        if (a.getPlatformKind() == AMD64Kind.DOUBLE) {
+            getLIRGen().append(new AMD64Ternary.ThreeOp(VFMADD231SD, AVXSize.XMM, result, asAllocatable(a), asAllocatable(b), asAllocatable(c)));
+        } else {
+            assert a.getPlatformKind() == AMD64Kind.SINGLE;
+            getLIRGen().append(new AMD64Ternary.ThreeOp(VFMADD231SS, AVXSize.XMM, result, asAllocatable(a), asAllocatable(b), asAllocatable(c)));
         }
         return result;
     }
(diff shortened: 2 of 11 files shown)


compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64Ternary.java
```diff
@@ -0,0 +1,103 @@
...
+        private final AVXSize size;
+
+        @Def({REG, HINT}) protected AllocatableValue result;
+        @Use({REG}) protected AllocatableValue x;
+        /**
+         * This argument must be Alive to ensure that result and y are not assigned to the same
+         * register, which would break the code generation by destroying y too early.
+         */
+        @Alive({REG, STACK}) protected AllocatableValue y;
+        @Alive({REG, STACK}) protected AllocatableValue z;
```
jphalimi-intel: Not sure the annotation is correct here. Comment needs to be updated accordingly as well.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64Ternary.java
```diff
@@ -0,0 +1,103 @@
...
+            this.z = z;
+        }
+
+        @Override
+        public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
+            AMD64Move.move(crb, masm, result, x);
+            if (isRegister(z)) {
+                opcode.emit(masm, size, asRegister(result), asRegister(y), asRegister(z));
+            } else {
+                assert isStackSlot(z);
```
jphalimi-intel: I am pretty sure there are things missing here since we have 3 inputs.

compiler/src/org.graalvm.compiler.replacements.amd64/src/org/graalvm/compiler/replacements/amd64/AMD64GraphBuilderPlugins.java
```diff
@@ -171,6 +179,30 @@ private static void registerMathPlugins(InvocationPlugins plugins, AMD64 arch, B
...
             registerRound(r, "floor", RoundingMode.DOWN);
         }
+
+        if (arch.getFeatures().contains(CPUFeature.FMA)) {
+            registerFMA(r);
+        }
+    }
+
+    private static void registerFMA(Registration r) {
+        // TODO: have a way to get config.useFMAIntrinsics flag check here.
```
jphalimi-intel: I need to have access to config.useFMAIntrinsics here. Is there a shortcut that I can use?

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/nodes/FusedMultiplyAddNode.java
```diff
@@ -0,0 +1,92 @@
...
+    public FusedMultiplyAddNode(ValueNode a, ValueNode b, ValueNode c) {
+        super(TYPE, computeStamp(a.stamp(NodeView.DEFAULT), a), a, b, c);
+        assert a.getStackKind() == JavaKind.Float || a.getStackKind() == JavaKind.Double;
+        assert b.getStackKind() == JavaKind.Float || b.getStackKind() == JavaKind.Double;
+        assert c.getStackKind() == JavaKind.Float || c.getStackKind() == JavaKind.Double;
+    }
+
+    @Override
+    public Stamp foldStamp(Stamp stampX, Stamp stampY, Stamp stampZ) {
+        return getX().stamp(NodeView.DEFAULT);
```
jphalimi-intel: I think the stamp needs to be a combination of the three inputs, but I am not sure whether it is necessary. Is this code ok?

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64Ternary.java
mur47x111: `Alive` means we will not overwrite the original value. If the allocatable is a register, the value remains the same after this LIR instruction.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64Ternary.java
mur47x111: y might not be in a register. I think the easiest way would be to mark y `@Alive({REG})`. /cc @zapster

compiler/src/org.graalvm.compiler.replacements.amd64/src/org/graalvm/compiler/replacements/amd64/AMD64GraphBuilderPlugins.java
jphalimi-intel: I am confused here. I tried to import org.graalvm.compiler.hotspot.GraalHotSpotVMConfig to be able to use it in AMD64GraphBuilderPlugins, but it does not seem to be possible, probably because of the package this file is in? I am not sure whether using another class is good from a design perspective. FMA intrinsics are technically a Math method substitution. Why is there a need for two different classes? Isn't there a way to access the config from this class?

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64Ternary.java
zapster: Correct. :+1:

compiler/src/org.graalvm.compiler.replacements.amd64/src/org/graalvm/compiler/replacements/amd64/AMD64GraphBuilderPlugins.java
dougxc: You can pass a `boolean useFMAIntrinsics` to the `AMD64GraphBuilderPlugins` constructor. It is called from `org.graalvm.compiler.hotspot.amd64.AMD64HotSpotBackendFactory.createGraphBuilderPlugins` which is HotSpot specific code and thus has access to `GraalHotSpotVMConfig`.

jphalimi-intel: I can do this. However, is it something we want to do? I am concerned that it is not a scalable solution, because we need to pass flags independently. If moving forward, math or other method substitutions require Hotspot flags access, we will need to pass all the flags. What do you think? Would it be, for example, possible to make the Hotspot config accessible directly from any plugin class?

dougxc: No, we cannot make HotSpotVMConfig accessible to non-hotspot classes otherwise SVM cannot use these classes. If we start to accumulate a bunch such flags, we can create a `AMD64GraphBuilderPlugins.Params` helper classes to pass them.

compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
```diff
@@ -961,6 +964,18 @@ public Variable emitBitScanReverse(Value value) {
         return result;
     }
 
+    @Override
+    public Variable emitFusedMultiplyAdd(Value a, Value b, Value c) {
+        Variable result = getLIRGen().newVariable(LIRKind.combine(a, b, c).changeType(AMD64Kind.DWORD));
```
mur47x111: You don't need to change the result type to `DWORD`. 
```
Variable result = getLIRGen().newVariable(LIRKind.combine(a, b, c));
```
would be sufficient. If you want to be robust, try assert that a, b, c are of the same platform kind and are either `FLOAT` or `DOUBLE`.

compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
```diff
@@ -961,6 +964,18 @@ public Variable emitBitScanReverse(Value value) {
...
+        if (a.getPlatformKind() == AMD64Kind.QWORD) {
+            getLIRGen().append(new AMD64Ternary.ThreeOp(VFMADD231SD, AVXSize.QWORD, asAllocatable(a), asAllocatable(a), asAllocatable(b), asAllocatable(c)));
```
mur47x111: Since we are issuing an LIG instruction, we can replace `AVXSize.QWORD` with `AVXSize.XMM`. Same for line 974. Also please replace the first `asAllocatable(a)` with `result`.

compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -906,7 +906,8 @@ private AMD64Shift(String opcode, int code) {
         CPU_XMM(CPUFeature.AVX, null, CPU, null, XMM, null),
         AVX1_2_CPU_XMM(CPUFeature.AVX, CPUFeature.AVX2, CPU, null, XMM, null),
         BMI1(CPUFeature.BMI1, null, CPU, CPU, CPU, null),
-        BMI2(CPUFeature.BMI2, null, CPU, CPU, CPU, null);
+        BMI2(CPUFeature.BMI2, null, CPU, CPU, CPU, null),
+        FMA(CPUFeature.FMA, null, CPU, CPU, CPU, null);
```
mur47x111: Should be `FMA(CPUFeature.FMA, null, XMM, XMM, XMM, null);`

jphalimi-intel:
The patch is now working and should be ready, unless other code changes need to be made. Speedups are as follows on a JMH test that I wrote for FMA Float/Double:

FMA intri	C2		Graal    Graal + stub C2/Graal  Graal acceleration
Float	2,209.12	1,454.46	2,634.36	**19.25%	81.12%**
Double	4,727.66	1,896.46	4,717.77	**-0.21%	148.77%**

mur47x111:
Thanks. I will start integrating this.

compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
```diff
@@ -961,6 +964,21 @@ public Variable emitBitScanReverse(Value value) {
...
     }
 
+    @Override
+    public Variable emitFusedMultiplyAdd(Value a, Value b, Value c) {
+        Variable result = getLIRGen().newVariable(LIRKind.combine(a, b, c));
+        assert a.getPlatformKind().equals(b.getPlatformKind());
+        assert b.getPlatformKind().equals(c.getPlatformKind());
+
+        if (a.getPlatformKind() == AMD64Kind.DOUBLE) {
+            getLIRGen().append(new AMD64Ternary.ThreeOp(VFMADD231SD, AVXSize.XMM, result, asAllocatable(a), asAllocatable(b), asAllocatable(c)));
```
mur47x111: Since we are using 231 instructions, the order of the operands should be `c, a, b`.

mur47x111:
FYI, `Math.fma` is not supported on Java 8. Thus we have to provide a proxy in `GraalService`, see https://github.com/oracle/graal/commit/91d1f738aae472001574376e10363938b87756f3

jphalimi-intel:
Thanks for the info. I noticed that some automatic build did not pass because of fma, I assume it was the root cause. Thanks for updating the patch.

PR #1173, opened 2019-04-17
Defensiveness around the polyglot launcher class names.

bobmcwhirter:
In the even the property is unset or empty, avoid either
an NPE or a ClassNotFoundException regarding an empty string,
since `splitAsStream` under JDK11 will apparently split an
empty string into a stream of 1 empty string, instead of
an empty stream of no strings.

### DIFF
--- a/sdk/src/org.graalvm.launcher/src/org/graalvm/launcher/PolyglotLauncher.java
+++ b/sdk/src/org.graalvm.launcher/src/org/graalvm/launcher/PolyglotLauncher.java
@@ -222,8 +222,13 @@ public final class PolyglotLauncher extends Launcher {
 
     static {
         if (IS_AOT) {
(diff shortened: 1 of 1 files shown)


bobmcwhirter:
I'll repush with my @redhat address in a moment.

bobmcwhirter:
re-pushed, dunno why the OCA check hasn't run.

gilles-duboscq:
Hi Bob,
Thank you for the PR. I have some changes i'm working on where i completely removed this code. If that's OK with you i'd like to hang on this PR for a bit: next week i should either have merged my changes in which case I think we could close this PR or I won't have merged my changes in which case we can integrate your changes.

bobmcwhirter:
Sure. Just tripped over this from my JDK11 work. I’m not picky.

gilles-duboscq:
Bob, i'm closing this since this is superseded by #1245

PR #1174, opened 2019-04-17
Fold base-pointer-size as a constant.

bobmcwhirter:
Helps to allow for native-image to be native-image'd on aarch64.


bobmcwhirter:
Ping @sanzinger or @cstancu...

peter-hofer:
Thank you for the submission!

PR #1188, opened 2019-04-24
Allow LINUX_JNI platform for JDK 11

### LINKED ISSUE #1187: [native-image] Using LINUX_JNI fails for Java 11
At least 2 classes are referred that don't exist anymore in Java 11, resulting in NullPointerExceptions. The classes are referred to from PosixJavaLangSubstituteFeature.duringSetup and PosixJavaNetSubstitutionsFeature.duringSetup

johanvos:
This fixes #1187

### DIFF
--- a/substratevm/src/com.oracle.svm.core.jdk9/src/com/oracle/svm/core/jdk/Target_java_util_logging_SimpleFormatter_JDK9OrLater.java
+++ b/substratevm/src/com.oracle.svm.core.jdk9/src/com/oracle/svm/core/jdk/Target_java_util_logging_SimpleFormatter_JDK9OrLater.java
@@ -43,7 +43,7 @@ class FormatAccessors {
 
     private static String format = null;
 
-    public static String getFormat() {
+    public static String getFormat(java.util.logging.SimpleFormatter parent) {
         if (format == null) {
             /*
              * If multiple threads are doing the initialization at the same time it is not a problem
@@ -53,6 +53,10 @@ class FormatAccessors {
 
         }
         return format;
+    }
+
+    public static void setFormat(java.util.logging.SimpleFormatter parent, String f) {
+        format = f;
     }
 }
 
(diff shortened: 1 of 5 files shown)


bobvandette:
If you look in the sources, you'll see that in JDK11, there's a #ifndef STATIC_BUILD around this function so it will NOT be provided by the static libraries.  You're going to have to find a way of making this function only be available for JDK11+.

johanvos:
Is there a reason why it is not provided by the static libraries? Or can we change it in JDK11? (first in jdk/jdk and then backport to 11?)

bobvandette:
The problem is that these functions are defined in the Hotspot VM as well as libjava which will result in a link error if you attempt to combine all static libs and VM into a single executable binary.  You can propose a fix that makes these functions unique but you can't just simply add them to libjava.

johanvos:
Since they are defined in HotSpot VM, I think it makes sense to define them in JvmFuncs.c as well.
But for Java 8, jio_snprintf is still defined in libjava.a so there is a link error when jio_snprintf is added in JvmFuncs.c and linked with libjava.a .
For Java 11, jio_snprintf is already under #ifndef STATIC_BUILD in libjava.

bobvandette:
Yes, we really didn't support static builds in JDK8.  I should have updated the patch that builds the static libraries for GraalVM to add the #ifndef STATIC_BUILD.  That's probably the best solution but will take time.

johanvos:
It might be easier to conditionally add the jio_snprintf in JvmFuncs.c, with something like an #ifndef JDK8 ?

bobvandette:
I was just looking into that.  I haven't tried this but Aleksandar just told me that if you add "use_jdk_headers" : True in the suite.py jvm project, the path to jni.h should be added to the compile.  You can then use JNI_VERSION defines.  You'll have to fix a few issues where I've manually added some jni.h defines but this seems like a workable solution.

vjovanov:
This PR has been merged.

PR #1193, opened 2019-04-25
Merge two consecutive ldr/str to ldp/stp for AArch64.

XiaohongGong:
This patch will merge two ldr/str to one ldp/stp, like:

```
    str x1, [sp, #128]
    str x2, [sp, #120]
```

will be merged to:

```
    stp x2, x1, [sp, #120]
```

The same to the load operation.

We have to begin this optimization after register allocation,
so that some loads/stores generated for spilling can also be optimized.
Besides we should take care of the following tips:
1. The second ldr/str is not bound to a label.
2. We need to update the Infopoint list when an ImplicitException is
recorded at the second ldr/str position. In this patch, I updated its
position to the first ldr/str position.
3. Some basic ldr/str instruction form checkings when merging to ldp/stp.
More details can be found in my patch.

Here are two jmh cases for this patch:

```
private static int pairLoad() {
    int ret = 0;
    for (int i = 0; i < 10000000; i++) {
        A aa = arr[i];
        ret += aa.a + aa.b;
    }
    return ret;
}

@Benchmark
public int jmhPairLoad() {
    return pairLoad();
}

private static int pairStore(int m, int n) {
    int ret = 0;
    int factor = m + n;
    for (int i = 0; i < 100000; i++) {
        A.c = i;
        A.d = i;
        ret += factor * (A.c + A.d);
    }
    return ret;
}

@Benchmark
public int jmhPairStore() {
    return pairStore(5, 6);
}
```

Add the results:

```
                                            Without this patch
Benchmark                 Mode    Cnt       Score           Error         Units
LdStMerging.jmhPairLoad       avgt     15    120222.590  ? 5719.105    us/op
LdStMerging.jmhPairStore      avgt     15      114.279    ?  0.055     us/op

                                               With this patch
Benchmark                 Mode    Cnt       Score           Error         Units
LdStMerging.jmhPairLoad       avgt     15     116779.081  ? 2964.116   us/op
LdStMerging.jmhPairStore      avgt     15       83.901     ?  1.324      us/op
```

Change-Id: I6c1e3f54e7a1152c41499e9d7aeb50aa7bc62514

### DIFF
--- a/compiler/src/org.graalvm.compiler.asm.aarch64.test/src/org/graalvm/compiler/asm/aarch64/test/AArch64LoadStoreMergingAssemblerTest.java
+++ b/compiler/src/org.graalvm.compiler.asm.aarch64.test/src/org/graalvm/compiler/asm/aarch64/test/AArch64LoadStoreMergingAssemblerTest.java
@@ -1,0 +1,311 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, Arm Limited. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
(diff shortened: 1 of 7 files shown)


XiaohongGong:
Hi there,
Can someone please take a look at this patch? Thank you!

sanzinger:
Looks good to me. Tests are ok and we can merge this change.

PR #1199, opened 2019-04-26
Always have $$StringConcat indy to be EAGER

### LINKED ISSUE #1198: [native-image] StringConcat indy delayed
Running native-image with the following class will fail with the following error: 

`Error: Class that is marked for delaying initialization to run time got initialized during image building: Hello3$$StringConcat/0x00000007c09ff440`

```
public class Hello3 {

    String a = "a";
    String b = "b";

    public static void main(String[] args) throws Exception  {
        System.err.println("Hello, world");
        Hello3 me = new Hello3();
        me.bar();
    }

    void bar() {
        System.err.println("a + b = "+a+b);
    }

}
```

johanvos:
This fixes #1198.
Similar to Lambda's, have StringConcat classes eager.


vjovanov:
This change looks good. What I need to do in addition (in the future) is make the names stable for PGO. When I do that, we should register those classes explicitly so users have a reason that explains why this class was initialized at build time.

This is already merged, just the build bot did not pick it up:
https://github.com/oracle/graal/commit/b0e65d25d7b27c692d2454423f45b6ca6788cdd4

PR #1277, opened 2019-05-14
Implement ClassLoader.getPlatformClassLoader()

### LINKED ISSUE #1276: [native-image] no support for platformclassloader
While the relevance of ClassLoader.getPlatformClassLoader() is different, it is important that calls to it do not cause the VM to halt. Specifically, calls in the JDK core libs rely on ClassLoader.getPlatformClassLoader() to return a valid object.
Currently, ClassLoader.getPlatformClassLoader returns an Exception in NativeImage.

johanvos:
add add-opens for jdk.internal.logger (used by e.g. Formatter in core JDK classes)

This fixes #1276


vjovanov:
LGTM. It just needs to compile on JDK 8.

Issue #1275, opened 2019-05-14
AArch64: AssertionError in unittest RegisterPackageNFITest


pli-arm:
@axel22 
Graal unittest RegisterPackageNFITest fails on AArch64 after below patch merged.
```
commit 14fef247085c2e2726282f0c7e9e12c8d4ac82f9
Author: Aleksandar Prokopec <aleksandar.prokopec@oracle.com>
Date:   Wed Apr 24 15:36:28 2019 +0200

    Ensure that canonicalization of SignedRemNode can analyze usages of the
    existing node in more scenarios.
```
Failure is caused by an assertion error. See stack trace below.
```
java.lang.AssertionError: HotSpotMethod<OptimizedCallTarget.callRoot(Object[])> HotSpotMethod<AArch64IntegerArithmeticSnippets.iremSnippet(int, int, boolean)>
        at jdk.internal.vm.compiler/org.graalvm.compiler.graph.NodeSourcePosition.verifyRootMethod(NodeSourcePosition.java:80)
        at jdk.internal.vm.compiler/org.graalvm.compiler.graph.Graph.verifySourcePositions(Graph.java:1188)
        at jdk.internal.vm.compiler/org.graalvm.compiler.phases.common.inlining.InliningUtil.updateSourcePosition(InliningUtil.java:716)
        at jdk.internal.vm.compiler/org.graalvm.compiler.replacements.SnippetTemplate.inlineSnippet(SnippetTemplate.java:1539)
        at jdk.internal.vm.compiler/org.graalvm.compiler.replacements.SnippetTemplate.instantiate(SnippetTemplate.java:1604)
        at jdk.internal.vm.compiler/org.graalvm.compiler.replacements.InstanceOfSnippetsTemplates.lower(InstanceOfSnippetsTemplates.java:101)
        at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.meta.DefaultHotSpotLoweringProvider.lower(DefaultHotSpotLoweringProvider.java:266)
        at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.aarch64.AArch64HotSpotLoweringProvider.lower(AArch64HotSpotLoweringProvider.java:75)
        at jdk.internal.vm.compiler/org.graalvm.compiler.nodes.java.InstanceOfNode.lower(InstanceOfNode.java:114)
        at jdk.internal.vm.compiler/org.graalvm.compiler.phases.common.LoweringPhase$Round.process(LoweringPhase.java:476)
[...]

PR #1298, opened 2019-05-17
Fix missing source position issue.

XiaohongGong:
This fixes issue #1275 (https://github.com/oracle/graal/issues/1275)

In Graal, nodes with potential runtime exception (like SignedRemNode)
should have Java source position bound. And this position information
should be well propagated when lowering or canonicalizing a node. But
in the PartialEscapePhase, when replacing nodes to new ones, position
information is lost.

This patch surrounds node replacement code in GraphEffectList.java with
the graph.withNodeSourcePosition() call to make sure we won't lose the
source position info from the old node.

Change-Id: Ife36ebe1ea847382c26ec6d66e93a3f49ef97d15


XiaohongGong:
Hi there,

This PR fixes https://github.com/oracle/graal/issues/1275 on AArch64. Can you please take a look at it? Thanks!

PR #1309, opened 2019-05-20
Fix search for "nearby" bin/java(.exe).

bobmcwhirter:
By default, native-image would jump too high up before
looking for a nearby bin/java(.exe) prior to just looking
around in $JAVA_HOME.

Still, searches the same paths, but slowly jumps up from
the `rootDir` instead of immediately searching `rootDir.getParent()`.

Additionally, refactored to avoid duplicate logic.

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -440,21 +440,39 @@ public class NativeImage {
 
         @Override
         public Path getJavaExecutable() {
-            Path javaHomePath = rootDir.getParent();
-            Path binJava = Paths.get("bin", OS.getCurrent() == OS.WINDOWS ? "java.exe" : "java");
-            if (Files.isExecutable(javaHomePath.resolve(binJava))) {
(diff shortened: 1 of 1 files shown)


olpaw:
Once the trivial/style related issued are fixed (i.e. the travis gates pass) I will pick it up.

bobmcwhirter:
Right on, will fix and push again in a few.

Thanks!

bobmcwhirter:
Done

olpaw:
@bobmcwhirter I think latest master changes on master make this PR obsolete. Do you agree?

bobmcwhirter:
Seems Giles is fixing it in parallel yeah. 

Would be easier to track if all contributors (including ORCL) used the same PR process.

bobmcwhirter:
Actually Giles is doing AdoptOpenJDK (another PR of mine). I’ll have to check master commits wrt nearby bin/java once I’m not on my phone.

olpaw:
> Would be easier to track if all contributors (including ORCL) used the same PR process.

I agree.

bobmcwhirter:
Yah, seems fixed by

https://github.com/oracle/graal/commit/404359e56616d386531933d6a8d65fc95cbcbc17

PR #1311, opened 2019-05-20
Cleanup on aisle aarch64.

bobmcwhirter:
Renaming methods to better match what they do.
Renaming classes to better express what they do.


bobmcwhirter:
I think your CI is losing its mind.

    Imported suite 'graal-js' not found (binary or source).

sanzinger:
Looks good to merge. Will do that tomorrow.

PR #1318, opened 2019-05-22
Allow os.name and os.arch to be set when cross-compiling

### LINKED ISSUE #1317: [native image] os.name and os.arch System properties for target
Currently, os.name and os.arch are unmodifiable. The code mentions
/* We do not support cross-compilation for now. */

As a consequence, compiling for a different target platform will have an incorrect os.name and os.version System property.

johanvos:
This fixes #1317

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
@@ -56,8 +56,6 @@ public abstract class SystemPropertiesSupport {
                      * in other classes, so changing them would be tricky.
                      */
                     "line.separator", "path.separator", "file.separator",
-                    /* We do not support cross-compilation for now. */
-                    "os.arch", "os.name",
                     /* For our convenience for now. */
                     "file.encoding", "sun.jnu.encoding",
                     "java.class.version",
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -100,6 +98,11 @@ protected SystemPropertiesSupport() {
         initializeProperty("java.ext.dirs", "");
         initializeProperty("java.library.path", "");
 
+        String targetName = System.getProperty("svm.targetName");
```
olpaw: This is quite loosely coupled (i.e. allows to lie about `os.name` and  `os.arch` independent from actually doing cross-compilation). Is this a temporary solution? (I.e. the setting of `os.name` and  `os.arch` should be driven by actual cross-compilation options.)

johanvos: svm.targetArch is already used for actual cross-compilation, https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/macho/MachOObjectFile.java#L83

olpaw: Hmm, maybe I miss something but that seems to just tweak how relocations are generated in the output object file. Would it make more sense to derive `os.arch` from the `org.graalvm.nativeimage.Platform` we are building for?

olpaw: I.e. something like
```
if (Platform.includedIn(Platform.AArch64.class)) {
   ...
```

PR #1325, opened 2019-05-23
Track constantReason for VMConstants along with JavaConstants.

bobmcwhirter:


bobmcwhirter:
I think that CI failure might be... spurious?

bobmcwhirter:
Ping?

PR #1326, opened 2019-05-23
No-op module reflection modifications.

bobmcwhirter:
The WeakPairMap inside Module$ReflectionData was changing,
sometimes, during heap generation of native-image.

Delete the class and no-op the relevant methods that used it.

Might be a sub-optimal solution, but avoids spurious native-image
failures at this point when using JDK11+.


bobmcwhirter:
Ping?

bobmcwhirter:
The `Target_java_lang_Module` already existed in this location. I just added a few more stubs.

Happy to move it though.

bobmcwhirter:
Actually, can't move it due to `DynamicHub` referencing it.

PR #1352, opened 2019-06-04
Fix condition for memory bulk zeroing

pli-arm:
In the snippet code for allocating new array/object, method fillMemory()
calls ZeroMemoryNode.zero() without checking the value to fill is zero
or not. And, to be more accurate, the value of "size - offset", which is
the actual number of bytes to be zeroed, should be compared against the
value of "MinimalBulkZeroingSize".


pli-arm:
@mur47x111 Could you help take a look at this? I see you are the original author of the bulk zeroing.

adinn:
That looks good to me Pengfei!

mur47x111:
Thanks! Will start integrating this.

PR #1355, opened 2019-06-06
Fix rounding conditions for SignedDivNode when canonicalized.

XiaohongGong:
The conditions that a division can be replaced by one shift
instruction are:
1. The divisor is power-of-2.
2. The dividend is positive, or its low bits are always 0.
Otherwise, the dividend needs to be rounded if we still want to
replace the division by shift when the divisor is power-of-2.

But the conditions to round the dividend in Graal is:
"if (stampX.canBeNegative() || (stampX.upMask() & (abs - 1)) != 0)"
So if stampX is positive, and its low bits are not always 0, it will
also be rounded. This will also generate the rounding instructions,
which is redundant.

This patch modifies the conditions to make sure that any positive
dividend is not rounded.

Change-Id: I3bf4eceb11f50d3aa6e0bc606f7927f20259a477

### DIFF
--- a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/SignedDivNode.java
+++ b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/SignedDivNode.java
@@ -122,7 +122,7 @@ public class SignedDivNode extends IntegerDivRemNode implements LIRLowerable {
             ValueNode dividend = forX;
             int log2 = CodeUtil.log2(abs);
             // no rounding if dividend is positive or if its low bits are always 0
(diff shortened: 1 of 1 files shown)


XiaohongGong:
Hi there,
Can anyone please help to take a look at this patch? Thank you!

XiaohongGong:
Hi @mur47x111 , could you please take a look at this PR? Thank you!

mur47x111:
Sorry for the delay. Will do.

mur47x111:
FYI, I have added a test case https://github.com/oracle/graal/commit/6e3d243fefa278e962f19058440c967fc5835afd

XiaohongGong:
> FYI, I have added a test case [6e3d243](https://github.com/oracle/graal/commit/6e3d243fefa278e962f19058440c967fc5835afd)

Thank you!

PR #1359, opened 2019-06-06
AArch64 - Read return address correctly.

bobmcwhirter:
Since the FrameContext uses `stp fp, lr, ...` to store the
`fp` and `lr` as the first two words in the frame,
the layout is actually:

--------- frame start ---------
lr (word)
fp (word)
--------- rest of frame -------
...

Therefore, when walking the stack, to read the return value
of a given frame with an effective frame-start address, the
immediately first word is the `lr`. Reading the second results
in reading the `fp` which ultimately results in a bus error
on AArch64.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/aarch64/AArch64FrameAccess.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/aarch64/AArch64FrameAccess.java
@@ -48,13 +48,13 @@ class AMD64FrameAccessFeature implements Feature {
 public class AArch64FrameAccess extends FrameAccess {
     @Override
     public CodePointer readReturnAddress(Pointer sourceSp) {
-        /* Read the return address, which is stored one word below the stack pointer. */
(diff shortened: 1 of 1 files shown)


bobmcwhirter:
@sanzinger This fixes that thing...

bobmcwhirter:
fwiw, extant `AArch64FrameMap` confirms the expected `lr` immediately at the start of the frame.

https://github.com/oracle/graal/blob/master/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64FrameMap.java#L35-L63

bobmcwhirter:
Also, the previous CI failures seem entirely bogus because CI couldn't find `java` early in the builds.

I've rebased and repushed to trigger a new build anyhow.

bobmcwhirter:
Rebased to force CI to try again, harder.

bobmcwhirter:
Ping?

Can I get a review/merge of this before too much time passes and I have to rebase again?

adinn:
This looks correct to me. In fact I find it somewhat bizarre that the original code was written as was.

bobmcwhirter:
This blocks further aarch64 progress.

PR #1365, opened 2019-06-07
fix openjdk8 download in Travis

dougxc:


bobmcwhirter:
Also seems your integration with GitHub may be broken.

CI passed this build, but GH thinks the check is still in-progress...

dougxc:
Weird. If it's still a problem once this fix is merged, I'll investigate further.

bobmcwhirter:
Metacircular CI irony, right there.

PR #1371, opened 2019-06-10
Avoid NPE when parsing static initializers.

bobmcwhirter:
Due to the fact an unresolved type may (previous to this commit)
throw a parse exception when attempting to perform automatic unsafe
substitutions, this can result in a null slot being left in the
universe of types array.

Which then ends up being an unresolve type error of type 'null'
subsequently when attempting to resolve null as a type.

Allow the graph-builder to respect the the default false value of
`unresolvedIsError` by not throwing in such a case.

### DIFF
--- a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/graphbuilderconf/GraphBuilderConfiguration.java
+++ b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/graphbuilderconf/GraphBuilderConfiguration.java
@@ -27,6 +27,7 @@ package org.graalvm.compiler.nodes.graphbuilderconf;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 import org.graalvm.compiler.core.common.type.StampPair;
 
@@ -196,6 +197,8 @@ public final class GraphBuilderConfiguration {
 
     private final boolean eagerResolving;
     private final boolean unresolvedIsError;
+    // Optional due to visibility of NativeImageOptions
+    private final Optional<Boolean> allowIncompleteClasspath;
     private final BytecodeExceptionMode bytecodeExceptionMode;
     private final boolean omitAssertions;
     private final List<ResolvedJavaType> skippedExceptionTypes;
@@ -226,6 +229,7 @@ public final class GraphBuilderConfiguration {
 
     private GraphBuilderConfiguration(boolean eagerResolving,
                     boolean unresolvedIsError,
+                    Optional<Boolean> allowIncompleteClasspath,
                     BytecodeExceptionMode bytecodeExceptionMode,
                     boolean omitAssertions,
                     boolean insertFullInfopoints,
@@ -236,6 +240,7 @@ public final class GraphBuilderConfiguration {
         this.eagerResolving = eagerResolving;
         this.unresolvedIsError = unresolvedIsError;
         this.bytecodeExceptionMode = bytecodeExceptionMode;
+        this.allowIncompleteClasspath = allowIncompleteClasspath;
         this.omitAssertions = omitAssertions;
         this.insertFullInfopoints = insertFullInfopoints;
         this.trackNodeSourcePosition = trackNodeSourcePosition;
@@ -254,6 +259,7 @@ public final class GraphBuilderConfiguration {
         GraphBuilderConfiguration result = new GraphBuilderConfiguration(
                         eagerResolving,
                         unresolvedIsError,
+                        allowIncompleteClasspath,
                         bytecodeExceptionMode,
                         omitAssertions,
                         insertFullInfopoints,
@@ -274,6 +280,21 @@ public final class GraphBuilderConfiguration {
         return new GraphBuilderConfiguration(
                         eagerResolving,
                         newUnresolvedIsError,
+                        allowIncompleteClasspath,
+                        bytecodeExceptionMode,
+                        omitAssertions,
+                        insertFullInfopoints,
+                        trackNodeSourcePosition,
+                        retainLocalVariables,
+                        skippedExceptionTypes,
+                        plugins);
+    }
+
+    public GraphBuilderConfiguration withAllowIncompleteClasspath(boolean newAllowIncompleteClasspath) {
+        return new GraphBuilderConfiguration(
+                        eagerResolving,
+                        unresolvedIsError,
+                        Optional.of(newAllowIncompleteClasspath),
                         bytecodeExceptionMode,
                         omitAssertions,
                         insertFullInfopoints,
@@ -287,6 +308,7 @@ public final class GraphBuilderConfiguration {
         return new GraphBuilderConfiguration(
                         newEagerResolving,
                         unresolvedIsError,
+                        allowIncompleteClasspath,
                         bytecodeExceptionMode,
                         omitAssertions,
                         insertFullInfopoints,
@@ -300,6 +322,7 @@ public final class GraphBuilderConfiguration {
         return new GraphBuilderConfiguration(
                         eagerResolving,
                         unresolvedIsError,
+                        allowIncompleteClasspath,
                         bytecodeExceptionMode,
                         omitAssertions,
                         insertFullInfopoints,
@@ -312,6 +335,7 @@ public final class GraphBuilderConfiguration {
     public GraphBuilderConfiguration withBytecodeExceptionMode(BytecodeExceptionMode newBytecodeExceptionMode) {
         return new GraphBuilderConfiguration(eagerResolving,
                         unresolvedIsError,
+                        allowIncompleteClasspath,
                         newBytecodeExceptionMode,
                         omitAssertions,
                         insertFullInfopoints,
@@ -325,6 +349,7 @@ public final class GraphBuilderConfiguration {
         return new GraphBuilderConfiguration(
                         eagerResolving,
                         unresolvedIsError,
+                        allowIncompleteClasspath,
                         bytecodeExceptionMode,
                         newOmitAssertions,
                         insertFullInfopoints,
@@ -338,6 +363,7 @@ public final class GraphBuilderConfiguration {
         return new GraphBuilderConfiguration(
                         eagerResolving,
                         unresolvedIsError,
+                        allowIncompleteClasspath,
                         bytecodeExceptionMode,
                         omitAssertions,
                         newInsertFullInfopoints,
@@ -351,6 +377,7 @@ public final class GraphBuilderConfiguration {
         return new GraphBuilderConfiguration(
                         eagerResolving,
                         unresolvedIsError,
+                        allowIncompleteClasspath,
                         bytecodeExceptionMode,
                         omitAssertions,
                         insertFullInfopoints,
@@ -364,6 +391,7 @@ public final class GraphBuilderConfiguration {
         return new GraphBuilderConfiguration(
                         eagerResolving,
                         unresolvedIsError,
+                        allowIncompleteClasspath,
                         bytecodeExceptionMode,
                         omitAssertions,
                         insertFullInfopoints,
@@ -405,6 +433,7 @@ public final class GraphBuilderConfiguration {
         return new GraphBuilderConfiguration(
                         /* eagerResolving: */ false,
                         /* unresolvedIsError: */ false,
+                        /* allowIncompleteClasspath (defer to NativeImageOptions): */ Optional.empty(),
                         BytecodeExceptionMode.Profile,
                         /* omitAssertions: */ false,
                         /* insertFullInfopoints: */ false,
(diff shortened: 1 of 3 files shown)


cstancu:
@bobmcwhirter can you please share an example that exposes the issue? The automatic unsafe
substitutions is best effort and, in general, it should be side effect free in case of issues like a parse exception.

bobmcwhirter:
We have a class that ultimately gets fully deleted. But before that’s put into effect apparently someone tries to parse it. It’s static unit references a non-available class. Which does throw. But after the nextTypeId is consumed and leaves a null in the types array.

bobmcwhirter:
Fully substituted I mean.

cstancu:
Can you please share code so I can replicate the problem?

bobmcwhirter:
Sure I’ll try to work up a reproducer

bobmcwhirter:
Reproducing repository:

https://github.com/bobmcwhirter/incomplete-unsafe

Pre-built jar:

https://github.com/bobmcwhirter/incomplete-unsafe/releases/download/1.0.0.Alpha1/incomplete-unsafe-1.0.0.Alpha1.jar

Analysis:

When using `master#HEAD` against this jar, the result is:

```
$ mx native-image -cp /Users/bob/repos/redhat/incomplete-unsafe/target/incomplete-unsafe-1.0.0.Alpha1.jar com.redhat.incompleteunsafe.Main
[com.redhat.incompleteunsafe.main:90426]    classlist:   1,380.22 ms
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.oracle.svm.util.ReflectionUtil (file:/Users/bob/repos/oracle/graal/substratevm/mxbuild/dists/jdk11/pointsto.jar) to field java.nio.file.Files$FileTypeDetectors.installedDetectors
WARNING: Please consider reporting this to the maintainers of com.oracle.svm.util.ReflectionUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
[com.redhat.incompleteunsafe.main:90426]        (cap):   1,504.42 ms
[com.redhat.incompleteunsafe.main:90426]        setup:   3,077.06 ms
[com.redhat.incompleteunsafe.main:90426]     analysis:     490.63 ms
Fatal error: com.oracle.svm.core.util.VMError$HostedError: Found unsupported type: null
	at com.oracle.svm.core.util.VMError.shouldNotReachHere(VMError.java:66)
	at com.oracle.svm.hosted.SVMHost.dynamicHub(SVMHost.java:229)
	at com.oracle.svm.hosted.analysis.Inflation.checkType(Inflation.java:148)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1083)
	at com.oracle.svm.hosted.analysis.Inflation.checkObjectGraph(Inflation.java:136)
	at com.oracle.graal.pointsto.BigBang.checkObjectGraph(BigBang.java:603)
	at com.oracle.graal.pointsto.BigBang.finish(BigBang.java:558)
	at com.oracle.svm.hosted.NativeImageGenerator.runPointsToAnalysis(NativeImageGenerator.java:683)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:522)
	at com.oracle.svm.hosted.NativeImageGenerator.lambda$run$0(NativeImageGenerator.java:440)
```

When using this PR, the result is a binary native-image that works correctly.

The processed jar uses *no* Unsafe calls and requires no substitution. 

When the automatic unsafe substitutions attempt to execute, they seem to be occurring before normal explicit substitutions have occurred.

This results in the following flow:

`AnalysisUniverse#createType(..)` is called for `WrapperClass`

An `AnalysisType` is created, incrementing the atomic counter, reserving a slot in the `typesById`.

https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/meta/AnalysisUniverse.java#L261

https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/meta/AnalysisType.java#L193

Then `SVMHost#registerType(..)` is called:

https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SVMHost.java#L189

Which attempts automatic substitutions:

https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SVMHost.java#L199

Which calls `UnsafeAutomaticSubstitutionsProcessor#computeSubstitutions(..)`:

https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java#L301

Which then attempts to parse the class, which ultimately results in an exception being actually thrown:

https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java#L318

This thrown exception is swallowed somewhere and otherwise un-noted by the iteration of the types.

Causing `AnalysisUniverse#createType(..)` to exit early.  

With this PR, `mx native-image` can complete successfully.

This PR attempts to avoid the issue, maybe naively, by treating unresolved types during the parse of the type as non-errors, because as you noted, unsafe substitutions are best-effort.  And ultimately the error gets ignored *at this point* any how, and simply kicks the can further down the road for when the type graph is being later checked, and the `typesById` contains nulls because `AnalysisUniverse#createType(...)` reserved an index that it never actually populated.

A better solution, perhaps, would be to process explicit substitutions first. In this case, doing so would completely remove the `static {...}` initializer block and would avoid this unresolved type error completely, as it'd never be parsed, since the class was wholly substituted.

bobmcwhirter:
Also, fwiw, this change really just makes it so that the graph-builder-configure item of `unresolvedIsError=false` is respected. Right now, `=false` is ignored, and unresolved are actually indeed treated as errors, as that portion of the code is only checking the `NativeImageOptions.AllowIncompleteClasspath` option, ignoring the `unresolvedIsError` setting.

bobmcwhirter:
Anything else you need?

bobmcwhirter:
fwiw, this is blocking Quarkus from using Graal 19+.

cstancu:
@bobmcwhirter I'll look into it today. Thank you for the in-depth analysis!

cstancu:
The problem was "introduced" by https://github.com/oracle/graal/commit/9f1ded35b786e2cb9cb069c2c610a27c30b88f6d. Previously the bytecode parsing for `UnsafeAutomaticSubstitutionProcessor` was done via `org.graalvm.compiler.java.BytecodeParser`. That commit fixes problems caused by missing return types while analysing a static initializer. It switches the parsing for `UnsafeAutomaticSubstitutionProcessor` to a subclass of `SharedBytecodeParser` specifically because `SharedBytecodeParser` already does all the handling of incomplete classpaths. 

Now, running your test I can reproduce the reported problem, but the problem is avoided when running with `--allow-incomplete-classpath` becasue then, instead of throwing an exception from `SharedBytecodeParser.handleUnresolvedType()`, we alter the graph by synthesizing a `throw NoClassDefFoundError`. This is essentially what your fix does as well, because the graph builder configuration for the `UnsafeAutomaticSubstitutionProcessor` uses the default value of `unresolvedIsError` which is false `false`.

And this is essentially the original behavior of [`BytecodeParser.handleUnresolvedInvoke`](https://github.com/oracle/graal/blob/master/compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java#L1195): short circuit the parsing and introduce a synthetic node, a `DeoptimizeNode` in the general case. This was also the behavior during parsing for `UnsafeAutomaticSubstitutionProcessor` before the commit I mentioned above.

I think a better solution is to parameterize the `SharedBytecodeParser` with an `allowIncompleteClassPath` field which we always set to `true` for the `UnsafeAutomaticSubstitutionProcessor` and set it to the value of `-H:AllowIncompleteClasspath` for all other parsing needs.

You could even argue that we already have the parameter that we need, and that parameter is `graphBuilderConfig.unresolvedIsError`, but I need to think more about the implications of tying in the semantics of `-H:AllowIncompleteClasspath` and `graphBuilderConfig.unresolvedIsError`. That's why I was hesitant merging this PR as is. Although it clearly fixes the problem that your reported it essentially changes the semantics of `-H:AllowIncompleteClasspath`.

bobmcwhirter:
That all makes sense. I’ll adjust Monday and resubmit.

bobmcwhirter:
That CI failure is ... not my doing.

bobmcwhirter:
@cstancu how’s this look?

thomaswue:
Yes, we are working on fixing that CI failure related to Sulong. I will ping @cstancu.

rschatz:
See https://github.com/oracle/graal/pull/1395 for the gate failure fix. I'll merge it asap.

cstancu:
@bobmcwhirter the changes look good. I'll bring them in today. Thank you!

cstancu:
@bobmcwhirter I merged this but refactored it slightly. First, since `AllowIncompleteClasspath` is a native-image specific option and it shouldn't really be in the base `GraphBuilderConfiguration` I just made `allowIncompleteClassPath` a field of `SharedBytecodeParser`. Secondly, I used the same approach for the processing of unresolved methods and fields, not only types. See https://github.com/oracle/graal/commit/44ca69968a30466cb36bc2b7d9cfb0f96fbd2fbd.

bobmcwhirter:
Works for me.

If there’s a substitute type at the class level can we bail early from the
clinit replacements altogether?

cstancu:
@bobmcwhirter yes, we could to that. Currently we [suppress the warnings](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java#L876) for substituted types. We could simply bail out of the analysis [early](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java#L304) for substituted types.

bobmcwhirter:
Okie dokie. I may work that up.

dmlloyd:
I've just gotten this exception while testing native image building with Java 11 and the latest master (commit id 9185fb8915bfa79586a176bc78f8237ad68d61e1).  The stack trace looked like this:

```
2019-06-20 14:29:48 Fatal error: com.oracle.svm.core.util.VMError$HostedError: Found unsupported type: null
2019-06-20 14:29:48     at com.oracle.svm.core.util.VMError.shouldNotReachHere(VMError.java:66)
2019-06-20 14:29:48     at com.oracle.svm.hosted.SVMHost.dynamicHub(SVMHost.java:229)
2019-06-20 14:29:48     at com.oracle.svm.hosted.analysis.Inflation.checkType(Inflation.java:148)
2019-06-20 14:29:48     at java.base/java.lang.Iterable.forEach(Iterable.java:75)
2019-06-20 14:29:48     at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1083)
2019-06-20 14:29:48     at com.oracle.svm.hosted.analysis.Inflation.checkObjectGraph(Inflation.java:136)
2019-06-20 14:29:48     at com.oracle.graal.pointsto.BigBang.checkObjectGraph(BigBang.java:603)
2019-06-20 14:29:48     at com.oracle.graal.pointsto.BigBang.finish(BigBang.java:558)
2019-06-20 14:29:48     at com.oracle.svm.hosted.NativeImageGenerator.runPointsToAnalysis(NativeImageGenerator.java:683)
2019-06-20 14:29:48     at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:522)
2019-06-20 14:29:48     at com.oracle.svm.hosted.NativeImageGenerator.lambda$run$0(NativeImageGenerator.java:440)
2019-06-20 14:29:48     at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1407)
2019-06-20 14:29:48     at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
2019-06-20 14:29:48     at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
2019-06-20 14:29:48     at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
2019-06-20 14:29:48     at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
2019-06-20 14:29:48     at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
```

Is this expected?

dmlloyd:
Actually, I may not have been up to date with that commit like I thought.  I'm retesting one more time to be sure.

PR #1381, opened 2019-06-14
Add `InnocuousThread` to the warning whitelist.

bobmcwhirter:
Since it's a JDK class that we can't fix, and probably
can't convince the JDK team to use a field instead of
a local (since a local is all that is required), might as
well whitelist it from the unsafe automatic substitutions
warnings.

Meanwhile, let's spell "suppress" correctly while we're there.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
@@ -128,7 +128,7 @@ public class UnsafeAutomaticSubstitutionProcessor extends SubstitutionProcessor 
     private final AnnotationSubstitutionProcessor annotationSubstitutions;
     private final Map<ResolvedJavaField, ComputedValueField> fieldSubstitutions;
 
-    private final List<ResolvedJavaType> supressWarnings;
+    private final List<ResolvedJavaType> suppressWarnings;
 
     private ResolvedJavaMethod unsafeObjectFieldOffsetFieldMethod;
     private ResolvedJavaMethod unsafeObjectFieldOffsetClassStringMethod;
@@ -147,7 +147,7 @@ public class UnsafeAutomaticSubstitutionProcessor extends SubstitutionProcessor 
         this.snippetReflection = snippetReflection;
         this.annotationSubstitutions = annotationSubstitutions;
         this.fieldSubstitutions = new ConcurrentHashMap<>();
-        this.supressWarnings = new ArrayList<>();
+        this.suppressWarnings = new ArrayList<>();
     }
 
     public void init(ImageClassLoader loader, MetaAccessProvider originalMetaAccess, SVMHost hostVM) {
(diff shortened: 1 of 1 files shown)


bobmcwhirter:
Any thoughts on this? I hate seeing warnings during the build, particularly if, as an end-user, I can do nothing about them.

cstancu:
@bobmcwhirter I'll get this merged today. Thanks!

cstancu:
I replicated this and I see 2 warnings: 

```
Warning: RecomputeFieldValue.FieldOffset automatic substitution failed. The automatic substitution registration was attempted because a call to jdk.internal.misc.Unsafe.objectFieldOffset(Class, String) was detected in the static initializer of jdk.internal.misc.InnocuousThread. Detailed failure reason(s): Could not determine the field where the value produced by the call to jdk.internal.misc.Unsafe.objectFieldOffset(Class, String) for the field offset computation is stored. The call is not directly followed by a field store or by a sign extend node followed directly by a field store. 
Warning: RecomputeFieldValue.FieldOffset automatic substitution failed. The automatic substitution registration was attempted because a call to jdk.internal.misc.Unsafe.objectFieldOffset(Class, String) was detected in the static initializer of jdk.internal.misc.InnocuousThread. Detailed failure reason(s): Could not determine the field where the value produced by the call to jdk.internal.misc.Unsafe.objectFieldOffset(Class, String) for the field offset computation is stored. The call is not directly followed by a field store or by a sign extend node followed directly by a field store.
 ```

Looking at the `<clinit>` of `InnocuousThread` the warnings seem to come from these 2 unsafe operations:

```
    long tg = UNSAFE.objectFieldOffset(tk, "group");
    long gp = UNSAFE.objectFieldOffset(gk, "parent");
```

The warnings are indeed false postives because those offsets do not escape the `<clinit>`, therefore don't need to be recomputed. (Note to self: the warning message should show the name of the field for which the offset computation was detected.)

Additionally, `InnocuousThread.<clinit>` contains other offset computations that are correctly intercepted: 

```
THREAD_LOCALS = UNSAFE.objectFieldOffset(tk, "threadLocals");
INHERITABLE_THREAD_LOCALS = UNSAFE.objectFieldOffset(tk, "inheritableThreadLocals");
INHERITEDACCESSCONTROLCONTEXT = UNSAFE.objectFieldOffset(tk, "inheritedAccessControlContext");
CONTEXTCLASSLOADER = UNSAFE.objectFieldOffset(tk, "contextClassLoader");
```

This can be observed by setting the log level to info via `-H:UnsafeAutomaticSubstitutionsLogLevel=2`:

```
Info:RecomputeFieldValue.FieldOffset substitution automatically registered for jdk.internal.misc.InnocuousThread.THREAD_LOCALS, target element java.lang.Thread.threadLocals.
Info:RecomputeFieldValue.FieldOffset substitution automatically registered for jdk.internal.misc.InnocuousThread.INHERITABLE_THREAD_LOCALS, target element java.lang.Thread.inheritableThreadLocals.
Info:RecomputeFieldValue.FieldOffset substitution automatically registered for jdk.internal.misc.InnocuousThread.INHERITEDACCESSCONTROLCONTEXT, target element java.lang.Thread.inheritedAccessControlContext.
Info:RecomputeFieldValue.FieldOffset substitution automatically registered for jdk.internal.misc.InnocuousThread.CONTEXTCLASSLOADER, target element java.lang.Thread.contextClassLoader.

```

bobmcwhirter:
It seems that local and stack “usages” suffice to cause a warning. I’d
argue maybe only warn if a non-static non-final field. And I’d probably
still whitelist classes we can’t fix.

PR #1389, opened 2019-06-17
Infer stamp for the dividend of integer division and remainder.

XiaohongGong:
This patch creates more opportunities to the following optimizations inside
a counted loop when "d" is a power-of-2 constant:

    "x / d"  ->  "x >> log2(d)"
    "x % d"  ->  "x & (d - 1)"

The above optmization can be applied only if the dividend("x") is known to be
positive. Currently the graal compiler supports this canonicalization. But the
type of the dividend is unrestricted, which means it is not bounded. So most
of the integer divisions and remainders still cannot be canonicalized although
the dividend is positive. So if the compiler can compute the boundary of the
type of the dividend, it can be known as positive, negative or others.

This patch will compute a bounded stamp for the loop variable of a counted loop.
If the variable is used as the dividend of the division or remainder inside of
the loop, the bounded stamp will be used as the real stamp of the dividend.

Directly changing the stamp of the loop variable may generate some issues, for
example it may make a loop endless, or it may influence other loop optimizations.
So this patch just adds a new stamp named "loopStamp" for the counted loop which
represents the stamp of the loop variable only when it is used inside of the loop.
And as I mentioned above, the "loopStamp" will be treated as the new stamp for the
dividend. This operation is applied before the first CanonicalizerPhase, so that
the division and remainder can be optimized as early as possible.

Change-Id: Ia87221897ad67a0ee08bdf4562f357943612426e

### DIFF
--- a/compiler/mx.compiler/suite.py
+++ b/compiler/mx.compiler/suite.py
@@ -1215,7 +1215,10 @@ suite = {
     "org.graalvm.compiler.phases.common" : {
       "subDir" : "src",
       "sourceDirs" : ["src"],
-      "dependencies" : ["org.graalvm.compiler.phases"],
+      "dependencies" : [
+        "org.graalvm.compiler.phases",
+        "org.graalvm.compiler.loop",
+      ],
       "annotationProcessors" : [
         "GRAAL_NODEINFO_PROCESSOR",
         "GRAAL_OPTIONS_PROCESSOR"
@@ -1307,8 +1310,7 @@ suite = {
       "subDir" : "src",
       "sourceDirs" : ["src"],
       "dependencies" : [
-     "org.graalvm.compiler.loop",
-     "org.graalvm.compiler.phases.common",
+        "org.graalvm.compiler.phases.common",
        ],
       "annotationProcessors" : ["GRAAL_OPTIONS_PROCESSOR"],
       "checkstyle" : "org.graalvm.compiler.graph",
--- a/compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/phases/HighTier.java
+++ b/compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/phases/HighTier.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2019, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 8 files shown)


gilles-duboscq:
Hi XiaohongGong,

The stamp you compute is off by one stride in general, it only holds after the counter check. Since you use it in any `IntegerDivRemNode` usage of the counter, that could be before the counter check.

> Directly changing the stamp of the loop variable may generate some issues, for
example it may make a loop endless, or it may influence other loop optimizations.

I agree this can influence other optimizations but if it is correct it should definitely not lead to infinite loops.

Deriving stamps from induction variables is something we plan to do in general and i'm not sure it's worth adding specialized code for div/rem at the moment.

XiaohongGong:
Hi @gilles-duboscq,
    What do you mean about the "counter check"? Does the counter mean the `InductionVariable` of a counted loop? 

Thanks,
XiaohongGong

gilles-duboscq:
I mean the branch that looks like
```
if (counter >= end) {
 break;
}
```
At the beginning of the loop.
Before that, the counter could be past the "end".

XiaohongGong:
I understand what you mean. Thank you! Yes, the stamp I compute is off by one stride in general, that's why  I didn't set the new stamp to the loop variable directly.  The reason I compute the stamp like this is try to also optimize rem/div like this:

`for (int i = 10000; i >= 0; i--) { int x = i % 8}`

In general, the stamp for `i` should be `-1 ~ 10000`. Because the lowerBound is `-1`, the `i` will not be treated as a positive value. So the rem inside of the loop cannot be optimized. But actually the `i` inside the loop can be treated as a positive. So this will miss some optimization opportunities.

XiaohongGong:
I know it's better to reset the stamp for the loop variable, and just do nothing for the `IntegerDivRemNode`. But I don't know how to handle this problem, do you have any better ideas? Thanks!

gilles-duboscq:
> Yes, the stamp I compute is off by one stride in general, that's why I didn't set the new stamp to the loop variable directly.

Right but you would also have to check that the `IntegerDivRemNode` is in the body of the loop. I suppose right it would work because a `IntegerDivRemNode` before the counter check would prevent the loop from being detected as counted but we really don't want to rely on this kind of hard-to-discover invariants that might be broken in the future.

This case with the loop counting down could be handled by something like the `ConditionalEliminationPhase` which would give you a good stamp for `i` in the body of the loop already.

XiaohongGong:
> Right but you would also have to check that the `IntegerDivRemNode` is in the body of the loop.

Yes. Make sense! To make sure the `IntegerDivRemNode` is inside of the loop, I apply this phase at the first step of the HighTier, and check whether the dividend is the loop PhiNode. Maybe there are some other cases which make the `IntegerDivRemNode` outside of the loop body that I didn't cover. It will be safer to explicitly check the  `IntegerDivRemNode`.

XiaohongGong:
> This case with the loop counting down could be handled by something like the `ConditionalEliminationPhase` which would give you a good stamp for `i` in the body of the loop already.

Sounds good! So as you mentioned above that "Deriving stamps from induction variables is something we plan to do in general", I agree that it's not really worth to do something special for div/rem. So how do you think to handle this PR, keep it here to wait for the deriving work or close it currently?

gilles-duboscq:
I would rather close this for now. We can always re-open it if needed.

plokhotnyuk:
@XiaohongGong please look how efficiently can be contemporary compilers: https://godbolt.org/z/qnG_t4

XiaohongGong:
> @XiaohongGong please, look how efficient can be contemporary compilers: https://godbolt.org/z/qnG_t4
> ...speed of their resulting code goes up to [the limits](https://travisdowns.github.io/blog/2019/06/11/speed-limits.html)

Good tools! Thank you so much!

PR #1405, opened 2019-06-19
Better analysis for Unsafe automatic substitutions.

bobmcwhirter:
- If a class is substituted, avoid doing any of the work.
- If an unsafe usage is purely local, it's not a warning.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
@@ -49,6 +49,7 @@ import org.graalvm.compiler.graph.Node;
 import org.graalvm.compiler.graph.iterators.NodeIterable;
 import org.graalvm.compiler.java.BytecodeParser;
 import org.graalvm.compiler.java.GraphBuilderPhase;
+import org.graalvm.compiler.nodes.FrameState;
 import org.graalvm.compiler.nodes.Invoke;
 import org.graalvm.compiler.nodes.InvokeWithExceptionNode;
 import org.graalvm.compiler.nodes.StructuredGraph;
@@ -130,6 +131,8 @@ public class UnsafeAutomaticSubstitutionProcessor extends SubstitutionProcessor 
 
     private final List<ResolvedJavaType> suppressWarnings;
 
+    private static ResolvedJavaType resolvedUnsafeClass;
+
     private ResolvedJavaMethod unsafeObjectFieldOffsetFieldMethod;
     private ResolvedJavaMethod unsafeObjectFieldOffsetClassStringMethod;
     private ResolvedJavaMethod unsafeArrayBaseOffsetMethod;
@@ -179,6 +182,8 @@ public class UnsafeAutomaticSubstitutionProcessor extends SubstitutionProcessor 
             } catch (ClassNotFoundException cnfe) {
                 throw VMError.shouldNotReachHere(cnfe);
             }
+
+            resolvedUnsafeClass = originalMetaAccess.lookupJavaType(unsafeClass);
 
             Method unsafeObjectFieldOffset = unsafeClass.getMethod("objectFieldOffset", java.lang.reflect.Field.class);
             unsafeObjectFieldOffsetFieldMethod = originalMetaAccess.lookupJavaMethod(unsafeObjectFieldOffset);
@@ -250,9 +255,6 @@ public class UnsafeAutomaticSubstitutionProcessor extends SubstitutionProcessor 
          */
         try {
             suppressWarnings.add(originalMetaAccess.lookupJavaType(Class.forName("sun.security.provider.ByteArrayAccess")));
-            if (JavaVersionUtil.JAVA_SPEC >= 11) {
-                suppressWarnings.add(originalMetaAccess.lookupJavaType(Class.forName("jdk.internal.misc.InnocuousThread")));
-            }
         } catch (ClassNotFoundException e) {
             throw VMError.shouldNotReachHere(e);
         }
@@ -305,6 +307,13 @@ public class UnsafeAutomaticSubstitutionProcessor extends SubstitutionProcessor 
         if (hostType.isArray()) {
             return;
         }
+
+        if (annotationSubstitutions.findSubstitution(hostType).isPresent()) {
+            // Class is substituted, clinit will be eliminated, so bail early.
+            reportSkippedSubstitution(hostType);
+            return;
+        }
+
         /* Detect field offset computation in static initializers. */
         ResolvedJavaMethod clinit = hostType.getClassInitializer();
 
@@ -419,6 +428,11 @@ public class UnsafeAutomaticSubstitutionProcessor extends SubstitutionProcessor 
          */
         ResolvedJavaField offsetField = extractValueStoreField(unsafeObjectFieldOffsetInvoke.asNode(), FieldOffset, unsuccessfulReasons);
 
+        // No field, but doesn't escape local usage, ignore.
+        if (offsetField == null && unsuccessfulReasons.isEmpty()) {
+            return;
+        }
+
         /*
          * If the target field holder and name, and the offset field were found try to register a
          * substitution.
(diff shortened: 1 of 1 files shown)


bobmcwhirter:
@cstancu -- So this does better analysis, hoping to not have to add InnocuousThread to the whitelist, and also preventing warnings on user-provided code that uses the same patterns (JBoss-Threads, for one) for local unsafe offset usage.

dmlloyd:
Any updates on this?

bobmcwhirter:
ping?

bobmcwhirter:
No CI?

bobmcwhirter:
@cstancu Should I just close this and move on?

bobmcwhirter:
ping

plokhotnyuk:
@bobmcwhirter will it help in resolving of contradictions reproduced in [this issue](https://github.com/oracle/graal/issues/1515)?

bobmcwhirter:
I don’t think so. But I’ll look into this separately.

bobmcwhirter:
This PR is new-and-improved and addresses the issue in #1515 also.

bobmcwhirter:
This addresses the case of objectFieldOffset(field) where a try/catch is involved due to Class.getDeclaredField("name") having declared exceptions and being wrapped in a try/catch causing the offset field store itself to generally be stored via a temp variable.

This modification looks through possible Phi nodes to find the ultimate field store if one is not immediately available from the unsafe invoke itself.

cstancu:
@bobmcwhirter, thank you, I'll try to review/integrate this today.

bobmcwhirter:
@cstancu Any feedback on this one?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
```diff
@@ -305,6 +304,12 @@ public void computeSubstitutions(ResolvedJavaType hostType, OptionValues options
         if (hostType.isArray()) {
             return;
         }
+
+        if (annotationSubstitutions.findSubstitution(hostType).isPresent()) {
```
christianwimmer: I'm curious: for which class does this trigger?

bobmcwhirter: A class within JBoss-Threads for one. 

We substitute the entire class, where the original has some static initializer we're scrubbing away and replacing.

cstancu: We already suppress all warnings by default for substituted types; see the end of the `reportUnsuccessfulAutomaticRecomputation()` method. But I would like to at least see a debug message about this decision, i.e., guarded by `Options.UnsafeAutomaticSubstitutionsLogLevel.getValue() >= DEBUG_LEVEL`.

bobmcwhirter: I think ultimately, if I recall, we replaced classed `Foo` entirely, but the old `Foo` ended up with a parsing error due to Unsafe substitutions that were being pointlessly done attempting to parse a class that referenced some unresolvable types or something. I forget the details at this point.

I'm not sure what the log message would be... `Didn't perform Unsafe substitutions on a class Foo that has been wholly replaced`?

Happy to add it, though.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
```diff
@@ -682,6 +702,26 @@ private static ResolvedJavaField extractValueStoreField(ValueNode valueNode, Kin
             if (signExtendFieldStoreUsages.count() == 1) {
                 offsetField = signExtendFieldStoreUsages.first().field();
             }
+        } else {
+            // Determine if the usage of the offset escapes from local usage.
+            for (Node valueNodeUsage : valueNodeUsages) {
+                // Probably other valid node types could be added.
+                if (valueNodeUsage instanceof FrameState || valueNodeUsage instanceof MethodCallTargetNode) {
```
christianwimmer: I don't think `MethodCallTarget` is a valid reason without further qualification. We need to have a whitelist of allowed target methods here too, like `Unsafe` methods that read and write memory using the offset.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
```diff
@@ -682,6 +702,26 @@ private static ResolvedJavaField extractValueStoreField(ValueNode valueNode, Kin
...
+                    continue;
+                }
+                if (valueNodeUsage instanceof ValuePhiNode) {
+                    NodeIterable<StoreFieldNode> phiStores = valueNodeUsage.usages().filter(StoreFieldNode.class);
+                    // Store is on the far side of a Phi node, probably due to a temporary variable
+                    // being used involving a try/catch and a final field which desires single
+                    // assignment.
+                    if (phiStores.count() == 1) {
+                        offsetField = phiStores.first().field();
```
christianwimmer: So here you follow a single (out of possibly many) usages. That is not sound, since one of the other usages can actually be a disallowed usage. Or a store to another field that we are then missing.

Also the phi function has other inputs, i.e., the offset is merged with a different value. This means that the `offsetField` might be initialized with a different (non-field-offset) value in some cases.

In summary, I don't think there is a good way to handle phi nodes, that's why we hadn't added such a handling yet.

bobmcwhirter: Yeah... I was wondering. I'm still learning my way around Phi nodes and such. I'll revert this portion.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
bobmcwhirter: Okie dokie, I can make that a little more constrained, for sure.

cstancu:
@bobmcwhirter I apologize for taking me so much time to get back to this. I pulled in your changes and made some adjustments. They are now being checked through our internal gate. I integrated your analysis of purely local usage of unsafe API which removes the `InnocuousThread` (and classes with similar patterns) from the exception list, but I also tightened the rules of what a legal use of unsafe API is.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
cstancu: Just to reiterate: This case should already be fixed later on by:
https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java#L904
But if the substituted class references some missing types then you would end up with a parsing error as you mentioned. Bailing out early is better. The side effect is that you'll gate a "skipped unsafe substitution" message for every substituted type.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/UnsafeAutomaticSubstitutionProcessor.java
```diff
@@ -419,6 +428,11 @@ private void processUnsafeObjectFieldOffsetInvoke(
          */
         ResolvedJavaField offsetField = extractValueStoreField(unsafeObjectFieldOffsetInvoke.asNode(), FieldOffset, unsuccessfulReasons);
 
+        // No field, but doesn't escape local usage, ignore.
+        if (offsetField == null && unsuccessfulReasons.isEmpty()) {
```
cstancu, line 432: I don't like the idea of encoding the 'does-not-escape' decision in `unsuccessfulReasons.isEmpty()`. That may not always be true. The way the analysis is currently structured first the right-hand-side of an unsafe call is processed, and failure reasons recorded, then the left-hand-side. Therefore, the second part of the analysis that tries to understand where the unsafe value is stored may already have some failure reasons from the right-hand-side, e.g., non-constant arguments to the unsafe call. We need the right-hand-side processed first to extract bits of information needed by the recomputation itself such as target field name and holder class. So, a better solution is to wrap the result in a class: 
``` 
    static final class SearchResult {
        /** The field where the value is stored, if found. */
        final ResolvedJavaField valueStoreField;
        /**
         * Uses that can lead to the unsafe value having side effects that we didn't account for are
         * considered illegal.
         */
        final boolean illegalUseFound;
}
```
In this paradigm the "doesn't escape local usage" is a legal use and therefore ignored.

PR #1414, opened 2019-06-20
Java 11: Add java.io.UnixFileSystem.getNameMax0 implementation

### LINKED ISSUE #1413: Java 11: Unsupported method java.io.UnixFileSystem.getNameMax0(String) is reachable
Under Java 11, attempting to use `File#createTempFile` yields the above error.  PR coming right up.

dmlloyd:
Fixes #1413

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix.jdk11/src/com/oracle/svm/core/posix/Target_java_io_UnixFileSystem.java
+++ b/substratevm/src/com.oracle.svm.core.posix.jdk11/src/com/oracle/svm/core/posix/Target_java_io_UnixFileSystem.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2019, 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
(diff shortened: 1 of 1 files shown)


dmlloyd:
@olpaw, you renamed the class from `Target_java_io_UnixFileSystem` to `Target_java_io_UnixFileSystem_JDK11OrLater`, but you didn't rename the file to `Target_java_io_UnixFileSystem_JDK11OrLater.java`.  Was that intentional?

dmlloyd:
Also: should we expect that the other classes in the `posix.jdk11` subpackage will be similarly renamed?  In other words, should we always add this suffix to class names in that package?  It seems rather redundant, shadowing notwithstanding.

dmlloyd:
Also, also: should these classes maybe be put into an enclosing `PosixJava*Substitutions` class in the same manner as the jdk8 ones?

bobmcwhirter:
Please no. Class per file is my opinion if it matters.

dmlloyd:
Also^3: `com.oracle.svm.core.posix.Target_java_io_FileCleanable` does not have a platform annotation on it; is that wrong?

olpaw:
> Also^3: com.oracle.svm.core.posix.Target_java_io_FileCleanable does not have a platform annotation on it; is that wrong?

Yup. That's a latent bug. When we move to static JDK JNI-libs (like we already do for Windows) for Java 11 this will need to get fixed. Thanks for the hint.

olpaw:
> Was that intentional?

No. That was sloppy ;-)

olpaw:
> posix.jdk11 subpackage will be similarly renamed?

Obviously you only need to do this if they conflict with the classes already provided by the `posix` package.

dmlloyd:
> > posix.jdk11 subpackage will be similarly renamed?
> 
> Obviously you only need to do this if they conflict with the classes already provided by the `posix` package.

Well that's the question: being in a subpackage, no conflict is possible.  But, it *is* definitely possible that there could be, say, a style policy that dictates the classes shouldn't have the same *simple* name even if the qualified name is distinct.  That's more what I was asking about.

olpaw:
This is not how `mx` project overlays work. `src/com.oracle.svm.core.posix.jdk11/src/com/oracle/svm/core/posix/Target_java_io_UnixFileSystem.java` uses
```
package com.oracle.svm.core.posix;
```
same as all the classes in `src/com.oracle.svm.core.posix`. When you are on a JDK11 platform you get the classes defined in src/com.oracle.svm.core.posix.jdk11 in addition and usually they use the same package name as the base project (so you have to be careful with the qualified class names). @dougxc do we have documentation how `mx` project overlays should be used? Something like https://openjdk.java.net/jeps/238 but in the scope of defining the overlays in `mx`?

dougxc:
We don't have anything beyond https://github.com/graalvm/mx#versioning-sources-for-different-jdk-releases and the numerous usages in Graal as examples.

dmlloyd:
> same as all the classes in `src/com.oracle.svm.core.posix`. When you are on a JDK11 platform you get the classes defined in src/com.oracle.svm.core.posix.jdk11 in addition and usually they use the same package name as the base project (so you have to be careful with the qualified class names).

Ah, of course, I see. Thanks for the explanation.

PR #1418, opened 2019-06-21
Fix for #1417 - Invalid Class-Path entries.

### LINKED ISSUE #1417: Invalid classpath entry in MANIFEST.MF
Hey,

A user of Quarkus found a situation where native-image trips
https://github.com/quarkusio/quarkus/issues/2852
With a reproducer here https://github.com/vietk/QuarkusReproducer
It's seems related to the fact that maven adds to the MANIFEST.MF cp a non jar file

HotSpot is fine with it but native-image fails. Do you see it as a bug or a behavioral difference?

We did workaround it in Quarkus by excluding such references but that's more a maven + native-image disagreement at heart.

bobmcwhirter:
Fixes #1417. Be more defensive for MANIFEST.MF Class-Path entries
which sometimes includes non-jar/zip entries, because
Maven is goofy at times.

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -804,11 +804,17 @@ public class NativeImage {
                 }
 
                 for (Path jarFile : jarFileMatches) {
+                    String fileName = jarFile.getFileName().toString().toLowerCase();
+                    if (!(fileName.endsWith(".jar") || fileName.endsWith(".zip"))) {
+                        // not a .jar or .zip
+                        continue;
+                    }
                     URI jarFileURI = URI.create("jar:" + jarFile.toUri());
                     try (FileSystem jarFS = FileSystems.newFileSystem(jarFileURI, Collections.emptyMap())) {
                         Path nativeImageMetaInfBase = jarFS.getPath("/" + nativeImageMetaInf);
                         processNativeImageMetaInf(jarFile, nativeImageMetaInfBase, metaInfProcessor);
                     }
+
                 }
             }
         } catch (IOException | FileSystemNotFoundException e) {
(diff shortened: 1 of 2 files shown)


cstancu:
@bobmcwhirter I added `Fixes #1417.` in the description. This tells GitHub to automatically close the corresponding issue when this gets merged.

bobmcwhirter:
Thanks, I'll be sure to do that in the future.

olpaw:
This problem is already fixed on master. See 086fa120ea2dea40ce4607589a29ca388b0c862f (committed as part of 9c0a77f7307d429ee82c4c6f1f9ae023a9004d58)

Your PR would further open the gap between `native-image -cp` vs. `java -cp` behavior because `java -cp` does accept jar files no matter what file extension is used and so should we:
```
graalvm-ce-19.0.2> bin/java -cp xyz.jar Hello
Hello World! <--- So far so good ...
graalvm-ce-19.0.2> cp xyz.jar xyz
graalvm-ce-19.0.2> bin/java -cp xyz Hello
Hello World! <--- Ohh .... this also works!
```
Commit 086fa120ea2dea40ce4607589a29ca388b0c862f takes care that. As yours it will silently ignore invalid cp entries but it will also accept any jar file no matter its extension. BTW, this change is also on the list for getting backported. So this will land in a GraalVM 19 update release.

bobmcwhirter:
Though, there are other areas my PR addresses in terms of invalid jars. Not just the classloader, but when generation goes scrubbing around for MANIFEST.MF to look for its contents. I'm okay with your fix, but it should probably be applied further, in similar locations as mine?  Basically anywhere we attempt a Jar Filesystem creation.

olpaw:
Hmm not sure what you mean. The only other place is https://github.com/oracle/graal/blob/c12f7791882bc393f7fa4446fe2e8308748bf5dd/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java#L808 and there we already have a try-with-resources that ensures that files that are not a valid jar-file get silently ignored. Do you have something else in mind?

bobmcwhirter:
The problem is the type of caught exception, I think. I'm ending up with a `ProviderNotFoundException` when a maven XML file escapes through to there.

olpaw:
Oh, this is not good. I will change that so that it follows the same pattern used in 086fa120ea2dea40ce4607589a29ca388b0c862f. Thanks for being insistent :-)

PR #1424, opened 2019-06-24
Multithreading support for the LLVM backend

loicottet:
This PR allows multithreading for the LLVM backend by implementing support for the fixed register values of Native Image (thread pointer and heap base pointer) by passing them as arguments to functions.

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMIRBuilder.java
+++ b/compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMIRBuilder.java
@@ -77,7 +77,7 @@ public class LLVMIRBuilder {
              * doesn't recognize an address space cast (see pointerType()) as declaring a new
              * reference, but it does a function return value.
              */
-            gcRegisterFunction = addFunction("__svm_gc_register", functionType(objectType(), rawPointerType()));
+            gcRegisterFunction = addFunction(LLVMUtils.GC_REGISTER_FUNCTION_NAME, functionType(objectType(), rawPointerType()));
             LLVM.LLVMSetLinkage(gcRegisterFunction, LLVM.LLVMLinkOnceAnyLinkage);
             setAttribute(gcRegisterFunction, LLVM.LLVMAttributeFunctionIndex, "alwaysinline");
             setAttribute(gcRegisterFunction, LLVM.LLVMAttributeFunctionIndex, "gc-leaf-function");
@@ -100,7 +100,7 @@ public class LLVMIRBuilder {
 
     /* Functions */
 
-    LLVMValueRef addFunction(String name, LLVMTypeRef type) {
+    public LLVMValueRef addFunction(String name, LLVMTypeRef type) {
         return LLVM.LLVMAddFunction(module, name, type);
     }
 
@@ -119,7 +119,7 @@ public class LLVMIRBuilder {
         LLVM.LLVMSetLinkage(global, linkage);
     }
 
-    void setAttribute(LLVMValueRef func, long index, String attribute) {
+    public void setAttribute(LLVMValueRef func, long index, String attribute) {
         int kind = LLVM.LLVMGetEnumAttributeKindForName(attribute, attribute.length());
         LLVMAttributeRef attr;
         if (kind != 0) {
@@ -151,7 +151,7 @@ public class LLVMIRBuilder {
         setPersonalityFunction(getMainFunction(), personality);
     }
 
-    LLVMBasicBlockRef appendBasicBlock(String name, LLVMValueRef func) {
+    public LLVMBasicBlockRef appendBasicBlock(String name, LLVMValueRef func) {
         return LLVM.LLVMAppendBasicBlockInContext(context, func, name);
     }
 
@@ -212,7 +212,7 @@ public class LLVMIRBuilder {
         }
     }
 
-    static LLVMTypeRef typeOf(LLVMValueRef value) {
+    public static LLVMTypeRef typeOf(LLVMValueRef value) {
         return LLVM.LLVMTypeOf(value);
     }
 
@@ -290,8 +290,28 @@ public class LLVMIRBuilder {
         return LLVM.LLVMFunctionType(returnType, new PointerPointer<>(argTypes), argTypes.length, varargs ? TRUE : FALSE);
     }
 
-    private static LLVMTypeRef getReturnType(LLVMTypeRef functionType) {
+    public static LLVMTypeRef getReturnType(LLVMTypeRef functionType) {
         return LLVM.LLVMGetReturnType(functionType);
+    }
+
+    private static boolean isFunctionVarArg(LLVMTypeRef functionType) {
+        return LLVM.LLVMIsFunctionVarArg(functionType) == TRUE;
+    }
+
+    /**
+     * Creates a new function type based on the given one with the given argument types prepended to
+     * the original ones.
+     */
+    public LLVMTypeRef prependArguments(LLVMTypeRef functionType, LLVMTypeRef... typesToAdd) {
+        LLVMTypeRef returnType = getReturnType(functionType);
+        boolean varargs = isFunctionVarArg(functionType);
+        LLVMTypeRef[] oldTypes = getParamTypes(functionType);
+
+        LLVMTypeRef[] newTypes = new LLVMTypeRef[oldTypes.length + typesToAdd.length];
+        System.arraycopy(typesToAdd, 0, newTypes, 0, typesToAdd.length);
+        System.arraycopy(oldTypes, 0, newTypes, typesToAdd.length, oldTypes.length);
+
+        return functionType(returnType, varargs, newTypes);
     }
 
     private static LLVMTypeRef[] getParamTypes(LLVMTypeRef functionType) {
@@ -341,7 +361,7 @@ public class LLVMIRBuilder {
         return LLVM.LLVMGetPointerAddressSpace(pointerType) == LLVMUtils.TRACKED_POINTER_ADDRESS_SPACE;
     }
 
-    private static LLVMTypeRef getElementType(LLVMTypeRef pointerType) {
+    public static LLVMTypeRef getElementType(LLVMTypeRef pointerType) {
         return LLVM.LLVMGetElementType(pointerType);
     }
 
@@ -353,7 +373,7 @@ public class LLVMIRBuilder {
         return isPointer(type) && !isTracked(type);
     }
 
-    static boolean isVoidType(LLVMTypeRef type) {
+    public static boolean isVoidType(LLVMTypeRef type) {
         return LLVM.LLVMGetTypeKind(type) == LLVM.LLVMVoidTypeKind;
     }
 
@@ -372,7 +392,7 @@ public class LLVMIRBuilder {
         return true;
     }
 
-    private static String intrinsicType(LLVMTypeRef type) {
+    public static String intrinsicType(LLVMTypeRef type) {
         switch (LLVM.LLVMGetTypeKind(type)) {
             case LLVM.LLVMIntegerTypeKind:
                 return "i" + integerTypeWidth(type);
@@ -430,7 +450,7 @@ public class LLVMIRBuilder {
         return LLVM.LLVMConstReal(doubleType(), x);
     }
 
-    LLVMValueRef constantNull(LLVMTypeRef type) {
+    public LLVMValueRef constantNull(LLVMTypeRef type) {
         return LLVM.LLVMConstNull(type);
     }
 
@@ -458,11 +478,11 @@ public class LLVMIRBuilder {
         return func;
     }
 
-    private static LLVMValueRef getParam(LLVMValueRef func, int index) {
+    public static LLVMValueRef getParam(LLVMValueRef func, int index) {
         return LLVM.LLVMGetParam(func, index);
     }
 
-    LLVMValueRef getParam(int index) {
+    public LLVMValueRef getParam(int index) {
         return getParam(function, index);
     }
 
@@ -520,13 +540,13 @@ public class LLVMIRBuilder {
         LLVM.LLVMSetInitializer(global, value);
     }
 
-    LLVMValueRef register(String name) {
+    public LLVMValueRef register(String name) {
         String nameEncoding = name + "\00";
         LLVMValueRef[] vals = new LLVMValueRef[]{LLVM.LLVMMDStringInContext(context, nameEncoding, nameEncoding.length())};
         return LLVM.LLVMMDNodeInContext(context, new PointerPointer<>(vals), vals.length);
     }
 
-    LLVMValueRef buildReadRegister(LLVMValueRef register) {
+    public LLVMValueRef buildReadRegister(LLVMValueRef register) {
         LLVMTypeRef readRegisterType = functionType(longType(), metadataType());
         return buildIntrinsicCall("llvm.read_register.i64", readRegisterType, register);
     }
@@ -550,7 +570,7 @@ public class LLVMIRBuilder {
     /* Control flow */
     public static final AtomicLong nextPatchpointId = new AtomicLong(0);
 
-    LLVMValueRef buildCall(LLVMValueRef callee, LLVMValueRef... args) {
+    public LLVMValueRef buildCall(LLVMValueRef callee, LLVMValueRef... args) {
         return LLVM.LLVMBuildCall(builder, callee, new PointerPointer<>(args), args.length, DEFAULT_INSTR_NAME);
     }
 
@@ -630,11 +650,11 @@ public class LLVMIRBuilder {
         return buildCall(intrinsic, args);
     }
 
-    void buildRetVoid() {
+    public void buildRetVoid() {
         LLVM.LLVMBuildRetVoid(builder);
     }
 
-    void buildRet(LLVMValueRef value) {
+    public void buildRet(LLVMValueRef value) {
         LLVM.LLVMBuildRet(builder, value);
     }
 
@@ -703,7 +723,7 @@ public class LLVMIRBuilder {
 
     /* Comparisons */
 
-    LLVMValueRef buildIsNull(LLVMValueRef value) {
+    public LLVMValueRef buildIsNull(LLVMValueRef value) {
         return LLVM.LLVMBuildIsNull(builder, value, DEFAULT_INSTR_NAME);
     }
 
@@ -980,7 +1000,7 @@ public class LLVMIRBuilder {
         return LLVM.LLVMBuildGEP(builder, base, new PointerPointer<>(indices), indices.length, DEFAULT_INSTR_NAME);
     }
 
-    LLVMValueRef buildLoad(LLVMValueRef address, LLVMTypeRef type) {
+    public LLVMValueRef buildLoad(LLVMValueRef address, LLVMTypeRef type) {
         LLVMTypeRef addressType = LLVM.LLVMTypeOf(address);
         LLVMTypeRef pointedType = type;
         boolean postRegister = false;
@@ -1012,7 +1032,7 @@ public class LLVMIRBuilder {
         LLVM.LLVMBuildStore(builder, castedValue, castedAddress);
     }
 
-    LLVMValueRef buildArrayAlloca(int slots) {
+    public LLVMValueRef buildArrayAlloca(int slots) {
         return LLVM.LLVMBuildArrayAlloca(builder, rawPointerType(), constantInt(slots), DEFAULT_INSTR_NAME);
     }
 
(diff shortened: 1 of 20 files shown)


substratevm/src/com.oracle.svm.core.graal/src/com/oracle/svm/core/graal/snippets/CEntryPointSnippets.java
```diff
@@ -351,10 +351,10 @@ private static int enterIsolateMT(Isolate isolate) {
     public static int enterSnippet(IsolateThread thread) {
         Isolate isolate;
         if (MultiThreaded.getValue()) {
+            writeCurrentVMThread(thread);
```
vjovanov: If someone moves this around they will break the LLVM stuff without knowing why and waste a ton of time.

The problem here is that `r15` reads in the function are floating so they get hoisted to the point before the thread is written. 

I see the following solutions: 
 1) Verify that there are no reads scheduled before non-null writes and reads after a null write in every function.
 2) In functions that write to `r15` all the reads simply become fixed:
       a) By annotating the functions that write to `r15` and simply make all register reads fixed.
       b) By making a phase that handles that.
       c) By making fixed versions of functions that access `r15` and only call them in the code. 

Given how rare is this I would just go with 1) and put a comment in the code. 

@peter-hofer WDYT?

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMIRBuilder.java
```diff
@@ -151,6 +151,36 @@ public void setPersonalityFunction(LLVMValueRef personality) {
         setPersonalityFunction(getMainFunction(), personality);
     }
 
+    public LLVMValueRef createJNIWrapper(LLVMValueRef callee, long statepointId, int numArgs, int anchorIPOffset, LLVMBasicBlockRef currentBlock) {
```
vjovanov: This function deserves a comment.

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMIRBuilder.java
```diff
@@ -294,6 +324,22 @@ private static LLVMTypeRef getReturnType(LLVMTypeRef functionType) {
         return LLVM.LLVMGetReturnType(functionType);
     }
 
+    private static boolean isFunctionVarArg(LLVMTypeRef functionType) {
+        return LLVM.LLVMIsFunctionVarArg(functionType) == TRUE;
+    }
+
+    public LLVMTypeRef prependArguments(LLVMTypeRef functionType, LLVMTypeRef... typesToAdd) {
```
vjovanov: This one as well.

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMUtils.java
```diff
@@ -61,6 +62,16 @@
         public static final int NODE = 3;
     }
 
+    public abstract static class TargetSpecific {
```
vjovanov: Can you put documentation comments here? This will be the API for future architectures.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMFeature.java
```diff
@@ -142,3 +155,27 @@ private static void checkLLVMVersion() {
         }
     }
 }
+
+@AutomaticFeature
+@Platforms(Platform.AMD64.class)
+class LLVMAMD64TargetSpecificFeature implements Feature {
```
vjovanov: Can we add AArch64 as well?

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMNativeImageCodeCache.java
```diff
@@ -196,10 +196,11 @@ private LLVMStackMapInfo readStackMapSection(LLVMSectionIteratorRef sectionItera
 
     private void storeMethodOffsets(long codeSize) throws IOException {
         List<Integer> sortedMethodOffsets = textSymbolOffsets.values().stream().distinct().sorted().collect(Collectors.toList());
-        Integer gcRegisterOffset = textSymbolOffsets.get(SYMBOL_PREFIX + "__svm_gc_register");
-        if (gcRegisterOffset != null) {
-            sortedMethodOffsets.remove(gcRegisterOffset);
-        }
+        textSymbolOffsets.forEach((symbol, offset) -> {
```
vjovanov: A comment would be useful here.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMNativeImageCodeCache.java
```diff
@@ -196,10 +196,11 @@ private LLVMStackMapInfo readStackMapSection(LLVMSectionIteratorRef sectionItera
...
+            if (symbol.equals("__svm_gc_register") || symbol.startsWith(SYMBOL_PREFIX + "__svm_jni_wrapper_")) {
```
vjovanov: These should become constants as they are used in multiple cases.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMGenerator.java
```diff
@@ -24,38 +24,100 @@
...
+    @Override
+    public SubstrateRegisterConfig getRegisterConfig() {
+        return (SubstrateRegisterConfig) super.getRegisterConfig();
+    }
+
+    @Override
+    public void allocateRegisterSlots() {
+        if (!isEntryPoint) {
+            /*
+             * Non-entry point methods get the address to the special register in their argument
```
vjovanov: It would be nice to fit this on line.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMGenerator.java
```diff
@@ -89,4 +151,103 @@ public String getFunctionName(ResolvedJavaMethod method) {
...
+                return new LLVMUtils.LLVMVariable(getSpecialRegister(LLVMFeature.THREAD_POINTER_INDEX));
+            } else if (reg.getRegister().equals(getRegisterConfig().getHeapBaseRegister())) {
+                return new LLVMUtils.LLVMVariable(getSpecialRegister(LLVMFeature.HEAP_BASE_INDEX));
+            }
+        }
+        return super.emitMove(input);
+    }
+
+    @Override
+    public void emitMove(AllocatableValue dst, Value src) {
```
vjovanov: Wouldn't it be nicer if we could do this on a higher level? Here it looks like a fragile hack.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateNodeLLVMBuilder.java
```diff
@@ -59,6 +81,75 @@ public void emitCGlobalDataLoadAddress(CGlobalDataLoadAddressNode node) {
...
+        return new LLVMVariable(returnAddress);
+    }
+
+    private SubstrateLLVMGenerator getGenerator() {
+        return (SubstrateLLVMGenerator) gen;
+    }
+
+    @Override
+    protected int getParamIndex(int index) {
+        return (getGenerator().isEntryPoint() ? 0 : LLVMFeature.SPECIAL_REGISTER_COUNT) + super.getParamIndex(index);
```
vjovanov: Is the IsolateThread always at position 0? Also, what happens with entry points that accept the `Isolate` instead of the `IsolateThread`.

loicottet: This is just to offset the parameters if there are special registers passed as parameters as well. The function is maybe not too clear, `(getGenerator().isEntryPoint() ? 0 : LLVMFeature.SPECIAL_REGISTER_COUNT)` is just computing an offset to be added to the actual parameter number, I'll put this in an intermediary variable `offset` to clarify

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMUtils.java
loicottet: I was thinking about putting this in when we actually support ARM, instead of having bits of ARM support hanging around which might make it look like it works with ARM when it doesn't

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMGenerator.java
loicottet: done

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMFeature.java
loicottet: I was thinking about putting this in when we actually support ARM, instead of having bits of ARM support hanging around which might make it look like it works with ARM when it doesn't

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMNativeImageCodeCache.java
loicottet: done

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMUtils.java
loicottet: done

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMIRBuilder.java
loicottet: done

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMIRBuilder.java
loicottet: done

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMGenerator.java
loicottet: Done. I added emitReadRegister and emitWriteRegister to LIRGeneratorTool, that's way nicer that way indeed

substratevm/src/com.oracle.svm.core.graal/src/com/oracle/svm/core/graal/snippets/CEntryPointSnippets.java
vjovanov: @loicottet could you send the graph for this problem. /cc @christianwimmer

peter-hofer: `r15` is only ever written in these snippets, so I wouldn't put too much effort into a sophisticated solution. Are these reads hoisted in Graal or in LLVM? Could there be some sort of barrier (using the term loosely) that we can insert here or during lowering that prevents reording the reads before here?

loicottet: > @loicottet could you send the graph for this problem. /cc @christianwimmer


The problem is node 28 gets emitted before node 27.

loicottet: I have trouble determining when exactly I should emit a FixedNode instead of a FloatingNode. The node accessing the thread pointer is created when generating `VMThreads$StatusSupport.compareAndSetNativeToJava()`, so the most I can do is limit the change to uninterruptible methods. Limiting it to this exact function feels like a gross hack, is there another way I can find out whether a node is being emitted in an entry point?

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMNativeImageCodeCache.java
vjovanov: This is also done.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMFeature.java
vjovanov: OK.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/thread/VMThreadMTFeature.java
```diff
@@ -190,7 +192,16 @@ private static ValueNode currentThread(GraphBuilderContext b, boolean usedForAdd
...
         boolean isDeoptTarget = b.getMethod() instanceof SharedMethod && ((SharedMethod) b.getMethod()).isDeoptTarget();
-        if (isDeoptTarget || usedForAddress) {
+
+        /*
+         * Due to the fact that the LLVM backend handles reading the thread pointer in entry point
+         * methods as a stack slot load instead of a direct register access, a ReadRegisterFixedNode
+         * should be emitted in those cases so that the register read doesn't get hoisted above
+         * where the thread pointer gets stored in the stack slot.
+         */
+        boolean isLLVM = SubstrateOptions.CompilerBackend.getValue().equals("llvm");
```
vjovanov: I would just remove `isLLVM` here. It does not hurt to fix those registers on both backends.

loicottet: The emission of floating nodes was explicitly designed for efficiency, but I guess this isn't too frequent.

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMIRBuilder.java
```diff
@@ -77,7 +77,7 @@ protected LLVMIRBuilder(String functionName, LLVMContextRef context, boolean tra
              * doesn't recognize an address space cast (see pointerType()) as declaring a new
              * reference, but it does a function return value.
              */
-            gcRegisterFunction = addFunction("__svm_gc_register", functionType(objectType(), rawPointerType()));
+            gcRegisterFunction = addFunction(LLVMUtils.GC_REGISTER_FUNCTION_NAME, functionType(objectType(), rawPointerType()));
             LLVM.LLVMSetLinkage(gcRegisterFunction, LLVM.LLVMLinkOnceAnyLinkage);
             setAttribute(gcRegisterFunction, LLVM.LLVMAttributeFunctionIndex, "alwaysinline");
             setAttribute(gcRegisterFunction, LLVM.LLVMAttributeFunctionIndex, "gc-leaf-function");
```
peter-hofer: Can we use constant strings for these? I've seen `gc-leaf-function` a few times during this review.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateNodeLLVMBuilder.java
```diff
@@ -59,6 +85,102 @@ public void emitCGlobalDataLoadAddress(CGlobalDataLoadAddressNode node) {
...
+            threadData = gen.getBuilder().buildIntToPtr(threadData, gen.getBuilder().rawPointerType());
+            LLVMValueRef safepointCounterAddr = gen.getBuilder().buildGEP(threadData, gen.getBuilder().constantInt(Math.toIntExact(Safepoint.getThreadLocalSafepointRequestedOffset())));
+            LLVMValueRef safepointCount = gen.getBuilder().buildAtomicSub(safepointCounterAddr, gen.getBuilder().constantInt(1));
+            return gen.getBuilder().buildIsNull(safepointCount);
+        }
+        return super.emitCondition(condition);
+    }
+
+    @Override
+    protected LLVMValueRef emitCall(Invoke i, LoweredCallTargetNode callTarget, LLVMValueRef callee, long patchpointId, LLVMValueRef... args) {
```
peter-hofer: Perhaps I'm nitpicking, but I don't like single-letter names for parameters, in particular `i` which is really common as loop index.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateNodeLLVMBuilder.java
```diff
@@ -59,6 +85,102 @@ public void emitCGlobalDataLoadAddress(CGlobalDataLoadAddressNode node) {
...
+        LLVMValueRef[] args = super.getCallArguments(arguments, callType, targetMethod);
+        return gen.getCallArguments(args, callType, targetMethod);
+    }
+
+    protected LLVMValueRef emitCondition(LogicNode condition) {
+        if (condition instanceof SafepointCheckNode) {
+            LLVMValueRef threadData = getGenerator().getSpecialRegister(LLVMFeature.THREAD_POINTER_INDEX);
+            threadData = gen.getBuilder().buildIntToPtr(threadData, gen.getBuilder().rawPointerType());
+            LLVMValueRef safepointCounterAddr = gen.getBuilder().buildGEP(threadData, gen.getBuilder().constantInt(Math.toIntExact(Safepoint.getThreadLocalSafepointRequestedOffset())));
+            LLVMValueRef safepointCount = gen.getBuilder().buildAtomicSub(safepointCounterAddr, gen.getBuilder().constantInt(1));
```
peter-hofer: Since this applies (weak) atomicity guarantees to safepoint checks, can this have any drawbacks in the generated code? I suppose on x86 this could still be emitted as `dec` operation.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMGenerator.java
```diff
@@ -89,4 +149,102 @@ public String getFunctionName(ResolvedJavaMethod method) {
...
+        return specialRegister;
+    }
+
+    private LLVMValueRef getSpecialRegisterArgument(int index, ResolvedJavaMethod targetMethod) {
+        LLVMValueRef specialRegisterArg;
+        if (canModifySpecialRegisters(targetMethod)) {
+            assert (isEntryPoint || canModifySpecialRegisters);
+            specialRegisterArg = builder.buildBitcast(getSpecialRegisterPointer(index), builder.rawPointerType());
+        } else {
+            if (isEntryPoint || canModifySpecialRegisters) {
```
peter-hofer: (no point in nesting this `if`)

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMIRBuilder.java
```diff
@@ -1038,13 +1058,35 @@ void buildFence() {
...
-        LLVMValueRef castedAddress = buildBitcast(address, pointerType(expectedType, isObject(typeOf(address))));
-        return LLVM.LLVMBuildAtomicCmpXchg(builder, castedAddress, expectedValue, newValue, LLVM.LLVMAtomicOrderingSequentiallyConsistent, LLVM.LLVMAtomicOrderingSequentiallyConsistent, FALSE);
+        boolean isObject = isObject(expectedType);
+        LLVMTypeRef operationType = isObject ? rawPointerType() : expectedType;
+        LLVMValueRef castedAddress = isObject && isTracked(typeOf(address)) ? buildAddrSpaceCast(address, pointerType(operationType, false))
+                        : buildBitcast(address, pointerType(operationType, isTracked(typeOf(address))));
+        LLVMValueRef castedExpectedValue = isObject ? buildAddrSpaceCast(expectedValue, operationType) : expectedValue;
+        LLVMValueRef castedNewValue = isObject ? buildAddrSpaceCast(newValue, operationType) : newValue;
+
+        LLVMValueRef cas = LLVM.LLVMBuildAtomicCmpXchg(builder, castedAddress, castedExpectedValue, castedNewValue, LLVM.LLVMAtomicOrderingMonotonic, LLVM.LLVMAtomicOrderingMonotonic, FALSE);
```
peter-hofer: This change weakens the atomicity guarantee level, was the old level too strict?

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateNodeLLVMBuilder.java
loicottet: True, `invoke` would be more appropriante here

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateNodeLLVMBuilder.java
loicottet: This is indeed producing suboptimal code (the code generator is not smart enough to emit a `dec` instruction here. That should be a simple call to `buildSub` instead.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMGenerator.java
loicottet: True

substratevm/src/com.oracle.svm.core.graal/src/com/oracle/svm/core/graal/snippets/CEntryPointSnippets.java
vjovanov: This is fixed, right?

compiler/src/org.graalvm.compiler.core.llvm/src/org/graalvm/compiler/core/llvm/LLVMIRBuilder.java
vjovanov: Fixed.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateNodeLLVMBuilder.java
vjovanov: Fixed.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateNodeLLVMBuilder.java
vjovanov: This can be fixed in the subsequent PRs.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMGenerator.java
vjovanov: Fixed.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateNodeLLVMBuilder.java
vjovanov: @loicottet can you handle this?

loicottet: Yup, I’ll do it when I’m back

loicottet:
Closed by [bb19ae](https://github.com/oracle/graal/commit/bb19ae25a907411963fd228e4e47becd7d79f793)

Issue #1439, opened 2019-06-27
HotSpot assertion failed in Graal AArch64 compiler test suite


pli-arm:
We found a HotSpot assertion failure in AArch64 Graal compiler test suite. It can be reproduced with latest GitHub Graal working on JDK 14 by below command
```
mx unittest --suite compiler --fail-fast --verbose
```

Error message
```
# A fatal error has been detected by the Java Runtime Environment:
#
#  Internal Error (/home/penli01/jdk_src/src/hotspot/cpu/aarch64/nativeInst_aarch64.cpp:472), pid=141415, tid=141428
#  assert(nativeInstruction_at(verified_entry)->is_jump_or_nop() || nativeInstruction_at(verified_entry)->is_sigill_zombie_not_entrant()) failed: Aarch64 cannot replace non-jump with jump
#
# JRE version: OpenJDK Runtime Environment (14.0) (fastdebug build 14-internal+0-adhoc.penli01.jdksrc)
# Java VM: OpenJDK 64-Bit Server VM (fastdebug 14-internal+0-adhoc.penli01.jdksrc, mixed mode, tiered, jvmci, compressed oops, g1 gc, linux-aarch64)
# Problematic frame:
# V  [libjvm.so+0x113cedc]  NativeJump::patch_verified_entry(unsigned char*, unsigned char*, unsigned char*)+0x174
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport %p %s %c %d %P" (or dumping to /home/penli01/graal/compiler/core.141415)
#
# An error report file with more information is saved as:
# /home/penli01/graal/compiler/hs_err_pid141415.log
Compiled method (JVMCI) 1218434 72428       4      (method)
 total in heap  [0x0000ffff7a615310,0x0000ffff7a6154e8] = 472
 main code      [0x0000ffff7a615480,0x0000ffff7a615494] = 20
 stub code      [0x0000ffff7a615494,0x0000ffff7a615498] = 4
 oops           [0x0000ffff7a615498,0x0000ffff7a6154a8] = 16
 scopes data    [0x0000ffff7a6154a8,0x0000ffff7a6154b0] = 8
 scopes pcs     [0x0000ffff7a6154b0,0x0000ffff7a6154d0] = 32
 dependencies   [0x0000ffff7a6154d0,0x0000ffff7a6154d8] = 8
 JVMCI data     [0x0000ffff7a6154d8,0x0000ffff7a6154e8] = 16
```

[...]

PR #1442, opened 2019-06-27
Fix CodeGenTest issue for AArch64

pli-arm:
This fixes issue #1439

For Graal CodeGenTest cases, assembly code generated by assembler or macro assembler is installed into HotSpot as an nmethod. But AArch64 requires the instruction at nmethod verifed entry to be a jump or nop for method invalidation in sweeper thread. To meet the requirement, we patch the generated test code by inserting a nop at the beginning.

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot.aarch64/src/org/graalvm/compiler/hotspot/aarch64/AArch64HotSpotBackend.java
+++ b/compiler/src/org.graalvm.compiler.hotspot.aarch64/src/org/graalvm/compiler/hotspot/aarch64/AArch64HotSpotBackend.java
@@ -49,6 +49,7 @@ import org.graalvm.compiler.core.common.LIRKind;
 import org.graalvm.compiler.core.common.alloc.RegisterAllocationConfig;
 import org.graalvm.compiler.core.common.spi.ForeignCallLinkage;
 import org.graalvm.compiler.core.gen.LIRGenerationProvider;
+import org.graalvm.compiler.debug.DebugContext;
 import org.graalvm.compiler.hotspot.GraalHotSpotVMConfig;
 import org.graalvm.compiler.hotspot.HotSpotDataBuilder;
 import org.graalvm.compiler.hotspot.HotSpotGraalRuntimeProvider;
@@ -72,18 +73,24 @@ import org.graalvm.compiler.lir.gen.LIRGenerationResult;
 import org.graalvm.compiler.lir.gen.LIRGeneratorTool;
 import org.graalvm.compiler.nodes.StructuredGraph;
 import org.graalvm.compiler.nodes.spi.NodeLIRBuilderTool;
+import org.graalvm.compiler.serviceprovider.GraalUnsafeAccess;
 
 import jdk.vm.ci.aarch64.AArch64Kind;
 import jdk.vm.ci.code.CallingConvention;
+import jdk.vm.ci.code.CompilationRequest;
+import jdk.vm.ci.code.InstalledCode;
 import jdk.vm.ci.code.Register;
 import jdk.vm.ci.code.RegisterConfig;
 import jdk.vm.ci.code.StackSlot;
+import jdk.vm.ci.code.site.Mark;
 import jdk.vm.ci.hotspot.HotSpotCallingConventionType;
 import jdk.vm.ci.hotspot.HotSpotSentinelConstant;
 import jdk.vm.ci.hotspot.aarch64.AArch64HotSpotRegisterConfig;
 import jdk.vm.ci.meta.JavaKind;
 import jdk.vm.ci.meta.JavaType;
 import jdk.vm.ci.meta.ResolvedJavaMethod;
+
+import sun.misc.Unsafe;
 
 /**
  * HotSpot AArch64 specific backend.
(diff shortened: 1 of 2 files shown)


compiler/src/org.graalvm.compiler.asm.test/src/org/graalvm/compiler/asm/test/AssemblerTest.java
```diff
@@ -28,6 +28,7 @@
 
 import java.lang.reflect.Method;
 
+import jdk.vm.ci.aarch64.AArch64;
```
dougxc: We don't want to make this class depend on AArch64 or HotSpot specific semantics. I suggest moving the `AArch64BitCountAssemblerTest` class to be a sibling of `AArch64UncompressPointerTest`.

adinn:
HI Pengfei,

Is this the correct place to be doing this? If an AArch64 code buffer needs to start with a nop then perhaps the AArch64 tests that generate the code buffers ought to insert a nop? Would it not be better to 'fix' the tests rather than 'fix' class AssemblerTest?

compiler/src/org.graalvm.compiler.asm.test/src/org/graalvm/compiler/asm/test/AssemblerTest.java
dougxc: You should then be able to use `org.graalvm.compiler.hotspot.aarch64.AArch64HotSpotBackend.emitInvalidatePlaceholder(CompilationResultBuilder, AArch64MacroAssembler)` to avoid duplicating this `nop` inserting logic.

pli-arm:
> Is this the correct place to be doing this? If an AArch64 code buffer needs to start with a nop then perhaps the AArch64 tests that generate the code buffers ought to insert a nop? Would it not be better to 'fix' the tests rather than 'fix' class AssemblerTest?

I agree that adding a single line of `masm.nop()` in `AArch64BitCountAssemblerTest` is a simple and straightforward fix. But I chose to do this tricky thing in `AssemblerTest` in my patch because I'd like to make the assertion (AArch64 code buffer needs to start with a nop) transparent to unittest writers. My concern is that this crash may appear again when someone else committed new CodeGenTest cases without generating a nop() at the beginning. But I'll update this PR if you still prefer fixing the test case itself.

compiler/src/org.graalvm.compiler.asm.test/src/org/graalvm/compiler/asm/test/AssemblerTest.java
pli-arm: I can still use `masm.nop()` without relocating this source file. The purpose of adding this tricky logic in `AssemblerTest` is to avoid adding `nop()` in each test case and causing potential failures in the future. But I can change my fix if you don't like it.

dougxc:
The issue is that `"AArch64 code buffer needs to start with a nop"` is a HotSpot constraint, not an AArch64 constraint and so it should not be present in non-HotSpot code or tests.

To make the failure mode a little friendlier (Java assertion vs C++ assertion), you could override `org.graalvm.compiler.core.target.Backend.createInstalledCode(DebugContext, ResolvedJavaMethod, CompilationRequest, CompilationResult, InstalledCode, boolean, Object[])` in `org.graalvm.compiler.hotspot.aarch64.AArch64HotSpotBackend` to replicate the logic of the assertion in `NativeJump::patch_verified_entry`.

pli-arm:
Hi dougxc,

> The issue is that `"AArch64 code buffer needs to start with a nop"` is a HotSpot constraint, not an AArch64 constraint and so it should not be present in non-HotSpot code or tests.

I see your point, thanks.

> To make the failure mode a little friendlier (Java assertion vs C++ assertion), you could override `org.graalvm.compiler.core.target.Backend.createInstalledCode(DebugContext, ResolvedJavaMethod, CompilationRequest, CompilationResult, InstalledCode, boolean, Object[])` in `org.graalvm.compiler.hotspot.aarch64.AArch64HotSpotBackend` to replicate the logic of the assertion in `NativeJump::patch_verified_entry`.

I just tried adding a little more friendly Java assertion, but found it breaks other tests if I override `createInstalledCode()` and add assertion logic there because that method is also invoked from some test framework code and truffle code. Target code buffer does not always start with `nop` or `branch` in those invocations, even in non-AOT and non-OSR mode.

Also, I don't think it's good to add too much compiler logic code just for fixing a code-gen test case. So eventually, I decided to go back to the simplest test case fix in the very beginning. I've update this PR just now. Please help merge if it looks good.

dougxc:
> Target code buffer does not always start with nop or branch in those invocations, even in non-AOT and non-OSR mode.

Doesn't that indicate a real issue? If this code is to be installed and run on AArch64 HotSpot, I assume it has to abide by the constraint of a leading `nop`.

pli-arm:
> Doesn't that indicate a real issue? If this code is to be installed and run on AArch64 HotSpot, I assume it has to abide by the constraint of a leading `nop`.

I see the broken tests are mostly in truffle compilation tasks. The AArch64 HotSpot constraint is "instruction at verified entry point should be a nor or jump". Perhaps truffle verified entry point may be not at the start of the code buffer? Sorry for cannot answer your question because I'm new to Graal and don't have much knowledge on Truffle.

dougxc:
@sanzinger are we running any Truffle tests on an AARch64 fastdebug HotSpot build? Can you please ascertain whether we emitting code that violates the invariant for verified entry points on this platform.

sanzinger:
@dougxc  Will do later today.

pli-arm:
Hi @dougxc @sanzinger 

I have further investigation on this today and let me answer the question now.

> Doesn't that indicate a real issue? If this code is to be installed and run on AArch64 HotSpot, I assume it has to abide by the constraint of a leading `nop`.

That is not a real issue. Instead, that's my fault in my Java assertion code. Yesterday I have a guess that `nmethod verified entry point may be not at the start of the code buffer`. I've found this is TRUE after looking at the code at `AArch64HotSpotBackend.emitCodePrefix()`. AArch64 HotSpot may generate `unverified` entry point code before the `verified` entry point at the beginning of the code buffer for monomorphic calls. And when HotSpot invalidates an nmethod, it patches the instruction at the verified entry point. Hence, sometimes the instruction to be patched is not the leading instruction in the code buffer. So in my assertion code, I should search the instruction with the MARKID_VERIFIED_ENTRY mark first and then check if it is the nop placeholder.

I will add the Java assertion and update this PR soon.

pli-arm:
I've updated this PR with Java assertion added. This can detect the missing nop issue and no compiler test failed this time. Please help review.

compiler/src/org.graalvm.compiler.hotspot.aarch64/src/org/graalvm/compiler/hotspot/aarch64/AArch64HotSpotBackend.java
```diff
@@ -126,6 +133,44 @@ protected void bangStackWithOffset(CompilationResultBuilder crb, int bangOffset)
...
+            Object markId = mark.id;
+            if (markId instanceof Integer && (int) markId == config.MARKID_VERIFIED_ENTRY) {
+                // The nmethod verified entry is located at some pc offset.
+                verifiedEntryOffset = mark.pcOffset;
+                break;
+            }
+        }
+        Unsafe unsafe = GraalUnsafeAccess.getUnsafe();
+        int instruction = unsafe.getIntVolatile(targetCode, unsafe.arrayBaseOffset(byte[].class) + verifiedEntryOffset);
+        return instruction == 0xD503201F;   // AArch64 nop instruction
```
dougxc: It would be slightly nicer to use the assembler to generate the nop encoding:
```
        AArch64MacroAssembler asm = new AArch64MacroAssembler(getTarget());
        asm.nop();
        int nop = asm.getInt(0);
        return instruction == nop;
```
The cost of computing each time won't matter since this code is only executed when assertions are enabled.

sanzinger:
I doublechecked the truffle compilations from `SLCompileImmediatelyTestSuite` on JDK14.
No problems there.

```
mx unittest  -XX:NmethodSweepActivity=1000 -XX:+PrintMethodFlushing -Dgraal.TraceTruffleCompilation=true SLCompileImmediatelyTestSuite
```

I wonder why this error only happens during nmethod sweeping? As far as I remember the nop/jmp instruction is also checked when installing code?

pli-arm:
> It would be slightly nicer to use the assembler to generate the nop encoding

Done in my latest update.

PR #1456, opened 2019-07-01
Disable isolates and multithreading by default in the LLVM backend.

loicottet:
Disables unsupported features in the LLVM backend with an error message instead of crashing further down the pipeline.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
@@ -284,8 +284,20 @@ public class SubstrateOptions {
     public static final HostedOptionKey<String> CompilerBackend = new HostedOptionKey<String>("lir") {
         @Override
         protected void onValueUpdate(EconomicMap<OptionKey<?>, Object> values, String oldValue, String newValue) {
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -286,6 +286,16 @@ public static final long getTearDownFailureNanos() {
         protected void onValueUpdate(EconomicMap<OptionKey<?>, Object> values, String oldValue, String newValue) {
             if ("llvm".equals(newValue) && JavaVersionUtil.JAVA_SPEC > 8) {
                 EmitStringEncodingSubstitutions.update(values, false);
+
+                if ((Boolean) values.get(SpawnIsolates)) {
+                    System.out.println("Isolates are currently not supported by the LLVM backend. The build will continue with isolates disabled.");
```
loicottet: What's the preferred way of emitting this warning?

vjovanov: `com.oracle.svm.hosted.SVMHost#warn` If you have access to it. Otherwise, just use `System.out.println("warning: ")`.

vjovanov:
I would wait with this PR until the stack traces PR is merged. This part of the code is moved to the LLVM project.

PR #1458, opened 2019-07-01
Fix SLRootNode#setCloningAllowed()

fniephaus:
Before changing the value of `isCloningAllowed`, a ` CompilerDirectives.transferToInterpreterAndInvalidate()` is needed because it is a `@CompilationFinal` field.


smarr:
@fniephaus this will only invalidate the code writing the field, but won't invalidate any other code with the value compiled in.

fniephaus:
@smarr you're right and that might be a problem. But I'd say it depends on how this infrastructure is supposed to work: if cloning should be disallowed for all future calls, then this might be fine. Otherwise, an assumption is needed to invalidate corresponding compiled code as well.

fniephaus:
Closing this in favor of https://github.com/oracle/graal/pull/1732.

Issue #1408, opened 2019-06-20
BootstrapJVMCI is broken with EconomyCompilerConfiguration on AArch64


pli-arm:
After @thomaswue's below commit of restricting application of canonicalizations, BootstrapJVMCI is broken with -Dgraal.CompilerConfiguration=economy on AArch64.
```
commit 1419ca558c45504065b09825d57acbc6be822636
Author: Thomas Wuerthinger <thomas.wuerthinger@oracle.com>
Date:   Sun Jun 9 18:18:01 2019 +0200

    Restrict application of canonicalizations for better compilation speed.
```

Command line to reproduce this failure (only on AArch64, x86 is good)
```
mx vm -XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCI -XX:+UseJVMCICompiler -Djvmci.Compiler=graal \
  -Dgraal.CompilationFailureAction=ExitVM -Dgraal.CompilerConfiguration=economy -XX:+BootstrapJVMCI -version
```

Stack trace printed
```
Thread[JVMCI CompilerThread24,9,system]: Compilation of jdk.internal.org.objectweb.asm.SymbolTable.addConstantUtf8(String) failed:
org.graalvm.compiler.java.BytecodeParser$BytecodeParserError: java.lang.IllegalStateException: zip file closed
        at parsing jdk.internal.org.objectweb.asm.SymbolTable.addConstantUtf8(SymbolTable.java:800)
        at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.throwParserError(BytecodeParser.java:2470)
        at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.iterateBytecodesForBlock(BytecodeParser.java:3272)
        at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.processBlock(BytecodeParser.java:3074)
        at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.build(BytecodeParser.java:976)
        at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.buildRootMethod(BytecodeParser.java:870)
        at jdk.internal.vm.compiler/org.graalvm.compiler.java.GraphBuilderPhase$Instance.run(GraphBuilderPhase.java:84)
        at jdk.internal.vm.compiler/org.graalvm.compiler.java.GraphBuilderPhase.run(GraphBuilderPhase.java:55)
[...]

PR #1465, opened 2019-07-04
Use memory access stamp of Read/WriteNode to do address lowering.

XiaohongGong:
This patch fixes issue:https://github.com/oracle/graal/issues/1408.

Graal AddressLoweringPhase lowers the general address node to platform
specific address node, like AArch64AddressNode, and together with some
optimizations. It should use the right memory access type to set the
suitable address mode and scaleFactor. But currently it uses the node type.

An AArch64ReadNode is a node merges ZeroExtended/SignedExtended with
ReadNode. If the type of the ReadNode is 32bits. After extending, the
type of the AArch64ReadNode is 64bits, but its access type is still 32bits.
So if the compiler is lowering an AArch64ReadNode and using its node type,
the result address will be wrong.

This patch modifies the stamp of the address to the access stamp.

Change-Id: I0a805134bbff22f9df5626ccdeda95f41ed721e6

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ReadNode.java
+++ b/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ReadNode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2019, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2017, Red Hat Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
@@ -70,6 +70,11 @@ public class AArch64ReadNode extends ReadNode {
         AArch64Kind readKind = (AArch64Kind) lirgen.getLIRKind(accessStamp).getPlatformKind();
         int resultBits = ((IntegerStamp) stamp(NodeView.DEFAULT)).getBits();
         gen.setResult(this, arithgen.emitExtendMemory(isSigned, readKind, resultBits, (AArch64AddressValue) gen.operand(getAddress()), gen.state(this)));
+    }
+
+    @Override
+    public Stamp getAccessStamp() {
+        return accessStamp;
     }
 
     /**
--- a/compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/AddressLoweringByUsePhase.java
+++ b/compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/AddressLoweringByUsePhase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2019, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2017, Red Hat Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
@@ -69,7 +69,7 @@ public class AddressLoweringByUsePhase extends Phase {
             AddressNode lowered;
             if (node instanceof ReadNode) {
                 ReadNode readNode = (ReadNode) node;
-                Stamp stamp = readNode.stamp(NodeView.DEFAULT);
+                Stamp stamp = readNode.getAccessStamp();
                 address = readNode.getAddress();
                 lowered = lowering.lower(readNode, stamp, address);
             } else if (node instanceof JavaReadNode) {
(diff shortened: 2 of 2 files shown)


adinn:
Hi Xiaohong, This patch looks like it fixes the problem and it makes sense to add the getAccessStamp method because (as you explained) the access stamp and result stamp of a ReadNode are not always the same on AArch64.

However, I'm not clear how this got broken (by Thomas's change or anything else).

As far as I am aware the *only* case where the stamp of a ReadNode differs from its access stamp is when you have an AArch64ReadNode that is eliding a sign/zero extend. So, it looks like the bug could only happen if phase AArch64ReadReplacementPhase executes before AddressLoweringByUsePhase.

  AddressLoweringByUsePhase is inserted just before SchedulePhase (see method AArch64HotspotBackendFactory::createSuites)

  AddressLoweringByUsePhase is inserted immediately after FixReadsPhase or (if that is not present in the suite) ExpandLogicPhase (see method AddressLoweringHotSpotSuitesProvider::createSuites)

This doesn't add up because SchedulePhase is the last phase in the suite. It certainly has to precede 
FixReadsPhase and ExpandLogicPhase.

Perhaps Thomas or Stefan can shed light on why this breakage occurred?

thomaswue:
The patch looks good to me. @sanzinger, can you integrate?

Hard to say how my changes around canonicalizations could have triggered this. We are not running that many tests with the economy configuration, because it is a work in progress.

XiaohongGong:
> Hard to say how my changes around canonicalizations could have triggered this. We are not running that many tests with the economy configuration, because it is a work in progress.

The changes in `AddressLoweringHotSpotSuitesProvider.java` will make AddressLoweringPhase applied after AArch64ReadReplacementPhase with economy configuration. This makes the chance that lowering an address used by an AArch64ReadNode.

XiaohongGong:
> As far as I am aware the _only_ case where the stamp of a ReadNode differs from its access stamp is when you have an AArch64ReadNode that is eliding a sign/zero extend. So, it looks like the bug could only happen if phase AArch64ReadReplacementPhase executes before AddressLoweringByUsePhase.
> 
> AArch64ReadReplacementPhase is inserted just before SchedulePhase (see method AArch64HotspotBackendFactory::createSuites)
> 
> AddressLoweringByUsePhase is inserted immediately after FixReadsPhase or (if that is not present in the suite) ExpandLogicPhase (see method AddressLoweringHotSpotSuitesProvider::createSuites)

Yes, you are right. Because appending `AddressLoweringPhase` is after appending `AArch64ReadReplacementPhase`.  In Thomas's patch, modifications in `AddressLoweringHotSpotSuitesProvider.java` makes `AddressLoweringPhase` inserted before last `SchedulePhase` in economy mode, which will make `AArch64ReadReplacementPhase` before `AddressLoweringPhase`.

XiaohongGong:
> AddressLoweringByUsePhase is inserted immediately after FixReadsPhase or (if that is not present in the suite) ExpandLogicPhase (see method AddressLoweringHotSpotSuitesProvider::createSuites)

This is changed in Thomas's patch. In his patch, `AddressLoweringByUsePhase` is inserted before `UseTrappingNullChecksPhase` or (if that is not present in the suite) `SchedulePhase`.

adinn:
Ok, I see now where that was changed in AddressLoweringHotSpotSuitesProvider. Thanks for the explanation, Xiaohong. Your fix is definitely needed.

I'm concerned that it is so easy for the AArch64 code to get broken by changes to generic code like this one. It is not the first time this has happened and it is one of the reasons I have backed off from adding AArch64-specific optimizations to the Graal compiler. Unfortunately that means that the code quality for AArch64 is much poorer than it needs to be.

It would be a big help if the Travis checks and/or internal Oracle tests could be run automatically on AArch64 build as well as x86_64.

dougxc:
Unfortunately it doesn't appear as though ARM support will be added to Travis any time soon: https://github.com/travis-ci/travis-ci/issues/3376
To address this, we're trying to bulk up our internal gating with more ARM machines to improve support for this platform.

XiaohongGong:
> It would be a big help if the Travis checks and/or internal Oracle tests could be run automatically on AArch64 build as well as x86_64.

Yes, I think so.

XiaohongGong:
> To address this, we're trying to bulk up our internal gating with more ARM machines to improve support for this platform.

Thanks @dougxc , this will help us a lot！

XiaohongGong:
Hi @sanzinger  Can you take a look at this PR? Thanks!

XiaohongGong:
@dougxc @sanzinger Can you please take a look at this PR? This will influence our regular tests on AArch64, so thanks a lot if you can take a look at it. Thank you!

dougxc:
@sanzinger please merge this PR asap.
@XiaohongGong we now have AArch64 gates internally so this kind of regression should be caught earlier. Thanks for your patience.

adinn:
>  @XiaohongGong we now have AArch64 gates internally so this kind of regression should be caught earlier.

Hurrah, great news, Doug!

XiaohongGong:
> @sanzinger please merge this PR asap.
> @XiaohongGong we now have AArch64 gates internally so this kind of regression should be caught earlier. Thanks for your patience.

Good news, thanks so much!

XiaohongGong:
@sanzinger Can you please take a look at this PR? I have just rebased it. Thanks!

sanzinger:
I am going to integrate the changes now.

XiaohongGong:
Thank you so much!

sanzinger:
The internal tests are passing. The merge is in progress, the changes should appear on master in the next hours.

XiaohongGong:
@sanzinger  Thanks so much!

PR #1480, opened 2019-07-08
Fix `getCallerClass` methods to acquire the correct frame

### LINKED ISSUE #1479: Access checking fails due to wrong `getCallerClass()` results
I don't know the exact issue but I think `Reflection.getCallerClass()` is (maybe only sometimes) wrong now.  If you look at the stack trace of quarkusio/quarkus#3074, you can see that the result of `Reflection.getCallerClass()` within `Constructor` was `com.sun.xml.bind.v2.runtime.ClassBeanInfoImpl`, which is one frame lower than the class which should have been returned.  I'm thinking it might be a result of oracle/graal@a14d2f1, specifically the factoring out of `StackTraceUtils`. The problem occurs in 19.1.0 but not 19.0.x.

Reference: https://github.com/quarkusio/quarkus/pull/3074

/cc @christianwimmer

dmlloyd:
Fixes #1479

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/StackTraceUtils.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/StackTraceUtils.java
@@ -123,7 +123,6 @@ class BuildStackTraceVisitor extends JavaStackFrameVisitor {
 }
 
 class GetCallerClassVisitor extends JavaStackFrameVisitor {
-    private boolean foundCallee;
     Class<?> result;
 
     @Override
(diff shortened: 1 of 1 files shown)


dmlloyd:
OK this fix is actually wrong: it does move the frame being checked, but it does so in the wrong direction.  Doing an update.

dmlloyd:
This now fixes the problem seen in https://github.com/quarkusio/quarkus/pull/3074.

christianwimmer:
@dmlloyd Do you have a reduced test case for the Quarkus problem? All test cases of `getCallerClass` that I could come up with pass with the current master version, but fail with this PR.

dmlloyd:
That might be tricky.  I can possibly develop one using `Constructor`.  I think I might also need to research if there is special behavior for `@CallerSensitive` - that tickles my memory for some reason.

christianwimmer:
Actually I take it back, I managed to create a test case. The problem is that we probably do the filtering of `Constructor.newInstance` the wrong way: It needs to be filtered when `getCallerClass` is called from the actual constructor - but not when `Constructor.newInstance` itself calls `getCallerClass`.

I'm working on a fix.

christianwimmer:
Replaced by 506f2f188e732219020af3d82b2ec6a97446109a

Issue #1478, opened 2019-07-08
Link Fail JVM_Ge tClassContext Java_java_lang_SecurityManager  19.1.0 ,windows 7 native-image


lilili87222:
java.lib(SecurityManager.obj) : error LNK2019: unresolved external symbol JVM_Ge
tClassContext referenced in function Java_java_lang_SecurityManager_getClassCont
ext
java.lib(SecurityManager.obj) : error LNK2019: unresolved external symbol JVM_Cu
rrentLoadedClass referenced in function Java_java_lang_SecurityManager_currentLo
adedClass0
java.lib(SecurityManager.obj) : error LNK2019: unresolved external symbol JVM_Cu
rrentClassLoader referenced in function Java_java_lang_SecurityManager_currentCl
assLoader0
java.lib(SecurityManager.obj) : error LNK2019: unresolved external symbol JVM_Cl
assDepth referenced in function Java_java_lang_SecurityManager_classDepth
java.lib(SecurityManager.obj) : error LNK2019: unresolved external symbol JVM_Cl
assLoaderDepth referenced in function Java_java_lang_SecurityManager_classLoader
Depth0
G:\demo.exe : fatal error LNK1120: 5 u
nresolved externals

        at com.oracle.svm.hosted.image.NativeBootImageViaCC.write(NativeBootImag
eViaCC.java:343)
        at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator
.java:653)
        at com.oracle.svm.hosted.NativeImageGenerator.lambda$run$0(NativeImageGe
nerator.java:440)

olpaw:
It seems you are not using the Windows SDK for Windows 7 as described in https://github.com/oracle/graal/blob/master/substratevm/README.md

olpaw:
For reference:
```
Setting SDK environment relative to C:\Program Files\Microsoft SDKs\Windows\v7.1\.
Targeting Windows 7 x64 Debug

C:\Users\pwoegere>cd Downloads

C:\Users\pwoegere\Downloads>cd graalvm-ce-windows-amd64-19.1.0

C:\Users\pwoegere\Downloads\graalvm-ce-windows-amd64-19.1.0\graalvm-ce-19.1.0>notepad Hello.java

C:\Users\pwoegere\Downloads\graalvm-ce-windows-amd64-19.1.0\graalvm-ce-19.1.0>bin\javac.exe Hello.java

[...]

PR #1484, opened 2019-07-09
Add support for SecurityManager.getClassContext()

dmlloyd:
This method is used by some logging frameworks and in a few other places as a poor pre-9 alternative to `StackWalker`.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/StackTraceUtils.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/StackTraceUtils.java
@@ -37,16 +37,25 @@ import com.oracle.svm.core.stack.JavaStackWalker;
 
 public class StackTraceUtils {
 
+    private static final Class<?>[] NO_CLASSES = new Class<?>[0];
+    private static final StackTraceElement[] NO_ELEMENTS = new StackTraceElement[0];
+
     public static StackTraceElement[] getStackTrace(boolean filterExceptions, Pointer startSP, CodePointer startIP) {
         BuildStackTraceVisitor visitor = new BuildStackTraceVisitor(filterExceptions);
         JavaStackWalker.walkCurrentThread(startSP, startIP, visitor);
-        return visitor.trace.toArray(new StackTraceElement[0]);
+        return visitor.trace.toArray(NO_ELEMENTS);
     }
 
     public static StackTraceElement[] getStackTrace(boolean filterExceptions, IsolateThread thread) {
         BuildStackTraceVisitor visitor = new BuildStackTraceVisitor(filterExceptions);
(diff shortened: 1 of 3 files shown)


dmlloyd:
@christianwimmer on a similar topic... :)

This might fix or partially fix #1478.

cstancu:
The CI caught a style issue: https://travis-ci.org/oracle/graal/jobs/556310413. I'll fix it on our end, but in the future please keep on eye out for CI failures (even some if some of them are infrastructure failures it still catches style issues).

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/StackTraceUtils.java
```diff
@@ -155,3 +164,22 @@ public boolean visitFrame(FrameInfoQueryResult frameInfo) {
...
+
+    GetClassContextVisitor(final int skip) {
+        trace = new ArrayList<>();
+        this.skip = skip;
+    }
+
+    public boolean visitFrame(final FrameInfoQueryResult frameInfo) {
+        if (skip > 0) {
+            skip--;
+        } else if (StackTraceUtils.shouldShowFrame(frameInfo, true, false)) {
```
christianwimmer: @dmlloyd did you double-check on the HotSpot VM that `getClassContext` really is *not* filtering reflection methods (the `true` for the `showReflectFrames` parameter)? That would be a difference to `getCallerClass`.

dmlloyd: I believe it is not, but that is just from my memory.  I'll double-check it.

dmlloyd:
Sure, I can always fix style issues.  I haven't been able to get the style thing to run locally yet though (IIRC it blows up with a Python-ish stack trace).  So for now I'm reformatting using the Eclipse formatter and then hoping it is happy with the result.  When CI blows up I can go back to fix the additional style issues, though it might take me a bit of time depending what's happening (in summer, I have children running around everywhere so my schedule is erratic).

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/StackTraceUtils.java
dmlloyd: Ah, it is in fact doing so.  I'll fix that.

dmlloyd:
All fixed.

cstancu:
@dmlloyd did you forget to push your latest updates?

dmlloyd:
Oh! I pushed the wrong branch.  That's what I get for running commands from a partial history search :)  Fixed now.

cstancu:
Fixed in https://github.com/oracle/graal/commit/b9a93e22cc138cbc5595baa312aee536d08bafa6.

PR #1507, opened 2019-07-18
Turn comment into description for assertion

fniephaus:
Had to debug to find out what caused the assertion error. This turns the comment into a description for the assertion.

/cc @chumer


boris-spas:
Merged in 19ea1c73f07d2ed73b51bb66414501c07d12d4c0

PR #1523, opened 2019-07-23
Fixes 1519 - Silence unfixable Unsafe automatic substitutions in the JDK

### LINKED ISSUE #1519: JDK11: Silence additional unsafe offset warnings we can't fix.

bobmcwhirter:
Fixes #1519


cstancu:
@bobmcwhirter what is the exact warning in this case?

bobmcwhirter:
Unsafe offset usage not followed by a sign-extend or store, involving those additional JDK11 classes, just like InnocuousThread.

bobmcwhirter:
fwiw, I think my other PR obviates the need for this one. Not completely certain, but it attempts to deal with entire classes of usage instead of just whitelisting unfixable JDK whinging.

cstancu:
Yes, I thought that the other PR might actually fix this case too. Can you confirm that that's the case?

bobmcwhirter:
Yeah, lemme verify. Will get back to you later today or in the AM (US-Eastern).

bobmcwhirter:
Seems not.

```
Warning: RecomputeFieldValue.ArrayBaseOffset automatic substitution failed. The automatic substitution registration was attempted because a call to jdk.internal.misc.Unsafe.arrayBaseOffset(Class) was detected in the static initializer of java.util.concurrent.atomic.AtomicReferenceArray. Detailed failure reason(s):
```

Of course no actual detailed failure reasons actually follow in the output.

I'll dig in tomorrow some more.

bobmcwhirter:
I augmented #1405 which now does include a non-whitelisting fix for this class of warnings. Closing this PR.

PR #1538, opened 2019-07-24
1537 registernetarrays

johanvos:
Register 2 array classes for reflection. They are required by `java.net.NetworkInterface<clinit>`

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
@@ -126,6 +126,9 @@ class PosixJavaNetSubstitutionsFeature implements Feature {
             JNIRuntimeAccess.register(access.findClassByName("java.lang.Boolean").getDeclaredMethod("getBoolean", String.class));
 
             RuntimeReflection.register(access.findClassByName("java.net.InetAddressImpl"));
+            RuntimeReflection.register(access.findClassByName("[Ljava.net.NetworkInterface;"));
+            RuntimeReflection.register(access.findClassByName("[Ljava.net.InterfaceAddress;"));
+
             RuntimeReflection.register(access.findClassByName("java.net.Inet4AddressImpl"));
             RuntimeReflection.register(access.findClassByName("java.net.Inet6AddressImpl"));
             RuntimeReflection.registerForReflectiveInstantiation(access.findClassByName("java.net.Inet4AddressImpl"));


johanvos:
This PR also includes (and thus obsoletes) #1532 
I can split them again, if required.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
```diff
@@ -126,6 +126,9 @@ public void beforeAnalysis(BeforeAnalysisAccess access) {
             JNIRuntimeAccess.register(access.findClassByName("java.lang.Boolean").getDeclaredMethod("getBoolean", String.class));
 
             RuntimeReflection.register(access.findClassByName("java.net.InetAddressImpl"));
+            RuntimeReflection.register(access.findClassByName("[Ljava.net.NetworkInterface;"));
```
dmlloyd: Is this required even if the JNI-based platform is not used?

johanvos: That depends on the implementation. I would assume the non-JNI implementation does the same as the JNI implementation, in which case it is needed.

johanvos:
> This PR also includes (and thus obsoletes) #1532
> I can split them again, if required.

Rebased PR, since #1532 is merged.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java
dmlloyd: I believe this assumption is incorrect, but could you show where this requirement manifested itself?  Maybe there's some reflection I didn't find.

johanvos: Simply running HelloWorld fails.

dmlloyd: On what platform, and did you enable JNI for your test?  I haven't reproduced that, though I'm a few days out of date of master I believe.

dmlloyd: It appears that the problem is likely tied to JNI because of the `loadLibrary` in the `<clinit>` of `NetworkInterface`.  The reason it's important to restrict these cases to conditionals is that every class which is registered for reflection causes another segment of the execution graph to be excluded from dead code elimination, even if it's ultimately unused.

johanvos: There has been a change to rerun `NetworkInterface` clinit at runtime. This calls
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/net/NetworkInterface.java#L62
which initializes `DefaultInterface`which on macos will invoke https://github.com/openjdk/jdk/blob/master/src/java.base/macosx/classes/java/net/DefaultInterface.java#L70 which calls NetworkInterface.getAll() which requires the NetworkInterface[] to be present (https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/net/NetworkInterface.java#L351) 

We can do it conditionally, but it sounds dangerous to me to have 2 implementations. But I'll add a check.

dmlloyd: Yeah this is another JNI method.  The non-JNI versions do not need to access the array reflectively; it can be done directly: https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixJavaNetSubstitutions.java#L4536-L4542

As for having 2 implementations - it's not dangerous, it just means 2x testing.  But, that's the current status quo, so the topic is largely moot.

olpaw: @johanvos indeed, platform LINUX_JNI/DARWIN_JNI need to be able to reflectively create a NetworkInterface array in `NetworkInterface.getAll()` https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/aa318070b27849f1fe00d14684b2a40f7b29bf79/jdk/src/solaris/native/java/net/NetworkInterface.c#L447
and an InterfaceAddress array in https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/aa318070b27849f1fe00d14684b2a40f7b29bf79/jdk/src/solaris/native/java/net/NetworkInterface.c#L701 (which gets called from various places in the JNI code) Thanks for the fix!

@dmlloyd you are right that in the non-JNI case no reflection is required. Since `com.oracle.svm.core.posix.PosixJavaNetSubstitutionsFeature` has `@Platforms({InternalPlatform.LINUX_JNI.class, InternalPlatform.DARWIN_JNI.class})` the non-JNI code is not affected by any of this.

olpaw:
@johanvos this PR is also obsolete, right?

PR #1542, opened 2019-07-25
Fixes #1524 - Replace ForkJoinPool.commonPool() on static heap with d…

### LINKED ISSUE #1524: ForkJoinPool.commonPool() in the image heap
So while working on making camel running on quarkus I've run into some issues related to the ` ForkJoinPool.commonPool()` found in the image heap as result of how Caffeine constructs caches:

- https://github.com/ben-manes/caffeine/issues/332
- https://github.com/quarkusio/quarkus/issues/3300

As mention in the issues a possible workaround is to lazily initialized the pool but that's not always possible as 3th party libraries using Caffeine may not expose a way to do so. 
Another solution is to move the initialization of some classes at runtime which lead to have all the classes that depends on the offending one being needed to be initialized at runtime too which remove some benefits of build time initialization.

/cc @ben-manes

bobmcwhirter:
…eferred delegating implementation.

If ForkJoinPool.commonPool() is discovered to be reachable, replace it
with an implementation that is inactive and able to be written to the
static heap.

At runtime, all mutable methods directly defer to the actual runtime
ForkJoinPool.commonPool().

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/concurrent/DeferredCommonPool.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/concurrent/DeferredCommonPool.java
@@ -1,0 +1,208 @@
+/*
+ * Copyright (c) 2017, 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.core.concurrent;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinTask;
+import java.util.concurrent.ForkJoinWorkerThread;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Pure delegate implementation to ForkJoinPool.commonPool().
+ */
+public class DeferredCommonPool extends ForkJoinPool {
+
+    public DeferredCommonPool() {
+        super(1, new DisallowingForkJoinWorkerThreadFactory(), null, false);
+    }
+
+    @Override
+    public <T> T invoke(ForkJoinTask<T> task) {
+        return ForkJoinPool.commonPool().invoke(task);
+    }
+
+    @Override
+    public void execute(ForkJoinTask<?> task) {
+        ForkJoinPool.commonPool().execute(task);
+    }
+
+    @Override
+    public void execute(Runnable task) {
+        ForkJoinPool.commonPool().execute(task);
+    }
+
+    @Override
+    public <T> ForkJoinTask<T> submit(ForkJoinTask<T> task) {
+        return ForkJoinPool.commonPool().submit(task);
+    }
+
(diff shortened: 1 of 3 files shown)


bobmcwhirter:
My local test case that reproduced the issue prior to this change and works subsequent to it is simply:

```
package com.redhat.test;

import java.util.concurrent.ForkJoinPool;

public class CommonPoolTest {

    public static final ForkJoinPool POOL;

    static {
        POOL = ForkJoinPool.commonPool();
    }

    public static void main(String...args) {
        new CommonPoolTest().run();
    }

    public void run() {
        System.err.println("pool is: " + POOL);
    }
}
```

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/concurrent/DeferredCommonPool.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinTask;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Pure delegate implentation to ForkJoinPool.commonPool().
```
gsmet:
```suggestion
 * Pure delegate implementation to ForkJoinPool.commonPool().
```

bobmcwhirter:
Serves me right for using the github UI. 

Yes. I’ve signed it.

bobmcwhirter:
Rebased to fix CLA issue, included @gsmet attribution (thanks!)

lburgazzoli:
Tested against the project I've created to report https://github.com/ben-manes/caffeine/issues/332 and it fixes the issues

gsmet:
@christianwimmer @cstancu FWIW, this one is in the way of the upcoming Quarkus cache extension which is based on Caffeine.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/concurrent/DeferredCommonPool.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Pure delegate implementation to ForkJoinPool.commonPool().
+ */
+public class DeferredCommonPool extends ForkJoinPool {
+
+    public DeferredCommonPool() {
+        super(1);
```
christianwimmer: The delegate mechanism is an interesting idea. My concern is that we miss some necessary overrides: what if a new version of the JDK adds additional methods to `ForkJoinPool`? We need some safeguards. My first idea was to also override some internal methods of `ForkJoinPool`, i.e., the methods that actually start a thread / modify the queue, and throw an exception in those overrides. But they all seem to be private or package private. So the best I can come up with is to use the super constructor here that also provides a thread factory, and pass in a custom thread factory that throws an exception when a new thread is requested. That way we at least have the guarantee that the delegate pool does not start a thread itself.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/concurrent/ForkJoinPoolFeature.java
```diff
@@ -0,0 +1,57 @@
...
+import org.graalvm.nativeimage.Platforms;
+import org.graalvm.nativeimage.hosted.Feature;
+
+import com.oracle.svm.core.annotate.AutomaticFeature;
+
+@AutomaticFeature
+@Platforms(Platform.HOSTED_ONLY.class)
+class ForkJoinPoolFeature implements Feature {
+
+    static ForkJoinPoolFeature singleton() {
```
christianwimmer: I don't see any usage of this method

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/concurrent/ForkJoinPoolFeature.java
```diff
@@ -0,0 +1,57 @@
...
+        return ImageSingletons.lookup(ForkJoinPoolFeature.class);
+    }
+
+    private DeferredCommonPool commonPool = new DeferredCommonPool();
```
christianwimmer: field can be `final`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/concurrent/DeferredCommonPool.java
bobmcwhirter: I'll look into that.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/concurrent/ForkJoinPoolFeature.java
bobmcwhirter: Remnant, will remove.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/concurrent/ForkJoinPoolFeature.java
bobmcwhirter: Right on.

bobmcwhirter:
Requested changes have been pushed.

ben-manes:
Thanks @bobmcwhirter!

gsmet:
@christianwimmer @cstancu apparently, this one wasn't included in 19.2.0? Was it intentional? It's blocking our cache extension based on Caffeine. Thanks!

gsmet:
@christianwimmer so we have to wait for 19.3.0 now? Or do you plan to integrate it in a 19.2.1?

PR #1553, opened 2019-07-29
Implicitly support auto-reflection for usages of cls.getDeclaredFields()

### LINKED ISSUE #1470: Why SVM cannot automatically detect reflection on getDeclaredFields?
SVM requires configuration for reflections such as getDeclaredFields. It seems very easy to obtain such a fixed function name automatically. Why still need to explicitly set "allDeclaredFields=true" in the configuration file?

bobmcwhirter:
Fixes #1470

If a user invokes getDeclaredFields(), chances are they might want all declared
fields, and this usage will automatically add *all* declared fields to the
reflection data.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java
@@ -29,11 +29,15 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Executable;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import jdk.vm.ci.meta.ResolvedJavaField;
 import org.graalvm.compiler.api.replacements.SnippetReflectionProvider;
 import org.graalvm.compiler.nodes.ConstantNode;
 import org.graalvm.compiler.nodes.ValueNode;
@@ -167,6 +171,13 @@ public class ReflectionPlugins {
             }
         });
 
+        r.register1("getDeclaredFields", Receiver.class, new InvocationPlugin() {
+            @Override
+            public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver) {
+                return processGetFields(b, targetMethod, receiver, snippetReflection, true, analysis, hosted);
+            }
+        });
+
         r.register2("getField", Receiver.class, String.class, new InvocationPlugin() {
             @Override
             public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver, ValueNode name) {
@@ -251,6 +262,58 @@ public class ReflectionPlugins {
                     return false;
                 }
                 throwNoSuchFieldException(b, targetMethod, target);
+            } catch (NoClassDefFoundError e) {
+                /*
+                 * If the declaring class of the field references missing classes a
+                 * `NoClassDefFoundError` can be thrown. We intrinsify `it here.
+                 */
+                Method intrinsic = getIntrinsic(analysis, hosted, b, ExceptionSynthesizer.throwNoClassDefFoundErrorMethod);
+                if (intrinsic == null) {
+                    return false;
+                }
+                throwNoClassDefFoundError(b, targetMethod, e.getMessage());
+            }
+            return true;
+        }
+        return false;
+    }
+
+    private static boolean processGetFields(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver,
+                                           SnippetReflectionProvider snippetReflection, boolean declared, boolean analysis, boolean hosted) {
+        if (receiver.isConstant()) {
+            Class<?> clazz = snippetReflection.asObject(Class.class, receiver.get().asJavaConstant());
+
+            String target = clazz.getTypeName() + ".getDeclaredFields()";
+            try {
+                Field[] fields = declared ? clazz.getDeclaredFields() : clazz.getFields();
+                if (fields.length == 0) {
+                    return false;
+                }
+                Set<Field> intrinsic = new HashSet<>();
+                for (int i = 0; i < fields.length; ++i) {
+                    Field each = null;
+                    if ( analysis) {
+                        if (NativeImageOptions.ReportUnsupportedElementsAtRuntime.getValue()) {
+                            ResolvedJavaField annotated = b.getMetaAccess().lookupJavaField(fields[i]);
+                            if (annotated == null || !annotated.isAnnotationPresent(Delete.class)) {
+                                each = fields[i];
+                            }
+                        } else {
+                            each = fields[i];
+                        }
+                        if (each != null) {
+                            intrinsic.add(each);
+                        }
+                    }
+                }
+                Field[] intrinsicArray = null;
+                if (analysis) {
+                    intrinsicArray = intrinsic.toArray(new Field[intrinsic.size()]);
+                    ImageSingletons.lookup(ReflectionPluginRegistry.class).add(toAnalysisMethod(b.getMethod()), b.bci(), intrinsicArray);
+                } else {
+                    intrinsicArray = ImageSingletons.lookup(ReflectionPluginRegistry.class).get(toAnalysisMethod(b.getMethod()), b.bci());
+                }
+                pushConstant(b, targetMethod, snippetReflection.forObject(intrinsicArray), target);
             } catch (NoClassDefFoundError e) {
                 /*
                  * If the declaring class of the field references missing classes a
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java
@@ -460,8 +460,8 @@ public class ReflectionPlugins {
             Class<?> clazz = snippetReflection.asObject(Class.class, receiver.get().asJavaConstant());
             String target = clazz.getTypeName() + ( declared ? ".getDeclaredConstructors()" : ".getConstructors()");
             try {
-                Constructor[] constructors = declared ? clazz.getDeclaredConstructors() : clazz.getConstructors();
-                Constructor[] intrinsic = getIntrinsicArray(analysis, hosted, b, constructors);
+                Constructor<?>[] constructors = declared ? clazz.getDeclaredConstructors() : clazz.getConstructors();
+                Constructor<?>[] intrinsic = getIntrinsicArray(analysis, hosted, b, constructors);
                 if (intrinsic == null) {
                     return false;
                 }
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java
@@ -25,6 +25,7 @@
 package com.oracle.svm.hosted.snippets;
 
 import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Executable;
 import java.lang.reflect.Field;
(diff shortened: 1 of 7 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java
```diff
@@ -167,6 +171,13 @@ public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Rec
             }
         });
 
+        r.register1("getDeclaredFields", Receiver.class, new InvocationPlugin() {
+            @Override
+            public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver) {
+                return processGetFields(b, targetMethod, receiver, snippetReflection, true, analysis, hosted);
+            }
+        });
```
cstancu: `getFields()` could be intrinsified similarly. It looks like `processGetFields()` is ready to accept it.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java
```diff
@@ -267,6 +278,58 @@ private static boolean processGetField(GraphBuilderContext b, ResolvedJavaMethod
...
 
+    private static boolean processGetFields(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver,
+                                           SnippetReflectionProvider snippetReflection, boolean declared, boolean analysis, boolean hosted) {
+        if (receiver.isConstant()) {
+            Class<?> clazz = snippetReflection.asObject(Class.class, receiver.get().asJavaConstant());
+
+            String target = clazz.getTypeName() + ".getDeclaredFields()";
+            try {
+                Field[] fields = declared ? clazz.getDeclaredFields() : clazz.getFields();
+                if (fields.length == 0) {
```
cstancu: If the class doesn't declare any fields `getDeclaredFields()` could still be intrinsified to return an empty array.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/ReflectionPlugins.java
```diff
@@ -267,6 +278,58 @@ private static boolean processGetField(GraphBuilderContext b, ResolvedJavaMethod
...
+                    return false;
+                }
```
cstancu: Please refactor the code below, until the `pushConstant()` call, into a generic `getIntrinsicArray()` method. It can be reused for intrinsifying calls to `getMethods()`/`getDeclaredMethods()`/`getConstructors()`/`getDeclaredConstructors()` as well. (Feel free to extend this PR with those implementations too.)

bobmcwhirter:
Thanks for the feedback. Wasn’t sure if this was a viable feature to add but I’ll shore it up and expand it now.

cstancu:
The only potential issue that I see with this feature is that there is no filter for which elements to register for reflection, i.e., it will register all fields of `Foo` if `Foo.getDeclaredFields()` is discovered even though a user might want to only add a subset of those fields (using the reflection configuration files method for example). Maybe we should place this feature behind a flag so it can be easily disabled.

bobmcwhirter:
@cstancu 

So, I'm hitting a wall.

`getMethods()` works with JNI enabled and disabled.

`getDeclaredMethods()` works only with JNI _disabled_.

When it's enabled, instead of getting a constant value node for the `Foo.class`, it ends up being an invoke graph full of `JNIAccessibleClass` and such seemingly bumping through `ToReflectedMethod` and such.

I've poked around quite a bit and can't find where the delta is between `getMethods()` and `getDeclaredMethods()` when JNI is enabled.

I'd welcome any pointers you could provide as to what mechanism I should be looking at.

bobmcwhirter:
So, you can run the following native unit tests:

`GetFieldsTest`
`GetDeclaredFieldsTest`
`GetMethodsTest`
`GetDeclaredMethodsTest`

And all but the last pass. The last fails with an `NPE` due to somehow attempting to compile non-analysed code, ultimately from `copyMethods` within `getDeclaredMethods(...)` call chain. This seems to be happening because later on instead of a JavaConstant-able class, we have a `JNIAccessible.getClassObject()` invocation occurring, so the intrinsicification doesn't happen the 2nd time around. But the `copyMethods` doesn't end up in the ConstantPool, so the target method that isn't resolved is actually null, and the attempt to whinge about the unresolved method by name results in an NPE because a null has no method name.

*whew*

Oddly enough, if you find your way to `-H:-JNI` when running the test, then voilá, it passes and is happy.

So I've been unable to figure out why `getDeclaredMethods()` is the only one that triggers this thus far (I've not mucked with testing constructor reflecting yet), but not `getDeclaredFields()`.

Even when using a construct such as:

```
Class<?> c = Foo.class;
c.getMethods();
c.getDeclaredMethods();
```

within a single method, the first call to `getMethods()` always works, and resolves against the java constant for `Foo.class`, whereas the `getDeclaredMethods()` line immediately following, against the same local variable which should point to the same `Foo.class` is going via `ToReflectedMethod`/`JNIAccessibleClass`/the JNI dictionary stuffs.

Anyhow...

Output looks like

```
mx native-unittest GetDeclaredMethodsTest
Use /Users/bob/repos/oracle/graal/vm/mxbuild/darwin-amd64/GRAALVM_UNKNOWN/graalvm-unknown-19.3.0-dev/Contents/Home/bin/native-image for remaining image builds
Building junit image for matching: com.oracle.svm.test.GetDeclaredMethodsTest
Build on Server(pid: 39976, port: 54506)*
[com.oracle.svm.junit.svmjunitrunner:39976]    classlist:   1,827.63 ms
[com.oracle.svm.junit.svmjunitrunner:39976]        (cap):   1,347.29 ms
[com.oracle.svm.junit.svmjunitrunner:39976]        setup:   3,238.06 ms
[com.oracle.svm.junit.svmjunitrunner:39976]   (typeflow):   8,346.75 ms
[com.oracle.svm.junit.svmjunitrunner:39976]    (objects):  12,394.52 ms
[com.oracle.svm.junit.svmjunitrunner:39976]   (features):     394.43 ms
[com.oracle.svm.junit.svmjunitrunner:39976]     analysis:  21,416.57 ms
[com.oracle.svm.junit.svmjunitrunner:39976]     (clinit):     238.45 ms
[com.oracle.svm.junit.svmjunitrunner:39976]     universe:     775.73 ms
[com.oracle.svm.junit.svmjunitrunner:39976]      (parse):     243.43 ms
[com.oracle.svm.junit.svmjunitrunner:39976]      compile:     309.46 ms
Fatal error: org.graalvm.compiler.java.BytecodeParser$BytecodeParserError: java.lang.NullPointerException
	at parsing java.base@11.0.3/java.lang.Class.getDeclaredMethods(Class.java:2309)
	at method: Method[] java.lang.Class.getDeclaredMethods()
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.throwParserError(BytecodeParser.java:2469)
	at com.oracle.svm.hosted.phases.SharedGraphBuilderPhase$SharedBytecodeParser.throwParserError(SharedGraphBuilderPhase.java:92)
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.iterateBytecodesForBlock(BytecodeParser.java:3271)
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.processBlock(BytecodeParser.java:3073)
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.build(BytecodeParser.java:975)
	at com.oracle.svm.hosted.phases.HostedBytecodeParser.build(HostedGraphBuilderPhase.java:107)
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.buildRootMethod(BytecodeParser.java:869)
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.GraphBuilderPhase$Instance.run(GraphBuilderPhase.java:84)
	at jdk.internal.vm.compiler/org.graalvm.compiler.phases.Phase.run(Phase.java:49)
	at jdk.internal.vm.compiler/org.graalvm.compiler.phases.BasePhase.apply(BasePhase.java:197)
	at jdk.internal.vm.compiler/org.graalvm.compiler.phases.Phase.apply(Phase.java:42)
	at jdk.internal.vm.compiler/org.graalvm.compiler.phases.Phase.apply(Phase.java:38)
	at com.oracle.svm.hosted.code.CompileQueue.defaultParseFunction(CompileQueue.java:738)
	at com.oracle.svm.hosted.code.CompileQueue.doParse(CompileQueue.java:705)
	at com.oracle.svm.hosted.code.CompileQueue$ParseTask.run(CompileQueue.java:303)
	at com.oracle.graal.pointsto.util.CompletionExecutor.lambda$execute$0(CompletionExecutor.java:171)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1426)
	at com.oracle.svm.hosted.phases.SharedGraphBuilderPhase$SharedBytecodeParser.handleUnresolvedMethod(SharedGraphBuilderPhase.java:216)
	at com.oracle.svm.hosted.phases.SharedGraphBuilderPhase$SharedBytecodeParser.handleUnresolvedInvoke(SharedGraphBuilderPhase.java:182)
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.genInvokeStatic(BytecodeParser.java:1570)
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.genInvokeStatic(BytecodeParser.java:1545)
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.processBytecode(BytecodeParser.java:5124)
	at jdk.internal.vm.compiler/org.graalvm.compiler.java.BytecodeParser.iterateBytecodesForBlock(BytecodeParser.java:3266)
	... 19 more
```

cstancu:
@peter-hofer can you please take a look at the above issue?

bobmcwhirter:
Bear in mind I am running a JDK11 lifestyle on this. Maybe I’ll try JDK8ish tomorrow but nonetheless I’d like this to work with 11.

bobmcwhirter:
Added tests for `getConstructors()` and `getDeclaredConstructors()`.

The declared variant fails exactly as the `getDeclaredMethods()` test.

peter-hofer:
Reflection and JNI use separate data structures, so I don't see how JNI could directly interfere here. However, JNI's `ToReflectedMethod` itself calls `getDeclaredMethods` and `getDeclaredConstructors`. Could it be that state from multiple methods gets mixed up during intrinsification?

bobmcwhirter:
I honestly have no idea. Just instrumenting my plugin, the `receiver.isConstant()` returns false during one phase, because the receiver is an InvocationPluginReceiver representing an `Invoke(WithException)` around `JNIAccessibleClass.getClassObject()` as the argument for `getDeclaredMethods()`.

When that happens, it ends up attempting to compile through `Class.getDeclaredMethods()` and ends up barfing because `Class.copyMethods()` is missing from the constant-pool used by the `BytecodeParser` and not only do we end up with an un-resolved method, we end up holding a `null` and thus when it attempts to handle the unresolved, we get an NPE.

I have yet to figure out where/how `JNIAccessibleClass.getClassObject()` is being wired up.

But disabling JNI completely (`-H:-JNI`) allows the full build to work, and the result test-app to function correctly.

bobmcwhirter:
Any insight into this?

peter-hofer:
I'm debugging your code now. The unresolved method is `privateGetDeclaredMethods`, which apparently is never seen during the analysis phase, and therefore also not present in the hosted phase. I'll try to figure out why.

peter-hofer:
Actually, this must mean that the substitution code for `Class.getDeclaredMethods` is not used, since we replace this code.

peter-hofer:
I think our JNI code is merely triggering this problem because it uses `Class.getDeclaringMethods()` with a non-constant `Class` object, and the analysis seems to no longer be able through a non-replaced `getDeclaringMethods()` call, leading to methods not being detected as reachable and the image build failing due to these methods being unresolvable during compilation. @cstancu, since you are most familiar with the analysis, could you please look into this?

kaspernielsen:
I really believe this is the wrong solution. There are tons of code out there that iterates through all fields and members to find a subset of them. For example, find all methods annotated with @Inject for every registered bean/component. As I understand this patch would mean that every method and field would now automatically be registered?

If anything this should be opt-in and not opt-out, because unless you know it you would never realize that all the methods/fields were included in the image.

kaspernielsen:
Another question. What if one library requires that this feature is on, and another library requires that this feature is off? You would have no choice but to set it to true. 
I really believe that explicitly requiring people to add fields/methods is the right thing to do. If you allow implicit adding fields/methods it quickly turns into implicitly adding them for all code.

cstancu:
As @kaspernielsen mentioned this could lead to too many fields being registered for reflection. We prefer that reflection registration is either explicit or it is the result of using the Native Image agent to profile the application.

PR #1561, opened 2019-08-01
Relax reflection configuration errors for unresolved types.

### LINKED ISSUE #1548: [native-image] any option to make reflect config more relaxed?
On Graal 19.0.2/MacOS. Not a bug but perhaps a feature request or maybe there is already a way to achieve this, or maybe I'm thinking about it the wrong way. Thanks!

TL;DR - I'd like native-image not to hard fail if something about an entry in reflection-config.json can't be fully resolved (maybe a super type is missing or a type in a generic method reference) because I know the runtime is making decisions based on what types are around and won't be reflectively accessing those that are unresolvable.

I have my Spring feature that I'm using to compile Spring apps with native image. I wanted to try an annotation processor instead that would produce `reflection-config.json` files (and the other files) and a `native-image.properties` pointing to them.

I'll describe the fundamental problem, then more on my specific example below.

If I add some type `DefaultConfigurerAdapter` in `reflection-config.json`, the native-image command may fail if `DefaultConfigurerAdapter` is on the class path but the super type of `DefaultConfigurerAdapter` is not:

```
Error: Error parsing reflection configuration in jar:file:/Users/aclement/gits3/spring-boot-graal-processor/samples/commandlinerunner/unpack/BOOT-INF/lib/spring-boot-autoconfigure-2.2.0.M2plus.jar!/META-INF/native-image/org.springframework.boot/spring-boot-autoconfigure/reflection-config.json:
Class org.springframework.boot.autoconfigure.security.servlet.SpringBootWebSecurityConfiguration$DefaultConfigurerAdapter not found
Verify that the configuration matches the schema described in the -H:PrintFlags=+ output for option ReflectionConfigurationResources.
```
Slightly misleading message because it doesn't tell me the actual problem (that the superclass is missing) but I worked it out.  If I add the superclass then it works.  However, I know the app is fine here, because that super type isn't on the class path the runtime (at startup) will not in fact attempt to reflectively create DefaultConfigurerAdapter.

When I add my entry to reflection-config.json I'm really saying "depending on what is on the class path, I *might* need this to be reflectively accessible". Not definitely accessible - and if you can't resolve the hierarchy of it on a particular run of native-image, that is fine, it means this app isn't using it. The problem is that `reflection-config.json` was computed when `DefaultConfigurerAdapter` was built, because that seemed the right time to make the decision. With that model the problem is that I have no control on native-image that lets me say "if you can't fully resolve entries in the reflection-config.json, that's ok, don't worry, carry on". There is a `--report-unsupported-elements-at-runtime` but it doesn't quite cover this case (should it?).  Or a flag on each reflective-config entry like `errorIfUnresolvable: false/true` because this doesn't apply to all of them.

The alternative is me to compute it when the app is built rather than when the library is built. It just felt like whether some type T might need to be reflectively accessible (including in a reflection-config.json) can be made alongside the library containing that type. I could be persuaded this is the wrong way to think about it... just in some ways felt a lot of duplicate processing having to work this out in every app - effectively having to build duplicate knowledge of what the runtime is going to do into the annotation processor, rather than simply marking it as optional and letting the runtime do its thing.

I've basically worked around this in my feature. My feature is partially driven by a reflection configuration file but it pre-processes the entries before passing them to the native-image builder. The feature attempts to fully resolve the type and doesn't pass it on if parts are missing because it knows if it did it will see a hard failure later that it can't catch or deal with.

--- Spring specifics ---

This is how that specifically relates to Spring. Spring turns on/off a number of features based on what it detects on the class path as the app starts up. This behaviour is usually described in an auto configuration class you create for some feature. As an example here is the auto configuration for web security. I've distilled it to the relevant annotations:

```
@Configuration
@ConditionalOnClass(WebSecurityConfigurerAdapter.class)
public class SpringBootWebSecurityConfiguration {
	@Configuration
	static class DefaultConfigurerAdapter extends WebSecurityConfigurerAdapter {
	}
}
```

Note that auto configurations like this ship with spring-boot, but some of the classes it may depend on are in a spring-security jar that may or may not be on the class path at app startup (they are obviously on the class path when that code above is built). Notice the `@ConditionalOnClass` which says "only process this configuration if that class is around".  That conditions are checked at app startup and if it they fail no attempt will be made to reflectively construct the `DefaultConfigurerAdapter`. (Hence that is why it is ok for me if DefaultConfigurerAdapter is in the reflection-config and yet can't be resolved).

bobmcwhirter:
Fixes #1548.

Instead of throwing a hard error, display a warning and proceed
parsing the remainder. Useful for framework-type bits.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ReflectionConfigurationParser.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ReflectionConfigurationParser.java
@@ -44,9 +44,15 @@ public final class ReflectionConfigurationParser<T> extends ConfigurationParser 
     private static final String CONSTRUCTOR_NAME = "<init>";
 
     private final ReflectionConfigurationParserDelegate<T> delegate;
+    private final boolean allowIncompleteClasspath;
 
     public ReflectionConfigurationParser(ReflectionConfigurationParserDelegate<T> delegate) {
+        this(delegate, false);
+    }
+
+    public ReflectionConfigurationParser(ReflectionConfigurationParserDelegate<T> delegate, boolean allowIncompleteClasspath) {
         this.delegate = delegate;
+        this.allowIncompleteClasspath = allowIncompleteClasspath;
     }
 
     @Override
@@ -75,7 +81,8 @@ public final class ReflectionConfigurationParser<T> extends ConfigurationParser 
 
         T clazz = delegate.resolveType(className);
         if (clazz == null) {
-            throw new JSONParserException("Class " + className + " not found");
+            handleError("Could not resolve " + className + " for reflection configuration.");
+            return;
         }
         delegate.registerType(clazz);
 
@@ -127,7 +134,7 @@ public final class ReflectionConfigurationParser<T> extends ConfigurationParser 
                                     delegate.getTypeName(clazz));
                 }
             } catch (NoClassDefFoundError e) {
-                showWarning("Could not register " + delegate.getTypeName(clazz) + ": " + name + " for reflection. Reason: " + formatError(e) + ".");
+                handleError("Could not register " + delegate.getTypeName(clazz) + ": " + name + " for reflection. Reason: " + formatError(e) + ".");
             }
         }
     }
@@ -162,9 +169,9 @@ public final class ReflectionConfigurationParser<T> extends ConfigurationParser 
         try {
             delegate.registerField(clazz, fieldName, allowWrite, allowUnsafeAccess);
         } catch (NoSuchFieldException e) {
-            throw new JSONParserException("Field " + delegate.getTypeName(clazz) + "." + fieldName + " not found");
+            handleError("Field " + delegate.getTypeName(clazz) + "." + fieldName + " not found.");
         } catch (NoClassDefFoundError e) {
-            showWarning("Could not register field " + delegate.getTypeName(clazz) + "." + fieldName + " for reflection. Reason: " + formatError(e) + ".");
+            handleError("Could not register field " + delegate.getTypeName(clazz) + "." + fieldName + " for reflection. Reason: " + formatError(e) + ".");
         }
     }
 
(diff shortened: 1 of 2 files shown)


cstancu:
This is a good change but I think we should make it conditional on the `--allow-incomplete-classpath` option. And while we're here we should be consistent in how we treat the missing classes/fields/methods encountered during the parsing of reflection config. There are several locations in this file where we catch `NoClassDefFoundError` and show a warning. I think we should make those conditional too, i.e., show a warning if `--allow-incomplete-classpath` is set but throw an error otherwise. The error message should point to the `--allow-incomplete-classpath` option as we do in `SharedGraphBuilderPhase.reportUnresolvedElement()`. Additionally, the two instances of `NoSuchFieldException` and `NoSuchMethodException` that now throw an error should be treated similarly.

bobmcwhirter:
Right on, I'll follow these suggestions. Thanks!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ReflectionConfigurationParser.java
```diff
@@ -44,9 +44,15 @@
     private static final String CONSTRUCTOR_NAME = "<init>";
 
     private final ReflectionConfigurationParserDelegate<T> delegate;
+    private final boolean allowIncompleteClasspath;
 
     public ReflectionConfigurationParser(ReflectionConfigurationParserDelegate<T> delegate) {
+        this(delegate, false);
```
cstancu: @peter-hofer should the `ReflectionConfigurationParser` instance in `com.oracle.svm.configure.config.ConfigurationSet.loadTypeConfig()` be configured with `allowIncompleteClasspath` too? The reflection config there is the result of profiling, could it stil throw on missing classes?

bobmcwhirter:
I looked and a few other instances seemed to have no route to obtain a Hosted option.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ReflectionConfigurationParser.java
peter-hofer: That code only builds an in-memory reflection configuration without doing any lookups on the class path, so it should not throw.

peter-hofer:
I wonder why we currently only show warnings in some cases and fail fatally in others. Maybe there is a legitimate use case for that with which people could avoid using `--allow-incomplete-classpath` if they don't need it otherwise since it may conceal other problems during the build. @cstancu , you originally introduced those warnings, did you have a specific use case in mind back then?

cstancu:
@peter-hofer when this feature was initially implemented every exceptional case resulted in an error and we slowly started relaxing the restrictions over time for specific use cases. This was a long-awaited refactoring (thanks @bobmcwhirter)! I think that we can keep the restriction that missing elements in config files resulted from profiling will throw an error, at least until proven otherwise.

cstancu:
@bobmcwhirter this is now merged but the PR was not marked as merged automaticaly. I think that I pulled your PR in before the force-push, but I double checked and all changes are in the master. I'll close this manually. Thank you!

peter-hofer:
@cstancu , thanks for clarifying.

PR #1571, opened 2019-08-05
Throw MalformedURLException instead of UnsupportedFeatureError.

### LINKED ISSUE #558: SVM: Throw MalformedURLException instead of UnsupportedFeatureError for unsupported protocols
Would it make sense to throw `java.net.MalformedURLException` instead of `com.oracle.svm.core.jdk.UnsupportedFeatureError` for disabled and unsupported protocols? This seems more in line with the `java.net.URL` constructor's contract, can be handled without depending on an SVM API (which I'm not even sure where to get from), and is less likely to "crash" an app than throwing a `java.lang.Error`.

From the `java.net.URL` Javadoc:

> @exception  MalformedURLException  if no protocol is specified, or an unknown protocol is found, or {@code spec} is {@code null}.

bobmcwhirter:
Fixed #558.

If an unsupported protocol is requested, follow the JDK standard
by throwing a normal MalformedURLException with SVM-specific
explanatory message.


bobmcwhirter:
Any update on this?

bobmcwhirter:
This does help JDK behaviour compliance.

cstancu:
@bobmcwhirter this PR was superseded by afc34c7250a026b26f6e18ab44700dc899594938.

Issue #515, opened 2018-07-04
Confusing error message about schema when building native image


dsyer:
If the JSON configuration for allowed runtime reflection contains a class that is not on the class path, the tool tells me to verify that my configuration matches the schema:

```
error: Error parsing reflection configuration in /home/dsyer/dev/scratch/graal/reflect.json:
Class no.such.Object not found
Verify that the configuration matches the schema described in the -H:PrintFlags=+ output for option ReflectionConfigurationFiles.
```

The error is really that `no.such.Object` is not on the classpath, so it's confusing to tell me to check the JSON (which is perfectly valid).

Also running `native-image -H:PrintFlags+` is an error:

```
$ native-image -H:PrintFlags+ -cp target/classes/ HelloWorld
Build on Server(pid: 18485, port: 37717)
error: Could not find option 'PrintFlags+'. Did you mean one of these: PrintFlags. Use -H:PrintFlags= to list all available options.
```

so it isn't helpful even if the JSON is really malformed. If there is a way to print some schema hints from the command line, that's great, but I still don't know what it is.

cstancu:
Thanks for reporting this behavior. The correct syntax for `-H:PrintFlags=` is `-H:PrintFlags=<option-category>` where the `<option-category>` is one of `User, Expert, Debug`. `-H:ReflectionConfigurationFiles` is in the `User` category and the help message for it is `One or several (comma-separated) paths to JSON files that specify which program elements should be made available via reflection.`, so it doesn't actually show you the schema.

You are right, the error message is also wrong and it should instead tell you that the class is missing. By the way all this logic is in [`ReflectionConfigurationParser.java`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/config/ReflectionConfigurationParser.java#L55).

NusaibaNizam:
Hello, May I work on this issue?

spavlusieva:
hi @dsyer , this issue has been inactive for more than 180 days. We’ll close it to focus on current issues. In case it’s still relevant, please reopen it and we’ll take another look. Thank you!

PR #1573, opened 2019-08-06
Confusing error message about schema when building native image.

allandequeiroz:
Targets Issue #515

Hi there, I just gave a try to help with the issue mentioned above, to do so I've used [this small project](https://github.com/allandequeiroz/playground/tree/graal.515) to create a jar and execute my tests.

How I've performed the tests?
`mx native-image -H:ReflectionConfigurationFiles=/Users/a/dev/git/allan/playground/src/main/resources/main.json --debug-attach -jar /Users/a/dev/git/allan/playground/target/playground.jar`

In relation to the first problem "The confusing message", the output now looks like:
```
Listening for transport dt_socket at address: 8000
[playground:65401]    classlist:   2,925.69 ms
[playground:65401]        (cap):   1,788.70 ms
[playground:65401]        setup:  22,722.64 ms
Error: Class com.allandequeiroz.Mian not found in the classpath
Error: Use -H:+ReportExceptionStackTraces to print stacktrace of underlying exception
```

In relation to the second problem where the -H:PrintFlags flag usage instructions wasn't correct, the output now looks like:
```
Listening for transport dt_socket at address: 8000
[playground:85087]    classlist:   3,128.28 ms
[playground:85087]        (cap):   1,921.38 ms
[playground:85087]        setup:   2,554.16 ms
Error: Error parsing reflection configuration in /Users/a/dev/git/allan/playground/src/main/resources/main.json:
Class com.allandequeiroz.Mian not found int the classpath
Verify that the configuration matches the schema described in the -H:PrintFlags=<option-category>(User, Expert, Debug) output for option ReflectionConfigurationFiles.
Error: Use -H:+ReportExceptionStackTraces to print stacktrace of underlying exception
```


allandequeiroz:
> Hello Allan de Queiroz, thanks for contributing a PR to our project!
> 
> We use the [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) to make the copyright of contributions clear. We don't have a record of you having signed this yet, based on your email address allandequeiroz -(at)- gmail -(dot)- com. You can sign it at that link.
> 

That's strange, my name is into that very page and I've used that e-mail address to send my OCA assigned. 
Allan De Queiroz - OpenJDK GraalVM

dougxc:
I'll fix the OCA issue. The underlying problem is that the database used by the @graalvmbot is separate from the OCA website.

allandequeiroz:
Thank you, @dougxc

cstancu:
@allandequeiroz thank you for your contribution. There is some overlap between this PR and https://github.com/oracle/graal/pull/1561 which refactors the error handling in the `ReflectionConfigurationParser`. I think that throwing a more appropriate error rather than a [`JSONParserException`](https://github.com/oracle/graal/pull/1561/files#diff-4fcb72e73b27c8124071c23ef5e7a4deR260) is a good idea and allows for a better error message, maybe an `UnresolvedElementException` like we do in the [`SharedBytecodeParser`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/SharedGraphBuilderPhase.java#L238). Doing that should be easier after the refactoring by @bobmcwhirter is merged.

allandequeiroz:
Hi, @cstancu thanks for your feedback.

I've done a few changes but in relation to that particular problem of asking the user to check the schema when a class is not part of the classpath `UnresolvedElementException` looks a bit broad as well.

That particular block of code is clearly intentioned to deal with a missing class problem:
```
T clazz = delegate.resolveType(className);
        if (clazz == null) {
            ...
        }
```
Throwing an `UnresolvedElementException` here feels less accurate than `NoClassDefFoundError`, but is done anyway.

The second point is in relation to the error handling, I could use something like reflection to instantiate the exceptions and avoid the code duplication but looked a bit too much as well.

Also, I had to add `com.oracle.graal.pointsto` as a dependency to `com.oracle.svm.core` to make use of `UnresolvedElementException`.

I don't know if it was the expected result so, glad to make further changes if is the case.

allandequeiroz:
Hi @cstancu, I've just updated this branch to solve a conflict.
The results (information to the user) still an improvement compared with the version at the main branch but apparently, something is not right with the changes I've made in relation to the dependencies but I'm sure what is exactly, do you have any idea here? https://travis-ci.org/oracle/graal/jobs/570717764

Thanks

PR #1586, opened 2019-08-07
Allow Array.set(objArray, index, null) to not throw NPE.

### LINKED ISSUE #1585: Array.set(..) behaves differently
`java.lang.reflect.Array.set(Object array, int index, Object value)` behaves differently in SVM than in a regular JDK.

Pass in `null` as the value:
-> JDK: works fine
-> SVM: throws NPE

The JavaDoc says that this could throw a NPE, but it isn't exactly clear whether this refers to the `array` or the `value` param. Looks like JDK and SVM interpret this differently. Not sure who is right, but maybe it would make sense to align them.

bobmcwhirter:
Fixes #1585

Avoid checking assignability of the value to the array
element type if the value is null, and therefore `.getClass()`
is pretty much guaranteed to throw an NPE.

@cstancu I have no idea how to make Checkstyle actually stop uselessly whinging about the import in my test:

`/Users/bob/repos/oracle/graal/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/ArraySetTest.java:28: Illegal import - java.lang.reflect.Array.`

I've attempted a `// Checkstyle: off` but that doesn't seem sufficient.

I also question the need for such stringent checkstyle/linting/formatting in tests. Seems to just cause more non-functional work to make random compiler tools "happy" about code that doesn't even ship to end-users. Sorta discourages testing, in fact.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangReflectSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangReflectSubstitutions.java
@@ -364,7 +364,10 @@ final class Target_java_lang_reflect_Array {
                 return;
             }
         } else if (array instanceof Object[]) {
-            if (array.getClass().getComponentType().isAssignableFrom(value.getClass())) {
+            if (value == null) {
+                ((Object[]) array)[index] = null;
+                return;
+            } else if (array.getClass().getComponentType().isAssignableFrom(value.getClass())) {
                 ((Object[]) array)[index] = value;
                 return;
             }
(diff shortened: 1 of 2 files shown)


bobmcwhirter:
image

thomaswue:
Thanks for the bug fix contribution.

The accurate way to disable checkstyle is via "// Checkstyle: stop". One can also disable the formatter for a section. See this file as an example: https://github.com/oracle/graal/blob/e7df682fc26830b5e5df74996c948a087c23b3be/substratevm/src/com.oracle.svm.core.jdk8/src/com/oracle/svm/core/jdk8/zipfile/ZipFile.java

The stringent checkstyle/linting/formatting applies to the whole code base. This keeps the several million lines of code we are working on in a canonical form and in the case of checkstyle can also help catching basic bugs.

cstancu:
@bobmcwhirter in this case `// Checkstyle: allow reflection` should be enough.

bobmcwhirter:
I understand that these rules apply to the entire codebase, but perhaps I'm arguing that *some* shouldn't apply to tests.

While SVM might not want its own code to use `java.lang.reflect.*` stuff, users certainly might use it, and SVM should support it, so SVM should test using `java.lang.reflect.*` stuff and these rules get in the way of that.

Users do weird things, so should the tests.

dougxc:
To avoid CI round trips regarding Checkstyle, it should help to run `mx checkstyle` locally before committing. Combined with judicious use of `// Checkstyle: stop` and `// Checkstyle: resume`, you should find your time fighting the CI greatly reduced while being able to write tests for any pattern that comes to mind.

PR #1591, opened 2019-08-08
Allow -J-Xmx and -J-Xms to `native-image` command.

### LINKED ISSUE #1588: native-image --no-server -J-Xmx has no effect
Using `native-image --no-effect -J-Xmx2048m ...` has no effect, in spite of [the documentation saying it's supported](https://www.graalvm.org/docs/reference-manual/aot-compilation/), because the custom `-Xmx` is appended to the arguments after `native-image`'s own calculated `-Xmx`, and the JVM is taking the first `-Xmx` it finds.

This was also reported in a comment on #903, but that issue was closed.

By the way, why does it default to using 80% of the machines physical memory? On my machine, I'm running browsers and slack and all sorts of other stuff, I definitely never have 80% of my physical memory free to do just one thing, so I frequently see my native-image processes being OOM killed by the OS due to this. I don't understand why that default would be chosen, it only makes sense on a dedicated build machine, and even then, build tools (like Maven) that are likely invoking the graal compiler are probably already taking up more than 20% of the available memory.

bobmcwhirter:
Assume the user knows what they're doing if they ask for
specific values of Xmx/Xms when invoking `native-image`.

Fixes #1588

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
@@ -193,7 +193,13 @@ class DefaultOptionHandler extends NativeImage.OptionHandler<NativeImage> {
             if (headArg.equals("-J")) {
                 NativeImage.showError("The -J option should not be followed by a space");
             } else {
-                nativeImage.addCustomJavaArgs(headArg.substring(2));
+                if (headArg.startsWith("-J-Xmx")) {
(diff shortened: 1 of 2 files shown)


bobmcwhirter:
CI lost its mind.

bobmcwhirter:
Any interest in this?

olpaw:
@bobmcwhirter if the user wants custom xmx/xms values he should use the `--no-server` flag. We don't want to allow users to change mem settings for server-based builds otherwise e.g. a user using`-J-Xmx100000` and another one using `-J-Xmx100001` would no share the same warmed-up build server.

PR #1595, opened 2019-08-09
Simplify SLReadLocalVariableNode

fniephaus:
/cc @chumer

### DIFF
--- a/truffle/src/com.oracle.truffle.sl/src/com/oracle/truffle/sl/nodes/local/SLReadLocalVariableNode.java
+++ b/truffle/src/com.oracle.truffle.sl/src/com/oracle/truffle/sl/nodes/local/SLReadLocalVariableNode.java
@@ -44,7 +44,6 @@ import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.dsl.NodeField;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.frame.FrameSlot;
-import com.oracle.truffle.api.frame.FrameSlotKind;
 import com.oracle.truffle.api.frame.FrameUtil;
 import com.oracle.truffle.api.frame.VirtualFrame;
 import com.oracle.truffle.sl.nodes.SLExpressionNode;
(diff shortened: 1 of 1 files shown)


fniephaus:
Mh, I'm confused. `FrameWithBoxing` used to actually read the value:
https://github.com/oracle/graal/blob/e7cbc873c03a24a920b5c1b778d0e95e7b77ef88/compiler/src/org.graalvm.compiler.truffle.runtime/src/org/graalvm/compiler/truffle/runtime/FrameWithBoxing.java#L209-L212

So now that `FrameWithBoxing` is gone, what's the difference between using:

```java
frame.getFrameDescriptor().getFrameSlotKind(getSlot()) == FrameSlotKind.Long;
```

and 

```java
frame.isLong(getSlot());
```

@eregon?

fniephaus:
I should mentioned: I actually encountered situations where `frame.getFrameDescriptor().getFrameSlotKind(getSlot()) == FrameSlotKind.Long;` is `true`, but `frame.isLong(getSlot());` is `false`. How is that possible? I'm sure it's my fault, but I don't know where to look.

fniephaus:
Ok, I can explain the inconsistency: some part of my code set the frame slot kind to `Boolean` after an `Object` type value was stored in a frame. Nonetheless, I still don't understand when the `frame.is*` methods are actually useful.

woess:
Thanks, your fix is correct. One should use the Frame.is* methods to check the current type of the frame slot. The FrameSlotKind is only relevant for writes. Probably a copy&paste error.

fniephaus:
@woess thanks for the info. Is there any way a frame slot's content type can get out of sync with its corresponding FrameSlotKind of the frame descriptor slot? why is this information not managed at one place anyway?

woess:
Yes, the frame slot's content type can get out of sync with the FrameSlotKind of the frame descriptor. There may be multiple live Frame instances of the same FrameDescriptor and when the kind/type of a frame slot changes, the change is only reflected in the current Frame instance. A typical example for when that happens is a recursive call in which the frame slot kind/type changes, leaving all other frames of the same function on the stack out of sync with the FrameDescriptor. The next write of the frame slot will bring it back in sync with the FrameSlotKind.

woess:
> why is this information not managed at one place anyway?

Good question. The shared FrameSlotKind is used to ensure frame slots are consistently set values of the same type so as to avoid undesired polymorphism and allow virtualization, i.e. scalar replacement (the FrameSlotKind needs to be correctly set for that to work currently).

Issue #1600, opened 2019-08-10
[native-image] [questions] HTTPS support and supplying dependencies like libsunec.so or libsunec.dylib with the binary


neomatrix369:
**_Linux_**

I'm building an app that is using HTTPS to connect to third-party services. So far it does well when I run the native image in this form:

```
native-image-app -Djava.library.path=${JAVA_HOME}/jre/lib/amd64 [other CLI options]
```
and when the `${JAVA_HOME}/jre/lib/amd64` folder and its content with valid artifacts exist.

Otherwise, it silently terminates with an exit code of 99.

I'm using the `--static` flag to build the app, but this does not seem to have any effect on the final binary.

So I rebuilt the binary with the below options:

```
CLI_ARG_JAVA_LIB_PATH="-Djava.library.path=${JAVA_LIB_PATH}"
OPTIONS="${OPTIONS} --static ${CLI_ARG_JAVA_LIB_PATH}"
native-image ${OPTIONS} -jar ${JARFILE} ${IMAGE_NAME}
```

But that silently fails as well:
- _GraalVM_:  with exit code 99.
- _OpenJDK8_: Unexpected error - SSLException: java.lang.RuntimeException: Unexpected error: java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty

I then re-ran it by passing in the additional flags:

```
$ export CLI_ARG_JAVA_LIB_PATH="-Djava.library.path=${JAVA_LIB_PATH} -Djavax.net.ssl.trustStore=cacerts -Djavax.net.ssl.trustStorePassword=changeit"

$ native-image-app ${CLI_ARG_JAVA_LIB_PATH}
```

But still got the same error in the OpenJDK environment (not in the GraalVM environment).

---

**_macOS_**

I'm aware this support isn't available for the macOS. How is everyone coping with it? Are you supplying the necessary dependencies along with the native image binary? There are ways to pack it into a `.sh` file which unpacks the binaries, any other alternatives?

_The other issue here is that the native image built, works smoothly on GraalVM (when lib* dependencies are supplied) while does not work when the same is done on traditional JDK i.e. OpenJDK installations._

neomatrix369:
Solved the issue with `java.security.InvalidAlgorithmParameterException`  by following the discussion on https://github.com/oracle/graal/issues/1034.

neomatrix369:
Would still like to know alternative means to packing the native image with its dependencies.

Issue #1336, opened 2019-05-26
Single-binary ECC (ECDSA/ECDHE) TLS support for SubstrateVM


lvh:
#951 addresses the issue that SunEC becomes mandatory for GraalVM native-image binaries doing TLS. While this is fine for some uses of native-image, people sometimes want good TLS support, including ECC, in a single binary. My example in point for this is shipping CLI tools.

In #951, a workaround was discovered that easily disables the SunEC provider. A downside is that only DHE/RSA is supported. The purpose of this ticket is to examine how to get single binary that still supports ECDHE.

One way that this could be resolved is if native-image supported e.g. Sulong's LLVM implementation for System.loadLibrary, or native-image learned how to System.loadLibrary libraries that are actually embedded in the image. I have no idea if that's possible or realistic: but I'm not an Oracle employee so deep engineering on GraalVM itself feels a little scary :) I'm calling this approach "loading embedded dynamic libraries".

Another way to hack this together is to add libsunec.so (or .dylib or .dll) as an embedded resource, set `java.library.path` to a few well-known locations that are likely to be writable across platforms (presumably more than one such location), and write the dylib out at runtime  (and delete it when you're done). I'm assuming the runtime impact of that is not too terrible, but the implementation. I'm calling this approach "loading libraries shipped as resources". Because the code that calls System.loadLibrary is generated by SubstrateVM, it's not clear to me if I can make that use System.load with an explicit path, or if you have to modify `java.library.path`.

Another way that perhaps doesn't require as much hacking on Graal internals is to use a different provider. The best (on the metric of "satisfying my proclivities as a cryptographic engineer") provider available is probably [conscrypt](https://github.com/google/conscrypt), a provider published by Google using BoringSSL. I believe it's the default in modern Android. Unfortunately because BoringSSL itself is a C product, you end up with exactly the same problem that the SunEC provider has: a runtime dependency on a dynamic library.

While I generally don't love BouncyCastle (I am singularly convinced my dinky HTTPS-speaking CLI tool should not have a qTESLA implementation...) it is a pure-Java implementation and therefore "easy" for GraalVM to get working. For the rest of this ticket I'll be focusing on BouncyCastle.

To get replace the SunEC provider, two things need to happen:

1. Add an override that prevents SunEC from loading in the first place with `-J-Djava.security.properties=java.security.overrides`, where java.security.overrides has lines like `security.provider.3=something` (3 is the default index of SunEC)
[...]

Issue #1165, opened 2019-04-14
[SSL] the trustAnchors parameter must be non-empty


smoell:
I have a dockerized that uses the AWS SDK for Java and needs SSL support, unfortunately, I get the following error:

```
Caused by: io.netty.handler.codec.DecoderException: java.lang.RuntimeException: Unexpected error: java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty
at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:472)
at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:278)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1434)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:965)
at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:163)
at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:644)
at io.netty.channel.nio.NioEventLoop.processSelectedKeysPlain(NioEventLoop.java:544)
at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:498)
at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:458)
```

Dockerfile

```
FROM oracle/graalvm-ce:1.0.0-rc14 AS build-aot

FROM debian:9-slim
LABEL maintainer="Sascha Möllering <smoell@amazon.de>"

ENV javax.net.ssl.trustStore /cacerts

RUN apt-get update && apt-get install -y curl
COPY target/reactive-vertx /usr/bin/reactive-vertx
[...]

Issue #1034, opened 2019-03-04
Calling https URL throws Caused by: java.lang.RuntimeException: Unexpected error: java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty


codependent:
My native image makes a HTTPS call to a remote resource and throws a `Caused by: java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty
`

The code is pretty straightforward:

```
@Controller("/")
public class ExampleController {

    private static final Log LOG = LogFactory.getLog(ExampleController.class);
    private HttpClient httpClient;

    public ExampleController() {
        this.httpClient = httpClientFactory.create(HttpClientSettings.adapt(new ClientConfiguration()));
    }

    @Get("/http")
    public String httpClient() throws IOException {

        String url = "https://www.google.com/search?q=httpClient";
        HttpGet request = new HttpGet(url);

        // add request header
        HttpResponse response = httpClient.execute(request);

        System.out.println("Response Code : "
                + response.getStatusLine().getStatusCode());
        return "ok";
    }
}
```
I've followed the documentation at https://github.com/oracle/graal/blob/master/substratevm/URL-PROTOCOLS.md and https://github.com/oracle/graal/blob/master/substratevm/JCA-SECURITY-SERVICES.md to enable security services, adding `--enable-https`

The only thing I haven't done is adding `libsunec.so`. I've looked for it in various JDK installations (1.0.0-rc-12-grl, 8.0.202-zulu, and Oracle's jdk1.8.0_201.jdk), however none of the contained the .so file.

Is that the reason why I'm getting that error? If so, where can I get the needed .so file? An updated of the related doc would come in handy in case it is not shipped by default with any jdk.

cstancu:
@codependent does this fix work for you https://github.com/oracle/graal/issues/768#issuecomment-451776587? Otherwise, can you point to the repo where I can replicate this?

cstancu:
By the way `libsunec.so` is in GraalVM under `graalvm-ce-1.0.0-rc12/jre/lib/amd64/libsunec.so`. Various JDK distributions might have it in different locations. However, I don't think missing it is related to that issue.

codependent:
Regarding `libsunec.so`, I am using the mac graal distribuition which doesn't include it. I downloaded the linux version and copied it to my project (https://github.com/codependent/graal-app).

[...]

PR #1604, opened 2019-08-11
[Docs] Docs updated for libsunec.so for linux when building a native-image supporting https (security related builds)

neomatrix369:
substrateVM: Adding clarification about supplying `libsunec.so` for Linux with the native-image in the respective security docs for clarity

From previous attempts, it is found that only when GraalVM's version of `libsunec.so` for Linux is used, does the native image app run otherwise it silently fails, see issues #1600 and #1034. 

https://github.com/oracle/graal/issues/768, https://github.com/oracle/graal/issues/1336 and https://github.com/oracle/graal/issues/1165  are related and may also help.

_Happy for the texts in both the markdown files to be reworded as long as it helps everyone._


neomatrix369:
I corrected my previous commit as later on I found out using both GraalVM's version of `libsunec.{so|dylib}` and that of any other JDK implementation have small differences, in order for the later version to work, we need to also bundle `cacerts`. Happy to be proven wrong, but in the use-cases, I have tried, under both Linux and MacOS, this has been my experience and observation. I can help reproduce this if necessary.

cstancu:
`libsunec` is not statically linked in the image.

PR #1619, opened 2019-08-18
Issue 1598

Teinnsei:
I think the initialization of enum's static fields for analysinc is redundant, because we need only $VALUES field; as an alternative we can get by invoke `(Enum<?>[]) type.getJavaClass().getEnumConstants()`. If we remove the initialization of all the fields thet we don't need, in this case the issue would not be reproduce.
The rest of the static fields will be added by `BytecodeParser` if this fields will be read.

I'm new in Graal project, so I am not sure this opinion is 100% correct, please review it.


Teinnsei:
I send OCA approximately one hour ago

Teinnsei:
My option is not correct, because this bug can be reproduce like this:

```
import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;

public enum UnsafeEnum {
    INSTANCE;
    
    static AtomicReferenceFieldUpdater<AtomicRFUTest, Integer> OWNER_UPDATER$VALUES =
            AtomicReferenceFieldUpdater.newUpdater(AtomicRFUTest.class, Integer.class, "i");

}
```

PR #1637, opened 2019-08-23
Don't try to build symbol maps when only writing relocatable file

### LINKED ISSUE #1636: Don't try to write symbol map when building a relocatable file only
When NativeImageOptions.ExitAfterRelocatableImageWrite is set, there is no linker returned and there should be no linker invocation. 
When building an executable, this works fine. However, this ImageKind will now also introduce a main function in the relocatable file. This can be avoided by setting ImageKind to SharedLibrary, but this currently causes a NullPointerException as SharedLibraryViaCCBootImage will use the linker to generate a symbol map.

johanvos:
This fixes #1636

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/SharedLibraryViaCCBootImage.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/SharedLibraryViaCCBootImage.java
@@ -29,6 +29,7 @@ import java.lang.reflect.Method;
 import java.nio.file.Path;
 import java.util.List;
 
+import com.oracle.svm.hosted.NativeImageOptions;
 import org.graalvm.compiler.debug.DebugContext;
 
 import com.oracle.svm.core.LinkerInvocation;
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/SharedLibraryViaCCBootImage.java
```diff
@@ -54,6 +55,9 @@ public SharedLibraryViaCCBootImage(HostedUniverse universe, HostedMetaAccess met
     @Override
     public LinkerInvocation write(DebugContext debug, Path outputDirectory, Path tempDirectory, String imageName, BeforeImageWriteAccessImpl config) {
         LinkerInvocation inv = super.write(debug, outputDirectory, tempDirectory, imageName, config);
+        if (NativeImageOptions.ExitAfterRelocatableImageWrite.getValue()) {
+            return inv;
+        }
         writeHeaderFiles(outputDirectory, imageName, inv.getSymbolAliases(), false);
```
pejovica: Have you considered passing `Collections.emptyMap()` in case when `inv == null`?

olpaw:
This looks obsolete

PR #1638, opened 2019-08-23
Update sulong dynamic dependency

galderz:
* Sulong is now part of the graal repo, it should be referenced from the root of the repo.
* Otherwise you get error `Imported suite 'sulong' not found (binary or source).`

### DIFF
--- a/vm/README.md
+++ b/vm/README.md
@@ -40,17 +40,17 @@ To start the build, you can either run:
 
 1.
 ```bash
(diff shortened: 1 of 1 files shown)


galderz:
I work for Red Hat, I believe I should be covered under Red Hat's OCA.

ansalond:
Thank you for the PR. I'm getting it merged.

PR #1642, opened 2019-08-28
Delay insertion of membars for volatile accesses

rwestrel:
This is related to:
https://github.com/oracle/graal/pull/313

Volatile reads/writes can be emitted on aarch64 without explicit membar instructions. This requires the aarch64 backend to identify volatile reads/writes. When a volatile LoadField is lowered, rather than add the required membars, this change creates a VolatileRead node. When a volatile StoreField is lowerer, again required membars are not added, instead the Write is flagged as volatile.
In this change, required membars are added as a final pass of optimizations. A subsequent change would change the way aarch64 handles volatile accesses: membars wouldn't be emitted and instead the right instruction sequence would be emitted by the aarch64 backend.

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/LateMembarInsertionTest.java
+++ b/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/LateMembarInsertionTest.java
@@ -1,0 +1,235 @@
+/*
+ * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, Red Hat Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package org.graalvm.compiler.hotspot.test;
+
+import jdk.vm.ci.hotspot.HotSpotJVMCIRuntime;
+import jdk.vm.ci.meta.ConstantReflectionProvider;
+import jdk.vm.ci.meta.MetaAccessProvider;
+import jdk.vm.ci.meta.ResolvedJavaType;
+import jdk.vm.ci.runtime.JVMCI;
+import jdk.vm.ci.runtime.JVMCIBackend;
+import org.graalvm.collections.EconomicMap;
+import org.graalvm.compiler.core.common.type.Stamp;
+import org.graalvm.compiler.core.test.GraalCompilerTest;
+import org.graalvm.compiler.nodes.ConstantNode;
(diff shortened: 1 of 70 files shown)


sanzinger:
Hi @rwestrel 

this PR is pretty huge in this UI. Can you rebase 7d7c9ea29726f4210028b68bf695b891732f7eec on the latest master so we can remove the cherry-picked commits like 2267f5ffcd and so on. This makes the history a bit nicer too.

sanzinger:
We should remove commit 99d162ffcce2e9676d1eec2689b6f2182d44a1fa (`[GR-13707] Allow calling pthread_mutex_* and pthread_mutexattr_* functions with Sulong`) from this PR (its original commit is cd17e5a4145ba74f40b88f792f30cf85a41a4c11). Can't we have just base your change a512d7af5ea7aed3015f16692ce14052f91fe38b  on most recent master branch?

rwestrel:
@sanzinger: right. Leave me a bit of time to figure out what went wrong.

rwestrel:
I can't reproduce locally the sulong unittest failures from the travis build.

compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/LateMembarInsertionTest.java
```diff
@@ -0,0 +1,235 @@
...
+        return Arrays.stream(blocks).flatMap(b -> schedule.nodesFor(b).stream()).filter(n -> n instanceof MemoryAccess).map(
+                        n -> new TypePair(n instanceof ReadNode, classForAccess((FixedAccessNode) n))).collect(Collectors.toList());
+    }
+
+    private List<TypePair> compile(String test) {
+        StructuredGraph graph = getFinalGraph(getResolvedJavaMethod(test));
+        return getAccesses(graph);
+    }
+
+    private static Class<?> classForAccess(FixedAccessNode n) {
```
gilles-duboscq: If you don't make this static you can use `getMetaAccess` and `getConstantReflection` from the super-class instead of calling `JVMCI.getRuntime().getHostJVMCIBackend()`.
Also i'm not sure about the downcasts to be able to reflect back to the `Class` object. You could do the reverse: return a `ResolvedJavaType` here and get `ResolvedJavaType` for `VolatileAccess.class` via `getMetaAccess().lookupJavaType(VolatileAccess.class)`.

gilles-duboscq:
Why did you go for a subclass for the read node and a boolean field for the write node?

Thank you for the separation of `getLocationIdentity` and `getKilledLocationIdentity` this is long overdue!

rwestrel:
Thanks for reviewing this, Gilles.

> Why did you go for a subclass for the read node and a boolean field for the write node?

I started with both volatile reads and writes handled with a boolean field then I noticed volatile reads would need allowedUsageTypes = Memory in the NodeInfo annotation so I introduced a new class for volatile reads. Handling of write nodes is quite straightforward so the boolean field seems to work well enough.

Can you comment on the CI failures: not sure how my change affects sulong. Can they be a glitch of the CI or do they need further investigation?

gilles-duboscq:
Regarding the sulong issue, it's likely some problem with the LLVM version unrelated to your changes

compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/LateMembarInsertionTest.java
rwestrel: I went with your suggestion. I could move the test in core.test from hotspot.test as a result. But now travis fails right away with some apt-get errors.

thomaswue:
@rwestrel Don't worry about Travis. We regularly run into configuration issues with it. And we can only run a fraction of tests there anyway as it would otherwise overload the system. If locally "mx gate" works it is sufficient.

sanzinger:
FYI, the internal tests seem to pass. The merge process is running. The merge must be through in the next hours.

sanzinger:
FYI: The integration is in progress. This change requires some internal changes too before we can merge this change.

rwestrel:
> FYI: The integration is in progress. This change requires some internal changes too before we can merge this change.

Thanks for the update and for helping with this change.

rwestrel:
> FYI: The integration is in progress. This change requires some internal changes too before we can merge this change.

Any update on this?

dougxc:
Hi @rwestrel , sorry for the delay. @sanzinger is currently on vacation and will return to this integration as soon as he is back next week and keep you updated on his progress.

rwestrel:
> Hi @rwestrel , sorry for the delay. @sanzinger is currently on vacation and will return to this integration as soon as he is back next week and keep you updated on his progress.

Sounds good. Thanks.

sanzinger:
Hi, I am back now. I will continue later this week.

sanzinger:
I have rebased the PR to the latest master and increased the codeSizeIncrease - for some cases the code size increase is well above 2

rwestrel:
Excellent. Thank you.

PR #1643, opened 2019-08-29
Cleanup MethodHandlePlugin

mur47x111:
Use `ResolvedMethodHandleCallTargetNode` to revert to the original invoke if an invocation is not inlined with `MethodHandlePlugin`.

### DIFF
--- a/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/MethodHandlePlugin.java
+++ b/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/MethodHandlePlugin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2019, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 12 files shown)


compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/MethodHandlePlugin.java
```diff
@@ -90,26 +89,20 @@ public boolean handleInvoke(GraphBuilderContext b, ResolvedJavaMethod method, Va
                     argumentsList.initialize(i, b.append(argumentsList.get(i)));
                 }
 
-                boolean inlineEverything = false;
-                if (safeForDeoptimization) {
-                    // If a MemberName suffix argument is dropped, the replaced call cannot
-                    // deoptimized since the necessary frame state cannot be reconstructed.
-                    // As such, it needs to recursively inline everything.
-                    inlineEverything = args.length != argumentsList.size();
-                }
```
mur47x111: hi @christhalinger , I am adding JSR 292 deoptimization support and would like to use your expertise. 

First, I am wondering if we actually inline recursively with the current code. Or we simply need to force inline one layer? Would it be equivalent if we revert to the invoke with `MemberName`, e.g., using the `ResolvedMethodHandleCallTargetNode`?

Second, do you recall how the `PCDESC_is_method_handle_invoke` bit is used? We have observed different bits in C1 or C2 emitted code, i.e., C2 emits `PcDesc(pc=* offset=* bits=2)` for direct virtual method handle call only, while C1 emits `PcDesc(pc=* offset=* bits=2)` for all kinds of method handle call. 

Many thanks!

PR #1684, opened 2019-09-16
[truffle] Improve error message for guard negations

fniephaus:
/cc @chumer


chumer:
Thanks for the suggestion.
Unfortunately, the change is incorrect. Just specifying a generic specialization is enough, no replacement needed. I'd change it mention replacement in addition to declaration. Ok with you?

fniephaus:
Sure, I haven't found a case where a more generic specialization helped fix the problem. But if that is sometimes the case, I'm for adding a hint that a replacement might help as well. Thanks!

chumer:
Here you go:
```
@SuppressWarnings("unused")
    public abstract static class GenericOverrideTest extends Node {

        public abstract String execute(Object n);

        @Specialization(guards = "n == cachedN", limit = "3")
        protected String s0(int n, @Cached("n") int cachedN) {
            return "s0";
        }

        @Specialization
        protected String s1(int n) {
            return "s1";
        }

        @Fallback
        protected String f0(Object n) {
            return "f0";
        }
    }
```
But it is true strictly more generic cannot always be computed, especially with expression guards.

PR #1688, opened 2019-09-17
Unaligned bulk zeroing on aarch64

mur47x111:


compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
```diff
@@ -48,136 +49,178 @@
...
-            masm.subs(64, size, size, zvaLength);
-            masm.branchConditionally(AArch64Assembler.ConditionFlag.LT, tail);
-            masm.dc(AArch64Assembler.DataCacheOperationType.ZVA, base);
-            masm.add(64, base, base, zvaLength);
-            masm.jmp(zvaLoop);
-
-            // Handle bytes after ZVA loop.
-            masm.bind(tail);
-            masm.add(64, size, size, zvaLength);
+        try (AArch64MacroAssembler.ScratchRegister scratchRegister = masm.getScratchRegister()) {
```
mur47x111: @pli-arm could you please review this patch?

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
```diff
@@ -48,136 +49,178 @@
 public final class AArch64ZeroMemoryOp extends AArch64LIRInstruction {
     public static final LIRInstructionClass<AArch64ZeroMemoryOp> TYPE = LIRInstructionClass.create(AArch64ZeroMemoryOp.class);
 
-    @Use({REG}) protected AllocatableValue addressValue;
-    @Use({REG}) protected AllocatableValue lengthValue;
+    @Use({REG}) protected Value addressValue;
+    @Use({REG}) protected Value lengthValue;
 
+    @Temp({REG}) protected Value addressValueTemp;
```
pli-arm: Are these 2 temps redundant? I don't see they are used.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
```diff
@@ -48,136 +49,178 @@
...
+    @Temp({REG}) protected Value lengthValueTemp;
+
+    private final boolean isUnaligned;
```
pli-arm: Personally, I'd like the boolean variable to have a positive name (i.e use isAligned instead of isUnaligned) for better code readability.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
```diff
@@ -48,136 +49,178 @@
...
+
+            if (useDcZva && zvaLength > 0) {
+                // From ARMv8-A architecture reference manual D12.2.35 Data Cache Zero ID register:
+                // A valid ZVA length should be a power-of-2 value in [4, 2048]
+                assert (CodeUtil.isPowerOf2(zvaLength) && 4 <= zvaLength && zvaLength <= 2048);
+
+                masm.neg(64, alignmentBits, base);
+                masm.and(64, alignmentBits, alignmentBits, zvaLength - 1);
+
+                // Is size less than number of bytes to be pre-zeroed? Jump to post loop if so.
```
pli-arm: should be "post check", I think

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
```diff
@@ -48,136 +49,178 @@
...
+                Label postCheck = new Label();
+                masm.cmp(64, size, alignmentBits);
+                masm.branchConditionally(AArch64Assembler.ConditionFlag.LE, postCheck);
+                masm.sub(64, size, size, alignmentBits);
+
+                // Pre loop: align base according to the supported bulk zeroing stride.
+                Label preCheck = new Label();
```
pli-arm: For better code readability, I prefer the "new Label()" statements to be put altogether at the beginning of current scope (not interleaving with the masm logic).

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
```diff
@@ -48,136 +49,178 @@
...
+            masm.bind(tail);
+            if (isUnaligned) {
+                masm.cbz(64, size, done);
+                // We have to ensure size > 0 when entering the following loop
+                masm.align(crb.target.wordSize * 2);
+                Label perByteZeroingLoop = new Label();
+                masm.bind(perByteZeroingLoop);
+                masm.str(8, zr, AArch64Address.createPostIndexedImmediateAddress(base, 1));
+                masm.subs(64, size, size, 1);
+                masm.branchConditionally(AArch64Assembler.ConditionFlag.NE, perByteZeroingLoop);
```
pli-arm: The NE should be GT, I guess. (Jump back to the loop if we have remaining bytes to zero.)

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
mur47x111: This is intended. We use Temp annotation to mark these two allocatable values to be killed. Otherwise they may be reused in subsequent LIRs and could lead to segfault.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
```diff
@@ -180,8 +180,11 @@ protected void emitCode(CompilationResultBuilder crb, AArch64MacroAssembler masm
                     masm.add(64, size, size, 8);
                 }
             } else {
-                masm.cmp(64, size, 8);
-                masm.branchConditionally(ConditionFlag.LT, tail);
+                if (isUnaligned) {
+                    // After aligning base, we may have size less than 8. Need to check again.
+                    masm.cmp(64, size, 8);
+                    masm.branchConditionally(ConditionFlag.LT, tail);
```
pli-arm: The condition should be LE, I think.

In this else block, if "size" here is exactly 8, and base is not 16-byte aligned, we firstly store 8 bytes and subtract the size by 8. Then the size becomes 0, and the following code would do a STP (at line 198) without a check.

Please double check this again.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
mur47x111: I thought NE check here means size not zero, which should be equivalent to GT given that size>0 when entering this loop.

pli-arm: Yeah, you're right!

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ZeroMemoryOp.java
mur47x111: Good catch! This is for size<8 and we then perform per-byte zeroing. I think the actual fix to add `masm.jmp(mainCheck);` after storing 8 byte, i.e. L192.

mur47x111:
I have updated `AArch64ZeroMemoryOp`. Let me know if there is other potential issues.

pli-arm:
I don't find other issues. Thanks!

compiler/src/org.graalvm.compiler.hotspot.aarch64/src/org/graalvm/compiler/hotspot/aarch64/AArch64HotSpotRegisterAllocationConfig.java
```diff
@@ -101,7 +105,7 @@
         r0,  r1,  r2,  r3,  r4,  r5,  r6,  r7,
         r8,  r9,  r10, r11, r12, r13, r14, r15,
         r16, r17, r18, r19, r20, r21, r22, r23,
-        r24, r25, r26, /* r27, */ r28, /* r29, r30, r31 */
+        r24, r25, r26, r27, r28, r29, r30, r31,
```
mur47x111: @pli-arm @gilles-duboscq https://github.com/oracle/graal/issues/1704 is likely due to this change. Thought we have excluded them at `AArch64HotSpotRegisterConfig.initAllocatable`

mur47x111: Somehow `r27`, i.e., `heapBaseRegister` is used even when compressed oops is off.

PR #1691, opened 2019-09-18
fix bugs related to reading through incomplete phis during parsing (#1690)

dougxc:


kuaiwei:
BytecodeParser and FrameStateBuilder also use originalValue to check unbalanced monitor. Do we need fix these code?

dougxc:
Yes, we're also looking at those and will add further commits to this PR.

kuaiwei:
Hi Doug,
  I'm wondering when this patch can be merged into graal. We hit this issue in our production environment. I'm wondering when this patch can be merged.

Thanks

dougxc:
Hi @kuaiwei , it's currently going through internal review. The patch has become a little more complex to deal with all the uses of `GraphUtil.originalValue` correctly. I hope to merge it in the next 2 days.

dougxc:
Resolved by https://github.com/oracle/graal/commit/56e35c225709cb2564133eacf3d17bdd1f95ad90

PR #1692, opened 2019-09-18
Autovectorization

usrinivasan:
# Graal Autovectorization

Here's our first prototype for Autovectorization in Graal, as mentioned in our email to the [graal-dev mailing list in July](https://mail.openjdk.java.net/pipermail/graal-dev/2019-July/005858.html).

The feature can be enabled using the `-Dgraal.Autovectorize=true` flag. It's disabled by default.

## Current status

- Implementation of the [SLP algorithm](http://groups.csail.mit.edu/cag/slp/SLP-PLDI-2000.pdf).
- Operand packing and extraction.
- Code generation for AMD64.
- Not limited to loops.
	  The current implementation vectorizes everything to stress-test the implementation.
- Vector stamps.
- We're passing all `mx unittest` tests.
- Scimark 2.0 (but it's currently significantly slower, due to loop unrolling limitations & lack of a heuristic).
- Support for arbitrary vector lengths (within the vector register limit).

## Caveats

- Unrolling/ bounds elimination
	The CE `LoopPartialUnrollPhase` does not unroll loops containing array accesses due to guard lowering. We've explored several alternatives to this, like a separate unroll phase that runs before lowering, but this is also not ideal. Right now we have chosen to use one `LoopPartialUnrollPhase` so the implementation currently does not leverage SLP in loops.
- Heuristic
	We need to develop a heuristic to restrict vectorization to cases where it is beneficial. Right now we vectorize *everything* and performance takes a hit as a result. Heuristics that we need are: A loop unrolling heuristic, pair packing savings heuristic, pack filtering heuristic.

There are a few areas where we're unsure of the approach to take, as we don't want to break anything on the EE side. We'll explicitly call out areas in the patch in places where we think there might be a better approach.

@nvgrw @ddegazio @usrinivasan @christhalinger


compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
```diff
@@ -1162,6 +1335,23 @@ public Variable emitLoad(LIRKind kind, Value address, LIRFrameState state) {
         return result;
     }
 
+    @Override
+    public Variable emitVectorLoad(LIRKind vectorKind, int count, Value address, LIRFrameState state) {
+        AMD64AddressValue loadAddress = getAMD64LIRGen().asAddressValue(address);
+        Variable result = getLIRGen().newVariable(vectorKind);
+        // getLIRGen().append(new AMD64Unary.VectorReadMemory(result, loadAddress));
+        // Use the line below instead of the line above for the temporary-stack-space load op.
+        getLIRGen().append(new AMD64Packing.LoadStackOp(getLIRGen(), asAllocatable(result), loadAddress, count));
```
nvgrw: Currently we go via the stack to support arbitrary-size vectors. These could be slow and there might be a better approach.

compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
```diff
@@ -1162,6 +1335,23 @@ public Variable emitLoad(LIRKind kind, Value address, LIRFrameState state) {
...
+        return result;
+    }
+
+    @Override
+    public void emitVectorStore(LIRKind kind, int count, Value address, Value value, LIRFrameState state) {
+        AMD64AddressValue storeAddress = getAMD64LIRGen().asAddressValue(address);
+        // getLIRGen().append(new AMD64Unary.VectorWriteMemory(storeAddress, asAllocatable(value)));
+        getLIRGen().append(new AMD64Packing.StoreStackOp(getLIRGen(), storeAddress, asAllocatable(value), count));
```
nvgrw: Currently we go via the stack to support arbitrary-size vectors. These could be slow and there might be a better approach.

compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/VectorDescription.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public final int maxVectorWidth(Stamp stamp) {
+        if (stamp instanceof PrimitiveStamp) {
+            return maxVectorWidth((PrimitiveStamp) stamp);
+        }
+
+        return 1;
+    }
+
+    protected abstract int maxVectorWidth(PrimitiveStamp stamp);
```
nvgrw: We've created this class + method to determine the platform vector length. There could be a better way.

compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/VectorPrimitiveStamp.java
```diff
@@ -0,0 +1,117 @@
...
+import org.graalvm.compiler.debug.GraalError;
+
+import jdk.vm.ci.meta.Constant;
+import jdk.vm.ci.meta.JavaKind;
+import jdk.vm.ci.meta.MemoryAccessProvider;
+import jdk.vm.ci.meta.MetaAccessProvider;
+import jdk.vm.ci.meta.ResolvedJavaType;
+import jdk.vm.ci.meta.SerializableConstant;
+
+public abstract class VectorPrimitiveStamp extends ArithmeticStamp {
```
nvgrw: We've added new stamps that represent vector types to re-use many parts of the code generator. Since these changes are fairly invasive this may provide some challenges when integrating with EE

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/phases/LowTier.java
```diff
@@ -81,10 +99,30 @@ public LowTier(OptionValues options) {
...
         appendPhase(new UseTrappingNullChecksPhase());
 
+        if (Options.Autovectorize.getValue(options)) {
+            final String listValue = Options.AVList.getValue(options).trim();
+            List<String> list = new ArrayList<>();
+            if (!listValue.isEmpty()) {
+                list = Arrays.asList(listValue.split(","));
+            }
+
+            appendPhase(new IsomorphicPackingPhase(
```
nvgrw: IPP currently runs in the low tier in part to avoid having to make changes to other phases to support the vector stamp types & nodes. Interested to get your opinion on this approach.

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/vector/AMD64Packing.java
```diff
@@ -0,0 +1,507 @@
...
+    private static void move(CompilationResultBuilder crb, AMD64MacroAssembler masm, AMD64Address dst, Value src, AMD64Kind srcKind, Register scratch, AMD64Address backupSlot) {
+        if (isRegister(src)) {
+            reg2addr(crb, masm, srcKind, dst, asRegister(src));
+        } else if (isStackSlot(src)) {
+            // We move first to the scratch register, then to our destination.
+            // TODO: Is there a faster solution?
+            reg2addr(crb, masm, srcKind, backupSlot, scratch);
+            addr2reg(crb, masm, srcKind, scratch, (AMD64Address) crb.asAddress(src));
+            reg2addr(crb, masm, srcKind, dst, scratch);
+            addr2reg(crb, masm, srcKind, scratch, backupSlot);
```
nvgrw: Is there a better solution for this?

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/vector/AMD64Packing.java
```diff
@@ -0,0 +1,507 @@
...
+                final AMD64Address target = displace(packSpaceAddress, stackOffset);
+                move(crb, masm, target, value, scalarKind, asRegister(scratch), backupSlotAddress);
+
+                // Increment address offset so that we place the next value further into the
+                // temporary space.
+                stackOffset += scalarKind.getSizeInBytes();
+            }
+
+            // Write memory into vector register.
+            masm.vmovdqu(asRegister(result), packSpaceAddress);
```
nvgrw: Here (and throughout) we may wish to use an aligned instruction

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/vector/AMD64VectorShuffle.java
```diff
@@ -322,7 +333,12 @@ public ExtractByteOp(AllocatableValue result, AllocatableValue vector, int selec
 
         @Override
         public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
-            VPEXTRB.emit(masm, XMM, asRegister(result), asRegister(vector), selector);
+            if (isRegister(result)) {
```
nvgrw: These operations are not used in CE but may have an impact on EE

compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/vector/AMD64VectorShuffle.java
```diff
@@ -343,7 +359,12 @@ public ExtractShortOp(AllocatableValue result, AllocatableValue vector, int sele
 
         @Override
         public void emitCode(CompilationResultBuilder crb, AMD64MacroAssembler masm) {
-            VPEXTRW.emit(masm, XMM, asRegister(result), asRegister(vector), selector);
+            if (isRegister(result)) {
```
nvgrw: These operations are not used in CE but may have an impact on EE

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPartialUnrollPhase.java
```diff
@@ -64,7 +65,7 @@ protected void run(StructuredGraph graph, CoreProviders context) {
                         if (!LoopTransformations.isUnrollableLoop(loop)) {
                             continue;
                         }
-                        if (getPolicies().shouldPartiallyUnroll(loop)) {
+                        if (context instanceof MidTierContext && getPolicies().shouldPartiallyUnroll(loop, ((MidTierContext) context).getVectorDescription())) {
```
nvgrw: There ought to be a better way to pass this contextual data

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopTransformations.java
```diff
@@ -239,6 +239,18 @@ public static LoopBeginNode insertPrePostLoops(LoopEx loop) {
...
+        // The duplication above unmarks certain nodes in the loop's LoopFragmentWhole, but not in
+        // the loop's LoopFragmentInside (despite the LoopFragmentInside deriving from the
+        // LoopFragmentWhole).
+        //
+        // Invalidate the loop's LoopFragmentInside here, so that when it is used later it is
+        // consistent with the LoopFragmentWhole.
+        // `processPreLoopPhis` utilizes loop.inside() and asserts that have been duplicated.
+        // However, since the duplication above removes some nodes from the LoopFragmentWhole and
+        // therefore does not duplicate those nodes, the assertion fails as we encounter nodes that were not duplicated.
+        loop.invalidateInsideFragment();
```
nvgrw: This was a band-aid solution for an unrolling-related problem. Should probably be removed once unrolling is fixed.

compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/VectorizationLoopPolicies.java
```diff
@@ -0,0 +1,127 @@
...
+import org.graalvm.compiler.nodes.ValueNode;
+import org.graalvm.compiler.nodes.cfg.ControlFlowGraph;
+import org.graalvm.compiler.nodes.debug.ControlFlowAnchorNode;
+import org.graalvm.compiler.nodes.memory.Access;
+import org.graalvm.compiler.nodes.memory.WriteNode;
+import org.graalvm.compiler.options.OptionValues;
+
+import jdk.vm.ci.meta.MetaAccessProvider;
+
+public class VectorizationLoopPolicies implements LoopPolicies {
```
nvgrw: Policies to use for first unroll pass (now removed). If unrolling before guard lowering then this can be used. Otherwise this could be integrated with the default policy.

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/vectorization/DefaultAutovectorizationPolicies.java
```diff
@@ -0,0 +1,136 @@
...
+import org.graalvm.compiler.graph.Node;
+import org.graalvm.compiler.nodes.ValueNode;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class DefaultAutovectorizationPolicies implements AutovectorizationPolicies {
```
nvgrw: These are the heuristics that need some work

usrinivasan:
> * Hello Niklas Vangerow, thanks for contributing a PR to our project!
> 
> We use the [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) to make the copyright of contributions clear. We don't have a record of you having signed this yet, based on your email address nvangerow -(at)- twitter -(dot)- com. You can sign it at that link.
> 
> 
> * Hello David Degazio, thanks for contributing a PR to our project!
> 
> We use the [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) to make the copyright of contributions clear. We don't have a record of you having signed this yet, based on your email address ddegazio -(at)- twitter -(dot)- com. You can sign it at that link.
> 
> 
> * Hello Uma Srinivasan, thanks for contributing a PR to our project!
> 
> We use the [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) to make the copyright of contributions clear. We don't have a record of you having signed this yet, based on your email address usrinivasan -(at)- twitter -(dot)- com. You can sign it at that link.
> 

All the authors worked for Twitter Inc. at the time of contribution to this code. Twitter has signed the OCA.

compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/VectorPrimitiveStamp.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public String toString() {
+        final StringBuilder str = new StringBuilder();
+        if (hasValues()) {
+            str.append("v");
+            str.append(getElementCount());
+            str.append(" of ");
+            str.append(getScalar().toString());
+        } else {
+            str.append("<empty>");
```
mauhiz: let's return this constant without string building if `!hasValues()` ?

thomaswue:
Can you share the impact on compile time, code size and peak performance on the workloads as suggested by https://mail.openjdk.java.net/pipermail/graal-dev/2019-July/005862.html when the patch was initially proposed? Specifically interesting for us would be the measured impact for Twitter's services that run the GraalVM compiler in production.

From the summary above it seems there is a negative impact even on the Scimark benchmark. This is one of the workloads that should certainly profit from vectorization.

plokhotnyuk:
@thomaswue can I use [standard instructions](https://github.com/oracle/graal/blob/master/compiler/README.md) to build binaries for this PR?

I'm going to run [this](https://plokhotnyuk.github.io/jsoniter-scala/) benchmark for them with both values of the `-Dgraal.Autovectorize` option.

thomaswue:
@plokhotnyuk I would assume that the PR is at least in a state where you should be able to do that.

usrinivasan:
> @plokhotnyuk I would assume that the PR is at least in a state where you should be able to do that.

That's right! We use `mx build` to build the graal jar and use it with a jdk11 vm to run tests. Please let us know if you run into any issues.

helloguo:
Thanks for sharing the patch! Any update in terms of the next steps? (e.g. the fix of performance regression for Scimark benchmark, the plan of merging this PR into master). 

I ran a few workloads which are supposed to benefit from autovec. However, I did not see too much difference when I applied this PR. It's possible that the loops in my workload are not in a good shape. I guess my question is what would be good scenarios to use this patch? Or when should I not use this patch because of its current limitation?

thomaswue:
We discussed this briefly at the community workshop. There are a lot of possible interactions of vectorization with other loop transformations in the compiler due to different loop shapes. We do have support for auto-vectorization in our EE version. We will work towards having basic support for this feature also in CE.

helloguo:
@thomaswue Thanks for the information.

> We do have support for auto-vectorization in our EE version. We will work towards having basic support for this feature also in CE.

Any idea about the timeline for CE to have basic autovec support? Just curious when is a good time to come back and test it again on our workloads.

thomaswue:
@helloguo You should certainly test the EE version on your workloads ;). It is free for evaluation and I assume it could make a difference at Facebook. Note that auto-vectorization is only a small part of the CE/EE performance-related differences.

Productization of this feature in CE will require quite some engineering effort. The current plan is to have it in one of the 20.x releases next year. But take this as a rough estimate only.

apete:
I've done some linear algebra benchmarks comparing the EE and CE versions, using JMH. Here are the results doing matrix multiplication using Apache Commons Math (ACM), EJML and ojAlgo.

FillByMultiplying

Throughput ops/min on the y-axis, and matrix size (square) on the x-axis.

The results vary depending on the library, but the CE is typically slower than HotSpot and the EE typically faster. The difference between CE and EE can be significant.

christhalinger:
@thomaswue thanks for the update on a possible timeline!

gergo-:
I will close this pull request as it hasn't seen activity in years, and we have now added loop vectorization in the GraalVM compiler: https://github.com/oracle/graal/pull/14105

PR #1703, opened 2019-09-26
Enable Math.fma intrinsics for AArch64

pli-arm:
Math.fma returns the exact product of the first two arguments summed with the third argument and then rounded once. This behaviour has no difference with that of the AArch64 FMADD instruction. So this patch enables Math.fma intrinsics if VM option "UseFMA" is set and the JDK version is greater than 8.


thomaswue:
Thank you very much for this patch. @sanzinger will take care of final review and integration.

sanzinger:
Integration is in progress.

PR #1728, opened 2019-10-09
Ignore missing static libs if -H:+ExitAfterRelocatableFileWrite is set.

### LINKED ISSUE #1727: Allow developers to override location of static jdk libs
When there are no static JDK libs available, or when the developer prefers to use his own static libs (e.g. with different versions, or for different platforms), there is currently no option to override the default.

johanvos:
This fixes #1727

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageOptions.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageOptions.java
@@ -198,5 +198,4 @@ public class NativeImageOptions {
         }
         return analysisThreads;
     }
-
 }
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageOptions.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageOptions.java
@@ -198,4 +198,7 @@ public class NativeImageOptions {
         }
         return analysisThreads;
     }
+
+    @Option(help = "Specify a directory that contains static JDK libs (e.g. libjava.a etc)")//
+    public static final HostedOptionKey<String> StaticJDKLibs = new HostedOptionKey<>("");
 }
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageOptions.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageOptions.java
@@ -199,6 +199,4 @@ public class NativeImageOptions {
         return analysisThreads;
     }
 
-    @Option(help = "Specify a directory that contains static JDK libs (e.g. libjava.a etc)")//
-    public static final HostedOptionKey<String> StaticJDKLibs = new HostedOptionKey<>("");
 }
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java
@@ -41,6 +41,7 @@ import java.util.Map;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import com.oracle.svm.hosted.NativeImageOptions;
 import org.graalvm.compiler.api.replacements.SnippetReflectionProvider;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.Platform;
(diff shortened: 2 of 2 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageOptions.java
```diff
@@ -198,4 +198,7 @@ public static int getMaximumNumberOfAnalysisThreads(OptionValues optionValues) {
         }
         return analysisThreads;
     }
+
+    @Option(help = "Specify a directory that contains static JDK libs (e.g. libjava.a etc)")//
+    public static final HostedOptionKey<String> StaticJDKLibs = new HostedOptionKey<>("");
```
vjovanov: You can make this null by default.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java
```diff
@@ -165,6 +166,13 @@ public TargetDescription getTarget() {
         Path staticLibsDir = null;
         String hint = null;
 
+        /* In case a specific location for the static JDK libraries is provided, use that. */
+
+        String externalStaticLibsDir = NativeImageOptions.StaticJDKLibs.getValue();
```
vjovanov: And use `hasBeenSet` here.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java
```diff
@@ -165,6 +166,13 @@ public TargetDescription getTarget() {
         Path staticLibsDir = null;
         String hint = null;
 
+        /* In case a specific location for the static JDK libraries is provided, use that. */
+
```
vjovanov: Extra space.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java
```diff
@@ -165,6 +166,13 @@ public TargetDescription getTarget() {
...
+        String externalStaticLibsDir = NativeImageOptions.StaticJDKLibs.getValue();
+        if (!externalStaticLibsDir.isEmpty()) {
+            staticLibsDir = Paths.get(externalStaticLibsDir);
```
vjovanov: And finally, use `getValue` here.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageOptions.java
```diff
@@ -198,4 +198,7 @@ public static int getMaximumNumberOfAnalysisThreads(OptionValues optionValues) {
         }
         return analysisThreads;
     }
+
+    @Option(help = "Specify a directory that contains static JDK libs (e.g. libjava.a etc)")//
```
vjovanov: libs -> libraries

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageOptions.java
vjovanov: I would also use `Libraries` as we already have it with `CLibraries`.

johanvos:
> Instead of adding yet another hosted option I suggest we add checking if the specified com.oracle.svm.core.SubstrateOptions#CLibraryPath already contains the static libs. That means just run the same probing for all elements of `OptionUtils.flatten(",", SubstrateOptions.CLibraryPath.getValue())` in
> 
> https://github.com/oracle/graal/blob/519cb180a31c61f7735d64aeabf566236489ddcf/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java#L184

There are actually 2 aspects to this:

1. In case -H:+ExitAfterRelocatableImageWrite is specified, you don't care about the static SDK libs at all. In that case, they won't be in CLibraryPath either. But this can be better dealt with by checking the ExitAfterRelocatableImageWrite first, and returning if that is set.

2. In case you are cross-compiling, you don't want to use the static libraries that you use at compiletime. In case the CLibraryPath is not used for that purpose (e.g. it does not contain native code relevant at compiletime), we can indeed reuse that option.

olpaw:
> But this can be better dealt with by checking the ExitAfterRelocatableImageWrite first, and returning if that is set.

Yes that makes sense.

>     2. In case you are cross-compiling, you don't want to use the static libraries that you use at compiletime. In case the CLibraryPath is not used for that purpose (e.g. it does not contain native code relevant at compiletime), we can indeed reuse that option.

That is the case. None of the libs that are on `CLibraryPath` are _executed_ at image-build time. They are only used for _linking_ at image-build time. I.e. In the case of cross-compiling to - let's say - aarch64 on an amd64 host you would have aarch64 libraries on `CLibraryPath`.

johanvos:
I modified the PR to satisfy goal #1 (don't bail if there are no static JDK libs found, but ExitAfterRelocatableImageWrite is set).
I propose to tackle goal #2 (allow to override the location) in a different PR, at a later moment.

PR #1734, opened 2019-10-10
Fail truffle+sdk gates if checkcopyrights fails

fniephaus:
`mx.checkcopyrights(['--primary'])` does not fail the gate if the check fails, so valid its return value in the same way Sulong does:
https://github.com/oracle/graal/blob/832cee40053360183c4b15a443a552142c15012e/sulong/mx.sulong/mx_sulong.py#L167-L168


chumer:
We deliberately don't use that tool in the SDK gate. We check the copyright in Java files using checkstyle.
The checkcopyrights is only there to check and update version numbers that we update from time to time. But we do not require the year to be correct in the gate to save us from gate failures just because of that. That tool is still handy to generate the versions from time to time.

fniephaus:
Sure, just wanted to make sure this is not an oversight.

PR #1768, opened 2019-10-23
[native-image] Increase array base mask for huge hybird Class object

### LINKED ISSUE #1766: [SVM] LayoutEncoding.getArrayBaseOffset() causes assertion failure
We found the implementation of [layout encoding and decoding](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/LayoutEncoding.java) is asymmetric:
```java
// encoding
public static int forArray(boolean isObject, int arrayBaseOffset, int arrayIndexShift) {
    int tag = isObject ? ARRAY_TAG_OBJECT_VALUE : ARRAY_TAG_PRIMITIVE_VALUE;
    int encoding = (tag << ARRAY_TAG_SHIFT) | (arrayBaseOffset << ARRAY_BASE_SHIFT) | (arrayIndexShift << ARRAY_INDEX_SHIFT_SHIFT);
    // assertions
    assert !isInstance(encoding) && isArray(encoding);
    assert isObjectArray(encoding) == isObject;
    assert isPrimitiveArray(encoding) != isObject;
    assert getArrayBaseOffset(encoding).equal(WordFactory.unsigned(arrayBaseOffset));
    ^^^^^^^^^^^^^^^^^^^^^^^^^^ throw AssertionError
    assert getArrayIndexShift(encoding) == arrayIndexShift;
    return encoding;
}
// decoding
public static UnsignedWord getArrayBaseOffset(int encoding) {
    return WordFactory.unsigned((encoding >> ARRAY_BASE_SHIFT) & ARRAY_BASE_MASK);
    ^^^^^^^^^^^^^^^^
}
```
In encoding method, `arrayBaseOffset` has 22 bits which range from 8th to 29th inclusively, but in decoding method, `arrayBaseOffset` only contains 8 bits because of array base mask (255), this
would cause a potential problem if array base offset excesses 255, we can reproduce this scenario by the following case:
```python
import sys

TMPL = "public class C{} implements I{}, I{}, I{} {{}}"
TMPL2 = "public interface I{} {{}}"

if __name__ == "__main__":

    N = 1000
    for i in range(0,N):
        with open('C{}.java'.format(i), 'w') as f:
            f.write(TMPL.format(i,i,((i + 3) % N), ((i+5) % N)))
        with open('I{}.java'.format(i), 'w') as f:
            f.write(TMPL2.format(i))

    with open('Main.java', 'w') as f:
        f.write("public class Main {")
        f.write("public static void main(String[] args) {")
        f.write("int hash = 0;");
        for i in range(0,N):
            f.write("C{} c{} = new C{}();".format(i, i, i));
            f.write("hash ^= c{}.hashCode();".format(i));
        f.write("System.exit(hash);}}")

```
Commands:
```bash
python gen.py
javac Main.java
native-image -J-ea Main
```
This script will generate many interfaces and classes, and lead to big interface-bits.
​native-image will throw an AssertError due to the previous problem. Is this a bug or an intentional design of native-image?
Can we change the ARRAY_BASE_MASK from 255 to  (1 << (Integer.SIZE - ARRAY_TAG_BITS - ARRAY_BASE_SHIFT)) - 1 ？

kelthuzadx:
Fix #1766

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/LayoutEncoding.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/LayoutEncoding.java
@@ -47,8 +47,8 @@ public class LayoutEncoding {
     private static final int ARRAY_INDEX_SHIFT_SHIFT = 0;
     private static final int ARRAY_INDEX_SHIFT_MASK = 255;
     private static final int ARRAY_BASE_SHIFT = 8 + ARRAY_INDEX_SHIFT_SHIFT;
-    private static final int ARRAY_BASE_MASK = 255;
     private static final int ARRAY_TAG_BITS = 2;
+    private static final int ARRAY_BASE_MASK = (1 << (Integer.SIZE - ARRAY_TAG_BITS - ARRAY_BASE_SHIFT)) - 1;
     private static final int ARRAY_TAG_SHIFT = Integer.SIZE - ARRAY_TAG_BITS;
     private static final int ARRAY_TAG_PRIMITIVE_VALUE = ~0x00; // 0xC0000000 >> 30
     private static final int ARRAY_TAG_OBJECT_VALUE = ~0x01; // 0x80000000 >> 30


peter-hofer:
Thank you for the patch @kelthuzadx ! We were already working on a patch that adds extra bits and also makes the assertions into guarantees with better error messages, it has been merged in bbfd63218ccd7f869228508a8a8a86773e59c2e4

PR #1772, opened 2019-10-24
rework handling of volatile accesses on aarch64

rwestrel:
This builds on top of https://github.com/oracle/graal/pull/1642 and implements volatile field accesses on aarch64 without explicit membar instructions.

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
+++ b/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
@@ -46,6 +46,7 @@ import org.graalvm.compiler.lir.aarch64.AArch64AddressValue;
 import org.graalvm.compiler.lir.aarch64.AArch64ArithmeticLIRGeneratorTool;
 import org.graalvm.compiler.lir.aarch64.AArch64ArithmeticOp;
 import org.graalvm.compiler.lir.aarch64.AArch64BitManipulationOp;
+import org.graalvm.compiler.lir.aarch64.AArch64Move;
 import org.graalvm.compiler.lir.aarch64.AArch64Move.LoadOp;
 import org.graalvm.compiler.lir.aarch64.AArch64Move.StoreConstantOp;
 import org.graalvm.compiler.lir.aarch64.AArch64Move.StoreOp;
@@ -507,6 +508,14 @@ public class AArch64ArithmeticLIRGenerator extends ArithmeticLIRGenerator implem
     }
 
     @Override
+    public Variable emitVolatileLoad(LIRKind kind, Value address, LIRFrameState state) {
+        AllocatableValue loadAddress = asAllocatable(address);
+        Variable result = getLIRGen().newVariable(getLIRGen().toRegisterKind(kind));
+        getLIRGen().append(new AArch64Move.VolatileLoadOp((AArch64Kind) kind.getPlatformKind(), result, loadAddress, state));
+        return result;
+    }
+
+    @Override
     public void emitStore(ValueKind<?> lirKind, Value address, Value inputVal, LIRFrameState state) {
         AArch64AddressValue storeAddress = getLIRGen().asAddressValue(address);
         AArch64Kind kind = (AArch64Kind) lirKind.getPlatformKind();
@@ -521,6 +530,14 @@ public class AArch64ArithmeticLIRGenerator extends ArithmeticLIRGenerator implem
         }
         AllocatableValue input = asAllocatable(inputVal);
         getLIRGen().append(new StoreOp(kind, storeAddress, input, state));
+    }
+
+    @Override
+    public void emitVolatileStore(ValueKind<?> lirKind, Value addressVal, Value inputVal, LIRFrameState state) {
+        AArch64Kind kind = (AArch64Kind) lirKind.getPlatformKind();
+        AllocatableValue input = asAllocatable(inputVal);
+        AllocatableValue address = asAllocatable(addressVal);
+        getLIRGen().append(new AArch64Move.VolatileStoreOp(kind, address, input, state));
     }
 
     @Override
--- a/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/ArithmeticLIRGeneratorTool.java
+++ b/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/ArithmeticLIRGeneratorTool.java
@@ -100,7 +100,11 @@ public interface ArithmeticLIRGeneratorTool {
 
     Variable emitLoad(LIRKind kind, Value address, LIRFrameState state);
 
+    Variable emitVolatileLoad(LIRKind kind, Value address, LIRFrameState state);
+
     void emitStore(ValueKind<?> kind, Value address, Value input, LIRFrameState state);
+
+    void emitVolatileStore(ValueKind<?> kind, Value address, Value input, LIRFrameState state);
 
     @SuppressWarnings("unused")
     default Value emitFusedMultiplyAdd(Value a, Value b, Value c) {
--- a/compiler/src/org.graalvm.compiler.asm.aarch64.test/src/org/graalvm/compiler/asm/aarch64/test/TestProtectedAssembler.java
+++ b/compiler/src/org.graalvm.compiler.asm.aarch64.test/src/org/graalvm/compiler/asm/aarch64/test/TestProtectedAssembler.java
@@ -133,16 +133,6 @@ class TestProtectedAssembler extends AArch64Assembler {
     }
 
     @Override
-    protected void ldar(int size, Register rt, Register rn) {
-        super.ldar(size, rt, rn);
-    }
-
-    @Override
-    protected void stlr(int size, Register rt, Register rn) {
-        super.stlr(size, rt, rn);
-    }
-
-    @Override
     public void ldaxr(int size, Register rt, Register rn) {
         super.ldaxr(size, rt, rn);
     }
--- a/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
+++ b/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
@@ -1402,7 +1402,7 @@ public abstract class AArch64Assembler extends Assembler {
      * @param rt general purpose register. May not be null or stackpointer.
      * @param rn general purpose register.
      */
-    protected void ldar(int size, Register rt, Register rn) {
+    public void ldar(int size, Register rt, Register rn) {
         assert size == 8 || size == 16 || size == 32 || size == 64;
         int transferSize = NumUtil.log2Ceil(size / 8);
         exclusiveLoadInstruction(LDAR, rt, rn, transferSize);
@@ -1415,7 +1415,7 @@ public abstract class AArch64Assembler extends Assembler {
      * @param rt general purpose register. May not be null or stackpointer.
      * @param rn general purpose register.
      */
-    protected void stlr(int size, Register rt, Register rn) {
+    public void stlr(int size, Register rt, Register rn) {
         assert size == 8 || size == 16 || size == 32 || size == 64;
         int transferSize = NumUtil.log2Ceil(size / 8);
         // Hack: Passing the zero-register means it is ignored when building the encoding.
@@ -1471,7 +1471,7 @@ public abstract class AArch64Assembler extends Assembler {
      */
     private void exclusiveStoreInstruction(Instruction instr, Register rs, Register rt, Register rn, int log2TransferSize) {
         assert log2TransferSize >= 0 && log2TransferSize < 4;
-        assert rt.getRegisterCategory().equals(CPU) && rs.getRegisterCategory().equals(CPU) && !rs.equals(rt);
+        assert rt.getRegisterCategory().equals(CPU) && rs.getRegisterCategory().equals(CPU) && (instr == STLR || !rs.equals(rt));
         int transferSizeEncoding = log2TransferSize << LoadStoreTransferSizeOffset;
         emitInt(transferSizeEncoding | instr.encoding | rs2(rs) | rn(rn) | rt(rt));
     }
--- a/compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64VolatileAccessesTest.java
+++ b/compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64VolatileAccessesTest.java
@@ -1,0 +1,194 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, Red Hat Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package org.graalvm.compiler.core.aarch64.test;
+
+import jdk.vm.ci.aarch64.AArch64;
+import jdk.vm.ci.aarch64.AArch64Kind;
+import org.graalvm.compiler.core.test.MatchRuleTest;
+import org.graalvm.compiler.lir.LIRInstruction;
+import org.graalvm.compiler.lir.aarch64.AArch64Move;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.function.Predicate;
+
+import static org.junit.Assume.assumeTrue;
+
+public class AArch64VolatileAccessesTest extends MatchRuleTest {
+    private static volatile byte volatileByteField;
+    private static volatile short volatileShortField;
+    private static volatile int volatileIntField;
+    private static volatile long volatileLongField;
+    private static volatile float volatileFloatField;
+    private static volatile double volatileDoubleField;
+
+    private static Predicate<LIRInstruction> storePredicate(AArch64Kind kind) {
+        return op -> (op instanceof AArch64Move.VolatileStoreOp && ((AArch64Move.VolatileStoreOp) op).getKind() == kind);
+    }
+
+    private static Predicate<LIRInstruction> loadPredicate(AArch64Kind kind) {
+        return op -> (op instanceof AArch64Move.VolatileLoadOp && ((AArch64Move.VolatileLoadOp) op).getKind() == kind);
+    }
+
+    @Before
+    public void checkAArch64() {
+        assumeTrue("skipping AArch64 specific test", getTarget().arch instanceof AArch64);
+    }
+
+    public static byte volatileByteFieldLoad() {
+        return volatileByteField;
+    }
+
+    @Test
+    public void test01() {
+        checkLIR("volatileByteFieldLoad", loadPredicate(AArch64Kind.BYTE), 1);
+        volatileByteField = 42;
+        test("volatileByteFieldLoad");
+    }
+
+    public static short volatileShortFieldLoad() {
+        return volatileShortField;
+    }
+
+    @Test
+    public void test02() {
+        checkLIR("volatileShortFieldLoad", loadPredicate(AArch64Kind.WORD), 1);
+        volatileShortField = 42;
+        test("volatileShortFieldLoad");
+    }
+
+    public static int volatileIntFieldLoad() {
+        return volatileIntField;
+    }
+
+    @Test
+    public void test03() {
+        checkLIR("volatileIntFieldLoad", loadPredicate(AArch64Kind.DWORD), 1);
+        volatileIntField = 42;
+        test("volatileIntFieldLoad");
+    }
+
+    public static long volatileLongFieldLoad() {
+        return volatileLongField;
+    }
+
+    @Test
+    public void test04() {
+        checkLIR("volatileLongFieldLoad", loadPredicate(AArch64Kind.QWORD), 1);
+        volatileLongField = 42;
+        test("volatileLongFieldLoad");
+    }
+
+    public static float volatileFloatFieldLoad() {
+        return volatileFloatField;
+    }
+
+    @Test
+    public void test05() {
+        checkLIR("volatileFloatFieldLoad", loadPredicate(AArch64Kind.SINGLE), 1);
+        volatileFloatField = 42;
+        test("volatileFloatFieldLoad");
+    }
+
+    public static double volatileDoubleFieldLoad() {
+        return volatileDoubleField;
+    }
+
+    @Test
+    public void test06() {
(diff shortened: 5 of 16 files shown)


sanzinger:
FYI: I have started reviewing and internal testing of this PR.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/ArithmeticLIRGeneratorTool.java
```diff
@@ -100,7 +100,9 @@
 
     Variable emitLoad(LIRKind kind, Value address, LIRFrameState state);
 
-    void emitStore(ValueKind<?> kind, Value address, Value input, LIRFrameState state);
+    Variable emitVolatileLoad(LIRKind kind, Value address, LIRFrameState state);
+
+    void emitStore(ValueKind<?> kind, Value address, Value input, LIRFrameState state, boolean isVolatile);
```
sanzinger: I feel a bit uneasy by this API change. The `isVolatile` flag is only relevant for a single platform. But API suggests that this call always emits a volatile store, which is wrong. An assertion in the backends would help here.

I would go for adding an assertion to each platform (`LLVM`, `SPARC` and `AMD64`) that this flag must be false since they never emit volatile stores.

This also requires to replace the `WriteNode` in the `InsertMembarsPhase` with a new node having `volatileAccess==false`. Then we can add an assertion to all these platforms which assumes `isVolatile==false`.

rwestrel: Thanks for reviewing this. Wouldn't you then also have InsertMembarsPhase turn all VolatileReadNode into a regular ReadNode?

sanzinger: Thx. I'll test it.

sanzinger: BootstrapWithSystemAssertions fails with this exception:

```
gate: 06 Nov 2019 20:04:17(+03:04) BEGIN: BootstrapWithSystemAssertionsEconomy
env MX_SUBPROCESS_COMMAND_FILE=/.tmp/mx_subprocess_command.55OjMd MX_PRIMARY_SUITE_PATH=/b/b/e/main/compiler MX__SUITEMODEL=sibling \
/compiler/mxbuild/linux-amd64/graaljdks/jdk1.8-cmp/bin/java -server -XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCI -d64 -Dgraal.PrintGraph=Network -Dgraal.CompilationWatchDogStartDelay=60.0D -esa -Dgraal.DetailedAsserts=true -Dgraal.CompilerConfiguration=economy -Dgraal.CompilationFailureAction=ExitVM -XX:+UseJVMCICompiler -Dgraal.DumpOnError=true -Dgraal.PrintGraphFile=true -XX:-TieredCompilation -XX:+BootstrapJVMCI -version
Bootstrapping JVMCI[Use -Dgraal.LogFile=<path> to redirect Graal log output to a file.]
[thread:11] scope: JVMCI CompilerThread6
    [thread:11] scope: JVMCI CompilerThread6.Compiling.GraalCompiler
    Context: StructuredGraph:130{HotSpotMethod<HotSpotJVMCIRuntime.runtime()>}
          [thread:11] scope: JVMCI CompilerThread6.Compiling.GraalCompiler.BackEnd.EmitLIR.LIRGeneration
          Exception raised in scope JVMCI CompilerThread6.Compiling.GraalCompiler.BackEnd.EmitLIR.LIRGeneration: org.graalvm.compiler.graph.GraalGraphError: org.graalvm.compiler.debug.GraalError: should not reach here
          	at node: 12|VolatileRead
          	at org.graalvm.compiler.graph.GraalGraphError.transformAndAddContext(GraalGraphError.java:136)
          	at org.graalvm.compiler.core.gen.NodeLIRBuilder.doBlock(NodeLIRBuilder.java:378)
          	at org.graalvm.compiler.core.LIRGenerationPhase.emitBlock(LIRGenerationPhase.java:85)
          	at org.graalvm.compiler.core.LIRGenerationPhase.run(LIRGenerationPhase.java:75)
          	at org.graalvm.compiler.core.LIRGenerationPhase.run(LIRGenerationPhase.java:46)
          	at org.graalvm.compiler.lir.phases.LIRPhase.apply(LIRPhase.java:117)
          	at org.graalvm.compiler.lir.phases.LIRPhase.apply(LIRPhase.java:109)
          	at org.graalvm.compiler.core.gen.LIRCompilerBackend.emitLIR0(LIRCompilerBackend.java:155)
          	at org.graalvm.compiler.core.gen.LIRCompilerBackend.emitLIR(LIRCompilerBackend.java:113)
          	at org.graalvm.compiler.core.gen.LIRCompilerBackend.emitBackEnd(LIRCompilerBackend.java:83)
          	at org.graalvm.compiler.core.target.Backend.emitBackEnd(Backend.java:279)
          	at org.graalvm.compiler.core.GraalCompiler.compile(GraalCompiler.java:145)
          	at org.graalvm.compiler.core.GraalCompiler.compileGraph(GraalCompiler.java:129)
          	at org.graalvm.compiler.hotspot.HotSpotGraalCompiler.compileHelper(HotSpotGraalCompiler.java:227)
          	at org.graalvm.compiler.hotspot.HotSpotGraalCompiler.compile(HotSpotGraalCompiler.java:245)
          	at org.graalvm.compiler.hotspot.CompilationTask$HotSpotCompilationWrapper.performCompilation(CompilationTask.java:170)
          	at org.graalvm.compiler.hotspot.CompilationTask$HotSpotCompilationWrapper.performCompilation(CompilationTask.java:84)
          	at org.graalvm.compiler.core.CompilationWrapper.run(CompilationWrapper.java:179)
          	at org.graalvm.compiler.hotspot.CompilationTask.runCompilation(CompilationTask.java:334)
          	at org.graalvm.compiler.hotspot.HotSpotGraalCompiler.compileMethod(HotSpotGraalCompiler.java:144)
          	at org.graalvm.compiler.hotspot.HotSpotGraalCompiler.compileMethod(HotSpotGraalCompiler.java:109)
          	at jdk.vm.ci.hotspot.HotSpotJVMCIRuntime.compileMethod(HotSpotJVMCIRuntime.java:667)
          Caused by: org.graalvm.compiler.debug.GraalError: should not reach here
          	at org.graalvm.compiler.debug.GraalError.shouldNotReachHere(GraalError.java:47)
          	at org.graalvm.compiler.core.amd64.AMD64ArithmeticLIRGenerator.emitVolatileLoad(AMD64ArithmeticLIRGenerator.java:1167)
          	at org.graalvm.compiler.nodes.memory.VolatileReadNode.generate(VolatileReadNode.java:54)
          	at org.graalvm.compiler.core.gen.NodeLIRBuilder.emitNode(NodeLIRBuilder.java:482)
          	at org.graalvm.compiler.core.gen.NodeLIRBuilder.doRoot(NodeLIRBuilder.java:473)
          	at org.graalvm.compiler.core.gen.NodeLIRBuilder.doBlock(NodeLIRBuilder.java:376)
          	... 20 more
```

Commandline: `mx gate -t BootstrapWithSystemAssertionsEconomy`

rwestrel: The economy configuration must be broken then because it doesn't emit membars for volatile accesses at lowering and it doesn't run the InsertMembarsPhase. I see 3 ways to fix this:

- run the economy configuration with LateMembars to false. AFAIU, we would not be able to run code compiled with the economy configuration and the regular configuration in the same process on aarch64 because correctness requires that we uniformly use either membars or acquire/release instructions. (maybe @adinn can confirm that)

- Add InsertMembarsPhase to the economy configuration but that doesn't seem like a good fit for an economy configuration

- Have all platform support volatile accesses at the LIR level. So x86 would emit a regular access instruction surrounded by whatever membar instructions a required.

adinn: > correctness requires that we uniformly use either membars or acquire/release instructions. (maybe @adinn can confirm that)

Well, yes, strictly we really ought to to uniformly use either membars or acquire/release instructions. That's because the barrier model for ordering and the ldar/stlr model operate slightly differently so straight mixing can lead to failures like the Dekker problem. However, they can be made compatible (at least with current hardware) but only if you live with a slightly raised cost for the barrier-based version. The cost is adding a membar (dmb ish) before the ldr that implements a volatile load. n.b. that's as well as the acquire barrier (dmb ishsd) that follows the load.

That's what we do in OpenJDK to allow C1 barrier-based translation of volatile accesses and C2 stlr/ldar-based translation of volatile accesses to interoperate. The extra barrier at load is not needed to keep threads executing C1 code consistent. However, it ensures that threads executing a load via C1-generated code see the right values when other threads are executing stores using C2-generated code.

There is also the question of the various Unsafe intrinsics to consider though. We had to make sure they also operated using the correct behaviours.

rwestrel: What about the new patch that I just pushed? With LateMembars on, behavior of lowering depends on the platform. On aarch64, lowering never adds barriers (whether compiled under the economy or community configuration). All generated code uses the new barrier less volatile access scheme and there's no consistency issue between compiler configurations. On x86/sparc, lowering adds barriers for the economy configuration but not for the community configuration (barriers are then added late). This fixes the economy configuration that I broke with my previous change.

And, right, something needs to be done for Unsafe intrinsics. I'm not sure how to handle yet that so I'd like to keep it for a separate change.

rwestrel: That doesn't play well with native image. I'm trying to find a cleaner way to do this.

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/Aarch64VolatileAccessesTest.java
```diff
@@ -0,0 +1,195 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.function.Predicate;
+
+import static org.graalvm.compiler.core.common.GraalOptions.LateMembars;
+import static org.junit.Assume.assumeTrue;
+
+public class Aarch64VolatileAccessesTest extends MatchRuleTest {
```
XiaohongGong: The class name should be "AArch64VolatileAccessesTest" ？

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
```diff
@@ -506,9 +507,23 @@ public Variable emitLoad(LIRKind kind, Value address, LIRFrameState state) {
     }
 
     @Override
-    public void emitStore(ValueKind<?> lirKind, Value address, Value inputVal, LIRFrameState state) {
-        AArch64AddressValue storeAddress = getLIRGen().asAddressValue(address);
+    public Variable emitVolatileLoad(LIRKind kind, Value address, LIRFrameState state) {
+        AllocatableValue loadAddress = asAllocatable(address);
```
XiaohongGong: Does it need to ensure the address as a base register one for volatile access? See the "ensureBaseRegisterAddress()" in https://github.com/oracle/graal/pull/313

rwestrel: Sorry I didn't get back to you earlier. This PR is on hold until the related bug https://github.com/oracle/graal/pull/1844 is fixed. Also, as Andrew mentioned, unsafe volatile accesses handling is missing from https://github.com/oracle/graal/pull/1642. That needs to be fixed before this PR is integrated. I have a patch that I haven't submitted for this. But again I need https://github.com/oracle/graal/pull/1844 fixed first.

XiaohongGong: Thank you so much for the updating! I have a patch to handle the unsafe atomics. So with your patch, do you mean we use the "LateMembar" option on AArch64 to decide whether to use LDAR/STLR for java volatile access? If so, we should also use this option to handle the unsafe atomic. But in my patch I use the jvm option "-XX:+UseBarriersForVolatile" as the switch. Because whatever the option value is, the unsafe atomic always generate a single op. And it's strange to use "LateMembar" for it. So do you have any better idea or option to handle this? Actually we should use the same option to control Java volatile access and unsafe atomics.

XiaohongGong: Does the "LateMembar" option also work for SubstrateVM?

rwestrel: > Does the "LateMembar" option also work for SubstrateVM?

I'm not familiar enough with SVM to answer that question

rwestrel: To clarify, the thing that I have a patch for is Unsafe.get*Volatile/Unsafe.put*Volatile. You're right that hotspot's option UseBarriersForVolatile should be factored int. I'm unsure how to handle this but I'll make sure I'll take it into account when I rework that patch.

dougxc: @christianwimmer can you please address the SVM question above.

XiaohongGong: @rwestrel  I have a PR (https://github.com/oracle/graal/pull/2181) for the unsafe atomic issues. I only used the `"-XX:+UseBarriersForVolatile"` to decide the codes for atomics. And I think maybe I should also consider the graal option `“LateMambars”`, so you think so? And I also adds a same option for substratevm.  And anyway the options to decide codes for Java volatile access and unsafe atomics should keep the same. So if you have any better idea to handle this, please let me know. Thanks!

compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/Aarch64VolatileAccessesTest.java
rwestrel: Thanks. I made that change.

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
rwestrel: > Does it need to ensure the address as a base register one for volatile access? See the "ensureBaseRegisterAddress()" in #313

I'm confused by this. Isn't:
@Use protected AllocatableValue address;
sufficient?
@adinn: what do you think?

rwestrel: > @rwestrel I have a PR (#2181) for the unsafe atomic issues. It only used the `"-XX:+UseBarriersForVolatile"` to decide the codes for atomics. And I think maybe I should also consider the graal option `“LateMambars”`, do you think so? And I also adds the same option for substratevm. And anyway the options to decide codes for Java volatile access and unsafe atomics should keep the same. So if you have any better idea to handle this, please let me know. Thanks!

LateMembars got removed recently. The updated I push yesterday relies on UseBarriersForVolatile.

rwestrel:
I rebased and updated the patch. This includes SVM support but I wasn't sure how to test it so I can't tell if that part is functional.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateTargetDescription.java
```diff
@@ -50,6 +51,16 @@ public SubstrateTargetDescription(Architecture arch, boolean isMP, int stackAlig
...
 
+    /**
+     * We include all flags that enable AArch64 CPU instructions as we want best possible
+     * performance for the code.
+     *
+     * @return All the flags that enable AArch64 CPU instructions.
+     */
+    public static EnumSet<AArch64.Flag> allAArch64Flags() {
+        return EnumSet.of(AArch64.Flag.UseBarriersForVolatile);
+    }
```
XiaohongGong, line 62: I think this will always enable "UseBarriersForVolatile" . So for SVM, I think we could deal with this flag in two ways:
1. Always disable "UseBarriersForVolatile". That means always using "ldar/stlr" for volatile access.  Just never add `AArch64.Flag.UseBarriersForVolatile` to the EnumSet here.  This could avoid the issues caused by mixing use memory barriers like issues mentioned in https://github.com/oracle/graal/issues/2045.

2. Add an option for SVM like hotspot `-XX:+UseBarriersForVolatile`. And add this flag accord to the option. I also added the same option for SVM in my PR: https://github.com/oracle/graal/pull/2181.

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
XiaohongGong: Yes, cool! Actually there are some controversy for the vm option "-XX:+UseBarriersForVolatile" (See thread https://mail.openjdk.java.net/pipermail/aarch64-port-dev/2020-March/008679.html and https://mail.openjdk.java.net/pipermail/aarch64-port-dev/2020-April/008706.html). Although the final resolution is in discussion now,  this flag `UseBarriersForVolatile` might be removed in Graal in future.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateTargetDescription.java
rwestrel: Right. I got that backward. Thanks for catching it. I will fix it.

So does it sound ok if I go with 1. as you suggest and let you take care of 2. with PR #2181?

XiaohongGong: I think it's better to always use "ldar/stlr' for volatile access in SVM. So any opinion about it? @adinn @christianwimmer  Thanks!

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
adinn: > > Does it need to ensure the address as a base register one for volatile access? See the "ensureBaseRegisterAddress()" in #313
> 
> I'm confused by this. Isn't:
> @use protected AllocatableValue address;
> sufficient?
> @adinn: what do you think?

The method ensureBaseRegisterAddress() that XiaohonGong is referring to from my earlier patch deals with the restricted addressing modes available when using ldar/stlr. If the address is not actually a plain BASE_REGISTER_ONLY address then an lea is planted (it is actually coded as a Move insn) and the target of that lea is passed to the ladr/stlr. I think he may be pointing out the need to do this?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateTargetDescription.java
adinn: Yes, that's probably a better default. This was always really meant to be a diagnostic flag. However, we made it a product flag to provide an escape hatch just in case we ran into issues with early HW releases where ldar/stlr did not behave as expected. Since SVM does not have to worry about what setting is picked by an underlying JVM and since we know that all important current HW impementations of ldar/stlr are reliable it makes sense to always make this false.

rwestrel: I reverted that part of the patch

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
rwestrel: Right. But isn't that already guaranteed by
```
@Use protected AllocatableValue address;
```
where
```
    public static @interface Use {

        OperandFlag[] value() default OperandFlag.REG;
    }
```
? Doesn't that tell the register allocator that the address input needs to be in a register?

adinn: Hmm, I think you must be mistaking me for a compiler expert :-)

XiaohongGong: Basically there are several address modes in AArch64. The `address` here is actually a `AArch64AddressValue`, which might has other address modes except `BASE_REGISTER_ONLY`. You can have a look at `AArch64AddressValue` and  `AArch64Address.AddressingMode`.

rwestrel:
```
    public static final class VolatileLoadOp extends AArch64LIRInstruction {
        public static final LIRInstructionClass<VolatileLoadOp> TYPE = LIRInstructionClass.create(VolatileLoadOp.class);
        protected final AArch64Kind kind;
        @State protected LIRFrameState state;
        @Def protected AllocatableValue result;
        @Use protected AllocatableValue address;
```
It's not an AArch64AddressValue, it's AllocatableValue so I think it's guaranteed to be assigned a register.

adinn: Yes, you are right here, Roland (it's such a while since I looked at this I I had to reread the code to be sure). AArch64AddressValue is a Value but is not an instance of AllocatableValue. So, asAllocatable will detect that and plant a move (it calls emitMove) returning a Register as the AllocatableValue passed to ldar/stlr.

rwestrel: Thanks for the confirmation that's your understanding too.

XiaohongGong: Hi @rwestrel , the VM option "UseBarriersForVolatile" is finally to be determined to obsolete in jdk (See: https://bugs.openjdk.java.net/browse/JDK-8243339). I think we should remove it for Graal either.  I will also update my Unsafe atomic PR by removing it in future. Thanks!

rwestrel: @XiaohongGong do you think I should update this patch to remove UseBarriersForVolatile now or do we want to wait for JDK-8243339 to be pushed (and backported?) first?

XiaohongGong: I think we can remove it now, since it is not actually used ever before in Graal. Thanks!

rwestrel: I just pushed an updated patch with no reference to UseBarriersForVolatile

rwestrel:
@sanzinger this patch is good to go on our side. The CI failure is unrelated to my change AFAICT. Is anything holding it up on your side?

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
XiaohongGong: OK！Thanks so much, it looks ok to me.

sanzinger:
@rwestrel sorry for the long delay. I am looking into this again.

sanzinger:
@rwestrel The recent update made it better. The PR is scheduled for merge (With merge conflict resolved)

rwestrel:
> @rwestrel The recent update made it better. The PR is scheduled for merge (With merge conflict resolved)

thanks!

PR #1795, opened 2019-10-30
Generate complete stack trace string

zhangruoxu:
This pull request tries to fix a bug during generate stack traces as a string. 

A stack trace during class initialization has been pruned in the method `ConfigurableClassInitialization.relevantStackTrace()`, where the top `START_OF_THE_TRACE `, i.e., the beginning four irrelevant stack frames are skipped. Therefore, in the method `getTraceString()`, where a stack trace is transformed into a string, it is unnecessary to skip the beginning four frames, since those have already been removed. 

In this pull request, the trace string is generated from the first element of a stack trace.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ConfigurableClassInitialization.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ConfigurableClassInitialization.java
@@ -340,7 +340,7 @@ public class ConfigurableClassInitialization implements ClassInitializationSuppo
     private static String getTraceString(StackTraceElement[] trace) {
         StringBuilder b = new StringBuilder();
 
-        for (int i = START_OF_THE_TRACE; i < trace.length; i++) {
+        for (int i = 0; i < trace.length; i++) {
             StackTraceElement stackTraceElement = trace[i];
             b.append("\tat ").append(stackTraceElement.toString()).append("\n");
         }


zhangruoxu:
Using the following command to compile the following code snippet can reproduce the bug. 

```
native-image --initialize-at-build-time=A -H:+TraceClassInitialization TestTraceClassInit
```

```java
public class TestTraceClassInit {
        A a = new A();
        a.foo();
        B b = new B();
        b.bar();
    }
}

class A {
    private static B b;

    static {
        b = new B();
    }

    public void foo() {

    }
}

class B {
    public void bar() {}
}
```
Current output:
```
Error: Classes that should be initialized at run time got initialized during image building:
 B the class was requested to be initialized at build time (from the command line). A caused initialization of this class with the following trace: 

# Without any stack trace dumped here
```

Expected output:
```
Error: Classes that should be initialized at run time got initialized during image building:
 B the class was requested to be initialized at build time (from the command line). A caused initialization of this class with the following trace: 
	at B.<clinit>(TestTraceClassInit.java)
	at A.<clinit>(TestTraceClassInit.java:14)
```

zhangruoxu:
I am an employee from Alibaba Group. We have already signed the OpenJDK OCA at the company level. Please check our OCA. Cheers :-)
@graalvmbot

zhangruoxu:
@vjovanov Dear Vojin, I was wondering if this small pull request could be reviewed so that I could know whether it is a correct fix.

vjovanov:
Your fix will be included in 19.3. Thank you!

PR #1798, opened 2019-11-01
Fix NPE in coverage when a source has no path

chrisseaton:
Also generally add source name support to coverage.


eregon:
Seems related to https://github.com/oracle/graal/pull/1800.
I think it would be useful if @boris-spas reviews both.

chrisseaton:
Yes I think this one is a proper fix, while #1800 is a workaround which could probably be better fixed by doing something else, that's why they're separate.

boris-spas:
Thanks all, I'm aware of this an on it as soon as I can!

boris-spas:
Should be fixed by https://github.com/oracle/graal/commit/531e9b22510fecb992fab857b49581763565cd2f
Thanks!

PR #1799, opened 2019-11-01
Implement .p2align as a no-op

chrisseaton:
Could be interesting to pass the alignment all the way down into Graal!


chrisseaton:
It's from the Google *snappy* compression routine, which we want to compile unmodified, so I just needed a way to recognise and ignore the directive.

For example https://github.com/google/snappy/blob/b02bfa754ebf27921d8da3bd2517eab445b84ff9/snappy.cc#L693

They say they wanted the alignment as they observed that it reduced branch misprediction. Does Graal have a machine model for the impact of alignment on the branch prediction caches to align a loop automatically?

I wasn't suggesting that it's needed to pass it down into the compiler, but it would be an interesting experiment to add a `CompilerDirectives.alignmentHint(n)`.

dougxc:
Graal does some manual aligning of loops (e.g. https://github.com/oracle/graal/blob/1b0b825a62506f52c8c01a0df30737c2fb98e625/compiler/src/org.graalvm.compiler.lir.amd64/src/org/graalvm/compiler/lir/amd64/AMD64ArrayEqualsOp.java#L630) but I'm not aware of anything more general. @tkrodriguez ?

tkrodriguez:
I don't think we expose a high level alignment directive but it wouldn't be hard to implement.  You'd still get our automatic loop header alignment but you could perform your compiles with -Dgraal.LoopHeaderAlignment=1 to effectively disable it.

PR #1800, opened 2019-11-01
Workaround NPEs in coverage

chrisseaton:
This is more of an issue than a PR. I have to use these workarounds to avoid NPEs when running coverage on simple Ruby programs like this:

```ruby
require 'chunky_png'

a = ChunkyPNG::Image.from_file('a.png')
b = ChunkyPNG::Image.from_file('b.png')
c = a.compose(b, 0, 0)
c.save('c.png')
```

I don't really understand it enough to be able to fix the underlying problem.


chrisseaton:
CI failure seems like a flake.

boris-spas:
Should be fixed by https://github.com/oracle/graal/commit/531e9b22510fecb992fab857b49581763565cd2f
Thanks!

PR #1819, opened 2019-11-08
Introduce ability to specify which timezones to include in the native imagee…

jaikiran:
Currently when a native-image is built it by defaults adds _only_ `GMT`, `UTC` and the default zone id as the timezones that are initialized and made available in the generated image. Of course, there's `IncludeAllTimeZones` option which includes all the available timezones into the target image.

However, it would be better to expose an additional option which lets users decide which specific timezones to include in the target image. Quarkus project has seen issues like these[1] where even a `GMT` expressed as a different id `Etc/GMT` isn't available in the target image. Of course, that's just one use case. Having a more finer control on which timezones get included in the target image will help users in configuring the native-image build accordingly or allowing frameworks like Quarkus an ability to add them intelligently.

This is my first contribution to the graal project. However, I have a signed and approved OCA (as part of the JDK project) and here's my record in the census https://openjdk.java.net/census#jpai

I have done manual tests with these changes to verify that using the new `IncludeTimeZones` option indeed includes the specific timezones in the target image and also made sure the exisitng `IncludeAllTimeZones` option isn't regressed. I am willing to add an automated test case (couldn't find documentation for it) if someone can guide me or point me to how to do that.

[1] https://github.com/quarkusio/quarkus/issues/5269

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeZoneSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeZoneSubstitutions.java
@@ -76,18 +76,27 @@ final class Target_java_util_TimeZone {
 @AutomaticFeature
 final class TimeZoneFeature implements Feature {
     static class Options {
+        private static final TimeZone defaultZone = TimeZone.getDefault();
+
         @Option(help = "When true, all time zones will be pre-initialized in the image.")//
         public static final HostedOptionKey<Boolean> IncludeAllTimeZones = new HostedOptionKey<>(false);
+
+        @Option(help = "The time zones, in addition to the default zone of the host, that will be pre-initialized in the image.")//
+        public static final HostedOptionKey<String[]> IncludeTimeZones = new HostedOptionKey<>(new String[]{"GMT", "UTC", defaultZone.getID()});
     }
 
     @Override
     public void afterRegistration(AfterRegistrationAccess access) {
-        TimeZone defaultZone = TimeZone.getDefault();
-        String[] supportedZoneIDs = Options.IncludeAllTimeZones.getValue() ? TimeZone.getAvailableIDs() : new String[]{"GMT", "UTC", defaultZone.getID()};
+        final String[] supportedZoneIDs;
+        if (Options.IncludeAllTimeZones.getValue()) {
+            supportedZoneIDs = TimeZone.getAvailableIDs();
+        } else {
+            supportedZoneIDs = Options.IncludeTimeZones.getValue();
+        }
         Map<String, TimeZone> supportedZones = Arrays.stream(supportedZoneIDs)
                         .map(TimeZone::getTimeZone)
                         .collect(toMap(TimeZone::getID, tz -> tz, (tz1, tz2) -> tz1));
-        ImageSingletons.add(TimeZoneSupport.class, new TimeZoneSupport(supportedZones, defaultZone));
+        ImageSingletons.add(TimeZoneSupport.class, new TimeZoneSupport(supportedZones, Options.defaultZone));
     }
 }
 


jaikiran:

Please check for the user `jpai` https://openjdk.java.net/census#jpai

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeZoneSubstitutions.java
```diff
@@ -78,12 +78,23 @@ public static TimeZone getTimeZone(String id) {
...
         TimeZone defaultZone = TimeZone.getDefault();
-        String[] supportedZoneIDs = Options.IncludeAllTimeZones.getValue() ? TimeZone.getAvailableIDs() : new String[]{"GMT", "UTC", defaultZone.getID()};
+        final String[] supportedZoneIDs;
+        if (Options.IncludeAllTimeZones.getValue()) {
+            supportedZoneIDs = TimeZone.getAvailableIDs();
+        } else if (Options.IncludeTimeZones.hasBeenSet()) {
+            final String[] timeZones = Options.IncludeTimeZones.getValue().split(",");
+            supportedZoneIDs = Arrays.stream(timeZones).map(String::trim).toArray(String[]::new);
+        } else {
+            supportedZoneIDs = new String[]{"GMT", "UTC", defaultZone.getID()};
```
vjovanov: This line feels wrong. Would it be possible to make this a default value of `IncludeTimeZones`?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeZoneSubstitutions.java
```diff
@@ -78,12 +78,23 @@ public static TimeZone getTimeZone(String id) {
     static class Options {
         @Option(help = "When true, all time zones will be pre-initialized in the image.")//
         public static final HostedOptionKey<Boolean> IncludeAllTimeZones = new HostedOptionKey<>(false);
+
+        @Option(help = "A comma separated list of time zones that will be pre-initialized in the image.")//
+        public static final HostedOptionKey<String> IncludeTimeZones = new HostedOptionKey<>(null);
```
vjovanov: I would also make this a `String[]`. That way you will be able to add time zones in different places on the command line (or jars) and they will compose (which I believe should be the right behavior).

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeZoneSubstitutions.java
```diff
@@ -78,12 +78,23 @@ public static TimeZone getTimeZone(String id) {
     static class Options {
         @Option(help = "When true, all time zones will be pre-initialized in the image.")//
         public static final HostedOptionKey<Boolean> IncludeAllTimeZones = new HostedOptionKey<>(false);
```
vjovanov: Not sure this is the right default value. How much does `-H:+IncludeAllTimeZones` cost in terms of image size?

Anyhow, this is out of the scope of this PR.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeZoneSubstitutions.java
jaikiran: Hello @vjovanov,

Thank you very much for your review and sorry about the late response.

I've taken your input on this one and updated this PR to include this change. These values now form the default value for the new option.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeZoneSubstitutions.java
jaikiran: Done. I wasn't aware of this support. Thank you for pointing me to this. With this change, the param will now be passed (from command line) multiple times:

```
../../mx/mx native-image -H:IncludeTimeZones=Europe/Berlin -H:IncludeTimeZones=GMT HelloWorld 
```

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeZoneSubstitutions.java
jaikiran: > How much does -H:+IncludeAllTimeZones cost in terms of image size?

We have been having this similar conversation in the Quarkus project https://groups.google.com/forum/#!msg/quarkus-dev/J1afzAhpShI/A5vmUGJLAQAJ

My experiments (with a couple of different programs, one being the `HelloWorld` example used in `substratevm` project) have shown that when `IncludeAllTimeZones` isn't set then the filesize is something along lines of 2.9M:

```
ls -lh helloworld 
-rwxr-xr-x  1 jaikiran  jaikiran   2.9M Nov 14 23:22 helloworld
```
and when you set `IncludeAllTimeZones`, the size increases very slightly to:
```
-rwxr-xr-x  1 jaikiran  jaikiran   3.3M Nov 14 23:50 ./helloworld
```

So not a big difference really. So yes, by default including all timezones probably will help application frameworks like Quarkus, where in certain situations there isn't really an easy way to determine at build time which timezones will be needed in the native image (imagine an application which communicates with a DB server whose timezone information can only be obtained at runtime).

But yes, like you say, the default value of `IncludeAllTimeZones` is something that needs to be discussed and sorted out separately.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeZoneSubstitutions.java
```diff
@@ -76,18 +76,27 @@ public static TimeZone getTimeZone(String id) {
 @AutomaticFeature
 final class TimeZoneFeature implements Feature {
     static class Options {
+        private static final TimeZone defaultZone = TimeZone.getDefault();
```
vjovanov: Not a fan of static state, but I guess this one is OK to stay here.

jaikiran:
@vjovanov Thank you very much for reviewing this and merging it. 

For the question of the default value for `IncludeAllTimeZones`, I have created a separate issue here https://github.com/oracle/graal/issues/1833

aruffie:
I confirm too, I have the bug here: https://github.com/quarkiverse/quarkus-univocity-parsers/pull/3/commits/5321516c5126adf349e9e7f060bfbf46883c6cbd   the timezone is different between my non-native test phase and the native test phase. I think the "-H:+IncludeAllTimeZones" isn't take in account

PR #1824, opened 2019-11-12
AArch64: Fixed a bug in reading return address.

tshull226:
Code for reading a frame's return address was not being generated
correctly in one place. As shown in AArch64FrameMap.java, the return
address is at the top of the frame.

Change-Id: I3465956f82846530dfe0eb9e50782540ed4d1056


tshull226:
I'm currently working for Arm, so I fall under their organizational OCA.

bobmcwhirter:
LGTM

sanzinger:
Thank you for contribution. I am integrating this change.

PR #1825, opened 2019-11-12
Limit thread scaling and promptly shutdown pool

tshull226:
When running on a large number of cores, within the native image class
loader many threads were being created. Futhermore, these threads were
not being shutdown until the end of execution.

Creating too many threads can cause the process to hang. This is because
each thread makes a call to read /dev/random to get a random seed.
However, it is possible for the system to run out of entropy if too many
threads are created in a short period of time, causing the read of
/dev/random to hang.

To resolve this, I limit the number of threads created by the native
image class loader and also ensure its ForkJoinPool is shutdown
promptly.

Change-Id: I225ba57ab89fee3b7aafcca9913f4ca676aec6bd


tshull226:
Hi. Is there a chance this PR will be merged?

PR #1834, opened 2019-11-16
[Docs] Fix a typo

284km:
s/occurence/occurrence/


dougxc:
Thanks for the spelling fix. If you don't mind, I'll include it in an existing PR to avoid the overhead of pushing such a small change through the integration process.

284km:
Thanks for your explanation. That would be great. I hope to avoid the overhead.

zakkak:
This has been fixed.

PR #1844, opened 2019-11-20
fix broken economy configuration

rwestrel:
https://github.com/oracle/graal/pull/1642 broke the economy configuration: barriers are no lower added at lowering time under the expectation that the InsertMembarsPhase will add them but it's not part of the economy configuration. I suppose we don't want it in the economy configuration and instead want lowering to add the membars when running the economy configuration.


rwestrel:
Can some comment on the status of the economy configuration? Is it supported/used? As reported here it's broken (on all platforms, not only aarch64) but there doesn't much interest in fixing it.

sanzinger:
Hi @rwestrel 

I am working on this PR now. Will give you an update shortly.

sanzinger:
I have ran the tests and on AArch64, the `LateMembarInsertionTest` fails because `MembarNode` is put around the `WriteNode` in the `InsertMembarsPhase`:
```
2) org.graalvm.compiler.core.test.LateMembarInsertionTest#test08
java.lang.AssertionError: expected:<true> but was:<false>
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.graalvm.compiler.core.test.LateMembarInsertionTest.verifyMembars(LateMembarInsertionTest.java:217)
	at org.graalvm.compiler.core.test.LateMembarInsertionTest.test08(LateMembarInsertionTest.java:192)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at com.oracle.mxtool.junit.MxJUnitWrapper.runRequest(MxJUnitWrapper.java:310)
	at com.oracle.mxtool.junit.MxJUnitWrapper.main(MxJUnitWrapper.java:226)
```

sanzinger:
Please provide a fix for this error above. Then I am able to integrate it.

rwestrel:
> Please provide a fix for this error above. Then I am able to integrate it.

Thanks for looking at this Stefan. I'm working on an update to the patch.

sanzinger:
Our internal gate checks are in progress.

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/LoweringPhase.java
```diff
@@ -236,14 +238,25 @@ private void setLastFixedNode(FixedWithNextNode n) {
...
+            return expandsVolatileMembarsLate;
+        }
     }
 
     private final CanonicalizerPhase canonicalizer;
     private final LoweringTool.LoweringStage loweringStage;
+    private Boolean expandsVolatileMembarsLate;
 
-    public LoweringPhase(CanonicalizerPhase canonicalizer, LoweringTool.LoweringStage loweringStage) {
+    public LoweringPhase(CanonicalizerPhase canonicalizer, LoweringTool.LoweringStage loweringStage, Boolean expandsVolatileMembarsLate) {
```
dougxc: Why is this a boxed Boolean? Also, I only see it ever being set to true - under what conditions do you expect it to be set to false?

rwestrel: Thanks for looking at this.
The boxed Boolean is a mistake (must have trusted the IDE refactoring too much).
There are 2 LoweringPhase constructors. Second one doesn't take the expandsVolatileMembarsLate argument and sets it to false.That one is called from the EconomyHighTier constructor. So expandsVolatileMembarsLate should only be true when called from the HighTier constructor.
This said I now wonder if that does the right thing: snippets would always have expandsVolatileMembarsLate set to false and I suppose a snippet could have a volatile access too.
An other way of doing that would be to force LateMembars to false in the OptionValues in the case of the economic configuration.

dougxc: Sorry, I missed the fact that you added an extra LoweringPhase constructor.
@tkrodriguez can you please weigh in on Roland's observation about volatiles in snippets.

rwestrel: Some context: the end goal with all this is to generate a code pattern on aarch64 that doesn't contain membar instructions. Using that code pattern is only legal if all code that is emitted uses the membar less pattern. Mixing membar volatile accesses with non membar volatile accesses in the same process must not happen. So:
on non aarch64, we want LateMembars true for the CommunityCompilerconfiguration and false for the EconomyCompilerConfiguration
on aarch64 once the non membar pattern is supported, LateMembars would need to be true for both the CommunityCompilerconfiguration and EconomyCompilerConfiguration.
The patch I proposed tries to achieve this by overriding the value of LateMembars with an extra boolean that's passed around but it seems hard to do globally. Actually the root cause is that option values are inconsistent. If the economy configuration is enabled then LateMembars must be off. It's a common pattern in hotspot to force values of command line flags so they are consistent. I don't see much of that in Graal.

dougxc: >  If the economy configuration is enabled then LateMembars must be off

Isn't this a matter of ensuring in the `EconomyLowTier` constructor that `LateMembars` is not explicitly set?

It is indeed a little harder to ensure consistency of options in Graal because they are not as global as HotSpot options. This allows for options to be overridden in certain contexts (e.g. testing, snippets) while not disturbing other concurrently executing contexts.

rwestrel: > Isn't this a matter of ensuring in the `EconomyLowTier` constructor that `LateMembars` is not explicitly set?

So you're saying changing the value of the LateMembars option is the way to go?

LateMembars is on by default and we want it off at lowering time for the Economy configuration. Once https://github.com/oracle/graal/pull/1772 is in, we would want if off only on x86.

dougxc: Why is `LateMembars` even an option? It sounds like it's controlling something that should never be controlled by a user.

rwestrel: > Why is `LateMembars` even an option? It sounds like it's controlling something that should never be controlled by a user.

To help troubleshooting. But you're right that it's not something that the user should change. So if we remove the option where would we store a global flag that tells whether membars should be added at lowering time or as an independent phase or not at all?

dougxc: Do we really need a global? At every point that `LateMembar` is currently read, do we not have sufficient context (i.e. architecture, compiler config, etc) to determine how to treat membars? If not, I would try and thread that info through somehow.

rwestrel: From the DefaultJavaLoweringProvider, AFAICT, there's no way to tell what compiler configuration is being used. Threading that info through is what the proposed patch does, right? Except I now think snippets are not covered and they should be. But getting than bit of information through to the snippet machinery seems quite involved. Couldn't it be attached to some objects that's carried around? The Graph object maybe?

dougxc: Where in the snippet machinery is this information needed?

rwestrel: SnippetTemplate() creates a LoweringPhase() which AFAIU would also need that information

dougxc: Seems like we want something similar to `org.graalvm.compiler.nodes.StructuredGraph.guardsStage`. What do you think @tkrodriguez @gilles-duboscq ?

rwestrel: FYI, I pushed a new change that propagates expandsVolatileMembarsLate to snippets. Wasn't as complicated as I thought but that looks pretty ugly and fragile to me.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/SnippetTemplate.java
```diff
@@ -558,13 +558,15 @@ public ValueNode findLength(FindLengthMode mode, ConstantReflectionProvider cons
         private final GuardsStage guardsStage;
         private final LoweringTool.LoweringStage loweringStage;
         private int hash;
+        private boolean expandsVolatileMembarsLate;
```
dougxc: This `expandsVolatileMembarsLate` field must participate in `CacheKey.equals` and `CacheKey.hashCode`.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/SnippetTemplate.java
```diff
@@ -374,9 +374,9 @@ protected Lazy lazy() {
 
         protected int nextParamIdx;
 
-        public Arguments(SnippetInfo info, GuardsStage guardsStage, LoweringTool.LoweringStage loweringStage) {
+        public Arguments(SnippetInfo info, GuardsStage guardsStage, LoweringTool.LoweringStage loweringStage, boolean expandsVolatileMembarsLate) {
```
dougxc: I think the `loweringStage` argument is always coming from a call to `org.graalvm.compiler.nodes.spi.LoweringTool.getLoweringStage()` in which case the `LoweringTool` object can be passed to this constructor and the constructor can extract the `loweringStage` and `expandsVolatileMembarsLate` values itself.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/SnippetTemplate.java
rwestrel: Right. loweringState is not hashed with the other fields. I suppose it's ok to leave expandsVolatileMembarsLate and only change equals()?

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/SnippetTemplate.java
rwestrel: I made that change and pushed an updated patch.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/SnippetTemplate.java
dougxc: Yes.

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/LoweringPhase.java
dougxc: Doesn't look any worse than threading `loweringStage` through to the snippets in my opinion.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/SnippetTemplate.java
dougxc: Looks good.

rwestrel: Thanks for your help with this.

dougxc: No problem. Note that @sanzinger will have to make some changes in the enterprise code base to adapt to these changes so probably won't get around to re-submitting for merge until next week (he's currently on vacation).

sanzinger: > I think the loweringStage argument is always coming from a call to org.graalvm.compiler.nodes.spi.LoweringTool.getLoweringStage()

This assumption is wrong. The enterprise implementation uses snippets outside of lowerings as well, where we don't have any `LoweringTool`.
I think the fact that "volatile field reads/writes with barriers" is a property of the entire graph and should be attached there. I could either think of a member in `StructuredGraph` or even an additional property such as `StoreFieldNode.requireMembar` or `LoadFieldNode.requireMembar`.

tkrodriguez:
I think we should eliminate LateMembars as an option.  Late expansion should be the only way to do it and I think it should be built directly into the lowering.  Someday when we have a C1 replacement based on economy we might have both configurations in use at the same time and since snippets are shared we can't (don't want to) support mixed handling of the membars.

rwestrel:
> I think we should eliminate LateMembars as an option. Late expansion should be the only way to do it and I think it should be built directly into the lowering. Someday when we have a C1 replacement based on economy we might have both configurations in use at the same time and since snippets are shared we can't (don't want to) support mixed handling of the membars.

I'm fine with dropping the LateMembars option. Beyond that, would moving the membar insertion to the last pass of lowering (from VolatileReadNode to a ReadNode surrounded by membars) be ok? The economy and community configurations would them behave the same (and this bug would be fixed). On Aarch64 I would have to make sure the lowering is skipped.

tkrodriguez:
Yes I think that would be the right way to do it.

rwestrel:
I pushed a reworked fix that adds barriers around a volatile access during the last pass of lowering. I also removed the LateMembars option. I added a simplify() method to VolatileReadNode so VolatileReadNodes that have no use other than memory edges are removed. I noticed that, with barriers added at lowering time, useless VolatileReadNodes sometimes leave MemBarNodes behind otherwise.

sanzinger:
Sorry for the long delay. I was working on internal projects. I'll try to look at it tomorrow again.

sanzinger:
The change looks good and is scheduled for merge. You don't have to resolve the merge conflict, I have merged it on our end.

rwestrel:
> The change looks good and is scheduled for merge. You don't have to resolve the merge conflict, I have merged it on our end.

Thanks for getting this in.

dougxc:
According to git bisect, this PR is the cause of https://bugs.openjdk.java.net/browse/JDK-8247832. I used this test to do the bisecting:
```
public class FuzzerTest1 extends GraalCompilerTest {

    static class Obj {
        volatile long l = -4290070568L;
        short s = -15583;
    }

    public int vMeth(Obj o) {
        int i18 = 0;
        for (int i = 0; i < 3; i++) {
            i18 = (int) o.l;
            o.s += (short) o.l;
            // System.out.println(o.s);
        }
        return o.s + i18;
    }

    @Test
    public void testVMeth() {
        ArgSupplier obj = () -> new Obj();
        test("vMeth", obj);
    }
}
```
Would you be able to look into this @rwestrel ?

rwestrel:
> Would you be able to look into this @rwestrel ?

Sure.

dougxc:
Thanks!

Issue #1849, opened 2019-11-20
[Bug] DisallowedImageHeapObjectFeature suggests incorrect option


aguibert:
Currently if native image fails due to a disallowed image in the heap, we get an error message like this:
```
com.oracle.svm.core.util.UserError$UserException: Detected a started Thread in the image heap. 
Threads running in the image generator are no longer running at image run time. Object has been 
initialized without the native-image initialization instrumentation and the stack trace can't be tracked. 
The object was probably created by a class initializer and is reachable from a static field. You can 
request class initialization at image run time by using the option 
--initialize-at-build-time=<class-name>. Or you can write your own initialization methods and call 
them explicitly from your main entry point.
```

The message incorrectly suggests using the `--initialize-at-build-time=<class-name>` option in order to initialize a class at run time. Instead, it should be suggesting the `--initialize-at-run-time=<class-name>` option.

This issue was acknowledged here: https://github.com/oracle/graal/issues/1669#issuecomment-529870061 but it looks like it has been forgotten, so I thought I'd raise an issue to track it

Should be a quick fix to this line:
https://github.com/oracle/graal/blob/92fcb5847cee89dffac156f0c6d850bccb7aadd5/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/DisallowedImageHeapObjectFeature.java#L129
But I can't sign the OCA to propose the fix on my own.

PR #1854, opened 2019-11-21
Fix the error message in DisallowedImageHeapObjectFeature to refer to initialize-at-run-time instead of initialize-at-build-time

jaikiran:
Fixes https://github.com/oracle/graal/issues/1849.

The commit here fixes the error message printed out by `DisallowedImageHeapObjectFeature`.


vjovanov:
Thanks, I have one more of these coming soon.

PR #1881, opened 2019-11-25
Keeping tailored jar files as resources

ziyilin:
This patch has two major contributions:
1. When getResource methods(_java.lang.ClassLoader.getResource, java.lang.ClassLoader.getSystemResource, java.lang.ClassLoader.getResourceAsStream, java.lang.ClassLoader.getSystemResourceAsStream, java.lang.ClassLoader.getResources, java.lang.ClassLoader.getSystemResources, java.lang.System.getResource, java.lang.System.getResourceAsStream_) get resources from a jar file, reserve a tailored jar file in file system and ship it together with the executable native image file. It keeps compatible with tradational Java behavior. 
**-Dsvm.resources.jarLocation=** gives the preserve location, and the default value is resources/jars.
**-Dsvm.resource.preserveJarRes=true** to enable this patch, by default its false.

    Jar file is tailored to keep 1) non-class files and 2) class files that are explicitly declared as resource in getResource methods.
    The URL protocols returned by getResource methods in Java, SVM and this patch are listed in the following table:

Resource locations | Protocol in Java |Protocol in SVM | Protocol in this patch|
---------- |-----------|-------|-------- |
In directory| file: | resource: | resource:
In Jar | jar:file: | resource: | jar:file:

  We developed this patch because we use the result of getResource as an index to search the containing jar file for other configuration files. It's too expensive to modify all of such usage.

2. Fixes a bug: no inner class is matched using resource config file generated by native agent.
 The agent generates resource config file with resource's actual name, but that name is used as a regular expression during build time. So the "$" in inner class is taken as an end of line.

The test case [jarResTest.zip](https://github.com/oracle/graal/files/3885252/jarResTest.zip) can verify this patch.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
@@ -638,15 +638,13 @@ public final class DynamicHub implements JavaKind.FormatWithToString, AnnotatedE
     @Substitute
     private InputStream getResourceAsStream(String resourceName) {
         final String path = resolveName(getName(), resourceName);
-        List<byte[]> arr = Resources.get(path);
-        return arr == null ? null : new ByteArrayInputStream(arr.get(0));
+        return getClassLoader0().getResourceAsStream(path);
     }
 
     @Substitute
     private URL getResource(String resourceName) {
         final String path = resolveName(getName(), resourceName);
-        List<byte[]> arr = Resources.get(path);
-        return arr == null ? null : Resources.createURL(path, arr.get(0));
+        return getClassLoader0().getResource(path);
     }
 
     private String resolveName(String baseName, String resourceName) {
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JarResourcePreserver.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JarResourcePreserver.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) 2019 Alibaba Group Holding Limited. All Rights Reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation. Alibaba designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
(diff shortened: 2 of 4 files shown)


ziyilin:
> Hello cengfeng.lzy, thanks for contributing a PR to our project!
> We use the Oracle Contributor Agreement to make the copyright of contributions clear. We don't have a record of you having signed this yet, based on your email address cengfeng -(dot)- lzy -(at)- alibaba-inc -(dot)- com. You can sign it at that link.

I'm from Alibaba inc.. We have signed the OpenJDK OCA at company level. Please take a check, thanks.@graalvmbot

christianwimmer:
The benefit of a native image is that the output is a single self-contained binary where all dependencies are linked in statically. Loading resources at run time from a secondary .jar file makes it more complicated to distribute the binary.

Why do you think that embedding the information directly into the native image is not sufficient?

ziyilin:
This requirement is from our practical scenario. The following code snippet is extracted from our production code:
```
Enumeration<URL> urls = appClassLoader
.getResources("META-INF/MANIFEST.MF");
while (urls.hasMoreElements()) {
    URL url = urls.nextElement();
    if (ResourceUtils.isJarURL(url)) {
        try {
            URLConnection con = url.openConnection();
            Assert.isInstanceOf(JarURLConnection.class, con);
            JarURLConnection jarCon = (JarURLConnection) con;
            JarFile jarFile = jarCon.getJarFile();
            //Retrieve Spring configurations from jar file
            retrieveSpringConfigs(jarFile);
         }
        catch (Throwable t) {
            throw new RuntimeException(t);
        }
    }
}
```
The result from getResource method is used as an index to locate the jar file, and the acutal purpose is to retrieve something else from certain jar files.  Such usage is quite common in our program, and current embeded resource mechanism cannot fulfill the requirement.

However, the best approach to support such requirement would be embedding jar files into the single binary executable file but keep the JDK API unchanged.  We are looking for the possible solution, too.

christianwimmer:
> However, the best approach to support such requirement would be embedding jar files into the single binary executable file but keep the JDK API unchanged.

I agree with that. We are currently working on better `ClassLoader` support (the first PRs are integrated, see https://github.com/oracle/graal/commit/5bd27d29a4cf434238de7fb2bfb208d82bd7d534). Now that we have proper `ClassLoader` objects in the image heap, you can implement a proper URL protocol that accesses a virtual file system that is populated at image build time and can be accessed at run time. If you start working on that, please create an issue where you explain the design you want to implement, and coordinate with @cstancu so that we do not duplicate any work.

In the meantime, I don't think we should merge this PR because it does not provide a good user experience.

ziyilin:
> you can implement a proper URL protocol that accesses a virtual file system that is populated at image build time and can be accessed at run time

This is what exactly comes out in my mind at the very beginning, but didn't implement it due to project schedule.  I'll see how to make it happen.

Issue #1883, opened 2019-11-25
[native-image] NullPointerException during security services registration


gwenneg:
Hi GraalVM team!

I'm working on the GraalVM `19.3.0` integration into Quarkus and a `NullPointerException` is being thrown during three of the Quarkus extensions integration tests.

Here's an example of NPE (this one is from a local build but we have the same failures in the Quarkus CI):

```
[INFO] [io.quarkus.deployment.pkg.steps.NativeImageBuildStep] /Library/Java/JavaVirtualMachines/graalvm-unknown-java8-19.3.0/Contents/Home/jre/bin/native-image -J-Djava.util.logging.manager=org.jboss.logmanager.LogManager -J-Dsun.nio.ch.maxUpdateArraySize=100 -J-Dvertx.logger-delegate-factory-class-name=io.quarkus.vertx.core.runtime.VertxLogDelegateFactory -J-Dvertx.disableDnsResolver=true -J-Dio.netty.leakDetection.level=DISABLED -J-Dio.netty.allocator.maxOrder=1 --initialize-at-build-time= -H:InitialCollectionPolicy=com.oracle.svm.core.genscavenge.CollectionPolicy$BySpaceAndTime -jar quarkus-integration-test-elytron-security-oauth2-999-SNAPSHOT-runner.jar -H:FallbackThreshold=0 -H:+ReportExceptionStackTraces -H:+TraceSecurityServices -H:-AddAllCharsets -H:EnableURLProtocols=http,https --enable-all-security-services -H:+JNI --no-server -H:-UseServiceLoaderFeature -H:+StackTrace quarkus-integration-test-elytron-security-oauth2-999-SNAPSHOT-runner
[quarkus-integration-test-elytron-security-oauth2-999-SNAPSHOT-runner:47933]    classlist:  19,354.31 ms
[quarkus-integration-test-elytron-security-oauth2-999-SNAPSHOT-runner:47933]        (cap):   6,483.14 ms
[quarkus-integration-test-elytron-security-oauth2-999-SNAPSHOT-runner:47933]        setup:  10,970.08 ms
13:43:26,876 INFO  [org.jbo.threads] JBoss Threads version 3.0.0.Final
[quarkus-integration-test-elytron-security-oauth2-999-SNAPSHOT-runner:47933]     analysis:  59,683.57 ms
Fatal error: java.lang.NullPointerException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
[...]

PR #1893, opened 2019-11-26
Prevent NullPointerException in SecurityServicesFeature

jaikiran:
Fixes https://github.com/oracle/graal/issues/1883

The `SecurityServicesFeature` adds an (internal) callback to register certain `java.security.Provider.Service` related classes for reflection. This callback is triggered if/when the `Provider.Service#newInstance` method is reachable in the application. When this callback gets triggered, it uses the `java.util.Provider#knownEngines` (internal) private field to get hold of certain details. This `knownEngines` will only contain details of services that are shipped within the JDK. As a result, if the application uses non-JDK related services, this `knownEngines` will not have information for such service types. The code, which tries to do the reflection registration, however doesn't check for `null` in such cases and that results in the reported NPE.

The commit here adds a `null` check for such cases. This PR also includes a test case which reproduces this issue and verifies the fix. The test case was run locally as follows:

```
mx build
mx native-unittest com.oracle.svm.test.SecurityServiceTest --build-args -H:+EnableAllSecurityServices --features=com.oracle.svm.test.SecurityServiceTest\$TestFeature
```
Note: The `EnableAllSecurityServices` is necessary to trigger this issue.

I haven't added this test's `TestFeature` to the `gate` features (https://github.com/oracle/graal/blob/master/substratevm/mx.substratevm/mx_substratevm.py#L429) since I don't know if that's expected to be done for any new test cases.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
@@ -278,11 +278,20 @@ public class SecurityServicesFeature extends JNIRegistrationUtil implements Feat
             try {
                 /*
                  * Access the Provider.knownEngines map and extract the EngineDescription
-                 * corresponding to the serviceType. From the EngineDescription object extract the
-                 * value of the constructorParameterClassName field then, if the class name is not
-                 * null, get the corresponding Class<?> object and return it.
+                 * corresponding to the serviceType. Note that the map holds EngineDescription(s) of
+                 * only those service types that are shipped in the JDK. From the EngineDescription
+                 * object extract the value of the constructorParameterClassName field then, if the
+                 * class name is not null, get the corresponding Class<?> object and return it.
                  */
                 /* EngineDescription */Object engineDescription = knownEngines.get(serviceType);
+                /*
+                 * This isn't an engine known to the Provider (which actually means that it isn't
+                 * one that's shipped in the JDK), so we don't have the predetermined knowledge of
+                 * the constructor param class.
+                 */
+                if (engineDescription == null) {
+                    return null;
+                }
                 String constrParamClassName = (String) consParamClassNameField.get(engineDescription);
                 if (constrParamClassName != null) {
                     return access.findClassByName(constrParamClassName);
--- a/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/SecurityServiceTest.java
+++ b/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/SecurityServiceTest.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2019, 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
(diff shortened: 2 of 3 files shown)


substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/SecurityServiceTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Tests the {@code SecurityServicesFeature}
+ */
+public class SecurityServiceTest {
+    public static class TestFeature implements Feature {
+        @Override
+        public void beforeAnalysis(BeforeAnalysisAccess access) {
```
cstancu: This should be done in a previous phase, `Feature.duringSetup()`, to be sure that `SecurityServicesFeature.beforeAnalysis()` sees the complete provider list.

In general, the processing order of features is non-deterministic, but it can be enforced via `Feature.getRequiredFeatures()`.

cstancu:
@jaikiran thanks for your contribution! Yes, `TestFeature` needs to be added to the `gate` features otherwise the gate fails: https://travis-ci.org/oracle/graal/jobs/617274177#L1913

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
```diff
@@ -278,11 +278,18 @@ private static void linkJaas(DuringAnalysisAccess duringAnalysisAccess) {
...
-                 * null, get the corresponding Class<?> object and return it.
+                 * corresponding to the serviceType. Note that the map holds EngineDescription(s) of
+                 * only those service types that are shipped in the JDK. From the EngineDescription
+                 * object extract the value of the constructorParameterClassName field then, if the
+                 * class name is not null, get the corresponding Class<?> object and return it.
                  */
                 /* EngineDescription */Object engineDescription = knownEngines.get(serviceType);
+                // this isn't a engine known to the Provider (which actually means that it isn't one
+                // that's shipped in the JDK), so we don't have the predetermined knowledge of the
+                // constructor param class
```
cstancu: Nitpicking, but can you please use the `/**/` style for the above comment. Also, use capitalization and end it with a `.`. We like to keep our comment style consistent as much as possible. :)

gwenneg:
FYI, this commit fixed the Quarkus issues described in https://github.com/quarkusio/quarkus/pull/5358. It would be great if it could be included in GraalVM 19.3.1 as we will probably need it to integrate GraalVM 19.3 into Quarkus.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/SecurityServiceTest.java
jaikiran: Thank you for those inputs @cstancu, they are useful. I've updated this PR to take them into account.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
jaikiran: Done :)

jaikiran:
> @jaikiran thanks for your contribution! Yes, TestFeature needs to be added to the gate features otherwise the gate fails.

Indeed. I've updated the PR to include that in the gate as well as add the `-H:+EnableAllSecurityServices` since it's needed by this testcase.

cstancu:
@jaikiran it's all merged, thanks again!

@gwenneg yes, the fix will be back-ported to 19.3.1.

PR #1894, opened 2019-11-27
Wrap AArch64 immediate addresses

teshull:
Due to the limited number of bits available when using immediate
addressing in AArch64, it is necessary to guard against immediate
operand overflow. If overflow does occur, then a register-based
addressing mode must be used instead, and the address should be loaded
via mov(z|k) instructions into a register.

Fixed SVM backend to guard against unchecked non hard-coded immediates.

Change-Id: I7d1030abbf554a60b8263ef9a3ed6ffc50c6ebf7

### DIFF
--- a/substratevm/src/com.oracle.svm.core.graal.aarch64/src/com/oracle/svm/core/graal/aarch64/SubstrateAArch64Backend.java
+++ b/substratevm/src/com.oracle.svm.core.graal.aarch64/src/com/oracle/svm/core/graal/aarch64/SubstrateAArch64Backend.java
@@ -235,24 +235,26 @@ public class SubstrateAArch64Backend extends SubstrateBackend implements LIRGene
          * metadata is registered for the end of the instruction just works.
          */
         int startPos = masm.position();
-        try (ScratchRegister scratch = masm.getScratchRegister()) {
-            Register tempRegister = scratch.getRegister();
+        try (ScratchRegister scratch1 = masm.getScratchRegister(); ScratchRegister scratch2 = masm.getScratchRegister()) {
+            Register tempRegister = scratch1.getRegister();
+            Register overflowRegister = scratch2.getRegister();
             // Save PC
             masm.adr(tempRegister, 4); // Read PC + 4
             crb.recordIndirectCall(startPos, masm.position(), null, state);
-            masm.str(64, tempRegister, AArch64Address.createPairUnscaledImmediateAddress(anchor, runtimeConfiguration.getJavaFrameAnchorLastIPOffset()));
+            masm.str(64, tempRegister, AArch64Address.createWrappedUnscaledImmediateAddress(masm, anchor, runtimeConfiguration.getJavaFrameAnchorLastIPOffset(), overflowRegister));
             // Save SP
             masm.mov(64, tempRegister, sp);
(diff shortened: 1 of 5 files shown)


tshull226:
Yeah, you're right - I thought the code would look nicer that way, but it's not worth creating opportunities for mistakes. I have fixed the code now so that the scratch register resource lifetimes surround the address/memory operations.

tshull226:
I believe the previous issues have been resolved. Can this PR please be assigned to someone?

substratevm/src/com.oracle.svm.core.graal.aarch64/src/com/oracle/svm/core/graal/aarch64/SubstrateAArch64Backend.java
```diff
@@ -171,6 +172,20 @@ public SubstrateAArch64Backend(Providers providers) {
         super(providers);
     }
 
+    static class AArch64AddressFactory {
+        static AArch64Address createWrappedUnscaledImmediateAddress(AArch64MacroAssembler masm, Register base, int imm, int immBits, ScratchRegister scratch) {
```
sanzinger: I think we should pass `Register` rather than `ScratchRegister` here. It gives the caller the choice whether a scratch or other register is used here.

Can't we move this static method into `AArch64Address`?

tshull226: Hi. Sorry for the delay - got caught up making sure I could graduate :). Yes, I'll change the ScratchRegister to a Register.

Originally I put `AArch64AddressFactory` in SubstrateAArch64Backend because I didn't want to pollute the non-substrate code. However, It looks like the same method is being called a couple other places too. The use of it in `AArch64HotSpotPatchReturnAddressOp` looks fishy to me. I'll look more into it tomorrow

adinn: Congratulations on graduating, Tom! I'll leave Stefan to comment on the patch :-)

tshull226:
Note I don't have access to an Arm machine at the moment so it didn't test the new update. Also, I want to change `createPairUnscaledImmediateAddress`. This method isn't being used anywhere now, but it isn't performing the correct size check for stp/ldp instructions

XiaohongGong:
> Note I don't have access to an Arm machine at the moment so it didn't test the new update.

Hi @tshull226 maybe @a74nh or I could help to do the test on arm machine if you need the help. Thanks!

sanzinger:
Hi @teshull @graalvmbot  has seen your contributions are now with your private email address. You have already signed [OCA list](https://www.oracle.com/technetwork/community/oca-486395.html). I'll whiltelist your email.

teshull:
Thanks @sanzinger. Too, I've been meaning to put together a more thorough fix, but have been busy with other tasks. I'll try to find time to do so soon.

teshull:
I ended up doing this a different way (https://github.com/oracle/graal/commit/eaf58f096727bad1c3fda677efe6bf902fbd6b1e), but now immediate operands cannot be silently truncated

PR #1903, opened 2019-11-28
Fixed AArch64 relocation and patching mistakes in SVM.

tshull226:
Currently, in SVM relocation is only enabled for adrp/ldr,adrp/add, and
sequences of moves. However, this limitation was previously not always
observed in SVM code.

Also fixed an error in the backend where it was loading an address into
a register, instead of loading the value at the address

Change-Id: I5b4999e73cbd926c4e877f5f43803a08ed7be5b3

### DIFF
--- a/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
+++ b/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2020, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2018, Red Hat Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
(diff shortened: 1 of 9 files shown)


tshull226:
Hi. What is the status of merging this pull request?

tshull226:
Note that i just added more relocation fixes as well as corrected patching logic for AArch64

sanzinger:
Thank you for the contribution. This code looks good to me. I would go and merge this to master and then your second PR with runtime code installation. Then we can think of  putting my suggestions with patching on top of it. 

At the moment, tests are running. If everything goes well we can merge this PR.

tshull226:
Sounds good. Let me know if the patch needs any changes

tshull226:
Hi @sanzinger. Given that it is unlikely I will be able to get the make the suggested changes to #1977 and get past internal reviews before the holidays deadlock everything, maybe it is best to merge this PR by itself first?

sanzinger:
I have prepared this PR already for merging. I will try to get this PR in today.

tshull226:
Great! Thank you.

sanzinger:
Hi, unfortunately I did not get the requested internal reviews to merge this PR before christmas. Anyway, now it is in conflict with I have merged this afternoon. Let me fix the conflicts.

sanzinger:
The change 58438f91fbe026f24878f268b2058b6368d0d905 introduces some optimization when loading constants. The 64 bit constant loading may work, but the 32 bit version with `movn` needs some additional thought. The simplest way to solve this is when annotation is required for `mov32`, we just not use the `movn` instruction.

tshull226:
Yes, I just saw this. To compensate, I just added new state to indicate whether within a sequence of moves which are skipped, which are negated, and which are used normally.

sanzinger:
FYI, I have fixed the compilation error. Tests are running. When they are finsihed the merge should happen over night.

tshull226:
Great! Thank for fixing the issue and getting this merged!

PR #1926, opened 2019-12-04
Explain missing stack traces in polyglot exceptions

chrisseaton:
People in practice seem to get very confused why some polyglot exceptions can't print a stack trace.

From:

```
org.graalvm.polyglot.PolyglotException: org.truffleruby.language.control.ReturnException
Original Internal Error: 
org.truffleruby.language.control.ReturnException
```

To:

```
org.graalvm.polyglot.PolyglotException: org.truffleruby.language.control.ReturnException
Original Internal Error: 
org.truffleruby.language.control.ReturnException
(there is no stack trace because the error is a control flow exception)
```

@chumer

### DIFF
--- a/truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotExceptionImpl.java
+++ b/truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotExceptionImpl.java
@@ -51,6 +51,7 @@ import java.util.ListIterator;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 
+import com.oracle.truffle.api.nodes.ControlFlowException;
 import org.graalvm.polyglot.Engine;
 import org.graalvm.polyglot.PolyglotException;
 import org.graalvm.polyglot.PolyglotException.StackFrame;
@@ -254,6 +255,9 @@ final class PolyglotExceptionImpl extends AbstractExceptionImpl implements com.o
             if (isInternalError()) {
                 s.println("Original Internal Error: ");
                 s.printStackTrace(exception);
+                if (exception instanceof ControlFlowException) {
+                    s.println("(there is no stack trace because the error is a control flow exception)");
+                }
             }
         }
     }


chrisseaton:
@chumer can someone be assigned for this PR, please?

chumer:
I don't quite understand why this is necessary. Internal errors showing up as PolyglotException are *always* a bug. See https://www.graalvm.org/sdk/javadoc/org/graalvm/polyglot/PolyglotException.html#isInternalError--

ReturnException control flow exceptions should probably not be shown to the user. If these are exceptions that need to be shown to the user, then they must implement TruffleException and return false in TruffleException#isInternal. I suspect that such exceptions should be caught in the language impl and the return value should be returned, do I misunderstand?

chrisseaton:
Yes, seeing this exception at all is a failure case, but it's a failure case that seems to pop up reasonably often, and this addition makes it more clear why the information is so limited when it does pop up.

For example we have a bug in TruffleRuby at the moment where the compiler omits a catch for a `ReturnException` when it should not, so the user sees this error. The user dealing with it couldn't understand why there was no context. This improves that.

@fniephaus you thumbs-upped - did you see a need for this as well?

fniephaus:
Our students have encountered some of these situations and had problems providing a stack trace when there was indeed no stack trace printed.

I wonder if it would make sense to let ControlFlowExceptions return a dummy StackTraceElement instead (I think that's what they're called; I can't check on mobile). This way the message would show up in all places that print the stack trace.

chumer:
Do you think we should also add a note that internal errors are bugs and encourage users to report them them? Or is the current message understood by users?

Feels weird to special case ControlFlowExceptions for this. It would be reasonable to have other internal errors without stack traces. Why not print a message in general if there is no stack trace? E.g. I want to make interop exceptions stack trace less eventually in the future.

chrisseaton:
> Do you think we should also add a note that internal errors are bugs and encourage users to report them them? Or is the current message understood by users?

Yes that'd be useful. I added this to TruffleRuby: https://github.com/oracle/truffleruby/pull/1851/commits/c0777b06361ff4a3953b69caa195e757c292c57c#diff-6a7c6bcf636a4b656b7fcc59078aaf6eR237

> Feels weird to special case ControlFlowExceptions for this. It would be reasonable to have other internal errors without stack traces. Why not print a message in general if there is no stack trace? E.g. I want to make interop exceptions stack trace less eventually in the future.

Yes could make it more general. This is where I see the error in practice so I've added it for that.

> The idea with a single stack trace element sounds also interesting, but currently we don't have source information for most of the ControlFlowExceptions because we want to keep them to be singletons.

Couldn't the singletons have the single stack trace element as well? It'd be synthetic anyway so has no meaning that could come from an instance.

chumer:
> Couldn't the singletons have the single stack trace element as well?

This feels like a hack, as it never will represent a real location. When other languages parse stack traces they should not see such dummy stack trace entries, so we would need to filter there again. I'd rather go with printing if the stack trace is empty.

boris-spas:
my 2 cents: Who is the "user" in this scenario? If it's e.g. a ruby developer the message being printed is not appropriate (what is a control flow exception?)
```
(there is no stack trace because the error is a control flow exception)
```
It should just say that it's an internal error and that it should be reported to the developers

chrisseaton:
> my 2 cents: Who is the "user" in this scenario?

Someone implementing a Truffle language, or someone testing a Truffle language.

> It should just say that it's an internal error and that it should be reported to the developers

That's less information, and less useful for people reporting and investigating the problem.

boris-spas:
So I think the main thing stopping this is the fact that it's not general enough. As @chumer said:

> Feels weird to special case ControlFlowExceptions for this. It would be reasonable to have other internal errors without stack traces. Why not print a message in general if there is no stack trace? E.g. I want to make interop exceptions stack trace less eventually in the future.

chumer:
Let us add a message whenever there is no stack frames + another note about internal error messages should be reported as a bug.

chrisseaton:
I'm going to close this. I'll just keep it in my local branch if it's not general enough for upstreaming.

chumer:
Afaik @boris-spas is working on a general solution for this.

PR #1936, opened 2019-12-05
Re-enable unittest StringIndexOfTest for AArch64

pli-arm:
The Graal compiler unittest for String.indexOf was disabled for AArch64 before because of the bug JDK-8215792. Now the issue has been fixed and the fix has been backported. We should re-enable the test now.


pli-arm:
Hi @sanzinger , could you help review this trivial change?

sanzinger:
Looks good to me. Will test and integrate it this week.

PR #1960, opened 2019-12-09
Use a darwin AMD64 based UContextRegisterDumper in case of darwin AArch64

johanvos:
This patch was mentioned a couple of months ago by someone from the GraalVM team.
It is sub-optimal since it is not providing a real solution for a register dump on iOS, and it uses the compilerBackend as the condition between using a real registerDumper or a fake one.
On the other hand, this functionality makes little sense on iOS, so from a functional point, this patch seems ok.
Without this patch, native-image fails to compile applications, as there is no Signal$ucontext_t method for Platform.DARWIN_AARCH64.class and neither using the LINUX_AARCH64 or the DARWIN_AMD64 platform will do (since either the architecture or the OS is wrong).

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/aarch64/AArch64UContextRegisterDumper.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/aarch64/AArch64UContextRegisterDumper.java
@@ -24,7 +24,6 @@
  */
 package com.oracle.svm.core.posix.aarch64;
 
-import com.oracle.svm.core.SubstrateOptions;
 import org.graalvm.compiler.api.replacements.Fold;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.Platform;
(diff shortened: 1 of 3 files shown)


substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/aarch64/AArch64UContextRegisterDumper.java
```diff
@@ -50,7 +51,26 @@
 class AArch64UContextRegisterDumperFeature implements Feature {
     @Override
     public void afterRegistration(AfterRegistrationAccess access) {
-        ImageSingletons.add(UContextRegisterDumper.class, new AArch64UContextRegisterDumper());
+        UContextRegisterDumper regdumper;
+        if (SubstrateOptions.CompilerBackend.getValue().equals("llvm")) {
```
vjovanov: @loicottet is this still necessary? Did we fix the segfault handler for AArch64 on LLVM?

loicottet: I think register dumping should work fine on ARM, there's no need to disable it for LLVM.

johanvos: It works on Linux AArch64, and it works on Darwin AMD64, but not on Darwin AArch64

vjovanov: I guess the check should verify if we are running on Darwin for iOS (separate platform or not) instead of the LLVM backend then.

vjovanov: Also, if it is not too hard I would simply implement this. Sooner or later there will be an issue that requires this dumper to be present: we had the same situation on DARWIN_AMD64 a few months ago.

johanvos: will do.

johanvos: I reverted the previous approach, and now have using DARWIN_AARCH64 the same as DARWIN_AMD64. The names of the context fields are similar between Darwin arm64 and amd64. I had to narrow the Platform to only LINUX_AARCH64 in the AArch64UContextRegisterDumper.

PR #1965, opened 2019-12-10
Fixing debug usage performance defect in GraalDebugHandlersFactory

jkoppel:
According to my inspections: The modified code invokes node.toString, though this is wasted effort when debug logs are not enabled.

This is a direct violation of the policy laid out in https://github.com/oracle/graal/blob/a9db7dd44c3a6a37f6805bc93e7c366a201d2e9a/compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/VerifyDebugUsage.java . However, the checker misses this code because of the indirection. It is, however, caught by our Yogo tool, which can search for patterns in a way that does not care about indirection. 

I unfortunately did not see an easy fix other than the explicit if-check.

Thanks to @thomaswue for directing us to VerifyDebugUsage.

### DIFF
--- a/compiler/src/org.graalvm.compiler.printer/src/org/graalvm/compiler/printer/GraalDebugHandlersFactory.java
+++ b/compiler/src/org.graalvm.compiler.printer/src/org/graalvm/compiler/printer/GraalDebugHandlersFactory.java
@@ -74,14 +74,16 @@ public class GraalDebugHandlersFactory implements DebugHandlersFactory {
     private static class NodeDumper implements DebugDumpHandler {
         @Override
         public void dump(DebugContext debug, Object object, String format, Object... arguments) {
-            if (object instanceof Node) {
-                Node node = (Node) object;
-                String location = GraphUtil.approxSourceLocation(node);
-                String nodeName = node.toString(Verbosity.Debugger);
-                if (location != null) {
-                    debug.log("Context obj %s (approx. location: %s)", nodeName, location);
-                } else {
-                    debug.log("Context obj %s", nodeName);
+            if (debug.isLogEnabled()) {
+                if (object instanceof Node) {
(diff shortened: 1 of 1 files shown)


dougxc:
Hi @jkoppel , thanks for the PR. Indeed this is unnecessary overhead when both `-Dgraal.Dump` and `-Dgraal.Log` are specified. I can merge your change if you sign the OCA.

jkoppel:
Awesome! I'll get to signing  that within the day.

Our tool found lots of similar problems within the SubstrateVM code. It appears VerifyDebugUsage has not run on that. Is there a reason for this, or should we fix similar issues within that segment of this codebase?

dougxc:
I'm surprised that this is not over SVM code given the existence of `com.oracle.svm.enterprise.hosted.test.CheckSVMInvariants.SVMInvariantsTool` (/cc @davleopo ). I'd be interested to know which SVM classes are not being checked.

jkoppel:
OCA signed

jkoppel:
For SubstrateVM: The first two alarms I looked at were because our tool checks UserError.guarantee,  but VerifyDebugUsage only checks GraalError.guarantee, though perhaps it should also check UserErrr.guarantee

dougxc:
Good suggestion. Please let me know once your name shows up at [OCA](http://www.oracle.com/technetwork/community/oca-486395.html) and I will integrate this PR and add the test for `UserError.guarantee` in the process.

jkoppel:
@dougxc  Well, the processing took a long time, but it's done. Ready to merge.

jkoppel:
@dougxc  What's preventing merge?

dougxc:
Transient failures in internal testing. I've restarted the attempt to merge.

PR #1970, opened 2019-12-11
Fix typo in Javadoc-link and merge common cases during bytecode processing

jjfumero:


compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java
```diff
@@ -1148,8 +1148,8 @@ protected void cleanupFinalGraph() {
     }
 
     /**
-     * Creates the frame state after the start node of a graph for an {@link IntrinsicContext
-     * intrinsic} that is the parse root (either for root compiling or for post-parse inlining).
+     * Creates the frame state after the start node of a graph for an {@link IntrinsicContext}
+     * intrinsic that is the parse root (either for root compiling or for post-parse inlining).
```
dougxc: Why this fix? The original is perfectly valid according to https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html

compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java
```diff
@@ -5264,8 +5264,10 @@ public final void processBytecode(int bci, int opcode) {
             case IF_ICMPLE      : genIfSame(JavaKind.Int, Condition.LE); break;
             case IF_ACMPEQ      : genIfSame(JavaKind.Object, Condition.EQ); break;
             case IF_ACMPNE      : genIfSame(JavaKind.Object, Condition.NE); break;
-            case GOTO           : genGoto(); break;
-            case JSR            : genJsr(stream.readBranchDest()); break;
+            case GOTO           : // fall through
```
dougxc: This switch statement is ordered by opcode number. I don't see this change as bringing enough value to justify changing the ordering.

compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java
jjfumero: Ok, my version of IntelliJ did not recognize that as a link. I thought was a mistake

jjfumero:
I think this PR can be closed though

PR #1977, opened 2019-12-11
AArch64 runtime code installation

tshull226:
Added support for AArch64 runtime code installation. This required many
changes, including:

1) New "Synchronized" Safepoint States
After new code is installed, on AArch64 each thread is required to issue an instruction to synchronize its instruction cache. For this to happen, new safepoint states have been added to signal that this synchronization action must be performed before returning to Java code.

2) Support for Self Modifying Code Installation
In addition to the thread coordination, installing new code requires it to be flushed to the "Point of Unification". Added runtime support for this.

3) Correct Runtime Code Installation
Before the runtime code installation for determining direct call patching and trampolines was very AMD64 specific. I abstracted support until two platform-specific classes and added AArch64 support.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/RuntimeCodeCache.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/RuntimeCodeCache.java
@@ -64,7 +64,7 @@ public class RuntimeCodeCache {
         public static final RuntimeOptionKey<Boolean> TraceCodeCache = new RuntimeOptionKey<>(false);
 
         @Option(help = "Allocate code cache with write access, allowing inlining of objects", type = OptionType.Expert)//
-        public static final RuntimeOptionKey<Boolean> WriteableCodeCache = new RuntimeOptionKey<>(null);
+        public static final RuntimeOptionKey<Boolean> WriteableCodeCache = new RuntimeOptionKey<>(false);
     }
 
     private final RingBuffer<CodeCacheLogEntry> recentCodeCacheOperations = new RingBuffer<>(30, () -> new CodeCacheLogEntry());
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java
@@ -394,9 +394,6 @@ public class NativeImageGenerator {
                 architecture = new AMD64(features, SubstrateTargetDescription.allAMD64Flags());
             }
             assert architecture instanceof AMD64 : "using AMD64 platform with a different architecture";
-            if (RuntimeCodeCache.Options.WriteableCodeCache.getValue() == null) {
-                RuntimeOptionValues.singleton().update(RuntimeCodeCache.Options.WriteableCodeCache, false);
-            }
             boolean inlineObjects = SubstrateOptions.SpawnIsolates.getValue() && RuntimeCodeCache.Options.WriteableCodeCache.getValue();
             int deoptScratchSpace = 2 * 8; // Space for two 64-bit registers: rax and xmm0
             return new SubstrateTargetDescription(architecture, true, 16, 0, inlineObjects, deoptScratchSpace);
@@ -410,9 +407,6 @@ public class NativeImageGenerator {
                 architecture = new AArch64(features, EnumSet.noneOf(AArch64.Flag.class));
             }
             assert architecture instanceof AArch64 : "using AArch64 platform with a different architecture";
-            if (RuntimeCodeCache.Options.WriteableCodeCache.getValue() == null) {
-                RuntimeOptionValues.singleton().update(RuntimeCodeCache.Options.WriteableCodeCache, false);
-            }
             boolean inlineObjects = SubstrateOptions.SpawnIsolates.getValue() && RuntimeCodeCache.Options.WriteableCodeCache.getValue();
             int deoptScratchSpace = 2 * 8; // Space for two 64-bit registers.
             return new SubstrateTargetDescription(architecture, true, 16, 0, inlineObjects, deoptScratchSpace);


tshull226:
Note that this PR relies on PR #1903. Only the last commit contains the runtime code installation support.

compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
```diff
@@ -3002,19 +3004,73 @@ public void dc(DataCacheOperationType type, Register src) {
...
-    void annotateImmediateMovSequence(int pos, int numInstrs) {
-        if (codePatchingAnnotationConsumer != null) {
-            codePatchingAnnotationConsumer.accept(new MovSequenceAnnotation(pos, numInstrs));
+    public static class PatchableCodeAnnotation extends CodeAnnotation {
+        protected PatchableCodeAnnotation(int instructionStartPosition) {
+            super(instructionStartPosition);
+        }
+
+        @SuppressWarnings("unused")
+        void patch(int codePos, int relative, byte[] code) {
```
dougxc: Could this be an `abstract` method instead of throwing `shouldNotReachHere`?

tshull226:
Currently to maintain AArch64 I$ coherence, I am forcing all threads to issue an ISB after a return to Java code whenever new code is installed. However, I am not doing anything when code is invalidated. I believe this should be correct, unless  somewhere calls are being patched to point to the new code. If patching does occur, then more cache maintenance operations are needed.

compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64Assembler.java
tshull226: Done.

tshull226:
I should also note that currently this patch requires object inlining to be disabled. The is default anyway when `RuntimeCodeCache.Options.WriteableCodeCache` is not set. In the future I will work to remove this limitation.

sanzinger:
Hi,
thank you for your contribution. This helps a lot. I was working on the same issue for some time. I have first focussed on the code patcihng.
I don't like the pieces of code, which always implement the bit fizzling code all over again, with the potential to have bugs in it. I was interested to have some kind of specification in the Assembler which describes the patch instructions in a generic way, so it can be used by all code the same way over different platforms. This should also resolve the object inlining case as well.

I'll share the code with you when I have it working.

tshull226:
Hi @sanzinger,
    Thank you for the update. If you want to share your code with me before it's I'd also be happy to help. For AArch64, patching must consider many cases, including:

1. non-contiguous immediates (such as in adr/adrp lo and high bits)
2. different shift sizes of immediates
3. different sizes of immediates
4. some immediates considered signed, others unsigned
5. may consist of multiple instructions to convey 1 value (Mov Sequence)
6. adrp's relative calculation needs special consideration

There's probably other cases I'm unaware of as well.

The `PatcherUtil` I added in AArch64Assembler at least does the bit mangling in one place.


The reason why runtime code installation for object inlining doesn't work is that this inlining is performed via multiple mov instructions, not a single instruction. This breaks assumptions both in the RuntimeCodeInstaller as well as in other components, including:

RuntimeCodeInfoAccess
ReferenceAccessImpl
SubstrateReferenceMap


I *think* this inlined object reference issue is isolated to the runtime installed code, but it does have many implications for this. @christianwimmer, do you know offhand of any other possible implications of code inlined object pointers being stored in multiple instructions as opposed to 1.

sanzinger:
Hi Thomas,

sorry I did not make it last week to share a working version. I am currently working on getting ahead of the backlog here on github.

However, I pushed my immediate version of this generic patcher to my github repo: https://github.com/oracle/graal/compare/master...sanzinger:sa/GR-18864_improve_patching

The goal is to describe formally, how the patching has to be done. This of course has to be done byte-wise. So the most important classes are `Assembler.PatchGroup` and `Assembler.PatchDefinition`.

You can see in the `BitSpecTest.testPatchAdrpLdr` how the multi-instruction patching works. The simplest case is of course, just patching a 2^n bit value flat into memory (`AMD64BaseAssembler.patcher`).

Apart form that, I think this PR looks good. I will review it in detail soon. I am thinking of merging your PRs first, and then update the patching suggestions I have mentioned above.

tshull226:
Hi @sanzinger,
    Thank you for sharing your patching work. Unfortunately, I did not have much time to look at the code today. I should have more time to look through your code and provide comments either Tuesday or Wednesday.

sanzinger:
FYI, I am trying to resolve the merge conflicts.

tshull226:
I'm just looking at this now as well. It looks like some of my code needs to be changed to incorporate the notion of STATUS_IN_VM. Also, I need to think more about how STATUS_IN_VM is transitioning back to STATUS_IN_JAVA to confirm it doesn't need to use synchronized operation. In the meantime, my merged code is located here: https://github.com/tshull226/graal/tree/aarch64_runtime_code_installation_new. I should be able to devote tomorrow afternoon to making sure everything is still working correctly

sanzinger:
I have stopped with the merge. I'll wait on your update.

christianhaeubl:
Thanks for contributing and sorry that I join this review rather late.

It should not be necessary to introduce any new thread states for synchronizing the instruction cache after code installation. I would roughly suggest the following approach as it should be easier to implement & maintain:
- Introduce a new thread local value that holds a list of runnable actions that must be executed after a thread does a transition back to Java.
- In the VM operation that does the code installation, register for every thread an action that synchronizes the instruction cache.
- When a thread does a transition back to Java, check the thread local and execute all registered actions.

I suppose the instruction cache must be synchronized for every CPU core?
- Couldn't a thread that briefly detaches from the VM miss the instruction cache synchronization? E.g., a thread attaches to the VM, executes some Java code, detaches, reattaches and executes Java code again.
- Is it guaranteed that a thread context-switch synchronizes the instruction cache? Otherwise, the explicit instruction cache synchronization would not necessarily be visible on all cores.

tshull226:
Hi @christianhaeubl. Okay, I can switch to triggering ISBs via a new thread local value which is read on a transition back to Java code. In terms of the functionality of the ISB, it is needed to flush the processor pipeline whenever the I$ is updated to reinitiate the fetch process. A context switch should take care of this by itself. I think you are right though that an ISB should be issued for newly attached threads. I will add this as well.

In addition, I have worked some on adding support for AArch64 with object inlining enabled. Basically, I created a notion of split objects (since the reference will be split across multiple mov instructions) and created a new referencemap, encoder/decoder, referenceadjuster methods,  and objectreferencevistor methods to handle this case. I will create an issue before going much further  to make sure others agree with my approach. I hope you can provide your input on this as well.

christianhaeubl:
Thanks.
Yes, feel free to create an issue so that we can give you early feedback on your approach.

sanzinger:
I think we can separate out the first part of runtime code installation/trampoline patching and so on. So we keep the size of the PR smaller.

tshull226:
That works for me. I'll make a new PR for the coordination & code synchronization features and then will update this PR to depend on those changes.

sanzinger:
@tshull226 I have made some changes and used more a composition style since we use subclassing for another purpose. I hope this is fine for you.

Additionally, I had to add the 32 and 16 bit relocations as well.

tshull226:
Your changes look good to me. I just added support for patching adrp/ldr sequences loading 8 and 16 bit values.

sanzinger:
FYI, there was a different version merged. But all the changes should be there.

PR #1992, opened 2019-12-16
Update supported max java version to 15.

XiaohongGong:
The aot and graalunit jtreg tests fail when run with graal after
the java version of the jdk/jdk master being updated to 15. This
is caused by the class file version checking in Graal. It should
be inside the defined java version range. Currently the max java
version is defined to 14, which should be updated to 15.

Change-Id: I4431ceda3ea2f62efdc5ebee3c8ba13c319b4a4d


dougxc:
Thanks @XiaohongGong , I'll integrate this now. Just to be sure, no class file changes were made yet in JDK 15?

XiaohongGong:
> Just to be sure, no class file changes were made yet in JDK 15?

I think currently no class file changes. See https://hg.openjdk.java.net/jdk/jdk/rev/27d52615ba1b. This patch also modify the `MAJOR_VERSION_JAVA_MAX` of graal classfile.

PR #1059, opened 2019-03-12
Use add(extended) to generate array index offset for AArch64.

XiaohongGong:
We want to find a better way to implement array offset calculation.
The old pattern for an int array member loading is:

     and      x4, x3, #0xffffffff
     lsl      x4, x4, #2
     add      x4, x4, #0x10
     ldr      x0, [x2, x4]

And the optimized pattern is:
       
    add      x1, x2, w3, uxtw #2
    ldr      x0, [x1, #16]

Here is a jmh unit test for int array member loading:

```
private final int func(int[] arr) {
        int ret = 0;
        for (int i=0; i < field0; i++) {
            ret += arr[i];
        }
        return ret;
}
```

And the jmh performance results:

```
without this patch       with this patch       units
    1173501.595               948290.420         us/op
```

Here are the additional issues fixed related to this patch:
1. Finetune MarkBasePointers to get better derived reference marking.
  Currently there is an issue when marking live base pointers. The liveness
of the derived reference will decide whether its base pointer is marked.
If one derived reference is dead at some position, its base pointer will be
removed. So if there are two or more derived references that have the same
base pointer, one of them died will make other derived reference cann't
find its base pointer when building the oop map.
  This patch add a map which will record the base pointer and its derived
references. One base pointer can be removed only when all its derived references
are dead.

2. Set RegisterPriority of basepointers according to the STATE_FLAGS when
building its interval.
  As we all know, the constant value will be re-materialized while not be
spilled to stack when there are not free registers. And its location will be
set to invalid after register allocation. So does the constant reference.
So there will be some exceptions when building the oop map, because the compiler
need to save the locations of live references. If a derived reference has a
constant base pointer which is re-materialized, the compiler will not get its
location. So the jvmci will treat the derived reference as a normal reference
because its base part is null, which may cause the gc crash if the derived
reference is not an real object.
  To resolve this issue, I set the register priority of base pointers to "ShouleHaveRegister",
thus the compiler will not materialize the constant base pointer when doing register
allocation. Also the LIRFrameState has a STATE_FLAG for state values, that is
either a register or a stack slot.

3. Fix an issue on ConstantLoadOptimization. When a constant value is replaced with a
new one, we should also check whether this value is a base pointer of a derived reference.
If so, the base pointer of the derived reference should also be replaced with the new value.
Or there will be some issues when doing register allocation.

4. Fix an issue in MatchProcessor, so that we can process platform specific
node as a matchable node.

mx bootstrap passes.

[...]

PR #2006, opened 2019-12-18
MarkBasePointersPhase:  Base pointers do not die at the definition of…

sanzinger:
@XiaohongGong This change is related to point 1 in your PR #1059. It simplifies the `MarkBasePointersPhase`.

This change also outlined an issue in the `ConstantLoadOptimization` where it did not update the basepointers in `LIRKind`. The simplest solution I could find here is to not optimize those cases, where the constant is used as a base pointer. Otherwise it would require to also update all dependent uses, which is not simply possible.

I want to merge this PR independently from #1059.

### DIFF
--- a/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/constopt/ConstantLoadOptimization.java
+++ b/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/constopt/ConstantLoadOptimization.java
@@ -35,6 +35,7 @@ import java.util.Deque;
 import java.util.EnumSet;
 import java.util.List;
 
+import org.graalvm.compiler.core.common.LIRKind;
 import org.graalvm.compiler.core.common.cfg.AbstractBlockBase;
 import org.graalvm.compiler.core.common.cfg.BlockMap;
 import org.graalvm.compiler.debug.CounterKey;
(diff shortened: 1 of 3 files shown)


XiaohongGong:
Hi @sanzinger ,

It's great to just simplify the `MarkBasePointersPhase`. Thank you! And for the `ConstantLoadOptimization` issue, I also fixed it by just replacing the base pointer to the new one (See `compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/constopt/UseEntry.java ` in [1059](https://github.com/oracle/graal/pull/1059) ).  But it seems that this cannot make sure all the dependent usages are updated. So it's better to close the `ConstantLoadOptimization` for constant base pointer like what you did. Thank you! 
So this patch is ok for me. And once it is merged, I will rebase my PR. Thank you!

Xiaohong

sanzinger:
> But it seems that this cannot make sure all the dependent usages are updated. So it's better to close the ConstantLoadOptimization for constant base pointer like what you did. 

The problem is, that updating the basepointer in `LIRKind` is not sufficient. There is only a single instance of LIRKind referenced in all usages. If we introduce new constants, we also need to create a new LIRKind object and put it accordingly into the usages. Which requires to either recreate the usages as well or update the (currently final) field  `Value.valueKind`.

sanzinger:
FYI, the initial version was wrong. It was just disabling the entire optimization. This time it works.

XiaohongGong:
> FYI, the initial version was wrong. It was just disabling the entire optimization. This time it works.

I tested it locally, and the updated version also works well. Thanks!

PR #2016, opened 2019-12-22
Transition --engine options in the Truffle optimization documentation

chrisseaton:
Are we transitioning all these options over to `--engine`? Why don't some of them have `--engine` equivalents? I've only transitioned the documentation that I wrote here - there are more examples over the code base. Should we get on and move to the new style options everywhere and start to warn when people use the old ones? I'd rather one way to do it.

Related to https://github.com/oracle/truffleruby/pull/1863.

### DIFF
--- a/truffle/docs/Optimizing.md
+++ b/truffle/docs/Optimizing.md
@@ -42,17 +42,17 @@ for more details about instrumenting branches and boundaries.
 
 #### Observing what is being compiled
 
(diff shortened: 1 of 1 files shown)


eregon:
@tzezula is currently working on migrating the remaining options to `--engine`.
Once that's done we plan to warn when using old options and of course update the documentation.

PR #2018, opened 2019-12-23
gu would not work under proxy authentication

YaSuenag:
I tried to use `gu` under proxy authentication, but it did not work.

Our proxy has basic authentication, so I set `$http_proxy` and `$https_proxy` to `http://<user>:<passphrase>@<proxy host>:<port>`. But `ProxyConnectionFactory` in GraalVM did not handle it.

My patch would work in our environment when removing `Basic` from `jdk.http.auth.tunneling.disabledSchemes` in `jre/lib/net.properties`.

### DIFF
--- a/vm/src/org.graalvm.component.installer/src/org/graalvm/component/installer/remote/ProxyConnectionFactory.java
+++ b/vm/src/org.graalvm.component.installer/src/org/graalvm/component/installer/remote/ProxyConnectionFactory.java
@@ -26,9 +26,11 @@ package org.graalvm.component.installer.remote;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.Authenticator;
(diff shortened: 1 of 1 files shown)


YaSuenag:
I'm OpenJDK Reviewer (ysuenaga). Is it not enough?

zicklag:
I've just run into this issue with my proxy. Is there any chance of getting this merged? Or otherwise a way to workaround?

fniephaus:
> I've just run into this issue with my proxy. Is there any chance of getting this merged? Or otherwise a way to workaround?

You could download the artifacts manually and install them via `gu -L`.

Apologies that this PR did not get merged in 2019. The good news is that GraalVM updater has been removed and is no longer needed (https://github.com/oracle/graal/issues/6855).

Issue #2017, opened 2019-12-23
Any specific build instructions to build the graal repo on MacOS 10.15.2?


jaikiran:
I just tried a few times to build the graal vm locally using the following command:

```
cd vm
mx --env ce build
```
This keeps failing with:

```
Building com.oracle.truffle.llvm.tests.native with GNU Make... [dependency LLVM_TOOLCHAIN updated]
graal/sulong/tests/com.oracle.truffle.llvm.tests.native/src/callbackLibrary.cgraal/sulong/tests/com.oracle.truffle.llvm.tests.native/src/caller.c::3030::1010::  fatal errorfatal error: : 

            'stdlib.h''stdio.h'  filefile  notnot  foundfound

#include <stdlib.h>#include <stdio.h>

         ^~~~~~~~~~         ^~~~~~~~~

1 error generated.
1 error generated.
make: *** [bin/caller.o] Error 1
make: *** Waiting for unfinished jobs....
make: *** [bin/callbackLibrary.o] Error 1
```
I'm on mac OS 10.15.2 with XCode (latest) installed along with CommandLineTools. `clang -v` shows:
```
Apple clang version 11.0.0 (clang-1100.0.33.16)
Target: x86_64-apple-darwin19.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

dougxc:
Maybe this helps: https://stackoverflow.com/questions/58232595/missing-system-headers-usr-include-on-macos-catalina

jaikiran:
Hello @dougxc, that indeed helped. Thank you very much.

I don't know if it's my lack of knowledge of C ecosystem or if it's a genuine issue which others might run into. If it indeed is something that others should be aware, then adding a note to the readme document would help. So I've opened https://github.com/oracle/graal/pull/2020/ to address it.

fniephaus:
IIRC, this worked for me: 
```bash
SDKROOT=`xcrun --show-sdk-path` mx --env ce build
```

jaikiran:
Closing this, since I no longer run into this, plus the suggestions in this issue already helped me get past this when I first ran into it.

PR #2020, opened 2019-12-23
Add a note to the README to get past potential builds issues on MacOS

jaikiran:
Added a brief note to get past build issues on MacOS which are noted in https://github.com/oracle/graal/issues/2017

### DIFF
--- a/vm/README.md
+++ b/vm/README.md
@@ -43,6 +43,31 @@ In our CI, we build it using:
 - `binutils`: `2.23.2`
 - `llvm`: `3.8` on linux; `4.0.1` on darwin
 
+---
+**NOTE:**
+
+On MacOS 10.15 (Catalina) you might run into build errors of the form:
+
+```
+'stdlib.h' file not found
+
+#include <stdlib.h>
+
+         ^~~~~~~~~~
+
+1 error generated.
+```
+
+Make sure you have installed `Xcode` and `CommandLineTools` on your MacOS and then set the following environment variable to point to the `/usr/include` directory within the `Xcode` installation:
+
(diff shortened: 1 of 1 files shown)


vm/README.md
```diff
@@ -43,6 +43,32 @@ In our CI, we build it using:
 - `binutils`: `2.23.2`
 - `llvm`: `3.8` on linux; `4.0.1` on darwin
 
+---
+**NOTE:**
+
+On certain versions of MacOS you might run into build errors of the form:
```
dougxc: Please specify which versions. This seems specific to 10.15 (Catalina).

vm/README.md
```diff
@@ -43,6 +43,32 @@ In our CI, we build it using:
...
+
+```
+'stdlib.h''stdio.h'  filefile  notnot  foundfound
```
dougxc: That looks like interleaved output. Can you please add `--serial` to the `build` command to get non-interleaved output.

vm/README.md
jaikiran: @dougxc, thank you for the review. I've updated the PR with the change.

vm/README.md
jaikiran: Done. PR updated with more clear error message.

fniephaus:
Does ```SDKROOT=`xcrun --show-sdk-path` mx --env ce build``` work for you?

jaikiran:
> Does SDKROOT=`xcrun --show-sdk-path` mx --env ce build work for you?

@fniephaus, I just gave that a try and this command worked too. Thank you.

PR #2046, opened 2019-12-30
Ignore /sdk/latest_graalvm and /sdk/latest_graalvm_home

chrisseaton:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,5 @@ workingsets.xml
 bench-results.json
 jmh_result.json
 /vm/src/installer/dist/
+/sdk/latest_graalvm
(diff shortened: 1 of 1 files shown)


chrisseaton:
@eregon

eregon:
It's already ignored in https://github.com/oracle/graal/blob/32bcf03dbf8aa555a714f1a115a4d4667beeff3f/sdk/.gitignore, do you see it as untracked in git?

However, those should start with a `/` like in `vm/.gitignore`, could you update the PR with that?

chrisseaton:
Looks like this was fixed upstream then. I think this was while TruffleRuby was behind on the Graal version so I wouldn't have seen.

PR #2049, opened 2020-01-02
AArch64: Only disable foreign call math intrinsics for native image.

XiaohongGong:
AArch64 specific Math intrinsics are all disabled for svm. Some of them are
foreign calls that don't have the backend implementations for the related
IR on AArch64, like sin/cos/tan/exp/log/log10/pow. For these intrinsics,
the svm actually registers the original Math functions as the foreign call.
And it already has the optimized code sequences for the native methods in
StrictMath.

Howerever, other intrinsics that have the backend implementations are also
disabled, like rint/ceil/floor/fma. These ones can be registered for svm. So
this patch narrows the usage of the registerMathPlugin flag to make it only
disabe the foreign call intrinsics for AArch64.

Change-Id: I9c05b311cdbc56c34917fc321d44494b4b28eefd


XiaohongGong:
Hi there, could anyone please take a look at this PR? Thank you so much!

sanzinger:
This change looks good to me. I am going to integrate it.

XiaohongGong:
> This change looks good to me. I am going to integrate it.

Thanks for your review!

PR #2052, opened 2020-01-06
Add --engine.TraceCompilationCompact option, which fits better in limited terminal space

chrisseaton:
I don't know how big your screens are, but for me on typical development hardware, software, and default sizes of things, `--engine.TraceCompilation` prints lines that are far too long. Rather than being able to easily observe what is being compiled, I just see a wall of characters that is extremely hard to scan.

Screenshot 2020-01-06 at 00 11 43

I've added an option `--engine.TraceCompilationCompact`, which prints just enough to see what methods are being compiled, and lets you track your eye down the list of names much more easily.

Screenshot 2020-01-06 at 00 12 05

@eregon suggested I modify the original option to support something like `--engine.TraceCompilation=compact`. That doesn't seem to be easy, due to having to pair it up to the legacy `--vm.Dgraal` equivalent, it being documented as a stable option, and having to differentiate between not being set, and set to a default, which I'm not sure the current option system supports?

### DIFF
--- a/compiler/src/org.graalvm.compiler.truffle.options/src/org/graalvm/compiler/truffle/options/PolyglotCompilerOptions.java
+++ b/compiler/src/org.graalvm.compiler.truffle.options/src/org/graalvm/compiler/truffle/options/PolyglotCompilerOptions.java
@@ -265,6 +265,9 @@ public final class PolyglotCompilerOptions {
 
     // Tracing
 
+    @Option(help = "Print compact information for compilation results.", category = OptionCategory.EXPERT, stability = OptionStability.STABLE)
+    public static final OptionKey<Boolean> TraceCompilationCompact = new OptionKey<>(false);
+
     @Option(help = "Print information for compilation results.", category = OptionCategory.EXPERT, stability = OptionStability.STABLE)
     public static final OptionKey<Boolean> TraceCompilation = new OptionKey<>(false);
 
--- a/compiler/src/org.graalvm.compiler.truffle.runtime/src/org/graalvm/compiler/truffle/runtime/EngineData.java
+++ b/compiler/src/org.graalvm.compiler.truffle.runtime/src/org/graalvm/compiler/truffle/runtime/EngineData.java
@@ -53,6 +53,7 @@ import static org.graalvm.compiler.truffle.options.PolyglotCompilerOptions.Split
 import static org.graalvm.compiler.truffle.options.PolyglotCompilerOptions.SplittingMaxNumberOfSplitNodes;
 import static org.graalvm.compiler.truffle.options.PolyglotCompilerOptions.SplittingMaxPropagationDepth;
 import static org.graalvm.compiler.truffle.options.PolyglotCompilerOptions.SplittingTraceEvents;
+import static org.graalvm.compiler.truffle.options.PolyglotCompilerOptions.TraceCompilationCompact;
 import static org.graalvm.compiler.truffle.options.PolyglotCompilerOptions.TraceCompilation;
 import static org.graalvm.compiler.truffle.options.PolyglotCompilerOptions.TraceCompilationDetails;
 import static org.graalvm.compiler.truffle.options.PolyglotCompilerOptions.TraceSplitting;
@@ -118,6 +119,7 @@ public final class EngineData {
     @CompilationFinal public boolean multiTier;
     @CompilationFinal public boolean returnTypeSpeculation;
     @CompilationFinal public boolean argumentTypeSpeculation;
+    @CompilationFinal public boolean traceCompilationCompact;
     @CompilationFinal public boolean traceCompilation;
     @CompilationFinal public boolean traceCompilationDetails;
     @CompilationFinal public boolean backgroundCompilation;
@@ -169,6 +171,7 @@ public final class EngineData {
 
         this.returnTypeSpeculation = getPolyglotOptionValue(options, ReturnTypeSpeculation);
         this.argumentTypeSpeculation = getPolyglotOptionValue(options, ArgumentTypeSpeculation);
+        this.traceCompilationCompact = getPolyglotOptionValue(options, TraceCompilationCompact);
         this.traceCompilation = getPolyglotOptionValue(options, TraceCompilation);
         this.traceCompilationDetails = getPolyglotOptionValue(options, TraceCompilationDetails);
         this.backgroundCompilation = getPolyglotOptionValue(options, BackgroundCompilation);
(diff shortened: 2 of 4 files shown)


fniephaus:
Would it help to truncate long method descriptions to something like 50 chatacters? I believe the 2nd log would still not fit into an 80 character terminal.

eregon:
One thing that seems very long here is the absolute paths.
I'm not sure if this is caused by TruffleRuby or this reporting.
I think the CPUSampler uses paths relative to CWD, which are a lot shorter.

chrisseaton:
I *think* Ruby makes the paths absolute. It fixed some problems in the past. This is ancient history and it'd be worth trying to fix now.

eregon:
This is what formats the SourceSection:
https://github.com/oracle/graal/blob/8126984f400fe8410df65200ffbde9347c9e7717/compiler/src/org.graalvm.compiler.truffle.runtime/src/org/graalvm/compiler/truffle/runtime/debug/TraceCompilationListener.java#L175

So it uses Source#getName() which in TruffleRuby is absolute currently (the same as `getPath()` except for the main file). It's kind of needed for that main file case IIRC.

https://github.com/oracle/graal/blob/8126984f400fe8410df65200ffbde9347c9e7717/tools/src/com.oracle.truffle.tools.profiler/src/com/oracle/truffle/tools/profiler/impl/ProfilerCLI.java#L135 is what formats the SourceSection for CPUSampler.
Maybe we should use that for TraceCompilation?

chumer:
> So it uses Source#getName() which in TruffleRuby is absolute currently (the same as getPath() except for the main file). It's kind of needed for that main file case IIRC.

Source#getName() is intended to be the file name not the path, why would you store the path there? The output for the compilation is a lot more reasonable if you do that.

chrisseaton:
> Source#getName() is intended to be the file name not the path, why would you store the path there?

I think we've been storing the full path longer than there was a `Source` API and then we had things depending on it being like that. I will try changing it and see what it breaks.

boris-spas:
To be honest, I don't think this should be merged. 
The long lines issue can be addressed with a simple pipe (e.g. `| less -S +F` or to a file) and adding another option to deal with this seems like overkill.

chrisseaton:
Let me try how it looks again after fixing these Ruby file names.

eregon:
FWIW I think the more compact output is useful, also for end users which e.g. want to check warmup and which of their methods get compiled. For them, there is no value in all the implementation details like Graal Nodes.

Passing it to a pager is always a second-thought thing, so it ends up being impractical (e.g., needs to do a full re-run which might take a while).
My current workaround is to reduce the font size of my terminal and use a 2560px-wide screen, which has obvious drawbacks such as being harder to read.

One can see in the first screenshot from Chris that the output is already too long, even if we removed the `Source` field entirely.

chrisseaton:
I'm going to close this issue for now. I don't have the time to fix Ruby's `getName()` / `getPath()` thing right now.

tzezula:
Regarding the `engine.TraceCompilation=compact` we already have options which have `boolean` value in legacy `graal` mode and enum value in `engine` mode.
An example of such an option is `TrufflePerformanceWarningsAreFatal` which maps to `PerformanceWarningsAreFatal`engine option.

eregon:
Another issue shown in the screenshots is that the first column seems to also include the source path when I think it's expected it doesn't (should be just module+method name). I'll check with other languages. Tracked in https://github.com/oracle/truffleruby/projects/2#card-34932122

eregon:
One reason TruffleRuby cannot simply use `Source#getPath()` is paths are not supposed to be canonicalized, e.g., in `$LOADED_FEATURES`. So we use Source#getName() to basically store the path expanded according to Ruby semantics. We'll probably need to store that another way to use the default `Source#getName()` for Ruby files.
Maybe Truffle shouldn't force to canonicalize the path, and rather just make it an absolute path if relative, and let the caller decide if it wants realpaths or just absolute paths.

Also the main file, given to `ruby foo/bar.rb` should have a path of `foo/bar.rb` in backtraces, for which neither Source#getPath() nor the default Source#getName() works. That one we can handle as a special case as there is only one main file (at least when using RubyLauncher).

eregon:
With https://github.com/oracle/truffleruby/commit/e276e66f38306842602833fb2357a9416580c889 I changed Source#getName() in Ruby to be the filename, so the output is quite a bit shorter now:
```
Before:
[engine] opt done         block in Optcarrot::ROM#initialize /home/eregon/code/truffleruby-ws/truffleruby/bench/optcarrot/lib/optcarrot/rom.rb:93 <opt> <split-4a177e33>|ASTSize      13/   13 |Time     2(   1+1   )ms |DirectCallNodes I    0/D    0 |GraalNodes    28/   85 |CodeSize          241 |CodeAddress 0x7fa768000000 |Source /home/eregon/code/truffleruby-ws/truffleruby/bench/optcarrot/lib/optcarrot/rom.rb:93
[engine] opt done         block in Optcarrot::CPU#add_mappings /home/eregon/code/truffleruby-ws/truffleruby/bench/optcarrot/lib/optcarrot/cpu.rb:117 <opt> <split-617a120e>|ASTSize      40/   72 |Time    91(   5+86  )ms |DirectCallNodes I    2/D    0 |GraalNodes   263/  536 |CodeSize         1553 |CodeAddress 0x7fa769454000 |Source /home/eregon/code/truffleruby-ws/truffleruby/bench/optcarrot/lib/optcarrot/cpu.rb:117
After:
[engine] opt done         block in Optcarrot::ROM#initialize <split-7ecbd25c>         |ASTSize      13/   13 |Time    19(  17+1   )ms |DirectCallNodes I    0/D    0 |GraalNodes    28/   63 |CodeSize          261 |CodeAddress 0x7f2d75d23710 |Source      rom.rb:93
[engine] opt done         block in Optcarrot::CPU#add_mappings <split-2783d935>       |ASTSize      40/   74 |Time    46(  41+4   )ms |DirectCallNodes I    2/D    0 |GraalNodes   265/  349 |CodeSize         1213 |CodeAddress 0x7f2d76045210 |Source     cpu.rb:117
```

eregon:
@norswap made the output of TraceCompilation significantly less wide, so now it should fit a lot better in terminals.
For instance:
```
Before:
[engine] opt done         block in Optcarrot::CPU#add_mappings <split-2783d935>       |ASTSize      40/   74 |Time    46(  41+4   )ms |DirectCallNodes I    2/D    0 |GraalNodes   265/  349 |CodeSize         1213 |CodeAddress 0x7f2d76045210 |Source     cpu.rb:117
After:
[engine] opt done     Kernel#dup                                                  |AST    5|Time    6(   5+0   )ms|Tier 2|Inlined   0Y   0N|IR     5/    5|CodeSize     88|Addr 0x7f78f996d3d0|Src (core):1
```

cc @chrisseaton

chrisseaton:
Cheers

PR #2060, opened 2020-01-09
AArch64: Elide NarrowNode for Extend.

qc1iu:
Extending value from narrow type to a wider one will generate a "sxt" or
"and" instruction, and if its input is a NarrowNode, the narrow could be
eliminated.

        and     w0, w2, #0xffff
        sxth    x0, w0

With this patch, the generated assembly above can be optimized to below:

        sxth    x0, w2

Change-Id: Ic1db0f2b11a9d80ca0fab9b042bedff3de66e95a

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64ElideL2ITest.java
+++ b/compiler/src/org.graalvm.compiler.core.aarch64.test/src/org/graalvm/compiler/core/aarch64/test/AArch64ElideL2ITest.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2019, Arm Limited. All rights reserved.
+ * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020, Arm Limited. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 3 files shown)


sanzinger:
Looks good to me.

gergo-:
This looks good to me in principle, but I'd like to understand more about the context. Does this code pattern occur outside of the synthetic matcher tests? If yes, is it only on AArch64 or on other platforms as well?
`ZeroExtendNode`/`SignExtendNode` already have canonicalization rules for a nested `NarrowNode` (including the possibly hypothetical edge case, not handled here, of the narrow being wider than the extension). I would expect these generic canonicalizations to be sufficient for actual compilations on all platforms. If they aren't, it would be good to understand why not, and to fix things in a target-independent way rather than separately in each backend.

qc1iu:
Thanks for your attention🙇‍.

First, I think this pattern `SignExtendNode(NarrowNode value)` is target-dependent. Eliding the `NarrowNode` is based on the architecture characters. 

For AArch64(from The A64 instruction set), 
> When a 32-bit register form is selected:
> • Writes to the W register set bits [63:32] of the X register to zero.

I'm not familiar with AMD64 and didn't test, perhaps it can also work if it followed similar rules with AArch64. But we can not sure **_all_** the architectures followed these rules.

Besides, in my view, the mid-end IR should only responsible for java semantic and there indeed have some significant differences between mid-end and back-end. 
One thing is the type. Basically we can regard at least the general registers as data containers without type, so it's not hard to understand that we can do some optimizations or simplifies in back-end ignoring type which however is a limitation in mid-end.

All in all, if this patch is worked well, I would like to merge this patch first and give the matter a further thought. And we will keep the focus on this problem.

adinn:
@gergo- Yes, this is an AArch64-specific thing because of the architecture feature noted by @qc1iu. We profit from similar opportunities for this sort of optimization in the Hotspot C2 compiler for AArch64 that are not available for x86_64.

gergo-:
Thanks for the discussion, this makes more sense to me now. We will try to merge this PR soon.

PR #2068, opened 2020-01-13
AArch64 Speculative Execution Mitigation Fix

teshull:
Use full system barrier to mitigate speculative exec.

Originally a CSDB was being used. However, this is not sufficient, as
this instruction must be used in conjuction with a data processing
instruction to mitigate speculative execution of memory accesses. For
now, have switched to using a full system barrier (DSB SY + ISB) to
ensure subsequent memory accesses are not executed speculatively.


tshull226:
Note that only the last commit is the speculative execution changes (I based this on #2022 as they both need ISBs). Once #2022 is merged I will rebase this on the master

tshull226:
You can see https://developer.arm.com/support/arm-security-updates/speculative-processor-vulnerability/download-the-whitepaper for more information about when CSDB can be used vs. DSB SYS + ISB

Issue #2062, opened 2020-01-10
Unsupported java.lang.Class.getConstantPool() is reachable from parseParameterAnnotations


ziyilin:
I got the following complains:
```
Error: com.oracle.svm.hosted.substitute.DeletedElementException: Unsupported method java.lang.Class.getConstantPool() is reachable: The declaring class of this element has been substituted, but this element is not present in the substitution class
To diagnose the issue, you can add the option --report-unsupported-elements-at-runtime. The unsupported element is then reported at run time when it is accessed the first time.
Trace:
        at parsing java.lang.System$2.getConstantPool(System.java:1275)
Call path from entry point to java.lang.System$2.getConstantPool(Class):
        at java.lang.System$2.getConstantPool(System.java:1275)
        at java.lang.reflect.Executable.parseParameterAnnotations(Executable.java:83)
        at com.oracle.svm.reflect.Executable_parseParameterAnnotations_e04e8678cbe5147ec3ca5f53b872cc8ea631d03c_12307.invoke(Unknown Source)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.apache.xml.dtm.DTMException.printStackTrace(DTMException.java:350)
        at org.apache.xml.dtm.DTMException.printStackTrace(DTMException.java:286)
        at com.oracle.svm.jni.functions.JNIFunctions.ExceptionDescribe(JNIFunctions.java:749)
        at com.oracle.svm.core.code.IsolateEnterStub.JNIFunctions_ExceptionDescribe_b5412f7570bccae90b000bc37855f00408b2ad73(generated:0)
```
What is  this class?
> com.oracle.svm.reflect.Executable_parseParameterAnnotations_e04e8678cbe5147ec3ca5f53b872cc8ea631d03c_12307

How it is generated? If it is generated by Substrate VM, it should not call the unimplemented method.

cstancu:
That's a method stub generated as a result of `Executable.parseParameterAnnotations()` being registered for reflection. Do you (or a framework you use) explicitly register this method for reflection? In normal execution `Executable.parseParameterAnnotations()` is never used in SVM, the only method that uses is `Executable.sharedGetParameterAnnotations()` is substituted with a custom implementation.

ziyilin:
I see. We have configured all methods in Executable for reflection:
```
{
  "name":"java.lang.reflect.Executable",
  "allDeclaredMethods":true,
  "allPublicMethods":true
}
```

This could be the reason.

[...]

PR #2070, opened 2020-01-14
Substitute parseParameterAnnotations method

ziyilin:
Original Executable.parseParameterAnnotations can access not supported
method java.lang.Class.getConstantPool. And Executable.parseParameterAnnotations could be reached by reflection calls. 

This patch can fix issue https://github.com/oracle/graal/issues/2062

### DIFF
--- a/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/target/Target_java_lang_reflect_Executable.java
+++ b/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/target/Target_java_lang_reflect_Executable.java
@@ -104,6 +104,13 @@ public final class Target_java_lang_reflect_Executable {
     }
 
     @Substitute
+    @SuppressWarnings({"unused", "hiding"})
+    Annotation[][] parseParameterAnnotations(byte[] parameterAnnotations) {
+        Target_java_lang_reflect_Executable holder = ReflectionHelper.getHolder(this);
+        return ReflectionHelper.requireNonNull(holder.parseParameterAnnotations(parameterAnnotations), "Parameter annotations must be computed during native image generation");
+    }
+
+    @Substitute
     public AnnotatedType getAnnotatedReceiverType() {
         Target_java_lang_reflect_Executable holder = ReflectionHelper.getHolder(this);
         /* The annotatedReceiverType can be null. */


cstancu:
I don't see how this PR would actually work. You are substituting `Executable.parseParameterAnnotations()` with a call to itself. This will eventually result in a stack overflow if this method is ever called. Its original implementation depends on constant pool data, which is not available at run time, that's why we precompute and cache the results where this method is used. The problem is that you are trying to indiscriminately register all methods in `Executable` for reflection. `Executable.parseParameterAnnotations()` is an internal API of `Executable` and it should probably not be accessed via reflection anyway. You should find the methods that your app actually needs and only register those. You can use the native-image-agent to do so.

ziyilin:
Thanks, I'll close the PR. 
However, the reflection configration was generated by agent. I have found the trace to Executable.getDeclaredMethods. This is from the agent's trace file:
```
{
  "caller_class": "org.codehaus.groovy.reflection.stdclasses.CachedSAMClass$1",
  "args": [],
  "function": "getDeclaredMethods",
  "tracer": "reflect",
  "class": "java.lang.reflect.Executable"
},
```
The related source code from groovy-2.5.7:

```
    private static Method[] getDeclaredMethods(final Class c) {
        try {
            Method[] methods = AccessController.doPrivileged(new PrivilegedAction<Method[]>() {
                public Method[] run() {
                    return c.getDeclaredMethods();
                }
            });
            if (methods!=null) return methods;
        } catch (java.security.AccessControlException ace) {
            // swallow and do as if no method is available
        }
        return EMPTY_METHOD_ARRAY;
    }

    private static void getAbstractMethods(Class c, List<Method> current) {
        if (c==null || !Modifier.isAbstract(c.getModifiers())) return;
        getAbstractMethods(c.getSuperclass(), current);
        for (Class ci : c.getInterfaces()) {
            getAbstractMethods(ci, current);
        }
        for (Method m : getDeclaredMethods(c)) {
            if (Modifier.isPrivate(m.getModifiers())) continue;
            if (Modifier.isAbstract(m.getModifiers())) current.add(m);
        }
    }
```
From the above source code we can see the abstract methods in` java.lang.Executable` are added in a list for future usage, while others are ignored. So `parseParameterAnnotations` will be ignored, and we can safely ingore the reach of unsupported feature in this case.
But it is not convenient to do such manual analysis. Is there any approach to automatilize the analysis?

cstancu:
@peter-hofer for a traced call to 
```
{
  "caller_class": "org.codehaus.groovy.reflection.stdclasses.CachedSAMClass$1",
  "args": [],
  "function": "getDeclaredMethods",
  "tracer": "reflect",
  "class": "java.lang.reflect.Executable"
},
```
the agent generates this reflection config
```
{
  "name":"java.lang.reflect.Executable",
  "allDeclaredMethods":true,
  "allPublicMethods":true
}
```
This is problematic because most of the methods in `java.lang.reflect.Executable` are substitued and the left over methods can lead to issues if are reached via reflective calls, by, e.g., calling native methods whose implementation is not available. Can the agent be configured to exclude these methods? This issue could affect other substituted JDK classes too.

peter-hofer:
@cstancu no, we don't have support in the agent for this, and I think this would be best addressed in the image generation process. After all, the user program does request all declared and public methods of `Executable`, and the agent just puts this in the resulting configuration. In my opinion, the nicest way to approach this is that in the image generation process, we enable reflective access to all methods of `Executable` that will exist at runtime and are not deleted with substitutions. If a method is not safe to call, it should throw an exception at runtime. I wonder, which methods do you expect to not work?

cstancu:
The issue that led to this PR is #2062. The problematic method is `java.lang.reflect.Executable.parseParameterAnnotations` which calls a deleted method:
```
Error: com.oracle.svm.hosted.substitute.DeletedElementException: Unsupported method java.lang.Class.getConstantPool() is reachable: The declaring class of this element has been substituted, but this element is not present in the substitution class
To diagnose the issue, you can add the option --report-unsupported-elements-at-runtime. The unsupported element is then reported at run time when it is accessed the first time.
Trace:
        at parsing java.lang.System$2.getConstantPool(System.java:1275)
Call path from entry point to java.lang.System$2.getConstantPool(Class):
        at java.lang.System$2.getConstantPool(System.java:1275)
        at java.lang.reflect.Executable.parseParameterAnnotations(Executable.java:83)
        at com.oracle.svm.reflect.Executable_parseParameterAnnotations_e04e8678cbe5147ec3ca5f53b872cc8ea631d03c_12307.invoke(Unknown Source)
        at java.lang.reflect.Method.invoke(Method.java:498)
...
```
I think we could substitute `java.lang.reflect.Executable.parseParameterAnnotations` with  `throw VMError.unsupportedFeature("Executable.parseParameterAnnotations() is not available at run time.");`.

peter-hofer:
I just checked and `@Delete` on the method(s) should work as well.

PR #2091, opened 2020-01-23
Thermometer

chrisseaton:
At Shopify we're trying to optimise a critical production application running on Truffle at scale.

We've found this *thermometer* to be a useful, practical tool for understanding the optimisation state of a program that is compact enough to scannable quickly, and approachable enough for users who aren't experts in Truffle.

From the documentation (which explains everything in more depth) the user usually sees something like this:

```
[thermometer] INFO:   0.75s  🥶    0°    0.77 MB   0.03 GB   55 ▶  2 ▶  66  (  0, 312 )  15 ▼
[thermometer] INFO:   1.06s  🤔   55°    0.77 MB   0.05 GB   34 ▶  2 ▶  68  (  0, 354 )  15 ▼
[thermometer] INFO:   1.37s  😊   94°    0.77 MB   0.07 GB    0 ▶  0 ▶  69  (  0, 356 )  15 ▼
[thermometer] INFO:   1.72s  😊  100°    0.77 MB   0.01 GB    0 ▶  0 ▶  69  (  0, 356 )  15 ▼
[thermometer] INFO:   2.04s  😊  100°    0.77 MB   0.05 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   2.37s  😊  100°    0.77 MB   0.12 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   2.71s  😊  100°    0.77 MB   0.12 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   3.04s  😊  100°    0.77 MB   0.17 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   3.37s  😊  100°    0.77 MB   0.22 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   3.70s  😊  100°    0.77 MB   0.20 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   4.04s  😊  100°    0.77 MB   0.25 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   4.43s  😊  100°    0.77 MB   0.01 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   4.70s  😊  100°    0.77 MB   0.01 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   5.04s  😊  100°    0.77 MB   0.07 GB    0 ▶  0 ▶  70  (  0, 356 )  15 ▼
[thermometer] INFO:   5.58s  🤮   42°    0.81 MB   0.03 GB    8 ▶  2 ▶ 107  (  0, 368 )  22 ▼
[thermometer] INFO:   5.70s  🤔   74°    0.81 MB   0.18 GB    8 ▶  2 ▶ 107  (  0, 368 )  22 ▼
[thermometer] INFO:   6.07s  😊   91°    0.81 MB   0.06 GB    0 ▶  1 ▶ 111  (  0, 373 )  22 ▼
[thermometer] INFO:   6.37s  😊  100°    0.81 MB   0.12 GB    0 ▶  1 ▶ 111  (  0, 373 )  22 ▼
```

The key:

```
5.58s  🤮   42°    0.81 MB   0.03 GB    8 ▶  2 ▶ 107  (  0, 368 )  22 ▼
  ┃     ┃    ┃       ┃         ┃        ┃     ┃     ┃    ┃   ┃     ┗━ deoptimizations and invalidations
  ┃     ┃    ┃       ┃         ┃        ┃     ┃     ┃    ┃   ┗━━━━━━━ dequeued
  ┃     ┃    ┃       ┃         ┃        ┃     ┃     ┃    ┗━━━━━━━━━━━ failed
  ┃     ┃    ┃       ┃         ┃        ┃     ┃     ┗━━━━━━━━━━━━━━━━ finished
  ┃     ┃    ┃       ┃         ┃        ┃     ┗━━━━━━━━━━━━━━━━━━━━━━ running
  ┃     ┃    ┃       ┃         ┃        ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━ queued
  ┃     ┃    ┃       ┃         ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ used heap (total - free)
  ┃     ┃    ┃       ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ code loaded
  ┃     ┃    ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 'temperature' (see below)
  ┃     ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ indicator (see below)
  ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ wall clock
```

The *temperature* is really the percentage of samples taken during the period
where the top-most Truffle method activation is running in compiled code. Method
preludes set a per-thread flag to indicate whether they're compiled or not. A
separate high-priority timer thread samples this flag.

The indicator is set as follows, in priority order

* 😡 if there was a failure
* 🤮 if there was a deoptimization
* 😊 if temperature is 90 or higher
* 🤔 if temperature is less than 90, or code has been loaded
* 🥶 if temperature is less than 50

You can also create a benchmark from unmodified production code by attaching a counter. For example turning the route you are testing into a benchmark without any extra tools.

`--thermometer.IterationPoint=demo.rb:15` will install an iterations-per-second
counter on any statements at this location. You should ensure there is just one
statement at this location as each statement run will count as an iteration.

```
5.43s  😊  100°   94.382 K i/s    0.81 MB   0.04 GB    0 ▶  1 ▶ 110  (  0, 375 )  23 ▼
                     ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ iterations per second
```

I think that graphing time spent in compiled code alongside the current iterations per second is a novel idea.

The tool can also log this information, and produce graphs, including comparing. For example this is the comparative optimisation characteristic of the hosted and native configurations on an application.

image

As I say, at Shopify we've found this to be a very practical tool for working on optimisation of Truffle languages at scale.


fniephaus:
+1 for integrating it! I've had some time to play around with this and GraalSqueak. It worked right away and the feedback it produces is quite useful to get a feeling if/when a stable state is reached.

chrisseaton:
There's some holes I had to poke through the API to do this - very open to better ways to get the same information.

chrisseaton:
> Missing documentation

Do you mean Javadoc? I thought the readme was pretty comprehensive (includes diagrams!)?

https://github.com/oracle/graal/blob/a764128ef6536048722d9410fefb7f623595fe1d/tools/docs/Thermometer-Manual.md

> Missing tests

Yes if the idea is acceptable I'll go back and write the tests.

> Run style and API checks

I can do that.

> CompilationStateBackdoor and the engine backdoor

> CompilationState API exposed in instrumentation. We should probably try to reduce that somehow and/or make it more generally useful.

Yes - I wasn't sure how to thread this particular thing through the API and went with a hack to get it working. I'm open to a better way of doing it. I did try using the listener system to count this information, but there was some issue - I think you can't tell the difference between a dequeue that has started being compiled from one that hasn't so you can't count the size of the compilation queues correctly.

> The way CompilationRootTag works. in particular if (tag == StandardTags.CompilationRootTag.class && instrumentedNode.getParent() instanceof RootNode) is a rough assumption about the language AST. There is no reason why the first node of the a root node should be instrumentable. Better would be to solve this instrumentation internally in the CallTarget of the runtime implementation. So all languages get CompilationRoot support automatically. Maybe we can even deprecate the RootTag in the end.

Yes if we can do that that'd be great.

> Do you want us to fix the remaining issues, or do you plan to continue working on this?

I'm happy both for other people to change the PR or me to keep doing the work with some direction.

The things I need most help with are the compilation state API and the compilation root tag.

> On other words, can I directly push to this pr?

Hmm now we've got a practical problem here - I'll have to see if I can add collaborators to the public Shopify Graal repo where this branch is.

fniephaus:
> Hmm now we've got a practical problem here - I'll have to see if I can add collaborators to the public Shopify Graal repo where this branch is.

When you create a PR, you can thick a box to allow maintainers to push to the branch of the PR:

image

chumer:
> The things I need most help with are the compilation state API and the compilation root tag.

Ok. I will try to spend some time on it next week. If it gets too big I might split the compilation root tag into a separate PR.
Regarding CompilationRoot: Why didn't you just use the existing RootTag?
@boris-spas  pointed out that the name CompilationRootTag is kind of misleading. As we have a CompilerDirectives.isCompilationRoot() which means something different. So we probably need a new name. (not good idea atm)

Another thing that I don't quite like is the notion of temperature in degrees. I find using 100 degrees as the maximum value confusing. Wouldn't it be more appropriate to show a percentage instead? What do I miss?

Some people might dislike the use of emoticons for encoding/distracting reasons. I don't really have an opinion on that, what is your take?

chrisseaton:
> Regarding CompilationRoot: Why didn't you just use the existing RootTag?

I don't think that covers OSR roots does it?

> Another thing that I don't quite like is the notion of temperature in degrees. I find using 100 degrees as the maximum value confusing. Wouldn't it be more appropriate to show a percentage instead? What do I miss?

I guess it's a big whimsical. The key measurement this tool does is the percentage time in compiled code, which tells you how warmed-up the application is, so it's a temperature.

> Some people might dislike the use of emoticons for encoding/distracting reasons. I don't really have an opinion on that, what is your take?

The reason I used the emoji was so that it's more easily scannable to get a general impression of the health of the application's optimisation. The colour is the key bit (for systems that choose to render emoji in colour, which is most).

The arrow characters are also non-ASCII.

Something that I'm try to battle is oceans of similar looking ASCII characters in these traces. How can we make it more understandable and scannable while making it as compact as possible.

I'm open to suggestions on the temperature  ° and the Unicode characters as long as people are saying how they're going to keep it as usable as possible.

chumer:
> I don't think that covers OSR roots does it?

True. Is that the only reason? Could we go forward with this PR reusing the RootTag, fixing this issue in a separate one? That would speed up the integration of this.

> I'm open to suggestions on the temperature °

I think we can keep the temperature label, but maybe change the ° to %? I could infer that ° means % here, but maybe others can't? Other opinions?

> and the Unicode characters

I don't have an opinion. I'll let @boris-spas  comment.

chrisseaton:
> Is that the only reason? Could we go forward with this PR reusing the RootTag, fixing this issue in a separate one? That would speed up the integration of this.

Yes. I'd just note that an additional limitation is that OSR may be missed.

> I think we can keep the temperature label, but maybe change the ° to %? I could infer that ° means % here, but maybe others can't? Other opinions?

I did also think that ° as in celsius might confuse Americans, so I'm ok to change after letting any other opinions be heard.

chumer:
While I like the compact representation of the thermometer. Could it be useful to print a thermometer legend info message that explains each column when the application is starting up? I will forget what each column means immediately.

chrisseaton:
> Could it be useful to print a thermometer legend info message that explains each column when the application is starting up?

I've had that feedback from other users in practice, so ok.

fniephaus:
+1 for using %.

I kind of like the use of emoticons. They make the tool more fun, which I consider a plus, especially when I think about our students and using the tool for educational purposes.

chrisseaton:
This is a good example of the kind of information you can see for a change using this tool.

https://github.com/oracle/truffleruby/issues/1398#issuecomment-580371896

chrisseaton:
I hope to get this mergeable soon, but may take me a few weeks go through it. I'll also keep adding new functionality that we're using at Shopify.

PR #2096, opened 2020-01-25
Explain why implicitExceptionBytes may be null

luhenry:
Because not all versions of the JDK have the field aot_metadata._nul_chk_table_begin, you do not always
want to generate this field depending on the target JDK's version. This difference is taken care of by
HotSpotGraalServices which has different implementations based on the target JDK's version. The
HotSpotGraalServices.getImplicitExceptionBytes returns null in case the field does not exist.

Additionally, harmonize how exceptionBytes is generated with the other variables.

This change was previously discussed in the graal-dev mailing list (https://mail.openjdk.java.net/pipermail/graal-dev/2020-January/005935.html)


luhenry:
I'm covered by the OCA as I am a Microsoft employee.

luhenry:
Hey, quick follow up on this PR. Is there anything you like/don't like? Any approach you would prefer to take? Thank you!

dougxc:
It is currently in the process of being merged. The merge queue is a little backed up but it should be integrated within 24 hours.

dean-long:
@dougxc Is this ready to merge now?  (I don't see to have permission.)

PR #2102, opened 2020-01-27
[truffle] Make three TruffleLanguage methods public

fniephaus:
Language implementers need to use these three methods and because they are `protected`, they are rather hard to find. And it's not possible to override visibility in Java.

/cc @eregon @chumer

### DIFF
--- a/truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
+++ b/truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleLanguage.java
@@ -1282,7 +1282,7 @@ public abstract class TruffleLanguage<C> {
      * @see com.oracle.truffle.api.dsl.CachedLanguage
      * @since 0.27
      */
-    protected static <T extends TruffleLanguage<?>> T getCurrentLanguage(Class<T> languageClass) {
+    public static <T extends TruffleLanguage<?>> T getCurrentLanguage(Class<T> languageClass) {
         return LanguageAccessor.engineAccess().getCurrentLanguage(languageClass);
     }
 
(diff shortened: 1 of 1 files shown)


boris-spas:
To be honest "they are rather hard to find" seems like a poor reason to increase visibility.

fniephaus:
Most of my nodes have access to the language, and yet I almost decided to use `HomeFinder` because I couldn't find `getLanguageHome()`:
image

It was @eregon who told me about the API, and I think he also agrees that some of these methods should be public. I don't see a reason why a language implementer shouldn't have access to the language's home directory for example.

boris-spas:
I think my objection still holds.

Having no public methods is a very clear message that this class is not meant to be used for object that does things. Having some public methods might give the wrong impression that one should try to get an instance of TruffleLanguage and do things with it. It's an API design thing.

@chumer Thoughts?

chumer:
The reasons they are protected is because it is intended to be used by subclasses only. 
If a language has intent to expose them then they can create a public method themselves. The idea is that the language is in control whether it wants to do more things to resolve the language home. If we would make this public by default the API would be redundant, and people might use the wrong one (particularly applies to getLanguageHome). There is also the case that a language does not want to expose a static context lookup at all.

In SL we decided to expose it[1], but it is actually only used in tests. So we could get rid of it.
[1] https://github.com/oracle/graal/blob/master/truffle/src/com.oracle.truffle.sl/src/com/oracle/truffle/sl/SLLanguage.java#L405

Accessing the language or the context using getCurrentLanguage and getCurrentContext is discouraged when using Nodes. You should use `@CachedContext` and `@CachedLanguage` instead.

eregon:
I think from a usability point of view of Truffle, this is annoying and hard to find, especially for `getLanguageHome()`. I would think most Truffle language developers which are not also Truffle developers would agree with this.

I don't see any risk, because it's impossible to get an instance of TruffleLanguage itself anyway (it's an abstract class), only of one of the subclasses.

Many times, we spent too much time to find those hidden protected methods, and it feels so silly to just redefine them as public in every language.

Also, since `getLanguageHome()` is `final` it means it can't be overridden as public with the same name, so one needs to come up with a new silly name like `getTruffleLanguageHome()`.

A typical example is the context needs access to the language home. And in every such case one has to guess where to find the method (hidden in the IDE as shown by Fabio), declare it public with another name and finally be able to use it in a very sensible place.

getCurrentLanguage()/getCurrentContext() need a Class argument, so in practice redefining them allows to remove that extra argument, and it's less necessary to have them `public`.

> There is also the case that a language does not want to expose a static context lookup at all.

What kind of use-case would that be? It's on the language implementer side, of course the context is used in many places, including slow paths which don't have nodes.

Issue #2111, opened 2020-01-29
getPackageInfo should not attempt to load classes


galderz:
Master breaks one of quarkus test (see [quarkus#6685](https://github.com/quarkusio/quarkus/issues/6685)).

The relevant part is:
```
Caused by: com.oracle.svm.core.jdk.UnsupportedFeatureError: JDK11OrLater: Target_java_lang_ClassLoader.loadClass(Target_java_lang_Module, String)
	at com.oracle.svm.core.util.VMError.unsupportedFeature(VMError.java:101)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:247)
	at java.lang.Package.getPackageInfo(Package.java:420)
	at java.lang.Package.getAnnotations(Package.java:467)
	at org.eclipse.yasson.internal.AnnotationIntrospector.collectAnnotations(AnnotationIntrospector.java:757)
```

One possible fix would be to add a substitution in `getPackageInfo` to avoid ending attempting to load classes. A PR follows up.

eginez:
related to https://github.com/oracle/graal/issues/2093

galderz:
Thanks @eginez! Sorry I missed the original issue ☺️

eginez:
Fixed by 871f978

PR #2112, opened 2020-01-29
Substitute package info to avoid attempting to load classes #2111

galderz:
Fix for #2111

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
@@ -733,6 +733,22 @@ final class Target_java_lang_Package {
             return null;
         }
     }
+
+    @Alias private Class<?> packageInfo;
+
+    @Substitute
+    @TargetElement(onlyWith = JDK11OrLater.class)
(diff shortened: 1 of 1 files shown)


galderz:
I work for Red Hat, I should be covered under Red Hat's OCA.

dmlloyd:
I suspect the code will fail the style gate, but I don't recall the command to check it...

galderz:
@dmlloyd indeed it breaks, looking...

galderz:
Fixed `mx eclipseformat` recommendations.

dmlloyd:
Still fails: https://travis-ci.org/oracle/graal/jobs/643518329#L6412

galderz:
Fixed it once more, let's see...

eginez:
@galderz  This should no longer be necessary after: https://github.com/oracle/graal/commit/871f9784f3d7df3556f9388187cdabcdb53751d0

If `package-info` has been registered for reflection the class will now be found in the image.

dmlloyd:
That's interesting.  We discussed and dismissed such a change on the basis that JPMS support is already incomplete, and the result could be incorrect as a result.  But, fixed is fixed I guess! :slightly_smiling_face:

eginez:
Completely understandable. I am closing this PR and its accompanying issue

PR #2113, opened 2020-01-29
[native-image]  AArch64: SVM inlined object code installation

teshull:
Added support for inlined objects within runtime installed code.

In AArch64 these objects are represented through a sequence of mov instructions and hence the reference is not contiguous within the code. To handle this, had to add multiple components, including:
    
1. Means of adjusting split references
2. Ability to encode split references within reference maps
3. Ability to walk split references in visitors

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/GCRefAccessor.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/GCRefAccessor.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2020, 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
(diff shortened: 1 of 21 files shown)


tshull226:
Note that I have a second commit which is a hack around what I think is an issue in SVM's analysis. In `AArch64ReferenceAccessImpl.java` I am calling the method uncompressReference(UnsignedWord w). However, in the points to analysis it thinks the UnsignedWord's stamp is an ObjectStamp. I haven't looked into this issue very deeply, but I believe the stamp of the UnsignedWord is supposed to be internally converted to a PrimitiveStamp, and that this is not happening.

For reference, I was using the same uncompressReference(UnsignedWord w) method in my original implementation off a branch from around Dec. 5 (f2aa6489a660c925eb693cf445f9b8dd450b8d4b), and it was working fine. Something must have changed since then.

tshull226:
Another thing to point out is write `AArch64GCRefAccessor.java`'s writeRefUpdate method is assuming that it is being called from within a safepoint. I could change it to a) very this assumption and/or b) to call VMThreads.ActionOnTransitionToJavaSupport.requestAllThreadsSynchronizeCode(); from within a VMOperation.

tshull226:
It turns out the issue I referenced above about analysis failing for uncompressReference(UnsignedWord w) was my fault -- I was referencing it directly instead of via ReferenceAccess.singleton().

teshull:
Background for closing this PR - having inline objects in runtime-installed code requires having a writable code cache, which has many security consequences. Given this, it is not worth the maintenance effort to support inlined objected in runtime code for AArch64

PR #2126, opened 2020-02-04
relax check on version info for MacOS toolchain

johanvos:
Old compilers use "Apple LLVM version" where new compilers use "Apple clang version"

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/codegen/CCompilerInvoker.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/codegen/CCompilerInvoker.java
@@ -211,7 +211,7 @@ public abstract class CCompilerInvoker {
         @Override
         protected CompilerInfo createCompilerInfo(Scanner scanner) {
             try {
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/codegen/CCompilerInvoker.java
```diff
@@ -211,7 +211,7 @@ protected String getDefaultCompiler() {
         @Override
         protected CompilerInfo createCompilerInfo(Scanner scanner) {
             try {
-                while (scanner.findInLine("Apple LLVM version ") == null) {
+                while (scanner.findInLine("Apple.*version ") == null) {
```
loicottet: As there are only two possible values I'd test for both explicitly instead of using a wildcard.

chrisseaton:
Perhaps we should gain an understanding of when this version string changed? What macOS are Oracle using? Otherwise we risk cargo culting in these strings.

loicottet:
Fixed by [98ffbbe](https://github.com/oracle/graal/commit/98ffbbedb3b)

Issue #1911, opened 2019-12-02
release file insufficiently distinguishes java versions


ekmett:
I have a language for which I'd like to support both jdk1.8 and jdk11 builds, more or less automagically depending on where `GRAALVM_HOME` is set.

With the current release file it is rather awkward to tell if we're looking at a `GRAALVM_HOME` that is built with `jdk11` or `jdk1.8`. 

The `release` file names the `GRAALVM_VERSION`, but no `JAVA_VERSION`, forcing me to at best guess that it is java 11 based on the URL of the `component_catalog`, but that strikes me as a particularly brittle distinction to draw.

dougxc:
You can use the standard `java.specification.version` system property, just like [JavaVersionUtil](https://github.com/oracle/graal/blob/a9db7dd44c3a6a37f6805bc93e7c366a201d2e9a/compiler/src/org.graalvm.compiler.serviceprovider/src/org/graalvm/compiler/serviceprovider/JavaVersionUtil.java#L32) does.

ekmett:
I can, but that does mean running a small java program in the middle of another build to get at the information, rather than looking at a nice properties file. 

Currently I might have 2-3 JVMs to work with, one for whatever is being used to run `gradle` itself (often set by IntelliJ), which is what I'd see if I tried accessing `java.specification.version` in my build script, one for wherever the `JAVA_HOME` is actually set, one for whatever is set as `GRAALVM_HOME`, which is where I try to find `gu` when running registered builds of my language project, but otherwise isn't needed now that JDK11 works out of the box with enough options.

dougxc:
Maybe this is a possibility:
```
java -XshowSettings:properties 2>&1 | grep java.specification.version
    java.specification.version = 1.8
```

gastaldi:
Would it be possible to add the JDK version to the `native-image -version` output?

jaikiran:
> Would it be possible to add the JDK version to the native-image -version output?

https://github.com/oracle/graal/pull/2148 contains an implementation of this proposal.

ansalond:
In the [upcoming GraalVM 20.0 release](https://www.graalvm.org/docs/release-notes/version-roadmap/), the GraalVM release file will be a superset of the one of the base JDK. It will therefore contain a `JAVA_VERSION` field.

jaikiran:
I think this is now resolved based on what @ansalond noted as well as https://github.com/oracle/graal/pull/2148 being merged.

[...]

PR #2148, opened 2020-02-13
Print Java version in the output of "native-image --version" command

jaikiran:
The `native-image --version` command currently prints:
```
GraalVM Version 19.3.1 CE
```
Given that the Graal VM ships in more than one variant of Java version (i.e. Java 8 and Java 11) and given that we have noticed there can be considerable difference in behaviour of native image generation depending on which variant is used, it is going to be useful to print out, in that `native-image --version` output, the Java version on which that Graal VM is based on.

This has been asked in a couple of places:
https://github.com/oracle/graal/issues/1911#issuecomment-565874506
https://github.com/quarkusio/quarkus/issues/6162

The commit here enhances the `native-image --version` output to even include the Java version. With this change, the output should look something like:
```
GraalVM Version 19.3.1 CE (Java Version 11.0.6)
```
(where Java 11.0.6 was the version this Graal VM was built upon).

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
@@ -81,6 +81,7 @@ class DefaultOptionHandler extends NativeImage.OptionHandler<NativeImage> {
                 if (!NativeImage.graalvmConfig.isEmpty()) {
                     message += " " + NativeImage.graalvmConfig;
                 }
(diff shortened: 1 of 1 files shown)


jaikiran:
@olpaw @dougxc, a general question - how does one track which commit made into which release of Graal VM? Like the commit in this PR doesn't seem like it made it to the 20.0 release. Does it mean it will be part of 20.0.1? Is there a easy way to track this?

dougxc:
The top level `release` file in a GraalVM binary shows the commit ids of each component in the binary. From that, it should be possible to work out whether a specific commit is in a release.

jaikiran:
Thank you @dougxc

Issue #2092, opened 2020-01-23
Add range check elimination in counted loops to Graal CE


rwestrel:
Graal CE lacks range check elimination in counted loops that C2 supports with so called loop predication.
For instance C2 can optimize the range check in the following code snippet out of loop:
       for (int i = start; i < stop; i++) {
            array[2 * i + offset] = i;
        }
C2 loop predication is implemented in:
http://hg.openjdk.java.net/jdk/jdk/file/34138fe5f9f7/src/hotspot/share/opto/loopPredicate.cpp

I'm working on a port of that logic to Graal and will provide it as a PR once I have working.

wirthi:
I understand that the patch based on #2153 was accept, so closing this ticket.

PR #2067, opened 2020-01-13
C2 style range check smearing

rwestrel:
C2 optimizes range checks of the form i + constant |<| array.length under what is referred to range check smearing. This change implements similar optimizations in Graal.

For instance, in the range check sequence:

i + 2 |<| array.length
i |<| array.length
i+1 |<| array.length

the first 2 range checks guarantee the 3rd one cannot fail so the 3rd one is useless and is candidate for elimination. However, accesses that depend on the 3rd range check need to depend on the 2 first, if the 3rd check is eliminated. I implemented that by adding a second guard input to GuardedNodes.

With the range check sequence:
i + 2 |<| array.length
i+1 |<| array.length
i |<| array.length

the 2nd check would be redundant if the checks are reordered so the i+1 range check is last. The change I propose attempts to float the i+1 guard below the i guard.

Finally, if reordering guards is not an option, this change proceeds speculatively: in the sequence above, the i+1 change can be speculatively widened to a i |<| array.length check which then covers the 3rd change as well.

PR #2153, opened 2020-02-13
range check predication

rwestrel:
Implementation of https://github.com/oracle/graal/issues/2092

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/RangeCheckPredicatesTest.java
+++ b/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/RangeCheckPredicatesTest.java
@@ -1,0 +1,1144 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, Red Hat Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package org.graalvm.compiler.hotspot.test;
+
+import jdk.vm.ci.hotspot.HotSpotJVMCIRuntime;
+import jdk.vm.ci.hotspot.HotSpotVMConfigAccess;
+import jdk.vm.ci.hotspot.HotSpotVMConfigStore;
+import jdk.vm.ci.meta.DeoptimizationReason;
+import jdk.vm.ci.meta.ProfilingInfo;
+import jdk.vm.ci.meta.ResolvedJavaMethod;
+import jdk.vm.ci.meta.SpeculationLog;
+import org.graalvm.collections.EconomicMap;
+import org.graalvm.compiler.core.common.cfg.Loop;
+import org.graalvm.compiler.core.test.GraalCompilerTest;
+import org.graalvm.compiler.graph.Node;
+import org.graalvm.compiler.nodes.StructuredGraph;
+import org.graalvm.compiler.nodes.calc.IntegerBelowNode;
+import org.graalvm.compiler.nodes.cfg.Block;
+import org.graalvm.compiler.options.OptionKey;
+import org.graalvm.compiler.options.OptionValues;
+import org.graalvm.compiler.serviceprovider.JavaVersionUtil;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.HashSet;
+import java.util.List;
+
+import static jdk.vm.ci.meta.DeoptimizationReason.BoundsCheckException;
+import static jdk.vm.ci.meta.DeoptimizationReason.LoopLimitCheck;
+import static org.graalvm.compiler.core.common.GraalOptions.LoopPredicationMainPath;
+
+public class RangeCheckPredicatesTest extends GraalCompilerTest {
+    @SuppressWarnings("unused") private static int volatileField;
+    final HotSpotVMConfigStore configStore = HotSpotJVMCIRuntime.runtime().getConfigStore();
+    final HotSpotVMConfigAccess access = new HotSpotVMConfigAccess(configStore);
+    final boolean useJVMCICompiler = access.getFlag("UseJVMCICompiler", Boolean.class);
+
+    private final SpeculationLog speculationLog;
+
+    @Override
+    protected SpeculationLog getSpeculationLog() {
+        speculationLog.collectFailedSpeculations();
+        return speculationLog;
+    }
+
+    private static OptionValues getOptionsMainPath() {
+        return getInitialOptions();
+    }
+
+    private static OptionValues getOptionsAllPaths() {
+        EconomicMap<OptionKey<?>, Object> overrides = OptionValues.newOptionMap();
+        overrides.put(LoopPredicationMainPath, false);
+        return new OptionValues(getInitialOptions(), overrides);
+    }
+
+    public RangeCheckPredicatesTest() {
+        speculationLog = getCodeCache().createSpeculationLog();
+    }
+
+    private void runOutOfBound(String methodName, int size, boolean loopLimitCheck, Object... testParameters) {
+        final ResolvedJavaMethod method = getResolvedJavaMethod(methodName);
+        final int[] array = new int[size];
+
+        ProfilingInfo profile = method.getProfilingInfo();
+        int deoptimizationCountBoundsCheck = profile.getDeoptimizationCount(BoundsCheckException);
+        int deoptimizationCountLoopLimitCheck = profile.getDeoptimizationCount(LoopLimitCheck);
+        int extraBoundsCheck;
+        int extraLoopLimitCheck;
+        if (!loopLimitCheck) {
+            // Running with UseJVMCICompiler off causes hotspot to account for trap twice
+            extraBoundsCheck = (useJVMCICompiler || JavaVersionUtil.JAVA_SPEC <= 8) ? 1 : 2;
+            extraLoopLimitCheck = 0;
+        } else {
+            extraBoundsCheck = (useJVMCICompiler || JavaVersionUtil.JAVA_SPEC <= 8) ? 0 : 1;
+            extraLoopLimitCheck = 1;
+        }
+        Object[] args = new Object[testParameters.length + 1];
+        args[0] = array;
+        System.arraycopy(testParameters, 0, args, 1, testParameters.length);
+        Result result = executeActual(getOptionsMainPath(), method, null, args);
+        Assert.assertNotNull(result.exception);
+        Assert.assertTrue(result.exception instanceof ArrayIndexOutOfBoundsException);
+        profile = method.getProfilingInfo();
+        Assert.assertEquals(deoptimizationCountBoundsCheck + extraBoundsCheck, profile.getDeoptimizationCount(BoundsCheckException));
+        Assert.assertEquals(deoptimizationCountLoopLimitCheck + extraLoopLimitCheck, profile.getDeoptimizationCount(LoopLimitCheck));
+    }
+
+    private void runNoOutOfBound(String methodName, int size, OptionValues options, Object... testParameters) {
+        final ResolvedJavaMethod method = getResolvedJavaMethod(methodName);
+        final int[] array = new int[size];
+        final HashSet<DeoptimizationReason> deoptimizationReasons = new HashSet<>();
+        deoptimizationReasons.add(BoundsCheckException);
+        deoptimizationReasons.add(LoopLimitCheck);
+        Object[] args = new Object[testParameters.length + 1];
+        args[0] = array;
+        System.arraycopy(testParameters, 0, args, 1, testParameters.length);
+        Result result = executeActualCheckDeopt(options, method, deoptimizationReasons, null, args);
+        Assert.assertNull(result.exception);
+    }
+
+    private boolean noRangeCheckInLoop(String method) {
+        StructuredGraph graph = getFinalGraph(getResolvedJavaMethod(method), getOptionsMainPath());
+        final StructuredGraph.ScheduleResult schedule = graph.getLastSchedule();
+        final List<Loop<Block>> loops = schedule.getCFG().getLoops();
+        Assert.assertEquals(1, loops.size());
+        final Loop<Block> loop = loops.get(0);
+        return loop.getBlocks().size() == 2;
+    }
+
+    private void verifyNoRangeCheckInLoop(String method) {
+        Assert.assertTrue(noRangeCheckInLoop(method));
+    }
+
+    private void verifyRangeCheckInLoop(String method) {
+        Assert.assertFalse(noRangeCheckInLoop(method));
+    }
+
+    public static void rangeCheckPredicatesLoopUpScalePos1(int[] array, int start, int stop) {
+        for (int i = start; i < stop; i++) {
+            array[i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScalePos1() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScalePos1");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScalePos1", 1000, getOptionsMainPath(), 0, 1000);
+        runOutOfBound("rangeCheckPredicatesLoopUpScalePos1", 1000, false, 0, 1001);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScalePos2(int[] array, int start, int stop) {
+        for (int i = start; i < stop; i++) {
+            array[i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScalePos2() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScalePos2");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScalePos2", 1000, getOptionsMainPath(), 0, 1000);
+        runOutOfBound("rangeCheckPredicatesLoopUpScalePos2", 1000, false, -1, 1000);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScalePos3(int[] array, int start, int stop, int offset) {
+        for (int i = start; i < stop; i++) {
+            array[i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScalePos3() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScalePos3");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScalePos3", 1000, getOptionsMainPath(), 5, 1005, -5);
+        runOutOfBound("rangeCheckPredicatesLoopUpScalePos3", 1000, false, 5, 1006, -5);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScalePos4(int[] array, int start, int stop, int offset) {
+        for (int i = start; i < stop; i++) {
+            array[i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScalePos4() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScalePos4");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScalePos4", 1000, getOptionsMainPath(), 5, 1005, -5);
+        runOutOfBound("rangeCheckPredicatesLoopUpScalePos4", 1000, false, 4, 1005, -5);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScalePos5(int[] array, int start, int stop, int offset) {
+        for (int i = start; i < stop; i++) {
+            array[2 * i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScalePos5() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScalePos5");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScalePos5", 1000, getOptionsMainPath(), -5, 495, 10);
+        runOutOfBound("rangeCheckPredicatesLoopUpScalePos5", 1000, false, -6, 495, 10);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScalePos6(int[] array, int start, int stop, int offset) {
+        for (int i = start; i < stop; i++) {
+            array[2 * i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScalePos6() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScalePos6");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScalePos6", 1000, getOptionsMainPath(), -5, 495, 10);
+        runOutOfBound("rangeCheckPredicatesLoopUpScalePos6", 1000, false, -5, 496, 10);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScalePos7(int[] array, int start, int stop) {
+        for (int i = start; i < stop; i++) {
+            array[2 * i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScalePos7() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScalePos7");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScalePos7", 1000, getOptionsMainPath(), 0, 500);
+        runOutOfBound("rangeCheckPredicatesLoopUpScalePos7", 1000, false, 0, 501);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScalePos8(int[] array, int start, int stop) {
+        for (int i = start; i < stop; i++) {
+            array[2 * i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScalePos8() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScalePos8");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScalePos8", 1000, getOptionsMainPath(), 0, 500);
+        runOutOfBound("rangeCheckPredicatesLoopUpScalePos8", 1000, false, -1, 500);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScaleNeg1(int[] array, int start, int stop) {
+        for (int i = start; i < stop; i++) {
+            array[-i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScaleNeg1() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScaleNeg1");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScaleNeg1", 1000, getOptionsMainPath(), -999, 1);
+        runOutOfBound("rangeCheckPredicatesLoopUpScaleNeg1", 1000, false, -999, 2);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScaleNeg2(int[] array, int start, int stop) {
+        for (int i = start; i < stop; i++) {
+            array[-i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScaleNeg2() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScaleNeg2");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScaleNeg2", 1000, getOptionsMainPath(), -999, 1);
+        runOutOfBound("rangeCheckPredicatesLoopUpScaleNeg2", 1000, false, -1000, 1);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScaleNeg3(int[] array, int start, int stop, int offset) {
+        for (int i = start; i < stop; i++) {
+            array[-i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScaleNeg3() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScaleNeg3");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScaleNeg3", 1000, getOptionsMainPath(), -1004, -4, -5);
+        runOutOfBound("rangeCheckPredicatesLoopUpScaleNeg3", 1000, false, -1004, -3, -5);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScaleNeg4(int[] array, int start, int stop, int offset) {
+        for (int i = start; i < stop; i++) {
+            array[-i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScaleNeg4() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScaleNeg4");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScaleNeg4", 1000, getOptionsMainPath(), -1004, -4, -5);
+        runOutOfBound("rangeCheckPredicatesLoopUpScaleNeg4", 1000, false, -1005, -4, -5);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScaleNeg5(int[] array, int start, int stop, int offset) {
+        for (int i = start; i < stop; i++) {
+            array[-2 * i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScaleNeg5() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScaleNeg5");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScaleNeg5", 1000, getOptionsMainPath(), -495, 5, 9);
+        runOutOfBound("rangeCheckPredicatesLoopUpScaleNeg5", 1000, false, -495, 6, 9);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScaleNeg6(int[] array, int start, int stop, int offset) {
+        for (int i = start; i < stop; i++) {
+            array[-2 * i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScaleNeg6() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScaleNeg6");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScaleNeg6", 1000, getOptionsMainPath(), -495, 5, 9);
+        runOutOfBound("rangeCheckPredicatesLoopUpScaleNeg6", 1000, false, -496, 5, 9);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScaleNeg7(int[] array, int start, int stop) {
+        for (int i = start; i < stop; i++) {
+            array[-2 * i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScaleNeg7() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScaleNeg7");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScaleNeg7", 1000, getOptionsMainPath(), -499, 1);
+        runOutOfBound("rangeCheckPredicatesLoopUpScaleNeg7", 1000, false, -500, 1);
+    }
+
+    public static void rangeCheckPredicatesLoopUpScaleNeg8(int[] array, int start, int stop) {
+        for (int i = start; i < stop; i++) {
+            array[-2 * i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopUpScaleNeg8() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopUpScaleNeg8");
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScaleNeg8", 1000, getOptionsMainPath(), -499, 1);
+        runOutOfBound("rangeCheckPredicatesLoopUpScaleNeg8", 1000, false, -499, 2);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScalePos1(int[] array, int start, int stop) {
+        for (int i = stop; i > start; i--) {
+            array[i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScalePos1() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScalePos1");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScalePos1", 1000, getOptionsMainPath(), -1, 999);
+        runOutOfBound("rangeCheckPredicatesLoopDownScalePos1", 1000, false, -1, 1000);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScalePos2(int[] array, int start, int stop) {
+        for (int i = stop; i > start; i--) {
+            array[i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScalePos2() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScalePos2");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScalePos2", 1000, getOptionsMainPath(), -1, 999);
+        runOutOfBound("rangeCheckPredicatesLoopDownScalePos2", 1000, false, -2, 999);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScalePos3(int[] array, int start, int stop, int offset) {
+        for (int i = stop; i > start; i--) {
+            array[i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScalePos3() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScalePos3");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScalePos3", 1000, getOptionsMainPath(), 4, 1004, -5);
+        runOutOfBound("rangeCheckPredicatesLoopDownScalePos3", 1000, false, 4, 1005, -5);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScalePos4(int[] array, int start, int stop, int offset) {
+        for (int i = stop; i > start; i--) {
+            array[i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScalePos4() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScalePos4");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScalePos4", 1000, getOptionsMainPath(), 4, 1004, -5);
+        runOutOfBound("rangeCheckPredicatesLoopDownScalePos4", 1000, false, 3, 1004, -5);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScalePos5(int[] array, int start, int stop, int offset) {
+        for (int i = stop; i > start; i--) {
+            array[2 * i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScalePos5() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScalePos5");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScalePos5", 1000, getOptionsMainPath(), -6, 494, 10);
+        runOutOfBound("rangeCheckPredicatesLoopDownScalePos5", 1000, false, -7, 494, 10);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScalePos6(int[] array, int start, int stop, int offset) {
+        for (int i = stop; i > start; i--) {
+            array[2 * i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScalePos6() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScalePos6");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScalePos6", 1000, getOptionsMainPath(), -6, 494, 10);
+        runOutOfBound("rangeCheckPredicatesLoopDownScalePos6", 1000, false, -6, 495, 10);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScalePos7(int[] array, int start, int stop) {
+        for (int i = stop; i > start; i--) {
+            array[2 * i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScalePos7() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScalePos7");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScalePos7", 1000, getOptionsMainPath(), -1, 499);
+        runOutOfBound("rangeCheckPredicatesLoopDownScalePos7", 1000, false, -1, 500);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScalePos8(int[] array, int start, int stop) {
+        for (int i = stop; i > start; i--) {
+            array[2 * i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScalePos8() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScalePos8");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScalePos8", 1000, getOptionsMainPath(), -1, 499);
+        runOutOfBound("rangeCheckPredicatesLoopDownScalePos8", 1000, false, -2, 499);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScaleNeg1(int[] array, int start, int stop) {
+        for (int i = stop; i > start; i--) {
+            array[-i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScaleNeg1() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScaleNeg1");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScaleNeg1", 1000, getOptionsMainPath(), -1000, 0);
+        runOutOfBound("rangeCheckPredicatesLoopDownScaleNeg1", 1000, false, -1000, 1);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScaleNeg2(int[] array, int start, int stop) {
+        for (int i = stop; i > start; i--) {
+            array[-i] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScaleNeg2() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScaleNeg2");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScaleNeg2", 1000, getOptionsMainPath(), -1000, 0);
+        runOutOfBound("rangeCheckPredicatesLoopDownScaleNeg2", 1000, false, -1001, 0);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScaleNeg3(int[] array, int start, int stop, int offset) {
+        for (int i = stop; i > start; i--) {
+            array[-i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScaleNeg3() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScaleNeg3");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScaleNeg3", 1000, getOptionsMainPath(), -1005, -5, -5);
+        runOutOfBound("rangeCheckPredicatesLoopDownScaleNeg3", 1000, false, -1005, -4, -5);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScaleNeg4(int[] array, int start, int stop, int offset) {
+        for (int i = stop; i > start; i--) {
+            array[-i + offset] = i;
+        }
+    }
+
+    @Test
+    public void testLoopDownScaleNeg4() {
+        verifyNoRangeCheckInLoop("rangeCheckPredicatesLoopDownScaleNeg4");
+        runNoOutOfBound("rangeCheckPredicatesLoopDownScaleNeg4", 1000, getOptionsMainPath(), -1005, -5, -5);
+        runOutOfBound("rangeCheckPredicatesLoopDownScaleNeg4", 1000, false, -1006, -5, -5);
+    }
+
+    public static void rangeCheckPredicatesLoopDownScaleNeg5(int[] array, int start, int stop, int offset) {
+        for (int i = stop; i > start; i--) {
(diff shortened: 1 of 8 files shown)


thomaswue:
Thank you for the suggested contribution!

Larger changes to the compiler pipeline need to be evaluated based on the criteria outlined at https://mail.openjdk.java.net/pipermail/graal-dev/2019-July/005862.html. So an assessment of peak performance / compilation time / code size impact is required for at least renaissance suite plus SpecJVM2008. I think for this particular case compilation time & code size shouldn't be affected much (ideally maybe still verify that this assumption holds); but the peak performance metric would certainly be interesting. Please run those also with -esa and CompilationFailureAction=ExitVM to make sure that everything runs accurately. Would be good to do this also for the CTW runs to check for correctness. I did a quick test on the branch and there were several failures.

High level point regarding the code is that we typically prefer to have optimizations in separate phases that can be turned on/off. 

This could be security sensitive and will require some detailed manual review as well. I would prefer to do that after understanding the performance impact and the patch being able to run the larger benchmarks.

rwestrel:
> Larger changes to the compiler pipeline need to be evaluated based on the criteria outlined at https://mail.openjdk.java.net/pipermail/graal-dev/2019-July/005862.html. So an assessment of peak performance / compilation time / code size impact is required for at least renaissance suite plus SpecJVM2008. I think for this particular case compilation time & code size shouldn't be affected much (ideally maybe still verify that this assumption holds); but the peak performance metric would certainly be interesting. Please run those also with -esa and CompilationFailureAction=ExitVM to make sure that everything runs accurately. Would be good to do this also for the CTW runs to check for correctness. I did a quick test on the branch and there were several failures.

Hi Thomas. Thanks for looking at this. I will try to reproduce the failures and provide performance numbers.

dougxc:
@rwestrel what's the relationship between this PR and #2067?

rwestrel:
> @rwestrel what's the relationship between this PR and #2067?

In both cases I added and extra input to Guarded nodes so there's some common code. Otherwise they are independent.

rwestrel:
> Please run those also with -esa and CompilationFailureAction=ExitVM to make sure that everything runs accurately. Would be good to do this also for the CTW runs to check for correctness. I did a quick test on the branch and there were several failures.

Can you specify where you're seeing failures?
Is:
mx  ctw  -esa -DCompileTheWorld.Config="CompilationFailureAction=ExitVM"
what you're suggesting?
It runs fine AFAICT with jdk 8.
Is there a recommended way of running Renaissance? Number of iterations etc.?

> High level point regarding the code is that we typically prefer to have optimizations in separate phases that can be turned on/off.

The patch has this:
        if (ConditionalElimination.getValue(options)) {
            appendPhase(new IterativeConditionalEliminationPhase(canonicalizer, true));
        }

Is it not the correct way of doing this?

rwestrel:
I also ran all of renaissance and specjvm2008 with -esa -Dgraal.CompilationFailureAction=ExitVM successfully.

plokhotnyuk:
@rwestrel will your improvement help to eliminate range checks for each store operation in case of sequential access to arrays, like [here](https://github.com/plokhotnyuk/jsoniter-scala/blob/5f4fc9bfd256d7d2a527a0c0ca3b23226e8a5992/jsoniter-scala-core/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonWriter.scala#L815-L871)?

Currently GraalVM CE generates a range check branch for each store operation:
```
[info]   0.29%  │││  0x00007f287777e071: cmp    %ebx,%r9d
[info]          │││  0x00007f287777e074: jbe    0x00007f287777e889  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@36 (line 815)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.10%  │││  0x00007f287777e07a: mov    %ebx,%ecx
[info]   0.18%  │││  0x00007f287777e07c: inc    %ecx               ;*iadd {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@52 (line 817)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.41%  │││  0x00007f287777e07e: xchg   %ax,%ax
[info]   0.23%  │││  0x00007f287777e080: cmp    %ecx,%r9d
[info]          │││  0x00007f287777e083: jbe    0x00007f287777e65c  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@56 (line 817)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.02%  │││  0x00007f287777e089: lea    0x2(%rbx),%eax     ;*iadd {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@62 (line 818)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.31%  │││  0x00007f287777e08c: cmp    %eax,%r9d
[info]          │││  0x00007f287777e08f: jbe    0x00007f287777e898  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@69 (line 818)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.41%  │││  0x00007f287777e095: lea    0x3(%rbx),%ebp     ;*iadd {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@89 (line 820)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.16%  │││  0x00007f287777e098: nopl   0x0(%rax,%rax,1)
[info]   0.06%  │││  0x00007f287777e0a0: cmp    %ebp,%r9d
[info]          │││  0x00007f287777e0a3: jbe    0x00007f287777e87a  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@93 (line 820)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.27%  │││  0x00007f287777e0a9: lea    0x4(%rbx),%r13d    ;*iadd {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@99 (line 821)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.41%  │││  0x00007f287777e0ad: cmp    %r13d,%r9d
[info]          │││  0x00007f287777e0b0: jbe    0x00007f287777e86b  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@106 (line 821)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.23%  │││  0x00007f287777e0b6: lea    0x5(%rbx),%r14d    ;*iadd {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@126 (line 823)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.08%  │││  0x00007f287777e0ba: nopw   0x0(%rax,%rax,1)
[info]   0.22%  │││  0x00007f287777e0c0: cmp    %r14d,%r9d
[info]          │││  0x00007f287777e0c3: jbe    0x00007f287777e85c  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@130 (line 823)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.59%  │││  0x00007f287777e0c9: mov    %rdx,(%rsp)
[info]   0.31%  │││  0x00007f287777e0cd: lea    0x6(%rbx),%edx     ;*iadd {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@137 (line 824)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.08%  │││  0x00007f287777e0d0: cmp    %edx,%r9d
[info]          │││  0x00007f287777e0d3: jbe    0x00007f287777e851  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@144 (line 824)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.23%  │││  0x00007f287777e0d9: lea    0x7(%rbx),%edi     ;*iadd {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@162 (line 826)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.39%  │││  0x00007f287777e0dc: nopl   0x0(%rax)
[info]   0.39%  │││  0x00007f287777e0e0: cmp    %edi,%r9d
[info]          │││  0x00007f287777e0e3: jbe    0x00007f287777e846  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@166 (line 826)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
...
```
and then proceeds with stores to an array:
```
...
[info]   0.37%  │││  0x00007f287777e350: mov    %r10,0x88(%rsp)
[info]   0.20%  │││  0x00007f287777e358: mov    %r11d,%r10d
[info]   0.04%  │││  0x00007f287777e35b: shr    $0x18,%r10d        ;*iushr {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@43 (line 816)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.45%  │││  0x00007f287777e35f: movswl 0x10(%rbx,%r10,2),%r10d  ;*saload {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@44 (line 816)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.37%  │││  0x00007f287777e365: mov    %r10b,0x10(%rcx,%r8,1)  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@56 (line 817)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.31%  │││  0x00007f287777e36a: sar    $0x8,%r10d         ;*ishr {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@67 (line 818)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.06%  │││  0x00007f287777e36e: mov    %r10b,0x10(%rax,%r8,1)  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@69 (line 818)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.61%  │││  0x00007f287777e373: mov    %r11d,%r10d
[info]   0.16%  │││  0x00007f287777e376: shr    $0x10,%r10d        ;*iushr {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@76 (line 819)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.31%  │││  0x00007f287777e37a: and    $0xff,%r10d        ;*iand {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@80 (line 819)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.02%  │││  0x00007f287777e381: movswl 0x10(%rbx,%r10,2),%r10d  ;*saload {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@81 (line 819)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.59%  │││  0x00007f287777e387: mov    %r10b,0x10(%rbp,%r8,1)  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@93 (line 820)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.23%  │││  0x00007f287777e38c: sar    $0x8,%r10d         ;*ishr {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@104 (line 821)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.33%  │││  0x00007f287777e390: mov    %r10b,0x10(%r13,%r8,1)  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@106 (line 821)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.35%  │││  0x00007f287777e395: mov    %r11d,%r10d
[info]   0.10%  │││  0x00007f287777e398: shr    $0x8,%r10d         ;*iushr {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@113 (line 822)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.29%  │││  0x00007f287777e39c: and    $0xff,%r10d        ;*iand {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@117 (line 822)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.47%  │││  0x00007f287777e3a3: movswl 0x10(%rbx,%r10,2),%r10d  ;*saload {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@118 (line 822)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.22%  │││  0x00007f287777e3a9: mov    %r10b,0x10(%r14,%r8,1)  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@130 (line 823)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.22%  │││  0x00007f287777e3ae: sar    $0x8,%r10d         ;*ishr {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@142 (line 824)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
[info]   0.16%  │││  0x00007f287777e3b2: mov    %r10b,0x10(%rdx,%r8,1)  ;*bastore {reexecute=0 rethrow=0 return_oop=0}
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeUUID@144 (line 824)
[info]          │││                                                ; - com.github.plokhotnyuk.jsoniter_scala.core.JsonWriter::writeVal@13 (line 274)
```
Please use the following command to run the benchmark which reproduces that:
```
sbt -java-home /usr/lib/jvm/graalvm-ce-java11 'jsoniter-scala-benchmark/jmh:run -jvmArgsAppend "-Dgraal.UseBranchesWithin32ByteBoundary=true" -prof perfasm -p size=128 ArrayOfUUIDsWriting.jsoniterScalaPrealloc'
```

rwestrel:
> @rwestrel will your improvement help to eliminate range checks for each store operation in case of sequential access to arrays

Not unless range checks are in a loop. https://github.com/oracle/graal/pull/2067 is the one that I would expect to help. I'll give a try a report back.

rwestrel:
> > @rwestrel will your improvement help to eliminate range checks for each store operation in case of sequential access to arrays
> 
> Not unless range checks are in a loop. #2067 is the one that I would expect to help. I'll give a try a report back.

@plokhotnyuk  see my comment: https://github.com/oracle/graal/pull/2067#issuecomment-601281807

thomaswue:
@rwestrel Thanks for running renaissance & specjvm2008. What are the performance results for the branch vs master? Maybe you can add #2067 on top for another round of measurements.

rwestrel:
Benchmark results:

# specjvm2008

first score column is baseline (with #2067 enabled), second score column is with this enabled. Higher is better. Last column gives percentage change.

|Benchmark               | Mode  |Cnt  |    Score |    Error  |  Units  |    Score |    Error  |  Units | change (%)|
|------------------------|-------|-----|----------|-----------|---------|----------|-----------|--------|----------|
|Compiler.compiler       |thrpt  | 30  |  573.531 |±  65.496  |ops/min  |  594.114 |±  51.021  |ops/min | 3.5       |
|Compiler.sunflow        |thrpt  | 30  |  987.158 |±  90.243  |ops/min  |  969.761 |±  90.849  |ops/min |-1.7       |
|Compress.test           |thrpt  | 15  | 1201.700 |±   7.195  |ops/min  | 1196.536 |±   5.869  |ops/min |-0.4       |
|CryptoAes.test          |thrpt  | 15  |   70.792 |±   0.745  |ops/min  |   78.665 |±   0.573  |ops/min | 11.1      |
|CryptoRsa.test          |thrpt  | 15  | 4139.534 |±  15.008  |ops/min  | 4135.276 |±   9.983  |ops/min |-0.1       |
|CryptoSignVerify.test   |thrpt  | 15  |15509.796 |±  87.868  |ops/min  |17401.964 |±  95.872  |ops/min | 12.1      |
|Derby.test              |thrpt  | 30  | 1444.629 |± 114.643  |ops/min  | 1431.280 |±  93.747  |ops/min |-0.9       |
|MpegAudio.test          |thrpt  | 15  |  414.409 |±   2.552  |ops/min  |  414.034 |±   2.528  |ops/min |-0.0       |
|ScimarkFFT.large        |thrpt  | 15  |  160.540 |±   1.854  |ops/min  |  160.510 |±   6.148  |ops/min |-0.0       |
|ScimarkFFT.small        |thrpt  | 15  | 1147.278 |±   2.105  |ops/min  | 1144.735 |±   4.865  |ops/min |-0.2       |
|ScimarkLU.large         |thrpt  | 15  |   10.987 |±   0.204  |ops/min  |   12.311 |±   0.124  |ops/min | 12.0      |
|ScimarkLU.small         |thrpt  | 15  |  711.646 |±   0.809  |ops/min  |  713.199 |±   1.543  |ops/min | 0.2       |
|ScimarkMonteCarlo.test  |thrpt  | 15  | 5466.865 |± 137.679  |ops/min  | 5473.633 |±  95.200  |ops/min | 0.1       |
|ScimarkSOR.large        |thrpt  | 15  |  141.184 |±   0.238  |ops/min  |  141.044 |±   0.115  |ops/min |-0.0       |
|ScimarkSOR.small        |thrpt  | 15  |  676.367 |±   0.739  |ops/min  |  676.153 |±   0.870  |ops/min |-0.0       |
|ScimarkSparse.large     |thrpt  | 15  |   83.793 |±   0.157  |ops/min  |   83.731 |±   0.173  |ops/min |-0.0       |
|ScimarkSparse.small     |thrpt  | 15  |  728.918 |±  28.617  |ops/min  |  728.358 |±  27.768  |ops/min |-0.0       |
|Serial.test             |thrpt  | 30  |18771.242 |±  66.475  |  ops/s  |18913.206 |± 298.057  |  ops/s | 0.7       |
|Sunflow.test            |thrpt  | 15  |  260.490 |±   4.078  |ops/min  |  262.032 |±  12.919  |ops/min | 0.5       |
|XmlTransform.test       |thrpt  | 30  |  992.192 |±  30.799  |ops/min  | 1002.233 |±  30.845  |ops/min | 1.0       |
|XmlValidation.test      |thrpt  | 30  | 2221.914 |±  20.753  |ops/min  | 2207.065 |±  14.754  |ops/min |-0.6       |

# Renaissance

First result column is baseline, second is with this enabled, third is percentage change. To compute statistics I ignored first iteration as warmup doesn't always seem to be finished.

| Benchmark                         | Time             | Time             |  %  |
|-----------------------------------|------------------|------------------|-----|
|akka-uct (actors)                  |10341.50 ± 260.00 |10187.82 ± 273.65 |-1.4 |
|als (apache-spark)                 |2234.68 ± 104.12  |2129.95 ± 104.96  |-4.6 |
|chi-square (apache-spark)          |898.17 ± 22.59    |888.68 ± 16.87    |-1.0 |
|db-shootout (database)             |3437.66 ± 37.79   |3503.45 ± 88.31   | 1.9 |
|dec-tree (apache-spark)            |1855.36 ± 97.05   |1821.24 ± 67.61   |-1.8 |
|dotty (scala-dotty)                |1382.24 ± 202.65  |1372.51 ± 208.80  |-0.7 |
|finagle-chirper (twitter-finagle)  |2713.12 ± 268.48  |2824.35 ± 253.34  | 4.0 |
|finagle-http (twitter-finagle)     |2562.50 ± 204.63  |2555.81 ± 112.83  |-0.2 |
|fj-kmeans (jdk-concurrent)         |1944.43 ± 8.41    |2086.92 ± 7.19    | 7.3 |
|future-genetic (jdk-concurrent)    |2001.90 ± 14.23   |1988.63 ± 19.09   |-0.6 |
|gauss-mix (apache-spark)           |981.65 ± 26.10    |1030.93 ± 23.31   | 5.0 |
|log-regression (apache-spark)      |2650.36 ± 64.16   |2660.02 ± 66.27   | 0.3 |
|mnemonics (jdk-streams)            |4728.73 ± 67.87   |4553.60 ± 66.87   |-3.7 |
|movie-lens (apache-spark)          |6256.62 ± 217.42  |6169.73 ± 169.74  |-1.3 |
|naive-bayes (apache-spark)         |1062.36 ± 128.01  |963.54 ± 103.98   |-9.3 |
|neo4j-analytics (neo4j)            |4729.81 ± 87.74   |4861.57 ± 100.18  | 2.7 |
|page-rank (apache-spark)           |3659.88 ± 61.44   |3515.14 ± 81.71   |-3.9 |
|par-mnemonics (jdk-streams)        |4181.60 ± 15.01   |3982.03 ± 21.40   |-4.7 |
|philosophers (scala-stm)           |2566.91 ± 121.94  |2711.12 ± 92.72   | 5.6 |
|reactors (actors)                  |20974.97 ± 475.15 |20802.28 ± 873.91 |-0.8 |
|rx-scrabble (rx)                   |237.30 ± 16.30    |233.03 ± 13.22    |-1.7 |
|scala-doku (scala-sat)             |4227.46 ± 47.11   |4307.52 ± 85.19   | 1.8 |
|scala-kmeans (scala-stdlib)        |185.46 ± 4.37     |184.75 ± 3.57     |-0.3 |
|scala-stm-bench7 (scala-stm)       |1173.26 ± 44.37   |1152.17 ± 62.01   |-1.7 |
|scrabble (jdk-streams)             |229.70 ± 11.33    |232.59 ± 10.39    | 1.2 |

thomaswue:
Are the regressions on fj-kmeans, philosophers, and gauss-mix real or just measurement artefacts? Maybe you can run those with more warmups / repeat the measurements for those 3 to check this. Are there expected scenarios where this makes things slower (e.g., with lower frequency loops)?

rwestrel:
> Are the regressions on fj-kmeans, philosophers, and gauss-mix real or just measurement artefacts? Maybe you can run those with more warmups / repeat the measurements for those 3 to check this. Are there expected scenarios where this makes things slower (e.g., with lower frequency loops)?

The implementation I propose is fairly conservative so while a slow down is not impossible that would really be surprising. I'll rerun renaissance as you suggest.

rwestrel:
> > Are the regressions on fj-kmeans, philosophers, and gauss-mix real or just measurement artefacts? Maybe you can run those with more warmups / repeat the measurements for those 3 to check this. Are there expected scenarios where this makes things slower (e.g., with lower frequency loops)?
> 
> The implementation I propose is fairly conservative so while a slow down is not impossible that would really be surprising. I'll rerun renaissance as you suggest.

New runs with double the number of iterations, 3 runs for each.
fj-kmeans baseline runs:
2237.70 +/- 35.92
2186.86 +/- 33.72
2236.25 +/- 35.22

fj-kmeans with the patch and #2067:
2211.56 +/- 35.49
2183.96 +/- 33.47
2142.94 +/- 29.44

philosophers baseline runs:
2651.49 +/- 72.53
2913.75 +/- 56.76
2747.96 +/- 57.44

philosophers with the patch and #2067:
2650.14 +/- 69.95
2612.05 +/- 41.70
2721.60 +/- 58.88

gauss-mix baseline runs:
1034.32 +/- 21.61
1036.14 +/- 20.57
1045.59 +/- 21.76

gauss-mix with the patch and #2067:
1042.52 +/- 22.94
1042.84 +/- 23.10
1044.68 +/- 22.27

So no regressions AFAICT and most likely no improvement either.

rwestrel:
@thomaswue what's your take on the performance measurement? Are you still concerned by regressions? Are there any other measurements you'd say are required?

thomaswue:
Sorry for dropping the ball. This looks fine from a performance impact perspective. I've requested Gilles for a code/architecture review.

rwestrel:
> Sorry for dropping the ball. This looks fine from a performance impact perspective. I've requested Gilles for a code/architecture review.

Great. Thanks.

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
```diff
@@ -0,0 +1,399 @@
...
+                        }
+                    }
+                }
+            } catch (Throwable t) {
+                throw debug.handle(t);
+            }
+        }
+    }
+
+    private static EconomicSet<Node> collectMainPath(LoopsData data, LoopEx loop) {
```
gilles-duboscq, line 134: Could you clarify what this is about in a comment? Is it about avoiding paths that are conceptually loop exits but have not yet been marked as such in the computation of `loop.whole()`?

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/memory/FixedAccessNode.java
```diff
@@ -42,10 +45,12 @@
  * does not include a null check on the object.
  */
 @NodeInfo
-public abstract class FixedAccessNode extends DeoptimizingFixedWithNextNode implements AddressableMemoryAccess, GuardedMemoryAccess, OnHeapMemoryAccess, IterableNodeType {
+public abstract class FixedAccessNode extends DeoptimizingFixedWithNextNode implements AddressableMemoryAccess, GuardedMemoryAccess, OnHeapMemoryAccess, IterableNodeType, Canonicalizable {
     public static final NodeClass<FixedAccessNode> TYPE = NodeClass.create(FixedAccessNode.class);
 
     @OptionalInput(InputType.Guard) protected GuardingNode guard;
+    @OptionalInput(InputType.Guard) protected GuardingNode extraGuard;
```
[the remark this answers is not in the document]
rschatz: We already have that: https://github.com/oracle/graal/blob/master/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/extended/MultiGuardNode.java

rwestrel:
Hi Gilles,

Thanks for reviewing this.

> Does this have some relationship with #2067?

They interact with each other and share the same extra guard machinery. They are otherwise independent. Actually, I would have preferred #2067 to be reviewed first. It's simpler I think.

Anyway I will rework both patches to take advantage of MultiGuardNode.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/memory/FixedAccessNode.java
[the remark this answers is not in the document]
rwestrel: > We already have that: https://github.com/oracle/graal/blob/master/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/extended/MultiGuardNode.java

Great. I missed that it already existed. I'll see how I can use it.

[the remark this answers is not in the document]
rwestrel: I've been reworking the patches using MultiGuardNode and all the extra guard stuff is gone now.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/FloatingGuardedNode.java
```diff
@@ -57,4 +61,36 @@ public void setGuard(GuardingNode guard) {
...
+            final GuardingNode tmp = this.extraGuard;
+            extraGuard = guard;
+            guard = tmp;
+        }
+        if (guard == extraGuard && guard != null) {
+            setExtraGuard(null);
+        }
+        if (graph() != null) {
+            // See if this node already exists
+            final FloatingGuardedNode duplicate = graph().findDuplicate(this);
```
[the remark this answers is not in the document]
rwestrel: That change doesn't exist in an upcoming patch.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/IntegerBelowNode.java
```diff
@@ -77,6 +80,27 @@ public Node canonical(CanonicalizerTool tool, ValueNode forX, ValueNode forY) {
         return this;
     }
 
+    public ValueNode getRangeCheckIndex() {
```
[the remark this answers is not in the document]
rwestrel: I moved this and related methods to ConditionalEliminationPhase$Instance as static methods.

compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/InductionVariable.java
```diff
@@ -115,4 +115,15 @@ public ValueNode extremumNode() {
      * Deletes any nodes created within the scope of this object that have no usages.
      */
     public abstract void deleteUnusedNodes();
+
+    /**
+     * Range check predication support.
+     */
+    public abstract boolean isConstantScale();
```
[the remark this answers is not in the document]
rwestrel: I'm looking for range check expressions of the form scale * i + offset in loop:
for (int i = ; i < ; i += stride) {
LoopPredicationPhase.verifyScaleAndOffset() implements logic that's quite similar with a chain of instanceof. It felt cleaner to implement it as methods of InductionVariable.
A sketch of that API is:
BasicInductionVariable.offset() is 0
BasicInductionVariable.scale() is 1
DerivedScaledInductionVariable.offset() is base.offset()
DerivedScaledInductionVariable.scale() is scale * base.scale()
DerivedOffsetInductionVariable.offset() is base.offset() + offset
DerivedOffsetInductionVariable.scale() is base.scale()

The API is a bit awkward because we would really need to return a 3 state value. For instance InductionVariable.offset() needs to return an offset node or a special value indicating a zero offset or a special value indicating that we can't tell the offset because the expression is too complicated. To achieve the same effect I split the 2 API points in 2:

offsetNode() returns either a node or null if it can't tell what the offset is
offsetIsZero() returns true if the offset is 0

Same with scale() with should return either a constant or some special value that tells scale is non constant or unknown.

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
rwestrel: In C2 guards do not float. There are pinned in whatever cfg path they were found at parse time. The C2 optimization pass that this PR emulates looks for candidate guards by starting from the loop backedge and following the dominator chain to the loop head. So any guard that's pinned in a path that's not always executed is not a candidate for predication. This code emulates that behavior (or at least it tries to).

I'm not sure what's the rationale for the C2 implementation is. Maybe it was simpler to do it that way and it has work well enough that it wasn't revised for a long time. I could see how guards that are not always executed could sometimes fail, cause the whole loop predication optimization to be skipped on recompilation and result in a performance anomaly.

The reason I include this conservative heuristic is that it feels safer from a performance point of view.

gilles-duboscq: OK, thanks for the details

compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/InductionVariable.java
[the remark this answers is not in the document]
gilles-duboscq: i think the issue comes from the fact that when you have a single induction variable "offset" and "scale" don't make much sense.

I think i see what you are trying to do here but without this context i can't really understand it that's why I think one option is to not put these method in the `InductionVariable` or to pass a second IV as argument so it's clear that it's an offset/scale relative to some other given IV.

Looking at the code in the phase i think the best is to pass the expected `BasicInductionVariable` to those functions so they can directly do the verification and there should be no need for `verifyScaleAndOffset`.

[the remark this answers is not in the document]
rwestrel: Right that makes sense now. I updated the PR.

rwestrel:
Patch is updated based on Gilles's feedback. CI failures are unrelated to that patch.

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
gilles-duboscq: Should this also be limited to guards that are anchored below the counter check of the loop?

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
```diff
@@ -0,0 +1,282 @@
...
+        if (offset == null || scale == null || !loop.isOutsideLoop(offset)) {
+            return;
+        }
+
+        long scaleCon = scale;
+
+        final GuardNode upperGuard = getGuardNode(loop, guard, range, graph, scaleCon, offset, true);
+        final GuardNode lowerGuard = getGuardNode(loop, guard, range, graph, scaleCon, offset, false);
+
+        GuardingNode combinedGuard = null;
```
gilles-duboscq, line 239: Could we just do
```
guard.replaceAtUsagesAndDelete(MultiGuardNode.combine(lowerGuard, upperGuard));
```
and let `MultiGuardNode` combine if they want to during canonicalization.
In particular we need to create the `MultiGuardNode` regardless of whether the `guard` has usages.

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
```diff
@@ -0,0 +1,282 @@
...
+        if (iv.isConstantScale(counter)) {
+            scale = iv.constantScale(counter);
+        }
+
+        ValueNode offset = null;
+
+        if (iv.offsetIsZero(counter)) {
+            offset = graph.unique(ConstantNode.forInt(0));
+        } else {
+            offset = iv.offsetNode(counter);
```
gilles-duboscq, line 227: Why is the offset allowed to be a variable but not the scale?

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
```diff
@@ -0,0 +1,282 @@
...
+                    combinedGuard = MultiGuardNode.combine(lowerGuard, upperGuard);
+                }
+                ((GuardedNode) usage).setGuard(combinedGuard);
+            }
+        }
+        assert guard.hasNoUsages();
+        guard.safeDelete();
+    }
+
+    private static GuardNode getGuardNode(LoopEx loop, GuardNode guard, ValueNode range, StructuredGraph graph, long scaleCon, ValueNode offset, boolean upper) {
```
gilles-duboscq, line 256: This could also use `initNode` and `extremumNode` from `InductionVariable` to from the upper and lower guards

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
```diff
@@ -0,0 +1,282 @@
...
+                for (LoopEx loop : data.countedLoops()) {
+                    final FrameState state = loop.loopBegin().stateAfter();
+                    final BytecodePosition pos = new BytecodePosition(null, state.getMethod(), state.bci);
+                    SpeculationLog.SpeculationReason reason = LOOP_PREDICATION.createSpeculationReason(pos);
+                    if (speculationLog.maySpeculate(reason)) {
+                        final CountedLoopInfo counted = loop.counted();
+                        final InductionVariable counter = counted.getCounter();
+                        final Condition condition = ((CompareNode) counted.getLimitTest().condition()).condition().asCondition();
+                        // Only inner most loops with int induction variable
+                        if (loop.loop().getChildren().isEmpty() &&
```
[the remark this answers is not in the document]
rwestrel: Makes sense. I'll make that change.

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
rwestrel: OK. A MultiGuardNode with MultiGuardNode inputs is not canonicalized right now AFAICT. Does that need to be changed?

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
rwestrel: I suppose this would work with non constant scales. The patch requires a constant scale because it's a copy of C2's algorithm and C2 requires it. Can this be left as a future improvement?

rwestrel:
@gilles-duboscq Thanks for the review comments. I pushed an update of the patch that should address all comments. I also changed the logic of the LoopPredicationMainPath to something much simpler that filter out guards whose anchor doesn't dominate the back edge. CI failures are unrelated to my change AFAICT.

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+        final ValueNode upperNode = MathUtil.add(graph, MathUtil.mul(graph, extremumNode, ConstantNode.forLong(scaleCon, graph)),
+                        IntegerConvertNode.convert(offset, StampFactory.forInteger(64), graph, NodeView.DEFAULT));
+        final LogicNode upperCond = IntegerBelowNode.create(upperNode, rangeLong, NodeView.DEFAULT);
+
+        final ValueNode initNode = IntegerConvertNode.convert(counter.initNode(), StampFactory.forInteger(64), graph, NodeView.DEFAULT);
+        final ValueNode lowerNode = MathUtil.add(graph, MathUtil.mul(graph, initNode, ConstantNode.forLong(scaleCon, graph)),
+                        IntegerConvertNode.convert(offset, StampFactory.forInteger(64), graph, NodeView.DEFAULT));
+        final LogicNode lowerCond = IntegerBelowNode.create(lowerNode, rangeLong, NodeView.DEFAULT);
+
```
XiaohongGong, line 206: Hi @rwestrel it's a great patch that could eliminate the range check! However, I have a question about the `lowerCond`. Is it expected to compare against "0" while not the array range like C2? The `lowerGuard` should be GE 0, isn't it?

rwestrel: Thanks for looking at this. The comparison in PhaseIdealLoop::rc_predicate() is:

  if (overflow) {
    // Integer expressions may overflow, do long comparison
    range = new ConvI2LNode(range);
    register_new_node(range, ctrl);
    cmp = new CmpULNode(max_idx_expr, range);
  } else {
    cmp = new CmpUNode(max_idx_expr, range);
  }
  register_new_node(cmp, ctrl);
  BoolNode* bol = new BoolNode(cmp, BoolTest::lt);

so max_idx_expr |<| range

for a range check  in the loop that follows the pattern: ... |<| range

How does the graal patch differ?

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
XiaohongGong: I checked the codes, and it seems initNode/extremumNode is not directly equals to "min/max" for an induction when the directly is `DOWN`.  So maybe it needs to handle the difference between "UP" and "DOWN", doesn't it?

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
XiaohongGong: This is used as the `upperGuard`, right? Maybe I'm not right, so the `lowerGuard` is not used to guard "min_idx >= 0"

rwestrel: PhaseIdealLoop::rc_predicate() is called twice from PhaseIdealLoop::loop_predication_impl_helper(), once for the lower bound and once for the upper bound. So that code above is used for both lower and upper guards.

XiaohongGong: Yeah, you are right. It's also compared against the array length in C2. And maybe that because it's a kind of unsigned comparison for range check, which make it no need to compare against to zero. Thanks for your more explanation!

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
gilles-duboscq: That's right which one is min vs. max depends on direction

rwestrel: For range check:
scale * iv + offset |<| range
and lower <= iv <= upper
we need 2 guards:
scale * lower + offset |<| range
scale * upper + offset |<| range

We test both ends of the range of values taken by the iv with the same formula. There's no need to handle UP and DOWN differently as long as we indeed add a guard for each ends of the iteration range.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/extended/MultiGuardNode.java
```diff
@@ -69,6 +69,20 @@ public Node canonical(CanonicalizerTool tool) {
         } else if (guards.size() == 1) {
             // Only a single guard left => replace multi-guard with that single guard.
             return guards.get(0);
+        } else {
+            // Merge chain of MultiGuardNodes.ra
```
gilles-duboscq, line 73: `canonical` methods have the contract that they shouldn't mutate the node so this should rather return a new node.

Strictly speaking `guards.trim()` above is a mutation but i suppose it's accepted because it only changes some internal representation without being observable in meaningful ways outside this node.

compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/RangeCheckPredicatesTest.java
```diff
@@ -0,0 +1,1164 @@
...
+        flag[0] = true;
+        flag[1] = false;
+        runNoOutOfBound("rangeCheckPredicatesLoopUpScalePosSpeculation1", 2000, getOptionsAllPaths(), 0, 2000);
+        final ResolvedJavaMethod method = getResolvedJavaMethod("rangeCheckPredicatesLoopUpScalePosSpeculation1");
+        StructuredGraph graph = getFinalGraph(method, getOptionsAllPaths());
+        StructuredGraph.ScheduleResult schedule = graph.getLastSchedule();
+        List<Loop<Block>> loops = schedule.getCFG().getLoops();
+        Assert.assertEquals(1, loops.size());
+        Loop<Block> loop = loops.get(0);
+        for (Block block : loop.getBlocks()) {
```
gilles-duboscq, line 997: Could this just use `countRangeChecksInLoop`? Same at the end of this method

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopPredicationPhase.java
XiaohongGong: Yes, indeed! Thanks!

compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/InductionVariable.java
```diff
@@ -115,4 +115,38 @@ public ValueNode extremumNode() {
...
+     */
+    public boolean isConstantScale(InductionVariable ref) {
+        return this == ref;
+    }
+
+    /**
+     * this = C * ref + n, returns C.
+     */
+    public long constantScale(InductionVariable ref) {
+        assert isConstantScale(ref);
```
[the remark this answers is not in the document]
rwestrel: Ok

compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedOffsetInductionVariable.java
```diff
@@ -152,6 +152,37 @@ private ValueNode op(ValueNode b, ValueNode o) {
...
 
+    @Override
+    public boolean isConstantScale(InductionVariable ref) {
+        return super.isConstantScale(ref) || base.isConstantScale(ref);
+    }
+
+    @Override
+    public long constantScale(InductionVariable ref) {
+        assert isConstantScale(ref);
+        if (super.isConstantScale(ref)) {
```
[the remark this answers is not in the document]
rwestrel: OK

compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedOffsetInductionVariable.java
```diff
@@ -152,6 +152,37 @@ private ValueNode op(ValueNode b, ValueNode o) {
...
+            return super.constantScale(ref);
+        }
+        return base.constantScale(ref) * (value instanceof SubNode && base.valueNode() == value.getY() ? -1 : 1);
+    }
+
+    @Override
+    public boolean offsetIsZero(InductionVariable ref) {
+        if (super.offsetIsZero(ref)) {
```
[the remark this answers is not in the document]
rwestrel: Ok

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/extended/MultiGuardNode.java
rwestrel: I reworked it so it returns a new node

compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/RangeCheckPredicatesTest.java
rwestrel: right!

rwestrel:
@gilles-duboscq all your comments should have been taken care of in the new patch. The CI failure is unrelated to the change AFAICT.

compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedScaledInductionVariable.java
```diff
@@ -134,6 +134,34 @@ public void deleteUnusedNodes() {
         GraphUtil.tryKillUnused(scale);
     }
 
+    @Override
+    public boolean isConstantScale(InductionVariable ref) {
+        return super.isConstantScale(ref) || (scale.isConstant() && base.isConstantScale(ref));
```
gilles-duboscq, line 139: Sorry i was not very clear last time, i think we should do the same as in `DerivedOffsetInductionVariable` in this class: use the simple expressions rather than the "super" methods.

To avoid back-and-forth delays i'm anyway taking the PR in this state for further testing. In the meanwhile could you consider doing those extra changes?

rwestrel: Sorry. I should have reviewed other similar uses. I just pushed an updated patch. Thanks for the careful review and for your help with this.

gilles-duboscq:
Hey Roland, i'm back from vacation, I have have pushed this through internal testing and it looks good so far.
I'll look into integrating this PR into master.

rwestrel:
> Hey Roland, i'm back from vacation, I have have pushed this through internal testing and it looks good so far.
> I'll look into integrating this PR into master.

I'm back too. That sounds good. Thanks for the update.

rwestrel:
@gilles-duboscq 
Hi Gilles,
Do you an update on this?

gilles-duboscq:
The last internal approvals came in. This is in the merge queue! Sorry for the long delay.
The good news is that it should be in 20.3.0 :)

rwestrel:
> The last internal approvals came in. This is in the merge queue! Sorry for the long delay.
> The good news is that it should be in 20.3.0 :)

Great. Do you have an ETA for when this should be in?
Thanks.

gilles-duboscq:
This went in 2 weeks ago but i see now that it was an older version (c60e710). I think it's functionally equivalent:
```diff
--- <unnamed>
+++ <unnamed>
@@ -1,4 +1,4 @@
-From d3178a2d8de05a75507b5c52692708af3f1534aa Mon Sep 17 00:00:00 2001
+From c60e7109d2f5cdf81592c730e4c8bc529b7e94fe Mon Sep 17 00:00:00 2001
 From: rwestrel <rwestrel@redhat.com>
 Date: Tue, 16 Jun 2020 15:02:04 +0200
 Subject: [PATCH] range check predication
@@ -1445,7 +1445,7 @@
 +    }
 +}
 diff --git a/compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedOffsetInductionVariable.java b/compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedOffsetInductionVariable.java
-index b1f75de74c12..d6b110a3d353 100644
+index b1f75de74c12..9c3e7479faf6 100644
 --- a/compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedOffsetInductionVariable.java
 +++ b/compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedOffsetInductionVariable.java
 @@ -152,6 +152,37 @@ private ValueNode op(ValueNode b, ValueNode o) {
@@ -1454,7 +1454,7 @@
  
 +    @Override
 +    public boolean isConstantScale(InductionVariable ref) {
-+        return this == ref || base.isConstantScale(ref);
++        return super.isConstantScale(ref) || base.isConstantScale(ref);
 +    }
 +
 +    @Override
@@ -1487,7 +1487,7 @@
      public String toString() {
          return String.format("DerivedOffsetInductionVariable base (%s) %s %s", base, value.getNodeClass().shortName(), offset);
 diff --git a/compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedScaledInductionVariable.java b/compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedScaledInductionVariable.java
-index a6881371b43a..6b468cf590dd 100644
+index a6881371b43a..bc726179eecf 100644
 --- a/compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedScaledInductionVariable.java
 +++ b/compiler/src/org.graalvm.compiler.loop/src/org/graalvm/compiler/loop/DerivedScaledInductionVariable.java
 @@ -134,6 +134,34 @@ public void deleteUnusedNodes() {
@@ -1496,21 +1496,21 @@
  
 +    @Override
 +    public boolean isConstantScale(InductionVariable ref) {
-+        return this == ref || (scale.isConstant() && base.isConstantScale(ref));
++        return super.isConstantScale(ref) || (scale.isConstant() && base.isConstantScale(ref));
 +    }
 +
 +    @Override
 +    public long constantScale(InductionVariable ref) {
 +        assert isConstantScale(ref);
-+        if (this == ref) {
-+            return 1;
++        if (super.isConstantScale(ref)) {
++            return super.constantScale(ref);
 +        }
 +        return scale.asJavaConstant().asLong() * base.constantScale(ref);
 +    }
 +
 +    @Override
 +    public boolean offsetIsZero(InductionVariable ref) {
-+        if (this == ref) {
++        if (super.offsetIsZero(ref)) {
 +            return true;
 +        }
 +        return base.offsetIsZero(ref);
```
Sorry about this mix-up

rwestrel:
> This went in 2 weeks ago but i see now that it was an older version ([c60e710](https://github.com/oracle/graal/commit/c60e7109d2f5cdf81592c730e4c8bc529b7e94fe)). I think it's functionally equivalent:

Excellent! This PR is not closed which I assumed meant the change was not in. Thanks again for all your help along the way.
What do you intend to do with the small change that was missed?

gilles-duboscq:
I will push it separately

PR #2167, opened 2020-02-20
Don't use the build compiler check in case LLVM backend is used.

### LINKED ISSUE #2166: cross-compilation to Android fails due to strict compiler check
 - GraalVM version or commit id if built from source: **20.1-dev**
 - CE or EE: **CE**
 - Build Time or run time failure: **buld-time**
 - JDK version: **JDK 11**
 - Native compiler information:<details><summary>Run the following to capture compiler version</summary>
   - In windows: `cl.exe`
   - In macOS : `cc -v`
   - In Linux: `gcc --version`
</details>

cross-compilation from Linux AMD64 for Android AArch64 fails, since NativeImageGenerator will invoke LinuxCCompilerInvoker.verifyCompiler to see if the build compiler is ok for the target.
When using LLVM though, we don't rely on this build compiler.

I propose to skip this step if the LLVM backend is used.

johanvos:
This fixes #2166

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java
@@ -848,7 +848,9 @@ public class NativeImageGenerator {
                 prepareLibC();
 
                 CCompilerInvoker compilerInvoker = CCompilerInvoker.create(tempDirectory());
-                compilerInvoker.verifyCompiler();
+                if (!"llvm".equals(SubstrateOptions.CompilerBackend.getValue())) {
+                    compilerInvoker.verifyCompiler();
+                }
                 ImageSingletons.add(CCompilerInvoker.class, compilerInvoker);
 
                 nativeLibraries = setupNativeLibraries(imageName, aConstantReflection, aMetaAccess, aSnippetReflection, cEnumProcessor, classInitializationSupport, debug);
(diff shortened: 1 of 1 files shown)


loicottet:
The LLVM backend still relies on the native compiler in the general case though, I'd make the check dependent on ExitAfterRelocatableImageWrite instead.

loicottet:
@johanvos can you check that omitting the check only when `ExitAfterRelocatableImageWrite` is on solves your problem?

johanvos:
yes, that works. I added a commit so that the verify is only skipped in case the LLVMBackend is user AND -H:+ExitAfterRelocatableImageWrite is set.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java
```diff
@@ -848,7 +848,9 @@ private void setupNativeImage(String imageName, OptionValues options, Map<Method
                 prepareLibC();
 
                 CCompilerInvoker compilerInvoker = CCompilerInvoker.create(tempDirectory());
```
olpaw: How about:
```
boolean nativeToolchainNeed = !"llvm".equals(SubstrateOptions.CompilerBackend.getValue());
if (nativeToolchainNeed) {
    CCompilerInvoker compilerInvoker = CCompilerInvoker.create(tempDirectory());
    compilerInvoker.verifyCompiler();
    ImageSingletons.add(CCompilerInvoker.class, compilerInvoker);
}
```
Not sure it makes sense to bring `ExitAfterRelocatableImageWrite` in. Even with ExitAfterRelocatableImageWrite you still need a native toolchain for query code compilation, or am I missing something?

Instead of `"llvm".equals(SubstrateOptions.CompilerBackend.getValue())` it'd use `ImageSingletons.contains(LLVMFeature.class)`

johanvos: we use pre-generated CAP files and CAPcache for cross-platform targets, so there is no compiler needed.

loicottet:
Fixed by [824fc477](https://github.com/oracle/graal/commit/824fc477439251e7dd599f0416988502037a3662)

PR #2172, opened 2020-02-21
sulong: Don't create bitcast_x86_fp80 test on AArch64

a74nh:
Test bitcast_x86_fp80 is X86 only.
Add architecture check to LLIRTestGen.java to blacklist it.

### DIFF
--- a/sulong/tests/com.oracle.truffle.llvm.tests.llirtestgen/src/com/oracle/truffle/llvm/tests/llirtestgen/LLIRTestGen.java
+++ b/sulong/tests/com.oracle.truffle.llvm.tests.llirtestgen/src/com/oracle/truffle/llvm/tests/llirtestgen/LLIRTestGen.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2019, Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2020, Oracle and/or its affiliates.
  *
  * All rights reserved.
  *
@@ -68,6 +68,28 @@ public class LLIRTestGen {
         boolean isFloat();
 
         String toName();
+    }
+
+    enum Architecture {
+        AMD64,
+        AArch64;
+
+        private static Architecture findCurrent() {
+            final String name = System.getProperty("os.arch");
+            if (name.equals("amd64") || name.equals("x86_64")) {
+                return AMD64;
+            }
+            if (name.equals("aarch64")) {
+                return AArch64;
+            }
+            throw new IllegalArgumentException("unknown architecture: " + name);
+        }
+
+        private static final Architecture current = findCurrent();
+
+        public static Architecture getCurrent() {
+            return current;
+        }
     }
 
     private enum ScalarType implements Type {
@@ -758,6 +780,11 @@ public class LLIRTestGen {
                         "fsub_x86_fp80", // Fails with managed sulong
                         "fdiv_x86_fp80"  // Fails with managed sulong
         ));
+        Architecture arch = Architecture.getCurrent();
+        if (arch == Architecture.AArch64) {
+            filenameBlacklist.add("bitcast_x86_fp80");
+        }
+
         filenameBlacklist = filenameBlacklist.stream().map(s -> makeBlacklistFilename(outputDir, s)).collect(Collectors.toSet());
 
         for (Type type : allTypes) {


a74nh:
> Hello Alan Hayward, thanks for contributing a PR to our project!
> 
> We use the [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) to make the copyright of contributions clear. We don't have a record of you having signed this yet, based on your email address alan -(dot)- hayward -(at)- arm -(dot)- com. You can sign it at that link.
> 

Hi, I'm an employee of Arm and I'll be contributing under their signature. Please let me know if there is anything else I need to do there.

a74nh:
My patch is just changing a sulong test file, so I don't think that my changes could be causing this failure:
 
```
There was 1 failure:
1) org.graalvm.compiler.hotspot.test.LambdaStableNameTest#checkStableLamdaNameForRunnableAndAutoCloseable
java.lang.NumberFormatException: For input string: "6f55x00"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:652)
	at java.base/java.math.BigInteger.<init>(BigInteger.java:543)
	at org.graalvm.compiler.hotspot.test.LambdaStableNameTest.assertLambdaName(LambdaStableNameTest.java:83)
	at org.graalvm.compiler.hotspot.test.LambdaStableNameTest.checkStableLamdaNameForRunnableAndAutoCloseable(LambdaStableNameTest.java:61)
```

zapster:
Hi Alan! Thanks for your contribution.
> My patch is just changing a sulong test file, so I don't think that my changes could be causing this failure

You are right. These are compiler failures and unrelated to your change. cc @dougxc 

Your change looks good to me. I'll take care of getting it in.
Out of curiosity, what is the state of the sulong tests on AArch64? Can we build all/most of them already? If so, I guess we should add it to the CI to ensure that we do not regress.

a74nh:
> Your change looks good to me. I'll take care of getting it in.

Thanks.

> Out of curiosity, what is the state of the sulong tests on AArch64? Can we build all/most of them already? If so, I guess we should add it to the CI to ensure that we do not regress.

There are about 40 tests which crash the JVM causing the entire testsuite to exit. There are about another 40 failures or so on top of that. Looks like it's the same issues over and over (vargs, passing structs by value, bitwise shifts etc). Hopefully I should be raising quite a few tests over the next few weeks fixing them,

How easy is it to blacklist tests in the CI? If so, I can get a list of the crashing tests to you.

zapster:
> There are about 40 tests which crash the JVM causing the entire testsuite to exit. There are about another 40 failures or so on top of that.

Cool, that sounds quite promising already!

> How easy is it to blacklist tests in the CI? If so, I can get a list of the crashing tests to you.

Currently, there is no white/black list for architectures, but it should be relatively simple to add. How are you executing the tests (e.g. `mx gate ...` or `mx unittest ...`)?

a74nh:
> 
> Currently, there is no white/black list for architectures, but it should be relatively simple to add. How are you executing the tests (e.g. `mx gate ...` or `mx unittest ...`)?

Ususally:
$ <update a test file>
$ mx build --project SULONG_TEST_SUITES
$ mx unittest --very-verbose SulongSuite

Although I have sometime used:
$ mx gate --tag sulongBasic

Still trying to figure out the best way to run everything.

zapster:
> 
> Ususally:
> $
> $ mx build --project SULONG_TEST_SUITES
> $ mx unittest --very-verbose SulongSuite

:+1: 
Just a hind, `mx gate --tag sulong` does the same in just a single command. (But doing it manually is totally fine.)

> 
> Although I have sometime used:
> $ mx gate --tag sulongBasic

:+1: 

> Still trying to figure out the best way to run everything.

_everything_ is kind of tricky since not all tests do work on all platform. Even on linux/amd64 we always run certain gate tags selectively. (Have a look at sulong/ci.hocon if you are interested in what we are running on what platform.)

I suggest we should try to make things work tag by tag, starting with `--tag sulong`. Once all tests work on AArch64, I can add them to the CI. Please let me know when that is the case.

(Tip: In case didn't know it already, `mx -v gate --dry-run` gives quite some infos about the different gates and their tags.)

PR #2173, opened 2020-02-21
sulong: test libc/getchar/getchar should use int

a74nh:
On AArch64, chars are unsigned by default.
The test libc/getchar/getchar requires "c" to be signed, otherwise
an infinite loop will occur. Switch to int (signed).


a74nh:
Travis failure is "checkStableLamdaNameForRunnableAndAutoCloseable" which has nothing to do with this patch.

Issue #2045, opened 2019-12-30
[AArch64] Renaissance benchmark hangs when Graal is used


pli-arm:
Recently, we noticed that on some types of AArch64 CPUs, the Renaissance benchmark hangs when Graal is used as the JVMCI compiler. This issue can be reproduced by below test command on Graal master with jdk11u.
```
cd graal/compiler
mx vm -XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCI -XX:+UseJVMCICompiler \
  -Djvmci.Compiler=graal -jar renaissance-gpl-0.10.0.jar future-genetic
```
where renaissance-gpl-0.10.0.jar can be downloaded by
```
wget https://github.com/renaissance-benchmarks/renaissance/releases/download/v0.10.0/renaissance-gpl-0.10.0.jar
```

We have some investigations on this issue. Below are some details. TL;DR

The future-genetic case in the Renaissance benchmark is a test which makes heavy use of the `java.util.concurrent` library. It submits a map-reduce job but hangs with a phenomenon that all the worker threads are parked. The hanging issue only appears on AArch64 when we use Graal as the JVMCI compiler. If we just use C1 or C2, everything is fine. This issue is very hard to be reproduced in a small Java program. As we noticed that the library code called by this case has a lot of volatile accesses, we guessed that Graal might have something wrong when compiling loads/stores with volatile.

To prove our guess, we dumpped the assembly code compiled for volatile load by each compiler. The C1 compiler uses a normal load instruction surrounded by two dmb instructions for a volatile load. Code is like
```
dmb ish
ldr [xxx]
dmb ishld
```
while C2 has optimized the volatile load to a below AArch64 load-acquire instruction.
```
ldar [xxx]
```
In the Graal compiler, we only generates one dmb after the load, like below. There's no dmb before.
```
ldr [xxx]
dmb ishld
```

To test that the missing DMB is the cause of hanging, we tried to modify the Graal's compiled to what C1's and/or C2's like. We found that, either by hard-coding one more DMB before the Graal compiled LDR or by replacing the LDR+DMB to a single LDAR, the hanging issue disappers. To see why C1 needs one more DMB before the load, we found JDK-8179954 (https://bugs.openjdk.java.net/browse/JDK-8179954) stating that the mixed use of STLR and LDR+DMB may cause some errors. We also found this problem mentioned in the comments of @adinn's PR #1772.

[...]

PR #2181, opened 2020-02-24
AArch64: Conditionally use barriers for atomic operations.

XiaohongGong:
Graal always uses LDAXR+STLXR for Unsafe atomic operations. However this is not sufficient on AArch64 for the VM internal atomics that need the full barrier semantics (See [1]).

Another issue in Graal is for the java volatile read. Currently graal uses the LDR+DMB for volatile read, which might have issues when it is used together with STLR/STLXR (See [2]). And this has
already made several Renaissance benchmarks hang on some machines (See [3]).

This patch resolved these issues with the modifications:
  1. Add a "conservative" field for "LocationIdentity" to decide whether to use the full semantic barriers for VM internal atomic operations.
  2. Always use the optimized "ldaxr/stlxr" pattern for java atomics.
  3. Temporarily insert a full barrier after STLXR for the normal atomics to make the Renaissance benchmark work well. This can be removed once "LDAR" is used for java volatile read (See [4]).

[1] https://patchwork.kernel.org/patch/3575821/
[2] https://bugs.openjdk.java.net/browse/JDK-8179954
[3] https://github.com/oracle/graal/issues/2045
[4] https://github.com/oracle/graal/pull/1772

Change-Id: Ifa57c9733b1dd1a2e2208e885a11932475a4e3b0

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64LIRGenerator.java
+++ b/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64LIRGenerator.java
@@ -152,21 +152,24 @@ public abstract class AArch64LIRGenerator extends LIRGenerator {
     }
 
     @Override
-    public Variable emitLogicCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, Value trueValue, Value falseValue) {
-        Variable prevValue = newVariable(expectedValue.getValueKind());
-        Variable scratch = newVariable(LIRKind.value(AArch64Kind.DWORD));
-        append(new CompareAndSwapOp(prevValue, loadReg(expectedValue), loadReg(newValue), asAllocatable(address), scratch));
+    public Variable emitLogicCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, Value trueValue, Value falseValue, boolean useBarriers) {
+        emitCompareAndSwap(address, expectedValue, newValue, useBarriers);
         assert trueValue.getValueKind().equals(falseValue.getValueKind());
+        assert isIntConstant(trueValue, 1) && isIntConstant(falseValue, 0);
         Variable result = newVariable(trueValue.getValueKind());
-        append(new CondMoveOp(result, ConditionFlag.EQ, asAllocatable(trueValue), asAllocatable(falseValue)));
+        append(new CondSetOp(result, ConditionFlag.EQ));
         return result;
     }
 
     @Override
-    public Variable emitValueCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue) {
-        Variable result = newVariable(newValue.getValueKind());
-        Variable scratch = newVariable(LIRKind.value(AArch64Kind.WORD));
-        append(new CompareAndSwapOp(result, loadReg(expectedValue), loadReg(newValue), asAllocatable(address), scratch));
+    public Variable emitValueCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, boolean useBarriers) {
+        return emitCompareAndSwap(address, expectedValue, newValue, useBarriers);
+    }
+
+    private Variable emitCompareAndSwap(Value address, Value expectedValue, Value newValue, boolean useBarriers) {
+        Variable result = newVariable(expectedValue.getValueKind());
+        Variable scratch = newVariable(LIRKind.value(AArch64Kind.DWORD));
+        append(new CompareAndSwapOp(result, loadReg(expectedValue), loadReg(newValue), asAllocatable(address), scratch, useBarriers));
         return result;
     }
 
--- a/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
+++ b/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -62,13 +62,17 @@ public class AArch64AtomicMove {
         @Alive protected AllocatableValue addressValue;
         @Temp protected AllocatableValue scratchValue;
 
-        public CompareAndSwapOp(AllocatableValue result, Value expectedValue, AllocatableValue newValue, AllocatableValue addressValue, AllocatableValue scratch) {
+        private final boolean useBarriers;
+
+        public CompareAndSwapOp(AllocatableValue result, Value expectedValue, AllocatableValue newValue,
+                        AllocatableValue addressValue, AllocatableValue scratch, boolean useBarriers) {
             super(TYPE);
             this.resultValue = result;
             this.expectedValue = expectedValue;
             this.newValue = newValue;
             this.addressValue = addressValue;
             this.scratchValue = scratch;
+            this.useBarriers = useBarriers;
         }
 
         @Override
@@ -80,10 +84,20 @@ public class AArch64AtomicMove {
             Register address = asRegister(addressValue);
             Register result = asRegister(resultValue);
             Register newVal = asRegister(newValue);
+            if (useBarriers) {
+                masm.dmb(AArch64Assembler.BarrierKind.ANY_ANY);
+                emitCompareAndSwap(masm, size, address, result, newVal, false, false);
+                masm.dmb(AArch64Assembler.BarrierKind.ANY_ANY);
+            } else {
+                emitCompareAndSwap(masm, size, address, result, newVal, true, true);
+            }
+        }
+
+        private void emitCompareAndSwap(AArch64MacroAssembler masm, int size, Register address, Register result, Register newVal, boolean acquire, boolean release) {
             if (AArch64LIRFlagsVersioned.useLSE(masm.target.arch)) {
                 Register expected = asRegister(expectedValue);
                 masm.mov(size, result, expected);
-                masm.cas(size, result, newVal, address, true /* acquire */, true /* release */);
+                masm.cas(size, result, newVal, address, acquire, release);
                 AArch64Compare.gpCompare(masm, resultValue, expectedValue);
             } else {
                 // We could avoid using a scratch register here, by reusing resultValue for the
@@ -93,10 +107,10 @@ public class AArch64AtomicMove {
                 Label retry = new Label();
                 Label fail = new Label();
                 masm.bind(retry);
-                masm.ldaxr(size, result, address);
+                masm.loadExclusive(size, result, address, acquire);
                 AArch64Compare.gpCompare(masm, resultValue, expectedValue);
                 masm.branchConditionally(AArch64Assembler.ConditionFlag.NE, fail);
-                masm.stlxr(size, scratch, newVal, address);
+                masm.storeExclusive(size, scratch, newVal, address, release);
                 // if scratch == 0 then write successful, else retry.
                 masm.cbnz(32, scratch, retry);
                 masm.bind(fail);
--- a/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGeneratorTool.java
+++ b/compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGeneratorTool.java
@@ -151,9 +151,9 @@ public interface LIRGeneratorTool extends DiagnosticLIRGeneratorTool, ValueKindF
 
     void emitNullCheck(Value address, LIRFrameState state);
 
-    Variable emitLogicCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, Value trueValue, Value falseValue);
+    Variable emitLogicCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, Value trueValue, Value falseValues, boolean useBarriers);
 
-    Value emitValueCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue);
+    Value emitValueCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, boolean useBarriers);
 
     /**
      * Emit an atomic read-and-add instruction.
--- a/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
+++ b/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
@@ -896,6 +896,42 @@ public class AArch64MacroAssembler extends AArch64Assembler {
         }
     }
 
+    /* exclusive access */
+    /**
+     * Load exclusive. Natural alignment of address is required.
+     *
+     * @param size size of memory read in bits. Must be 8, 16, 32 or 64.
+     * @param rt general purpose register. May not be null or stackpointer.
+     * @param rn general purpose register.
+     * @param acquire memory model flag. Decide whether the load has acquire semantics.
+     */
+    public void loadExclusive(int size, Register rt, Register rn, boolean acquire) {
+        if (acquire) {
+            ldaxr(size, rt, rn);
+        } else {
+            ldxr(size, rt, rn);
+        }
+    }
+
+    /**
+     * Store exclusive. Natural alignment of address is required. rs and rt may not point to the
+     * same register.
+     *
+     * @param size size of bits written to memory. Must be 8, 16, 32 or 64.
+     * @param rs general purpose register. Set to exclusive access status. 0 means success,
+     *            everything else failure. May not be null, or stackpointer.
+     * @param rt general purpose register. May not be null or stackpointer.
+     * @param rn general purpose register.
+     * @param release memory model flag. Decide whether the store has release semantics.
+     */
+    public void storeExclusive(int size, Register rs, Register rt, Register rn, boolean release) {
+        if (release) {
+            stlxr(size, rs, rt, rn);
+        } else {
+            stxr(size, rs, rt, rn);
+        }
+    }
+
     /**
      * Conditional move. dst = src1 if condition else src2.
      *
--- a/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64LIRGenerator.java
+++ b/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64LIRGenerator.java
@@ -266,12 +266,12 @@ public abstract class AMD64LIRGenerator extends LIRGenerator {
     }
 
     @Override
-    public Variable emitLogicCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, Value trueValue, Value falseValue) {
+    public Variable emitLogicCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, Value trueValue, Value falseValue, boolean useBarriers) {
         return LIRValueUtil.asVariable(emitCompareAndSwap(true, accessKind, address, expectedValue, newValue, trueValue, falseValue));
     }
 
     @Override
-    public Value emitValueCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue) {
+    public Value emitValueCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, boolean useBarriers) {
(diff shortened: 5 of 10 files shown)


compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
```diff
@@ -80,23 +85,35 @@ public void emitCode(CompilationResultBuilder crb, AArch64MacroAssembler masm) {
...
+                masm.dmb(AArch64Assembler.BarrierKind.ANY_ANY);
+                emitCompareAndSwap(masm, size, address, result, newVal, false /* acquire */, false /* release */);
+                masm.dmb(AArch64Assembler.BarrierKind.ANY_ANY);
+            } else {
+                emitCompareAndSwap(masm, size, address, result, newVal, true /* acquire */, true /* release */);
+                if (!AArch64LIRFlagsVersioned.useLSE(masm.target.arch)) {
+                    // Temporarily add a fullbarrier after STLXR. This can be removed when LDAR is
+                    // used for java volatile read.
+                    masm.dmb(AArch64Assembler.BarrierKind.ANY_ANY);
+                }
```
XiaohongGong: Temporarily insert a full barrier after each atomic operations, even though it is a non-conservative model. This is added here to resolve the renassiance banchmark hang issue (https://github.com/oracle/graal/issues/2045) when "STLXR" is used together with "LDR/DMB" for java volatile read. This issue can also be resolved once "LDAR/STLR" is used for java volatile access. So once the PR https://github.com/oracle/graal/pull/1772 is merged, the temporarily codes can be deleted.

XiaohongGong:
Hi there, could anyone please take a look at this PR? Thank you so much!

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
adinn: Yes that will do until the other PR gets pushed.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGeneratorTool.java
```diff
@@ -155,6 +155,15 @@ default boolean mayEmbedConstantLoad(Constant constant) {
 
     Value emitValueCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue);
 
+    default Variable emitJavaLogicCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue,
```
tkrodriguez: I think it would be cleaner to pass a new flag into the existing methods instead of adding a whole new method.  Other platforms might ignore it but that's fine.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/AbstractCompareAndSwapNode.java
```diff
@@ -50,6 +50,7 @@
     @Input ValueNode expectedValue;
     @Input ValueNode newValue;
     @OptionalInput(State) FrameState stateAfter;
+    protected final boolean isJavaVolatile;
```
tkrodriguez: I think I understand the practical meaning of this flag but I'm not sure the naming is clear enough.  So basically in ARM there are 2 ways of doing volatiles and this flag is meant to distinguish between these 2 cases?  In almost all cases Java volatile should be considered the normal one I think and maybe this flag should be something like isNonStandardVolatile?  Actually what are there cases where we use a CompareAndSwapNode against a location which is acquired using the other sequence?

adinn: There are indeed two ways of generating compare and swap code. The choice of generation strategy depends on two factors: what location is being updated and what value has been set for a AArch64 flag (UseBarriersForVolatile).

If UseBarriersForVolatile is true then all compare and swap operations use ldxr and stxr combined with dmb instructions. 

If UseBarriersForVolatile is false then the nature of the memory location comes into play.

When the location is a Java field then ldaxr and stlxr need to be be used. This is consistent with the semantics employed for generating Java volatile reads and implemented in PR 1772 where they translate to ldar and stlr.

However, since PR has not yet been merged1772 XiahongGong has added an extra dmb at the end of ldax/stlxr sequence to deal with the disparity between the translation this patch implements and the current translation for volatiles a combination of ldr/str with dmb instructions.

When the location is not a Java field then the code must use ldxr, stxr combined with dmb instructions. This is consistent with normal expectations regarding memory synchronization  non-Java data.

tkrodriguez: Ok, but why are the other locations that are using isJavaVolatile == false correct?  Are all of those clearly non-Java locations?  For instance in the context of SVM I think everything is a Java volatile.

adinn: As I understand it (I hope XiaohongHong will correct me if I misrepresent things) XiaohongHong has only wired up the unsafe plugins for compare and exchange/swap to pass true. Those are cases that are known to correspond to a Java application program explicitly posting a read or write to Java data. The relaxing of the semanticsni these cases is compatible with the optimization implemented by PR 1772 when UseBarriersForVolatile is false and improves performance. Or rather it wil be when 1772 is merged and we can omit the tarailing barriers that are stil in place for this patch.

In other cases where code internal to the JIT or Substrate creates an UnsafeCompareAndSwapNode or UnsafeCompareAndSwapNode via the original constructor the value defaults to false which means the generator uses a more heavyweight barrier-based translation. This alternative is always consistent (on all known AArch64 arches) whether or not PR 1772 is in place. It is just less performant.

The reason the other cases have not been changed to use ldaxr/stlxr  is because it was not clear whether these internal uses were expecting other accesses to the target locations to be simple Java volatile field reads and writes (which would be correctly handled by the JIT once PR 1772 was merged) or to require the full semantics that is normally expected for compare and swap. The idea was to leave them as is and then convert the relevant uses one by one as needed. If you can confirm that all cases where the existing code explicitly constructs an UnsafeCompareAndSwapNode or UnsafeCompareAndSwapNode are compatible with PR 1772 then we don't need to worry about the isJavaVolatile == false case and can always use this optimization.

tkrodriguez: Yes that's what I'm getting at.  So for instance does the CompareAndSwapNode that we emit in the MonitorSnippets using Pointer.compareAndSwapWord need to use the conservative pattern because HotSpot uses a different pattern or can it use the more optimal pattern?  In the context of this PR it might be helpful to eliminate the constructor that passes the default value to make it more clear where we have to think about the setting of this flag.
In the context of SVM it's really only talking to itself I think so I would guess that always using the optimal pattern would be possible though it could necessary to have an escape hatch.
Apart from the Unsafe plugin, selecting the kind of volatile based on the LocationIdentity might work fairly well.  In the MonitorSnippets case for instance the MARK_WORD_LOCATION could be used to select the right kind for Hotspot if it needs something different but still do the normal thing for SVM.  Anyway, I'd like to get more clarity on which cases we actually need.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGeneratorTool.java
XiaohongGong: Yes, in a sense it should be better. I was expecting to not have too much modifications for other platforms, especially the modifications to the common existing API. That's why I decided to add a new function here.

XiaohongGong:
> This looks fine. Obviously it will need a further update when PR 1772 is pulled. I only spoted one details that needs fixing.
> 
> At AArch64LIRGenerator:178 you have misspelled method name 'emitComapreAndSwap'. Calls also need correcting.

Thanks for your review. I will fix it later then. Thanks again!

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/AbstractCompareAndSwapNode.java
XiaohongGong: We mainly concerns the usages in SVM, like in `JNIJavaVM.addJavaVM` and others in isolated, and its future usages in the internal VM of native image like GC. I'm not very much sure about where we should add the conservative barriers. But just in case, I think we should add a choice for the developer.
For the `MonitorSnippets `in HotSpot, interpreter and C1 always uses the conservative model while C2 uses the same one with Graal now. 
For the `isJavaVolatile` flag, do you mean we should eliminate the constructor that will set the default value to it, or just delete it because we could also decide the memory model based on the `LocationIdentity`?

adinn: >  Anyway, I'd like to get more clarity on which cases we actually need

Well, yeah, me too ;-) Only I'm not really in a position to determine that. That's why the current change was chosen: to be conservative and minimally invasive while still allowing further optimizations to be applied gradually as opportunities were  identified.

You are right that compilation for Substrate can probably always use the optimized model and that includes, say, the MonitorSnippets case. However, that code is currently shared with the Hotspot generation case. I'm not really clear how to separate those cases and apply the right transformation and I don't believe XiahongGong is. Advice would be very welcome.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/gen/LIRGeneratorTool.java
XiaohongGong: Update the patch to use a new flag!

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/AbstractCompareAndSwapNode.java
XiaohongGong: > At AArch64LIRGenerator:178 you have misspelled method name 'emitComapreAndSwap'. Calls also need correcting.

Fixed in the updated version! Thanks!

XiaohongGong:
> At AArch64LIRGenerator:178 you have misspelled method name 'emitComapreAndSwap'. Calls also need correcting.

Fixed in the updated version! Thanks!

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/AbstractCompareAndSwapNode.java
XiaohongGong: Hi @tkrodriguez , I have renamed the `isJavaVolatile` flag to `useBarriers` and eliminated the constructor that sets the default value to it. Besides, I did the modifications compared to the last version:
1. Add a `conservative` field for `LocationIdentity` to decide whether to use the full semantic barriers for VM internal atomic operations. I only set the flag for the CAS used `LocationIdentity `in `MonitorSnippets` currently. People can set the flag for other atomic operations if needed in future.
2. Remove the reference of `UseBarriersForVolatile`, always use the optimized `ldaxr/stlxr `pattern for java Unsafe atomics.

Could you please take a look whether this version is more reasonable? Thanks!

theRealAph: The reason HotSpot uses the "conservative" memory barriers is that it's very old code, and no-one could guarantee that "merely" sequentially-consistent operations would be adequate. However, I believe that SC is really adequate for HotSpot and we quietly used such operations in the AArch64 port for years until someone noticed. Graal shouldn't need anything more than SC, and if it does anywhere, that's a bug in Graal. IMO.

XiaohongGong: Hi @theRealAph, thanks for looking at this PR. So do you mean "ldaxr/stlxr" is SC enough for vm internal atomics like monitor operations in Graal?  Actually we didn't met any issue for it.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
theRealAph: That condition on "useLSE" doesn't make much sense to me. LSE instructions don't have full barrier semantics any more than LDAXR/STLXR instructions do.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/java/AbstractCompareAndSwapNode.java
theRealAph: Yes, that's what I mean. However, it is possible that there is (broken) code elsewhere in Graal which assumes monitors, CAS, etc. have full-barrier semantics. I have no idea how to find such code.

XiaohongGong: Yes, I'm also worried about the possible scenario and actually have no idea to make a decision. However, the main purpose of this PR is to add the full-barrier alternative implementation for CAS, so that people can have a chance to use it if needed in future.  If there is not any known issue for the current pattern `(ldaxr/stlxr)` used in hotspot monitor (since we have used it for long times in C2), I think we can keep using it currently in Graal. But it's still worth to add the full-barrier alternative implementation in Graal.

tkrodriguez:
I'm looking at it now.  Can you fix the style problems and rebase to a recent master?

XiaohongGong:
> I'm looking at it now. Can you fix the style problems and rebase to a recent master?

Sure! I'v rebased my patch and the patch is currently in our internal testing. I'll update it here once the tests pass. BTW, the style issue exists for each PR. I noticed there is already a PR in review to fix it (https://github.com/oracle/graal/pull/2443#issuecomment-627569196).

tkrodriguez:
Sorry for the close/reopen.  Accidental button push.

tkrodriguez:
Could we set the useBarrier flag much later and without having to modify LocationIdentity for a very platform dependent property?   Something like adding NodeLIRBuilderTool.useConservativeAtomics(LocationIdentity identity) and using that in the generate methods?

XiaohongGong:
> Could we set the useBarrier flag much later

Yes, I think we can avoid to make the "useBarrier" as a final field. It can be a local variable which is decided by `LocationIdentity`. 

> NodeLIRBuilderTool.useConservativeAtomics(LocationIdentity identity)

Do you mean this can directly decide whether the input `identity` is conservative or not?  The implementation on AArch64 should be something like:
```
public boolean useConservativeAtomics(LocationIdentity identity) {
    if (identity == OBJECT_MONITOR_OWNER_LOCATION) {
        return true;
    }
   return false;
}
```

And people can add other `LocationIdentity` if it needs conservative barriers.

XiaohongGong:
Hi @tkrodriguez , I'v rebased the patch and updated it according to your comment.  Since PR 1772 has been merged, the new patch has removed the temporarily added `DMBs` after each `STLXR`.  So for the patch itself, it just adds an alternative implementation for "CAS", which has the full barrier semantics. The final `"CAS" ` pattern is decided by the flag `useBarriers` which is decided by the operated memory `LocationIdentity`.  Since we prefer to use the optimized `LDAXR+STLXR` pattern, it is false by default for all location now. It can be set to true by checking the memory `LocationIdentity`  for the VM internal `CAS` operation if it needs the full barrier semantic.

Could you please take a look at it again? Thank you so much!

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
XiaohongGong: This temporary codes have been removed since PR 1772 has been merged. Thanks!

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64LIRGenerator.java
```diff
@@ -151,21 +151,23 @@ public AArch64AddressValue asAddressValue(Value address) {
...
     @Override
-    public Variable emitLogicCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, Value trueValue, Value falseValue) {
-        Variable prevValue = newVariable(expectedValue.getValueKind());
-        Variable scratch = newVariable(LIRKind.value(AArch64Kind.DWORD));
-        append(new CompareAndSwapOp(prevValue, loadReg(expectedValue), loadReg(newValue), asAllocatable(address), scratch));
+    public Variable emitLogicCompareAndSwap(LIRKind accessKind, Value address, Value expectedValue, Value newValue, Value trueValue, Value falseValue, boolean useBarriers) {
+        emitCompareAndSwap(address, expectedValue, newValue, useBarriers);
         assert trueValue.getValueKind().equals(falseValue.getValueKind());
         Variable result = newVariable(trueValue.getValueKind());
-        append(new CondMoveOp(result, ConditionFlag.EQ, asAllocatable(trueValue), asAllocatable(falseValue)));
```
tkrodriguez, line 160: Shouldn't this be checking that true and false are 1 and 0 respectively?

tkrodriguez:
Yes that looks good.  Are you going to add the LocationIdentities to the new function?   We could do add a subclass like ConservativeAtomicLocationIdentity in HotSpotReplacementsUtil and check for instances of that type in useConservativeAtomics. That way we don't have to simply enumerate them.

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64LIRGenerator.java
XiaohongGong: It's 1 and 0 respectively. However, I think adding an assert is better. I will fix it. Thanks for reminding!

XiaohongGong:
> Yes that looks good. Are you going to add the LocationIdentities to the new function? We could do add a subclass like ConservativeAtomicLocationIdentity in HotSpotReplacementsUtil and check for instances of that type in useConservativeAtomics. That way we don't have to simply enumerate them.

Yeah, good idea! I think it's ok to add a special LocationIdentity. Thanks!

XiaohongGong:
> We could do add a subclass like ConservativeAtomicLocationIdentity in HotSpotReplacementsUtil and check for instances of that type in useConservativeAtomics.

So do you think if it's better to make the `ConservativeAtomicLocationIdentity` common for all VMs, while not hotspot special? Even if there is no need to add the conservative barriers for SubstrateVM now, I think we'd better to also make it open for SVM in case it is needed in future.

tkrodriguez:
I would be inclined to make it local to hotspot right now.  It can always be moved later if it's needed, which I think is unlikely.

tkrodriguez:
Don't you actually need to create some instances of HotSpotConservativeLocationIdentity?

XiaohongGong:
> Don't you actually need to create some instances of HotSpotConservativeLocationIdentity?

No, I don't. Actually I'm not sure which CAS must needs the conservative barriers. `MonitorSnippet` is the only one vm internal CAS usage in hotspot, which currently uses the optimized pattern like the C2 compiler. Considering that there is not any actual issue related to it, we'd keep the optimized CAS for it.  What do you think about it? Any suggestion is appreciate.

XiaohongGong:
Hi @tkrodriguez , any update about this PR? Thanks a lot if you can have a review for it!

XiaohongGong:
Hi @tkrodriguez  how is this PR going now? Thanks!

tkrodriguez:
The useConservativeAtomics stuff seems useless to me.  I guess at some point in this discussion it was decided that we never need to use them?  If so let's remove useConservativeAtomics and act as if it's always false.  I we need it back for some reason then we can add it.  So if you make that change and rebase I can push this.

XiaohongGong:
> The useConservativeAtomics stuff seems useless to me. I guess at some point in this discussion it was decided that we never need to use them? If so let's remove useConservativeAtomics and act as if it's always false. I we need it back for some reason then we can add it. So if you make that change and rebase I can push this.

Hi @tkrodriguez , thanks for going back to this PR. Yeah, I think it is not used currently. Regarding to removing the stuff, do you mean only removing the flag `useConservativeAtomics`, or removing all the related codes like `HotSpotConservativeLocationIdentity` and the flag `useBarriers`? If only removing ``useConservativeAtomics``, I think always setting "useBarries" to "false" is ok,  otherwise, I think we can just abandon this PR directly.  So which do you think is better? Thanks!

tkrodriguez:
Removing useConservativeAtomics and HotSpotConservativeLocationIdentity and passing false for useBarriers sounds fine.

XiaohongGong:
> Removing useConservativeAtomics and HotSpotConservativeLocationIdentity and passing false for useBarriers sounds fine.

OK, thank you so much! I will update the patch soon!

XiaohongGong:
Hi @tkrodriguez I'v updated the patch to the latest one. Could you please take a look at it again? Thanks!

Issue #2192, opened 2020-02-26
Unexpected "Unsupported type jdk.internal.reflect.MethodAccessorGenerator is reachable" error


galderz:
Just spotted a Quarkus/Graal master issue, see (https://github.com/quarkusio/quarkus/issues/7422):

```
$ mvn clean install -Dnative -pl integration-tests/optaplanner-jackson
...
Error: com.oracle.svm.hosted.substitute.DeletedElementException: Unsupported type jdk.internal.reflect.MethodAccessorGenerator is reachable
To diagnose the issue, you can add the option --report-unsupported-elements-at-runtime. The unsupported element is then reported at run time when it is accessed the first time.
Detailed message:
Trace:
	at parsing jdk.internal.reflect.ReflectionFactory.generateConstructor(ReflectionFactory.java:513)
Call path from entry point to jdk.internal.reflect.ReflectionFactory.generateConstructor(Class, Constructor):
	at jdk.internal.reflect.ReflectionFactory.generateConstructor(ReflectionFactory.java:513)
	at jdk.internal.reflect.ReflectionFactory.newConstructorForSerialization(ReflectionFactory.java:506)
	at java.io.ObjectStreamClass.getSerializableConstructor(ObjectStreamClass.java:1516)
	at java.io.ObjectStreamClass$2.run(ObjectStreamClass.java:509)
	at java.io.ObjectStreamClass$2.run(ObjectStreamClass.java:484)
	at com.oracle.svm.core.jdk.Target_java_security_AccessController.doPrivileged(SecuritySubstitutions.java:81)
	at sun.security.pkcs11.SunPKCS11.initToken(SunPKCS11.java:1028)
	at sun.security.pkcs11.SunPKCS11$TokenPoller.run(SunPKCS11.java:852)
	at java.lang.Thread.run(Thread.java:834)
	at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:517)
	at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:193)
	at com.oracle.svm.core.code.IsolateEnterStub.PosixJavaThreads_pthreadStartRoutine_e1f4a8c0039f8337338252cd8734f63a79b5e3df(generated:0)
```

I suspect this is caused by the following change https://github.com/oracle/graal/commit/eab7e93b585febf7c2c68948af70157732767963

The call path above is not under Quarkus control, so I'd assume this is a regression on svm?

[...]

PR #2194, opened 2020-02-26
ReflectionFactory.generateConstructor should return null #2192

galderz:
Fix for #2192


dmlloyd:
I think a better fix would be to restore the generated serialization constructor and just treat it the same way as a regular reflection constructor.  @olpaw looks like this was from a change that you put in: eab7e93b585

Then if someone needs one of these partial constructors for a special serialization library, they can grab it during static initialization.

BTW, would it be possible to please ensure that there are PRs for every change?  It makes it much easier to keep track of what's going on - when things are pushed directly into master we always end up being surprised. /cc @maxandersen

olpaw:
See https://github.com/oracle/graal/issues/2192#issuecomment-591955700

galderz:
Closing since this is not the right fix for this.

PR #2204, opened 2020-03-02
Replace java.lang.Compiler.command usage with proper native-image API methods

olpaw:
`java.lang.Compiler.command` is deprecated and will eventually be removed in the next LTS. Thus we have to get rid of our "use" of `java.lang.Compiler.command` (exposing SubstrateVM specific functionality via Compiler.command. See https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/CompilerCommandPlugin.java) and instead create proper native-image API methods this functionality.


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
```diff
@@ -200,8 +209,12 @@ public static int run(int paramArgc, CCharPointerPointer paramArgv) {
      * @return maximum length of C chars that can be stored in the program argument part of the
      *         contiguous memory block without writing into the environment variables part.
      */
-    public static long getCRuntimeArgumentBlockLength() {
-        VMError.guarantee(argv.notEqual(WordFactory.zero()) && argc > 0, "Requires JavaMainWrapper.run(int, CCharPointerPointer) entry point!");
+    public static int getCRuntimeArgumentBlockLength() {
+        try {
+            checkArgumentBlockSupported();
+        } catch (UnsupportedOperationException e) {
```
christianwimmer: I don't like catching exceptions as normal control flow (and on Windows this is normal control flow). Better would be to make the `check` method return a boolean and in the callers either return `-1` or throw an exception.

olpaw: Will do.

PR #1546, opened 2019-07-26
Support latest Windows SDK

pquiring:
This one line change "should" allow substrateVM to build on latest Windows SDK.
The legacy_stdio_definitions.lib includes old names for stdio.
See bug#948 for more info.

PR #2208, opened 2020-03-03
Add legacy_stdio_definitions.lib for a successful native-image with VS2017

mardukbp:
When running `native-image` on Windows 10 with the VS2017 build tools, the linker complains that symbols related to printing functions are missing (e.g. unresolved external symbol __imp_sprintf , etc.). As mentioned in the PR #1546  adding `legacy_stdio_definitions.lib` solves this issue. However, this PR has not been merged because apparently its author had some issue signing the OCA.


pejovica:
Closing as won't fix.

Static libs compiled with VS2010, which is the case on JDK 8, are generally incompatible with newer VS versions and adding this would not change much in that regard.

PR #2213, opened 2020-03-03
Class initialization optimisation

jovanstevanovic:
For every static field or static method access, to those classes that we need to initialize at runtime, the class initialization method is generated only once.


compiler/src/org.graalvm.compiler.phases/src/org/graalvm/compiler/phases/RemovingUnneededClassInitClosure.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    protected EconomicSet<String> processBlock(Block block, EconomicSet<String> currentState) {
+        for (Node node : block.getNodes()) {
+            if (node instanceof InvokeWithExceptionNode) {
+                InvokeWithExceptionNode invokeWithExceptionNode = (InvokeWithExceptionNode) node;
+                if (invokeWithExceptionNode.callTarget() != null && invokeWithExceptionNode.callTarget().targetMethod() != null
+                        && invokeWithExceptionNode.callTarget().targetMethod().getName().equals("ensureInitialized")
+                        && invokeWithExceptionNode.callTarget().targetMethod().getDeclaringClass().getUnqualifiedName().equals("Class")) {
+
+                    JavaConstant javaConstant = invokeWithExceptionNode.callTarget().arguments().get(0).asJavaConstant();
```
vjovanov: Make 0 a constant and give it a descriptive name.

compiler/src/org.graalvm.compiler.phases/src/org/graalvm/compiler/phases/RemovingUnneededClassInitClosure.java
```diff
@@ -0,0 +1,84 @@
...
+                    if (javaConstant != null) {
+                        String hashCode = Integer.toString(javaConstant.hashCode());
+                        if (!currentState.contains(hashCode)) {
+                            currentState.add(hashCode);
```
vjovanov: There can be collisions, you need to use something unique.

PR #2215, opened 2020-03-04
Move analysis universe cleanup before Feature.beforeCompilation() call.

cstancu:
Feature.beforeCompilation() should not depend on the analysis universe.


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java
```diff
@@ -565,31 +565,31 @@ private void doRun(Map<Method, CEntryPointData> entryPoints,
...
+
                 ImageHeapLayouter heapLayouter = ImageSingletons.lookup(ImageHeapLayouter.class);
                 heap = new NativeImageHeap(aUniverse, hUniverse, hMetaAccess, heapLayouter);
 
                 BeforeCompilationAccessImpl config = new BeforeCompilationAccessImpl(featureHandler, loader, aUniverse, hUniverse, hMetaAccess, heap, debug, runtime);
                 featureHandler.forEachFeature(feature -> feature.beforeCompilation(config));
 
                 runtime.updateLazyState(hMetaAccess);
-
-                bigbang.getUnsupportedFeatures().report(bigbang);
```
christianwimmer: This line should remain the last one before the `catch`, otherwise we could miss unsupported features that get accumulated in the `UnsupportedFeatures` object.

cstancu: The `ShortestInvokeChainPrinter` used by the unsupported features reporting needs `AnalysisMethod.typeFlow` which is cleaned up by  `bigbang.cleanupAfterAnalysis()`. I think we should actually move everything after ` bigbang.cleanupAfterAnalysis();` outside of the catch, `beforeCompilation` should not accumulate unspported features.

christianwimmer: Yes, sounds good

PR #2227, opened 2020-03-06
Add watchdog to diagnose deadlocks during image building

christianwimmer:


substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/util/CompletionExecutor.java
```diff
@@ -164,6 +166,7 @@ public void execute(DebugContextRunnable command) {
                         if (timing != null) {
                             startTime = System.nanoTime();
                         }
+                        heartbeatCallback.run();
```
olpaw: For a series of `com.oracle.graal.pointsto.util.CompletionExecutor#execute`s where isSequential() returns true wouldn't we also want `heartbeatCallback.run()` to be called for each?

christianwimmer: The sequential mode is more a relic from the past for testing, but for completeness I added it there too.

PR #2236, opened 2020-03-08
Allow Windows-dependent conditional compilation

gwenneg:
This PR makes os-dependent conditional compilation available for Windows.

Now that Windows is no longer an experimental platform in GraalVM, this feature will certainly be needed in the near future in Quarkus.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/os/IsDefined.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/os/IsDefined.java
@@ -47,6 +47,11 @@ public class IsDefined {
     }
 
     @Fold
+    public static final boolean isWindows() {
+        return Platform.includedIn(Platform.WINDOWS.class);
+    }
+
+    @Fold
     public static final boolean __solaris__() {
         return false;
     }


pejovica:
@gwenneg For this purpose, we usually just use `Platform.includedIn(Platform.WINDOWS.class)`, which is also a public API.

On the other hand, the `IsDefined` class is actually obsolete, but unfortunately we missed it when we removed the old substitution code (98e43052b1279181cc115c0795f88709ae304575). So while there is no need for you to do this, if you would like to prepare a PR that removes it altogether, that would be great.

gwenneg:
Thanks for your answer @pejovica.

Does `Platform.includedIn(Platform.WINDOWS.class)` allow os-dependent conditional compilation like what I proposed in this PR? (the `@Fold` annotation is really important here)

We're relying on `@Fold` to do some os-dependent conditional compilation in the current Quarkus codebase (see [here](https://github.com/quarkusio/quarkus/blob/master/core/runtime/src/main/java/io/quarkus/runtime/graal/DiagnosticPrinter.java#L47)). Will it still be possible in the future?

pejovica:
> Does `Platform.includedIn(Platform.WINDOWS.class)` allow os-dependent conditional compilation like what I proposed in this PR? (the @Fold annotation is really important here)

Yes, because calls to `Platform.includedIn` are also constant folded (as stated in javadoc).

> We're relying on IsDefined and @Fold to do some os-dependent conditional compilation in the current Quarkus codebase (see here). Will it still be possible in the future?

Certainly, except that once `IsDefined` is removed, you may need to adjust a bit.

gwenneg:
That's good to know, thanks for this information @pejovica!

pejovica:
Closing as won't fix, since we already have a suitable API.

PR #2240, opened 2020-03-09
[wasm] Use numArgs to decrement stackPointer

fniephaus:
`numArgs` is a PE constant, so this should not affect compiled code. However, this avoids two redundant length checks in the interpreter.


fniephaus:
Outdated.

PR #2241, opened 2020-03-10
native-image: Include the original cause that led to the failure of CollectionPolicy instantiation

jaikiran:
In Quarkus, we have seen more than one user reporting an exception, during the native-image generation, of the form:

```
Error: policy com.oracle.svm.core.genscavenge.CollectionPolicy cannot be instantiated.
com.oracle.svm.core.util.UserError$UserException: policy com.oracle.svm.core.genscavenge.CollectionPolicy cannot be instantiated.
        at com.oracle.svm.core.util.UserError.abort(UserError.java:65)
        at com.oracle.svm.core.genscavenge.HeapPolicy.instantiatePolicy(HeapPolicy.java:87)
        at com.oracle.svm.core.genscavenge.CollectionPolicy.getInitialPolicy(CollectionPolicy.java:59)
        at com.oracle.svm.core.genscavenge.GCImpl.<init>(GCImpl.java:148)
        at com.oracle.svm.core.genscavenge.HeapImpl.<init>(HeapImpl.java:106)
        at com.oracle.svm.core.genscavenge.graal.HeapFeature.afterRegistration(HeapFeature.java:65)
        at com.oracle.svm.hosted.NativeImageGenerator.lambda$setupNativeImage$11(NativeImageGenerator.java:812)
        at com.oracle.svm.hosted.FeatureHandler.forEachFeature(FeatureHandler.java:63)
        at com.oracle.svm.hosted.NativeImageGenerator.setupNativeImage(NativeImageGenerator.java:812)
        at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:528)
        at com.oracle.svm.hosted.NativeImageGenerator.lambda$run$0(NativeImageGenerator.java:445)
```
like here https://github.com/quarkusio/quarkus/issues/7684 and here https://github.com/quarkusio/quarkus/issues/2414

The error reported doesn't include the root cause which caused the failure. 

The commit here now includes the original cause of the exception to help in situations like these.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapPolicy.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapPolicy.java
@@ -85,7 +85,7 @@ public class HeapPolicy {
         try {
             result = policy.getDeclaredConstructor().newInstance();
         } catch (Exception ex) {
-            throw UserError.abort("policy " + className + " cannot be instantiated.");
+            throw UserError.abort(ex, "policy " + className + " cannot be instantiated.");
         }
 
         if (!policyClass.isInstance(result)) {


dmlloyd:
I think that this exception is caused by incorrect escaping - the collection policy is actually a nested class of `CollectionPolicy`; since nested classes contain a `$`, if there is inadequate escaping of the class name and it passes through Docker (for example) which removes a layer of escaping, then it looks like an environment variable that is undefined, thus the suffix gets lost.

I'm not sure the additional exception cause will help debug this exact error, but it might be useful in other situations.

jaikiran:
> I'm not sure the additional exception cause will help debug this exact error, but it might be useful in other situations.

If that's what's happening, I think then this piece of code will try to instantiate (using the no-org constructor) the `com.oracle.svm.core.genscavenge.CollectionPolicy` which is an abstract class and the exception that gets thrown as a result of that (after this proposed change) will have a clear message stating that as the cause, which IMO should be an hint enough that the escaping went wrong.

jaikiran:
> which IMO should be an hint enough that the escaping went wrong

Well, now that you mentioned it @dmlloyd, I went back and looked at that exception that's being already thrown:
```
policy com.oracle.svm.core.genscavenge.CollectionPolicy cannot be instantiated
```
 and compared it to the code throwing it:
```
throw UserError.abort("policy " + className + " cannot be instantiated.");
```
So it's already printing the `className` as `com.oracle.svm.core.genscavenge.CollectionPolicy`, which means it's a clear hint that the escaping has caused it to the rest of the passed class name following the `$` sign.

So yes, like you note we may not need this change for the current issue, but it still can be useful in other failures.

jaikiran:
Closing this as this is no longer relevant and the linked issue in Quarkus ended up being solved in a different way.

Issue #2216, opened 2020-03-04
Add a phase to eliminate redundant constant mov.


XiaohongGong:
**Feature request**

Please include the following information:

**Is your feature request related to a problem? Please describe.**
A `ConstantNode` with multiple usages can be safely matched by a rule. So it will not be consumed during the match rule process even it might be actually unused after its usage is consumed. This does not have any influence for the constants inside 16-bits size. But if a constant is larger than 16-bits, it cannot be inlined and must be moved to a register on AArch64. So a redundant constant mov might exist in the generated codes.

For example, the original codes:
 
    lsr    w0, w0, #6
    orr    w1, wzr, #0xfffff
    and    w0, w0, w1

will be optimized to:

    orr    w1, wzr, #0xfffff
    ubfx   w0, w0, #6, #20

with the match rule:

    @MatchRule("(And (UnsignedRightShift=shift a Constant=b) Constant=c)")

Obviously the `orr w1, wzr, #0xfffff` is redundant that can be removed safely.

**Describe the solution you'd like.**
We can add a phase on LIR to remove the unused constant load like `ConstantLoadOptimization` in `PreAllocationOptimizationStage`.

**Describe who do you think will benefit the most.**
Any users of GraalVM.

**Describe alternatives you've considered.**

**Additional context.**
Add any other context about the feature request here.
For example, link to the relevant projects, documentation, standards.

**Express whether you'd like to help contributing this feature**
I'd like to contribute a PR to fix it. Actually I have prepared a patch for it and it is in our internal testing.

XiaohongGong:
Has created a PR here: https://github.com/oracle/graal/pull/2248.

PR #20, opened 2016-01-28
Fix ForeignAccess of ComplexNumbersB

grimmerm:
Bug fix

PR #2248, opened 2020-03-11
AArch64: Eliminate redundant constant load instructions.

XiaohongGong:
This patch fixes issue: https://github.com/oracle/graal/issues/2216

Unused constant load won't be deleted after match rules. As a result, it might generate redundant "mov/orr" instructions if the constants cannot be inlined to codes.

For example, the original codes:

      lsr    w0, w0, #6
      orr    w1, wzr, #0xfffff
      and    w0, w0, w1

will be optimized to the following pattern by adding match rules:

      orr    w1, wzr, #0xfffff
      ubfx   w0, w0, #6, #20

However, the constant `"0xfffff"` is not deleted even though it will never be used. To fix it, this patch inserts a phase applied to LIRInstructions. It's used to eliminate the unused constant load ops before code generation. Although it is limited to constant load ops currently, it can also be extended to remove other kinds of unused instructions if needed in future.

Change-Id: Ibed66714c12e02f400ab494683dee8826ede01a1

### DIFF
--- a/compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
+++ b/compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
@@ -49,6 +49,7 @@ import org.graalvm.compiler.core.match.ComplexMatchValue;
 import org.graalvm.compiler.core.match.MatchPattern;
 import org.graalvm.compiler.core.match.MatchRuleRegistry;
 import org.graalvm.compiler.core.match.MatchStatement;
+import org.graalvm.compiler.core.match.SharedMatchValue;
 import org.graalvm.compiler.debug.DebugCloseable;
 import org.graalvm.compiler.debug.DebugContext;
 import org.graalvm.compiler.debug.GraalError;
@@ -137,6 +138,7 @@ public abstract class NodeLIRBuilder implements NodeLIRBuilderTool, LIRGeneratio
 
     private final NodeMatchRules nodeMatchRules;
     private EconomicMap<Class<? extends Node>, List<MatchStatement>> matchRules;
+    private EconomicMap<Node, SharedMatchValue> sharedMatchValues;
 
     public NodeLIRBuilder(StructuredGraph graph, LIRGeneratorTool gen, NodeMatchRules nodeMatchRules) {
         this.gen = (LIRGenerator) gen;
@@ -146,6 +148,7 @@ public abstract class NodeLIRBuilder implements NodeLIRBuilderTool, LIRGeneratio
         OptionValues options = graph.getOptions();
         if (MatchExpressions.getValue(options)) {
             matchRules = MatchRuleRegistry.lookup(nodeMatchRules.getClass(), options, graph.getDebug());
+            sharedMatchValues = EconomicMap.create();
         }
         traceLIRGeneratorLevel = TTY.isSuppressed() ? 0 : Options.TraceLIRGeneratorLevel.getValue(options);
 
@@ -224,6 +227,19 @@ public abstract class NodeLIRBuilder implements NodeLIRBuilderTool, LIRGeneratio
         assert nodeOperands != null && nodeOperands.get(x) == null : "operand cannot be set twice";
         assert !(x instanceof VirtualObjectNode);
         nodeOperands.set(x, operand);
+    }
+
+    public void setSharedMatchResult(Node node) {
+        assert nodeOperands != null && nodeOperands.get(node) == null : "operand cannot be set twice";
+        SharedMatchValue matchValue = sharedMatchValues.get(node);
+        if (matchValue == null) {
+            matchValue = new SharedMatchValue();
+        }
+        matchValue.incrementUsageCount();
+        sharedMatchValues.put(node, matchValue);
+        if (node.getUsageCount() == matchValue.getUsageCount()) {
+            nodeOperands.set(node, matchValue);
+        }
     }
 
     public LabelRef getLIRBlock(FixedNode b) {
@@ -387,6 +403,9 @@ public abstract class NodeLIRBuilder implements NodeLIRBuilderTool, LIRGeneratio
                     } else if (ComplexMatchValue.INTERIOR_MATCH.equals(operand)) {
                         // Doesn't need to be evaluated
                         debug.log("interior match for %s", valueNode);
+                    } else if (operand instanceof SharedMatchValue) {
+                        // Doesn't need to be evaluated.
+                        debug.log("shared match value for %s", valueNode);
                     } else if (operand instanceof ComplexMatchValue) {
                         debug.log("complex match for %s", valueNode);
                         // Set current position to the position of the root matched node.
--- a/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
+++ b/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64ArithmeticLIRGenerator.java
@@ -371,7 +371,7 @@ public class AArch64ArithmeticLIRGenerator extends ArithmeticLIRGenerator implem
         }
     }
 
-    private void emitBinaryConst(Variable result, AArch64ArithmeticOp op, AllocatableValue a, JavaConstant b) {
+    public void emitBinaryConst(Variable result, AArch64ArithmeticOp op, AllocatableValue a, JavaConstant b) {
         AllocatableValue x = moveSp(a);
         getLIRGen().append(new AArch64ArithmeticOp.BinaryConstOp(op, result, x, b));
     }
--- a/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
+++ b/compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64NodeMatchRules.java
@@ -74,6 +74,7 @@ import org.graalvm.compiler.nodes.memory.MemoryAccess;
 import jdk.vm.ci.aarch64.AArch64Kind;
 import jdk.vm.ci.code.CodeUtil;
 import jdk.vm.ci.meta.AllocatableValue;
+import jdk.vm.ci.meta.JavaConstant;
 import jdk.vm.ci.meta.JavaKind;
 import jdk.vm.ci.meta.Value;
 
@@ -438,9 +439,11 @@ public class AArch64NodeMatchRules extends NodeMatchRules {
         }
         if ((0 == shift1 + shift2) || (src.getStackKind().getBitCount() == shift1 + shift2)) {
             return builder -> {
-                Value a = operand(src);
-                Value b = x instanceof LeftShiftNode ? operand(shiftAmt2) : operand(shiftAmt1);
-                return getArithmeticLIRGenerator().emitBinary(LIRKind.combine(a, b), AArch64ArithmeticOp.ROR, false, a, b);
+                AllocatableValue a = gen.asAllocatable(operand(src));
+                JavaConstant b = x instanceof LeftShiftNode ? shiftAmt2.asJavaConstant() : shiftAmt1.asJavaConstant();
+                Variable result = gen.newVariable(LIRKind.combine(a));
+                getArithmeticLIRGenerator().emitBinaryConst(result, AArch64ArithmeticOp.ROR, a, b);
+                return result;
             };
         }
         return null;
--- a/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
+++ b/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64ArithmeticLIRGenerator.java
@@ -692,21 +692,26 @@ public class AMD64ArithmeticLIRGenerator extends ArithmeticLIRGenerator implemen
     }
 
     private Variable emitShift(AMD64Shift op, OperandSize size, Value a, Value b) {
+        if (isJavaConstant(b)) {
+            return emitShiftConst(op, size, a, asJavaConstant(b));
+        }
         Variable result = getLIRGen().newVariable(LIRKind.combine(a, b).changeType(a.getPlatformKind()));
         AllocatableValue input = asAllocatable(a);
-        if (isJavaConstant(b)) {
-            JavaConstant c = asJavaConstant(b);
-            if (c.asLong() == 1) {
-                getLIRGen().append(new AMD64Unary.MOp(op.m1Op, size, result, input));
-            } else {
-                /*
-                 * c needs to be masked here, because shifts with immediate expect a byte.
-                 */
-                getLIRGen().append(new AMD64Binary.ConstOp(op.miOp, size, result, input, (byte) c.asLong()));
-            }
+        getLIRGen().emitMove(RCX_I, b);
+        getLIRGen().append(new AMD64ShiftOp(op.mcOp, size, result, input, RCX_I));
+        return result;
+    }
+
+    public Variable emitShiftConst(AMD64Shift op, OperandSize size, Value a, JavaConstant b) {
+        Variable result = getLIRGen().newVariable(LIRKind.combine(a).changeType(a.getPlatformKind()));
+        AllocatableValue input = asAllocatable(a);
+        if (b.asLong() == 1) {
+            getLIRGen().append(new AMD64Unary.MOp(op.m1Op, size, result, input));
         } else {
-            getLIRGen().emitMove(RCX_I, b);
-            getLIRGen().append(new AMD64ShiftOp(op.mcOp, size, result, input, RCX_I));
+            /*
+             * c needs to be masked here, because shifts with immediate expect a byte.
+             */
+            getLIRGen().append(new AMD64Binary.ConstOp(op.miOp, size, result, input, (byte) b.asLong()));
         }
         return result;
     }
--- a/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64NodeMatchRules.java
+++ b/compiler/src/org.graalvm.compiler.core.amd64/src/org/graalvm/compiler/core/amd64/AMD64NodeMatchRules.java
@@ -33,6 +33,7 @@ import static org.graalvm.compiler.asm.amd64.AMD64Assembler.AMD64BinaryArithmeti
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.AMD64RMOp.MOVSX;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.AMD64RMOp.MOVSXB;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.AMD64RMOp.MOVSXD;
+import static org.graalvm.compiler.asm.amd64.AMD64Assembler.AMD64Shift.ROL;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VADDSD;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VADDSS;
 import static org.graalvm.compiler.asm.amd64.AMD64Assembler.VexRVMOp.VMULSD;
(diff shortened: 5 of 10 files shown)


compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/RedundantInstructionElimination.java
```diff
@@ -0,0 +1,185 @@
...
+            };
+
+            InstructionValueConsumer useConsumer = (instruction, value, mode, flags) -> {
+                if (isVariable(value)) {
+                    DefEntry def = unUsedDefs.get((Variable) value);
+                    if (def != null) {
+                        assert def.value.equals(value);
+                        // Remove the instruction that defined this value from the unused map.
+                        LIRInstruction inst = def.instruction;
+                        unUsedDefs.filter(d -> (!d.instruction.equals(inst)));
```
zapster: At some point we tried to avoid allocating lambdas. I'd replace `d -> (!d.instruction.equals(inst))` with a static method (e.g. `RedundantInstructionElimination::filterDefEntry`).

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/RedundantInstructionElimination.java
```diff
@@ -0,0 +1,185 @@
...
+                if (instruction == null) {
+                    hasDead = true;
+                } else {
+                    instruction.setId(-1);
+                }
+            }
+
+            // delete unused instructions
+            if (hasDead) {
+                instructions.removeAll(Collections.singleton(null));
```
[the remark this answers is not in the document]
XiaohongGong: I'm not quite sure about this, but I think this is not guaranteed. For example `null` value LIRInstruction will cause `NullPointerException` in `MarkBasePointersPhase`. So I think it's necessary to delete them here.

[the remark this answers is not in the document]
zapster: Right, only later phases (after RA) can deal with that.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/RedundantInstructionElimination.java
XiaohongGong: Thanks for your review! Yeah, it makes sense. However, since `VariableMap` has the public API `public void filter(Predicate<T> predicate) ` to specially handle the filtration. I think it's better not to create another one. Another problem to use a static method in `RedundantInstructionElimination` is that the members of `VariableMap` are private and they are not expected to be public. We need to add another API to expose them if we want to go through them outside of class `VariableMap` .  But if you insist your opinion, I will try to fix it. Thanks!

XiaohongGong:
> (1) how often this happens 

This always happens if the conditions are matched. I tried to get the numbers of the deleted instructions by running the bootstrap and CTW.  And it has about `10` instructions for `bootstrap`, while about `260` instructions for `CTW`.

> (2) what the optimization gains in terms of performance

The performance is in testing. I will update the results later. Thanks!



> (3) what the phase costs in terms of compile time

It largely depends on the function complexity. The average costs is `3ms`. And the max cost is `358ms` for `bootstrap` and `408ms` for `CTW`. Most of the cases have lower than `5ms` cost. It only has the max cost for a small part cases.

> (4) whether this is also beneficial for AMD64 or if we should only do it on AArach64?

Currently this is not beneficial for AMD64, since constants can be inlined to codes on AMD64. But I think it might be useful for AMD64 in future if other kinds of instructions could be eliminated. Anyway I think it's ok to only do it on AArch64.

XiaohongGong:
> If possible such optimizations should be done on the graph (or at least during LIRGeneration) where we have more context. Although the Matcher is somewhere in-between, we might be possible to avoid emitting the instruction.

Yes, it's better to avoid the redundant instructions before LIRGeneration. I tried to eliminate it during Match rules, but it always has issues. The main problem is that the compiler doesn't know whether the `ConstantNode` is used or not after matching rules. The matcher won't delete its usages if it is matched.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/RedundantInstructionElimination.java
zapster: Ah, I missed that you are binding the local variable `inst`. Forget what I said.

zapster:
> It largely depends on the function complexity. The average costs is `3ms`. And the max cost is `358ms` for `bootstrap` and `408ms` for `CTW`. Most of the cases have lower than `5ms` cost. It only has the max cost for a small part cases.

What is that in relation to the overall compile time/overall backend time?

zapster:
> Anyway I think it's ok to only do it on AArch64.

Sounds good.

XiaohongGong:
> > It largely depends on the function complexity. The average costs is `3ms`. And the max cost is `358ms` for `bootstrap` and `408ms` for `CTW`. Most of the cases have lower than `5ms` cost. It only has the max cost for a small part cases.
> 
> What is that in relation to the overall compile time/overall backend time?

I'm sorry that I got a wrong average number of the costs yesterday. I took the time unit to `ms` which makes the accuracy not enough.  With the `System.nanoTime`, I got the new average number here: 

    Costs of this phase:    0.637ms
    Rate to whole compilation:    0.0067
    Rate to backend compilation:    0.0182

XiaohongGong:
> > Anyway I think it's ok to only do it on AArch64.
> 
> Sounds good.

Hi @zapster I have made this phase AArch64 special in my latest PR. Thanks again for your advice!

XiaohongGong:
> > (2) what the optimization gains in terms of performance
> 
> The performance is in testing. I will update the results later. Thanks!

I tested the performance with this micro jmh benchmark:

     private final int constantLoad(int length){
        int ret = 0;
        for (int i = 0; i < length; i++) {
            int a = (i >>> 4) & 0xfffff;
            int b = (i >>> 5) & 0x1fffff;
            int c = (i >>> 3) & 0x3ffffff;
            int d = (i >>> 6) & 0x7fffff;
            int e = (i >>> 7) & 0x3fffffff;
            int f = (i >>> 8) & 0xffffff;
            ret += a + b + c + d + e + f;
        }
        return ret;
    }

And the results:

                          Benchmark            Mode  Cnt       Score      Error  Units
    Without this phase:  jmhConstantLoad       avgt    5  245319.204 ? 1851.511  us/op
    With this phase:     jmhConstantLoad       avgt    5  244671.797 ? 54.937  us/op

It has slight improvement for this case. The constants inside of the loop will be optimized outside, and they are only loaded once. So I think that's the main reason of the low performance gain.

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/VariableMap.java
```diff
@@ -87,4 +85,38 @@ public void filter(Predicate<T> predicate) {
...
+        }
+
+        boolean empty = true;
+        for (T e : content) {
+            if (e != null) {
+                empty = false;
+                break;
+            }
+        }
+        return empty;
```
zapster, line 100: I'd rewrite this to 
```java
        for (T e : content) {
            if (e != null) {
                return false;
            }
        }
        return true;
```
Not a blocker, though.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64RedundantInstructionElimination.java
```diff
@@ -0,0 +1,190 @@
...
+            };
+
+            InstructionValueConsumer useConsumer = (instruction, value, mode, flags) -> {
+                if (isVariable(value)) {
+                    DefEntry def = unUsedDefs.get((Variable) value);
+                    if (def != null) {
+                        assert def.value.equals(value);
+                        // Remove the instruction that defined this value from the unused map.
+                        LIRInstruction inst = def.instruction;
+                        unUsedDefs.filter(d -> (!d.instruction.equals(inst)));
```
zapster, line 114: Since we are in SSA form, there can be only one instruction defining a variable. Thus, I think the code can be changed to `unUsedDefs.remove((asVariable(value))`.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64RedundantInstructionElimination.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        /**
+         * Collect unused instructions for a {@code block}.
+         */
+        private void collectUnusedInBlock(AbstractBlockBase<?> block) {
+            InstructionValueConsumer defConsumer = (instruction, value, mode, flags) -> {
+                if (isVariable(value) && isEligibleOp(instruction)) {
+                    DefEntry def = new DefEntry(block, instruction, value);
+                    unUsedDefs.put(((Variable) value), def);
```
zapster, line 103: We usually use `LIRValueUtil.asVariable(value)` instead of a direct cast, often with a static import so the code would change to `unUsedDefs.put(asVariable(value)), def);`. But it is a question of style preference and not a blocker.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64RedundantInstructionElimination.java
XiaohongGong: Actually, directly calling `remove` is all right for `ConstantLoadOp`. Thanks! My original concern is the `Def` with an array type like `@Def({REG, STACK}) private Value[] incomingValues;` of `LabelOp`. If an instruction has a definition with an array type. It actually has multiply definition values. And if any of the member is used, the instruction should not be removed. So I use the `filter` instead of "remove".  Anyway it seems this might not be possible for all the existed `ConstantLoadOp`.

zapster: > If an instruction has a definition with an array type. It actually has multiply definition values.

That is right. In that case, there is still no need to use `filter` (which can be expensive I think) but do something like this (more or less pseudocode):
```java
InstructionValueConsumer removeConsumer = (instruction, value, mode, flags) -> {
    if (isVariable(value)) {
        unUsedDefs.remove(asVariable(value));
    }
};
...
inst.visitEachOutput(removeConsumer);
```

XiaohongGong: OK, it seems reasonable! So as a conclusion, which do you think is better:  add these codes to handle the corner case, or just a directly `unUsedDefs.remove(asVariable(value))` is enough? I think the latter is ok. Your opinion? Thanks!

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64RedundantInstructionElimination.java
XiaohongGong: Godd catch! I will update it soon! Thank you!

compiler/src/org.graalvm.compiler.lir/src/org/graalvm/compiler/lir/VariableMap.java
XiaohongGong: Good catch! Thanks!

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64RedundantInstructionElimination.java
zapster: I agree. However, I'd add a comment why this is OK and an assertion that `inst` in no longer in any entry of the `unUsedDefs` map to catch error early if our assumption changes.

XiaohongGong: Good idea! I will update the patch soon! Thank you!

tkrodriguez:
@zapster yes this kind of problem is another reason that a high level transformation for matching would be better.  The main difficulty is that we'll need to either manually or automatically create generate container ValueNodes to replace the existing nodes.  Those nodes will have to implement whatever classes and interfaces define the semantics of the existing nodes.  It's likely there are a small set of those required.  Anyway, it just hasn't been a major priority.

tkrodriguez:
Regarding the fix itself, why can't we just avoid materializing those constants if they are swallowed by the matching?  It seems like some code is unnecessarily calling load or loadNonConst on those constants if all the users are folded away.

XiaohongGong:
Hi @zapster the latest patch is updated. Could you please take a look at it again? Thank you!

XiaohongGong:
> why can't we just avoid materializing those constants if they are swallowed by the matching?

I don't quite understand about this solution. Do you mean removing the IRs during matching rules?

tkrodriguez:
Generally if we have matched a IR node then we don't try to evaluate it at all at the LIR level.  I was thinking these would be marked as INTERIOR_MATCH and skipped [here](https://github.com/oracle/graal/blob/master/compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java#L387).  But I just realized that won't happen for constants which are allowed to have multiple users.  We could add something like a SharedMatchValue for those and track how many match usages are swallowing the constant.  If that number is less than the number of real usages then we emit it normally, i.e. as if getOperand returned null.  I think that would solve this efficiently.

XiaohongGong:
> Generally if we have matched a IR node then we don't try to evaluate it at all at the LIR level. I was thinking these would be marked as INTERIOR_MATCH and skipped [here](https://github.com/oracle/graal/blob/master/compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java#L387). But I just realized that won't happen for constants which are allowed to have multiple users. We could add something like a SharedMatchValue for those and track how many match usages are swallowing the constant. If that number is less than the number of real usages then we emit it normally, i.e. as if getOperand returned null. I think that would solve this efficiently.

Yeah, my initial idea is also to fix it during matching rules. However, I remember that I met some issues which might make the match codes complex. Anyway, I will try to rework it again! Thanks so much for your advice!

zapster:
> Hi @zapster the latest patch is updated. Could you please take a look at it again? Thank you!

The code looks good now but there seem to be travis CI issues (unused imports)

zapster:
On moving this forward, I think we can either push this (after fixing the style errors) or you continue investigation on a high-level optimization incorporating the ideas from @tkrodriguez. I think the latter should be done in any case since a high-level solution is more maintainable, but I'm also fine with merging this for now and reverting it afterwards if there is a better solution.

XiaohongGong:
Hi @zapster , thanks for your review! I'v updated the patch to resolve the unused import issue. I agree with you that the better solution is to fix it during matching rules. And I'm looking it at now.  However, I'm not sure whether it can works well, so I'm fine too to merge this for a temporary fix and revent it if the better solution is done. Thank you!

XiaohongGong:
Hi @zapster , I have already created a local fix to the matching codes. And it's in our internal testing now. So could you please not merge this patch for now? Maybe we could wait for the high-level fixing version. Thank you!

XiaohongGong:
Hi @tkrodriguez @zapster I'v updated the new version of the patch. Could you please take a look at it again? Thanks!

The new patch is almost a new one compared to the old version. It mainly fixes this issue during the matching process.  The modifications to the matching codes:
  1) Adds the `consumable` flag for matching. It is used to decide whether a node can be consumed by a matched rule. Originally it is decided by `singleUse`.
  2) Adds the class `SharedMatchValue`. It is used as the operand for a sharable and consumable node like `ConstantNode`, if the node is swallowed by the matched rules. It doesn't emit any codes. 

However, the consumed `ConstantNode` might be actually used during evaluating the match result `ComplexMatchValue`  by calling `operand(node)`. So even if it is only used for the matched rule, it might have the real usage. The compiler should emit its LIR if its operand is used.

zapster:
Since the Matcher is not my natural habitat I'm reassigning to @tkrodriguez. cc @dougxc

XiaohongGong:
> Since the Matcher is not my natural habitat I'm reassigning to @tkrodriguez. cc @dougxc

OK. thanks all the time!

tkrodriguez:
I'm happy to take over.

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
```diff
@@ -171,6 +174,10 @@ protected DebugInfoBuilder createDebugInfoBuilder(StructuredGraph graph, NodeVal
     @Override
     public Value operand(Node node) {
         Value operand = getOperand(node);
+        if (node instanceof ValueNode && operand instanceof SharedMatchValue) {
```
tkrodriguez, line 177: Why is this necessary?

XiaohongGong: This is to handle the case when `ConstantNode` is matched but not swallowed by the matched rule.  This is an example on aarch64:
```
 @MatchRule("(Add=op (UnsignedRightShift=x src Constant=shiftAmt1) (LeftShift src Constant=shiftAmt2))")
 public ComplexMatchResult rotationConstant(ValueNode op, ValueNode x, ValueNode src, ConstantNode shiftAmt1, ConstantNode shiftAmt2) {
    assert src.getStackKind().isNumericInteger();
    assert shiftAmt1.getStackKind().getBitCount() == 32;
    assert shiftAmt2.getStackKind().getBitCount() == 32;

    int shift1 = shiftAmt1.asJavaConstant().asInt();
    int shift2 = shiftAmt2.asJavaConstant().asInt();
    if (op instanceof AddNode && (0 == shift1 || 0 == shift2)) {
        return null;
    }
    if ((0 == shift1 + shift2) || (src.getStackKind().getBitCount() == shift1 + shift2)) {
    return builder -> {
                Value a = operand(src);
                Value b = x instanceof LeftShiftNode ? operand(shiftAmt2) : operand(shiftAmt1);
                return getArithmeticLIRGenerator().emitBinary(LIRKind.combine(a, b), AArch64ArithmeticOp.ROR, false, a, b);
            };
    }
    return null;
  }
```

The `ConstantNode` `shiftAmt1` and `shiftAmt2` are all matched to the rule. However, it's still used as an operand of the final `ROR`. See:
```
Value b = x instanceof LeftShiftNode ? operand(shiftAmt2) : operand(shiftAmt1);
```
The compiler cannot know the real usages inside of the `evaluate` function of the `ComplexMatchValue` during matching.

Another way to fix is to make it wrong when `operand()` of a matched `ConstantNode` is called. For the above rule, it needs some more modification at aarch64 backend which I don't think is friendly if fixing it.

tkrodriguez: I think if it wants them as a constant then it needs to construct them in that form instead of going through operand.  Why can't it call emitBinaryConst directly using the JavaConstant from the node?  You can't safely do potentially lazy evaluation using doRoot since it might be called in a non-dominating position.

XiaohongGong: Yes, I greatly agree with you. I was thinking that it needs more effort to fix the rules, since this case also exists on AMD64. It seems easier and simpler than what I thought before.  I will fix it later. Thanks!

XiaohongGong:
Hi @tkrodriguez I'v updated the patch. Please help to take a look at it. Thanks so much!

compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/gen/NodeLIRBuilder.java
```diff
@@ -226,6 +229,19 @@ public void setMatchResult(Node x, Value operand) {
         nodeOperands.set(x, operand);
     }
 
+    public void setSharedMatchResult(Node node) {
+        assert nodeOperands != null && nodeOperands.get(node) == null : "operand cannot be set twice";
+        SharedMatchValue matchValue = sharedMatchValues.get(node);
```
tkrodriguez, line 234: You know I don't really thing we need SharedMatchValue.  If we're going to maintain a side table we can just as easily store an Integer to track the count.  Then when it's equal to getUsageCount() we can store INTERIOR_MATCH into nodeOperands.

XiaohongGong: Yeah, agree! Actually the additional `SharedMatchValue` is not needed, and I think it's ok to remove it. Thanks for your review! And it's great that you could help to make it better. Thanks so much!

Issue #2220, opened 2020-03-05
Canonicalize negate node following a compatible right shift.


erik1iu:
**Feature request**

**Describe**

The pair `NegateNode` following a compatible `RightShiftNode` can be canonicalized to a singal `UnsignedRightShiftNode` . 

For example `-(x >> 31)`, the original codes:

```
sar    edx,0x1f        
neg    edx        
mov    eax,edx
```

can be optimized to

```
shr    edx,0x1f        
mov    eax,edx
```
Actually, it‘s a architecture independent optimization.

**Solution**
Improve the canonical phase in `NegateNode`.

**Status**
I'd like to contribute a PR to fix it. Actually I have prepared a patch for it and it is in our internal testing.

PR #63, opened 2016-02-19
Add loop node documentation. Make loop nodes a non-experimental feature.

chumer:
Long overdue documentation for loop nodes. They were also still marked as experimental, but they are used by most Truffle languages now.

PR #2249, opened 2020-03-11
Canonicalize negate node to unsigned right shift.

erik1iu:
This patch fixes issue: #2220 

Negate a right shift with the max shift amount equals to unsigned right
shift.

As the final code, just take AArch64 for example:

        asr x0, x2, #63
        neg x0, x0

After this patch, generated assembly above can be optimized to below:

        lsr x0, x2, #63

Change-Id: Ifd1c4f7885822b92798341f4728c887fddf7c654


erik1iu:
ping @tkrodriguez

tkrodriguez:
Looks good.  It's in the merge queue.

PR #2266, opened 2020-03-17
Avoid calling internal CoreFoundation methods on Darwin

loicottet:
This change ensures that iOS apps compiled with Graal can be accepted into the AppStore, as they don't use internal CoreFoundation methods anymore. This results in a slight startup performance decrease as shown by the numbers below, but this is acceptable to ensure apps will be approved.

Startup performance before the change
```
            Mean        Std.Dev.    Min         Median      Max
real        0.012       0.002       0.009       0.014       0.015
user        0.004       0.001       0.003       0.005       0.005
sys         0.005       0.001       0.004       0.005       0.006
```

Startup performance after the change
```
            Mean        Std.Dev.    Min         Median      Max
real        0.015       0.002       0.012       0.016       0.018
user        0.006       0.001       0.005       0.006       0.007
sys         0.006       0.001       0.004       0.006       0.008
```


olpaw:
> can be accepted into the AppStore, as they don't use internal CoreFoundation methods anymore

I agree. The ability to get images accepted into AppStore justifies the minor startup-time regression on OSX.

PR #2270, opened 2020-03-18
Introduced a small JVMTI agent framework and refactored native-image-agent.

gradinac:


substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NIHandles.java
```diff
@@ -0,0 +1,134 @@
...
+package com.oracle.svm.agent;
+
+import com.oracle.svm.agent.core.JavaHandles;
+import com.oracle.svm.jni.nativeapi.JNIEnvironment;
+import com.oracle.svm.jni.nativeapi.JNIMethodId;
+import com.oracle.svm.jni.nativeapi.JNIObjectHandle;
+
+import static com.oracle.svm.jni.JNIObjectHandles.nullHandle;
+
+public class NIHandles extends JavaHandles {
```
peter-hofer: I think `NativeImageAgentHandles` would be a better name.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/Agent.java
```diff
@@ -88,37 +72,34 @@
...
 import com.oracle.svm.core.configure.ConfigurationFiles;
 import com.oracle.svm.driver.NativeImage;
 import com.oracle.svm.jni.nativeapi.JNIEnvironment;
-import com.oracle.svm.jni.nativeapi.JNIErrors;
 import com.oracle.svm.jni.nativeapi.JNIJavaVM;
 import com.oracle.svm.jni.nativeapi.JNIObjectHandle;
-import com.oracle.svm.jni.nativeapi.JNIVersion;
 
-public final class Agent {
+public final class Agent extends AgentBase {
```
peter-hofer: Perhaps rename this to `NativeImageAgent` too. Then you could also rename `AgentBase` to `Agent` if you'd like.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/Agent.java
```diff
@@ -345,27 +324,19 @@ public static int onLoad(JNIJavaVM vm, CCharPointer options, @SuppressWarnings("
...
-        check(jvmti.getFunctions().SetEventNotificationMode().invoke(jvmti, JVMTI_ENABLE, JVMTI_EVENT_VM_INIT, nullHandle()));
-        check(jvmti.getFunctions().SetEventNotificationMode().invoke(jvmti, JVMTI_ENABLE, JVMTI_EVENT_VM_DEATH, nullHandle()));
-        check(jvmti.getFunctions().SetEventNotificationMode().invoke(jvmti, JVMTI_ENABLE, JVMTI_EVENT_THREAD_END, nullHandle()));
-
         setupExecutorServiceForPeriodicConfigurationCapture(configWritePeriod, configWritePeriodInitialDelay);
         return 0;
     }
 
-    private static int parseIntegerOrNegative(String number) {
+    private int parseIntegerOrNegative(String number) {
```
peter-hofer: Our style checking gate will likely complain that this could be declared `static`, other methods too.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/AgentBase.java
```diff
@@ -0,0 +1,212 @@
...
+
+    protected abstract void onVMStartCallback(JvmtiEnv jvmti, JNIEnvironment jni);
+
+    protected abstract void onVMInitCallback(JvmtiEnv jvmti, JNIEnvironment jni, @SuppressWarnings("unused") JNIObjectHandle thread);
+
+    protected abstract void onVmDeathCallback(JvmtiEnv jvmti, @SuppressWarnings("unused") JNIEnvironment jni);
+
+    protected abstract int getRequiredJVMTIVersion();
+
+    protected static JavaHandles handles;
```
peter-hofer: You can make this field non-static and access it via `singleton()`. In order to avoid casts, you could introduce a type parameter `<H extends JavaHandles>` that subclasses can specify for the field type and the return type of `constructJavaHandles`, but that might be overkill.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JavaHandles.java
```diff
@@ -0,0 +1,102 @@
...
+import com.oracle.svm.jni.nativeapi.JNIObjectHandle;
+import org.graalvm.nativeimage.c.type.CTypeConversion;
+
+import java.util.Arrays;
+import java.util.concurrent.locks.ReentrantLock;
+
+import static com.oracle.svm.core.util.VMError.guarantee;
+import static com.oracle.svm.jni.JNIObjectHandles.nullHandle;
+
+public abstract class JavaHandles {
```
peter-hofer: On second thought, perhaps `HandleSet` would be a better name.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/AgentBase.java
gradinac: I agree with this, however, I think we can't get away from having a static method in each agent subclass. Right now, `NativeImageAgent` has a static `handles()` method that does the type cast. For this approach, we would either have to re-implement singleton in the subclass with the appropriate cast, or keep a static handles method. If there are no other ways, I would go with reimplementing the `singleton()` method.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NIHandles.java
gradinac: I agree, will change it.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/Agent.java
gradinac: Yes, that would be perfect. Changing it.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/Agent.java
gradinac: I will make them static again.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JavaHandles.java
gradinac: I agree, will take care of it.

gradinac:
Thank you :)

SergejIsbrecht:
@gradinac, would it be possible to add the same callbacks from Hotspot for TLAB allocations in this PR?

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/HandleSet.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Arrays;
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.graalvm.nativeimage.c.type.CTypeConversion;
+
+import com.oracle.svm.jni.nativeapi.JNIEnvironment;
+import com.oracle.svm.jni.nativeapi.JNIMethodId;
+import com.oracle.svm.jni.nativeapi.JNIObjectHandle;
+
+public abstract class HandleSet {
```
vjovanov, line 39: I find HandleSet is an overly generic name. Can we give a name to this set that describes all the handles well.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/HandleSet.java
```diff
@@ -0,0 +1,103 @@
...
+    private final ReentrantLock globalRefsLock = new ReentrantLock();
+    private JNIObjectHandle[] globalRefs = new JNIObjectHandle[16];
```
vjovanov, line 41: What is 16?

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/HandleSet.java
gradinac: How about JNIHandleSet?

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/HandleSet.java
gradinac: The initial number of global references that are tracked. This array can grow dynamically, I think the size was chosen in accordance with the JVMTI per-thread guaranteed local handle capacity. but it could be different? CC @peter-hofer

vjovanov: It would be better to make it a const and give it a descriptive name.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/HandleSet.java
vjovanov: Would `BaseJVMTIHandleSet` or `JVMTIHandleSetBase` describe it well?

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JvmtiAgentBase.java
```diff
@@ -0,0 +1,186 @@
...
+import com.oracle.svm.core.c.function.CEntryPointOptions;
+import com.oracle.svm.core.c.function.CEntryPointSetup;
+import com.oracle.svm.core.util.VMError;
+import com.oracle.svm.jni.nativeapi.JNIEnvironment;
+import com.oracle.svm.jni.nativeapi.JNIErrors;
+import com.oracle.svm.jni.nativeapi.JNIJavaVM;
+import com.oracle.svm.jni.nativeapi.JNIObjectHandle;
+import com.oracle.svm.jni.nativeapi.JNIVersion;
+
+public abstract class JvmtiAgentBase<T extends HandleSet> {
```
dougxc, line 60: Where's the javadoc comment explaining the purpose and usage of this class?

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/HandleSet.java
dougxc: At least having a javadoc comment for the class would help clarify what kind of handles this class manages.

gradinac: @vjovanov I can rename it to `JVMTIHandleSetBase`. The reason for the `JNIHandleSet` is because the class manages `JNI` such as `JNIObjectHandle`/`JNIMethodId`. It could also be `JNIHandleSetBase`.
@dougxc I agree, I will add documentation for the class.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/HandleSet.java
gradinac: I agree, I will take care of it.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JvmtiAgentBase.java
gradinac: I will add documentation for this class.

gradinac:
> Excellent work! Hope this framework gets used by other agents as well.

Thank you :)

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JvmtiAgentBase.java
```diff
@@ -0,0 +1,275 @@
...
+     */
+    protected abstract int onUnloadCallback(JNIJavaVM vm);
+
+    /**
+     * JVMTI VMStart event callback.
+     *
+     * @param jvmti The JVMTI environment.
+     * @param jni The JNI environment of the thread running the JVMTI callback.
+     */
+    protected abstract void onVMStartCallback(JvmtiEnv jvmti, JNIEnvironment jni);
```
vjovanov, line 134: In this API I find inconsistencies with capitalization: why is VM here, and Vm later. Why is it `JNIJavaVM` but then we have a `JvmtiEnv`. We should stick to one naming convention in this API, whichever it is. \CC @peter-hofer

peter-hofer: `JNIEnvironment`, `JNIJavaVM` etc. are classes from our JNI implementation. JNI and JVMTI identifiers themselves also differ in that aspect: `jvmtiEnv` vs `JNIEnv`.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/HandleSet.java
peter-hofer: They are not JVMTI handles though (there is no such thing), these are a set of JNI handles frequently used by the agent, so I propose `AgentHandleSet`.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/HandleSet.java
peter-hofer: `INITIAL_GLOBAL_REFS_SIZE` or just a comment `// grows` should be fine.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JvmtiAgentBase.java
vjovanov: Well, if it is `onVMStartCallback` then it should be `onVMDeathCallback` instead of `onVmDeathCallback`. I will change this when merging.

peter-hofer: No argument there.

gradinac: I have updated it to `onVMDeathCallback`, that was an oversight on my part. I have also fixed some javadoc related warnings for `@link` embedded inside a `@see`.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
```diff
@@ -0,0 +1,176 @@
...
+
+import org.graalvm.nativeimage.c.type.CTypeConversion;
+
+import com.oracle.svm.jni.JNIObjectHandles;
+import com.oracle.svm.jni.nativeapi.JNIEnvironment;
+import com.oracle.svm.jni.nativeapi.JNIMethodId;
+import com.oracle.svm.jni.nativeapi.JNIObjectHandle;
+
+/**
+ * Helps with creation and manages JNI handles for JVMTI agents.
```
peter-hofer, line 41: `with the creation and management of`

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * A JVMTI agent must provide a subclass of this class. It should contain handles to classes and
+ * methods that are needed across different JVMTI calls (for example, a JNIObjectHandle to
+ * java/lang/Class).
+ *
+ * For JNI handles that are created in a JVMTI callback and that should survive a return from native
+ * code, a helper method {@link #newTrackedGlobalRef} is provided.
+ *
+ * See also: {@link com.oracle.svm.agent.core.JvmtiAgentBase}.
```
peter-hofer, line 50: You can use `@see` for this.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
```diff
@@ -0,0 +1,176 @@
...
+            javaLangClassGetName = Support.jniFunctions().getGetMethodID().invoke(env, javaLangClass, name.get(), signature.get());
+            guarantee(javaLangClassGetName.isNonNull());
+        }
+    }
+
+    /**
+     * Returns a local handle to a Java class object.
+     *
+     * If the class is not found, a RuntimeException is thrown. If the handle must survive a return
+     * from native code, {@link #newTrackedGlobalRef} should be used on the returned handle.
```
peter-hofer, line 72: `guarantee` at runtime is substituted and aborts the entire VM, see the substitution in `Target_com_oracle_svm_core_util_VMError.shouldNotReachHere`. The class _must_ exist.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
```diff
@@ -0,0 +1,176 @@
...
+    protected static JNIObjectHandle findClass(JNIEnvironment env, String className) {
+        try (CTypeConversion.CCharPointerHolder name = Support.toCString(className)) {
+            JNIObjectHandle h = Support.jniFunctions().getFindClass().invoke(env, name.get());
+            guarantee(h.notEqual(nullHandle()));
+            return h;
+        }
+    }
+
+    /**
+     * A convenience method to return a global handle to a Java class object.
```
peter-hofer, line 87: This documentation doesn't mention that the reference is freed eventually by this class and a user might believe they are responsible for freeing it. I would just reference method `newTrackedGlobalRef` here.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
```diff
@@ -0,0 +1,176 @@
...
+            if (isStatic) {
+                return Support.jniFunctions().getGetStaticMethodID().invoke(env, clazz, cname.get(), csignature.get());
+            } else {
+                return Support.jniFunctions().getGetMethodID().invoke(env, clazz, cname.get(), csignature.get());
+            }
+        }
+    }
+
+    /**
+     * Creates a global JNI handle of the specified local JNI handle.
```
peter-hofer, line 140: Document that this instance remembers the reference and destroys it in `destroy`.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
```diff
@@ -0,0 +1,176 @@
...
+            globalRefCount++;
+        } finally {
+            globalRefsLock.unlock();
+        }
+        return global;
+    }
+
+    /**
+     * Releases all global JNI handles. This class and the handles must not be used after a call to
+     * this method.
```
peter-hofer, line 167: This might be worth a `boolean destroyed` to check in asserts in the methods.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JvmtiAgentBase.java
```diff
@@ -0,0 +1,275 @@
...
+
+    /**
+     * Registers the agent singleton. This method must be called only once during a native image
+     * build.
+     *
+     * @param agentSingleton The agent implementation.
+     */
+    @SuppressWarnings({"rawtypes"})
+    protected static <T extends JNIHandleSetBase> void registerAgent(JvmtiAgentBase<T> agentSingleton) {
+        singleton = agentSingleton;
```
peter-hofer, line 188: Wasn't there a check that the agent hasn't been set before? I think it should stay.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageAgentJNIHandleSet.java
```diff
@@ -0,0 +1,132 @@
...
+    JNIObjectHandle javaLangClassLoader;
+    public JNIObjectHandle javaLangSecurityException;
+    public JNIObjectHandle javaLangNoClassDefFoundError;
+    public JNIObjectHandle javaLangNoSuchMethodError;
+    JNIObjectHandle javaLangNoSuchMethodException;
+    public JNIObjectHandle javaLangNoSuchFieldError;
+    JNIObjectHandle javaLangNoSuchFieldException;
+    JNIObjectHandle javaLangClassNotFoundException;
+    JNIObjectHandle javaLangRuntimeException;
+    JNIObjectHandle javaUtilMissingResourceException;
```
peter-hofer, line 50: Why are these no longer `final`?

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageTracingAgent.java
```diff
@@ -644,10 +605,10 @@ public static void onUnload(@SuppressWarnings("unused") JNIJavaVM vm) {
         }
 
         /*
-         * Agent shutdown is tricky: apparently we can still have events at the same time as this
-         * function executes, so we would need to synchronize. We could do this with a combined
-         * shared+exclusive lock, but that adds some cost to all events. We choose to leak a few
-         * handles and some memory for now -- this agent isn't supposed to be attached only
+         * JvmtiAgentBase shutdown is tricky: apparently we can still have events at the same time
```
peter-hofer, line 608: That's more about JVMTI agents in general. Leaving `Agent shutdown [...]` or using `JVMTI agent shutdown [...]` would be better.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
gradinac: I agree, will change it.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
gradinac: That's a good point, changed. I had tried {@see {@link}} but that didn't work - hence why it was this way.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
gradinac: I agree, the substitution changes the exception and that exception cannot be caught. I will update the docs.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
gradinac: I agree, will add a note for this.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
gradinac: Will document this part here too.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
gradinac: I agree, will add it. The user should not call `destroy` manually, it should be called from the agent framework once the agent unloads. Will add a note.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JvmtiAgentBase.java
gradinac: I agree, will add the check.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageAgentJNIHandleSet.java
gradinac: This might've been lost when I was tweaking the modifiers. I have re-added final.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageTracingAgent.java
gradinac: I have changed it back to `Agent shutdown [...]`.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
```diff
@@ -0,0 +1,188 @@
...
+ * A JVMTI agent must provide a subclass of this class. It should contain handles to classes and
+ * methods that are needed across different JVMTI calls (for example, a JNIObjectHandle to
+ * java/lang/Class).
+ *
+ * For JNI handles that are created in a JVMTI callback and that should survive a return from native
+ * code, a helper method {@link #newTrackedGlobalRef} is provided.
+ *
+ * @see com.oracle.svm.agent.core.JvmtiAgentBase
+ */
+public abstract class JNIHandleSetBase {
```
peter-hofer, line 52: I think this class is useful even without subclassing (although at least one instance is needed for the agent). I'd consider dropping the `Base` name suffix and change some methods from `protected` to `public`.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageTracingAgent.java
```diff
@@ -88,53 +70,58 @@
...
 import com.oracle.svm.jni.nativeapi.JNIEnvironment;
-import com.oracle.svm.jni.nativeapi.JNIErrors;
 import com.oracle.svm.jni.nativeapi.JNIJavaVM;
 import com.oracle.svm.jni.nativeapi.JNIObjectHandle;
-import com.oracle.svm.jni.nativeapi.JNIVersion;
+import org.graalvm.nativeimage.hosted.Feature;
 
-public final class Agent {
-    public static final String AGENT_NAME = "native-image-agent";
+public final class NativeImageTracingAgent extends JvmtiAgentBase<NativeImageAgentJNIHandleSet> {
```
peter-hofer, line 86: I'd rather keep the name `NativeImageAgent`, which is also consistent with the name of the build artifact. Tracing is only part of the agent functionality, the other major part being the restriction mode.

substratevm/src/com.oracle.svm.agent.core/src/com/oracle/svm/agent/core/JNIHandleSetBase.java
gradinac: I agree, I have dropped the base and changed the methods from protected to public.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageTracingAgent.java
gradinac: I agree with you, I have changed the name.

PR #2271, opened 2020-03-19
sulong: Disable LLsuite tests on AArch64

a74nh:
These patches will enable the LLsuite tests to fully pass.

I'm a little concerned about the long term viability of this suite on non x86-linux platforms. All the tests are .ll files which contain 'target triple = "x86_64-unknown-linux-gnu"'. I guess as long as all the tests just use generic IR nodes, then they should work cross platform.
I'm assuming the goal isn't to eventually, for example, run 80bit X86 specific nodes on non-X86.


a74nh:
Travis failures are building: /home/travis/build/oracle/graal/truffle/src/com.oracle.truffle.api.instrumentation.test/src/com/oracle/truffle/api/instrumentation/test/InstrumentationNoExceptionsTest.java

Which this patch doesn't touch

rschatz:
Hi Alan!

I ran your changes in our internal gate, and it seems to work fine. Looking at the travis history, these failures seem to have creeped into master. I'm not sure why they fail on travis but not in our internal gate, we should be running the same tests there. I'll investigate.

> I'm a little concerned about the long term viability of this suite on non x86-linux platforms. All the tests are .ll files which contain 'target triple = "x86_64-unknown-linux-gnu"'. I guess as long as all the tests just use generic IR nodes, then they should work cross platform.

Yes, these tests should all be relatively simple. The problem is that some bitcode pattern that we have observed to happen in practice are sometimes very hard to produce from small C programs, because they are generated only as part of certain optimizations in the context of bigger programs. That's what these tests are for.

If it turns out to be a problem, we can always make them completely platform specific.

Thanks for your work in this area! I'll merge this PR asap.

a74nh:
> > I'm a little concerned about the long term viability of this suite on non x86-linux platforms. All the tests are .ll files which contain 'target triple = "x86_64-unknown-linux-gnu"'. I guess as long as all the tests just use generic IR nodes, then they should work cross platform.
> 
> Yes, these tests should all be relatively simple. The problem is that some bitcode pattern that we have observed to happen in practice are sometimes very hard to produce from small C programs, because they are generated only as part of certain optimizations in the context of bigger programs. That's what these tests are for.
> 
> If it turns out to be a problem, we can always make them completely platform specific.

Ok, sure. That's a lot more reasonable.

> 
> Thanks for your work in this area! I'll merge this PR asap.

Thanks.

PR #2274, opened 2020-03-20
test use of regex to allow red Hat copyrights

adinn:
This PR is in part a test to check that we can use regex-based copyright checks in the substrate tree as per the existing checks in the compiler tree. This should allow extra copyrights (also the Intel Library credit) to pass the gate. If it works then the patch can be used to enable this new behaviour but see the caveat below.

This PR adds the required copyright regex pattern files in substratevm/mx.substratevm/copyrights. It also modifies the copyright on ObjectFile.java to test that an extra copyright is permissible. That latter change needs to be backed out before pulling the PR if this change works. The main purpose of the PR is not to check that this works (it is pretty clear that it does) but rather to ensure nothing else breaks.


adinn:
n.b. rebased changes on latest master after build failure

adinn:
Fixed this problem using checkstyle file instead

Issue #2287, opened 2020-03-25
ImageClassLoader holds on to large number of irrelevant Method(s) and Field(s)


jaikiran:
The `ImageClassLoader` seems to be responsible for initializing all classes on the classpath during native image generation (during classlist phase). During this process, it loads the class (through `Class.forName` with `initialize` param as `false`) and then for each such loaded class, it goes on to check if annotations can be loaded from that class (i.e. `AnnotatedElement.getAnnotations` doesn't throw any error). If it can't load the annotations on that class, for whatever reason, it skips this class and moves on to the next. 

For classes that can successfully load the annotations, it does further checks to see if there are certain Graal specific annotations that classify it as a "hosted only" class or an "application class". Based on this check, it stores that class in a relevant `Set` (called `applicationClasses` or `hostedOnlyClasses`). All this seems fine.

For each such identified class, it then goes on to "find system elements". This involves checking each method and field on the class to see if any annotations are available on them. If there are any annotations available on them, then the `ImageClassLoader` stores that Method/Field into a `Set` called `systemMethods`, `systemFields` respectively. These `Set`s are (only) used whenever there's a call on the `ImageClassLoader` to find and return any method/field that may have certain specific annotations (for example, the `CEntryPoint`, `CConstant`, `CFunction`, `NativeImageReinitialize` and such). The problem in this code is the current `annotationsAvailable` method which returns `true` even if there may be 0 annotations on the method/field:

```
private static boolean annotationsAvailable(AnnotatedElement element) {
        try {
            element.getAnnotations();
            return true;
        } catch (Throwable t) {
            handleClassLoadingError(t);
            return false;
        }
    }

```
This causes the `ImageClassLoader` to unnecessarily hold on to a large number of unnecessary `Method` and `Field` instances that aren't really relevant and thus cause both unnecessary additional memory usage as well as unnecessary time spent on iterating over these additional methods/fields whenever a call is made on the `ImageClassLoader` to find any specific annotation.

jaikiran:
I've issued https://github.com/oracle/graal/pull/2288 with a potential fix.

adinn:
Hmm, that looks worth having ;-) Nice work, Jaikiran!

jaikiran:
Hello @dougxc, I'm closing this since the linked PR has been merged (thanks!). However, I don't know what milestone to add as the fix version for this issue.

PR #2288, opened 2020-03-25
Fix 2287 - ImageClassLoader should only hold on to method/field that have annotations

jaikiran:
The commit in this PR fixes the issue noted in https://github.com/oracle/graal/issues/2287.

This commit changes the implementation of `annotationsAvailable` (private) method to return `true` only if there are any annotations available on the element. 
This commit also introduces a new `canLoadAnnotations` method which returns `true` if there were no errors like loading the annotations on an element (through the use of `AnnotatedElement#getAnnotations()`). This new method is used/necessary to retain the current semantics of how the `ImageClassLoader` decides to hold on to "hosted only classes" and "application classes" (i.e. these classes are held on to, even if they don't have any annotations, since those classes are then checked/used when there's a call on the `ImageClassLoader` to find any sub-classes of a specific type).

Here's the exact difference in the number of methods and fields against a sample application (based on Quarkus) that this `ImageClassLoader` holds on to:

Without this fix:
--------------------
Application classes 25419
Hosted only classes 640
System methods 202701
System fields 61007

With this fix:
----------------
Application classes 25419
Hosted only classes 640
System methods 16698
System fields 5925

Notice the drastic difference in numbers of methods and fields. I've run a bunch of Quarkus native image application tests with this change and haven't noticed any issues/regressions in the actual functionality.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ImageClassLoader.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ImageClassLoader.java
@@ -257,10 +257,31 @@ public final class ImageClassLoader {
         }
     }
 
-    private static boolean annotationsAvailable(AnnotatedElement element) {
+    /**
+     * @param element The element to check
+     * @return Returns true if the annotations on the {@code element} can be loaded without any
+     *         errors.
+     */
+    private static boolean canLoadAnnotations(AnnotatedElement element) {
         try {
             element.getAnnotations();
             return true;
+        } catch (Throwable t) {
+            handleClassLoadingError(t);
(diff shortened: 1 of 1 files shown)


jaikiran:
Hello @dougxc, @christianwimmer, @cstancu, just a gentle ping to see if this can be reviewed. I'll be happy to do any changes that you think are relevant here. I was hoping this would make it into 20.1.0 release (which is in April?).

dougxc:
> which is in April?

No, May: https://www.graalvm.org/docs/release-notes/version-roadmap/

jaikiran:
> No, May: https://www.graalvm.org/docs/release-notes/version-roadmap/

Thanks. There's enough time then.

PR #2294, opened 2020-03-26
Capture NoClassDefFoundError and report with source type

galderz:
Some NCDE errors can appear when trying to use a graalvm built with OpenJDK 11 to build native images.   

This is due to OpenJDK 11 using JVMCI code that relies on Class.getDeclaredMethods as opposed to using JDK 13+ JVMCI code present in Labs JDK.

The code below catches those NCDE and reports them in a way that we can see which type is the one that contains the import that's causing the issues. 

For more details see https://github.com/quarkusio/quarkus/issues/8081

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
@@ -37,10 +37,12 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 
+import com.oracle.svm.hosted.NativeImageOptions;
 import org.graalvm.collections.Pair;
 import org.graalvm.compiler.api.replacements.SnippetReflectionProvider;
 import org.graalvm.compiler.debug.DebugHandlersFactory;
 import org.graalvm.compiler.graph.Node;
+import org.graalvm.compiler.debug.GraalError;
 import org.graalvm.compiler.options.Option;
 import org.graalvm.compiler.options.OptionType;
 import org.graalvm.compiler.options.OptionValues;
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
```diff
@@ -368,17 +369,22 @@ static boolean declaresDefaultMethods(ResolvedJavaType type) {
...
+             */
+            for (ResolvedJavaMethod method : Inflation.toWrappedType(type).getDeclaredMethods()) {
+                if (method.isDefault()) {
+                    assert !Modifier.isStatic(method.getModifiers()) : "Default method that is static?";
+                    return true;
+                }
             }
+            return false;
+        } catch (NoClassDefFoundError e) {
+            throw new GraalError(e, "Unable to resolve all classes in class: %s", type.toClassName());
```
vjovanov, line 386: Is there an action we could suggest to the user that gets this error?

Also, should we re-throw if `AllowIncompleteClasspath` is enabled? Maybe we should just ignore in that case and return false?

galderz:
@vjovanov Apologies for the delay. Addressed your concerns and updated the PR.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
cstancu: @vjovanov was the intention here to return `false` when `AllowIncompleteClasspath` is *enabled*? This got merged with the condition flipped: 
```   
if (!NativeImageOptions.AllowIncompleteClasspath.getValue()) {
       return false;
}
```
Either way  returning `false` (assume that the interface does not have `default` methods) doesn't seem like a good idea. What are the implications of this on the class initialization logic that depends on it?

galderz: @cstancu I was just looking at this and noticed the same thing. Indeed it seems the condition is flipped.

@vjovanov @cstancu Shall we need the check for `AllowIncompleteClasspath` assuming it's fixed to be:

```
if (NativeImageOptions.AllowIncompleteClasspath.getValue()) {
       return false;
}
```

galderz: I've created https://github.com/oracle/graal/issues/2556 to fix this.

PR #2296, opened 2020-03-27
fix optionality of EpsilonGC and ZGC flag values

dougxc:


ilopmar:
I've tried this fix in our Micronaut-GraalVM pipeline that was failing and with this it works again: https://gitlab.com/micronaut-projects/micronaut-graal-tests/-/jobs/488564687

PR #2302, opened 2020-03-31
Delay memory barrier insertion for unsafe volatile accesses

rwestrel:
Follow up to #1642 for unsafe volatile accesses


compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/extended/RawVolatileLoadNode.java
```diff
@@ -0,0 +1,56 @@
...
+import org.graalvm.word.LocationIdentity;
+
+import static org.graalvm.compiler.nodeinfo.NodeCycles.CYCLES_2;
+import static org.graalvm.compiler.nodeinfo.NodeSize.SIZE_1;
+
+@NodeInfo(cycles = CYCLES_2, size = SIZE_1)
+public class RawVolatileLoadNode extends RawLoadNode implements SingleMemoryKill {
+    public static final NodeClass<RawVolatileLoadNode> TYPE = NodeClass.create(RawVolatileLoadNode.class);
+
+    public RawVolatileLoadNode(ValueNode object, ValueNode offset, JavaKind accessKind, LocationIdentity locationIdentity) {
```
davleopo, line 43: Minor: This needs to propagate the nodeclass (TYPE) to the super class ctor

davleopo:
Hi @rwestrel thanks for the contribution. I added a minor comment that should be fixed before integration.

More general question: whats the rationale for a dedicated new class instead of adding a field to the rawloadnode?

rwestrel:
Thanks for reviewing this.

> Hi @rwestrel thanks for the contribution. I added a minor comment that should be fixed before integration.

I pushed a new change with this fixed.
 
> More general question: whats the rationale for a dedicated new class instead of adding a field to the rawloadnode?

RawVolatileLoadNode implements SingleMemoryKill. RawLoadNode does not and it felt cleaner to have a subclass than to have RawLoadNode implements SingleMemoryKill when regular loads don't affect memory. Also AFAIU, we would then need a LocationIdentity.NO_LOCATION to be returned by getKilledLocationIdentity() for regular loads and make sure that new constant is correctly handled elsewhere.

davleopo:
> RawVolatileLoadNode implements SingleMemoryKill. RawLoadNode does not and it felt cleaner to have a subclass than to have RawLoadNode implements SingleMemoryKill when regular loads don't affect memory. Also AFAIU, we would then need a LocationIdentity.NO_LOCATION to be returned by getKilledLocationIdentity() for regular loads and make sure that new constant is correctly handled elsewhere.

Makes sense, I did not think about the memorykill there. No you are right, having a killing operation that semantically is not even a kill and then handle NO_LOCATION everywhere else sounds wrong. 
I will do integration and internal testing then once the travis tasks passed.

rwestrel:
FYI, I pushed a small fix that caused duplicated memory barriers and also fixed the test case that should have caught that.

davleopo:
ok, got it.
I started internal testing, seems there are some failures but only minor stuff what I can see so far. If nothing serious shows up I will do the internal fixes tomorrow,. If something serious comes up I will get back with follow up discussions here.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/extended/RawStoreNode.java
```diff
@@ -58,21 +58,28 @@
...
+    private boolean isVolatile;
 
     public RawStoreNode(ValueNode object, ValueNode offset, ValueNode value, JavaKind accessKind, LocationIdentity locationIdentity) {
-        this(object, offset, value, accessKind, locationIdentity, true);
+        this(object, offset, value, accessKind, locationIdentity, true, false, null, false);
     }
 
-    public RawStoreNode(ValueNode object, ValueNode offset, ValueNode value, JavaKind accessKind, LocationIdentity locationIdentity, boolean needsBarrier) {
-        this(object, offset, value, accessKind, locationIdentity, needsBarrier, null, false);
+    public RawStoreNode(ValueNode object, ValueNode offset, ValueNode value, JavaKind accessKind, LocationIdentity locationIdentity, boolean isVolatile) {
```
davleopo, line 67: There slipped a tiny bug through: A few lines down is a node intrinsic that used this ctor with needsbarrier and this commit changed it to isVolatile, this caused some of our internal tests to fail. I fixed that internally and started the integration process.

rwestrel: Thanks for fixing that one. And thanks for the quick turnaround on that one.

PR #2334, opened 2020-04-10
Not initialize class in Agent

ziyilin:
Initializing class in Agent when a java.lang.Class.forName is intercepted may lead to missing cathes of reflection, jni, proxy and resources invoked inside target class' static initializer.

This PR can fix the issue https://github.com/oracle/graal/issues/2335.

### DIFF
--- a/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
+++ b/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
@@ -180,15 +180,11 @@ final class BreakpointInterceptor {
         boolean allowed = (accessVerifier == null || accessVerifier.verifyForName(jni, callerClass, className));
         Object result = false;
         if (allowed) {
-            boolean initializeValid = true;
             boolean classLoaderValid = true;
-            CIntPointer initializePtr = StackValue.get(CIntPointer.class);
             WordPointer classLoaderPtr = StackValue.get(WordPointer.class);
             if (bp.method == handles().javaLangClassForName3) {
-                initializeValid = (jvmtiFunctions().GetLocalInt().invoke(jvmtiEnv(), nullHandle(), 0, 1, initializePtr) == JvmtiError.JVMTI_ERROR_NONE);
                 classLoaderValid = (jvmtiFunctions().GetLocalObject().invoke(jvmtiEnv(), nullHandle(), 0, 2, classLoaderPtr) == JvmtiError.JVMTI_ERROR_NONE);
             } else {
-                initializePtr.write(1);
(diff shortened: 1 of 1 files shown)


peter-hofer:
Thank you for your submission @ziyilin! I think in this case, we don't need `initializeValid` at all and can declare and set `initializePtr` right before the call. Do you mind if I make those changes or would you prefer to do that?

peter-hofer:
One drawback is of course that we will no longer observe when a class fails to initialize. Have you experimented with temporarily disabling `recursive` for the `forName` call too?

peter-hofer:
> One drawback is of course that we will no longer observe when a class fails to initialize. Have you experimented with temporarily disabling `recursive` for the `forName` call too?

... although that needs a more sophisticated mechanism to ensure that the `forName` call is not intercepted, but everything else is.

ziyilin:
> Thank you for your submission @ziyilin! I think in this case, we don't need initializeValid at all and can declare and set initializePtr right before the call. Do you mind if I make those changes or would you prefer to do that?

I have modified the submit accordingly.

> One drawback is of course that we will no longer observe when a class fails to initialize. Have you experimented with temporarily disabling recursive for the forName call too?

I don't think it's a drawback. Suppose the target class exists but the initialization fails, the agent will not write the trace into the configuration file. Without the reflection configration information, the native image will not run into the same failure but report a class not found exception at runtime. These two different behaviors may confuse the user.  Wouldn't it be better if user can see the same exceptions?

peter-hofer:
> I have modified the submit accordingly.

Thanks!

> > One drawback is of course that we will no longer observe when a class fails to initialize. Have you experimented with temporarily disabling recursive for the forName call too?
> 
> I don't think it's a drawback. Suppose the target class exists but the initialization fails, the agent will not write the trace into the configuration file. Without the reflection configration information, the native image will not run into the same failure but report a class not found exception at runtime. These two different behaviors may confuse the user. Wouldn't it be better if user can see the same exceptions?

In general, I agree. What I am concerned about is code is probing if some class can be loaded (and initialized) and it's not a problem if not, which could be the case with some kind of plug-in mechanism. If it's the class initialization that fails, we would miss that and introduce a type into the configuration that breaks the image build. But then at least it should fail in an obvious way.

peter-hofer:
> What I am concerned about is code is probing if some class can be loaded (and initialized) and it's not a problem if not, which could be the case with some kind of plug-in mechanism. If it's the class initialization that fails, we would miss that and introduce a type into the configuration that breaks the image build. But then at least it should fail in an obvious way.

We've discussed this internally and this situation will likely still result in the class being initialized at runtime so the code can deal with it the same way then. I'll be merging your change together with other agent changes soon, thanks again for your submission!

PR #2343, opened 2020-04-14
Initialize classes with simple side-effect-free class initializers before static analysis

christianwimmer:


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ameta/AnalysisConstantReflectionProvider.java
```diff
@@ -225,22 +225,14 @@ private JavaConstant replaceObject(JavaConstant value) {
...
      * all (because it is the hosted assertion status), we instead return the appropriate runtime
-     * assertion status.
+     * assertion status. Field loads are also intrinsified early in
+     * {@link com.oracle.svm.hosted.phases.EarlyConstantFoldLoadFieldPlugin}, but we could still see
+     * such a field here if user code, e.g., accesses it via reflection.
      */
     private static JavaConstant interceptAssertionStatus(AnalysisField field, JavaConstant value) {
-        if (Modifier.isStatic(field.getModifiers()) && field.isSynthetic() && field.getName().startsWith("$assertionsDisabled")) {
-            String unsubstitutedName = field.wrapped.getDeclaringClass().toJavaName();
-            if (unsubstitutedName.startsWith("java.") || unsubstitutedName.startsWith("javax.") || unsubstitutedName.startsWith("sun.")) {
```
christianwimmer, line 233: This is an intentional semantic change that cleans up the assertion status for JDK classes. Historically, some assertions in the JDK failed because not all parts of the JDK were supported yet in a native image. But by now every assertion failing in JDK code is actually a bug that we want to know about and fix.

PR #2354, opened 2020-04-16
Add LLVM Round Instrinsic

kipply:
llvm spec; https://llvm.org/docs/LangRef.html#llvm-round-intrinsic

Referenced code from llvm ceil/floor. I found this missing as it was used by a Ruby library, [oj](https://github.com/ohler55/oj). 

https://github.com/Shopify/graal/issues/1

### DIFF
--- a/sulong/projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/BasicNodeFactory.java
+++ b/sulong/projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/BasicNodeFactory.java
@@ -1362,6 +1362,8 @@ public class BasicNodeFactory implements NodeFactory {
                     return LLVMPowNodeGen.create(args[1], args[2]);
                 case "llvm.powi.f80":
                     return LLVMPowNodeGen.create(args[1], args[2]);
+                case "llvm.round.f80":
+                    return LLVMCMathsIntrinsicsFactory.LLVMRoundNodeGen.create(args[1]);
                 case "llvm.fabs.f32":
                 case "llvm.fabs.f64":
                 case "llvm.fabs.f80":
--- a/sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/nodes/intrinsics/c/LLVMCMathsIntrinsics.java
+++ b/sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/nodes/intrinsics/c/LLVMCMathsIntrinsics.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2019, Oracle and/or its affiliates.
+ * Copyright (c) 2016, 2020, Oracle and/or its affiliates.
  *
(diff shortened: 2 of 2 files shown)


sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/nodes/intrinsics/c/LLVMCMathsIntrinsics.java
```diff
@@ -248,6 +248,26 @@ protected LLVM80BitFloat doIntrinsic(LLVM80BitFloat value) {
...
+        }
+
+        @Specialization
+        protected double doIntrinsic(double value) {
+            return Math.round(value);
+        }
+
+        @Specialization
+        protected LLVM80BitFloat doIntrinsic(LLVM80BitFloat value) {
+            double result = doIntrinsic(value.getDoubleValue());
```
kipply, line 266: I believe this is imprecise

rschatz: Yes, that might lose precision, but in practice it's rarely a problem, since we're rounding anyway. This is only a problem for numbers that are big enough that the precision of `double` is not enough to represent a difference of `1.0` (i.e. numbers bigger than `2^52`).

Is that a problem for your use-case? Then we'd need to do a native call, similar to what we do for regular arithmetic. Unfortunately we don't have any way to directly do 80 bit floating point arithmetic in Java yet.

kipply: Yep sounds right. Probably not a problem for our use cases! Thanks~

sulong/projects/com.oracle.truffle.llvm.parser.factories/src/com/oracle/truffle/llvm/parser/factories/BasicNodeFactory.java
```diff
@@ -1362,6 +1362,8 @@ protected LLVMExpressionNode getLLVMBuiltin(FunctionDeclaration declaration, LLV
                     return LLVMPowNodeGen.create(args[1], args[2]);
                 case "llvm.powi.f80":
                     return LLVMPowNodeGen.create(args[1], args[2]);
+                case "llvm.round.f80":
```
rschatz, line 1365: I guess we can also put `llvm.round.f32` and `llvm.round.f64` here, right?

kipply: Yep, let me know if you want me to commit that or if you want to add it when you merge

kipply:
@rschatz 
The closest my trace got me was to this function in a library; https://github.com/ohler55/oj/blob/develop/ext/oj/strict.c#L103 
might be something buried in https://github.com/ohler55/oj/blob/develop/ext/oj/parse.c#L754?

PR #2360, opened 2020-04-18
Support WithExceptionNode in IntrinsicGraphBuilder

christianwimmer:
Fixes a problem introduced by the recent addition of ArrayCopyWithExceptionNode. When System.arraycopy is compiled as a root method, the IntrinsicGraphBuilder needs to properly link the control flow and also create the exception unwind nodes.

### DIFF
--- a/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/IntrinsicGraphBuilder.java
+++ b/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/IntrinsicGraphBuilder.java
@@ -35,6 +35,7 @@ import org.graalvm.compiler.debug.DebugCloseable;
 import org.graalvm.compiler.debug.DebugContext;
 import org.graalvm.compiler.debug.GraalError;
 import org.graalvm.compiler.graph.NodeSourcePosition;
+import org.graalvm.compiler.nodes.AbstractBeginNode;
 import org.graalvm.compiler.nodes.CallTargetNode;
 import org.graalvm.compiler.nodes.CallTargetNode.InvokeKind;
 import org.graalvm.compiler.nodes.FixedNode;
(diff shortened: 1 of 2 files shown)


compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/IntrinsicGraphBuilder.java
```diff
@@ -134,12 +146,46 @@ protected IntrinsicGraphBuilder(OptionValues options, DebugContext debug, CorePr
...
+                WithExceptionNode withExceptionNode = (WithExceptionNode) fixedNode;
+                AbstractBeginNode normalSuccessor;
+                if (withExceptionNode instanceof SingleMemoryKill) {
+                    normalSuccessor = graph.add(KillingBeginNode.create(((SingleMemoryKill) withExceptionNode).getKilledLocationIdentity()));
+                } else if (withExceptionNode instanceof MultiMemoryKill) {
+                    normalSuccessor = graph.add(MultiKillingBeginNode.create(((MultiMemoryKill) withExceptionNode).getKilledLocationIdentities()));
+                } else {
+                    assert !(withExceptionNode instanceof MemoryKill) : "missing a MemoryKill";
+                    normalSuccessor = graph.add(new BeginNode());
+                }
```
gergo-, line 160: This should just be:
```suggestion
                AbstractBeginNode normalSuccessor = withExceptionNode.createNextBegin();
```
Looks good to me otherwise.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/IntrinsicGraphBuilder.java
```diff
@@ -134,12 +146,46 @@ protected IntrinsicGraphBuilder(OptionValues options, DebugContext debug, CorePr
...
             }
         }
     }
 
+    /**
+     * Currently unimplemented here, but implemented in subclasses that need it.
+     *
+     * @param exceptionObject The node that needs an exception state.
+     */
+    protected void setExeptionState(ExceptionObjectNode exceptionObject) {
```
gergo-, line 185: typo: `Exeption`

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/IntrinsicGraphBuilder.java
christianwimmer: Thanks, didn't know that method exists. But I'm glad it does.

PR #12, opened 2016-01-25
Interop: annotation processor to generate classes for message resolution

grimmerm:
There is a need to simplify interop API to:
- make it more type-safe
- encourage exception handling
- make its use less verbose

PR #2364, opened 2020-04-20
Apply trapping nullcheck to the uncompressed target for compressed po…

XiaohongGong:
…inter.

Trapping nullcheck might generate two uncompress instructions for the same compressed oop in Graal. One is inserted by the backend when it emits nullcheck. If the pointer is a compressed object, it should be uncompressed before the nullcheck is emitted. And another one is generated by the normal uncompressing operation. These two instructions are duplicated with each other.
The generated codes on AArch64 like:

      ldr   w0, [x0,#112]
      lsl   x2, x0, #3      ; uncompressing (first)
      ldr   xzr, [x2]       ; implicit exception: deoptimizes
      ......                ; fixed operations
      lsl   x0, x0, #3      ; uncompressing (second)
      str   w1, [x0,#12]

A simple way to avoid this is to creat a new uncompression node for the nullcheck, and let the value numbering remove the duplicated one if possible. Since the address lowering of AMD64 can handle the uncompressing computation for address, the created uncompression node is wrapped to an address node and the nullcheck is finally applied on the address.

With the modification, the codes above could be optimized to:

      ldr   w0, [x0,#112]
      lsl   x0, x0, #3      ; uncompressing
      ldr   xzr, [x0]       ; implicit exception: deoptimizes
      ......                ; fixed operations
      str   w1, [x0,#12]

Change-Id: Iabfe47bbf984ed11c42555f84bdd0ccf2a5bdddb

### DIFF
--- a/compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/UseTrappingNullChecksPhase.java
+++ b/compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/UseTrappingNullChecksPhase.java
@@ -274,7 +274,7 @@ public class UseTrappingNullChecksPhase extends BasePhase<LowTierContext> {
 
         if (trappingNullCheck == null) {
             // Need to add a null check node.
-            trappingNullCheck = graph.add(new NullCheckNode(value, deoptReasonAndAction, deoptSpeculation));
+            trappingNullCheck = graph.add(NullCheckNode.create(value, deoptReasonAndAction, deoptSpeculation));
             graph.replaceSplit(ifNode, trappingNullCheck, nonTrappingContinuation);
             debug.log("Inserted NullCheckNode %s", trappingNullCheck);
         }
--- a/compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/AbstractPointerStamp.java
+++ b/compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/AbstractPointerStamp.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -160,5 +160,9 @@ public abstract class AbstractPointerStamp extends Stamp {
     @Override
     public JavaKind getStackKind() {
         return JavaKind.Illegal;
+    }
+
+    public boolean isCompressed() {
+        return false;
     }
 }
--- a/compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/phases/LowTier.java
+++ b/compiler/src/org.graalvm.compiler.core/src/org/graalvm/compiler/core/phases/LowTier.java
@@ -70,9 +70,9 @@ public class LowTier extends BaseTier<LowTierContext> {
         appendPhase(new FixReadsPhase(true,
                         new SchedulePhase(GraalOptions.StressTestEarlyReads.getValue(options) ? SchedulingStrategy.EARLIEST : SchedulingStrategy.LATEST_OUT_OF_LOOPS_IMPLICIT_NULL_CHECKS)));
 
-        appendPhase(canonicalizerWithoutGVN);
-
         appendPhase(new UseTrappingNullChecksPhase());
+
+        appendPhase(canonicalizerWithoutGVN);
 
         appendPhase(new DeadCodeEliminationPhase(Required));
 
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/meta/AddressLoweringHotSpotSuitesProvider.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/meta/AddressLoweringHotSpotSuitesProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2020, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2017, Red Hat Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
@@ -56,11 +56,13 @@ public class AddressLoweringHotSpotSuitesProvider extends HotSpotSuitesProvider 
         Suites suites = super.createSuites(options);
 
         ListIterator<BasePhase<? super LowTierContext>> findPhase = suites.getLowTier().findPhase(UseTrappingNullChecksPhase.class);
-        if (findPhase == null) {
+        if (findPhase != null) {
+            findPhase.add(addressLowering);
+        } else {
             findPhase = suites.getLowTier().findPhase(SchedulePhase.class);
+            findPhase.previous();
+            findPhase.add(addressLowering);
         }
-        findPhase.previous();
-        findPhase.add(addressLowering);
 
         return suites;
     }
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/nodes/type/KlassPointerStamp.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/nodes/type/KlassPointerStamp.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -146,6 +146,7 @@ public final class KlassPointerStamp extends MetaspacePointerStamp {
         }
     }
 
+    @Override
     public boolean isCompressed() {
         return encoding != null;
     }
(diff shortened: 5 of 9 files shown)


XiaohongGong:
Hi there, could anyone please take a look at this PR? Thanks!

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/UseTrappingNullChecksPhase.java
```diff
@@ -287,4 +292,22 @@ private static void replaceWithTrappingNullCheck(AbstractDeoptimizeNode deopt, I
...
     }
+
+    private static NullCheckNode tryUseUncompressedNullCheck(ValueNode value, IsNullNode isNullNode) {
+        assert isNullNode.getValue().equals(value);
+        CompressionNode uncompressed = null;
+        for (Node usage : value.usages()) {
+            if (usage instanceof CompressionNode) {
+                if (((CompressionNode) usage).getOp() == CompressionNode.CompressionOp.Uncompress) {
+                    assert ((CompressionNode) usage).getValue().equals(value);
+                    uncompressed = (CompressionNode) usage;
```
dougxc, line 303: Couldn't there be multiple uncompress usages?

XiaohongGong: I searched the codes that the `uncompress` is mostly generated together with its `input`. And If there are multiple uncompress usages, would they be optimized to a single one by GVN? If it exists the corner cases, I think we'd better add the restriction of the single usage of uncompress.

dougxc: Yes, please add the restriction and only apply the transformation if there is a single uncompress usage.

XiaohongGong: OK! Thank you for review! I will update soon!

XiaohongGong: Hi @dougxc I'v updated the patch according to your comment. So could you please take a look at it again? Thanks!

dougxc: Looks good. I've started the process of integration.

XiaohongGong: Thanks again for your review!

tkrodriguez:
Wouldn't it make more sense to change NullCheckNode to take an AddressNode so that the normal address computation optimizations take care of this?  It just seems like you're hacking value numbering into the UseTrappingNullChecksPhase.  Any path which builds an address in the LIR is going to have this kind of problem.

XiaohongGong:
> Any path which builds an address in the LIR is going to have this kind of problem

Actually, I agree with you that it's better to apply this transformation to `NullCheckNode`. Thanks!

> Wouldn't it make more sense to change NullCheckNode to take an AddressNode so that the normal address computation optimizations take care of this?

I don't quite understand about this. Do you mean we could wrap the uncompressed object to an `AddressNode` by setting its base to the uncompressed object and offset to 0? Or wrap the narrowed object to an `AddressNode` ? I didn't find the related computation of `AddressNode ` that could take care of this issue. If it actually exists, could you please give me more details? Thanks!

Another concern about `AddressNode` is that it has the relationship with the reference map that is used for GC. And I'm not sure whether any pointer can be converted to an AddressNode without any regression. So how do you think about this? Thanks!

tkrodriguez:
My thought was that a NullCheckNode would just be created with OffsetAddressNode with a 0 displacement then the existing address optimizations would take care of commoning the computation where possible.  It seems like there's a phase ordering problem though because the AddressLoweringPhase runs just before UseTrappingNullChecksPhase so we can't just build an OffsetAddressNode any more.  The only way around that I can imagine is for AddressNode to have a method that returns an AddressNode suitable for use by NullCheckNode so each platform dependent AddressNode would return its own subclass.  That all seems overly complicated though.
Why can't this code simply create an uncompress operation and let value numbering just merge them if it's possible?  The direct search and substitutions just doesn't seem right to me.

XiaohongGong:
> My thought was that a NullCheckNode would just be created with OffsetAddressNode with a 0 displacement then the existing address optimizations would take care of commoning the computation where possible.

I checked the optimizations codes for `AArch64AddressNode`. Currently it doesn't have any significant transformations for `BASE_REGISTER_ONLY` address mode on AArch64. Besides the issue I found is for the duplicated uncompress operations. So I don't think it's so worth to use an `AddressNode` here.  And actually it also has the phase order problem.



> Why can't this code simply create an uncompress operation and let value numbering just merge them if it's possible?

It's a very good idea! But it has the similar issues that will make codes complicated: `CompressionNode` is an abstract class. We have to create a `HotSpotCompressionNode` or a `SubstrateCompressionNode`.  It seems creating an VM dependent node at the common phase is complicated.




> The direct search and substitutions just doesn't seem right to me.

Originally the `IsNull` is applied on the uncompressed object. It is canonicalized inside `IsNullNode` which converts the null check to the `input` of the `CompressionNode`.  So do you think it's better to stop this conversion for `CompressionNode`?
The related codes:

    if (value instanceof ConvertNode) {
        ConvertNode convertNode = (ConvertNode) value;
        if (convertNode.mayNullCheckSkipConversion()) {
              value = convertNode.getValue();
              continue;
        }
        /*
         * CompressionNode.mayNullCheckSkipConversion returns false when linear pointer
         * compression is enabled.
         */
         if (value instanceof CompressionNode) {
              CompressionNode compressionNode = (CompressionNode) value;
              nullConstant = compressionNode.nullConstant();
              value = compressionNode.getValue();
              continue;
          }
    }

XiaohongGong:
> > Why can't this code simply create an uncompress operation and let value numbering just merge them if it's possible?
> 
> It's a very good idea! But it has the similar issues that will make codes complicated: `CompressionNode` is an abstract class. We have to create a `HotSpotCompressionNode` or a `SubstrateCompressionNode`. It seems creating an VM dependent node at the common phase is complicated.

A simple way to create the uncompress operation is directly cloning one from the found `CompressionNode`.  It looks weird, but we can let value numbering to do the merging safely.

tkrodriguez:
Ok, I think I see what's happening.  For AMD64 we'd prefer all memory operations to have the shift visible in the address so that we can fold it into the addressing modes.  I'd guess that's the thinking behind the mayNullCheckSkipConversion transformation.  So I don't think we want to undo that.  Since UseTrappingNullChecks runs after address lowering that CompressionNode should have already been swallowed on platforms where that's possible.  So on AMD64 if there were no CompressionNodes of that value left then your original change would leave it alone, but if there happened to be unrelated ones it would merge them together, which I don't think we want to do.
I guess I'm left with the idea that AddressNode should have a helper method that returns an address suitable for a NullCheckNode.  I don't love that change either but it seems like AMD64 and aarch64 want this code to behave differently.

XiaohongGong:
> So on AMD64 if there were no CompressionNodes of that value left then your original change would leave it alone, but if there happened to be unrelated ones it would merge them together, which I don't think we want to do.

So do you think cloning a new `CompressionNode` from the existed one could resolve this issue?  Like this change:

    diff --git a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/nodes/HotSpotCompressionNode.java b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/nodes/HotSpotCompressionNode.java
    index f3bdf2f52a..0e7beffebd 100644
    --- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/nodes/HotSpotCompressionNode.java
    +++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/nodes/HotSpotCompressionNode.java
    @@ -79,6 +79,11 @@ public final class HotSpotCompressionNode extends CompressionNode {
         }
     }

    +    @Override
    +    public CompressionNode clone() {
    +        return new HotSpotCompressionNode(this.op, this.value, this.encoding);
    +    }
    +




> I guess I'm left with the idea that AddressNode should have a helper method that returns an address suitable for a NullCheckNode. I don't love that change either but it seems like AMD64 and aarch64 want this code to behave differently.

Do you mean we'd better do something for AddressNode to fix this issue? So do you know the main reason that `UseTrappingNullChecksPhase` should run after address lowering phase?

tkrodriguez:
I think cloning is wrong for AMD64 because you can get an extra shift if that path instead of using the addressing modes directly.
It's not obvious to me that there's any technical reason that UseTrappingNullChecksPhase runs after address lowering.  I think you could try reordering them and using an OffsetAddressNode for the NullCheckNode.  That would be nicest if it works.

XiaohongGong:
> I think cloning is wrong for AMD64 because you can get an extra shift if that path instead of using the addressing modes directly.

Yes, I agree with you.  Since the base of the `OffsetAddressNode` should be the uncompressed object, I mean it's unavoidable to create a new uncompress operation for it. Or the JVM will crash at least on AArch64.  We can create the uncompress operation in `UseTrappingNullChecksPhase` by cloning one from an existed one. Or we can also try to do the uncompressing in `AddressLowingPhase` for each back-end. However, this needs to make the codes more complex.

XiaohongGong:
> It's not obvious to me that there's any technical reason that UseTrappingNullChecksPhase runs after address lowering. I think you could try reordering them and using an OffsetAddressNode for the NullCheckNode. That would be nicest if it works.

I think it's ok to reorder the two phases. I tried it locally and our internal tests pass without any failure.

XiaohongGong:
Hi @tkrodriguez ,  I have updated my patch to make the `NullCheckNode`  take an `OffsetAddressNode` as the input if the original object is compressed. I also reordered `UseTrappingNullChecksPhase` and `AddressLoweringPhase` to make it work well. Besides I noticed that `AMD64AddressNode` is `Simplifiable ` which applies `canonicalizeIndex` in the `simplify` function. 
I think there should run a `CanonicalizerPhase`  after `AddressLoweringPhase`. So I also move the last `CanonicalizerPhase` after it. Could you please to take a look at it? Thank you!

XiaohongGong:
Hi @tkrodriguez  How is this PR going? Any update about it? Thanks so much!

XiaohongGong:
Hi @tkrodriguez I rebased this PR and resolved the conflict. So how is this PR going now. Any comments, please let me know. Thanks so much!

PR #2393, opened 2020-04-24
Provide java.lang.Terminator.setup() via --install-exit-handlers

olpaw:
This PR provides a fix for #1592

Testing with

```java
public class Main {
        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
            System.out.println("Shutdown happens!");
        }));
        
        try {
            System.out.println("Zzzzzzz.....");
            Thread.sleep(Long.MAX_VALUE);
        } catch (InterruptedException e) {
            System.out.println("Good morning!");
            e.printStackTrace();
        }
    }
}
```

shows an image size increase of 48Kb when native-image --install-exit-handlers is used (from 6,802,048 bytes to 6,850,976 bytes)

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
@@ -51,9 +51,7 @@ import org.graalvm.word.PointerBase;
 import org.graalvm.word.UnsignedWord;
 import org.graalvm.word.WordFactory;
 
-import com.oracle.svm.core.annotate.Alias;
 import com.oracle.svm.core.annotate.AlwaysInline;
-import com.oracle.svm.core.annotate.TargetClass;
 import com.oracle.svm.core.c.CGlobalData;
 import com.oracle.svm.core.c.CGlobalDataFactory;
 import com.oracle.svm.core.c.function.CEntryPointOptions;
@@ -135,10 +133,6 @@ public class JavaMainWrapper {
         int exitCode;
         try {
             if (SubstrateOptions.ParseRuntimeOptions.getValue()) {
-                if (SubstrateOptions.InstallExitHandlers.getValue()) {
-                    Target_java_lang_Terminator.setup();
-                }
-
                 /*
                  * When options are not parsed yet, it is also too early to run the startup hooks
                  * because they often depend on option values. The user is expected to manually run
@@ -292,10 +286,4 @@ public class JavaMainWrapper {
         @RawField
         void setArgv(CCharPointerPointer value);
     }
-}
-
-@TargetClass(className = "java.lang.Terminator")
-final class Target_java_lang_Terminator {
-    @Alias
-    static native void setup();
 }
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
@@ -51,12 +51,16 @@ import org.graalvm.word.PointerBase;
 import org.graalvm.word.UnsignedWord;
 import org.graalvm.word.WordFactory;
 
+import com.oracle.svm.core.annotate.Alias;
 import com.oracle.svm.core.annotate.AlwaysInline;
+import com.oracle.svm.core.annotate.RecomputeFieldValue;
+import com.oracle.svm.core.annotate.TargetClass;
 import com.oracle.svm.core.c.CGlobalData;
 import com.oracle.svm.core.c.CGlobalDataFactory;
 import com.oracle.svm.core.c.function.CEntryPointOptions;
 import com.oracle.svm.core.c.function.CEntryPointSetup.EnterCreateIsolatePrologue;
 import com.oracle.svm.core.jdk.InternalVMMethod;
+import com.oracle.svm.core.jdk.Package_jdk_internal_misc;
 import com.oracle.svm.core.jdk.RuntimeSupport;
 import com.oracle.svm.core.option.RuntimeOptionParser;
 import com.oracle.svm.core.thread.JavaThreads;
@@ -133,6 +137,10 @@ public class JavaMainWrapper {
         int exitCode;
         try {
             if (SubstrateOptions.ParseRuntimeOptions.getValue()) {
+                if (SubstrateOptions.InstallExitHandlers.getValue()) {
+                    Target_java_lang_Terminator.setup();
+                }
+
                 /*
                  * When options are not parsed yet, it is also too early to run the startup hooks
                  * because they often depend on option values. The user is expected to manually run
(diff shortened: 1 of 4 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
```diff
@@ -133,6 +135,12 @@ public static int runCore(int paramArgc, CCharPointerPointer paramArgv) {
         int exitCode;
         try {
             if (SubstrateOptions.ParseRuntimeOptions.getValue()) {
+                if (SubstrateUtil.isPosix()) {
```
christianwimmer, line 138: Why does that not work the same way on Windows too?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
```diff
@@ -133,6 +135,12 @@ public static int runCore(int paramArgc, CCharPointerPointer paramArgv) {
...
+                    if (SubstrateOptions.InstallExitHandlers.getValue()) {
```
christianwimmer, line 139: but the check for || (SubstrateUtil.isLinux() && ProcessProperties.getProcessId() ==1) is missing here? Otherwise how do we get out-of-the-box correct handling for the Docker init process?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -425,6 +425,11 @@ public static boolean useLLVMBackend() {
     @Option(help = "Show native-toolchain information and image-build settings", type = User)//
     public static final HostedOptionKey<Boolean> DumpTargetInfo = new HostedOptionKey<>(false);
 
+    @Platforms({Platform.LINUX.class, Platform.DARWIN.class})
```
christianwimmer, line 428: Having a platform specific API option complicates all sorts of build scripts, so I think we should avoid that in general.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
olpaw: I decided against making the installation of the handlers dependent on  `ProcessProperties.getProcessId() == 1`.

I tested what the JVM does when running in docker and without. Regarding signal handling the JVM makes no distinction whatsoever. So to be as close as possible to the behaviour of the JVM I do not predicate on PID1. Also since we require the user to opt-in via `--install-exit-handlers` there should be no regressions for anyone not interested in having this JVM-like behaviour for SIGHUP/SIGINT/SIGTERM exit handling.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
olpaw: I will have a look what the JVM does on Windows and make --install-exit-handlers behave the same. Then this is no more posix-only.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
olpaw: I agree. Also providing proper behaviour for Windows will make this use of `@Platforms` obsolete. See https://github.com/oracle/graal/pull/2393#discussion_r415622948

olpaw:
@pejovica reminded me that there is `java.lang.Terminator`. That is exactly what we should use here.

olpaw:
Arrg using the code from `java.lang.Terminator.setup()` does not work. Investigating ...

pejovica:
You need to reset `java.lang.Terminator.handler` field, or you could reuse `com.oracle.svm.enterprise.pgo.substitutions.Target_java_lang_Terminator` isntead.

olpaw:
Tested if this also works on Windows
```
C:\Users\pwoegere\OLabs\graal\substratevm>main.exe
Zzzzzzz.....
[CTRL-C]
Shutdown happens!
```

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
```diff
@@ -287,3 +295,16 @@ public static void enter(int paramArgc, CCharPointerPointer paramArgv) {
         void setArgv(CCharPointerPointer value);
     }
 }
+
+@TargetClass(className = "java.lang.Terminator")
+final class Target_java_lang_Terminator {
+    @Alias @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.Reset) //
+    private static Target_jdk_internal_misc_Signal_Handler handler;
```
pejovica, line 302: For this to work on JDK 11 (`handler` field changed type) you could declare it as:
```suggestion
    private static Object handler;
```

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
```diff
@@ -133,6 +137,10 @@ public static int runCore(int paramArgc, CCharPointerPointer paramArgv) {
         int exitCode;
         try {
             if (SubstrateOptions.ParseRuntimeOptions.getValue()) {
+                if (SubstrateOptions.InstallExitHandlers.getValue()) {
+                    Target_java_lang_Terminator.setup();
+                }
```
pejovica, line 142: Maybe you should register a startup hook instead.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
olpaw: Nahh, that doesn't work.
```
Error: Type mismatch:
    class java.lang.Object private static java.lang.Object com.oracle.svm.core.Target_java_lang_Terminator.handler
    interface jdk.internal.misc.Signal$Handler private static jdk.internal.misc.Signal$Handler java.lang.Terminator.handler
```
Instead it's much easier to use: https://github.com/oracle/graal/pull/2393/commits/6bebeb45724956f33357ff14caf63fec5757f0ce

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
olpaw: I agree: ec15571605e2cf13390133e908dea82b7884654e

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java
christianwimmer: OK. Having a different automatic behavior when running as Docker pid 1 is probably a bad idea.

PR #2394, opened 2020-04-24
Add `vm` suite command to fetch required labsjdk

lazar-mitrovic:
Adds `mx fetch-labsjdk` command to `graal/vm` suite which downloads JDK required for GraalVM building and outputs corresponding `JAVA_HOME` export command to `stdout`.


dougxc:
I like this idea but I think it should be a general mx command instead of something specific to the `vm` suite. I envisage something like this:
```
> mx fetch-jdk --from common.json --to /usr/lib/jvm --shell bash
[0] openjdk8-8u242-jvmci-20.1-b01
[1] labsjdk-ce-11.0.6+9-jvmci-20.1-b01
Select JDK> 0
Install https://github.com/graalvm/openjdk8-jvmci-builder/releases/download/jvmci-20.1-b01/openjdk-8u242-jvmci-20.1-b01-linux-amd64.tar.gz into /usr/lib/jvm/openjdk1.8.0_242 [Yn]? y
Downloading...
Installing...
Run the following to set JAVA_HOME in your shell:
export JAVA_HOME=/usr/lib/jvm/
```

If `--from` is omitted, `common.json` is searched for in the current and parent directories.
If `--to` is omitted, it can be the [default system JDK location](https://github.com/graalvm/mx/blob/28e6a5c6b77df50886d83fc79ec09a689a13d348/select_jdk.py#L53).
If `--shell` is omitted, it can be taken from `$SHELL`.

Note that only JDKs with known public URLs should be listed.

vm/mx.vm/mx_vm.py
```diff
@@ -186,3 +187,76 @@ def rm_links(self):
...
+        untar.extract(jdk_path)
+        os.remove(join(jdk_path, labs_archive))
+        os.rename(join(jdk_path, curr_jdk_folder), jdk_folder)
+
+    if mx.get_os() == 'darwin':
+        jdk_folder = join(jdk_folder, 'Contents', 'Home')
+
+    print('export JAVA_HOME={}'.format(jdk_folder))
+
+mx.update_commands(_suite, {
```
vjovanov, line 260: This can become an annotation on the function (`@mx.command`).

vm/mx.vm/mx_vm.py
```diff
@@ -186,3 +187,76 @@ def rm_links(self):
 
     def __str__(self):
         return "Generating GraalVM symlinks in the vm suite"
+
+
+def fetch_labsjdk(args):
```
vjovanov, line 192: This function needs to have documentation like other `mx` commands.

vm/mx.vm/mx_vm.py
```diff
@@ -186,3 +187,76 @@ def rm_links(self):
...
+    parser = ArgumentParser(prog='mx fetch-labsjdk')
+    parser.add_argument('--java-version', action='store', help='numeric value representing requested JDK version')
+    parser.add_argument('--java-distribution', action='store', help='ce or ee')
+    parser.add_argument('-q', '--quiet', action='store_true', help='suppress logging output')
```
vjovanov, line 196: I personally need two more optional flags for this feature to be complete: 1) (only on macOS) `--no-contents-home` that will strip the `Contents/Home` nonsense on macOS, and 2) `--dest-name` as typing `labsjdk-ce-11.0.6+9-jvmci-20.1-b01` in `sdkman` is annoying. I just use `11-ce` as a shorthand.

vm/mx.vm/mx_vm.py
```diff
@@ -34,9 +34,10 @@
...
 import mx_vm_gate
 
-import os
-from os.path import join, relpath
 
+import os, json, shutil
+from os.path import join, relpath, exists, abspath
+from argparse import ArgumentParser
 
 _suite = mx.suite('vm')
```
vjovanov, line 42: I would put this in `sdk` as then all projects have access to it.

vjovanov:
We need to use this script in the CI as a primary way to get the JDK. That will assure that it never gets outdated.

vm/mx.vm/mx_vm.py
dougxc: Is the `Contents/Home` prefix that much of a problem if script completely manages the installation of JDKs and setting of `JAVA_HOME`? My concern is that we can run into issues if we make the layout of JDKs not match the default for a platform (and which external tools expect).

vm/mx.vm/mx_vm.py
dougxc: Any good reason not to put it in mx itself?

dougxc:
> We need to use this script in the CI as a primary way to get the JDK. That will assure that it never gets outdated.

By this you mean the Travis CI obviously as the internal CI already uses `common.json`.

vm/mx.vm/mx_vm.py
vjovanov: Not a problem, just a convenience. I use this layout with `sdkman` and all works fine.

lazar-mitrovic:
Moved to `mx` project 
https://github.com/graalvm/mx/pull/214

Issue #1216, opened 2019-04-30
Include package-info into the native image


gsmet:
Hi,

Maybe it's simple but I wasn't able to figure out how to include a `package-info` into a native image.

We have a user trying to use JAXB and the generated `package-info` contains important information necessary for JAXB to work properly, namely:
```
@javax.xml.bind.annotation.XmlSchema(namespace = "http://www.w3.org/2005/Atom", elementFormDefault = javax.xml.bind.annotation.XmlNsForm.QUALIFIED)
package mapper.codegen.eforum;
```

I tried to see if the reflection API would allow me to register it but it doesn't look like it does. I tried with a reflection JSON file but it doesn't seem to change anything either.

I couldn't find any literature on the subject so I thought I might as well ask.

Thanks.

cstancu:
We don't currently support this. Can you provide a simple example? How would you access the `package-info` at runtime?

tcollignon:
Hi @cstancu 

I have provide a simple example here : https://github.com/tcollignon/quarkus-jaxb-test

If I run it with graalvm native image it fail like I described here : https://github.com/quarkusio/quarkus/issues/2199#issuecomment-487703960

Do you need more information? 

Thx you very much

cstancu:
Thank you for your example. Can you show me an example of how JAXB is accessing `package-info` via reflection? The error in https://github.com/quarkusio/quarkus/issues/2199#issuecomment-487703960 `java.lang.ClassCastException: org.jboss.threads.JBossThread cannot be cast to java.lang.Thread` seems to be related to https://github.com/oracle/graal/issues/1200.

tcollignon:
For the question "Can you show me an example of how JAXB is accessing package-info via reflection? ", I'm afraid that I can't help you :( I think it's in quarkus ? @gsmet I'm wrong?

gsmet:
It really doesn't do anything fancy: it gets the package from the class then tries to access the package info.

From `java.lang.Package`:
```java
    private Class<?> getPackageInfo() {
        if (packageInfo == null) {
            try {
                packageInfo = Class.forName(pkgName + ".package-info", false, loader);
            } catch (ClassNotFoundException ex) {
                // store a proxy for the package info that has no annotations
                class PackageInfoProxy {}
                packageInfo = PackageInfoProxy.class;
            }
        }
        return packageInfo;
    }
```

So we would need a way to somehow register the `package-info` for reflection.

cstancu:
This works with `native-image`:
[...]

PR #2400, opened 2020-04-27
Support package-annotation-related calls

zhangruoxu:
In this pull request, four package-annotation-related calls, i.e., `Package.getAnnotation()`, `Package.getAnnotations()`, `Package.getDeclaredAnnotation()`, `Package.getDeclaredAnnotations()`, are supported via the substitution mechanism. 

Currently, obtaining annotations attached to other reflective meta-objects, such as `Class` and `Method`, are supported since those objects are created and stored as constants in the `Class.ReflectionData` during image generation. However, `Package` objects are not handled in particular. Consequently, package-annotation-related calls return `null` and do not work as expected in the native images. 

Our solution is straightforward. First, in a feature, all the annotations attached to packages are obtained and stored into a map of an `ImageSingleton`, whose key is the package name and value is the corresponding annotations attached. Next, we leverage the substitution mechanism to replace original package-annotation-related calls with calls that obtain annotations from the afore-mentioned map. In this way,  package-annotation-related calls can return correct annotations. 

A test case is attached to this pull request. 
[pkgAnnotation.zip](https://github.com/oracle/graal/files/4537215/pkgAnnotation.zip)

First, unzip a jar file from the `pkgAnnotation.zip`, then use the following command to compile it. 
```
native-image -cp pkgAnnotation.jar packageinfo.GetPackageAnnotation
```

Excepted output:
```
Test Package.getAnnotation().
packageinfo.Author$$ProxyImpl, author: Anthony
packageinfo.Version$$ProxyImpl, version: 1.0

Test Package.getAnnotations().
packageinfo.Version$$ProxyImpl
packageinfo.Author$$ProxyImpl
packageinfo.Version$$ProxyImpl
packageinfo.Author$$ProxyImpl
```

Current output:
```
Test Package.getAnnotation().
Annotation is null.
Annotation is null.

Test Package.getAnnotations().
```

Please feel free to give any comments and suggestions to this pull request. 

Cheers,
Yifei


tfaller:
Package annotations already work. Like for other annotations you have to include the package in the reflection config of your native image configuration. Example for package `com.example.test`

```
[
  {
    "name" : "com.example.test.package-info"
  }
]
```
Note: package-info is the special place where package related docs and annotations are.
See also issue #1216

PR #2403, opened 2020-04-27
Expect AVX3Threshold on all JDKs >= 11

### LINKED ISSUE #2402: [OpenJDK 11u] HelloWorld native-image generation fails with enabled assertions
A graal VM built from OpenJDK 11u (upcoming 11.0.8) will hit an assertion on generating native images. A simple `HelloWorld` will do. I'm setting `java.vm.version` to a version which looks like a released version of OpenJDK so as to actually enable assertions at runtime. Developer builds won't show this issue.

Reproducer:
```
$ ./substratevm/svmbuild/vm/bin/native-image -J-Djava.vm.version=11.0.8+1 -J-ea -cp substratevm HelloWorld foo-bar
Exception in thread "main" jdk.vm.ci.common.JVMCIError: VM config values not expected to be present in JDK 11 linux-amd64 (java.home=/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_E6EC3842EE_JAVA11/graalvm-e6ec3842ee-java11-20.2.0-dev, java.vm.name=OpenJDK 64-Bit Server VM, java.vm.version=11.0.8+1):
    AVX3Threshold at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.GraalHotSpotVMConfig.<init>(GraalHotSpotVMConfig.java:157) [value: 4096]
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.GraalHotSpotVMConfigAccess.reportErrors(GraalHotSpotVMConfigAccess.java:224)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.GraalHotSpotVMConfig.<init>(GraalHotSpotVMConfig.java:67)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.HotSpotGraalRuntime.<init>(HotSpotGraalRuntime.java:161)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.HotSpotGraalCompilerFactory.createCompiler(HotSpotGraalCompilerFactory.java:156)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.HotSpotGraalCompilerFactory.createCompiler(HotSpotGraalCompilerFactory.java:134)
	at jdk.internal.vm.compiler/org.graalvm.compiler.hotspot.HotSpotGraalCompilerFactory.createCompiler(HotSpotGraalCompilerFactory.java:52)
	at jdk.internal.vm.ci/jdk.vm.ci.hotspot.HotSpotJVMCIRuntime.getCompiler(HotSpotJVMCIRuntime.java:425)
	at com.oracle.svm.hosted.c.GraalAccess.getGraalCapability(GraalAccess.java:50)
	at com.oracle.svm.hosted.c.GraalAccess.getOriginalTarget(GraalAccess.java:38)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.isValidArchitecture(NativeImageGeneratorRunner.java:221)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.verifyValidJavaVersionAndPlatform(NativeImageGeneratorRunner.java:426)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:232)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:518)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:117)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner$JDK9Plus.main(NativeImageGeneratorRunner.java:546)
```

Note that [JDK-8221092](https://bugs.openjdk.java.net/browse/JDK-8221092) got backported to OpenJDK 11.0.6 which introduced this flag.

jerboaa:
Instead of conditionalizing flag AVX3Threshold on a JVMCI-versioned
JDK 11, expect it for any JDK >= 11

Closes #2402

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
@@ -36,6 +36,7 @@ import org.graalvm.compiler.debug.Assertions;
 import org.graalvm.compiler.hotspot.JVMCIVersionCheck.Version;
 import org.graalvm.compiler.hotspot.JVMCIVersionCheck.Version2;
 import org.graalvm.compiler.hotspot.JVMCIVersionCheck.Version3;
+import org.graalvm.compiler.serviceprovider.GraalServices;
 import org.graalvm.compiler.serviceprovider.JavaVersionUtil;
 
 import jdk.vm.ci.common.JVMCIError;
@@ -137,6 +138,7 @@ public class GraalHotSpotVMConfigAccess {
     }
 
     public static final int JDK = JavaVersionUtil.JAVA_SPEC;
+    public static final int JDK_UPDATE = GraalServices.getJavaUpdateVersion();
     public static final boolean IS_OPENJDK = getProperty("java.vm.name", "").startsWith("OpenJDK");
     public static final Version JVMCI_VERSION;
     public static final boolean JVMCI;
--- a/compiler/src/org.graalvm.compiler.serviceprovider.jdk11/src/org/graalvm/compiler/serviceprovider/GraalServices.java
+++ b/compiler/src/org.graalvm.compiler.serviceprovider.jdk11/src/org/graalvm/compiler/serviceprovider/GraalServices.java
@@ -437,4 +437,8 @@ public final class GraalServices {
         }
         return VirtualObject.get(type, id);
     }
+
+    public static int getJavaUpdateVersion() {
+        return Runtime.version().update();
+    }
 }
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java
@@ -154,7 +154,7 @@ public class GraalHotSpotVMConfig extends GraalHotSpotVMConfigAccess {
     private final boolean useSquareToLenIntrinsic = getFlag("UseSquareToLenIntrinsic", Boolean.class, false, isJDK8OrJDK11Plus);
     public final boolean useVectorizedMismatchIntrinsic = getFlag("UseVectorizedMismatchIntrinsic", Boolean.class, false, isJDK11Plus);
     public final boolean useFMAIntrinsics = getFlag("UseFMA", Boolean.class, false, JDK >= 9);
-    public final int useAVX3Threshold = getFlag("AVX3Threshold", Integer.class, 4096, osArch.equals("amd64") && (JVMCI ? JDK >= 11 : JDK >= 14));
+    public final int useAVX3Threshold = getFlag("AVX3Threshold", Integer.class, 4096, osArch.equals("amd64") && (JDK >= 14 || (JDK == 11 && JDK_UPDATE >= 6)));
 
     public final boolean preserveFramePointer = getFlag("PreserveFramePointer", Boolean.class);
 
--- a/compiler/src/org.graalvm.compiler.serviceprovider.jdk13/src/org/graalvm/compiler/serviceprovider/GraalServices.java
+++ b/compiler/src/org.graalvm.compiler.serviceprovider.jdk13/src/org/graalvm/compiler/serviceprovider/GraalServices.java
@@ -554,4 +554,8 @@ public final class GraalServices {
     public static VirtualObject createVirtualObject(ResolvedJavaType type, int id, boolean isAutoBox) {
         return VirtualObject.get(type, id, isAutoBox);
     }
+
+    public static int getJavaUpdateVersion() {
+        return Runtime.version().update();
+    }
 }
--- a/compiler/src/org.graalvm.compiler.serviceprovider.jdk8/src/org/graalvm/compiler/serviceprovider/GraalServices.java
+++ b/compiler/src/org.graalvm.compiler.serviceprovider.jdk8/src/org/graalvm/compiler/serviceprovider/GraalServices.java
@@ -36,6 +36,8 @@ import java.util.List;
 import java.util.ServiceConfigurationError;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import jdk.vm.ci.code.VirtualObject;
 import jdk.vm.ci.meta.ResolvedJavaField;
@@ -433,5 +435,18 @@ public final class GraalServices {
     @SuppressWarnings("unused")
     public static VirtualObject createVirtualObject(ResolvedJavaType type, int id, boolean isAutoBox) {
         return VirtualObject.get(type, id, isAutoBox);
+    }
+
+    public static int getJavaUpdateVersion() {
+        // JDK 8: Only simplified patterns like 25.242-b08 or 25.241-b07-jvmci-20.1-b01
+        // are being recognized. Update represents the numerical value after the first
+        // dot and before the first dash
+        Pattern p = Pattern.compile("\\d+\\.([^-]+)-.*");
+        String vmVersion = Services.getSavedProperties().get("java.vm.version");
+        Matcher matcher = p.matcher(vmVersion);
+        if (!matcher.matches()) {
+            throw new InternalError("Unexpected java.vm.version value: " + vmVersion);
+        }
+        return Integer.parseInt(matcher.group(1));
     }
 }
--- a/compiler/src/org.graalvm.compiler.serviceprovider/src/org/graalvm/compiler/serviceprovider/GraalServices.java
+++ b/compiler/src/org.graalvm.compiler.serviceprovider/src/org/graalvm/compiler/serviceprovider/GraalServices.java
@@ -244,4 +244,13 @@ public final class GraalServices {
     public static VirtualObject createVirtualObject(ResolvedJavaType type, int id, boolean isAutoBox) {
         throw shouldNotReachHere();
     }
+
+    /**
+     * Gets the update-release counter for the current Java runtime.
+     *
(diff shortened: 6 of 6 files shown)


dougxc:
The check will now fail on earlier JDK 11s that don't have this flag so the check needs to take the update version into account.

jerboaa:
> The check will now fail on earlier JDK 11s that don't have this flag so the check needs to take the update version into account.

I take it `>= 11.0.6` would be OK?

dougxc:
I think so.

jerboaa:
> I take it `>= 11.0.6` would be OK?

This added reflectiveness makes it quite ugly. Perhaps there is some less clunky way to do this?

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
```diff
@@ -146,6 +149,19 @@ public static boolean jvmciGE(Version v) {
         JVMCI_VERSION = Version.parse(vmVersion);
         JVMCI_PRERELEASE = vmVersion.contains("SNAPSHOT") || vmVersion.contains("internal") || vmVersion.contains("-dev");
         JVMCI = JVMCI_VERSION != null || JVMCI_PRERELEASE;
+        int jdkUpdateVers = UNSET;
```
dougxc, line 152: You should use a regular expression to parse the value of the `java.vm.version` system property. Here are some sample values:
11.0.7+8-LTS-jvmci-20.1-b02
25.241-b07-jvmci-20.1-b01

jerboaa: `UNSET` is just the default. I was trying to avoid all this version parsing mess. In particular, for this issue I'm not sure it's worth doing the right thing for JDK 8, is it? If that's needed, I'd rather have a specialized class which only deals with that and is unit-tested.

So if `Runtime.version().update()` cannot be used, how about `Runtime.Version.parse(<value>)` where `<value>` could be coming from `java.vm.version` property as you say.

jerboaa: So I'd be something like: http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/e6178063e9dd/test/sun/misc/Version/Version.java#l139 for `java.vm.version` parsing for JDK 8. For JDK 9+ we really should leverage the JDK for the parsing rather than writing our own.

dougxc: For 8, *only* JVMCI JDKs can be used (currently) so we can rely on JVMCI versions there.

dougxc: I think it's simplest to just use regular expressions. Something like this:
```
diff --git a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java
index 45eeb0aff99..4d02f306304 100644
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java
@@ -154,7 +154,7 @@ public class GraalHotSpotVMConfig extends GraalHotSpotVMConfigAccess {
     private final boolean useSquareToLenIntrinsic = getFlag("UseSquareToLenIntrinsic", Boolean.class, false, isJDK8OrJDK11Plus);
     public final boolean useVectorizedMismatchIntrinsic = getFlag("UseVectorizedMismatchIntrinsic", Boolean.class, false, isJDK11Plus);
     public final boolean useFMAIntrinsics = getFlag("UseFMA", Boolean.class, false, JDK >= 9);
-    public final int useAVX3Threshold = getFlag("AVX3Threshold", Integer.class, 4096, osArch.equals("amd64") && (JVMCI ? JDK >= 11 : JDK >= 14));
+    public final int useAVX3Threshold = getFlag("AVX3Threshold", Integer.class, 4096, osArch.equals("amd64") && (JVMCI ? JDK >= 11 : JDK == 11 ? JDK_UPDATE >= 6 : JDK >= 14));
 
     public final boolean preserveFramePointer = getFlag("PreserveFramePointer", Boolean.class);
 
diff --git a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
index b99cbd6975b..b9f84d97132 100644
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
@@ -30,6 +30,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
 import java.util.stream.Collectors;
 
 import org.graalvm.compiler.debug.Assertions;
@@ -137,6 +139,7 @@ public class GraalHotSpotVMConfigAccess {
     }
 
     public static final int JDK = JavaVersionUtil.JAVA_SPEC;
+    public static final int JDK_UPDATE;
     public static final boolean IS_OPENJDK = getProperty("java.vm.name", "").startsWith("OpenJDK");
     public static final Version JVMCI_VERSION;
     public static final boolean JVMCI;
@@ -146,6 +149,13 @@ public class GraalHotSpotVMConfigAccess {
         JVMCI_VERSION = Version.parse(vmVersion);
         JVMCI_PRERELEASE = vmVersion.contains("SNAPSHOT") || vmVersion.contains("internal") || vmVersion.contains("-dev");
         JVMCI = JVMCI_VERSION != null || JVMCI_PRERELEASE;
+
+        String javaVersion = getProperty("java.vm.version");
+        Matcher m = Pattern.compile("(?:1.8.0_(\\d+))|(?:(\\d+)\\.(\\d+)\\.(\\d+))").matcher(javaVersion);
+        if (!m.matches()) {
+            throw new JVMCIError("Unexpected java.version value: " + javaVersion);
+        }
+        JDK_UPDATE = Integer.parseInt(m.group(1) != null ? m.group(1) : m.group(4));
     }
 
     private final List<String> missing = new ArrayList<>();
```

jerboaa: That's what I mean. It'll fail for all sorts of weird versions. I'm pretty sure this fails for my hypothetical JDK 8 here: `java.vm.version = 25.242-b08`. Since we don't need to cover JDK 8 - covered by jvmci, why would we want to craft a custom parser which then fails in strange ways. Example for JDK 14 I have here: `java.vm.version = 14+36`.

dougxc: Can there really be a lot of variation in values for the `java.version` system property? The specification for [Runtime.Version](https://download.java.net/java/GA/jdk14/docs/api/java.base/java/lang/Runtime.Version.html) includes a regular expression for a _version number_ and I think `java.version` will always be a version number as opposed a more variable _version string_. In any case a version string is specified to always start with a version number.

If your really want to rely on using `Runtime.Version` for parsing `java.version` (which I'm not objecting to), then you will have to use an [overlay](https://github.com/graalvm/mx/blob/master/README.md#versioning-sources-for-different-jdk-releases) which is a larger change.

jerboaa:
@dougxc I've updated it to use JDK parsing logic for JDK 11+ and a simple regex for JDK 8. Hope that seems reasonable.

dougxc:
We only use reflection where absolutely necessary in Graal, especially in Graal initialization code. Can you please replace the reflection with a `int[] getJavaVersion()` method in `GraalServices`.

jerboaa:
> We only use reflection where absolutely necessary in Graal, especially in Graal initialization code. Can you please replace the reflection with this new method in `GraalServices`:
> 
> ```
>     /**
>      * Gets the update-release counter for the current Java runtime.
>      *
>      * @see "https://download.java.net/java/GA/jdk14/docs/api/java.base/java/lang/Runtime.Version.html"
>      */
>     public static int getJavaUpdateVersion() {
>         throw shouldNotReachHere();
>     }
> ```

OK.

compiler/src/org.graalvm.compiler.serviceprovider.jdk11/src/org/graalvm/compiler/serviceprovider/GraalServices.java
```diff
@@ -437,4 +437,8 @@ public static VirtualObject createVirtualObject(ResolvedJavaType type, int id, b
         }
         return VirtualObject.get(type, id);
     }
+
+    public static int getJavaUpdateVersion() {
+        return Runtime.version().update();
```
jerboaa, line 442: @dougxc Not sure if you rather wanted `Runtime.Version.parse(System.getProperty("java.vm.version")).update()` here. Let me know if so.

dougxc: No, this is exactly what we want. If `Runtime.Version.parse(System.getProperty("java.vm.version")).update() != Runtime.version().update()`, then I'd say there are bigger problems.

jerboaa:
Thanks for the review and your help!

PR #2408, opened 2020-04-27
Class initialization optimization.

jovanstevanovic:


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java
```diff
@@ -1294,6 +1295,11 @@ private static Suites modifySuites(SubstrateBackend backend, Suites suites, Feat
         pos.next();
         pos.add(new StackValuePhase());
 
+        if (hosted) {
+            position = highTier.findPhase(DeadStoreRemovalPhase.class);
+            position.add(new RemoveRedundantClassInitPhase(runtimeCallProviders.getMetaAccess()));
```
christianwimmer, line 1300: You have a phase ordering problem here: You insert the phase before inlining because you rely on the ensure initialized method to not be inlined. But that means you are missing a lot of optimization potential because many more checks could be removed after inlining and after high-level optimizations such as loop peeling.

The solution for this is to model class initialization as a separate Graal node that is then lowered to the actual check using a snippet. Then you can insert this phase much lower in the compilation pipeline.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/RemoveRedundantClassInitClosure.java
```diff
@@ -0,0 +1,135 @@
...
+                        classInitMethod == invokeWithExceptionNode.callTarget().targetMethod()) {
+
+                    ValueNode valueNode = invokeWithExceptionNode.callTarget().arguments().get(WHICH_ARGUMENT_INDEX_IS_CLASS);
+                    if (valueNode.isJavaConstant()) {
+                        JavaConstant javaConstant = valueNode.asJavaConstant();
+                        if (!currentState.contains(javaConstant)) {
+                            currentState.add(javaConstant);
+                            nodesToRemain.put(invokeWithExceptionNode, invokeWithExceptionNode.next());
+                        } else {
+                            nodesToRemain.removeKey(invokeWithExceptionNode);
```
christianwimmer, line 98: How can this node ever be in `nodesToRemain` already?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/RemoveRedundantClassInitClosure.java
```diff
@@ -0,0 +1,135 @@
...
+    public EconomicMap<InvokeWithExceptionNode, AbstractBeginNode> getNodesToRemain() {
+        return nodesToRemain;
+    }
+
+    @Override
+    protected EconomicSet<JavaConstant> getInitialState() {
+        this.nodesToBeRemoved = EconomicMap.create(Equivalence.IDENTITY);
+        this.nodesToRemain = EconomicMap.create(Equivalence.IDENTITY);
+
+        return EconomicSet.create(Equivalence.IDENTITY);
```
christianwimmer, line 78: Why do you explicitly request `Equivalence.IDENTITY` for this set (and all the other sets/maps)? While all constants should be canonicalized, there is no need to restrict the optimization in case the canonicalization did not happen.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/RemoveRedundantClassInitPhase.java
```diff
@@ -0,0 +1,72 @@
...
+        // After closure processing, they are removed (with propagation) in the code below.
+        MapCursor<InvokeWithExceptionNode, AbstractBeginNode> cursor = removeRedundantClassInitClosure.getNodesToBeRemoved().getEntries();
+        while (cursor.advance()) {
+            InvokeWithExceptionNode invokeWithExceptionNode = cursor.getKey();
+            AbstractBeginNode survivingSuccessor = invokeWithExceptionNode.killKillingBegin();
+
+            invokeWithExceptionNode.graph().removeSplitPropagate(invokeWithExceptionNode, survivingSuccessor);
+        }
+
+        // During the inline phase, class initialization nodes aren't inlined. So in this phase, we need to do it manually.
```
christianwimmer, line 64: Actually in your whole analysis you are assuming (without any check) that the trivial method inlining is not already inlining the ensure initialized method. The node-based approach of course takes care of that problem too.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/RemoveRedundantClassInitClosure.java
jovanstevanovic: My mistake. I removed it.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/RemoveRedundantClassInitClosure.java
jovanstevanovic: Changed to default equivalence type.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java
jovanstevanovic: Answer related to PR closing. I will move this changelist to private PR, as soon as I get access. That's the reason why I closed this PR in advance, without merge.

Issue #2378, opened 2020-04-22
[native-image] AArch64: Using GDB errors with unknown CFA value


a74nh:
Take a simple java program. On AArch64, create a native image using the debug info flag. Then load it into GDB. Break at main, run, then backtrace. GDB errors with unknown CFA value.

 ```
$ native-image -H:GenerateDebugInfo=1 FPAdder

$ gdb fpadder
GNU gdb (GDB) 10.0.50.20200416-git
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "aarch64-unknown-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./fpadder...
(gdb) b main
Breakpoint 1 at 0x2300c: main. (2 locations)
(gdb) r
Starting program: /home/alahay01/javatut/fpadder
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/aarch64-linux-gnu/libthread_db.so.1".

Breakpoint 1, 0x0000aaaaaaade528 in com.oracle.svm.core.code.IsolateEnterStub::JavaMainWrapper_run_5087f5482cc9a6abc971913ece43acb471d2631b(int, org.graalvm.nativeimage.c.type.CCharPointerPointer)(void) ()
    at com/oracle/svm/core/code/IsolateEnterStub.java:5
5	com/oracle/svm/core/code/IsolateEnterStub.java: No such file or directory.
(gdb) bt
#0  0x0000aaaaaaade528 in com.oracle.svm.core.code.IsolateEnterStub::JavaMainWrapper_run_5087f5482cc9a6abc971913ece43acb471d2631b(int, org.graalvm.nativeimage.c.type.CCharPointerPointer)(void) ()
    at com/oracle/svm/core/code/IsolateEnterStub.java:5
../../src/binutils-gdb/gdb/dwarf2/frame.c:1077: internal-error: Unknown CFA rule.
A problem internal to GDB has been detected,
[...]

PR #2411, opened 2020-04-28
AArch64 Debug Info fixes

adinn:
This patch fixes several problems with debug info on AArch64. Associated issue is https://github.com/oracle/graal/issues/2378.

Fixed functionality includes

.debug_frame section CIE records:
  establish correct stack call frame address at entry 

.debug_line section:
  restate initial line position at end of method prologue  -- workaround for bad gdb behaviour

.debug_frame section FDE records:
  notify gdb of save/restore for frame and link register 

In terms of user experience what this achieves is that gdb now
    - places method breaks at the first rather than the second line of a method
    - does not crash when you try to display a stack backtrace
    - correctly identifies the return address and stack top for the previous frame

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfDebugInfo.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfDebugInfo.java
@@ -163,7 +163,7 @@ public class DwarfDebugInfo extends DebugInfoBase {
     private DwarfARangesSectionImpl dwarfARangesSection;
     private DwarfLineSectionImpl dwarfLineSection;
     private DwarfFrameSectionImpl dwarfFameSection;
-    private ELFMachine elfMachine;
+    public final ELFMachine elfMachine;
 
     public DwarfDebugInfo(ELFMachine elfMachine, ByteOrder byteOrder) {
         super(byteOrder);
@@ -202,9 +202,5 @@ public class DwarfDebugInfo extends DebugInfoBase {
 
     public DwarfLineSectionImpl getLineSectionImpl() {
         return dwarfLineSection;
-    }
-
-    public ELFMachine getELFMachine() {
-        return elfMachine;
     }
 }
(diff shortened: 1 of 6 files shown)


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfDebugInfo.java
```diff
@@ -201,4 +203,8 @@ public DwarfARangesSectionImpl getARangesSectionImpl() {
     public DwarfLineSectionImpl getLineSectionImpl() {
         return dwarfLineSection;
     }
+
+    public ELFMachine getELFMachine() {
```
olpaw, line 207: You don't need an accessor if you make `elfMachine` a public final field.

adinn: Sure, no point making the API surface larger than it needs to be. Fixed

adinn:
@olpaw  Hmm, Travis seems to have failed for some spurious reason. Can this be pulled despite that failure?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLineSectionImpl.java
```diff
@@ -478,6 +479,32 @@ private int writeLineNumberTable(DebugContext context, ClassEntry classEntry, by
...
+             * On AArch64 gdb expects to see a line record at the start of the method and a second
+             * one at the end of the prologue marking the point where the method code begins for
+             * real. If we don't provide it then gdb will skip to the second line record when we
+             * place a breakpoint on the method.
+             *
+             * We can identify the end of the prologue for normal methods by noting where the stack
+             * frame height is first adjusted. This should normally be no more a few instructions in
+             * total.
+             */
+            if (isAArch64() && !primaryEntry.getFrameSizeInfos().isEmpty()) {
```
olpaw, line 492: If doing the same also for AMD64 does not have any negative side effects then maybe it's not worth to make this piece architecture specific.

olpaw:
> @olpaw Hmm, Travis seems to have failed for some spurious reason. Can this be pulled despite that failure?

Sure. Not sure though if it will merge before Monday.

PR #2413, opened 2020-04-28
Fix static image IPv6 support.

### LINKED ISSUE #2373: IPv6 static native image throws java.net.SocketException: Protocol family unavailable
**Describe the issue**
Executing a static native image binary built using muslC throws a SocketException when using an IPv6 address

Please include both build steps as well as run steps
1. Install scala build tool `SBT` https://www.scala-sbt.org/download.html
2. Clone https://github.com/drocsid/https_example
3. Execute the script native-image-build-cmd.sh which builds a static image using muslC
4. Execute the nativeImage  binary and get the stack trace.

```
java.net.SocketException: Protocol family unavailable
	at com.oracle.svm.jni.JNIJavaCallWrappers.jniInvoke_ARRAY:Ljava_net_SocketException_2_0002e_0003cinit_0003e_00028Ljava_lang_String_2_00029V(JNIJavaCallWrappers.java:0)
	at com.oracle.svm.jni.functions.JNIFunctions$NewObjectWithObjectArrayArgFunctionPointer.invoke(JNIFunctions.java)
	at com.oracle.svm.jni.functions.JNIFunctions.ThrowNew(JNIFunctions.java:801)
	at sun.nio.ch.Net.bind0(Net.java)
	at sun.nio.ch.Net.bind(Net.java:455)
	at sun.nio.ch.Net.bind(Net.java:447)
	at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:227)
	at java.nio.channels.ServerSocketChannel.bind(ServerSocketChannel.java:162)
	at org.http4s.blaze.channel.nio1.NIO1SocketServerGroup.$anonfun$bind$1(NIO1SocketServerGroup.scala:217)
	at scala.util.Try$.apply(Try.scala:210)
	at org.http4s.blaze.channel.nio1.NIO1SocketServerGroup.bind(NIO1SocketServerGroup.scala:216)
	at org.http4s.blaze.channel.nio1.NIO1SocketServerGroup$$anon$1.bind(NIO1SocketServerGroup.scala:79)
	at org.http4s.server.blaze.BlazeServerBuilder.$anonfun$resource$5(BlazeServerBuilder.scala:341)
	at scala.Function1.$anonfun$andThen$1(Function1.scala:85)
	at scala.Function1.$anonfun$andThen$1(Function1.scala:85)
	at cats.effect.IO$Map.apply(IO.scala:1504)
	at cats.effect.IO$Map.apply(IO.scala:1502)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:142)
	at cats.effect.internals.IORunLoop$.startCancelable(IORunLoop.scala:41)
	at cats.effect.internals.IOBracket$BracketStart.run(IOBracket.scala:88)
	at cats.effect.internals.Trampoline.cats$effect$internals$Trampoline$$immediateLoop(Trampoline.scala:67)
	at cats.effect.internals.Trampoline.startLoop(Trampoline.scala:35)
	at cats.effect.internals.TrampolineEC$JVMTrampoline.super$startLoop(TrampolineEC.scala:89)
	at cats.effect.internals.TrampolineEC$JVMTrampoline.$anonfun$startLoop$1(TrampolineEC.scala:89)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:94)
	at cats.effect.internals.TrampolineEC$JVMTrampoline.startLoop(TrampolineEC.scala:89)
	at cats.effect.internals.Trampoline.execute(Trampoline.scala:43)
	at cats.effect.internals.TrampolineEC.execute(TrampolineEC.scala:42)
	at cats.effect.internals.IOBracket$BracketStart.apply(IOBracket.scala:69)
	at cats.effect.internals.IOBracket$BracketStart.apply(IOBracket.scala:49)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:139)
	at cats.effect.internals.IORunLoop$.start(IORunLoop.scala:34)
	at cats.effect.internals.IOBracket$.$anonfun$apply$1(IOBracket.scala:42)
	at cats.effect.internals.IOBracket$.$anonfun$apply$1$adapted(IOBracket.scala:32)
	at cats.effect.internals.IORunLoop$RestartCallback.start(IORunLoop.scala:345)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:122)
	at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:359)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:380)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:323)
	at cats.effect.internals.IOShift$Tick.run(IOShift.scala:35)
	at cats.effect.internals.PoolUtils$$anon$2$$anon$3.run(PoolUtils.scala:52)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.lang.Thread.run(Thread.java:834)
	at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:527)
	at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:193)
```

java --version
openjdk 11.0.6 2020-01-14
OpenJDK Runtime Environment GraalVM CE 20.0.0 (build 11.0.6+9-jvmci-20.0-b02)
OpenJDK 64-Bit Server VM GraalVM CE 20.0.0 (build 11.0.6+9-jvmci-20.0-b02, mixed mode, sharing)

OS: Debian 4.19.0-4-amd64 bullseye
Architecture: AMD64

**More details**
https://github.com/drocsid/https_example/blob/aab787f5ce12e621002f1e661a2972f63df351c8/README.md

gradinac:
The issue is described in more detail here: [#2373 comment](https://github.com/oracle/graal/issues/2373#issuecomment-619147560 )

This PR fixes support for IPv6 in static native-images. It also adds a check to see whether the `JVM_FindLibraryEntry` has been called while running a static native-image binary, and force the program to exit if that is the case, instead of allowing implementation specific behavior.

JDK 11 call that checks for IPv6 support can be located at: https://github.com/graalvm/labs-openjdk-11/blob/master/src/java.base/unix/native/libnet/net_util_md.c, function `jint  IPv6_supported()`, called during `libnet` initialization.

Closes #2373

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/VMFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/VMFeature.java
@@ -32,10 +32,13 @@ import java.util.stream.Collectors;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.Platform;
 import org.graalvm.nativeimage.hosted.Feature;
+import org.graalvm.word.PointerBase;
+import org.graalvm.word.WordFactory;
 
 import com.oracle.svm.core.SubstrateOptions;
 import com.oracle.svm.core.VM;
 import com.oracle.svm.core.annotate.AutomaticFeature;
+import com.oracle.svm.core.c.CGlobalData;
 import com.oracle.svm.core.c.CGlobalDataFactory;
 import com.oracle.svm.core.c.libc.LibCBase;
 import com.oracle.svm.core.util.VMError;
@@ -47,6 +50,7 @@ import com.oracle.svm.hosted.c.codegen.CCompilerInvoker;
 public class VMFeature implements Feature {
 
     private NativeLibraries nativeLibraries;
+    private static final String STATIC_BINARY_MARKER_SYMBOL_NAME = "__svm_vm_is_static_binary";
 
     @Override
     public void beforeAnalysis(BeforeAnalysisAccess a) {
@@ -77,8 +81,7 @@ public class VMFeature implements Feature {
         addCGlobalDataString("Target.LibC", ImageSingletons.lookup(LibCBase.class).getClass().getName());
 
         addCGlobalDataString("Target.Libraries", String.join("|", nativeLibraries.getLibraries()));
-        addCGlobalDataString("Target.StaticLibraries", nativeLibraries.getStaticLibraries().stream()
-                        .map(Path::getFileName).map(Path::toString).collect(Collectors.joining("|")));
+        addCGlobalDataString("Target.StaticLibraries", nativeLibraries.getStaticLibraries().stream().map(Path::getFileName).map(Path::toString).collect(Collectors.joining("|")));
         if (ImageSingletons.contains(CCompilerInvoker.class)) {
             addCGlobalDataString("Target.CCompiler", ImageSingletons.lookup(CCompilerInvoker.class).compilerInfo.toString());
         }
@@ -89,6 +92,9 @@ public class VMFeature implements Feature {
             nativeLibraries.getStaticLibraries().stream().map(current::relativize).map(Path::toString).forEach(x -> System.out.println("#   " + x));
             System.out.println("# Other libraries: " + String.join(",", nativeLibraries.getLibraries()));
         }
+
+        CGlobalData<PointerBase> isStaticBinaryMarker = CGlobalDataFactory.createWord(WordFactory.unsigned(SubstrateOptions.StaticExecutable.hasBeenSet() ? ~1 : 0), STATIC_BINARY_MARKER_SYMBOL_NAME);
+        CGlobalDataFeature.singleton().registerAsAccessedOrGet(isStaticBinaryMarker);
     }
 
     private static void addCGlobalDataString(String infoType, String content) {
--- a/substratevm/src/com.oracle.svm.native.jvm.posix/src/JvmFuncs.c
+++ b/substratevm/src/com.oracle.svm.native.jvm.posix/src/JvmFuncs.c
@@ -44,6 +44,16 @@
 #define OS_OK 0
 #define OS_ERR -1
 
+/* Set by native-image during image build time. Indicates whether the built image is a static binary. */
+extern int __svm_vm_is_static_binary;
+/*
+    The way JDK checks IPv6 support on Linux involves checking if this function exists using JVM_FindLibraryEntry. That
+    function in turn calls dlsym, which is a bad idea in a static binary. Since this function is guaranteed to exist on
+    every POSIX platform we support, we ask the linker to provide us with the symbol value, which we will then return to
+    the test code.
+*/
+int inet_pton(int af, const char *src, void *dst);
+
 #ifdef JNI_VERSION_9
     #define JVM_INTERFACE_VERSION 6
 #else
@@ -78,7 +88,24 @@ JNIEXPORT int JNICALL JVM_Connect(int fd, struct sockaddr* him, socklen_t len) {
 }
 
 JNIEXPORT void* JNICALL JVM_FindLibraryEntry(void* handle, const char* name) {
-    return dlsym(handle, name);
+    /*
+        Calls to this function from a static binary are inherently unsafe. On some libc implementations, it may result
+        in a segfault, while on others, the symbol could be wrongly not found. As of JDK11, this function is invoked in
+        only one place: IPv6 support checking.
+        As a safeguard, we restrict access to only known used symbols from the JDK code. If a future version introduces
+        a dependency on another symbol, it could result in hard-to-find bugs. Therefore, calling this function from a
+        static binary with an unknown symbol terminates the program.
+    */
+    if (__svm_vm_is_static_binary) {
+        if (strcmp(name, "inet_pton") == 0) {
+            return inet_pton;
+        }
+        fprintf(stderr, "Internal error: JVM_FindLibraryEntry called from a static native image. Results may be unpredictable. Please report this issue to the SubstrateVM team.");
+        fflush(stderr);
+        exit(1);
+    } else {
+        return dlsym(handle, name);
+    }
 }
 
 JNIEXPORT int JNICALL JVM_GetHostName(char* name, int namelen) {


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/VMFeature.java
```diff
@@ -89,6 +92,9 @@ public void afterAnalysis(AfterAnalysisAccess access) {
             nativeLibraries.getStaticLibraries().stream().map(current::relativize).map(Path::toString).forEach(x -> System.out.println("#   " + x));
             System.out.println("# Other libraries: " + String.join(",", nativeLibraries.getLibraries()));
         }
+
+        CGlobalData<PointerBase> isStaticBinaryMarker = CGlobalDataFactory.createWord(WordFactory.unsigned(SubstrateOptions.StaticExecutable.hasBeenSet() ? ~0 : 0), STATIC_BINARY_MARKER_SYMBOL_NAME);
```
olpaw, line 96: I'd use `-1` instead of `~0` (although a matter of taste).

substratevm/src/com.oracle.svm.native.jvm.posix/src/JvmFuncs.c
```diff
@@ -78,7 +88,24 @@ JNIEXPORT int JNICALL JVM_Connect(int fd, struct sockaddr* him, socklen_t len) {
...
-    return dlsym(handle, name);
+    /*
+        Calls to this function from a static binary are inherently unsafe. On some libc implementations, it may result
+        in a segfault, while on others, the symbol could be wrongly not found. As of JDK11, this function is invoked in
+        only one place: IPv6 support checking.
+        As a safeguard, we restrict access to only known used symbols from the JDK code. If a future version introduces
+        a dependency on another symbol, it could result in hard-to-find bugs. Therefor, calling this function from a
+        static binary with an unknown symbol terminates the program.
+    */
+    if (__svm_vm_is_static_binary) {
```
olpaw, line 99: This is neat!

substratevm/src/com.oracle.svm.native.jvm.posix/src/JvmFuncs.c
```diff
@@ -78,7 +88,24 @@ JNIEXPORT int JNICALL JVM_Connect(int fd, struct sockaddr* him, socklen_t len) {
...
+        if (strcmp(name, "inet_pton") == 0) {
+            return inet_pton;
+        }
+        fprintf(stderr, "Internal error: JVM_FindLibraryEntry called from a static native image. Results may be unpredictable. Please report this issue to the SubstrateVM team.");
```
olpaw, line 103: Use
```
(*env)->FatalError(env, "Internal error: JVM_FindLibraryEntry called from a static native image. Results may be unpredictable. Please report this issue to the SubstrateVM team.");
```
instead. It has the advantage that we see the whole stack trace that caused the situation.

olpaw: Ahh that's not possible here because we don't get `JNIEnv` passed into `JVM_FindLibraryEntry`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/VMFeature.java
gradinac: I have changed it to -1 instead of ~0. :)

bobvandette:
If the symbol inet_pton exist in the generated static image then why doesn't dlsym find it?
We're passing RTLD_DEFAULT which should look for the symbols in the executable.  Does musl libc dlsym not work the same way as libc or is inet_pton not declared as a global symbol in the executable image?

bobvandette:
In the event that RTLD_DEFAULT doesn't work when linked against musl libc, you might want to take a look at other uses of RTLD_DEFAULT in the JDK native libraries (nio and net) to make sure they won't impact native-image generated executables.  

jdk15/open/src/java.base$ search RTLD_DEFAULT
./linux/native/libnio/fs/LinuxNativeDispatcher.c:    my_fgetxattr_func = (fgetxattr_func*)dlsym(RTLD_DEFAULT, "fgetxattr");
./linux/native/libnio/fs/LinuxNativeDispatcher.c:    my_fsetxattr_func = (fsetxattr_func*)dlsym(RTLD_DEFAULT, "fsetxattr");
./linux/native/libnio/fs/LinuxNativeDispatcher.c:    my_fremovexattr_func = (fremovexattr_func*)dlsym(RTLD_DEFAULT, "fremovexattr");
./linux/native/libnio/fs/LinuxNativeDispatcher.c:    my_flistxattr_func = (flistxattr_func*)dlsym(RTLD_DEFAULT, "flistxattr");
./unix/native/libjli/java_md_common.c:       findBootClass = (FindClassFromBootLoader_t *)dlsym(RTLD_DEFAULT,
./unix/native/libjli/java_md_solinux.c:        fptr = (int (*)())dlsym(RTLD_DEFAULT, "main");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_openat64_func = (openat64_func*)dlsym(RTLD_DEFAULT, "openat");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_fstatat64_func = (fstatat64_func*)dlsym(RTLD_DEFAULT, "fstatat");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_openat64_func = (openat64_func*) dlsym(RTLD_DEFAULT, "openat64");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_fstatat64_func = (fstatat64_func*) dlsym(RTLD_DEFAULT, "fstatat64");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_unlinkat_func = (unlinkat_func*) dlsym(RTLD_DEFAULT, "unlinkat");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_renameat_func = (renameat_func*) dlsym(RTLD_DEFAULT, "renameat");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_futimesat_func = (futimesat_func*) dlsym(RTLD_DEFAULT, "futimesat");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_lutimes_func = (lutimes_func*) dlsym(RTLD_DEFAULT, "lutimes");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_futimens_func = (futimens_func*) dlsym(RTLD_DEFAULT, "futimens");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_fdopendir_func = (fdopendir_func*) dlsym(RTLD_DEFAULT, "fdopendir64");
./unix/native/libnio/fs/UnixNativeDispatcher.c:    my_fdopendir_func = (fdopendir_func*) dlsym(RTLD_DEFAULT, "fdopendir");
./unix/native/libnet/net_util_md.c:    ipv6_fn = JVM_FindLibraryEntry(RTLD_DEFAULT, "inet_pton");
./unix/native/libnet/DefaultProxySelector.c:                (g_type_init_func*)dlsym(RTLD_DEFAULT, "g_type_init");
./unix/native/libnet/DefaultProxySelector.c:                (gconf_client_get_default_func*)dlsym(RTLD_DEFAULT,
./unix/native/libnet/DefaultProxySelector.c:                        (gconf_client_get_string_func*)dlsym(RTLD_DEFAULT,
./unix/native/libnet/DefaultProxySelector.c:                        (gconf_client_get_int_func*)dlsym(RTLD_DEFAULT,
./unix/native/libnet/DefaultProxySelector.c:                        (gconf_client_get_bool_func*)dlsym(RTLD_DEFAULT,

bobvandette:
Also, if this issue is a musl libc problem, your flag that you are using to enable the work-around should be based on a MUSLC specific flag and not the fact that you are building statically.  If you build a static library for Linux, you might not need this work-around.

bobvandette:
Could someone try using the latest muslC library version 1.2.0 to see if it resolves this problem without this change?

gradinac:
@bobvandette This issue is not musl specific, although that is where we found the first manifestation of it. It is specific to static native-images.
The `inet_pton` symbol is provided by the implementing libc (in this case - musl c, but it can also be provided by glibc).
The issue lies in calling `dlsym`, a function meant for searching through the dynamic symbol table for a symbol provided by a loaded dynamic library. As the JDK libraries were not meant to be used in a static executable, they make use of `dlsym` for some operations.
When using JDK as it was intended, libc will be dynamically loaded inside the JVM at runtime, which then behaves as expected.
This issue can manifest itself in different ways. For example, on the last few musl releases and the glibc release provided on Ubuntu 18.04 (2.27), the symbol will not be found. For the glibc version provided in Oracle Linux 6, we would get a segfault.
Thank you for the list of `dlsym` usages, I will investigate them and see if any more conflicts can arise in that area.

gradinac:
One note: Right now, musl is supported only on JDK11 with the primary purpose of creating static native-images, so I will focus on the `labs-openjdk-11` source code.

gradinac:
Here is a list of files inside `java.base` which use `RTLD_DEFAULT`:

 - ./macosx/native/libjli/java_md_macosx.c
 - ./aix/native/libnio/ch/AixPollPort.c
 - ./linux/native/libnio/fs/LinuxNativeDispatcher.c
 - ./unix/native/libnio/fs/UnixNativeDispatcher.c
 - ./unix/native/libjli/java_md_common.c
 - ./unix/native/libjli/java_md_solinux.c
 - ./unix/native/libnet/DefaultProxySelector.c
 - ./unix/native/libnet/net_util_md.c
 - ./unix/native/libjava/jdk_util_md.c

As we are only interested in Linux, we will remove files specific to other systems:

 - ./linux/native/libnio/fs/LinuxNativeDispatcher.c
 - ./unix/native/libnio/fs/UnixNativeDispatcher.c
 - ./unix/native/libjli/java_md_common.c
 - ./unix/native/libjli/java_md_solinux.c
 - ./unix/native/libnet/DefaultProxySelector.c
 - ./unix/native/libnet/net_util_md.c
 - ./unix/native/libjava/jdk_util_md.c
---
Here are the noted issues, broken down by file groups:

 - ./linux/native/libnio/fs/LinuxNativeDispatcher.c
 - ./unix/native/libnio/fs/UnixNativeDispatcher.c

These two files appear to grab functions necessary for file manipulation. These appear to be used and this could cause us issues in a static native-image. I will investigate this further and see if I can create an example that misbehaves in a static native-image.

---
 - ./unix/native/libjli/java_md_common.c

SVM does not rely on class loaders for a native-image. This part of the code should not be reachable. If it were somehow reached,
the referenced function `JVM_FindClassFromBootLoader` would exit fatally in our JVM implementation (JvmFuncsFallbacks.c):
```c
JNIEXPORT jobject JNICALL JVM_FindClassFromBootLoader(JNIEnv *env) {
    (*env)->FatalError(env, "JVM_FindClassFromBootLoader called:  Unimplemented");
    return NULL;
}
```

---
 - ./unix/native/libjli/java_md_solinux.c

This file appears to only be used by the `java` launcher, which SVM also does not rely on.

---
 - ./unix/native/libnet/DefaultProxySelector.c

This file appears to only be used for Darwin by looking at com/oracle/svm/hosted/jdk/JNIRegistrationJava.java:165:
```java
if (isDarwin()) {
            List<Method> darwinMethods = Arrays.asList(
                            method(a, "apple.security.KeychainStore", "_scanKeychain"),
                            method(a, "apple.security.KeychainStore", "_releaseKeychainItemRef", long.class),
                            method(a, "apple.security.KeychainStore", "_addItemToKeychain", String.class, boolean.class, byte[].class, char[].class),
                            method(a, "apple.security.KeychainStore", "_removeItemFromKeychain", long.class),
                            method(a, "apple.security.KeychainStore", "_getEncodedKeyData", long.class, char[].class));
            if (JavaVersionUtil.JAVA_SPEC >= 11) {
                /*
                 * JNI method implementations depending on CoreService are present in the following
                 * jdk classes sun.nio.fs.MacOXFileSystemProvider (9+),
                 * sun.net.spi.DefaultProxySelector (9+)
                 */
                ArrayList<Method> methods = new ArrayList<>(darwinMethods);
                methods.addAll(Arrays.asList(method(a, "sun.nio.fs.MacOSXFileSystemProvider", "getFileTypeDetector"),
                                method(a, "sun.net.spi.DefaultProxySelector", "getSystemProxies", String.class, String.class),
                                method(a, "sun.net.spi.DefaultProxySelector", "init")));

                a.registerReachabilityHandler(CORESERVICES_LINKER, methods.toArray(new Object[]{}));
            } else {
                a.registerReachabilityHandler(CORESERVICES_LINKER, darwinMethods.toArray(new Object[]{}));
            }
        }
```
---
 - ./unix/native/libnet/net_util_md.c

This file was covered by the linked IPv6 bug report and the above PR.

---
 - ./unix/native/libjava/jdk_util_md.c
 This contains a JDK_FindJvmEntry function:

```c
void* JDK_FindJvmEntry(const char* name) {
    return dlsym(RTLD_DEFAULT, name);
}
```
This function does not appear to be used anywhere.

I will post an update on the (Unix|Linux)NativeDispatcher as soon as I create an example.
Thank you for pointing me this way! If this does not work, it would be an issue for static native-images.

gradinac:
@bobvandette 
As mentioned above, the two problematic files appeared to be:
 - ./linux/native/libnio/fs/LinuxNativeDispatcher.c
 - ./unix/native/libnio/fs/UnixNativeDispatcher.c

The issues they cause are a bit subtle. Here is an example program that will use these under the hood:
`$ cat DispatcherTest.java`
```java
import java.io.IOException;
import java.nio.file.DirectoryStream;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.Paths;
import java.nio.file.SecureDirectoryStream;

public class DispatcherTest {

    public static void main(String[] args) throws IOException {
        Files.createDirectories(Paths.get("foo"));
        Files.createFile(Paths.get("foo/a"));
        Files.createFile(Paths.get("foo/b"));
        Files.createDirectory(Paths.get("foo/c"));
        Files.createFile(Paths.get("foo/c/d"));
        Files.createFile(Paths.get("foo/c/e"));

        try (DirectoryStream<Path> stream = Files.newDirectoryStream(Paths.get("foo"))) {
            System.out.println(stream.getClass().getName());
            stream.forEach(e -> System.out.println(e));
        }
    }
}
```
Compile the program and build three native-images:
 - a dynamically linked native-image
 - a statically linked native-image, linked against glibc
 - a statically linked native-image, linked against musl
```bash
$ javac DispatcherTest.java
$ native-image -H:Name=dispatchertest_dynamic DispatcherTest
[dispatchertest_dynamic:29146]    classlist:   1,410.72 ms,  0.96 GB
[dispatchertest_dynamic:29146]        (cap):     478.89 ms,  0.96 GB
[dispatchertest_dynamic:29146]        setup:   1,527.40 ms,  0.96 GB
[dispatchertest_dynamic:29146]     (clinit):     147.70 ms,  1.21 GB
[dispatchertest_dynamic:29146]   (typeflow):   3,418.74 ms,  1.21 GB
[dispatchertest_dynamic:29146]    (objects):   3,152.54 ms,  1.21 GB
[dispatchertest_dynamic:29146]   (features):     166.22 ms,  1.21 GB
[dispatchertest_dynamic:29146]     analysis:   7,031.93 ms,  1.21 GB
[dispatchertest_dynamic:29146]     universe:     220.27 ms,  1.21 GB
[dispatchertest_dynamic:29146]      (parse):     729.50 ms,  1.69 GB
[dispatchertest_dynamic:29146]     (inline):   1,544.70 ms,  1.69 GB
[dispatchertest_dynamic:29146]    (compile):   5,237.63 ms,  2.31 GB
[dispatchertest_dynamic:29146]      compile:   7,894.18 ms,  2.31 GB
[dispatchertest_dynamic:29146]        image:   1,101.07 ms,  2.31 GB
[dispatchertest_dynamic:29146]        write:     170.99 ms,  2.31 GB
[dispatchertest_dynamic:29146]      [total]:  19,525.55 ms,  2.31 GB
$ native-image -H:Name=dispatchertest_static_glibc DispatcherTest --static
[dispatchertest_static_glibc:29441]    classlist:   1,425.79 ms,  0.96 GB
[dispatchertest_static_glibc:29441]        (cap):     458.75 ms,  0.96 GB
[dispatchertest_static_glibc:29441]        setup:   1,407.80 ms,  0.96 GB
[dispatchertest_static_glibc:29441]     (clinit):     143.72 ms,  1.22 GB
[dispatchertest_static_glibc:29441]   (typeflow):   3,320.96 ms,  1.22 GB
[dispatchertest_static_glibc:29441]    (objects):   3,208.59 ms,  1.22 GB
[dispatchertest_static_glibc:29441]   (features):     194.13 ms,  1.22 GB
[dispatchertest_static_glibc:29441]     analysis:   7,078.00 ms,  1.22 GB
[dispatchertest_static_glibc:29441]     universe:     237.84 ms,  1.22 GB
[dispatchertest_static_glibc:29441]      (parse):     625.68 ms,  1.22 GB
[dispatchertest_static_glibc:29441]     (inline):   1,039.58 ms,  1.67 GB
[dispatchertest_static_glibc:29441]    (compile):   5,115.07 ms,  2.28 GB
[dispatchertest_static_glibc:29441]      compile:   7,177.93 ms,  2.28 GB
[dispatchertest_static_glibc:29441]        image:     999.39 ms,  2.28 GB
[dispatchertest_static_glibc:29441]        write:     230.61 ms,  2.28 GB
[dispatchertest_static_glibc:29441]      [total]:  18,801.81 ms,  2.28 GB
$ native-image -H:Name=dispatchertest_static_muslc DispatcherTest --static -H:UseMuslC=bundle/
[dispatchertest_static_muslc:29645]    classlist:   1,472.22 ms,  0.96 GB
[dispatchertest_static_muslc:29645]        (cap):     417.77 ms,  0.96 GB
[dispatchertest_static_muslc:29645]        setup:   1,420.21 ms,  0.96 GB
[dispatchertest_static_muslc:29645]     (clinit):     164.23 ms,  1.21 GB
[dispatchertest_static_muslc:29645]   (typeflow):   3,516.51 ms,  1.21 GB
[dispatchertest_static_muslc:29645]    (objects):   3,279.18 ms,  1.21 GB
[dispatchertest_static_muslc:29645]   (features):     197.01 ms,  1.21 GB
[dispatchertest_static_muslc:29645]     analysis:   7,325.19 ms,  1.21 GB
[dispatchertest_static_muslc:29645]     universe:     230.16 ms,  1.21 GB
[dispatchertest_static_muslc:29645]      (parse):     722.85 ms,  1.68 GB
[dispatchertest_static_muslc:29645]     (inline):     951.77 ms,  1.68 GB
[dispatchertest_static_muslc:29645]    (compile):   5,852.68 ms,  2.30 GB
[dispatchertest_static_muslc:29645]      compile:   8,092.48 ms,  2.30 GB
[dispatchertest_static_muslc:29645]        image:   1,256.85 ms,  2.30 GB
[dispatchertest_static_muslc:29645]        write:     183.83 ms,  2.30 GB
[dispatchertest_static_muslc:29645]      [total]:  20,230.90 ms,  2.30 GB
```
By running each example, we get:
```bash
$ java DispatcherTest 
sun.nio.fs.UnixSecureDirectoryStream
foo/b
foo/a
foo/c
$ rm -rf foo
$ ./dispatchertest_dynamic 
sun.nio.fs.UnixSecureDirectoryStream
foo/b
foo/a
foo/c
$ rm -rf foo
$ ./dispatchertest_static_glibc 
sun.nio.fs.UnixDirectoryStream
foo/b
foo/a
foo/c
$ rm -rf foo
$ ./dispatchertest_static_muslc 
sun.nio.fs.UnixDirectoryStream
foo/b
foo/a
foo/c
```
A note for the static image with glibc: this was compiled under Ubuntu 18.04.
Although the functionality remains the same, the underlying types that implement it change from a `UnixSecureDirectoryStream` to `UnixDirectoryStream` as the JDK has (wrongfully) not detected support for some system calls. We can see these capabilities get set:
```c
    if (my_openat64_func != NULL &&  my_fstatat64_func != NULL &&
        my_unlinkat_func != NULL && my_renameat_func != NULL &&
        my_futimesat_func != NULL && my_fdopendir_func != NULL)
    {
        capabilities |= sun_nio_fs_UnixNativeDispatcher_SUPPORTS_OPENAT;
    }
```
As can be seen from above, it affects both glibc and muslc. It could also affect other libcs, although I have not tested it. As the issue stems from `dlsym`, we should fix this once and have the fix affect every used glibc. I have two ideas that come to mind:
 1. Make native-image specific changes to the JDK static libraries. We provide a static version of all static libraries. The same static libraries are used for both dynamic and static native-images. If we wanted to make the choice, `dlopen` or just linking the symbol,  a compile time constant, we would have to provide two sets of JDK libraries, one for static native-images, and one for dynamic. We could also take the approach from this PR and configure them using a symbol provided by native-image.
 2. Provide our own implementation for `dlfcn.h` which would handle the logic of getting symbol addresses inside a static native-image. For known symbols, like the ones used in `UnixNativeDispatcher`, we would return a compile time constant. For others, we would either throw a fatal error or simply return a symbol not found. I assume we could force the usage of our implementation simply by passing it first during linking.

@olpaw @bobvandette What do you think?

gradinac:
Static linking against muslc-1.20 yields the same result:
```bash
$ ./dispatchertest_static_muslc_newest
sun.nio.fs.UnixDirectoryStream
foo/b
foo/a
foo/c
```

bobvandette:
Well the good news is that both files LinuxNativeDispatcher and UnixNativeDispatcher are using dlsym to look up functions for "optional" behavior.  The extended attributes method for LinuxNativeDispatcher is not even available on multiple platforms and the openAt support appears to be optional as well.  It's pretty difficult to alter the JDK sources since it would be best to make these changes upstream so they don't get lost when merging into the labsjdk.  The other problem is that we'd have to back port these changes to JDK8, 11 after integrating them into the mainline (JDK15).  If I were to do a change, I'd suggest that the JDK gets changed to not use dlsym but instead use the JVM_FindLibraryEntry so you can override it's behavior.

bobvandette:
Just to be complete, could you write a small "C" program that calls dlsym looking for one of these missing functions and link it statically to confirm that dlsym actually fails.  I just want to make sure there isn't something going on during native-image generation that is stripping symbols.  There's definitely some logic in native-image that attempts to reduce the size of the symbol table.  You should also dump the symbols in your generated image to make sure there's a global symbol in the image.

gradinac:
I agree with you, I will make a small C program that will test this with various options and on different libcs and post an update here.

gradinac:
@bobvandette I have created a small dockerized test to show the behaviour of dlsym. Here's the repo: https://github.com/gradinac/dlsym-static-test
It consists of a small C program that uses `dlsym` to look up fprintf. fprintf is also used directly in the program to force a symbol dependency. The source is compiled on four different platforms, here are the results:
```bash
$ ./run-tests.sh
Running test on Ubuntu 18.04
Hello World!
fprintf called from a symbol obtained using dlsym
-------------------------
Hello World!
Symbol fprintf not found!
Running test on Oracle Linux 7.7
Hello World!
fprintf called from a symbol obtained using dlsym
-------------------------
Hello World!
Symbol fprintf not found!
Running test on Oracle Linux 6.10
Hello World!
fprintf called from a symbol obtained using dlsym
-------------------------
./compile_and_run_tests.sh: line 8:    20 Segmentation fault      (core dumped) ./main-static
Running test on Alpine Linux 3.11
Hello World!
fprintf called from a symbol obtained using dlsym
-------------------------
Hello World!
Symbol fprintf not found!
```
On each platform, dlsym works fine in a dynamic binary.
On Ubuntu 18.04, Oracle Linux 7.7 and Alpine 3.11, dlsym does not locate the symbol.
On Oracle Linux 6.10, dlsym crashes.

drocsid:
Will you guys eventually fix IPV6?

gradinac:
Hey @drocsid! Sorry for the long wait!
We will merge the fix for this today and we will backport it to 20.2 so that it is available in the upcoming release

gradinac:
The fix has been merged and backported to 20.2

PR #2414, opened 2020-04-29
Prevent exception from being swallowed in maven-plugin

bdemers:
Adds root cause to MojoExecutionException when an IllegalAccessError is thrown


bdemers:
I'm trying to track down an error on windows, when the native-image-maven-plugin is used, but the root cause is ignored

olpaw:
@bdemers your change looks fine. Please sign the OCA so that I can merge your change. Thanks.

bdemers:
@olpaw, I sent it in last night, let me known if it doesn’t show up in the system!

olpaw:
> @olpaw, I sent it in last night, let me known if it doesn’t show up in the system!

@dougxc how long does it usually take nowadays to get graalvmbot to see submitted OCAs? Is that process still manual?

dougxc:
Yep, still manual. First we need to see the OCA show up [here](http://www.oracle.com/technetwork/community/oca-486395.html). Then we need to edit the OCA database used by @graalvmbot .

bdemers:
@olpaw @dougxc I'm listed on the OCA page now 😄

Issue #2381, opened 2020-04-23
Add callee-based filters to the tracing agent


sdeleuze:
As [pointed out](https://graalvm.slack.com/archives/CN9KSFB40/p1587030771130700?thread_ts=1587026997.128900&cid=CN9KSFB40) by @shelajev on GraalVM Slack, the tracing agent currently only support  [caller-based filters](https://github.com/oracle/graal/blob/master/substratevm/CONFIGURE.md#caller-based-filters).

This issue is a feature request to add callee-based filters in order to be able to leverage GraalVM and Surefire Maven plugins to generate GraalVM configuration files (reflection, proxy and resource ones) via Maven configuration. We would like to leverage that in the next release of [Spring GraalVM native](https://github.com/spring-projects-experimental/spring-graal-native) support (agent and hybrid modes).

Few notes:
 - Since the goal is to use only Maven configuration, using bash based filtering is not an option hence the need for such feature
 - The support would be pretty similar to [caller-based filters](https://github.com/oracle/graal/blob/master/substratevm/CONFIGURE.md#caller-based-filters) currently supported
 - Such calle-based filters would be typically used to filter test related packages (or classes) like `org.junit.**`, `org.mockito.**` or `org.springframework.test.**`.

Thanks to tag this issue with the `spring` label.

peter-hofer:
@sdeleuze , @sbrannen is #2422 what you had in mind?

sdeleuze:
@peter-hofer Thanks, yes it looks like what I had in mind. Feel free to rename this issue and/or close it if needed.

sbrannen:
Yes, I think that looks good. Thanks!

PR #2422, opened 2020-04-30
Introduce configurable access filter in agent.

peter-hofer:
Solves #2381. Called _access filter_ in contrast to the proposed _callee filter_ because in agent terms, we don't observe _calls_ to these methods, we observe _lookups_ or more generally, _accesses_.


substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/AccessAdvisor.java
```diff
@@ -67,16 +67,18 @@ public static RuleNode copyBuiltinCallerFilterTree() {
         return internalCallerFilter.copy();
     }
 
+    public static RuleNode copyBuiltinAccessFilterTree() {
+        RuleNode root = RuleNode.createRoot();
+        root.addOrGetChildren("**", RuleNode.Inclusion.Include);
```
gradinac, line 72: A small question - could this also work like `copyBuiltinCallerFilterTree`?  It would copy `accessWithoutCallerFilter`, and then we could remove
```
        if (callerClass.get() == null && !accessWithoutCallerFilter.treeIncludes(queriedClass.get())) {
            return true;
        }
```
from the `shouldIgnore` call?

sdeleuze:
Could you please consider inclusion in 20.1 of this PR in order to allow Spring GraalVM support to move forward on running the agent during tests within a pure Maven build?

peter-hofer:
> Could you please consider inclusion in 20.1 of this PR in order to allow Spring GraalVM support to move forward on running the agent during tests within a pure Maven build?

I think it's low-risk, so I'll try to get it merged.

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/AccessAdvisor.java
peter-hofer: I'm not sure if I understand correctly. It would probably be fine to merge the two filters, but `accessWithoutCallerFilter` is specifically for cases in which we don't know the caller class and it's very likely an internal access (JVMCI, Libgraal, JVM startup) where we assume it shouldn't end up in the user configuration, so I'd prefer to keep it separate.

gradinac: Ahh, I overlooked the purpose initially, my bad. I agree, it's best to keep it separate.

sdeleuze:
@peter-hofer First thanks for shipping that in 20.1. This feature works as expected for reflection, but for proxies I can't obtain the behavior I would like.

With following `access-filter.json` file:
```
{ "rules": [
  {"excludeClasses": "org.apache.maven.surefire.**"},
  {"excludeClasses": "net.bytebuddy.**"},
  {"excludeClasses": "org.apiguardian.**"},
  {"excludeClasses": "org.junit.**"},
  {"excludeClasses": "org.mockito.**"},
  {"excludeClasses": "org.springframework.test.**"},
  {"excludeClasses": "org.springframework.boot.test.**"}
]
}
```

The related entries are filtered from `reflect-config.json` but `proxy-config.json` contains:
```
[
  ["java.lang.reflect.ParameterizedType","org.springframework.core.SerializableTypeWrapper$SerializableTypeProxy","java.io.Serializable"],
  ["org.springframework.boot.context.properties.ConfigurationProperties","org.springframework.core.annotation.SynthesizedAnnotation"],
  ["org.springframework.boot.test.context.SpringBootTest","org.springframework.core.annotation.SynthesizedAnnotation"],
  ["org.springframework.test.context.BootstrapWith","org.springframework.core.annotation.SynthesizedAnnotation"]
]
```

Is there currently a way to filter `proxy-config.json` to get the following content based on previously mentioned `access-filter.json`?

```
[
  ["java.lang.reflect.ParameterizedType","org.springframework.core.SerializableTypeWrapper$SerializableTypeProxy","java.io.Serializable"],
  ["org.springframework.boot.context.properties.ConfigurationProperties","org.springframework.core.annotation.SynthesizedAnnotation"]
]
```

If not, should I create a related issue?

peter-hofer:
@sdeleuze , I checked and proxy classes are currently filtered only by caller, not with the new access filter. I created #2490 based on your comment, thank you!

PR #2423, opened 2020-05-01
Tweak compiler assertion

fniephaus:
AFAICT, there's no reason to use `#compilationConstant` here. `#partialEvaluationConstant` is to be preferred.
See https://github.com/oracle/graal/blob/8fb8fc502411b3c84308ff5cda4d6c1f775f3920/truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerAsserts.java#L102-L113


fniephaus:
>The corresponding code in LLVMLoopDispatchNode also needs to be changed. I hope you don't mind that I'll just fix that when integrating. No need for another iteration for such a simple thing.

Sure, please go ahead! Only opened the PR, so we don't forget to change this.

> Out of curiosity: Did you see any actual problem resulting from this? Or how did you stumble over that?

No, I didn't see a technical problem. However, we documented how bytecode interpreters can be implemented in Truffle a while ago (see p.3 of [this preprint](https://fniephaus.com/2018/icooolps18-graalsqueak.pdf)), and copied this line. At that time, the difference between the two APIs were not clear to us. In the meantime, we helped to improve the Javadoc for both assertions. This is just the last missing change that we wanted to push for a while.

PR #2427, opened 2020-05-03
Support MaxDirectMemorySize

christianwimmer:
Do not inherit the maximum direct memory size from image generator, and provide same runtime option `-XX:MaxDirectMemorySize` as the Java HotSpot VM.


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heap/Target_java_lang_ref_Reference.java
```diff
@@ -136,7 +136,25 @@
     @TargetElement(onlyWith = JDK8OrEarlier.class)
     @SuppressWarnings("unused")
     static boolean tryHandlePending(boolean waitForNotify) {
-        throw VMError.unimplemented();
```
christianwimmer, line 139: @peter-hofer We need to support this method, since it is reachable from `ByteBuffer.allocateDirect` and invoked before running out of native memory.

If you see an easy way to implement the "real" JDK 8 semantics of helping with reference processing, please do it. Otherwise I think this substitution should be good enough.

Issue #2431, opened 2020-05-04
Native image debug line info wrongly attributed when using gdb 9.1


adinn:
gdb 9.1 appears to be associating method code addresses with the wrong line numbers when presented with a native image containing debug info. An objdump of the binary indicates that the debug_lineinfo DWARF data has the correct line numbers. Also, the mismatch does not occur when debugging the same generated image with gdb 8.3. So, this may be a bug in gdb.

1. Clone latest graalvm repo and mx build
2. Write a simple Hello World program in Hello.java
3. Compile it `javac Hello.java`
4. Generate a native image with debug info  `mx native-image -J-ea -J-esa  -H:GenerateDebugInfo=1 Hello`
5. Start the debugger `gdb hello`
6. Set up the directories path `(gdb) set directories=sources/src:sources/graal:sources/jdk`
7. Break main `(gdb) break Hello::main` Note that a breakpoint address is printed but the breakpoint file or line is omitted (whereas it is correctly printed using gdb 8.3)
8. Run to breakpoint `(gdb) run` Note that when the breakpoint is it the breakpoint file and line are printed but the line number is incorrect and the listed line is incorrect (the line is correctly printed using gdb 8.3)

GraalVM version: latest head
commit 7062562ab88e167aecf647b3f36cbf5407952b26 (HEAD -> master, upstream/master, origin/master, origin/HEAD)
Merge: 1a74b765c01 366f8294fd0
Author: Andreas Woess <andreas.woess@oracle.com>
Date:   Tue Apr 28 12:41:16 2020 +0000

    [GR-22978] Intrinsify MethodHandleImpl.isCompileConstant.
    
    PullRequest: graal/6087

JDK version:
openjdk version "11.0.7" 2020-04-14
OpenJDK Runtime Environment (build 11.0.7+10-jvmci-20.1-b02)
OpenJDK 64-Bit Server VM (build 11.0.7+10-jvmci-20.1-b02, mixed mode)

gdb version:
GNU gdb (GDB) Fedora 9.1-3.fc32

adinn:
Yes, it should be closed.

PR #2433, opened 2020-05-04
correct computation of prologue size

adinn:
This fixes the problem found when using gdb 9.1 that was raised in issue #2431

The computation of the value for the prologue size field was incorrectly including that field in the byte count for remaining header + dir & file tables, meaning it was identifying a point 4 bytes into the line state machine instructions. With gdb 9.1 the debugger adds the prologue size to the prologue offset to identify the start of those instructions. With 8.3 it simply starts reading instructions from the point where the header + dir & file tables end which is why it went undetected.

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLineSectionImpl.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLineSectionImpl.java
@@ -320,7 +320,7 @@ public class DwarfLineSectionImpl extends DwarfSectionImpl {
         /*
          * 4 ubyte prologue length includes rest of header and dir + file table section.
          */
-        int prologueSize = classEntry.getLinePrologueSize() - 6;
+        int prologueSize = classEntry.getLinePrologueSize() - (4 + 2 + 4);
         pos = putInt(prologueSize, buffer, pos);
         /*
          * 1 ubyte min instruction length is always 1.


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLineSectionImpl.java
```diff
@@ -320,7 +320,7 @@ private int writeHeader(ClassEntry classEntry, byte[] buffer, int p) {
         /*
          * 4 ubyte prologue length includes rest of header and dir + file table section.
          */
-        int prologueSize = classEntry.getLinePrologueSize() - 6;
+        int prologueSize = classEntry.getLinePrologueSize() - 10;
```
olpaw, line 323: So the 10 here is from 4 (length field)  + 2 (version field) + 4 (prologue length field), right?

adinn: Yes, that's right and originally I failed to exclude the 4 byte slot for for the prologue field from the prologue length.

I thought about using a private static final field for this value here to allow the above to be documented in some javadoc. However, I thought that would be overkill as it was fairly obvious at this point that we had just written 10 bytes.

I also thought about asserting at the end of writing the header that the sum of the header size, file table size and directory size added up to the prologue field value + 10. However, that would involve reading back an int out of the byte array at offset 6 in order to validate what was fairly clearly written using that same computation. Once again that seemed to be overkill (so far there has been no need for any methods to read structured data from the byte[] so adding one to implement an assert seemed to be a step too far).

If you want I couled revise either or both those decisions?

olpaw: Nah ... just use `- (4 + 2 + 4)` instead of `-10`.

adinn: Done.

PR #2323, opened 2020-04-07
Support dynamic class loading and serialization

ziyilin:
This patch provides supports for dynamic class loading and serialization/deserialization features for substratevm.
**Overview**
The basic idea for supporting dynamic class loading is dumping classes generated at run time in a beforehand run with native-image-agent to a specified directory which is added to the classpath to build native-image. Serialization/deserialization is basically dynamic class loading + reflection.
This is implemented in 3 steps:

1. Intercept java.lang.ClassLoader.postDefineClass and sun.reflect.ClassDefiner.defineClass by native-image-agent to dump the dynamically generated class to file system.
2. Make sure the generated classes' names are fixed through Hotspot version runtime and native-image version runtime.
3. Substitute relevant classes in SubstrateVM to load the previously dumped classes.

**Features**:

- Add new agent option `-agentlib:native-image-agent=dynmaic-class-dump-dir=` to specify where to dump the dynamically generated classs, the default value is _"dynClass"_ if not specified.
- Both dynamically generated class and its stack trace are dumped.
- Dynamically generated class's name could be decided at runtime (e.g. runtime sequence number as postfix) or null when defineClass is invoked. The former is supported by using same rule to generate fixed names for both Agent runtime and native-image runtime. The latter is supported by retrieving class name from dumped class bytecode at native-image runtime.
- Support JDK serialization/deserialization which replies on dynamic class loading and reflection.

**Limitations**:

1. Don't support serializing proxied class. The proxied class is generated and cached in native-image at build time. The generated class name is profixed with a sequence number which could be different from Agent run and native-image build time run.
2. It is possible the jar on classpath has a different signature file from dynamically generated class and fail the check in `java.lang.ClassLoader.checkCerts(String name, CodeSource cs)` at native-image build time. Current solution is to delete jar's signature file before building.
3. Warning messages such as "WARNING: Method java.lang.Object.<clinit>() not found." will be reported at native-image build time. Because such method has been accessed via JNI calls at serialization time to calculate serializeVersionUID and the Agent has intercepted and recorded them.

**Tests**

We have prepared the `serial` test cases in SPECjvm2008 to test this patch. Simply unzip [specjvm2008-for-serial.zip](https://github.com/oracle/graal/files/4443123/specjvm2008-for-serial.zip), change $GRAALVM_HOME to your GraalVM home and run 

> ./SVMBenchmarks/serial.sh -cr

The script builds and runs `spec.benchmarks.serial.Main` in SPECjvm2008.
[...]

PR #2442, opened 2020-05-08
Support dynamic class loading

ziyilin:
**Overview**
This patch supports `java.lang.ClassLoader.defineClass(String name, byte[] b, int off, int len)` based dynamic class loading in native image. It is separated from the previous PR (https://github.com/oracle/graal/pull/2323). 
The basic idea for supporting dynamic class loading is dumping classes generated at run time in a beforehand run with native-image-agent to a specified directory which is added to the classpath for building native image. 
This is implemented in 3 steps:

1. Intercept `java.lang.ClassLoader.postDefineClass` by native-image-agent to dump the dynamically generated class to file system.
2. User should make sure the generated classes' names are fixed through Hotspot version runtime and native-image version runtime.
3. Substitute relevant classes in SubstrateVM to load the previously dumped classes at native image runtime.

**Features:**

- Add new agent option `-agentlib:native-image-agent=dynamic-class-dump-dir=` to specify where to dump the dynamically generated classes, the default value is "dynClass" if not specified.
- Both dynamically generated class and its stack trace are dumped.
- Dynamically generated class's name could be decided at runtime (e.g. runtime sequence number as post-fix) or null when `java.lang.ClassLoader.defineClass` is invoked. It is the user's job to make sure the dynamic generated class' name is constant through Agent runtime and native image runtime. The patch has supported the null name scenario.
- Different classloaders can dynamically define same name class at runtime. SubstrateVM doesn't support multiple classloaders yet. This patch will report error at Agent runtime if same name classes  are defined by multiple classloaders.
- All unsupported dynamic class loading usages shall be reported at the end of Agent run.

**Limitations:**
- This patch does not support defineClass with specified ProtectionDomain.
- It is possible the jar on classpath has a different signature file from dynamically generated class and fail the check in `java.lang.ClassLoader.checkCerts(String name, CodeSource cs)` at native-image build time. Current solution is to manually delete jar's signature file before building. This limitation can be shown by test: [testWithSignedJar.zip](https://github.com/oracle/graal/files/4770502/testWithSignedJar.zip)

- It is possible the dynamically generated class contents are changing from run to run. We make a class contents check at runtime based on the byte array's hash code to assure the runtime defined class is the same as the previously prepared one. But this check excludes SourceFile attribute because it often varies from runs but doesn't affect runtime behavior. This limitation can be shown by test: [testChangedClass.zip](https://github.com/oracle/graal/files/4770579/testChangedClass.zip)

**Tests:**
This patch can be tested by 
[testDynamicLoading.zip](https://github.com/oracle/graal/files/4626450/testDynamicLoading.zip)

### DIFF
--- a/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
+++ b/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
@@ -27,6 +27,7 @@ package com.oracle.svm.agent;
 import static com.oracle.svm.core.util.VMError.guarantee;
 import static com.oracle.svm.jni.JNIObjectHandles.nullHandle;
 import static com.oracle.svm.jvmtiagentbase.Support.callObjectMethod;
+import static com.oracle.svm.jvmtiagentbase.Support.callObjectMethodL;
 import static com.oracle.svm.jvmtiagentbase.Support.check;
 import static com.oracle.svm.jvmtiagentbase.Support.checkJni;
 import static com.oracle.svm.jvmtiagentbase.Support.checkNoException;
@@ -50,6 +51,7 @@ import static com.oracle.svm.jvmtiagentbase.Support.toCString;
 import static com.oracle.svm.jvmtiagentbase.jvmti.JvmtiEvent.JVMTI_EVENT_BREAKPOINT;
 import static com.oracle.svm.jvmtiagentbase.jvmti.JvmtiEvent.JVMTI_EVENT_CLASS_PREPARE;
 import static com.oracle.svm.jvmtiagentbase.jvmti.JvmtiEvent.JVMTI_EVENT_NATIVE_METHOD_BIND;
+import static com.oracle.svm.jvmtiagentbase.jvmti.JvmtiEvent.JVMTI_EVENT_CLASS_FILE_LOAD_HOOK;
 import static org.graalvm.word.WordFactory.nullPointer;
 
 import java.nio.ByteBuffer;
@@ -63,6 +65,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.locks.ReentrantLock;
 
+import com.oracle.svm.core.util.JavaClassUtil;
 import org.graalvm.compiler.core.common.NumUtil;
 import org.graalvm.nativeimage.StackValue;
 import org.graalvm.nativeimage.UnmanagedMemory;
@@ -126,7 +129,7 @@ final class BreakpointInterceptor {
     private static NativeImageAgent agent;
 
     private static Map<Long, Breakpoint> installedBreakpoints;
-
+    private static List<String> unsupportedExceptions = new ArrayList<>();
     /**
      * A map from {@link JNIMethodId} to entry point addresses for bound Java {@code native}
      * methods, NOT considering our intercepting functions, i.e., these are the original entry
@@ -1085,6 +1088,52 @@ final class BreakpointInterceptor {
         }
     }
 
+    @CEntryPoint
+    @CEntryPointOptions(prologue = AgentIsolate.Prologue.class)
+    private static void onClassFileLoadHook(@SuppressWarnings("unused") JvmtiEnv jvmti, JNIEnvironment jni,
+                    @SuppressWarnings("unused") JNIObjectHandle classBeingRedefined, JNIObjectHandle loader, CCharPointer name, @SuppressWarnings("unused") JNIObjectHandle protectionDomain,
+                    int classDataLen,
+                    CCharPointer classData, @SuppressWarnings("unused") CIntPointer newClassDataLen, @SuppressWarnings("unused") CCharPointerPointer newClassData) {
+        boolean nameIsNull = name.isNull();
+        if (isDynamicallyGenerated(jni, loader, nameIsNull, nameIsNull ? "" : fromCString(name))) {
+            byte[] contents = new byte[classDataLen];
+            CTypeConversion.asByteBuffer(classData, classDataLen).get(contents);
+            String definedClassName = nameIsNull ? JavaClassUtil.getClassName(contents) : fromCString(name);
+            ClassLoaderDefineClassSupport.trace(traceWriter, contents, definedClassName, true);
+        }
+    }
+
+    private static boolean isDynamicallyGenerated(JNIEnvironment jni, JNIObjectHandle classLoader, boolean inputNameIsNull, String definedClassName) {
+        boolean isDynamicallyGenerated;
+        // 1. Classloader is null, it's a system class.
+        // The class is not dynamically generated.
+        if (classLoader.equal(nullHandle())) {
+            isDynamicallyGenerated = false;
+        } else {
+            // 2. Don't have a name for class before defining.
+            // The class is dynamically generated.
+            if (inputNameIsNull) {
+                isDynamicallyGenerated = true;
+            } else {
+                // 3. A dynamically defined class always return null
+                // when call java.lang.ClassLoader.getResource(classname)
+                // This is the accurate but slow way.
+                String asResourceName = definedClassName.replace('.', '/') + ".class";
+                try (CCharPointerHolder resourceNameHolder = toCString(asResourceName);) {
+                    JNIObjectHandle resourceNameJString = jniFunctions().getNewStringUTF().invoke(jni, resourceNameHolder.get());
+                    if (agent.handles() == null) {
+                        // agent's handles is created at onVMStart.
+                        isDynamicallyGenerated = false;
+                    } else {
+                        JNIObjectHandle returnValue = callObjectMethodL(jni, classLoader, agent.handles().javaLangClassLoaderGetResource, resourceNameJString);
+                        isDynamicallyGenerated = returnValue.equal(nullHandle());
+                    }
+                }
+            }
+        }
+        return isDynamicallyGenerated;
+    }
+
     private static final CEntryPointLiteral<CFunctionPointer> onBreakpointLiteral = CEntryPointLiteral.create(BreakpointInterceptor.class, "onBreakpoint",
                     JvmtiEnv.class, JNIEnvironment.class, JNIObjectHandle.class, JNIMethodId.class, long.class);
 
@@ -1093,6 +1142,10 @@ final class BreakpointInterceptor {
 
     private static final CEntryPointLiteral<CFunctionPointer> onClassPrepareLiteral = CEntryPointLiteral.create(BreakpointInterceptor.class, "onClassPrepare",
                     JvmtiEnv.class, JNIEnvironment.class, JNIObjectHandle.class, JNIObjectHandle.class);
+
+    private static final CEntryPointLiteral<CFunctionPointer> onClassFileLoadHookLiteral = CEntryPointLiteral.create(BreakpointInterceptor.class, "onClassFileLoadHook",
+                    JvmtiEnv.class, JNIEnvironment.class, JNIObjectHandle.class, JNIObjectHandle.class, CCharPointer.class, JNIObjectHandle.class, int.class, CCharPointer.class, CIntPointer.class,
+                    CCharPointerPointer.class);
 
     public static void onLoad(JvmtiEnv jvmti, JvmtiEventCallbacks callbacks, TraceWriter writer, NativeImageAgent nativeImageTracingAgent,
                     boolean exptlClassLoaderSupport) {
@@ -1106,6 +1159,7 @@ final class BreakpointInterceptor {
         capabilities.setCanGenerateBreakpointEvents(1);
         capabilities.setCanAccessLocalVariables(1);
         capabilities.setCanGenerateNativeMethodBindEvents(1);
+        capabilities.setCanGenerateAllClassHookEvents(1);
         if (exptlClassLoaderSupport) {
             capabilities.setCanGetBytecodes(1);
             capabilities.setCanGetConstantPool(1);
@@ -1125,6 +1179,9 @@ final class BreakpointInterceptor {
 
         BreakpointInterceptor.boundNativeMethods = new HashMap<>();
         Support.check(jvmti.getFunctions().SetEventNotificationMode().invoke(jvmti, JvmtiEventMode.JVMTI_ENABLE, JVMTI_EVENT_NATIVE_METHOD_BIND, nullHandle()));
+
+        callbacks.setClassFileLoadHook(onClassFileLoadHookLiteral.getFunctionPointer());
+        Support.check(jvmti.getFunctions().SetEventNotificationMode().invoke(jvmti, JvmtiEventMode.JVMTI_ENABLE, JVMTI_EVENT_CLASS_FILE_LOAD_HOOK, nullHandle()));
     }
 
     public static void onVMInit(JvmtiEnv jvmti, JNIEnvironment jni) {
@@ -1266,6 +1323,19 @@ final class BreakpointInterceptor {
             } finally {
                 recursive.set(false);
             }
+        }
+    }
+
+    public static void reportExceptions() {
+        if (!unsupportedExceptions.isEmpty()) {
+            System.err.println(unsupportedExceptions.size() + " unsupported features are detected ");
+            StringBuilder errorMsg = new StringBuilder();
+            for (int i = 0; i < unsupportedExceptions.size(); i++) {
+                errorMsg.append(unsupportedExceptions.get(i)).append("\n");
+            }
+            throw new UnsupportedOperationException(errorMsg.toString());
+        } else {
+            unsupportedExceptions = null;
         }
     }
 
--- a/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/ClassLoaderDefineClassSupport.java
+++ b/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/ClassLoaderDefineClassSupport.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2021, Alibaba Group Holding Limited. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.agent;
+
+import com.oracle.svm.core.util.JavaClassUtil;
+import com.oracle.svm.jni.nativeapi.JNIEnvironment;
+
+import java.security.NoSuchAlgorithmException;
+
+import static com.oracle.svm.jvmtiagentbase.Support.getMethodFullNameAtFrame;
+
+/**
+ * Support dynamic class loading that is implemented by java.lang.ClassLoader.defineClass.
+ */
+public class ClassLoaderDefineClassSupport {
+
+    private static String calculateGeneratedClassSHA(byte[] values) {
+        String generatedClassHashCode;
+        try {
+            generatedClassHashCode = JavaClassUtil.getSHAWithoutSourceFileInfo(values);
+        } catch (NoSuchAlgorithmException e) {
+            generatedClassHashCode = null;
+        }
+        return generatedClassHashCode;
+    }
+
+    public static void trace(TraceWriter traceWriter, byte[] classContents, String generatedClassName, Object result) {
+        assert classContents != null;
+        if (generatedClassName != null && result != null) {
+            // Trace dynamically generated class in config file
+            traceWriter.traceCall("classDefiner", "onClassFileLoadHook", null, null, null, result, generatedClassName.replace('/', '.'), calculateGeneratedClassSHA(classContents), classContents);
+        }
+    }
+
+    public static StringBuilder getStackTrace(JNIEnvironment jni) {
+        StringBuilder trace = new StringBuilder();
+        int i = 0;
+        int maxDepth = 20;
+        while (i < maxDepth) {
+            String methodName = getMethodFullNameAtFrame(jni, i++);
+            if (methodName == null) {
+                break;
+            }
+            trace.append("    ").append(methodName).append("\n");
+        }
+        if (i >= maxDepth) {
+            trace.append("    ").append("...").append("\n");
+        }
+        return trace;
+    }
+}
--- a/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageAgentJNIHandleSet.java
+++ b/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageAgentJNIHandleSet.java
@@ -39,17 +39,15 @@ public class NativeImageAgentJNIHandleSet extends JNIHandleSet {
     final JNIObjectHandle javaLangClass;
     final JNIMethodId javaLangClassForName3;
     final JNIMethodId javaUtilEnumerationNextElement;
+    final JNIMethodId javaLangReflectMemberGetName;
+    final JNIMethodId javaLangReflectMemberGetDeclaringClass;
+    final JNIMethodId javaUtilEnumerationHasMoreElements;
+    final JNIMethodId javaLangClassLoaderGetResource;
+    final JNIObjectHandle javaLangClassLoader;
     final JNIMethodId javaLangClassGetDeclaredMethod;
     final JNIMethodId javaLangClassGetDeclaredConstructor;
     final JNIMethodId javaLangClassGetDeclaredField;
     final JNIMethodId javaLangClassGetName;
-
-    final JNIMethodId javaLangReflectMemberGetName;
-    final JNIMethodId javaLangReflectMemberGetDeclaringClass;
-
-    final JNIMethodId javaUtilEnumerationHasMoreElements;
-
-    final JNIObjectHandle javaLangClassLoader;
 
     final JNIMethodId javaLangObjectGetClass;
 
@@ -78,6 +76,7 @@ public class NativeImageAgentJNIHandleSet extends JNIHandleSet {
         javaLangClassGetDeclaredField = getMethodId(env, javaLangClass, "getDeclaredField", "(Ljava/lang/String;)Ljava/lang/reflect/Field;", false);
         javaLangClassGetName = getMethodId(env, javaLangClass, "getName", "()Ljava/lang/String;", false);
 
+        javaLangClassLoaderGetResource = getMethodId(env, findClass(env, "java/lang/ClassLoader"), "getResource", "(Ljava/lang/String;)Ljava/net/URL;", false);
         JNIObjectHandle javaLangReflectMember = findClass(env, "java/lang/reflect/Member");
         javaLangReflectMemberGetName = getMethodId(env, javaLangReflectMember, "getName", "()Ljava/lang/String;", false);
         javaLangReflectMemberGetDeclaringClass = getMethodId(env, javaLangReflectMember, "getDeclaringClass", "()Ljava/lang/Class;", false);
--- a/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/ConfigurationTool.java
+++ b/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/ConfigurationTool.java
@@ -195,6 +195,12 @@ public class ConfigurationTool {
                     set.getSerializationConfigPaths().add(requirePathUri(current, value));
                     break;
 
+                case "--dynamic-classes-input":
+                    set = inputSet; // fall through
+                case "--dynamic-classes-output":
+                    set.getDynamicClassesConfigPaths().add(requirePathUri(current, value));
+                    break;
+
                 case "--trace-input":
                     traceInputs.add(requirePathUri(current, value));
                     break;
@@ -256,7 +262,8 @@ public class ConfigurationTool {
         try {
             p = new TraceProcessor(advisor, inputSet.loadJniConfig(ConfigurationSet.FAIL_ON_EXCEPTION), inputSet.loadReflectConfig(ConfigurationSet.FAIL_ON_EXCEPTION),
                             inputSet.loadProxyConfig(ConfigurationSet.FAIL_ON_EXCEPTION), inputSet.loadResourceConfig(ConfigurationSet.FAIL_ON_EXCEPTION),
-                            inputSet.loadSerializationConfig(ConfigurationSet.FAIL_ON_EXCEPTION));
+                            inputSet.loadSerializationConfig(ConfigurationSet.FAIL_ON_EXCEPTION),
+                            inputSet.loadDynamicClassesConfig(ConfigurationSet.FAIL_ON_EXCEPTION));
         } catch (IOException e) {
             throw e;
         } catch (Throwable t) {
@@ -297,6 +304,11 @@ public class ConfigurationTool {
         for (URI uri : outputSet.getSerializationConfigPaths()) {
             try (JsonWriter writer = new JsonWriter(Paths.get(uri))) {
                 p.getSerializationConfiguration().printJson(writer);
+            }
+        }
+        for (URI uri : outputSet.getDynamicClassesConfigPaths()) {
+            try (JsonWriter writer = new JsonWriter(Paths.get(uri))) {
+                p.getDynamicClassesConfiguration().printJson(writer);
             }
         }
     }
--- a/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationDynamicClass.java
+++ b/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationDynamicClass.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2021, Alibaba Group Holding Limited. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.configure.config;
+
+import com.oracle.svm.configure.json.JsonPrintable;
+import com.oracle.svm.configure.json.JsonWriter;
+import com.oracle.svm.core.configure.ConfigurationFiles;
+
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public final class ConfigurationDynamicClass implements JsonPrintable {
+
+    private final String qualifiedJavaName;
+    private final String checksum;
+    private final Path classFilePath;
+    private final byte[] classContents;
+
+    public static ConfigurationDynamicClass newAgentTraceTimeConfig(String qualifiedJavaName, String checksum, byte[] classContents) {
+        return new ConfigurationDynamicClass(qualifiedJavaName, checksum, classContents, null);
+    }
+
+    public static ConfigurationDynamicClass newAgentMergeTimeConfig(String qualifiedJavaName, String checksum, Path classFile) {
+        return new ConfigurationDynamicClass(qualifiedJavaName, checksum, null, classFile);
+    }
+
+    private ConfigurationDynamicClass(String qualifiedJavaName, String checksum, byte[] classContents, Path classFilePath) {
+        assert qualifiedJavaName.indexOf('/') == -1 : "Requires qualified Java name, not internal representation";
+        assert !qualifiedJavaName.startsWith("[") : "Requires Java source array syntax, for example java.lang.String[]";
+        assert !(classContents == null && classFilePath == null) : "classContents and classFilePath can't be null at the same time";
+        this.qualifiedJavaName = qualifiedJavaName;
+        this.checksum = checksum;
+        this.classContents = classContents;
+        this.classFilePath = classFilePath;
+    }
+
+    public boolean contains(String definedClassName, String checksum2Match) {
+        return definedClassName != null && definedClassName.equals(qualifiedJavaName) && checksum2Match.equals(this.checksum);
+    }
+
+    @Override
+    public void printJson(JsonWriter writer) throws IOException {
+        // Mangle class' qualified name into a system independent format
+        String internalName = qualifiedJavaName.replace(".", ConfigurationFiles.MANGLE_SLASH);
+        Path basePath = writer.getPath().getParent();
+        Path dumpDir = basePath.resolve(ConfigurationFiles.DUMP_CLASSES_DIR);
+        Path dumpFile = dumpDir.resolve(internalName + ".class");
+        checkAndMkDir(dumpDir);
+        if (classContents != null) {
+            try (FileOutputStream stream = new FileOutputStream(dumpFile.toFile());) {
+                stream.write(classContents);
+            }
+        }
+        if (classFilePath != null) {
+            Files.copy(classFilePath, dumpFile);
+        }
+        writer.append('{').newline();
+        writer.quote("name").append(':').quote(qualifiedJavaName).append(',').newline();
+        writer.quote("classFile").append(":").quote(basePath.relativize(dumpFile).toString()).append(',').newline();
+        writer.quote("checksum").append(":").quote(checksum).newline();
+        writer.append("}");
+    }
+
+    private static void checkAndMkDir(Path dumpDirs) throws IOException {
+        if (!Files.exists(dumpDirs)) {
+            Files.createDirectories(dumpDirs);
+        } else if (!Files.isDirectory(dumpDirs)) {
+            throw new IOException("Existing " + dumpDirs + " is a file, but a directory is expected.");
+        }
+    }
+}
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/ClassForNameSupport.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/ClassForNameSupport.java
@@ -41,10 +41,18 @@ public final class ClassForNameSupport {
 
     /** The map used to collect registered classes. */
     private final EconomicMap<String, Class<?>> knownClasses = ImageHeapMap.create();
+    /** Store class name and checksum byte array as key-value pair. */
+    private final EconomicMap<String, byte[]> dynamicGeneratedClasses = ImageHeapMap.create();
 
     @Platforms(Platform.HOSTED_ONLY.class)
     public static void registerClass(Class<?> clazz) {
         singleton().knownClasses.put(clazz.getName(), clazz);
+    }
+
+    @Platforms(Platform.HOSTED_ONLY.class)
+    public static void registerDynamicGeneratedClass(Class<?> generatedClazz, String definedClassName, String checksum) {
+        registerClass(generatedClazz);
+        ImageSingletons.lookup(ClassForNameSupport.class).dynamicGeneratedClasses.put(definedClassName, checksum.getBytes());
     }
 
     public static Class<?> forNameOrNull(String className, boolean initialize) {
@@ -64,6 +72,14 @@ public final class ClassForNameSupport {
             throw new ClassNotFoundException(className);
         }
         return result;
+    }
+
+    public static String getDynamicClassChecksum(String className) throws ClassNotFoundException {
+        byte[] storedValue = ImageSingletons.lookup(ClassForNameSupport.class).dynamicGeneratedClasses.get(className);
+        if (storedValue == null) {
+            throw new ClassNotFoundException(className);
+        }
+        return new String(storedValue);
     }
 }
 
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_lang_ClassLoader.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_lang_ClassLoader.java
@@ -31,15 +31,16 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.security.NoSuchAlgorithmException;
 import java.security.ProtectionDomain;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Vector;
 import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.Enumeration;
+import java.util.Collections;
+import java.util.Vector;
 
 import com.oracle.svm.core.annotate.Alias;
 import com.oracle.svm.core.annotate.Delete;
@@ -50,6 +51,7 @@ import com.oracle.svm.core.annotate.TargetClass;
 import com.oracle.svm.core.annotate.TargetElement;
 import com.oracle.svm.core.hub.ClassForNameSupport;
 import com.oracle.svm.core.hub.DynamicHub;
+import com.oracle.svm.core.util.JavaClassUtil;
 import com.oracle.svm.core.util.VMError;
 
 import jdk.vm.ci.meta.MetaAccessProvider;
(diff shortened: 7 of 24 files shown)


peter-hofer:
Thank you for your contribution @ziyilin ! Before I go into details, I'd like to discuss the overall design (and I'd also appreciate input from @christianwimmer).

The current design, as I understand it, dumps class files for defined classes directly and immediately to a separate directory, bypassing the mechanism for writing trace events and generating configuration files. I understand the tracing approach involves some additional overhead, but it would provide better opportunities for tracking dynamically loaded classes and their metadata (like the hash code) and loading them during the image build. The current approach requires manual changes to the class path and uses a fake `Class.forName` reflection event to carry metadata and trigger loading the class.
What I'm considering is to write a trace event containing the Base64-encoded class data along with any metadata. From this, trace processing can write a JSON file with the class metadata, including its fully-qualified name, which might not be supported as a file name on each platform. This file can be written to the configuration directory so that the configuration is self-contained and users don't have to maintain a separate directory. Considering the class file data, I see three options:
1. Store raw `.class` files in a subdirectory of the configuration directory and use mangled class names or generated identifiers for file names. The class files are referenced from the (single) JSON file.
1. Store a single JSON file for each class in a subdirectory, containing its Base64-encoded class data and metadata, using a mangled class name. This would make it easier to delete individual undesirable classes (and avoid potential issues with binary files in version control, across platforms, etc.).
1. Store all metadata and Base64-encoded class data of dynamically loaded classes in a single JSON file. This would be in line what we currently do for other purposes, but probably be impractically large.
Options 1 and 2 have the advantage that the class data can be immediately written after trace processing rather than staying resident in memory until the application finishes.
The approach in general would keep all information in a single trace stream, and all processed information in a single configuration directory, simplifying the build and retaining the current workflows of users.

ziyilin:
The original design was for quick implementation to support the business requirement. I agree all things should be kept together in a single system. The option 1 seems good to me. In option 1, the json configuration file acts as an index, and the raw classes can be stored in a subdirectory independently. The configuration file is small so it's feasible for human reading and checking, the independently stored class files are also more flexible for future extension.

vjovanov:
Hi @ziyilin, I have made a rebase of your PR to test it out with our benchmarks. It works in all of the cases there. The link to the fixed-up version is here: 

 https://github.com/vjovanov/graal/tree/dynamicClassLoading

I am running the tests now and will keep you posted on the results.

Amazing work!

vjovanov:
All seems to work except a transient (flaky) failure in one of the tests. The reproducer should be, but I can't reproduce it:
```
mx --env ni-ce benchmark scala-dacapo-native-image:scalaxb -- --jvm=native-image --jvm-config=default-ce -Dnative-image.benchmark.stages=agent
```
It doesn't fail on my machine, but in the CI the error message is: 
```
--------- Standard output:
===== DaCapo 0.1.0-SNAPSHOT scalaxb starting =====
===== DaCapo 0.1.0-SNAPSHOT scalaxb PASSED in 18844 msec =====
Fatal error: java.lang.NullPointerException: null
    at com.oracle.svm.agent.AbstractDynamicClassGenerationSupport.<init>(AbstractDynamicClassGenerationSupport.java:97)
    at com.oracle.svm.agent.DynamicDefineClassSupport.<init>(DynamicDefineClassSupport.java:37)
    at com.oracle.svm.agent.AbstractDynamicClassGenerationSupport.getDynamicClassGenerationSupport(AbstractDynamicClassGenerationSupport.java:89)
    at com.oracle.svm.agent.BreakpointInterceptor.postDefineClass(BreakpointInterceptor.java:790)
    at com.oracle.svm.agent.BreakpointInterceptor.onBreakpoint(BreakpointInterceptor.java:1143)
```

I think we need to be defensive here against null values.

ziyilin:
> https://github.com/vjovanov/graal/tree/dynamicClassLoading

I have cloned this repo, but got a mx error:
```
[cengfeng.lzy@e69e13043.et15sqa /home/cengfeng.lzy/codebase/dynamicSVM/vm]
$mx --dynamicimports ./substratevm --java-home=/home/cengfeng.lzy/tools/labsjdk-ce-11.0.8-jvmci-20.2-b03 --disable-libpolyglot --disable-polyglot --force-bash-launchers=true --exclude-components=nju,nil,lg build
Traceback (most recent call last):
  File "/home/cengfeng.lzy/codebase/ajdk_graal/mx/mx.py", line 17180, in <module>
    main()
  File "/home/cengfeng.lzy/codebase/ajdk_graal/mx/mx.py", line 17065, in main
    primary = _discover_suites(primarySuiteMxDir, load=should_load_suites)
  File "/home/cengfeng.lzy/codebase/ajdk_graal/mx/mx.py", line 3506, in _discover_suites
    _register_visit(primary)
  File "/home/cengfeng.lzy/codebase/ajdk_graal/mx/mx.py", line 3502, in _register_visit
    _register_visit(discovered[_suite_import.name])
  File "/home/cengfeng.lzy/codebase/ajdk_graal/mx/mx.py", line 3504, in _register_visit
    s._load()
  File "/home/cengfeng.lzy/codebase/ajdk_graal/mx/mx.py", line 1604, in _load
    self._load_extensions()
  File "/home/cengfeng.lzy/codebase/ajdk_graal/mx/mx.py", line 2026, in _load_extensions
    mod = __import__(extensionsName)
  File "/home/cengfeng.lzy/codebase/dynamicSVM/truffle/mx.truffle/mx_truffle.py", line 56, in <module>
    import tck
  File "/home/cengfeng.lzy/codebase/dynamicSVM/truffle/mx.truffle/tck.py", line 189
    print message.format(args if args else [])
                ^
SyntaxError: invalid syntax
```
And there is no file `ni-ce` neither in the repo.

vjovanov:
@ziyilin try to run the following in the root of the repo:
```
cd vm
mx --env ni-cee build
mx --env ni-ce benchmark scala-dacapo-native-image:scalaxb -- --jvm=native-image --jvm-config=default-ce -Dnative-image.benchmark.stages=agent
```

ziyilin:
@vjovanov 
I cannot reproduce the issue either. This issue suggests that the assumption of dynamically generated class should always have a name is broken. A null check should be performed.

Where should I commit the modification? Still in this PR or your repo?

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    public static AbstractDynamicClassGenerationSupport getClassLoaderDefineClassSupport(JNIEnvironment jni, JNIObjectHandle callerClass,
+                    String generatedClassName, TraceWriter traceWriter, NativeImageAgent agent) throws IOException {
+        return new ClassLoaderDefineClassSupport(jni, callerClass, generatedClassName, traceWriter, agent);
+    }
+
+    protected AbstractDynamicClassGenerationSupport(JNIEnvironment jni, JNIObjectHandle callerClass,
+                    String generatedClassName, TraceWriter traceWriter, NativeImageAgent agent) throws IOException {
+        if (dynclassDumpDir == null) {
```
vjovanov, line 74: How can this be non-null?

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
```diff
@@ -0,0 +1,196 @@
...
+            System.out.println("Warning: dynamic-class-dump-dir= was not set in -agentlib:native-image-agent=. Dynamically generated classes will be dumped to the default location:" +
```
vjovanov, line 75: I would avoid printing in the agent as this could alter the behavior of the program. In other places the agent uses `System.err`.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
```diff
@@ -0,0 +1,196 @@
...
+                            DEFAULT_DUMP.toAbsolutePath().toString());
+            dynclassDumpDir = DEFAULT_DUMP;
+        }
+
+        if (!Files.exists(dynclassDumpDir)) {
+            Files.createDirectories(dynclassDumpDir);
+        } else if (!Files.isDirectory(dynclassDumpDir)) {
+            throw new IOException("File " + dynclassDumpDir + " already exists! Cannot create the same directory for class file dumping.");
```
vjovanov, line 83: I would mention here that the file in place is a file and that we want to create a dir there.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -678,6 +699,124 @@ private static boolean handleGetSystemResources(JNIEnvironment jni, Breakpoint b
...
+                    JNIObjectHandle resourceNameJString = jniFunctions().getNewStringUTF().invoke(jni, resourceNameHolder.get());
+                    JNIObjectHandle returnValue = callObjectMethodL(jni, self, agent.handles().javaLangClassLoaderGetResource, resourceNameJString);
+                    isDynamicallyGenerated = returnValue.equal(nullHandle());
+                }
+            }
+        }
+
+        // Reuse verifyForName
+        // CallerClass is always java.lang.ClassLoader, we only check the defined class
+        boolean allowed = (accessVerifier == null || accessVerifier.verifyForName(jni, nullHandle(), definedClassName));
```
vjovanov, line 754: In the new version of the code we need to pass an extra argument here. I am working with @d-kozak to see how to do that best.

vjovanov:
I have updated 

https://github.com/vjovanov/graal/tree/dynamicClassLoading

with the latest code that works. We should update this PR to that code so we can continue.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
```diff
@@ -0,0 +1,196 @@
...
+
+        // Get directory from package
+        Path dumpDirs = dumpFile.getParent();
+        if (!Files.exists(dumpDirs)) {
+            Files.createDirectories(dumpDirs);
+        } else if (!Files.isDirectory(dumpDirs)) {
+            throw new IOException("File " + dumpDirs + " already exists! Cannot create the same name directory for dumping class file.");
+        }
+        try (FileOutputStream stream = new FileOutputStream(dumpFile.toFile());) {
+            stream.write(values);
```
vjovanov, line 168: I think it would be necessary to use `com.oracle.svm.agent.NativeImageAgent#tryAtomicMove` as a guarantee that the file will always be complete.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
vjovanov: Fixed in my branch.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -678,6 +699,124 @@ private static boolean handleGetSystemResources(JNIEnvironment jni, Breakpoint b
...
+            // Check the caller's protectionDomain is null, otherwise don't support
+            String caller = getMethodFullNameAtFrame(jni, 1);
+            if ("java.lang.ClassLoader.defineClass(Ljava/lang/String;[BIILjava/security/ProtectionDomain;)Ljava/lang/Class;".equals(caller)) {
+                caller = getMethodFullNameAtFrame(jni, 2);
+                if (!"java.lang.ClassLoader.defineClass(Ljava/lang/String;[BII)Ljava/lang/Class;".equals(caller)) {
+                    // Reflections are supported in another way, no need to track dynamic class
+                    // loading
+                    // used by reflection.
+                    if (!caller.startsWith("sun.reflect")) {
+                        unsupportedExceptions.add("Don't support defineClass with ProtectionDomain parameter in native image. " +
```
vjovanov, line 772: I don't see why we would fail here. I think we should simply not collect those classes ATM. Actually, we could even collect them because in the future we definitely want to support the calls with the protection domain.

vjovanov:
The following benchmarks fail after all the updates:

  renaissance-native-image:philosophers
  renaissance-native-image:reactors
  renaissance-native-image:fj-kmeans

They all fail in the stage agent.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -132,7 +140,8 @@
     private static NativeImageAgent agent;
 
     private static Map<Long, Breakpoint> installedBreakpoints;
-
+    private static Set<String> definedClasses = new HashSet<>();
```
vjovanov, line 143: I think this data structure needs to handle concurrent updates?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/util/JavaClassUtil.java
```diff
@@ -0,0 +1,75 @@
...
+        if (sourceFileAttr.size() == 1) {
+            SourceFile sourFile = (SourceFile) sourceFileAttr.get(0);
+            int nameIndex = sourFile.getSourceFileIndex();
+            String originalSourceFile = ((ConstantUtf8) (jc.getConstantPool().getConstant(nameIndex))).getBytes();
+            if (originalSourceFile != null && originalSourceFile.length() > 0) {
+                jc.getConstantPool().setConstant(nameIndex, new ConstantUtf8(CONSTANT_PLACE_HOLDER));
+                return Arrays.hashCode(jc.getBytes());
+            }
+        }
+        return Arrays.hashCode(classDefinition);
```
vjovanov, line 66: Why not use SHA-256 or SHA-512 that are standard and can be used from any other language/tool?

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
```diff
@@ -0,0 +1,196 @@
...
+    protected JNIEnvironment jni;
+    protected JNIObjectHandle callerClass;
+    protected final String generatedClassName;
+    protected TraceWriter traceWriter;
+    protected NativeImageAgent agent;
+    protected int generatedClassHashCode = 0;
+    protected byte[] values = null;
+
+    private static Path dynclassDumpDir = null;
+    private static final Path DEFAULT_DUMP = Paths.get("dynClass");
```
vjovanov, line 60: I think it would be better to keep the dynamic classes in the sam folder as other config. This way it will be easier to use native image with `-H:ConfigurationFileDirectories=`. This is, for example the case in our benchmarks.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
ziyilin: Tracking along the data flow, the definition of `generatedClassName` is either from line 770 or line 772 of `BreakpointInterceptor.java` as shown in the following picture.
image

Line 770 gets the dynamically defined class' name from method `java.lang.ClassLoader.postDefineClass(Class<?> c, ProtectionDomain pd)`'s first parameter. It is not supposed to be null, so no further null check has been performed.

Line 772's assignment happens only when `nameFromDefineClassParam` is not null.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
ziyilin: Reporting at collecting time would let the user know the program is not supported as early as possible, so they don't have to wait to image compilation time or runtime. This unsupported message is reported at unLoad time so it wouldn't interfere the configuration collecting process.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/util/JavaClassUtil.java
ziyilin: I concerned the runtime performance. Here is not for security purpose to protect the class from malicious attach, but only for consistency verification, SHA-256 would be too heavy. But if there is a requirement to extend to other languages or tools, we can choose other standard but faster algorithms such as MD5.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
ziyilin: Agreed, as discussed previously (https://github.com/oracle/graal/pull/2442#issuecomment-676704131 and https://github.com/oracle/graal/pull/2442#issuecomment-677381880). I'll trace all dynamic generated classes in a new configuration file and store all generated classes under the same directory.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
vjovanov: One of the use cases for the agent is automatic collection of profiles in the cloud setting. For this purpose, I would like to have the agent behave exactly the same as the original. 

Could we put this kind of reporting behind a flag?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/util/JavaClassUtil.java
vjovanov: SHA-256 is about 11 cycles per byte. For a 10 kB class this would be 110k cycles or roughly 30 us. 

I think that is OK given that the VM will anyhow need to load, verify, interpret, and compile that code.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
peter-hofer: Thank you for your comments @vjovanov :+1: 
@ziyilin , please do address https://github.com/oracle/graal/pull/2442#issuecomment-676704131 and https://github.com/oracle/graal/pull/2442#issuecomment-677381880 and also consider which suggestions from the object serialization PR might apply here as well. Then I will review in more depth.

vjovanov:
The latest rebase is at:

  https://github.com/vjovanov/graal/tree/dynamicClassLoading

This includes the removal of the restrict mode.

ziyilin:
> The latest rebase is at:
> 
> https://github.com/vjovanov/graal/tree/dynamicClassLoading
> 
> This includes the removal of the restrict mode.

Thanks. I'm working on putting dynamic class loading information into a separated config file and all the dynamically dumped classes into the config directory.

vjovanov:
The biggest issue for me is now the fact that the agent fails at the end of execution. In all benchmarks this feature seems to be working, but then on exit, there is a failing exit code.

ziyilin:
I've changed the dynamic class configuration approach:
- A separated configuration file `META-INF/native-image/dynamicClassLoading-config.json` is used to store dynamically generated class information. Three attributes are saved for one class:
-- Class name
-- Class checksum
--Dumped class file name that are mangled 
- A directory `META-INF/native-image/dynClasses` to store all dumped class files.
- A new `DynamicClassLoadingFeature` class to resolve `dynamicClassLoading-config.json` file and dumped classes.

@vjovanov I tried to push this commit to your branch, but didn't have access to your repo. So I pushed them in my branch.

> The biggest issue for me is now the fact that the agent fails at the end of execution

Not sure what agent failure did you refer to. I run `mx --env ni-ce benchmark renaissance-native-image:philosophers -- --jvm=native-image --jvm-config=default-ce -Dnative-image.benchmark.stages=agent` and got a result like:

image

The stderr.log is empty and the stdout.log is:
```
====== philosophers (scala-stm) [default], iteration 0 started ======
GC before operation: completed in 64.396 ms, heap usage 202.286 MB -> 6.999 MB.
Camera thread performed 1156 scans.
WARNING: This benchmark provides no result that can be validated.
         There is no way to check that no silent failure occurred.
====== philosophers (scala-stm) [default], iteration 0 completed (46017.173 ms) ======

```

vjovanov:
Awesome progress, have you tried: 
```
mx --env ni-ce benchmark renaissance-native-image:log-regression -- --jvm=native-image --jvm-config=default-ce -Dnative-image.benchmark.stages=agent
```

For that one I get:
```
====== log-regression (apache-spark) [default], iteration 0 started ======
GC before operation: completed in 40.869 ms, heap usage 162.931 MB -> 29.401 MB.
WARNING: This benchmark provides no result that can be validated.
         There is no way to check that no silent failure occurred.
====== log-regression (apache-spark) [default], iteration 0 completed (16778.877 ms) ======
Failed in stage agent for renaissance-0.11.0-log-regression-default-ee with exit code 99
```

Also, for `scala-dacapo-native-image:scalaxb`:
```
$ mx --env ni-ce benchmark scala-dacapo-native-image:scalaxb -- --jvm=native-image --jvm-config=default-ce -Dnative-image.benchmark.stages=agent 

===== DaCapo 0.1.0-SNAPSHOT scalaxb starting =====
===== DaCapo 0.1.0-SNAPSHOT scalaxb PASSED in 8336 msec =====
native-image-agent: error when writing configuration files: java.nio.file.NoSuchFileException: /var/folders/h8/dcf1d1pd1p90zgnndn7cb8dh0000gn/T/1608292327654/dynClasses
java.nio.file.NoSuchFileException: /var/folders/h8/dcf1d1pd1p90zgnndn7cb8dh0000gn/T/1608292327654/dynClasses
	at sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
	at sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:149)
	at java.nio.file.Files.readAttributes(Files.java:1763)
	at java.nio.file.FileTreeWalker.getAttributes(FileTreeWalker.java:219)
	at java.nio.file.FileTreeWalker.visit(FileTreeWalker.java:276)
	at java.nio.file.FileTreeWalker.walk(FileTreeWalker.java:322)
	at java.nio.file.Files.walkFileTree(Files.java:2716)
	at java.nio.file.Files.walkFileTree(Files.java:2796)
	at com.oracle.svm.agent.ClassLoaderDefineClassSupport.deleteDumpedDir(ClassLoaderDefineClassSupport.java:106)
	at com.oracle.svm.agent.NativeImageAgent.onUnloadCallback(NativeImageAgent.java:518)
	at com.oracle.svm.jvmtiagentbase.JvmtiAgentBase.onUnload(JvmtiAgentBase.java:267)
357 unsupported features are detected 
java.lang.UnsupportedOperationException: Don't support defineClass with ProtectionDomain parameter in native image. Please try to use java.lang.ClassLoader.defineClass(String name, byte[] b, int off, int len) if possible.
    java.lang.ClassLoader.postDefineClass(Ljava/lang/Class;Ljava/security/ProtectionDomain;)V
    java.lang.ClassLoader.defineClass(Ljava/lang/String;[BIILjava/security/ProtectionDomain;)Ljava/lang/Class;
    java.security.SecureClassLoader.defineClass(Ljava/lang/String;[BIILjava/security/CodeSource;)Ljava/lang/Class;
    java.net.URLClassLoader.defineClass(Ljava/lang/String;Ljdk/internal/loader/Resource;)Ljava/lang/Class;
    java.net.URLClassLoader$1.run()Ljava/lang/Class;
    java.net.URLClassLoader$1.run()Ljava/lang/Object;
    java.security.AccessController.doPrivileged(Ljava/security/PrivilegedExceptionAction;Ljava/security/AccessControlContext;)Ljava/lang/Object;
    java.net.URLClassLoader.findClass(Ljava/lang/String;)Ljava/lang/Class;
    org.dacapo.harness.DacapoClassLoader.loadClass(Ljava/lang/String;Z)Ljava/lang/Class;
    java.lang.ClassLoader.loadClass(Ljava/lang/String;)Ljava/lang/Class;
    scalaxb.compiler.Main.main([Ljava/lang/String;)V
    jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Ljava/lang/reflect/Method;Ljava/lang/Object;[Ljava/lang/Object;)Ljava/lang/Object;
    jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Ljava/lang/Object;[Ljava/lang/Object;)Ljava/lang/Object;
    jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Ljava/lang/Object;[Ljava/lang/Object;)Ljava/lang/Object;
    java.lang.reflect.Method.invoke(Ljava/lang/Object;[Ljava/lang/Object;)Ljava/lang/Object;
    org.scalabench.benchmarks.scalaxb.Scalaxb.iterate(Ljava/lang/String;)V
    org.dacapo.harness.Benchmark.run(Lorg/dacapo/harness/Callback;Ljava/lang/String;)Z
    org.dacapo.harness.TestHarness.runBenchmark(Ljava/io/File;Ljava/lang/String;Lorg/dacapo/harness/TestHarness;)V
    org.dacapo.harness.TestHarness.main([Ljava/lang/String;)V
    Harness.main([Ljava/lang/String;)V
```

1. `java.nio.file.NoSuchFileException: /var/folders/h8/dcf1d1pd1p90zgnndn7cb8dh0000gn/T/1608292327654/dynClasses` is probably just a bug. 
2. "java.lang.UnsupportedOperationException: Don't support defineClass with ProtectionDomain parameter in native image." This is not the agent's job. The image builder must fail when we encounter that method.

vjovanov:
This PR is gaining popularity :) How are things going? Can we be of any assistance?

ziyilin:
I was a little busy with other stuffs.  Now getting back to this PR.

vjovanov:
@ziyilin any news on this PR? We need the implementation of this feature for 21.2.

ziyilin:
> @ziyilin any news on this PR? We need the implementation of this feature for 21.2.

Just pushed a new version, using SHA-256 for class checksum and fixing the conflicts and CI errors. Thought the update of https://github.com/oracle/graal/pull/2442#issuecomment-792408568 was still under review, didn't check the CI status and conflicts.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
ziyilin: removed the reporting here.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/util/JavaClassUtil.java
ziyilin: fixed

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
ziyilin: fixed

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
ziyilin: fixed

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
ziyilin: fixed

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
ziyilin: fixed

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/AbstractDynamicClassGenerationSupport.java
ziyilin: Fixed. The dumping process is moved to `NativeImageAgent` , so the class files are written to the same tmp directory along with other configuration files, no need to maintain an independent dumping process here any more.

ziyilin:
@vjovanov Is there any review updates?

peter-hofer:
@ziyilin I'm looking into it.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/ClassForNameSupport.java
```diff
@@ -41,12 +41,19 @@ static ClassForNameSupport singleton() {
 
     /** The map used to collect registered classes. */
     private final EconomicMap<String, Class<?>> knownClasses = ImageHeapMap.create();
+    private final EconomicMap<String, String> dynamicGeneratedClasses = ImageHeapMap.create();
```
peter-hofer, line 44: The hash code should be kept in a compact byte array (should be converted during parsing).

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/dynamicclassloading/hosted/DynamicClassLoadingFeature.java
```diff
@@ -0,0 +1,168 @@
...
+     */
+    private static void addURL2URLClassLoader(URL url, URLClassLoader classLoader) {
+        Method addURL = ReflectionUtil.lookupMethod(URLClassLoader.class, "addURL", URL.class);
+        try {
+            addURL.invoke(classLoader, url);
+        } catch (IllegalAccessException e) {
+            e.printStackTrace();
+        } catch (InvocationTargetException e) {
+            e.printStackTrace();
+        }
```
peter-hofer, line 89: Needs proper rethrowing.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/config/ConfigurationParserUtils.java
```diff
@@ -124,12 +131,14 @@ public boolean tryAdvance(Consumer<? super URL> action) {
...
         }).sum();
-
         return parsedCount;
     }
 
     private static void doParseAndRegister(ConfigurationParser parser, Reader reader, String featureName, Object location, HostedOptionKey<LocatableMultiOptionValue.Strings> option) {
         try {
+            if (featureName.equals("dynamicClassLoading") && location instanceof URL) {
+                dynamicClassLoadingDirs.add((URL) location);
+            }
```
peter-hofer, line 141: Not clean and the static field will not work when there is more than one image build in the process (server mode). You might need to add a method to collect the locations and return them so you can use them in the feature's `afterRegistration` method.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/dynamicclassloading/hosted/DynamicClassLoadingFeature.java
```diff
@@ -0,0 +1,168 @@
...
+        } catch (IllegalAccessException | InvocationTargetException e) {
+            throw new ClassNotFoundException(name);
+        }
+    }
+
+    @Override
+    public void afterRegistration(AfterRegistrationAccess a) {
+        FeatureImpl.AfterRegistrationAccessImpl access = (FeatureImpl.AfterRegistrationAccessImpl) a;
+        ImageClassLoader imageClassLoader = access.getImageClassLoader();
+        URLClassLoader cl = (URLClassLoader) imageClassLoader.getClassLoader();
```
peter-hofer, line 113: Assuming that the image class loader is a `URLClassLoader` and manipulating it through reflection does not seem like a good idea at all (especially with the module system). These steps need proper support in `ImageClassLoader` and `AbstractNativeImageClassLoaderSupport` and its subclasses and that probably entail creating a `ClassLoader` subclass for these purposes.

Moreover it seems very roundabout and not safe to reconstruct a class directory hierarchy, add it to the loader's class path (which potentially) and load the class from there. Is it not possible to just load the class from a byte array with the file's contents (other than that we need to call `defineClass` somehow)?

This _really_ needs fixing.

peter-hofer (changes requested):

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_lang_ClassLoader.java
```diff
@@ -414,21 +416,32 @@ private void clearAssertionStatus() {
...
+        throw VMError.unsupportedFeature("Define class with protection domain is not supported");
     }
 
     @Substitute
-    @SuppressWarnings({"unused", "static-method"})
+    @SuppressWarnings("unused")
     private Class<?> defineClass(String name, java.nio.ByteBuffer b, ProtectionDomain protectionDomain) {
-        throw VMError.unsupportedFeature("Defining classes from new bytecodes run time.");
+        throw VMError.unsupportedFeature("Define class with protection domain is not supported");
     }
```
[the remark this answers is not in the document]
ziyilin: `defineClass` APIs with `ProtectionDomain` parameter will call `ClassLoader.checkCerts` method to check if the class' certs match with the other same-package-classes'. However, Substrate VM does not store certs files from jar files to native image at build time, so there is always nothing to compare at image runtime. If the API is supported now, it behaves differently from what JDK does. I think finding a proper solution to reserve jar's certs files is beyond this PR's scope, so I leave these two APIs as unimplemented for now.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/dynamicclassloading/hosted/DynamicClassLoadingFeature.java
ziyilin: Fixed.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/dynamicclassloading/hosted/DynamicClassLoadingFeature.java
ziyilin: Now read the class file into byte array and then reflectively call `defineClass` with the byte array data.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/ClassForNameSupport.java
ziyilin: Fixed by changing `dynamicGeneratedClasses` to EconomicMap<String, byte[]>

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_lang_ClassLoader.java
```diff
@@ -526,3 +539,47 @@ public Object transform(MetaAccessProvider metaAccess, ResolvedJavaField origina
         }
     }
 }
+
+final class ClassLoaderHelper {
```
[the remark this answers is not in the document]
ziyilin: Fixed

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/ReflectionProcessor.java
```diff
@@ -247,6 +247,11 @@ public void processEntry(Map<String, ?> entry) {
                 resourceConfiguration.addBundle((String) args.get(2));
                 break;
             }
+            case "checksum": {
+                expectSize(args, 1);
+                configuration.getOrCreateType(clazzOrDeclaringClass).setDynamicClassChecksum((Integer) args.get(0));
+                break;
+            }
```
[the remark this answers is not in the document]
ziyilin: Removed this case. Not needed any more.

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationType.java
```diff
@@ -49,6 +49,8 @@
     private boolean allDeclaredConstructors;
     private boolean allPublicConstructors;
 
+    private int dynamicClassChecksum = 0;
+
```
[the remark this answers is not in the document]
ziyilin: Removed.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ConfigurationParser.java
```diff
@@ -82,4 +82,11 @@ protected static long asLong(Object value, String propertyName) {
         }
         throw new JSONParserException("Invalid long value '" + value + "' for element '" + propertyName + "'");
     }
+
+    static int asInteger(Object value, String propertyName) {
+        if (value instanceof Integer) {
+            return (int) value;
+        }
+        throw new JSONParserException("Invalid int value '" + value + "' for element '" + propertyName + "'");
+    }
```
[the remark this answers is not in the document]
ziyilin: Removed

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/ConfigurationTool.java
```diff
@@ -195,6 +195,12 @@ private static void generate(Iterator<String> argsIter, boolean acceptTraceFileA
                     set.getSerializationConfigPaths().add(requirePathUri(current, value));
                     break;
 
+                case "--dynamicClassLoading-input":
+                    set = inputSet; // fall through
+                case "--dynamicClassLoading-output":
+                    set.getDynamicClassLoadingConfigPaths().add(requirePathUri(current, value));
+                    break;
```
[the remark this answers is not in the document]
ziyilin: Fixed, and all _loading_ post-fixes are removed.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -573,6 +593,96 @@ private static boolean handleGetSystemResources(JNIEnvironment jni, Breakpoint b
...
+        boolean justAdded = definedClasses.add(definedClassName);
+        if (isDynamicallyGenerated) {
+            if (!justAdded) {
+                unsupportedExceptions.add("Class " + definedClassName + " has been defined before. Multiple definitions are not supported.\n" +
+                                ClassLoaderDefineClassSupport.getStackTrace(jni).toString());
+                return true;
+            }
+            // Check the caller is using byte array or directedBuffer
+            String caller = getMethodFullNameAtFrame(jni, 1);
+            boolean isByteArray = "java.lang.ClassLoader.defineClass(Ljava/lang/String;[BIILjava/security/ProtectionDomain;)Ljava/lang/Class;".equals(caller);
```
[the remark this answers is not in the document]
ziyilin: Fixed as the whole `postDefine` method is removed.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/ClassLoaderDefineClassSupport.java
```diff
@@ -0,0 +1,175 @@
...
+
+    private static final String DEFINE_CLASS_BYTEARRAY_SIG = "java.lang.ClassLoader.defineClass(Ljava/lang/String;[BII)Ljava/lang/Class;";
+    protected JNIEnvironment jni;
+    protected JNIObjectHandle callerClass;
+    protected final String generatedClassName;
+    protected TraceWriter traceWriter;
+    protected NativeImageAgent agent;
+    protected String generatedClassHashCode = null;
+    protected byte[] values = null;
+    private boolean isByteArray;
```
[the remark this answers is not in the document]
ziyilin: Fixed by removing them.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/ClassLoaderDefineClassSupport.java
```diff
@@ -0,0 +1,175 @@
...
+
+import static com.oracle.svm.jvmtiagentbase.Support.getIntArgument;
+import static com.oracle.svm.jvmtiagentbase.Support.getMethodFullNameAtFrame;
+import static com.oracle.svm.jvmtiagentbase.Support.getObjectArgument;
+import static com.oracle.svm.jvmtiagentbase.Support.jniFunctions;
+
+/**
+ * Support dynamic class loading that is implemented by java.lang.ClassLoader.defineClass.
+ */
+public class ClassLoaderDefineClassSupport {
```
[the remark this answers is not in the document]
ziyilin: Fixed by making `trace` method static.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -573,6 +593,96 @@ private static boolean handleGetSystemResources(JNIEnvironment jni, Breakpoint b
...
+            definedClassName = getClassNameOrNull(jni, getObjectArgument(1));
+        } else {
+            definedClassName = nameFromDefineClassParam;
+            // Filter out internal classes which are definitely not dynamically generated
+            // CallerClass is always java.lang.ClassLoader, we only check the defined class
+            AccessAdvisor postDefineCLassAccessAdvisor = new AccessAdvisor();
+            postDefineCLassAccessAdvisor.setInLivePhase(true);
+            if (postDefineCLassAccessAdvisor.shouldIgnore(new LazyValue<>(() -> definedClassName), new LazyValue<>(() -> null))) {
+                isDynamicallyGenerated = false;
+            }
```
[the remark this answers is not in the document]
ziyilin: Fixed as the whole `postDefine` method is removed.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -573,6 +593,96 @@ private static boolean handleGetSystemResources(JNIEnvironment jni, Breakpoint b
...
     }
 
+    /**
+     * java.lang.ClassLoader.postDefineClass is always called in java.lang.ClassLoader.defineClass,
+     * so intercepting postDefineClass is equivalent to intercepting defineClass but with extra
+     * benefit of being always able to get defined class' name even if defineClass' classname
+     * parameter is null.
+     */
+    @SuppressWarnings("unused")
+    private static boolean postDefineClass(JNIEnvironment jni, Breakpoint bp) {
```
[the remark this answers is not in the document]
ziyilin: You are right. I have removed the `postDefineClass` method and added a new method `onClassFileLoadHook` instead.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -1269,15 +1379,29 @@ private static void bindNativeBreakpoint(JNIEnvironment jni, NativeBreakpoint bp
...
         installedBreakpoints = null;
         nativeBreakpoints = null;
         observedExplicitLoadClassCallSites = null;
         traceWriter = null;
+        definedClasses = null;
     }
 
     private interface BreakpointHandler {
-        boolean dispatch(JNIEnvironment jni, Breakpoint bp);
+        boolean dispatch(JNIEnvironment jni, Breakpoint bp) throws IOException;
```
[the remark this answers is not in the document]
ziyilin: Fixed

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -573,6 +593,96 @@ private static boolean handleGetSystemResources(JNIEnvironment jni, Breakpoint b
...
+            JNIValue args = StackValue.get(2, JNIValue.class);
+            args.addressOf(0).setObject(getObjectArgument(1));
+            args.addressOf(1).setObject(getObjectArgument(2));
+            jniFunctions().getCallVoidMethodA().invoke(jni, self, bp.method, args);
+            if (clearException(jni)) {
+                // No need to proceed if any exception happens
+                result = false;
+            } else {
+                result = true;
+            }
```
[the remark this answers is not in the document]
ziyilin: Fixed by changing `postDefineClass` method to `onClassFileLoadHook` method.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -573,6 +593,96 @@ private static boolean handleGetSystemResources(JNIEnvironment jni, Breakpoint b
...
+        }
+
+        // CallerClass is always java.lang.ClassLoader, we only check the defined class
+        Object result = false;
+        boolean justAdded = definedClasses.add(definedClassName);
+        if (isDynamicallyGenerated) {
+            if (!justAdded) {
+                unsupportedExceptions.add("Class " + definedClassName + " has been defined before. Multiple definitions are not supported.\n" +
+                                ClassLoaderDefineClassSupport.getStackTrace(jni).toString());
```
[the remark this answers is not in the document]
ziyilin: Removed.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -169,6 +177,18 @@ private static void traceBreakpoint(JNIEnvironment env, JNIObjectHandle clazz, J
...
+                    JNIObjectHandle callerClass, String function, boolean allowWrite, boolean unsafeAccess, Object result,
+                    String fieldName) {
+        if (traceWriter != null) {
+            traceWriter.traceCall("reflect", function, getClassNameOr(env, clazz, null, TraceWriter.UNKNOWN_VALUE),
+                            getClassNameOr(env, declaringClass, null, TraceWriter.UNKNOWN_VALUE),
+                            getClassNameOr(env, callerClass, null, TraceWriter.UNKNOWN_VALUE), result, allowWrite, unsafeAccess,
+                            fieldName);
+            guarantee(!testException(env));
+        }
+    }
```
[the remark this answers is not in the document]
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/DynamicClassLoadingConfigurationParser.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public void parseAndRegister(Reader reader) throws IOException {
+        JSONParser parser = new JSONParser(reader);
+        Object json = parser.parse();
+        for (Object dynamicDefinedClass : asList(json, "first level of document must be an array of dynamic defined classes")) {
+            Map<String, Object> data = asMap(dynamicDefinedClass, "second level of document must be dynamic defined class descriptor objects ");
+            String definedClassName = asString(data.get("name"));
+            String dumpedFileName = asString(data.get("classFile"));
+            String checksum = asString(data.get("checksum"));
```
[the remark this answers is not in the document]
ziyilin: Fixed

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/TraceWriter.java
```diff
@@ -110,6 +110,19 @@ public void traceCall(String tracer, String function, Object clazz, Object decla
...
+        entry.put("definedClassName", definedClassName);
+        entry.put("checksum", checksum);
+        // classContents will be dumped to file and then set dumpedFileName
+        // at the end of tracing process
+        entry.put("dumpedFileName", null);
+        entry.put("classContents", classContents);
+        entry.put("result", result);
+        traceEntry(entry);
+    }
+
```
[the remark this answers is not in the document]
ziyilin: Fixed

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/ConfigurationTool.java
```diff
@@ -299,6 +306,11 @@ private static void generate(Iterator<String> argsIter, boolean acceptTraceFileA
                 p.getSerializationConfiguration().printJson(writer);
             }
         }
+        for (URI uri : outputSet.getDynamicClassLoadingConfigPaths()) {
+            try (JsonWriter writer = new JsonWriter(Paths.get(uri))) {
+                p.getResourceConfiguration().printJson(writer);
```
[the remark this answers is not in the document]
ziyilin: I've changed it to `p.getDynamicClassesConfiguration()`. But I'm not sure if we should provide this function. The changing of dynamic class configuration location requires to copy the dumped class files to new location, too.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageAgentJNIHandleSet.java
```diff
@@ -39,17 +39,27 @@
...
+    final JNIObjectHandle javaLangNoSuchMethodException;
+    public final JNIObjectHandle javaLangNoSuchFieldError;
+    final JNIObjectHandle javaLangNoSuchFieldException;
+    final JNIObjectHandle javaLangClassNotFoundException;
+    final JNIObjectHandle javaLangRuntimeException;
+    final JNIObjectHandle javaUtilMissingResourceException;
+    final JNIMethodId javaLangClassGetDeclaredMethod;
+    final JNIMethodId javaLangClassGetDeclaredConstructor;
+    final JNIMethodId javaLangClassGetDeclaredField;
+    final JNIMethodId javaLangClassGetName;
```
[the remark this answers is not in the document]
ziyilin: Fixed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/config/ConfigurationParserUtils.java
ziyilin: Removed the `dynamicClassLoadingDirs`. The original way is to fetch the dumped class file from all possible locations setting by `-H:DynamicClassesConfigurationFiles=` and `-H:DynamicClassesConfigurationResources=` , but the class file should be located in the `dynClasses` directory under the configuration file's directory. So we only need to follow the location of the currently parsed dynamicClass configuration file to look for the dumped class files.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -573,6 +593,96 @@ private static boolean handleGetSystemResources(JNIEnvironment jni, Breakpoint b
...
+            } else {
+                // 3. A dynamically defined class always return null
+                // when call java.lang.ClassLoader.getResource(classname)
+                // This is the accurate but slow way.
+                String asResourceName = definedClassName.replace('.', '/') + ".class";
+                try (CCharPointerHolder resourceNameHolder = toCString(asResourceName);) {
+                    JNIObjectHandle resourceNameJString = jniFunctions().getNewStringUTF().invoke(jni, resourceNameHolder.get());
+                    JNIObjectHandle returnValue = callObjectMethodL(jni, self, agent.handles().javaLangClassLoaderGetResource, resourceNameJString);
+                    isDynamicallyGenerated = returnValue.equal(nullHandle());
+                }
```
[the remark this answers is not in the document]
ziyilin: I have removed the `$/$$` way. 
This part of code is now working inside the `onClassFileLoadHook`. There are 3 checks: 
1. if the classloader is null, it is not dynamic class
2. if the class name is null, it is the dynamic class
3. Still check via `getResource`

There is still no guarantee that a dynamic class always has a null `ProtectionDomain` or null `CodeSource.location`. User can provide any `ProtectionDomain` to `defineClass`.
 `getResource` is slow, but is accurate.

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationDynamicClass.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    public boolean contains(String definedClassName, String checksum2Match) {
+        return definedClassName != null && definedClassName.equals(qualifiedJavaName) && checksum2Match.equals(this.checksum);
+    }
+
+    @Override
+    public void printJson(JsonWriter writer) throws IOException {
+        // Mangle class' qualified name into a system independent format
+        String internalName = qualifiedJavaName.replace(".", ConfigurationFiles.MANGLE_SLASH);
```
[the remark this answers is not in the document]
ziyilin: As now I directly read the class file into byte array to call `defineClass` instead of finding class from class path, there is no need to keep the package information in the file name. Actually the file name can be anything that is unique.
What do you think using the already calculated SHA value as file name? In this way the file name is unique, and furthermore, we can save one entry in configuration file. As the file name is from SHA, only one of `classFile` and `checksum` entry is needed.

ziyilin:
@peter-hofer I have dealed with all comments, please take a look. Thanks.

peter-hofer:
@ziyilin , thanks, I will.

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationDynamicClass.java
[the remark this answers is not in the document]
peter-hofer: Yes, I believe the way forward is using a unique hash code in the file name, potentially the unqualified name too for identifiability to the developer.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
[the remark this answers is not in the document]
peter-hofer: Do you have an example for a custom class loader that provides a `ProtectionDomain` and `CodeSource` with a URL while `getResource` for the same class would return `null`?

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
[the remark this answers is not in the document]
peter-hofer: `unsupportedExceptions` and `reportExceptions` are unused now.

peter-hofer:
@ziyilin : looks good now, thanks again for your contribution. I'll take care of it from here :slightly_smiling_face:

ziyilin:
@peter-hofer Great! Thank you for your reviewing.

peter-hofer:
@dzou @dthommes we renamed this feature _predefined classes_ because there really is not much dynamic about this mechanism at image runtime. You can enable tracking in the agent with option `experimental-class-define-support=true`, then any classes not loaded by a built-in loader will be stored in the subdirectory `agent-extracted-predefined-classes` of the configuration directory and added to `predefined-classes-config.json`. You will likely need to manually pick those classes which should actually be predefined during the image build, and exclude others. There are others restrictions, a class cannot be initialized at image build time and at runtime, cannot be "loaded" more than once in the entire VM (isolate).

Issue #2387, opened 2020-04-24
native-image fails on windows with command size limit (CreateProcess error=206)


Mobe91:
I am using io.quarkus:quarkus-maven-plugin:1.3.2.Final and GraalVM CE 20.0.0 (Java 11) to build a native image on Windows 10 that issues to following command: 
```
native-image.cmd -J-DCoordinatorEnvironmentBean.transactionStatusManagerEnable=false -J-Dsun.nio.ch.maxUpdateArraySize=100 -J-Djava.util.logging.manager=org.jboss.logmanager.LogManager -J-Dvertx.logger-delegate-factory-class-name=io.quarkus.vertx.core.runtime.VertxLogDelegateFactory -J-Dvertx.disableDnsResolver=true -J-Dio.netty.leakDetection.level=DISABLED -J-Dio.netty.allocator.maxOrder=1 -J-Duser.language=de -J-Dfile.encoding=Cp1252 --initialize-at-build-time= -H:InitialCollectionPolicy=com.oracle.svm.core.genscavenge.CollectionPolicy$BySpaceAndTime -H:+JNI -jar quarkus-runner.jar -H:FallbackThreshold=0 -H:+ReportExceptionStackTraces -H:-AddAllCharsets -H:-IncludeAllTimeZones -H:EnableURLProtocols=http,https --enable-all-security-services -H:-UseServiceLoaderFeature -H:+StackTrace quarkus-runner
```
This in turn launches another command targeting java.exe but the -imagecp argument is way too long which is why it fails with "CreateProcess error=206, file name or extension too long".

The problem is caused by `https://github.com/oracle/graal/blob/00e729aa62e22646015f3051be7f3c1ae066ff01/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java#L1169`. Rather than concatenating all the paths and passing it to the command, this should generate a classpath jar and pass it instead. Alternatively, an args file could be used for Java 9+ as described [here](https://stackoverflow.com/a/54270831/412446).

**Describe GraalVM and your environment:**
 - GraalVM version: **CE 20.0.0**
 - JDK major version: **11**
 - OS: **Windows 10**
 - Architecture: **AMD64**

quintesse:
I created a PR with a possible solution. Is that what you were looking for @Mobe91 ?

PR #2443, opened 2020-05-08
Use @ arguments file for running java as part of native-image on Windows

quintesse:
A very simple solution to #2387 using the @ argument support added to `java` in version 9.

The other solution offered in the issue seems much more complex just for having it work on Java 8. If that's essential it seems more useful to backport the @ argument support to version 8 instead.

Btw, the code could be simplified a bit more by removing the length check and just always generate the @ file on windows. Not sure what would be preferable though.

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -73,7 +73,6 @@ import java.util.stream.Collectors;
 
 import org.graalvm.compiler.options.OptionKey;
 import org.graalvm.compiler.serviceprovider.JavaVersionUtil;
-import org.graalvm.nativeimage.Platform;
 import org.graalvm.nativeimage.ProcessProperties;
 
 import com.oracle.graal.pointsto.api.PointstoOptions;
@@ -98,8 +97,6 @@ public class NativeImage {
     private static final String DEFAULT_GENERATOR_CLASS_NAME = "com.oracle.svm.hosted.NativeImageGeneratorRunner";
     private static final String DEFAULT_GENERATOR_9PLUS_SUFFIX = "$JDK9Plus";
     private static final String CUSTOM_SYSTEM_CLASS_LOADER = NativeImageSystemClassLoader.class.getCanonicalName();
-
-    private static final int WINDOWS_MAX_COMMAND_LENGTH = 32768;
 
     static final boolean IS_AOT = Boolean.getBoolean("com.oracle.graalvm.isaot");
 
@@ -1259,17 +1256,15 @@ public class NativeImage {
 
         int exitStatus = 1;
         try {
-            if (Platform.includedIn(Platform.WINDOWS.class)) {
-                int totalLength = command.stream().mapToInt(s -> s.length() + 1).sum();
-                if (totalLength >= WINDOWS_MAX_COMMAND_LENGTH) {
-                    Path argsFile = Files.createTempFile("native-image", "args");
-                    argsFile.toFile().deleteOnExit();
-                    Files.write(argsFile, (Iterable<String>) command.stream().skip(1).map(NativeImage::quoteFileArg)::iterator);
-                    List<String> atCommand = new ArrayList<>();
-                    atCommand.add(command.get(0));
-                    atCommand.add("@" + argsFile);
-                    command = atCommand;
-                }
+            if (config.useJavaModules()) {
+                /* For Java > 8 we use an argument file to pass the options to the builder */
+                Path argsFile = Files.createTempFile("native-image", "args");
+                argsFile.toFile().deleteOnExit();
+                Files.write(argsFile, (Iterable<String>) command.stream().skip(1).map(NativeImage::quoteFileArg)::iterator);
+                List<String> atCommand = new ArrayList<>();
+                atCommand.add(command.get(0));
+                atCommand.add("@" + argsFile);
+                command = atCommand;
             }
             ProcessBuilder pb = new ProcessBuilder();
             pb.command(command);
@@ -1282,11 +1277,11 @@ public class NativeImage {
     }
 
     private static String quoteFileArg(String arg) {
-        String result = arg.replaceAll(Pattern.quote("\\"), Matcher.quoteReplacement("\\\\"));
-        if (result.contains(" ")) {
-            result = "\"" + result + "\"";
+        String resultArg = arg.replaceAll(Pattern.quote("\\"), Matcher.quoteReplacement("\\\\"));
+        if (resultArg.contains(" ")) {
+            resultArg = "\"" + resultArg + "\"";
         }
-        return result;
+        return resultArg;
     }
 
     private static final Function<BuildConfiguration, NativeImage> defaultNativeImageProvider = config -> IS_AOT ? NativeImageServer.create(config) : new NativeImage(config);
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -45,7 +45,6 @@ import java.nio.file.LinkOption;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -1257,11 +1256,10 @@ public class NativeImage {
 
         int exitStatus = 1;
         try {
-            Path argsFile = null;
             Path[] argsFileBox = new Path[1];
             if (config.useJavaModules()) {
                 /* For Java > 8 we use an argument file to pass the options to the builder */
-                argsFile = Files.createTempFile("native-image", "args");
+                Path argsFile = Files.createTempFile("native-image", "args");
                 argsFileBox[0] = argsFile;
                 Runtime.getRuntime().addShutdownHook(new Thread() {
                     @Override
@@ -1270,9 +1268,7 @@ public class NativeImage {
                             argsFileBox[0].toFile().delete();
                         }
                     }
-                });                
-                argsFile = Files.createTempFile("native-image", "args");
-                argsFile.toFile().deleteOnExit();
+                });
                 Files.write(argsFile, (Iterable<String>) command.stream().skip(1).map(NativeImage::quoteFileArg)::iterator);
                 List<String> atCommand = new ArrayList<>();
                 atCommand.add(command.get(0));
@@ -1283,9 +1279,8 @@ public class NativeImage {
             pb.command(command);
             Process p = pb.inheritIO().start();
             exitStatus = p.waitFor();
-            if (exitStatus != 0 && isVerbose() && argsFile != null) {
+            if (exitStatus != 0 && isVerbose()) {
                 argsFileBox[0] = null;
-                showMessage("Argument file: " + argsFile);
             }
         } catch (IOException | InterruptedException e) {
             throw showError(e.getMessage());
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -1258,8 +1258,19 @@ public class NativeImage {
         int exitStatus = 1;
         try {
             Path argsFile = null;
+            Path[] argsFileBox = new Path[1];
             if (config.useJavaModules()) {
                 /* For Java > 8 we use an argument file to pass the options to the builder */
+                argsFile = Files.createTempFile("native-image", "args");
+                argsFileBox[0] = argsFile;
+                Runtime.getRuntime().addShutdownHook(new Thread() {
+                    @Override
+                    public void run() {
+                        if (argsFileBox[0] != null) {
+                            argsFileBox[0].toFile().delete();
+                        }
+                    }
+                });                
                 argsFile = Files.createTempFile("native-image", "args");
                 argsFile.toFile().deleteOnExit();
                 Files.write(argsFile, (Iterable<String>) command.stream().skip(1).map(NativeImage::quoteFileArg)::iterator);
@@ -1273,10 +1284,8 @@ public class NativeImage {
             Process p = pb.inheritIO().start();
             exitStatus = p.waitFor();
             if (exitStatus != 0 && isVerbose() && argsFile != null) {
-                /* Create a copy of the argument file for failure analysis */
-                Path argsFileCopy = Paths.get("native-image-failure.args");
-                Files.copy(argsFile, argsFileCopy, StandardCopyOption.REPLACE_EXISTING);
-                showMessage("Argument file: " + argsFileCopy);
+                argsFileBox[0] = null;
+                showMessage("Argument file: " + argsFile);
             }
         } catch (IOException | InterruptedException e) {
             throw showError(e.getMessage());
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -67,10 +67,13 @@ import java.util.function.Supplier;
 import java.util.jar.Attributes;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.graalvm.compiler.options.OptionKey;
 import org.graalvm.compiler.serviceprovider.JavaVersionUtil;
+import org.graalvm.nativeimage.Platform;
 import org.graalvm.nativeimage.ProcessProperties;
 
 import com.oracle.graal.pointsto.api.PointstoOptions;
@@ -95,6 +98,8 @@ public class NativeImage {
     private static final String DEFAULT_GENERATOR_CLASS_NAME = "com.oracle.svm.hosted.NativeImageGeneratorRunner";
     private static final String DEFAULT_GENERATOR_9PLUS_SUFFIX = "$JDK9Plus";
     private static final String CUSTOM_SYSTEM_CLASS_LOADER = NativeImageSystemClassLoader.class.getCanonicalName();
+
+    private static final int WINDOWS_MAX_COMMAND_LENGTH = 32768;
 
     static final boolean IS_AOT = Boolean.getBoolean("com.oracle.graalvm.isaot");
 
@@ -1226,8 +1231,7 @@ public class NativeImage {
 
     protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedHashSet<Path> cp, LinkedHashSet<String> imageArgs, LinkedHashSet<Path> imagecp) {
         /* Construct ProcessBuilder command from final arguments */
-        ProcessBuilder pb = new ProcessBuilder();
-        List<String> command = pb.command();
+        List<String> command = new ArrayList<>();
         command.add(canonicalize(config.getJavaExecutable()).toString());
         command.addAll(javaArgs);
         if (!bcp.isEmpty()) {
@@ -1255,12 +1259,34 @@ public class NativeImage {
 
         int exitStatus = 1;
         try {
+            if (Platform.includedIn(Platform.WINDOWS.class)) {
+                int totalLength = command.stream().mapToInt(s -> s.length() + 1).sum();
+                if (totalLength >= WINDOWS_MAX_COMMAND_LENGTH) {
+                    Path argsFile = Files.createTempFile("native-image", "args");
+                    argsFile.toFile().deleteOnExit();
+                    Files.write(argsFile, (Iterable<String>) command.stream().skip(1).map(NativeImage::quoteFileArg)::iterator);
+                    List<String> atCommand = new ArrayList<>();
+                    atCommand.add(command.get(0));
+                    atCommand.add("@" + argsFile);
+                    command = atCommand;
+                }
+            }
+            ProcessBuilder pb = new ProcessBuilder();
+            pb.command(command);
             Process p = pb.inheritIO().start();
             exitStatus = p.waitFor();
         } catch (IOException | InterruptedException e) {
             throw showError(e.getMessage());
         }
         return exitStatus;
+    }
+
+    private static String quoteFileArg(String arg) {
+        String result = arg.replaceAll(Pattern.quote("\\"), Matcher.quoteReplacement("\\\\"));
+        if (result.contains(" ")) {
+            result = "\"" + result + "\"";
+        }
+        return result;
     }
 
     private static final Function<BuildConfiguration, NativeImage> defaultNativeImageProvider = config -> IS_AOT ? NativeImageServer.create(config) : new NativeImage(config);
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -95,6 +95,8 @@
     private static final String DEFAULT_GENERATOR_9PLUS_SUFFIX = "$JDK9Plus";
     private static final String CUSTOM_SYSTEM_CLASS_LOADER = NativeImageSystemClassLoader.class.getCanonicalName();
 
+    private static final int WINDOWS_MAX_COMMAND_LENGTH = 8192;
```
dougxc, line 98: You could make this something closer to the [real limit](http://msdn.microsoft.com/en-us/library/ms682425%28VS.85%29.aspx) of 32768.

quintesse: I wasn't actually sure what the real limit was. Google turned up 2047 for some Windows versions and 8191 for others (see https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation). THe number of 32768 is also mentioned but for the size of the ENV vars.
If you are sure it's 32768 I'll be more than happy to change it.

quintesse: Ah duh, I see you actually provided a link. Sorry about the noise, I'll change it ASAP :-)

quintesse: Done.

quintesse:
Btw, the CI failed but not in something that seems to have to do anything with my PR?
It says the formatting of the file I changed is incorrect (but not in the part I changed) and to run `mx eclipseformat`. But when I run that it doesn't detect anything wrong (no changes are made). So I have no idea how to fix things.

quintesse:
About the OCA, I'm a Red Hat employee (tschotan@redhat.com), I sent in a signed copy of the agreement, but aren't we automatically covered already?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1203,11 +1204,30 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
         }
 
         int exitStatus = 1;
+        File argsFile = null;
         try {
+            if (OS.getCurrent() == OS.WINDOWS) {
```
olpaw, line 1209: Use `Platform.includedIn(Platform.WINDOWS.class)` instead of `OS.getCurrent() == OS.WINDOWS`. 
`Platform.includedIn(...)` is built-time constant. It ensures the code for special handling is not included in native-image images for Platforms != Windows.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1203,11 +1204,30 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
...
+                int totalLength = command.stream().mapToInt(s -> s.length() + 1).sum();
+                if (totalLength >= WINDOWS_MAX_COMMAND_LENGTH) {
+                    argsFile = File.createTempFile("native-image", "args");
+                    argsFile.deleteOnExit();
+                    Files.write(argsFile.toPath(), (Iterable<String>) command.stream().skip(1)::iterator);
```
olpaw, line 1214: Did you verify that this works if the arguments contain spaces (e.g. in paths)?
See https://docs.oracle.com/en/java/javase/14/docs/specs/man/java.html#java-command-line-argument-files

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1203,11 +1204,30 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
         }
 
         int exitStatus = 1;
+        File argsFile = null;
         try {
+            if (OS.getCurrent() == OS.WINDOWS) {
+                int totalLength = command.stream().mapToInt(s -> s.length() + 1).sum();
+                if (totalLength >= WINDOWS_MAX_COMMAND_LENGTH) {
+                    argsFile = File.createTempFile("native-image", "args");
```
olpaw, line 1212: `Files#createTempDirectory` is preferred.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1203,11 +1204,30 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
         }
 
         int exitStatus = 1;
+        File argsFile = null;
         try {
+            if (OS.getCurrent() == OS.WINDOWS) {
+                int totalLength = command.stream().mapToInt(s -> s.length() + 1).sum();
+                if (totalLength >= WINDOWS_MAX_COMMAND_LENGTH) {
```
olpaw, line 1211: If we make this the default for Java > 8 then we don't need to make this code Windows specific.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1203,11 +1204,30 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
...
+                    argsFile = File.createTempFile("native-image", "args");
+                    argsFile.deleteOnExit();
```
olpaw, line 1213: Do we really need `.deleteOnExit()` here if we already have `argsFile.delete();` in the finally-clause?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1203,11 +1204,30 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
...
+                }
+            }
+            ProcessBuilder pb = new ProcessBuilder();
+            pb.command(command);
             Process p = pb.inheritIO().start();
             exitStatus = p.waitFor();
         } catch (IOException | InterruptedException e) {
             throw showError(e.getMessage());
+        } finally {
+            if (argsFile != null) {
```
olpaw, line 1228: Use
```
`!useDebugAttach() && argsFile != null`
```
here.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: During testing I found out that hitting CTRL+C would not delete the file unless I also added the `.deleteOnExit()`. Now we could of course remove the `finally` block, because the process will never be long-lived, but it  seemed "untidy".

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: I will, but could you explain that to me?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: Yes, I said the same in the PR description. Is that what you want me to do? Just remove the platform check and make this the default for all platforms?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: Yes, I said the same in the PR description. Is that what you want me to do? Just remove the platform check and make this the default for all platforms?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: Ok

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
olpaw: I you run `native-image` with `--debug-attach` you would eventually want to inspect the `argsFile` to know exactly what was passed passed to the native-image builder. See
https://github.com/oracle/graal/blob/75df2c1dc9ce10c175261f9d003eb983625a81b6/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java#L153
https://github.com/oracle/graal/blob/75df2c1dc9ce10c175261f9d003eb983625a81b6/substratevm/src/com.oracle.svm.driver/resources/HelpExtra.txt#L9

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
olpaw: If we are confident after extensive testing that it causes no drawbacks then we should make that the default for Java > 8 (independent of the platform).

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
olpaw: > Now we could of course remove the finally block, because the process will never be long-lived, but it seemed "untidy".

Would be fine for me.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
olpaw: Looking at https://docs.oracle.com/en/java/javase/14/docs/specs/man/java.html#java-command-line-argument-files shows that some care is needed to ensure we do not run into problems with quoting and character escaping.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: Ok, done

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: OK, arguments that have spaces are now quoted (not sure we need to check for other whitespace chars as well? Given the fact it's almost all generated instead of user-supplied?) and backspaces are duplicated correctly.

olpaw: I'll this through all our gates and let you know if I find any quoting/escaping-related issues.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
olpaw: Please change the sources to use
```
        if (config.useJavaModules()) {
            /* For Java > 8 we use an argument file to pass the options to the builder */
```
instead of `if (Platform.includedIn(Platform.WINDOWS.class))` ... `if (totalLength >= WINDOWS_MAX_COMMAND_LENGTH) {`. Then I can run this PR in our CI to see if the argument file approach causes any trouble.

quintesse: Done

quintesse:
NB: I added an extra commit to fix a formatting problem in some other file that was tripping up the CI each time.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
olpaw: All internal CI gates are passing with a57da3f553aaec8ca353778bb5f24ccabff2fea1.

olpaw:
> NB: I added an extra commit to fix a formatting problem in some other file that was tripping up the CI each time.

Huh ... strange non of our internal CI gates has a problem with `substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/FirstObjectTable.java`. Do you have a link to the failing gate?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1204,6 +1205,18 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
 
         int exitStatus = 1;
         try {
+            if (config.useJavaModules()) {
+                /* For Java > 8 we use an argument file to pass the options to the builder */
+                Path argsFile = Files.createTempFile("native-image", "args");
+                argsFile.toFile().deleteOnExit();
```
dougxc, line 1211: Unconditionally deleting `argsFile` makes it harder to replay and debug failures. I would suggest only deleting it if `exitStatus` below is non-zero. In any case, no need to use `deleteOnExit` - just add a `finally` clause to the `try` statement.

olpaw:
Another problem is that this approach does not fix the problem for Windows native-images on Java 8. But its still an improvement over what we have now. cc @christianwimmer

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: @dougxc see the comments @olpaw made before. There was a try-finally deleting the file, but that doesn't delete the file when the user hits CTRL+C while deleteOnExit() does. So instead of having 2 ways of deleting it was decided to keep the one that works in more cases. For the debugging we an surround it with an `!useDebugAttach()` perhaps?

quintesse:
> Do you have a link to the failing gate?

Sure, this is one of them: https://travis-ci.org/github/oracle/graal/builds/686028526

olpaw:
> > Do you have a link to the failing gate?
> 
> Sure @olpaw , this is one of them: https://travis-ci.org/github/oracle/graal/builds/686028526

@quintesse Thanks! @dougxc looks like `/home/travis/build/oracle/graal/../eclipse/eclipse` is different from what we use in our internal CI.

dougxc:
Indeed:
https://github.com/oracle/graal/blob/master/common.hocon#L111
https://github.com/oracle/graal/blob/master/.travis.yml#L73

@quintesse can you please add this patch to this PR to see if it fixes the problem:
```
diff --git a/.travis.yml b/.travis.yml
index 24bc84f0c4f..bc55122bbfb 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ install:
         pip install pylint==1.1.0

         export ECLIPSE_TAR=${TRAVIS_BUILD_DIR}/../eclipse.tar.gz
-        wget https://archive.eclipse.org/eclipse/downloads/drops4/R-4.5.2-201602121500/eclipse-SDK-4.5.2-linux-gtk-x86_64.tar.gz -O ${ECLIPSE_TAR}
+        wget https://archive.eclipse.org/eclipse/downloads/drops4/R-4.14-201912100610/eclipse-SDK-4.14-linux-gtk-x86_64.tar.gz -O ${ECLIPSE_TAR}
         tar -C ${TRAVIS_BUILD_DIR}/.. -xzf ${ECLIPSE_TAR}
         export ECLIPSE_EXE=${TRAVIS_BUILD_DIR}/../eclipse/eclipse
```

quintesse:
> @quintesse can you please add this patch to this PR to see if it fixes the problem:

Sure @dougxc , done.

quintesse:
Oops, forgot I had to first remove my previous fix of course. Repushed.

quintesse:
Seems that fixed it @dougxc

jaikiran:
One of the users of Quarkus reported this issue recently https://github.com/quarkusio/quarkus/issues/10074. 

It would be good to have this PR merged and available in next release of Graal.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
dougxc: Ok, how about making a copy of the to-be-deleted-on-exit file if `exitStatus != 0`?

quintesse: Wouldn't that leave a whole bunch of files around on the user's system? Remember that this is not a debug option or anything, it will be used on every execution.

dougxc: I would suggest that the copy has a fixed name (e.g. `native-image-failure.args`) to avoid polluting the system. This file will deleted if it exists before running native-image only be left behind when native-image fails. In addition, all this can be activated only when `--verbose` is supplied. It's only in that scenario that someone may be interested in replaying the low-level native image command.

olpaw: @quintesse are you going to address https://github.com/oracle/graal/pull/2443#discussion_r443523222 or should we close this PR? cc @dougxc

quintesse: Done @olpaw @dougxc

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1255,14 +1257,41 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
...
+            }
+            ProcessBuilder pb = new ProcessBuilder();
+            pb.command(command);
             Process p = pb.inheritIO().start();
             exitStatus = p.waitFor();
+            if (exitStatus != 0 && isVerbose() && argsFile != null) {
+                /* Create a copy of the argument file for failure analysis */
+                Path tmpDir = Paths.get(System.getProperty("java.io.tmpdir"));
+                Path argsFileCopy = tmpDir.resolve("native-image-failure.args");
+                Files.copy(argsFile, argsFileCopy, StandardCopyOption.REPLACE_EXISTING);
```
dougxc, line 1279: How is the user supposed to find this copy? The location of the temp dir is not obvious on all OSes. For example, on my Mac it's `/var/folders/wr/4yqmgc5978z05qsv8j4sxjlr0000gn/T/` which is pretty hard to guess ;-) I would suggest putting the copy in the current working directory. In any case, you should issue a println saying where it is.

olpaw: How about using https://github.com/oracle/graal/blob/146a5548cf95c4803c10103f1e67a0e59dd93ce7/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java#L695 ?

quintesse: Implemented @dougxc 's suggestion. But personally I'd just a) not delete the temp file and b) print its location (only when `isVerbose()` is true). It's the simplest implementation, it uses temp files in a location meant for temp files and it's still easy to find (once we print the path).

dougxc: I still prefer the option of retaining the original temp file so that the command line can simply be copied and pasted for reproduction. You could achieve this with a "conditional" shutdown hook:
```
        Path argsFile = Files.createTempFile("native-image", "args");
        Path[] argsFileBox = {argsFile};
        Runtime.getRuntime().addShutdownHook(new Thread() {
            @Override
            public void run() {
                if (argsFileBox[0] != null) {
                    argsFileBox[0].toFile().delete();
                }
            }
        });

        ...

        if (exitStatus != 0 && isVerbose() && argsFile != null) {
           argsFileBox[0] = null;
        }
```

quintesse: Like this?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1255,14 +1257,50 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
...
+                atCommand.add("@" + argsFile);
+                command = atCommand;
+            }
+            ProcessBuilder pb = new ProcessBuilder();
+            pb.command(command);
             Process p = pb.inheritIO().start();
             exitStatus = p.waitFor();
+            if (exitStatus != 0 && isVerbose() && argsFile != null) {
+                argsFileBox[0] = null;
+                showMessage("Argument file: " + argsFile);
```
dougxc, line 1288: You can remove the `showMessage` line now.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1255,14 +1257,50 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
 
         int exitStatus = 1;
         try {
+            Path argsFile = null;
```
dougxc, line 1260: Move declaration of `argsFile` into the scope of `if (config.useJavaModules()) {`.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1255,14 +1257,50 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> bcp, LinkedH
...
+                argsFileBox[0] = argsFile;
+                Runtime.getRuntime().addShutdownHook(new Thread() {
+                    @Override
+                    public void run() {
+                        if (argsFileBox[0] != null) {
+                            argsFileBox[0].toFile().delete();
+                        }
+                    }
+                });                
+                argsFile = Files.createTempFile("native-image", "args");
```
dougxc, line 1274: These 2 lines should now be deleted.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
dougxc: Yes.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: But then how will people find the file as you mentioned before?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: ok

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
quintesse: oops, of course

quintesse:
Applied latest requested changes

dougxc (approved):

olpaw:
@quintesse you changed the PR while it was already merging. Therefore this PR did not close automatically. But your changes are on master: https://github.com/oracle/graal/commit/9a9039f3f3b6ff845dab153a6746155bdb7f4e9a

quintesse:
> you changed the PR while it was already merging

Hi, I did so because there was a failure in one of the integration tests telling me there was a formatting error. So I thought I had to fix that first. 

Thanks for merging.

PR #2446, opened 2020-05-11
AArch64: jaotc generates incorrect code for compressed OOPs with non-zero heap base

nick-arm:
On AArch64 if the VM jaotc is running in uses compressed oops with a
zero base it will not emit instructions to add/subtract the heap base
when compressing/decompressing oops. This causes a crash if the AOT'd
shared library is loaded into a VM with non-zero heap base.

Fix for JDK-8244164.


nick-arm:
I'm covered by the OCA signed by Arm Limited

nick-arm:
Any feedback on this? There's a jtreg test that can reproduce this issue here: http://cr.openjdk.java.net/~ngasson/8244164/webrev.0/test/hotspot/jtreg/compiler/aot/TestHeapBase.java.html

dougxc:
I'm putting this through the integration process now.

PR #2453, opened 2020-05-13
AArch64: use stlr for VMThread status update

teshull:
When updating the VMThread status a barrier should be present to ensure
all prior stores are visible to other threads. Failure to do so can
result in the garbage collector observing stale values.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.graal.aarch64/src/com/oracle/svm/core/graal/aarch64/SubstrateAArch64Backend.java
+++ b/substratevm/src/com.oracle.svm.core.graal.aarch64/src/com/oracle/svm/core/graal/aarch64/SubstrateAArch64Backend.java
@@ -242,12 +242,17 @@ public class SubstrateAArch64Backend extends SubstrateBackend implements LIRGene
         }
 
         if (SubstrateOptions.MultiThreaded.getValue()) {
-            /* Change the VMThread status from Java to Native. */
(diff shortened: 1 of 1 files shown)


teshull:
Note I am using a release barrier here instead of a store barrier. I think it is fine for loads/stores to be hoisted above the barrier, as at this point it should be in native code. Furthermore, before the transition back to Java code a lock must be grabbed which should also block subsequent code memory ops from executing. @christianhaeubl please let me know if you think my assumption is incorrect.

christianhaeubl:
Another option would be to do a full fence after a safepoint is requested (see `Safepoint.Master.freeze()`). This is probably cheaper as transitions to native can be quite frequent, while safepoints are rare.

adinn:
Hi Tom,

I agree with your argument that you should be able to get away with an stlr rather than a full dmb. The important thing is to order writes to Java fields which is what any reader of the threadStatus field will be reading.

I'm not sure I understand Christian's suggestion. I believe it is necessary to order visibility of stores prior to the status update at the point of write. If that is not done then a reader thread may see a change to threadStatus yet read stale values for the other fields. An ldar on the reader side (which is what will be compiled in because threadStatus is a volatile field) won't prevent that happening. Am I missing something?

christianhaeubl:
Yes, my previous comment was wrong, I initially thought that this would only affect the GC but that is not necessarily true. Thanks Andrew for pointing that out.

PR #2454, opened 2020-05-13
Allow user-defined image inclusion of builtin modules.

olpaw:
This is a proper fix for the workaround added in https://github.com/oracle/graal/commit/4a5f6ae8341071ff9225c93ea82db643a102320a


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ImageClassLoader.java
```diff
@@ -69,10 +69,15 @@
...
 import com.oracle.svm.util.ModuleSupport;
-import com.oracle.truffle.api.Truffle;
 
 public final class ImageClassLoader {
 
+    /*
+     * This cannot be a HostedOption because all Subclasses of OptionDescriptors from inside builtin
+     * modules need to be initialized prior to option parsing so that they can be found.
+     */
+    public static final String PROPERTY_IMAGEINCLUDEBUILTINMODULES = "substratevm.ImageIncludeBuiltinModules";
```
peter-hofer, line 79: Should this be named `native-image.ImageIncludeBuiltinModules`? I guess it depends whether this is potentially outward-facing and if we already have properties with either convention.

PR #2455, opened 2020-05-13
[GR-23150] Fix several JDK 15 related issues with native-image

bobvandette:
This PR fixes

1. Add isHidden substitution to java.lang.Class
2. Fix MethodHandleIntrinsic needed since Hidden classes were introduced
3. Fix issue with Nashorn being removed in JDK 15
4. Add loadLibrary method to BootLoader


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
```diff
@@ -487,10 +490,15 @@ private void processInvokeWithMethodHandle(GraphBuilderContext b, Replacements r
...
+            ValueNode singleNewInstance = null;
             for (Node node : graph.getNodes()) {
                 if (node == graph.start() || node instanceof ParameterNode || node instanceof ConstantNode || node instanceof FrameState) {
                     /* Ignore the allowed framework around the nodes we care about. */
-                } else if (node instanceof MethodCallTargetNode) {
+                } else if (node instanceof DynamicNewInstanceNode && singleNewInstance == null && ((DynamicNewInstanceNode) node).getInstanceType().isConstant()) {
+                    singleNewInstance = (DynamicNewInstanceNode) node;
+                } else if (node instanceof NewInstanceNode && singleNewInstance == null) {
+                    singleNewInstance = (NewInstanceNode) node;
+                } else if (node instanceof MethodCallTargetNode || node instanceof InstanceOfNode || node instanceof FixedGuardNode) {
```
christianwimmer, line 501: You cannot just ignore `InstanceOfNode` and `FixedGuardNode`. That would mean you omit necessary type checks.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
```diff
@@ -755,6 +756,12 @@ private boolean isAnonymousClass() {
         }
     }
 
+    @Substitute
+    @TargetElement(onlyWith = JDK15OrLater.class)
+    public boolean isHidden() {
+        throw VMError.shouldNotReachHere();
```
christianwimmer, line 762: We need to immediately have a complete implementation. So you need to preserve this information from the image generator in a flag in `DynamicHub` so that you can return it here.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
```diff
@@ -487,10 +490,15 @@ private void processInvokeWithMethodHandle(GraphBuilderContext b, Replacements r
              */
             Node singleFunctionality = null;
             ReturnNode singleReturn = null;
+            ValueNode singleNewInstance = null;
             for (Node node : graph.getNodes()) {
                 if (node == graph.start() || node instanceof ParameterNode || node instanceof ConstantNode || node instanceof FrameState) {
                     /* Ignore the allowed framework around the nodes we care about. */
-                } else if (node instanceof MethodCallTargetNode) {
+                } else if (node instanceof DynamicNewInstanceNode && singleNewInstance == null && ((DynamicNewInstanceNode) node).getInstanceType().isConstant()) {
```
christianwimmer, line 497: Why is such a  `DynamicNewInstanceNode` not canonicalized to a `NewInstanceNode`? Because the type is not initialized yet?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
```diff
@@ -516,12 +524,26 @@ private void processInvokeWithMethodHandle(GraphBuilderContext b, Replacements r
...
+                    if (singleNewInstance instanceof DynamicNewInstanceNode) {
+                        type = lookup(originalProviders.getConstantReflection().asJavaType(((DynamicNewInstanceNode) singleNewInstance).getInstanceType().asConstant()));
+                    }
+                    if (singleNewInstance instanceof NewInstanceNode) {
+                        type = lookup(((NewInstanceNode) singleNewInstance).instanceClass());
+                    }
+                    maybeEmitClassInitialization(b, true, resolvedTarget.getDeclaringClass());
+                    transplantedNewInstance = b.add(new SubstrateNewInstanceNode(type, true));
+                } else {
+                    maybeEmitClassInitialization(b, singleCallTarget.invokeKind() == InvokeKind.Static, resolvedTarget.getDeclaringClass());
```
christianwimmer, line 545: Why do you need this class initialization check for the invoke when there is a `singleNewInstance`?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
```diff
@@ -516,12 +524,26 @@ private void processInvokeWithMethodHandle(GraphBuilderContext b, Replacements r
 
             JavaKind returnResultKind = b.getInvokeReturnType().getJavaKind().getStackKind();
             ValueNode transplantedSingleFunctionality = null;
+            NewInstanceNode transplantedNewInstance = null;
             if (singleFunctionality instanceof Invoke) {
```
christianwimmer, line 528: What happens if there is a `singleNewInstance` but `singleFunctionality` is not an `Invoke`?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
bobvandette: I only added this during debugging thinking that it was the cause of a problem I was having. I'll remove it.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
bobvandette: I don't know for sure.  This was the initial cause of the failure requiring the fixes.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
bobvandette: Complete Hidden Class support is most likely going to take a lot of additional effort.  Creating Hidden classes is implemented in Hotspot and the hidden attribute is extracted via a native call to Hotspot.  My goal with this fix is to just get past these existing problems with JDK 15 and allow native-image to be used in the mean time.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
bobvandette: The processInvokeWithMethodHandle should always have an Invoke but I can add an assert.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
christianwimmer: Why would getting the information be difficult? You only need to call the `isHidden` method during image generation and store it in `DynamicHub`. `SVMHost.createHub` already has the `javaClass`.
You need to call `isHidden` via reflection since it doesn't exist in prior JDKs. But that's OK.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
bobvandette: I only see the DynamicNewInstanceNodes during the clinit phase of native-image.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
bobvandette: I can do that but there are other changes throughout the JDK that added support for isHidden like reflection and methodhandles.  These need to be looked at carefully.  Isn't it better to get an error if someone is trying to reference the isHidden attribute rather than to just report that is was or wasn't hidden during native-image generation?  I can go either way.

christianwimmer: Preserving the information from the image builder is the correct thing to do. There are never any new classes created at image run time, and the "hidden" attribute never changes at run time.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
bobvandette: This should be fixed.  Removed these and added a ClassCast check in the transplanted nodes.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
bobvandette: Ok, I'll add a reflected call to get the value and store it in a local DynamicHub field.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
bobvandette: Added a check for this situation.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java
bobvandette: Done.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
```diff
@@ -516,13 +528,26 @@ private void processInvokeWithMethodHandle(GraphBuilderContext b, Replacements r
...
-                maybeEmitClassInitialization(b, singleCallTarget.invokeKind() == InvokeKind.Static, resolvedTarget.getDeclaringClass());
+                if (singleNewInstance != null) {
+                    ResolvedJavaType type = null;
+                    if (singleNewInstance instanceof DynamicNewInstanceNode) {
+                        type = lookup(originalProviders.getConstantReflection().asJavaType(((DynamicNewInstanceNode) singleNewInstance).getInstanceType().asConstant()));
+                    }
+                    if (singleNewInstance instanceof NewInstanceNode) {
+                        type = lookup(((NewInstanceNode) singleNewInstance).instanceClass());
+                    }
+                    ValueNode newInstance = b.add(new NewInstanceNode(type, true));
```
christianwimmer, line 546: You also need a `maybeEmitClassInitialization` before the `NewInstanceNode`, otherwise you might allocate a non-initialized class.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
```diff
@@ -516,13 +528,26 @@ private void processInvokeWithMethodHandle(GraphBuilderContext b, Replacements r
...
+                    transplantedNewInstance = maybeEmitClassCast(b, classCastStamp, newInstance);
```
christianwimmer, line 547: How do you know the `classCastStamp` applies to the `newinstance` node?

christianwimmer: Actually the `maybeEmitClassCast` here is not necessary. A class cast is only removed above when the checked value is a `Invoke` or `LoadFieldNode`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationJavaNet.java
```diff
@@ -115,14 +114,6 @@ public void beforeAnalysis(BeforeAnalysisAccess a) {
         a.registerReachabilityHandler(JNIRegistrationJavaNet::registerDatagramPacketInit,
                         method(a, "java.net.DatagramPacket", "init"));
 
-        if (JavaVersionUtil.JAVA_SPEC >= 15) {
-            a.registerReachabilityHandler(JNIRegistrationJavaNet::registerDatagramSocketCheckOldImpl,
-                            method(a, "java.net.DatagramSocket", "checkOldImpl", java.net.DatagramSocketImpl.class));
-        } else {
-            a.registerReachabilityHandler(JNIRegistrationJavaNet::registerDatagramSocketCheckOldImpl,
```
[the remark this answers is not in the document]
bobvandette: Check the lasted version of this file in the PR.  I fixed this.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/IntrinsifyMethodHandlesInvocationPlugin.java
```diff
@@ -487,9 +492,14 @@ private void processInvokeWithMethodHandle(GraphBuilderContext b, Replacements r
              */
             Node singleFunctionality = null;
             ReturnNode singleReturn = null;
+            ValueNode singleNewInstance = null;
             for (Node node : graph.getNodes()) {
                 if (node == graph.start() || node instanceof ParameterNode || node instanceof ConstantNode || node instanceof FrameState) {
                     /* Ignore the allowed framework around the nodes we care about. */
+                } else if (node instanceof DynamicNewInstanceNode && singleNewInstance == null && ((DynamicNewInstanceNode) node).getInstanceType().isConstant()) {
```
christianwimmer, line 499: I just merged https://github.com/oracle/graal/commit/18bdfddcbedf05331a47d628f0b1f4be60a52985 which ensures that `DynamicNewInstanceNode` with a constant class is always constant folded to a `NewInstanceNode`. Please remove this check now (and everything else related to `DynamicNewInstanceNode` in this PR), it should no longer be necessary.

PR #2459, opened 2020-05-14
Debug info test

adinn:
This PR provides a test of the debuginfo capability on Linux. The patch provides an mx command.

    $ mx debuginfotest

which compiles a sample program to a native image and scripts a run of the image under gdb, checking that breaks, backtraces, stepping and line listing work as expected.

n.b. for the command to work it requires gdb to be in the path.

The test succeeds on my Linux x86_64 and AArch64 boxes.

I have also modified the travis config to run this test automatically on Linux aarch64 and x86_64. I have no idea if that works or not (yet :-)

### DIFF
--- a/substratevm/DEBUGINFO.md
+++ b/substratevm/DEBUGINFO.md
@@ -16,14 +16,16 @@ argument to the GenerateDebugInfo option.
 
 The GenerateDebugInfo option also enables caching of sources for any
 JDK runtime classes, GraalVM classes and application classes which can
-be located during native image generation. The cache is created under
-local subdirectory sources. It is used to configure source file search
-path roots for the debugger. Files in the cache are located in a
-directory hierarchy that matches the file path information included in
-the native image debug records. The source cache should contain all
-the files needed to debug the generated image and nothing more. This
-local cache provides a convenient way of making just the necessary
-sources available to the debugger/IDE when debugging a native image.
+be located during native image generation. By default the cache is
+created under local subdirectory sources (a command line option can be
+use dot specifiy an alternative location). It is used to configure
+source file search path roots for the debugger. Files in the cache are
+located in a directory hierarchy that matches the file path
+information included in the native image debug records. The source
+cache should contain all the files needed to debug the generated image
+and nothing more. This local cache provides a convenient way of making
+just the necessary sources available to the debugger/IDE when
+debugging a native image.
 
 The implementation tries to be smart about locating source files. It
 uses the current JAVA_HOME to locate the JDK src.zip when searching
@@ -57,7 +59,23 @@ separator:
         -cp apps/target/hello.jar:apps/target/greeter.jar \
         org.my.Hello
 
-Note that in both the examples above the DebugInfoSourceSearchPath
+By default the cache of application, GraalVM and JDK sources is
+located under local directory sources. Option DebugInfoSourceCacheRoot
+can be used to specify an alternative location for the top level
+directory. As an example, the following variant of the previous
+command specifies the same target but employs an absolute path:
+
+    $ SOURCE_CACHE_ROOT=$PWD/sources
+    $ mx native-image -H:GenerateDebugInfo=1 \
+        -H:DebugInfoSourceCacheRoot=$SOURCE_CACHE_ROOT \
+        -H:DebugInfoSourceSearchPath=apps/hello/target/hello-sources.jar,apps/greeter/target/greeter-sources.jar \
+        -cp apps/target/hello.jar:apps/target/greeter.jar \
+        org.my.Hello
+
+If you specify a root directory that does not yet exist it will be
+created during population of the cache.
+
+Note that in all the examples above the DebugInfoSourceSearchPath
 options are actually redundant. In the first case the classpath
 entries for apps/hello/classes and apps/greeter/classes will be used
 to derive the default search roots apps/hello/src and
@@ -65,6 +83,7 @@ apps/greeter/src. In the second case classpath entries
 apps/target/hello.jar and apps/target/greeter.jar will be used to
 derive the default search roots apps/target/hello-sources.jar and
 apps/target/greeter-sources.jar.
+
 
 What is currently implemented
 -----------------------------
--- a/substratevm/DEBUGINFO.md
+++ b/substratevm/DEBUGINFO.md
@@ -18,7 +18,7 @@ The GenerateDebugInfo option also enables caching of sources for any
 JDK runtime classes, GraalVM classes and application classes which can
 be located during native image generation. By default the cache is
 created under local subdirectory sources (a command line option can be
-use dot specifiy an alternative location). It is used to configure
+used to specifiy an alternative location). It is used to configure
 source file search path roots for the debugger. Files in the cache are
 located in a directory hierarchy that matches the file path
 information included in the native image debug records. The source
--- a/substratevm/mx.substratevm/mx_substratevm.py
+++ b/substratevm/mx.substratevm/mx_substratevm.py
@@ -245,6 +245,7 @@ class Tags(set):
 GraalTags = Tags([
     'helloworld',
     'helloworld_debug',
+    'debuginfotest',
     'test',
     'maven',
     'js',
@@ -367,6 +368,16 @@ def svm_gate_body(args, tasks):
                 helloworld(['--output-path', svmbuild_dir(), '--shared', '-H:GenerateDebugInfo=1'])  # Build and run helloworld as shared library
                 cinterfacetutorial(['-H:GenerateDebugInfo=1'])
                 clinittest([])
+
+        with Task('image debuginfotest', tasks, tags=[GraalTags.debuginfotest]) as t:
+            if t:
+                if svm_java8():
+                    javac_image(['--output-path', svmbuild_dir()])
+                    javac_command = ['--javac-command', ' '.join(javac_image_command(svmbuild_dir()))]
+                else:
+                    # Building javac image currently only supported for Java 8
+                    javac_command = []
+                debuginfotest(['--output-path', svmbuild_dir()] + javac_command)
 
         with Task('native unittests', tasks, tags=[GraalTags.test]) as t:
             if t:
@@ -700,6 +711,36 @@ def _helloworld(native_image, javac_command, path, build_only, args):
         if actual_output != expected_output:
             raise Exception('Unexpected output: ' + str(actual_output) + "  !=  " + str(expected_output))
 
+def _debuginfotest(native_image, javac_command, path, build_only, args):
+    mkpath(path)
+    parent = os.path.dirname(path)
+    print("parent=%s"%parent)
+    sourcepath = join(parent, 'src/com.oracle.svm.test/src/')
+    print("sourcepath=%s"%sourcepath)
+    sourcecache = join(path, 'sources')
+    print("sourcecache=%s"%sourcecache)
+    hello_source = join(sourcepath, 'hello/Hello.java')
+    print('javac %s'%(['-sourcepath', sourcepath, '-d', path, hello_source]))
+    mx.run(javac_command + ['-sourcepath', sourcepath, '-d', path, hello_source])
+
+    javaProperties = {}
+    for dist in suite.dists:
+        if isinstance(dist, mx.ClasspathDependency):
+            for cpEntry in mx.classpath_entries(dist):
+                if hasattr(cpEntry, "getJavaProperties"):
+                    for key, value in cpEntry.getJavaProperties().items():
+                        javaProperties[key] = value
+    for key, value in javaProperties.items():
+        args.append("-D" + key + "=" + value)
+
+    native_image_args = ["--native-image-info", "-H:Path=" + path, '-H:+VerifyNamingConventions', '-cp', path, '-Dgraal.LogFile=graal.log', '-H:GenerateDebugInfo=1', '-H:DebugInfoSourceSearchPath=' + sourcepath, '-H:DebugInfoSourceCacheRoot=' + join(path, 'sources'), 'hello.Hello'] + args
+    print('native_image {}'.format(native_image_args))
+    native_image(native_image_args)
+
+    if mx.get_os() == 'linux' and not build_only:
+        mx.run(['gdb', '-d', join(sourcecache, 'src'), '-d', join(sourcecache, 'graal'), '-d', join(sourcecache, 'jdk'), '-x', join(parent, 'mx.substratevm/testhello.py'), join(path, 'hello.hello')])
+
+
 def _javac_image(native_image, path, args=None):
     args = [] if args is None else args
     mkpath(path)
@@ -1007,6 +1048,29 @@ def helloworld(args, config=None):
         build_if_missing=True
     )
 
+
+@mx.command(suite_name=suite.name, command_name='debuginfotest', usage_msg='[options]')
+def debuginfotest(args, config=None):
+    """
+    builds a debuginfo Hello native image and tests it with gdb.
+    """
+    parser = ArgumentParser(prog='mx debuginfotest')
+    all_args = ['--output-path', '--javac-command', '--build-only']
+    masked_args = [_mask(arg, all_args) for arg in args]
+    parser.add_argument(all_args[0], metavar='<output-path>', nargs=1, help='Path of the generated image', default=[svmbuild_dir(suite)])
+    parser.add_argument(all_args[1], metavar='<javac-command>', help='A javac command to be used', default=mx.get_jdk().javac)
+    parser.add_argument(all_args[2], action='store_true', help='Only build the native image', default=False)
+    parser.add_argument('image_args', nargs='*', default=[])
+    parsed = parser.parse_args(masked_args)
+    javac_command = unmask(parsed.javac_command.split())
+    output_path = unmask(parsed.output_path)[0]
+    build_only = parsed.build_only
+    native_image_context_run(
+        lambda native_image, a:
+            _debuginfotest(native_image, javac_command, output_path, build_only, a), unmask(parsed.image_args),
+        config=config,
+        build_if_missing=True
+    )
 
 @mx.command(suite.name, 'cinterfacetutorial', 'Runs the ')
 def cinterfacetutorial(args):
--- a/substratevm/mx.substratevm/mx_substratevm.py
+++ b/substratevm/mx.substratevm/mx_substratevm.py
@@ -371,13 +371,7 @@ def svm_gate_body(args, tasks):
 
         with Task('image debuginfotest', tasks, tags=[GraalTags.debuginfotest]) as t:
             if t:
-                if svm_java8():
-                    javac_image(['--output-path', svmbuild_dir()])
-                    javac_command = ['--javac-command', ' '.join(javac_image_command(svmbuild_dir()))]
-                else:
-                    # Building javac image currently only supported for Java 8
-                    javac_command = []
-                debuginfotest(['--output-path', svmbuild_dir()] + javac_command)
+                debuginfotest(['--output-path', svmbuild_dir()])
 
         with Task('native unittests', tasks, tags=[GraalTags.test]) as t:
             if t:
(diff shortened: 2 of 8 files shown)


adinn:
@olpaw I am unable to get this to work sing automation. However, I can run mx debuginfo on my local Linux x86/aarch64 machines and use it to verify that debuginfo is working as expected.

I am not really sure of all that is going wrong with the automation. There are 3 failing gate tasks. the first is the style,fullbuild jdk8 task. I see that pylint is expecting python 2 code whereas the script  (mx.svmbuild/testhello.py) that is passed to gdb to exercise the debugger is (indeed, has to be) python 3. That explains why that task is failing.

Do you know of any way to get round the python mismatch?

I originally added two other gate tasks to try to run fullbuild,debuginfotest using jdk11 on Linux/x86_64 and Linux/aarch64. Neither worked. I don't really undertsand why. can Travis not currently run native image generation on jdk11?

I switched the x86_64 task to use jdk8. It appears to run the fullbuild and the task succeeds. However, there is no evidence that it is attempting to even execute the debuginfo task (there should be printout for the build of natiev program hello.hello and also output from gdb). Is this perhaps because the task requires the gdb package to be installed and gdd is not available?


I added two new travis tasks to run the debug

olpaw:
> Do you know of any way to get round the python mismatch?

AFAIK the Python version GDB uses depends on what Python was passed to the `--with-python=` configure option when that GDB got built. I know that newer Ubuntus build GDB with Python 3 now (but for Fedora I can remember that it still had a GDB built with Python 2 not so long ago). I would not be surprised if the machine that runs the Travis gate has a GDB of such kind.

A simple solution would be to build a well defined GDB + Python combination for the travis gate that gets downloaded during gate execution with wget (as we do it e.g. for eclipse-SDK-4.5.2-linux-gtk-x86_64.tar.gz) and use that for testing. To build such a GDB you'd approximately use
```
cd Python
./configure --prefix=${HOME}/.local/gdb-for-svm
make && make install
```
then for building GDB
```
cd binutils-gdb
./configure --prefix=${HOME}/.local/gdb-for-svm --with-python=${HOME}/.local/gdb-for-svm/bin/python3
make && make install
```
and create a tarball for ${HOME}/.local/gdb-for-svm that we put somewhere so that wget can download it. Ideally we would have that somewhere on https://github.com/graalvm (e.g. having a `gdb-python-builder` repo with releases there) cc @dougxc

olpaw:
>  I see that pylint is expecting python 2 code whereas the script (mx.svmbuild/testhello.py) that is passed to gdb to exercise the debugger is (indeed, has to be) python 3. That explains why that task is failing.

That one is easy. Add
```
# pylint: skip-file
```
at the top of `testhello.py`

olpaw:
> AFAIK the Python version GDB uses depends on what Python was passed to the `--with-python=` configure option when that GDB got built. I know that newer Ubuntus build GDB with Python 3 now (but for Fedora I can remember that it still had a GDB built with Python 2 not so long ago). I would not be

Looking at `.travis.yml` shows that Ubuntu trusty is used for our Travis gates.
If you add
```
      - gdb
```
to
```
addons:
  apt:
    packages:
```
you should get the following GDB:
https://launchpad.net/ubuntu/trusty/+source/gdb (a gdb 7.7.1 built against Python 3.4)
Maybe that is already sufficient for your testing.

adinn:
> I If you add
> 
>       - gdb
> to
> 
> addons:
>   apt:
>     packages:
> 
> you should get the following GDB:
> https://launchpad.net/ubuntu/trusty/+source/gdb (a gdb 7.7.1 built against Python 3.4)
> Maybe that is already sufficient for your testing.

I already added that to .travis.yml (it was done in the 3rd commit). so, if that means gdb is available on the Ubuntu ghost then somethign else must be stopping the debuginfotest task from being run. I don't understand why because I can run it locally using command mx debuginfotest.
(by the way, I am sure gdb 7.7 will test the code ok).

Is there any way I can either 1) stop pylint checking testhello.py or 2) make the style task check testhello.py with a python3 compatible version of pylint?

olpaw:
> Is there any way I can either 1) stop pylint checking testhello.py

See https://github.com/oracle/graal/pull/2459#issuecomment-630124441

adinn:
n.b. I looked up in the release notes for gdb on Ubuntu. 7.7.1 was used in the Trusty release and it uses python3 by default ()Matthias Klose explicitly flagged this).

olpaw:
> n.b. I looked up in the release notes for gdb on Ubuntu. 7.7.1 was used in the Trusty release and it uses python3 by default ()Matthias Klose explicitly flagged this). fedora has been using python3 for just as long.

I use Fedora on my host since forever and I remember when I was testing a Ubuntu container and I was surprised that they had already switched GDB to Python 3 when my host GDB was still using 2. But that happened years ago so might remember it wrong. No offense. Fedora is awesome ;-)

adinn:
I think I am pretty close to having this working now.

What I had failed to do was hook up the run_gate command to the debuginfotest command (I eventually worked through the script code and identified how this works -- Python is not my forte :-/). I van run mx gate debuginfotest succesfully on my local machine now.

I also had to reformat the test class code to keep eclipse happy which meant adjusting the few cases where I check for specific line numbers in listings and stack backtraces.

One thing is still an issue. I had a problem scheduling fullbuild,debuginfotest on Travis -- jdk11u fullbuild seems to fail on both AArch64 or x86_64. So, I am currently trying to run build,debuginfotest on jdk8u and jdk11u just for x86_64. If that works ok I will then see if I can switch the jdk11u task to AArch64. It would be good to cover both arches and both jdks.

adinn:
@olpaw I think this patch is now complete and ready for a review.

The new "build,debuginfotest" task scheduled for jdk8-Linux-x86_64 successfully verifies the debuginfo.

The three Travis failures are for unrelated issues (2 x style, apt get failure). In the first case my python script changes pass the pylint style check; the problem is cased by an unrelated Java style issue. In the 2nd and 3rd cases the outcome of the task is nothing to do with the changes I made.

I cannot yet run the debuginfotest task for jdk11-Linux-x86_64 or jdk11-Linux-aarch64. The jdk11 build for PRIMARY=substratevm always seems to fail with missing JVM_* symbols (I noticed the existing jdk11/substratevm task is commented out in .travis.yml). Once that is fixed I will clone the jdk8 task and edit the vm and arch to test these cases.

olpaw:
> The new "build,debuginfotest" task scheduled for jdk8-Linux-x86_64 successfully verifies the debuginfo.

Excellent. I should be able to get this merged next week.

adinn:
Thanks very much Paul. I will be very glad to have a sanity test in the gate.

substratevm/DEBUGINFO.md
```diff
@@ -16,14 +16,16 @@ argument to the GenerateDebugInfo option.
...
-local subdirectory sources. It is used to configure source file search
-path roots for the debugger. Files in the cache are located in a
-directory hierarchy that matches the file path information included in
-the native image debug records. The source cache should contain all
-the files needed to debug the generated image and nothing more. This
-local cache provides a convenient way of making just the necessary
-sources available to the debugger/IDE when debugging a native image.
+be located during native image generation. By default the cache is
+created under local subdirectory sources (a command line option can be
+use dot specifiy an alternative location). It is used to configure
```
olpaw, line 21: `s/use dot/used to`

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -700,6 +711,36 @@ def _collector(x):
         if actual_output != expected_output:
             raise Exception('Unexpected output: ' + str(actual_output) + "  !=  " + str(expected_output))
 
+def _debuginfotest(native_image, javac_command, path, build_only, args):
+    mkpath(path)
+    parent = os.path.dirname(path)
+    print("parent=%s"%parent)
```
olpaw, line 717: replace all `print` with `mx.log`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -469,4 +469,14 @@ protected void onValueUpdate(EconomicMap<OptionKey<?>, Object> values, Integer o
     @Option(help = "Search path for source files for Application or GraalVM classes (list of comma-separated directories or jar files)")//
     public static final HostedOptionKey<String[]> DebugInfoSourceSearchPath = new HostedOptionKey<String[]>(null) {
     };
+    @Option(help = "Directory under which to create source file cache for Application or GraalVM classes")//
+    public static final HostedOptionKey<String> DebugInfoSourceCacheRoot = new HostedOptionKey<String>("sources") {
+        @Override
+        protected void onValueUpdate(EconomicMap<OptionKey<?>, Object> values, String oldValue, String newValue) {
+            // disallow null or empty string
+            if (newValue == null || newValue.length() == 0) {
```
olpaw, line 477: It's better to issue an error message here instead of silently fixing a users bad option argument. Otherwise we are breeding such command line invocations.

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -1008,6 +1049,29 @@ def helloworld(args, config=None):
...
+@mx.command(suite_name=suite.name, command_name='debuginfotest', usage_msg='[options]')
+def debuginfotest(args, config=None):
+    """
+    builds a debuginfo Hello native image and tests it with gdb.
+    """
+    parser = ArgumentParser(prog='mx debuginfotest')
+    all_args = ['--output-path', '--javac-command', '--build-only']
+    masked_args = [_mask(arg, all_args) for arg in args]
+    parser.add_argument(all_args[0], metavar='<output-path>', nargs=1, help='Path of the generated image', default=[svmbuild_dir(suite)])
+    parser.add_argument(all_args[1], metavar='<javac-command>', help='A javac command to be used', default=mx.get_jdk().javac)
```
olpaw, line 1061: Providing a custom `javac-command` does not make much sense. Since your test is in the `com.oracle.svm.test` project it will be built automatically as part of running `mx build` anyway (building it again is redundant). You can use `classpath('com.oracle.svm.test')` to get the classpath argument you need to pass to native-image when you want to build `Hello`.

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -700,6 +711,36 @@ def _collector(x):
...
+    mkpath(path)
+    parent = os.path.dirname(path)
+    print("parent=%s"%parent)
+    sourcepath = join(parent, 'src/com.oracle.svm.test/src/')
+    print("sourcepath=%s"%sourcepath)
+    sourcecache = join(path, 'sources')
+    print("sourcecache=%s"%sourcecache)
+    hello_source = join(sourcepath, 'hello/Hello.java')
+    print('javac %s'%(['-sourcepath', sourcepath, '-d', path, hello_source]))
+    mx.run(javac_command + ['-sourcepath', sourcepath, '-d', path, hello_source])
```
olpaw, line 724: See https://github.com/oracle/graal/pull/2459/files#r430217539

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -700,6 +711,36 @@ def _collector(x):
         if actual_output != expected_output:
             raise Exception('Unexpected output: ' + str(actual_output) + "  !=  " + str(expected_output))
 
+def _debuginfotest(native_image, javac_command, path, build_only, args):
+    mkpath(path)
+    parent = os.path.dirname(path)
+    print("parent=%s"%parent)
+    sourcepath = join(parent, 'src/com.oracle.svm.test/src/')
```
olpaw, line 718: No need to hardcode `src/com.oracle.svm.test/src/`. You can use `mx.project('com.oracle.svm.test').source_dirs()`

substratevm/mx.substratevm/testhello.py
```diff
@@ -0,0 +1,227 @@
+# pylint: skip-file
+#
+# A test script for use from gdb. It can be sued to drive execution of
```
olpaw, line 3: `s/sued/used`

substratevm/mx.substratevm/testhello.py
```diff
@@ -0,0 +1,227 @@
...
+    # Create a checker to check gdb command output text.
+    # name - string to help identify the check if we have a failure.
+    # regexps - a list of regular expressions which must match.
+    # successive lines of checked
+    def __init__(self, name, regexps):
+        self.name = name
+        compiled = []
+        if not isinstance(regexps, list):
+            regexps = [regexps]
+        for regexp in regexps:
```
olpaw, line 37: you could use
```
compiled = [re.compile(regexp) for regexp in regexps]
```

substratevm/mx.substratevm/testhello.py
```diff
@@ -0,0 +1,227 @@
...
+    # match the corresponding regexp patterns provided when this
+    # Checker was created.
+    # text - the full output of a gdb comand run by calling
+    # gdb.execute and passing to_string = True.
+    # Exits with status 1 if there are less lines in the text
+    # than regexp patterns or if any line fails to match the
+    # corresponding pattern otherwise prints the text and returns
+    # the set of matches.
+    def check(self, text, skip_fails=True):
+        lines = text.split('\n')
```
olpaw, line 51: `s/split('\n')/splitlines()`

substratevm/mx.substratevm/testhello.py
```diff
@@ -0,0 +1,227 @@
...
+#
+
+def test():
+
+    # disable prompting to continue output
+    execute("set pagination off")
+    # set a break point at hello.Hello::main
+    # expect "Breakpoint 1 at 0x[0-9a-f]+: file hello.Hello.java, line 64."
+    exec_string = execute("break hello.Hello::main")
+    rexp = "Breakpoint 1 at 0x([0-9a-f]+): file hello/Hello\\.java, line 64\\."
```
olpaw, line 109: `0x([0-9a-f]+` is a common substring that might be worth reusing .i.e use
```
address_pattern_str = '0x([0-9a-f]+'
```

substratevm/mx.substratevm/testhello.py
```diff
@@ -0,0 +1,227 @@
...
+    checker = Checker('list hello.Hello.Greeter::greeter', rexp)
+    checker.check(exec_string, skip_fails=False)
+
+    # run a backtrace
+    # expect "#0  hello.Hello.greeter::greeter(java.lang.String[]).* at hello.Hello.java:31"
+    # expect "#1  0x[0-9a-f]+ in hello.Hello::main(java.lang.String[]).* at hello.Hello.java:64"
+    # expect "#2  0x[0-9a-f]+ in com.oracle.svm.core.code.IsolateEnterStub::JavaMainWrapper_run_.* at [a-z/]+/JavaMainWrapper.java:[0-9]+"
+    exec_string = execute("backtrace")
+    checker = Checker("backtrace hello.Hello.Greeter::greeter",
+                      ["#0[ \t]+hello\\.Hello\\.Greeter::greeter\\(java\\.lang\\.String\\[\\]\\).* at hello/Hello\\.java:31",
```
olpaw, line 173: Using raw strings will make you patterns much more readable.
https://docs.python.org/3/library/re.html?highlight=raw%20string%20notation

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
adinn: Ok. What would be the appropriate way to make the user aware of the error? Do I throw an exception to stop the native image run? Should Can log an error message and force an exit? Or do I log a warning message, reset the value to the default and carry on. Is there a standard API to implement the latter actions?

olpaw: See e.g. `NativeImageOptions.getCStandard()` on how we usually do this.

adinn: Ok, thanks. Got it.

substratevm/mx.substratevm/testhello.py
olpaw: After this change the following bits
```
compiled = []
for regexp in regexps:
    compiled.append(re.compile(regexp))
```
are dead code. Please remove them.

substratevm/mx.substratevm/testhello.py
```diff
@@ -101,12 +101,20 @@ def execute(command):
 
 def test():
 
+    # define some useful patterns
+    address_pattern = '0x[0-9a-f]+'
```
olpaw, line 105: :+1:

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -470,13 +472,13 @@ protected void onValueUpdate(EconomicMap<OptionKey<?>, Object> values, Integer o
...
-    public static final HostedOptionKey<String> DebugInfoSourceCacheRoot = new HostedOptionKey<String>("sources") {
-        @Override
-        protected void onValueUpdate(EconomicMap<OptionKey<?>, Object> values, String oldValue, String newValue) {
-            // disallow null or empty string
-            if (newValue == null || newValue.length() == 0) {
-                DebugInfoSourceCacheRoot.update(values, oldValue);
-            }
+    public static final HostedOptionKey<String> DebugInfoSourceCacheRoot = new HostedOptionKey<String>("sources");
+
+    public static String getDebugInfoSourceCacheRoot() {
```
olpaw, line 477: Now that you have that abstraction in-between you could take the checking one level further by returning a `Path` here and react on `InvalidPathException` during construction with a sensible `UserError.abort` message.

substratevm/DEBUGINFO.md
adinn: done

substratevm/mx.substratevm/mx_substratevm.py
adinn: done

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
adinn: done

substratevm/mx.substratevm/mx_substratevm.py
adinn: done

substratevm/mx.substratevm/mx_substratevm.py
adinn: done

substratevm/mx.substratevm/testhello.py
adinn: done

substratevm/mx.substratevm/testhello.py
adinn: done

substratevm/mx.substratevm/testhello.py
adinn: done

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
adinn: done

adinn:
Thanks for the review, Paul. I believe i have now made all corrections. Just waiting for the Travis jobs to finish without error (debuginfotest was ok last time but there was a lint 'line too long' error in mx_substratevm.py)

adinn:
Yes, the Travis jobs that relate to the patch are now ok (any errors are unrelated).

Thanks for improving the patch and also for the very helpful python lessons :-)

olpaw:
@adinn I had to add a few fixes to make our internal gates happy. Please have a look if you are fine with my changes. https://github.com/oracle/graal/pull/2459/commits/7cf361f6996d6b74c50b3ed7987c80c1e62dd5f9 https://github.com/oracle/graal/pull/2459/commits/93dabb297aaae737380d811c5eeb9ebe31d4607b https://github.com/oracle/graal/pull/2459/commits/92337dfada77f57874e53115ad986404a9855358

adinn:
Hi Paul,

Yes, I saw all the changes going in and they are fine. Thanks for fixing everything.

adinn:
Hi Paul, I am afraid I noticed a problem with one of your changes after this was merged. You added a   comment to Hello.java

    // Checkstyle: stop

which changed the line numbers checked for in the gdb output. So, the debuginfotest gate task is now failing. I'll post a fix asap.

olpaw:
> which changed the line numbers checked for in the gdb output

Arrg ... Sorry.

Issue #2466, opened 2020-05-15
`dlinfo` symbol emitted on Android when building with Truffle language support


lazar-mitrovic:
If `GraalPython` or `TruffleRuby` are used while building native-image for Android using Gluon Substrate, application will crash in runtime with:
```
java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "dlinfo" referenced by ...
```
This is due to optimization introduced in https://github.com/oracle/graal/commit/c9c0b98 which assumes that isolated namespaces are supported on all `Platform.LINUX` regardless of LibC used. 

Since this behavior is changed in https://github.com/oracle/graal/commit/b91269f I believe that proper fix would be to add Android's Bionic LibC class which should skip following check:

https://github.com/oracle/graal/blob/bbbcebe904b3e5fd802d10e7e6028245aa82f28b/substratevm/src/com.oracle.svm.truffle.nfi.posix/src/com/oracle/svm/truffle/nfi/posix/PosixTruffleNFIFeature.java#L145-L149

and fall straight to the dlopen in else branch.

Example usecase for this fix would be [GraalVM REPL](https://github.com/lazar-mitrovic/GraalREPL ) which due to this issue cannot be built from the master atm.

PR #2467, opened 2020-05-15
Add initial support for Bionic LibC

lazar-mitrovic:
This PR fixes https://github.com/oracle/graal/issues/2466

Note that unlike MuslC, path to LibC isn't required here since all linking is done externally in Gluon Substrate using native Android toolchain.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
@@ -435,6 +435,9 @@ public class SubstrateOptions {
     @Option(help = "file:doc-files/UseMuslCHelp.txt", type = OptionType.Expert)//
     public static final HostedOptionKey<String> UseMuslC = new HostedOptionKey<>(null);
 
+    @Option(help = "When set to true, sets the internally used libc to Bionic. Note that this does not currently download and link against Bionic libc, but serves as a workaround that makes it possible externally", type = OptionType.Expert)//
+    public static final HostedOptionKey<Boolean> UseBionicC = new HostedOptionKey<>(false);
+
     @Option(help = "When set to true, the image generator verifies that the image heap does not contain a home directory as a substring", type = User)//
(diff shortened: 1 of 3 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -435,6 +435,9 @@ public static boolean useLLVMBackend() {
     @Option(help = "file:doc-files/UseMuslCHelp.txt", type = OptionType.Expert)//
     public static final HostedOptionKey<String> UseMuslC = new HostedOptionKey<>(null);
 
+    @Option(help = "When set to true builds a native image using Bionic as the libc implementation", type = OptionType.Expert)//
```
vjovanov, line 438: comma after true

lazar-mitrovic: Fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -435,6 +435,9 @@ public static boolean useLLVMBackend() {
     @Option(help = "file:doc-files/UseMuslCHelp.txt", type = OptionType.Expert)//
     public static final HostedOptionKey<String> UseMuslC = new HostedOptionKey<>(null);
 
+    @Option(help = "When set to true, builds a native image using Bionic as the libc implementation", type = OptionType.Expert)//
```
gradinac, line 438: We should explicitly state that this does actually download and link against bionic libraries yet, but is used as a workaround to be able to do so externally.
We can change it to: "When set to true, sets the internally used libc to Bionic. Note that this does not currently download and link against Bionic libc, but serves as a workaround that makes it possible externally"

lazar-mitrovic: Fixed.

PR #2471, opened 2020-05-18
AArch64: Add Math.multiplyHigh intrinsic on AArch64.

XiaohongGong:
Math.multiplyHigh() could be implemented with a single instruction "smulh" on AArch64. This patch adds the intrinsic support for it when the JDK version is greater than 9.

Here is the main code of a JMH benchmark:
```
  @Benchmark
  public void jmhTimeMultiplyHigh() {
    long res = 0;
    for (int i = 0; i < 64; ++i) {
      long a = rand_long[i % 1024];
      long b = rand_long[(i + 1) % 1024];
      res += Math.multiplyHigh(a, b);
    }
    res_long = res;
  }
```

With this patch, it has about` 346% ~ 840%` performance improvement on different AArch64 machines.

Change-Id: I4758849d82895f4df708711af26ec8935949aafe

### DIFF
--- a/compiler/mx.compiler/suite.py
+++ b/compiler/mx.compiler/suite.py
@@ -1262,6 +1262,24 @@ suite = {
       "javaCompliance" : "9..11",
     },
 
+    "org.graalvm.compiler.replacements.jdk10.test" : {
+      "testProject" : True,
+      "subDir" : "src",
+      "sourceDirs" : ["src"],
+      "dependencies" : [
+        "org.graalvm.compiler.replacements.test"
+      ],
+      "requiresConcealed" : {
+        "jdk.internal.vm.ci" : [
+          "jdk.vm.ci.meta",
+          "jdk.vm.ci.code",
+          "jdk.vm.ci.aarch64",
+        ],
+      },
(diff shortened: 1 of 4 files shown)


XiaohongGong:
Hi, could you please take a look at this simple PR? The error of travis-ci is from the style check, which is not caused by my patch. Thanks so much!

Valchap:
Shouldn't you check if JAVA_SPEC >= 9 instead of 10 ?

XiaohongGong:
> Shouldn't you check if JAVA_SPEC >= 9 instead of 10 ?

I checked the jdk history and found this intrinsic is created after jdk10. And it's also aligned with the `"isJDK10OrHigher()"` in `CheckGraakIntrinsics.java`.

sanzinger:
Thank you for the contribution. Looks good to me. Will start the merge process.

XiaohongGong:
Hi @sanzinger  Thanks so much for your review!

sanzinger:
FYI: I have rebased the changes.

XiaohongGong:
Hi @sanzinger thanks for taking look at it and all the help!

PR #2483, opened 2020-05-20
Search cl.exe version string so that it works for non-English setups

olpaw:


mbbee:
Plz provide a better patch ... this is not robust ...
And build should not fail ...

PR #2489, opened 2020-05-22
AArch64: Add integer Math.abs intrinsics

XiaohongGong:
Math.abs for int/long type could be implemented with "cmp+cneg" on AArch64, which has a better performance than the current pattern:
```
    tbnz w1, #31, label
    mov w0, w1
    ...
  label:
    neg w0, w1
```
This patch adds the intrinsic support for the integer Math.abs to generate the `"AbsNode"` that will emit `"cmp+cneg"` finally.

Here is the main code of two micro benchmarks:
```
  @Benchmark
  public void jmhAbsInt() {
    int res = 0;
      int a = rand_int[(i) % NUM_RANDS];
      res += Math.abs(a) * Math.abs(a);
    }
    res_int = res;
  }

  @Benchmark
  public void jmhAbsLong() {
    long res = 0;
      long a = rand_long[i % NUM_RANDS];
      res += Math.abs(a) * Math.abs(a);
    }
    res_long = res;
  }
```
And the performance results on an A72 AArch64 machine:
Before:
```
Benchmark   Mode  Cnt    Score    Error  Units
jmhAbsInt   avgt   25  335.121 ±  8.971  ns/op
jmhAbsLong  avgt   25  327.854 ± 15.797  ns/op
```

After:
```
Benchmark   Mode  Cnt    Score   Error  Units
jmhAbsInt   avgt   25   59.390 ± 0.256  ns/op
jmhAbsLong  avgt   25  104.772 ± 0.459  ns/op
```
Change-Id: I73a3aa66153fc2f023d6151a82f199d73824f04b

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/CheckGraalIntrinsics.java
+++ b/compiler/src/org.graalvm.compiler.hotspot.test/src/org/graalvm/compiler/hotspot/test/CheckGraalIntrinsics.java
@@ -398,9 +398,12 @@ public class CheckGraalIntrinsics extends GraalTest {
         }
 
         if (isJDK13OrHigher()) {
+            if (!(arch instanceof AArch64)) {
+                add(toBeInvestigated,
+                                "java/lang/Math.abs(I)I",
+                                "java/lang/Math.abs(J)J");
+            }
             add(toBeInvestigated,
-                            "java/lang/Math.abs(I)I",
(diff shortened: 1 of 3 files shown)


XiaohongGong:
Hi @sanzinger could you please take a look at this simple patch? The errors of travis-ci is not caused by my patch. Thank you so much!

sanzinger:
Looks good. Will be merged asap.

xiagon01:
Hi @sanzinger  Thanks for looking at it!

Issue #2493, opened 2020-05-22
GraalVM generates wrong result due to speculative optimization


helloguo:
**Describe the issue**
GraalVM generates wrong result due to speculative optimization. 

**Steps to reproduce the issue**
Please include both build steps as well as run steps
1. Run this code
If you cannot reproduce it, you can simply increase the `int size = 1000000` to a lager value in `main()` function to make the repro work.
```
import java.util.HashMap;

class InputRow {
  public long[] row;
  public int len;

  public InputRow(int size, long initValue) {
    len = size;
    row = new long[size];
    for (int i = 0; i < size; i++) {
      row[i] = initValue;
    }
  }

  public long getLong(int idx) {
    assert (idx < len);
    return row[idx];
  }
}

class SimpleInput {
  public InputRow[] rows;
  public int len;
  public int idx;

  public SimpleInput(int size) {
    len = size;
    rows = new InputRow[size];
    idx = 0;
    for (int i = 0; i < size - 3; i = i + 3) {
      rows[i] = new InputRow(10, 50);
      rows[i + 1] = new InputRow(10, -1);
      rows[i + 2] = new InputRow(10, 100);
    }
    rows[size - 1] = new InputRow(10, 10);
  }

  public boolean hasNext() {
    return idx < len;
  }

  public InputRow next() {
    InputRow ret = rows[idx];
    idx++;
    return ret;
  }
}

class OutputElement {
  public long value1;
  public String value2;

  public OutputElement() {
    value1 = 0;
    value2 = "init";
  }

  public void write(int idx, long value) {
    assert (idx == 0);
    value1 = value;
  }

  public void write(int idx, String value) {
    assert (idx == 1);
    value2 = value;
  }
}

class SimpleOutput {
  public OutputElement[] output;
  public int len;
  public int idx;

  public SimpleOutput(int size) {
    output = new OutputElement[size];
    len = size;
    idx = 0;
  }

  public void append(OutputElement element) {
    assert (idx < len);
    output[idx] = element;
    idx++;
  }

  public void print() {
    HashMap<String, Integer> map = new HashMap<String, Integer>();
    for (int i = 0; i < len; i++) {
      long v1 = output[i].value1;
      String v2 = output[i].value2;
      if (map.containsKey(v2)) {
        int tmp = map.get(v2);
        map.put(v2, (tmp + 1));
      } else {
        map.put(v2, 1);
      }
    }
    System.out.print(map);
  }
}

public class Main {
[...]

PR #2511, opened 2020-05-27
Fix issue #2493: stop reordering two IF statemenets if one of their condition is signe…

lic9:
…d and the other is unsigned.


lic9:
Hi @dougxc, could you please help take a look? Thanks!

dougxc:
Thanks for the PR @lic9 . However, @tkrodriguez already has a fix in the pipeline that will address this problem.

dougxc:
Correction: it is already merged in 2dfa6ab92a02f4c018376190ced992be9dd066e3

lic9:
That’s much better! Thanks @dougxc!

helloguo:
Any plan to backport the fix 2dfa6ab into 19.3.1?

dougxc:
I think 19.3.3 would be the next back-port target right @gilles-duboscq?

gilles-duboscq:
Indeed

dougxc:
Ok, just confirming, it will be in 19.3.3.

PR #2499, opened 2020-05-25
Support cl.exe detection for Simplified-Chinese and other languages

olpaw:

Issue #2480, opened 2020-05-19
20.1.0 native-image-maven-plugin: Error: Unable to detect supported WINDOWS native software development toolchain.


mbbee:
With version 20.0.0 everything worked well with windows, but now:
```
Error: Unable to detect supported WINDOWS native software development toolchain.
com.oracle.svm.core.util.UserError$UserException: Unable to detect supported WINDOWS native software development toolchain.
        at com.oracle.svm.core.util.UserError.abort(UserError.java:68)
        at com.oracle.svm.hosted.c.codegen.CCompilerInvoker.<init>(CCompilerInvoker.java:72)
        at com.oracle.svm.hosted.c.codegen.CCompilerInvoker$WindowsCCompilerInvoker.<init>(CCompilerInvoker.java:93)
        at com.oracle.svm.hosted.c.codegen.CCompilerInvoker.create(CCompilerInvoker.java:84)
        at com.oracle.svm.hosted.NativeImageGenerator.setupNativeImage(NativeImageGenerator.java:851)
        at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:536)
        at com.oracle.svm.hosted.NativeImageGenerator.lambda$run$0(NativeImageGenerator.java:451)
        at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1407)
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
        at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
        at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
        at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
        at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
Error: Image build request failed with exit status 1
```

And yes, i use "x64 Native Tools Command Prompt for VS 2017" ...

genarks:
I had the same problem and tested CCompilerInvoker, it looks like it's failing to parse the first line of cl.exe.

https://github.com/oracle/graal/blob/e84af13963d67dfcfe231531a644decaa8b35176/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/codegen/CCompilerInvoker.java#L129

The code is.

[...]

PR #2513, opened 2020-05-28
fixe OS languages - Arch and Version detection

Ferhatos:
Fixe to support for any OS language

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/codegen/CCompilerInvoker.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/codegen/CCompilerInvoker.java
@@ -59,6 +59,9 @@ import jdk.vm.ci.amd64.AMD64;
 import jdk.vm.ci.code.Architecture;
 import jdk.vm.ci.sparc.SPARC;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public abstract class CCompilerInvoker {
 
     public final Path tempDirectory;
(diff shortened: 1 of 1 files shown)


olpaw:
@Ferhatos I don't see why wee need this. See https://github.com/oracle/graal/pull/2499#issuecomment-635246987

Ferhatos:
Hello @olpaw ,

I've seen that issue still there for German OS as commented, and that there is some specific code for a specific OS language. This PR is to be more generic without specific code, for any OS language environment.

Still don't agree ?

olpaw:
> I've seen that issue still there for German OS as commented, and that there is some specific code for a specific OS language. This PR is to be more generic without specific code, for any OS language environment.
> 
> Still don't agree ?

Thanks for your PR but so far, I do not see any reason for merging this PR.

I have - of course - also tested my current approach (what we now have on master) with
```
Microsoft (R) C/C++-Optimierungscompiler Version 19.26.28805 für x64
Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.

Syntax: cl [ Option... ] Dateiname... [ /link Linkeroption... ]
```
and it works fine, just like with the French, Simplified-Chinese and all other variants people posted on https://github.com/oracle/graal/issues/2480.

Your PR searches for the architecture string in an unstructured way (`Matcher archMatcher = Pattern.compile("x[0-9]+[0-9]+").matcher(versionMessage);`) which is suboptimal because it only works for the architectures that you have hard-coded so far. What about ARM or ARM64? Are we going to change the pattern every time we support a new architecture?
Also I don't see why the if-cascade for version number parts is better than using scanner with `hasNextInt/nextInt` + custom delimiter.

As I already said before, if it turns out the current approach is insufficient we will instead implement the approach mentioned here: https://github.com/oracle/graal/issues/2480#issuecomment-633469170

Ferhatos:
Hello @olpaw ,
Thank you for your detailed response. I missed other architectures and you're right.
I've seen later **if-cascades**, it can be optimized using _matcher group function_ (`matcher.group(index)`) by adding `()` in **RegExp**.

Thank you.
Any idea when can we use the fixe for languages ?

PR #2516, opened 2020-05-28
Removes cloned fdlibm and uses openJDK's fdlibm instead

### LINKED ISSUE #2514: Using upstream fdlibm
Hello all,

As discussed in https://oss.oracle.com/pipermail/graalvm-dev/2020-May/000059.html

Since `fdlibm` is part of the OpenJDK and can be built as a separate static library (already shipped with labsJDK) there is no apparent reason for GraalVM to keep using an older clone of `fdlibm`.

Could you please provide details on how to properly achieve this?

A _proof of concept_ patch can be seen in https://github.com/zakkak/mandrel/commit/91922abe1cd8ef81fb8a5292644fee521c541bfc 
Based on my experience we will need to annotate `Target_java_lang_StrictMath` with more than one `@CLibrary` to include `fdlibm` which is now part of `strictmath`.

Edit: Replaced discussion link to point to the latest email in the thread.

zakkak:
As discussed in
https://oss.oracle.com/pipermail/graalvm-dev/2020-May/000059.html

Since fdlibm is part of the OpenJDK and can be built as a separate
static library (already shipped with labsJDK) there is no apparent
reason for GraalVM to keep using an older clone of fdlibm.

Closes #2514


zakkak:
Thanks for the review @christianwimmer.

I have resolved all your comments.

Could we, however, at least allow for a single year to be used in the copyright as in other modules? (e.g. https://github.com/oracle/graal/blob/master/compiler/src/org.graalvm.compiler.graph/.checkstyle_checks.xml#L238)

` * Copyright (c) 2020, 2020, Red Hat Inc. All rights reserved.` is a bit weird :)

christianwimmer:
Thanks for the changes, looks all good now and should be merged soon.

> Could we, however, at least allow for a single year to be used in the copyright as in other modules?

That is an interesting detail question where people have different opinions and I don't have a definitive legal answer. For now, all SVM projects consistently always have two years, if we change that then we need to change it in all source files.

PR #2521, opened 2020-05-29
native junit tests for java.lang.StrictMath

zakkak:
Adds tests from http://hg.openjdk.java.net/jdk-updates/jdk11u/file/90ce1674f838/test/jdk/java/lang/StrictMath/ to ease testing for #2514


zakkak:
I am aware that we are not supposed to edit checkstyle rules, but I didn't know how else to stay compliant with the existing copyright notices. Please advise.

zakkak:
Hi @christianwimmer 

This is a kind reminder :)

christianwimmer:
Sorry for the delay.

Usually adding tests is easy to approve. But in this case, I'm questioning the need: why copy more than 4000 lines from one repository to another repository? If someone wants to run these tests, that can be done by getting them from the original location, where they are maintained and updated if something changes in the strictmath implementation. Since now SVM now no longer contains implementations of the strictmath functions, why add the tests?

zakkak:
Hi Christian,

Good point. I agree that duplicating the tests doesn't seem wise in the long run.

Issue #2523, opened 2020-05-30
[native-image] provide more information about missing JNI Method


johanvos:
**Describe the issue**
When native wants to invoke a Java method, the corresponding methodID is searched via `JNIFunctions.getMethodID`. In case the class holding this method is not registered for JNI, this `getMethodID` will throw a NullPointerException without information about the method we want to invoke.

This can be improved by checking on `hclazz == null` and in that case, the name and signature of the method can be printed.

johanvos:
PR #2524 has a fix for this.

peter-hofer:
Closing: the JNI specification requires `must not be NULL`. HotSpot does not have any conservative checks either.

PR #2524, opened 2020-05-30
Add info about the requested method name and signature

johanvos:
Add info about the requested method name and signature in case clazz is null (i.e. not registered).

Fix for #2523


peter-hofer:
Thanks for your contribution @johanvos. I think this behavior should be enabled by a separate option, for example `VerboseJNI`. It is a legitimate use case for JNI code to probe using `GetMethodID`, `GetStaticMethodID` and similar functions, for example to detect overriding members in a subclass, and that should not affect the program's output. Also, we already build a string for the `NoSuchMethodError` in the `getMethodID(Class<?> clazz, ...` which could be used here, and it would be great if it also worked with `FindClass`, `GetFieldID` and `GetStaticFieldID`.

peter-hofer:
Closing due to no activity.

PR #2537, opened 2020-06-03
Fix assertion error for debug builds on AArch64

SwapnilGaikwad:
Fix a minor typo in the Arithmetic LIR generator for AArch64 that
avoids assertion errors when they are enabled (using -ea flag).

### DIFF
--- a/compiler/mx.compiler/suite.py
+++ b/compiler/mx.compiler/suite.py
@@ -803,7 +803,6 @@ suite = {
         "jdk.internal.vm.ci" : [
           "jdk.vm.ci.meta",
           "jdk.vm.ci.code",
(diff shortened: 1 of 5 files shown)


SwapnilGaikwad:
> Hello Swapnil Gaikwad, thanks for contributing a PR to our project!
> 
> We use the [Oracle Contributor Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) to make the copyright of contributions clear. We don't have a record of you having signed this yet, based on your email address swapnil -(dot)- gaikwad -(at)- arm -(dot)- com. You can sign it at that link.
> 

Hi team, I'm an employee of Arm and I'll be contributing under their signature. Please let me know if there is anything else I need to do there.

Issue #2490, opened 2020-05-22
Agent access filter does not apply to proxy classes


peter-hofer:
With following `access-filter.json` file:
```
{ "rules": [
  {"excludeClasses": "org.apache.maven.surefire.**"},
  {"excludeClasses": "net.bytebuddy.**"},
  {"excludeClasses": "org.apiguardian.**"},
  {"excludeClasses": "org.junit.**"},
  {"excludeClasses": "org.mockito.**"},
  {"excludeClasses": "org.springframework.test.**"},
  {"excludeClasses": "org.springframework.boot.test.**"}
]
}
```

The related entries are filtered from `reflect-config.json` but `proxy-config.json` contains:
```
[
  ["java.lang.reflect.ParameterizedType","org.springframework.core.SerializableTypeWrapper$SerializableTypeProxy","java.io.Serializable"],
  ["org.springframework.boot.context.properties.ConfigurationProperties","org.springframework.core.annotation.SynthesizedAnnotation"],
  ["org.springframework.boot.test.context.SpringBootTest","org.springframework.core.annotation.SynthesizedAnnotation"],
  ["org.springframework.test.context.BootstrapWith","org.springframework.core.annotation.SynthesizedAnnotation"]
]
```

There is currently no way to filter `proxy-config.json` to get the following content based on previously mentioned `access-filter.json`:
```
[
  ["java.lang.reflect.ParameterizedType","org.springframework.core.SerializableTypeWrapper$SerializableTypeProxy","java.io.Serializable"],
  ["org.springframework.boot.context.properties.ConfigurationProperties","org.springframework.core.annotation.SynthesizedAnnotation"]
]
```

_Originally posted by @sdeleuze in https://github.com/oracle/graal/pull/2422#issuecomment-631043276_

sdeleuze:
Thanks for your feedback on my original comment, glad this could be fixed :-)

Please tag this issue with `spring` label, and it would be awesome if it could be backported to 20.1 branch in order to provide this feature to our users on July 14th via 20.1.1.

eginez:
should be ported to 20.1.1

PR #2544, opened 2020-06-05
Apply access filter to proxy interfaces.

peter-hofer:
Solves  #2490.


peter-hofer:
With this change, the agent will not introduce a proxy configuration entry as soon as one of its proxied interfaces is excluded by the access filter. (Symmetrically, the _restrict_ mode will permit creating any proxy object that contains at least one interface excluded by the access list). Given that we only handle entire permitted _interface lists_ in `proxy-config.json`, I don't see how we can currently cover the use case better.

PR #2547, opened 2020-06-08
Support ManagementFactory.getPlatformMBeanServer

christianwimmer:
Read JavaDoc of class `ManagementSupport` for an overview.


substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/CompleteGarbageCollectorMXBean.java
```diff
@@ -0,0 +1,105 @@
...
+import org.graalvm.nativeimage.Platform;
+import org.graalvm.nativeimage.Platforms;
+
+import com.oracle.svm.core.util.TimeUtils;
+import com.sun.management.GcInfo;
+
+import sun.management.Util;
+//Checkstyle: resume
+
+public final class CompleteGarbageCollectorMXBean implements com.sun.management.GarbageCollectorMXBean, NotificationEmitter {
```
christianwimmer, line 45: Note that the various `*MXBean` implementation classes existed before, they just got moved to their own files. So no need to review them (and no need to complain that many methods are still only stubs that throw an exception).

Issue #2558, opened 2020-06-10
Always support isolates (remove option SpawnIsolates) to simplify GC improvements.


peter-hofer:
Support for isolates using option `SpawnIsolates` has been enabled by default since September 2018, before the 1.0 release. It uses base-relative object references and a continuous image heap area. When disabling isolate support, references are absolute pointers and the image heap is divided between the native image object file's `data` and `rodata` sections, among a number of smaller differences. Providing the option to turn isolate support off requires continuous maintenance and testing efforts and will complicate introducing future GC improvements as proposed in #2386. Therefore, we should consider removing it.

peter-hofer:
Superseded by #2569.

PR #2559, opened 2020-06-10
Always support isolates (remove option SpawnIsolates) to simplify GC improvements.

peter-hofer:
Solves #2558 .


peter-hofer:
@loicottet , LLVM still disabled `SpawnIsolates` for performance reasons. How significant is this performance impact still?

PR #2563, opened 2020-06-12
Remove AArch64 feature fDMBATOMICS in SVM

a74nh:
The fDMBATOMICS CPU feature was only ever implemented on a
prototype CPU and never made it into a production CPU.

See https://bugs.openjdk.java.net/browse/JDK-8242469


sanzinger:
Thank you for your contribution. There are no usages attached to this flag, so it is safe to remove.

PR #2567, opened 2020-06-14
Fix reachability checks in the presence of @GenerateUncached.

kustosz:
Fixes a behavior of the Truffle specialization DSL, where uncached specializations would be marked unreachable, even if they are reachable in the node generated with `@GenerateUncached`.

The logic implemented here is: if `@GenerateUncached` is present, an uncached specialization cannot be shadowed by a cached one.

### DIFF
--- a/truffle/src/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/ReachabilityTest.java
+++ b/truffle/src/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/ReachabilityTest.java
@@ -304,7 +304,7 @@ public class ReachabilityTest {
 
     }
 
-    static abstract class ReachabilityUncached extends Node {
+    abstract static class ReachabilityUncached extends Node {
         abstract int execute();
 
         int foo() {
@@ -313,7 +313,7 @@ public class ReachabilityTest {
     }
 
     @GenerateUncached
-    static abstract class ReachabilityUncached1 extends ReachabilityUncached {
+    abstract static class ReachabilityUncached1 extends ReachabilityUncached {
         @Specialization
         int doCached(@Cached("foo()") int cached) {
             return cached;
@@ -326,7 +326,7 @@ public class ReachabilityTest {
     }
 
     @GenerateUncached
-    static abstract class ReachabilityUncached2 extends ReachabilityUncached {
+    abstract static class ReachabilityUncached2 extends ReachabilityUncached {
         @Specialization
         int doCached(@Cached("foo()") int cached) {
             return cached;
@@ -345,7 +345,7 @@ public class ReachabilityTest {
     }
 
     @GenerateUncached
-    static abstract class ReachabilityUncached3 extends ReachabilityUncached {
+    abstract static class ReachabilityUncached3 extends ReachabilityUncached {
         @Specialization
         int doCached1(@Cached("foo()") int cached) {
             return cached;
--- a/truffle/src/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/ReachabilityTest.java
+++ b/truffle/src/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/ReachabilityTest.java
@@ -42,12 +42,15 @@ package com.oracle.truffle.api.dsl.test;
 
 import java.math.BigInteger;
 
+import com.oracle.truffle.api.dsl.Cached;
+import com.oracle.truffle.api.dsl.GenerateUncached;
 import com.oracle.truffle.api.dsl.NodeChild;
 import com.oracle.truffle.api.dsl.NodeChildren;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.dsl.test.TypeSystemTest.Abstract;
 import com.oracle.truffle.api.dsl.test.TypeSystemTest.BExtendsAbstract;
 import com.oracle.truffle.api.dsl.test.TypeSystemTest.ValueNode;
+import com.oracle.truffle.api.nodes.Node;
 
 public class ReachabilityTest {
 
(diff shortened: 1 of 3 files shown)


kustosz:
I have signed the OCA on Friday and received an email confirmation of it being "accepted and processed as of now" :)

truffle/src/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/model/SpecializationData.java
```diff
@@ -353,10 +355,18 @@ public void setReachable(boolean reachable) {
         this.reachable = reachable;
     }
 
+    public void setUncached(boolean uncached) {
```
chumer, line 358: Maybe rename this field reachableUncached to makle it clearer what it means.
A specialization is not uncached by itself.
At the same time it would make sense to rename SpecializationData#reachable to reachableCached.

truffle/src/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/parser/NodeParser.java
```diff
@@ -1748,6 +1749,13 @@ private void collectIncludes(SpecializationData specialization, Set<Specializati
         }
     }
 
+    private static void initializeUncached(final NodeData node) {
```
chumer, line 1752: I think this could be merged into initializeReachability. It is fine if this one computes cached and uncached reachability.

truffle/src/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/model/SpecializationData.java
kustosz: This would be confusing. The `isUncached` field does not mean "is it reachable in the uncached node?", but rather "is this specialization taken into account when generating the uncached node?". It has nothing to do with actual reachability, just a helper field.
I've only cached it here to avoid some map lookups, but honestly maybe we'd better off just doing those lookups and not add more fields here?

truffle/src/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/parser/NodeParser.java
kustosz: Again, the `intializeReachability` methods computes both kinds of reachability. This one only initializes a helper field.

truffle/src/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/parser/NodeParser.java
```diff
@@ -1576,6 +1576,7 @@ private void initializeSpecializations(List<? extends Element> elements, final N
         initializeUninitialized(node);
         initializeOrder(node);
         initializePolymorphism(node); // requires specializations
+        initializeUncached(node);
```
chumer, line 1579: initializeUncached uses SpecializationData.getReplaces indirectly that gets initialized in resolveReplaces later.

How about we merge initializeUncached with resolveReplaces. And rename SpecializationData#uncached to Specialization#replaced. This probably needs to be a flattened set again in order to fix the failing test.

truffle/src/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/ReachabilityTest.java
```diff
@@ -301,4 +304,62 @@ int do1() throws RuntimeException {
...
+        }
+
+        @Specialization(replaces = "doCached")
+        int doUncached() {
+            return 1;
+        }
+    }
+
+    @GenerateUncached
+    abstract static class ReachabilityUncached2 extends ReachabilityUncached {
```
chumer, line 329: Here is a failing test.

```

 @GenerateUncached
    abstract static class ReachabilityUncached4 extends ReachabilityUncached {

        @Specialization
        int s0(@Cached("foo()") int cached) {
            return cached;
        }

        @Specialization(replaces = "s0")
        int s1() {
            return 1;
        }

        @Specialization(replaces = "s1")
        int s2() {
            return 1;
        }
    }

```

I think that one should work too but doesn't.

truffle/src/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/parser/NodeParser.java
kustosz: Done, indeed makes the code simpler :)

truffle/src/com.oracle.truffle.api.dsl.test/src/com/oracle/truffle/api/dsl/test/ReachabilityTest.java
kustosz: I don't understand what's wrong with this example. It fails with `Specialization is not reachable. It is shadowed by s0(int).`. And that is true – `s1` is replaced and as such is never a part of the uncached node. So normal rules apply here (as if `@GenerateUncached` was not present).

chumer: You are right. I tried somehow to construct a case where the missing handling of transitivity was a problem. But it seems it never was.

truffle/src/com.oracle.truffle.dsl.processor/src/com/oracle/truffle/dsl/processor/parser/NodeParser.java
chumer: Perfect. Thanks!

PR #2575, opened 2020-06-18
Default value in ErrorStateHolder::messageHolder might leak

YaSuenag:
`ErrorStateHolder::messageHolder` holds `CCharPointerHolder` which is instantiated by `CTypeConversionSupportImpl`. If we pass `""` (zero-length string) to `CTypeConversion::toString`, it would be assigned unmanaged memory which have `\0`.

`messageHolder` would be released at `PolyglotNativeAPI::resetErrorState`. `ErrorStateHolder::info` should not be null to happen releasing memory. However `info` is assigned to `WordFactory.nullPointer()` by default, so it would not be happen - default value of `messageHolder` would leak.

`ErrorStateHolder::messageHolder` is overwritten when an error is happen at `PolyglotNativeAPI::handleThrowable`, but it seems to assume all resources in `ErrorStateHolder` are certainly released at `PolyglotNativeAPI::withHandledErrors` (caller of `resetErrorState()`).

### DIFF
--- a/substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
+++ b/substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
@@ -134,7 +134,10 @@ public final class PolyglotNativeAPI {
 
     private static class ErrorStateHolder {
         public PolyglotExtendedErrorInfo info = WordFactory.nullPointer();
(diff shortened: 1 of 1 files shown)


YaSuenag:
Some CI jobs failed, but they seems not to be related my change.

gradinac:
Hey @YaSuenag, thank you for taking the time to contribute to this project!
You are indeed correct, from what I see - we would leak a single byte of memory for each different thread.
The PR fixes the issue, I have squashed your commits and I will merge this PR, though it might not autoclose once merged due to the squash. If that happens, I will close it manually after.

gradinac:
This PR has been squashed and merged (https://github.com/oracle/graal/commit/0b120f3dd16 and https://github.com/oracle/graal/commit/c8901eb038d95eb1ce30ff06f56eb86e317f948c) so I will be closing it.

PR #2580, opened 2020-06-18
 Interface class initialization dependencies does not include its superinterfaces

zhangruoxu:
In the native-image, the class `TypeInitializerGraph` finds the safety of initializing a class at build time. It creates a mapping, whose key is a class, say, `A`, and value is dependencies of initializing `A`, which includes `A`'s super classes and some superinterfaces. If one of `A`'s dependencies is unsafe to be initialized at build time, then `A` itself is promoted to be unsafe. 

Intuitively speaking, the afore-mentioned mapping indicates implicit class initialization introduced by inheritance. Currently, in  `TypeInitializerGraph`, for a class (including interface), its dependencies include its superclasses as well as any superinterfaces that declare any default methods. However, for an interface, according to JLS Section 12.4.1, 

> Initialization of an interface does not, of itself, cause initialization of any of its superinterfaces.

Consequently, including superinterfaces of an interface as dependencies makes that we lose some chance to safely initializing interfaces at build time. 

Let's see an example. 

```java
public class TestInitSuperInterface {
        System.out.println(IA.o);
    }

    static interface IA extends IB {
        Object o = "IA.o";
    }

    static interface IB {
        default void foo() {}
        Object s = bar();

        static String bar() {
            return Long.toString(System.currentTimeMillis());
        }
    }
}
```
In the code above, `IB` cannot be initialized at build time since its `<clinit>()` invokes a native method, i.e.,`System.currentTimeMillis()`. According to JLS, initializing `IA` does not introduce the initialization of `IB`, and `IA`'s `<clinit>()` does not include any unsafe operations so that it is safe to be initialized at build time. 

However, `TypeInitializerGraph` now treats `IB` as `IA`'s dependencies, so that `IA` cannot be initialized at build time.  

In this pull request, superinterfaces are not included in the dependencies of an interface. After applying this pull request, `IA` can be initialized at build time. 

Please feel free to give any comments about this pull request. 

Cheers,
Yifei


zhangruoxu:
Hi @dougxc, I was wondering if you guys could give some comments to this pull request :-)

zhangruoxu:
@vjovanov Dear Vojin, could you give some comments on this modification? Thanks for your time. Cheers :-)

PR #2581, opened 2020-06-18
Fix ENV_VARIABLE_PREFIX

### LINKED ISSUE #2576: `GRAALVM_CATALOG` variable does not work as documented
**Describe the issue**

The [`gu` documentation](https://www.graalvm.org/docs/reference-manual/install-components/#component-installation-modes) states that we can use `GRAALVM_CATALOG` to specify a custom catalog to download Graal components from. I am trying to do just that, and `gu` still tries to download from https://www.graalvm.org/component-catalog/graal-updater-component-catalog-java8.properties

Passing `GRAALVM_CATALOG` as -C argument works, but is not an option in this case, as we need it in the context of a Gradle build, and the plugin we use does not allow for custom `gu` options.

1. Install `graalvm-ce-java8-20.1.0`
2. Create an empty file named `test`
3. Set `export GRAALVM_CATALOG=file://$PWD/test`
4. Run `gu install native-image`

The last step should have failed, because the catalog is empty, but it actually succeeeds. There is more elaborate procedure to reproduce the issue if you have isolated network.

 - This is as of version 20.1.0 and documentation website as of 2020-06-18.
 - JDK major version: 8
 - OS: RHEL 6
 - Architecture: AMD64

**More details**
Add any other information about the problem here. Especially important are stack traces or log output. Feel free to link to gists or to screenshots if necessary.

```
-bash-4.1$ /app/localstorage/dimitrda/caches/com.palantir.graal/20.1.0/8/graalvm-ce-java8-20.1.0/bin/gu install native-image
Downloading: Component catalog from www.graalvm.org
Error: Error downloading component catalog from https://www.graalvm.org/component-catalog/graal-updater-component-catalog-java8.properties: www.graalvm.org. 
Please check your connection and proxy settings. If your machine is behind a proxy, environment variables (http_proxy, https_proxy, ...) must be set appropriately.

-bash-4.1$ set | grep GRAAL
GRAALVM_CATALOG=file:///home/dimitrda/fixer/acme-graal-updater-component-catalog-java8.properties
```

fniephaus:
I believe this fixes #2576.

### DIFF
--- a/vm/src/org.graalvm.component.installer/src/org/graalvm/component/installer/CommonConstants.java
+++ b/vm/src/org.graalvm.component.installer/src/org/graalvm/component/installer/CommonConstants.java
@@ -78,7 +78,7 @@ public class CommonConstants {
 
     public static final String SYSPROP_JAVA_VERSION = "java.specification.version"; // NOI18N
 
-    public static final String ENV_VARIABLE_PREFIX = "GRAAVLM_"; // NOI18N
+    public static final String ENV_VARIABLE_PREFIX = "GRAALVM_"; // NOI18N
 
     /**
      * Env variable that controls catalog URL.


boris-spas:
This seems to have been fixed by another party.

PR #2583, opened 2020-06-18
Add missing GraalVM packages present in JDK source cache

### LINKED ISSUE #2577: Expand JDK source cache filter to explode missing source files
The JDK source cache logic is not extracting classes that are present in lib/src.zip that are stored under modules. They need to be expanded so that native debugging can correctly function. Essentially, the JDK source cache package filter needs to be expanded.

I'll be sending a PR shortly.

galderz:
Closes #2577.

The sources for the packages added are stored in `lib/src.zip`. Hence, they need to be included as part of the sources that are extracted from there.


galderz:
Closing the PR until we have clearer clarification on the divergences between what one can build locally and what one can download. See #2577 for more info.

PR #2584, opened 2020-06-18
[gu] Ensure errors in Graal Updater are displayed

fniephaus:
In case `components.isEmpty()`, errors are not displayed to the user. While debugging, I found out that:
> Unsupported operation in filter. Only = is supported.

is thrown. With this fix in place, I would have seen the error straight away.


fniephaus:
Outdated.

PR #2588, opened 2020-06-19
Add intrinsics for floating point Math.min/Math.max on AArch64

SwapnilGaikwad:
Emit `fmin/fmax` instructions when `Math.min()/Math.max()` function is called with single or double precision floating point values on AArch64. The existing mechanism emits `cmp+csel` combination for the integer variants of `Math.min()/Math.max()`.

Relevant snippet of the microbenchmark is shown below (only the double variant for `Math.max()` is shown for brevity).

```
@Benchmark
public void jmhTimeMaxDouble() {
    double res = 0.0d;
        double a = rand_double[i % NUM_RANDS];
        double b = rand_double[(i + 1) % NUM_RANDS];
        res += Math.max(a, b);
    }
    res_double = res;
}
```

Performance results on an A72 AArch64 machine

Before:
```
Benchmark         Mode  Cnt    Score   Error  Units
jmhTimeMaxDouble  avgt   15  288.019 ± 5.645  ns/op
jmhTimeMaxFloat   avgt   15  283.060 ± 4.367  ns/op
jmhTimeMinDouble  avgt   15  282.940 ± 3.823  ns/op
jmhTimeMinFloat   avgt   15  279.740 ± 2.278  ns/op
```

After:
```
Benchmark         Mode  Cnt   Score   Error  Units
jmhTimeMaxDouble  avgt   15  98.860 ± 0.010  ns/op
jmhTimeMaxFloat   avgt   15  98.847 ± 0.008  ns/op
jmhTimeMinDouble  avgt   15  98.867 ± 0.024  ns/op
jmhTimeMinFloat   avgt   15  98.855 ± 0.016  ns/op
```

Change-Id: Ic3b9de4aa096a8af699cfe5d0f0b8c83ca86b208

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/FloatStamp.java
+++ b/compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/FloatStamp.java
@@ -910,6 +910,77 @@ public class FloatStamp extends PrimitiveStamp {
 
                     null, null, null,
 
+                    new BinaryOp.Max(true, true) {
+
+                        @Override
+                        public Constant foldConstant(Constant const1, Constant const2) {
+                            PrimitiveConstant a = (PrimitiveConstant) const1;
+                            PrimitiveConstant b = (PrimitiveConstant) const2;
+                            assert a.getJavaKind() == b.getJavaKind();
+                            switch (a.getJavaKind()) {
+                                case Float:
+                                    return JavaConstant.forFloat(Math.max(a.asFloat(), b.asFloat()));
+                                case Double:
+                                    return JavaConstant.forDouble(Math.max(a.asDouble(), b.asDouble()));
+                                default:
+                                    throw GraalError.shouldNotReachHere();
+                            }
+                        }
+
+                        @Override
+                        public Stamp foldStamp(Stamp s1, Stamp s2) {
+                            if (s1.isEmpty()) {
+                                return s1;
+                            }
+                            if (s2.isEmpty()) {
+                                return s2;
+                            }
+                            FloatStamp stamp1 = (FloatStamp) s1;
+                            FloatStamp stamp2 = (FloatStamp) s2;
+                            Stamp folded = maybeFoldConstant(this, stamp1, stamp2);
+                            if (folded != null) {
+                                return folded;
+                            }
+
+                            return new FloatStamp(stamp1.getBits(), Math.max(stamp1.lowerBound, stamp2.lowerBound), Math.max(stamp1.upperBound, stamp2.upperBound), false);
+                        }
+                    },
+
+                    new BinaryOp.Min(true, true) {
+
+                        @Override
+                        public Constant foldConstant(Constant const1, Constant const2) {
+                            PrimitiveConstant a = (PrimitiveConstant) const1;
+                            PrimitiveConstant b = (PrimitiveConstant) const2;
+                            assert a.getJavaKind() == b.getJavaKind();
+                            switch (a.getJavaKind()) {
+                                case Float:
+                                    return JavaConstant.forFloat(Math.min(a.asFloat(), b.asFloat()));
+                                case Double:
+                                    return JavaConstant.forDouble(Math.min(a.asDouble(), b.asDouble()));
+                                default:
+                                    throw GraalError.shouldNotReachHere();
+                            }
+                        }
+
+                        @Override
+                        public Stamp foldStamp(Stamp s1, Stamp s2) {
+                            if (s1.isEmpty()) {
+                                return s1;
+                            }
+                            if (s2.isEmpty()) {
+                                return s2;
+                            }
+                            FloatStamp stamp1 = (FloatStamp) s1;
+                            FloatStamp stamp2 = (FloatStamp) s2;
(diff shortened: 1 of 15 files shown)


compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/FloatStamp.java
```diff
@@ -910,6 +910,76 @@ public Stamp foldStamp(Stamp s) {
 
                     null, null, null,
 
+                    new BinaryOp.Max(true, true) {
```
teshull, line 913: Since Max/Min are associative, I believe also hooks for them must be added into `BinaryArithmeticNode.reassociate`

SwapnilGaikwad: Done!

teshull (approved):

compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/FloatStamp.java
```diff
@@ -910,6 +910,76 @@ public Stamp foldStamp(Stamp s) {
...
+                            if (s2.isEmpty()) {
+                                return s2;
+                            }
+                            FloatStamp stamp1 = (FloatStamp) s1;
+                            FloatStamp stamp2 = (FloatStamp) s2;
+                            Stamp folded = maybeFoldConstant(this, stamp1, stamp2);
+                            if (folded != null) {
+                                return folded;
+                            }
+                            return stamp1.unrestricted();
```
teshull, line 944: Can you tighten the bounds of the returned stamp? Since this is a max operation, we know the lower bound will be max(stamp1.lowerBound(), stamp2.lowerBound()) and the upper bound likewise will be the max of the two stamps' upper bounds.

compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/FloatStamp.java
```diff
@@ -910,6 +910,76 @@ public Stamp foldStamp(Stamp s) {
...
```
teshull, line 979: Same thing as the comment above, except this time the upper/lower bounds would be the min of the two stamps' respective values

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/MaxNode.java
```diff
@@ -0,0 +1,91 @@
...
+import org.graalvm.compiler.nodeinfo.NodeInfo;
+import org.graalvm.compiler.nodes.ConstantNode;
+import org.graalvm.compiler.nodes.NodeView;
+import org.graalvm.compiler.nodes.ValueNode;
+import org.graalvm.compiler.nodes.spi.NodeLIRBuilderTool;
+
+import jdk.vm.ci.meta.Value;
+
+@NodeInfo(shortName = "MAX")
+public class MaxNode extends BinaryArithmeticNode<Max> implements NarrowableArithmeticNode, BinaryCommutative<ValueNode> {
```
teshull, line 45: Can you create a shared abstract superclass for the Max and Min nodes, called MinMaxNode, and lift the common logic into this class?

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/MaxNode.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    protected BinaryOp<Max> getOp(ArithmeticOpTable table) {
+        return table.getMax();
+    }
+
+    public static ValueNode create(ValueNode x, ValueNode y, NodeView view) {
+        BinaryOp<Max> op = ArithmeticOpTable.forStamp(x.stamp(view)).getMax();
+        Stamp stamp = op.foldStamp(x.stamp(view), y.stamp(view));
```
teshull, line 60: Can you add the assert `assert stamp instanceof FloatStamp`?

teshull (changes requested):

compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/type/FloatStamp.java
SwapnilGaikwad: Done!

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/MaxNode.java
SwapnilGaikwad: Done, added MinMaxNode.java

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/MaxNode.java
SwapnilGaikwad: Done!

PR #2590, opened 2020-06-19
Optimizing VolatileReadNode out causes wrong memory state to propagate

rwestrel:
This is bug https://bugs.openjdk.java.net/browse/JDK-8247832

VolatileReadNode.simplify() removes VolatileReadNodes that have no use other than Memory uses. The current logic updates the uses' Memory input with the VolatileReadNode's lastLocationAccess. That's broken as the LocationIdentity of the VolatileReadNode and the one of its uses likely differ. To make that optimization possible, I suggest the VolatileReadNode keeps track of the entire memory state in its lastLocationAccess input with a MemoryMapNode.

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/LateMembarInsertionTest.java
+++ b/compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/LateMembarInsertionTest.java
@@ -195,45 +195,6 @@ public class LateMembarInsertionTest extends GraalCompilerTest {
         verifyMembars("volatileFieldStore", membarsExpected());
     }
 
-    // Unused field load should be optimized out and leave no barrier behind
-    @SuppressWarnings("unused")
-    public static void volatileFieldStoreUnusedVolatileFieldLoadVolatileFieldStore(int v2) {
-        VolatileAccess2.field = v2;
-        int v1 = VolatileAccess.field;
-        VolatileAccess2.field = v2;
-    }
-
-    @Test
-    public void test09() {
-        StructuredGraph graph = getFinalGraph(getResolvedJavaMethod("volatileFieldStoreUnusedVolatileFieldLoadVolatileFieldStore"));
-        List<TypePair> accesses = getAccesses(graph);
-
-        Assert.assertEquals(accesses.size(), 2);
-        Assert.assertEquals(accesses.get(0).getType(), volatileAccess2Type);
-        Assert.assertEquals(accesses.get(1).getType(), volatileAccess2Type);
-        Assert.assertTrue(accesses.get(0).isWrite());
-        Assert.assertTrue(accesses.get(1).isWrite());
-        Assert.assertEquals(membarsExpected() ? 4 : 0, getMembars(graph).size());
-    }
-
-    // Unused field load should be optimized out and leave no barrier behind
-    @SuppressWarnings("unused")
-    public static void unusedVolatileFieldLoadVolatileFieldStore(int v2) {
-        int v1 = VolatileAccess.field;
-        VolatileAccess2.field = v2;
-    }
-
-    @Test
-    public void test10() {
-        StructuredGraph graph = getFinalGraph(getResolvedJavaMethod("unusedVolatileFieldLoadVolatileFieldStore"));
-        List<TypePair> accesses = getAccesses(graph);
-
-        Assert.assertEquals(accesses.size(), 1);
-        Assert.assertEquals(accesses.get(0).getType(), volatileAccess2Type);
-        Assert.assertTrue(accesses.get(0).isWrite());
-        Assert.assertEquals(membarsExpected() ? 2 : 0, getMembars(graph).size());
-    }
-
     public static int unsafeVolatileFieldLoad(Object o, long offset) {
         return UNSAFE.getIntVolatile(o, offset);
     }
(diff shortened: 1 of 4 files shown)


davleopo:
@rwestrel   thanks for the proposed fix, awesome find, I will do some internal testing and I have to figure out what potential implications we have if the last location access can now be a memory map node. I am not sure we actually want to fix this problem that way.
The transformation killing the read is simply wrong if it does not check the locations of the usages and the location of the read (since it reads and kills different locations) and the read's last location access.

Also, given that we now need to explicitly fix up the graph again with `MemoryAccess#canonicalizeLastLocationAccess` seems like a source for bugs in the future.

Since you went full route to implement the optimization using the memory map node I guess this is relevant for certain workloads ?

rwestrel:
> Since you went full route to implement the optimization using the memory map node I guess this is relevant for certain workloads ?

The initial motivation is aarch64 where a VolatileReadNode is not lowered to a pair of membars + a read but instead implemented with a special instruction sequence. If a useless volatile read is left in the graph then on amd64, it's compiled down to a pair of (empty) membars (the read is eliminated) but on aarch64, it's a full blown volatile read with a memory access.

Now if we remove VolatileReadNode.simply(), given a VolatileReadNode is not lowered on aarch64, after FixReadsPhase I suppose VolatileReadNode is removed because it has no use.

So no, I don't know of a relevant workload that's affected but it seems aarch64 and amd64 should optimize the same way and having a high level VolatileReadNode should make it doable.

davleopo:
Hi @rwestrel 

Thanks again for your proposal to fix this bug.
I took some time to wrap my head around this problem and your solution since its touching some fundamentals of the graal IR design.

The fact that we only have an abbreviated memory graph in graal and that nodes accessing ANY_LOCATION are basically untouched (i.e. they must never float, move, nor have a last location access)
means that we cannot optimize the volatile read which accesses a location that differs from the killed location in local transformations done in canonical/simplify methods.

Removing something that kills ANY in Graal is not a local effect since we do not model that in the memory graph, and should thus not be done in canonical/simplify methods.
Given that the VolatileReadNode returns a field location identity for getLocationIdentity but kills ANY should also mean that the last location access (LLA) of the node
must be null to correctly reflect the rules about killing any. 
So I think the actual bug is twofold:
	-> the last location access of volatile reads must be null (get location identity should return what get killed location identity returns if the kill is to ANY)
	-> the simplification can never work for LLA of ANY so after the actual fix this code becomes dead and should be removed

We need a dedicated, more global (e.g. a phase), transformations to do this coherently with the entire memory graph (or maybe before).
Your solution adding a memory map node for volatile reads in this scenario may work, however it does not follow the design of the memory graph in graal and adds a level of complexity and size (new nodes and new edges) we 
wish to avoid.
We have not gone the route yet to have a full blown memory graph, basically what your changes start introducing, and we currently do not anticipate to do so in the near future.
This may change at some point, but is a separate issue.

I think for this particular bug we should do the following:
	(1) Integrate your test case
	(2) Remove the simplification of the volatile read node and let it return ANY for its getLocationIdentity
		-> this also means adjusting the LateMembarInsertion test for the fact that the reads in two test cases are not removed
	
To optimize volatile reads we have possible future directions to go:
	(a) Add a dedicated phase that optimizes volatile reads globally
	or
	(b) Drill open some logic in an existing phase (e.g. early read elimination) and perform the removal of volatile reads
	
In the past we did not perform any optimizations on volatile reads, and we have not seen immanent performance gains in doing so on amd64, so
for now I don't think we should go for a solution with such heavy design implications, for the sake of optimizing infrequent code patterns without a visible effect on important workloads.

If you prefer to not fix the bug that way, I can do the proposed changes of (1) and (2).
And you can work on a proper, more global, elimination of volatile reads in the future.

adinn:
Hi @davleopo Thanks for your feedback. You are right that the problem here is a fundamental issue with the current model of memory accesses in the Graal compiler. That was identified several years ago when we tried to find ways to optimize volatile operations for AArch64.

I am not sure I agree with your conclusion:

> I don't think we should go for a solution with such heavy design implications, for the sake of optimizing infrequent code patterns without a visible effect on important workloads

Is that claim based on experience with x86_64 workloads? Experience with OpenJDK on AArch64 suggests that the ability to optimize volatile accesses is important for performance on the latter architecture. While AArch64 has not been used extensively in the past few years that is unlikely to remain the status quo.

davleopo:
Hi @adinn ,

> You are right that the problem here is a fundamental issue with the current model of memory accesses in the Graal compiler. That was identified several years ago when we tried to find ways to optimize volatile operations for AArch64.

I would not call that issue, its a design decision of graal to not model the full state in the memory graph.

> Is that claim based on experience with x86_64 workloads? Experience with OpenJDK on AArch64 suggests that the ability to optimize volatile accesses is important for performance on the latter architecture. While AArch64 has not been used extensively in the past few years that is unlikely to remain the status quo.

Yes, I mentioned that above that we did not see any obvious gains on x86_64. 
I agree that regarding aarch64 the situation might be different and that it will be important in the future and thus should be optimized.

I am not arguing against optimizing the loads :-)  I think its an important and valuable optimization especially on aarch64 as you outlined, 
however, extending the memory graph to that extent just for the sake of volatile loads is an overkill, because then we have 2 "models/ways" of doing
memory optimizations in graal: the global phases we have now and the "local" kind of transformations/canonicalizations/simplifications done by the volatile read node.
 
If we are to actually start propagating a full blown memory graph in graal we'd have to rewrite the machinery we currently have to do memory optimizations
to support the local kind of transformations that could actually benefit from the additional data in the memory graph.
However, that is a fundamental engineering effort with (at least on amd64) questionable benefit currently, thus adding a dedicated phase
that does the removal of volatile fields is a straightforward solution that fits nicely into the existing design of our memory graph.

(/cc @tkrodriguez )

rwestrel:
> I think for this particular bug we should do the following:
> (1) Integrate your test case
> (2) Remove the simplification of the volatile read node and let it return ANY for its getLocationIdentity
> -> this also means adjusting the LateMembarInsertion test for the fact that the reads in two test cases are not removed

That sounds ok. I've just pushed a new patch that does what you recommend.

davleopo:
@rwestrel  thanks, I started internal testing and will integrate if nothing else comes up.

tkrodriguez:
Optimizing volatile memory accesses is certainly something we should pursue if there's performance to be had.  I can certainly believe that it would help aarch64 more than it might amd64.

PR #2596, opened 2020-06-22
Fix native-image-maven-plugin for Windows

### LINKED ISSUE #2151: Native-image-maven-plugin does not work in Windows
Hello, I am having trouble while compiling springboot using maven v3.1.1, the error message is
 

`Fatal error: com.oracle.svm.core.util.VMError$HostedError: Option name "IncludeResourceBundles" has multiple definitions: com.oracle.svm.core.jdk.LocalizationSupport$Options.IncludeResourceBundles and com.oracle.svm.core.jdk.LocalizationFeature$Options.IncludeResourceBundles
	at com.oracle.svm.core.util.VMError.shouldNotReachHere(VMError.java:66)
	at com.oracle.svm.hosted.option.HostedOptionParser.collectOptions(HostedOptionParser.java:88)
	at com.oracle.svm.hosted.option.HostedOptionParser.<init>(HostedOptionParser.java:61)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:206)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:454)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:115)`

Can someone help me ?

olpaw:
fixes #2151

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/OS.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/OS.java
@@ -52,6 +52,10 @@ public enum OS {
         return name().toLowerCase();
     }
 
+    public boolean isCurrent() {
+        return getCurrent() == this;
+    }
+
     private static OS findCurrent() {
         final String name = System.getProperty("os.name");
         if (name.equals("Linux")) {
--- a/substratevm/src/native-image-maven-plugin/src/main/java/com/oracle/substratevm/NativeImageMojo.java
+++ b/substratevm/src/native-image-maven-plugin/src/main/java/com/oracle/substratevm/NativeImageMojo.java
@@ -175,7 +175,7 @@ public class NativeImageMojo extends AbstractMojo {
         addClasspath(project.getArtifact());
         String classpathStr = classpath.stream().map(Path::toString).collect(Collectors.joining(File.pathSeparator));
 
-        Path nativeImageExecutable = getMojoJavaHome().resolve("bin").resolve(withExeSuffix("native-image"));
+        Path nativeImageExecutable = getMojoJavaHome().resolve("bin").resolve("native-image" + (OS.WINDOWS.isCurrent() ? ".cmd" : ""));
         if (Files.isExecutable(nativeImageExecutable)) {
             String nativeImageExecutableVersion = "Unknown";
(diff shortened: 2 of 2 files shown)


jonenst:
Hi @pejovica @olpaw 
What is this PR supposed to fix ?

I think this pr breaks my build (links to public builds that show the problem at the end).
[INFO] Executing: C:\hostedtoolcache\windows\jdk\11.0.0\x64\bin\native-image.cmd -cp [... many many jars ] --no-fallback --allow-incomplete-classpath -H:Class=com.powsybl.tools.Main -H:Name=olf
The command line is too long.

The only change I see in the log is that native-image.exe is changed to native-image.cmd.

The command line generated by native-image-maven-plugin in my case is about 15k characters (all the jars).

I read that
""A little bit of google later, and I learn that Windows has a 32767 character command line limit. That limit is even less, 8191, if trying to execute the process from within cmd.exe.""

I'm not a windows user so I don't know much about limits, but maybe switching from .exe to .cmd reduced the max command line length ?

Thanks,

builds:
https://github.com/powsybl/powsybl-open-loadflow/runs/1414612072 (KO) (using native-image-maven-plugin 20.3)
https://github.com/powsybl/powsybl-open-loadflow/runs/1414671494 (OK) (using native-image-maven-plugin 20.1)

olpaw:
The GraalVM directory layout was changed on Windows to be in line with other platforms (Linux & Darwin). Thus 
```
getMojoJavaHome().resolve("bin").resolve("native-image" + (OS.WINDOWS.isCurrent() ? ".cmd" : ""));
```
now points to a "link" (in the case of Windows, a batch script because we cannot use symlinks on Windows) that forwards to the real location of the native-image executable. Note that this is the same for other GraalVM applications in `$GRAALVM_HOME/bin` (`ployglot.cmd`, `js.cmd`, `gu.cmd`, ...).

As you observed this has the unpleasant side effect that now because of going through a batch-file the Windows character command line size hits us even harder. Please open a ticket for this.

As a workaround you could create an uberjar out of all theses jars (e.g. with `maven-shade-plugin`) that then gets feed into `native-image-maven-plugin`. In this case the command line should be much shorter.

PR #2611, opened 2020-06-25
Ds/gr 24427

dougxc:


zakkak:
@dougxc can we somehow limit the notifications generated by (https://github.com/oracle/graal/pull/2611/commits/dc5d4ff08a624233585fde0811d79be74e2f1adb) to be sent only when the pipeline is ran for the main repository and not for forks?

dougxc:
> @dougxc can we somehow limit the notifications generated by ([dc5d4ff](https://github.com/oracle/graal/commit/dc5d4ff08a624233585fde0811d79be74e2f1adb)) to be sent only when the pipeline is ran for the main repository and not for forks?

I've added `if: branch = master AND repo = oracle/graal` which should acheive that. Note that it's commented out for now so that I can test whether notifications work. I will uncomment it before merging.

Issue #2587, opened 2020-06-19
Cannot see native call frames in GDB


YaSuenag:
I tried to see native call frames of customized GraalVM which was built with `-H:GenerateDebugInfo=1` in native debugger (GDB) on Linux, but I could not see them in GDB.

I expect I can see call frames which are implemented by Java in libpolyglot. `objdump --dwarf=info libpolyglot.so` shows DWARF information for Java layer, so `GenerateDebugInfo` seems to be applied partially.

**Steps to reproduce the issue**

1. Build GraalVM with `-H:GenerateDebugInfo=1` with following patch

```
diff --git a/sdk/mx.sdk/mx_sdk_vm_impl.py b/sdk/mx.sdk/mx_sdk_vm_impl.py
index de295302e50..68343ab58d7 100644
--- a/sdk/mx.sdk/mx_sdk_vm_impl.py
+++ b/sdk/mx.sdk/mx_sdk_vm_impl.py
@@ -1893,6 +1893,7 @@ class GraalVmSVMNativeImageBuildTask(GraalVmNativeImageBuildTask):
         build_args = [
             '--macro:' + GraalVmNativeProperties.macro_name(self.subject.native_image_config),
             '-H:NumberOfThreads=' + str(self.parallelism),
+            '-H:GenerateDebugInfo=1',
         ]
         if self.subject.native_image_config.is_polyglot:
             build_args += ["--macro:truffle", "--language:all"]
diff --git a/substratevm/mx.substratevm/mx_substratevm.py b/substratevm/mx.substratevm/mx_substratevm.py
index cc551793ed9..6899a5df20b 100644
--- a/substratevm/mx.substratevm/mx_substratevm.py
+++ b/substratevm/mx.substratevm/mx_substratevm.py
@@ -976,6 +976,7 @@ mx_sdk_vm.register_graalvm_component(mx_sdk_vm.GraalVmJreComponent(
         "-Dorg.graalvm.polyglot.nativeapi.libraryPath=${.}/../../../polyglot/",
         "-H:CStandard=C11",
         "-H:+SpawnIsolates",
+        "-H:GenerateDebugInfo=1",
     ],
     polyglot_lib_jar_dependencies=[
         "substratevm:POLYGLOT_NATIVE_API",
```

2. Create C source file to crash

libpolyglot-crash.c

```
#include <stddef.h>
#include <polyglot_api.h>

main(){
  poly_thread dummy_thread; /* Not initialized */
[...]

PR #2612, opened 2020-06-25
Generate correct debug info for deoptimization target methods

adinn:
This is a proposed fix for issue #2587. It handles the case where deopt target methods are included in a native image. Compiled code for these methods is generated after the compiled code for normal compiled methods . In consequence, a class which contains deopt targets has code occupying two distinct code cache address ranges and code form different classes may be interleaved.

This is a problem because DWARF debug data (info, ranges, frame) must be generated as a sequence of entries associated with non-overlapping address ranges ordered low to high. So, a class containing deopt target methods must be processed twice, generating DWARF entries first for the normal methods and then for the deopt targets.

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
@@ -55,7 +55,8 @@ public class Range {
     /*
      * Create a primary range.
      */
-    public Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line, boolean isDeoptTarget) {
+    public Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line,
+                    boolean isDeoptTarget) {
         this(fileName, filePath, className, methodName, paramNames, returnTypeName, stringTable, lo, hi, line, isDeoptTarget, null);
     }
 
@@ -65,10 +66,12 @@ public class Range {
     public Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line, Range primary) {
         this(fileName, filePath, className, methodName, paramNames, returnTypeName, stringTable, lo, hi, line, false, primary);
     }
+
     /*
      * Create a primary or secondary range.
      */
-    private Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line, boolean isDeoptTarget, Range primary) {
+    private Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line,
+                    boolean isDeoptTarget, Range primary) {
         /*
          * Currently file name and full method name need to go into the debug_str section other
          * strings just need to be deduplicated to save space.
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
@@ -46,6 +46,7 @@ public class Range {
     private int lo;
     private int hi;
     private int line;
+    private boolean isDeoptTarget;
     /*
      * This is null for a primary range.
      */
@@ -54,14 +55,20 @@ public class Range {
     /*
      * Create a primary range.
      */
-    public Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line) {
-        this(fileName, filePath, className, methodName, paramNames, returnTypeName, stringTable, lo, hi, line, null);
+    public Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line, boolean isDeoptTarget) {
+        this(fileName, filePath, className, methodName, paramNames, returnTypeName, stringTable, lo, hi, line, isDeoptTarget, null);
     }
 
     /*
-     * Create a primary or secondary range.
+     * Create a secondary range.
      */
     public Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line, Range primary) {
+        this(fileName, filePath, className, methodName, paramNames, returnTypeName, stringTable, lo, hi, line, false, primary);
+    }
+    /*
+     * Create a primary or secondary range.
+     */
+    private Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line, boolean isDeoptTarget, Range primary) {
         /*
          * Currently file name and full method name need to go into the debug_str section other
          * strings just need to be deduplicated to save space.
@@ -76,6 +83,7 @@ public class Range {
         this.lo = lo;
         this.hi = hi;
         this.line = line;
+        this.isDeoptTarget = isDeoptTarget;
         this.primary = primary;
     }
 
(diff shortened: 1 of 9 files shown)


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
```diff
@@ -133,6 +133,10 @@ public String getFullMethodName() {
         return fullMethodName;
     }
 
+    public boolean isDeoptTarget() {
+        return methodName.endsWith("**");
```
olpaw, line 137: Might be a good idea to define a public static field `METHOD_NAME_DEOPT_SUFFIX` for `"**"` in `src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/meta/HostedMethod.java:247` and use that field also here.

adinn: Yes, will do.

adinn:
Hi @olpaw Just pushed a fix for the rework. I also corrected a problem setting the compile unit index for the deopt target methods compile unit (this was not caught by previous testing because assertions were not enabled when I built libployglot).

This patch now allows @YaSuenag's  libpolyglot crasherto be debugged. Also an objdump of all the debug sections produces no warnings or errors and an eyeball scan shows everything generated with the right values and in the right order.

olpaw:
@adinn thanks a lot. I will push it through out gates.

YaSuenag:
Thanks @adinn ! I got the following backtrace on GDB with your patch:

```
(gdb) bt
#0  com.oracle.svm.core.jdk.VMErrorSubstitutions::doShutdown(java.lang.String, java.lang.Throwable)(void) () at com/oracle/svm/core/log/RealLog.java:385
#1  0x00007f0a2e9ad0c9 in com.oracle.svm.core.util.VMError::shouldNotReachHere(java.lang.String, java.lang.Throwable)(void) () at com/oracle/svm/core/threadlocal/FastThreadLocalInt.java:51
#2  0x00007f0a2e9ad05c in com.oracle.svm.core.util.VMError::guarantee(boolean, java.lang.String)(void) () at com/oracle/svm/core/util/VMError.java:88
#3  0x0000000000000000 in ?? ()
```

I wonder why frame # 3 could not be shown. It seems to lack of DWARF for C API in libpolyglot.so.

YaSuenag:
In addition, RealLog.java:385 does not contain VMErrorSubstitutions::doShutdown.

```
383     @Override
384     public Log autoflush(boolean onOrOff) {
385         autoflush = onOrOff;
386         return this;
387     }
```

adinn:
Hi @YaSuenag 

    (gdb) bt
    #0  com.oracle.svm.core.jdk.VMErrorSubstitutions::doShutdown(java.lang.String, java.lang.Throwable)(void) () at com/oracle/svm/core/log/RealLog.java:385
    #1  0x00007f0a2e9ad0c9 in com.oracle.svm.core.util.VMError::shouldNotReachHere(java.lang.String, java.lang.Throwable)(void) () at com/oracle/svm/core/threadlocal/FastThreadLocalInt.java:51
    #2  0x00007f0a2e9ad05c in com.oracle.svm.core.util.VMError::guarantee(boolean, java.lang.String)(void) () at com/oracle/svm/core/util/VMError.java:88
    #3  0x0000000000000000 in ?? ()

Yes, that is the same stack backtrace I saw.

> I wonder why frame # 3 could not be shown. It seems to lack of DWARF for C API in libpolyglot.so.

This looks like it may be a separate issue to do with the frame section data for frame #2. The stack unwind has not found a valid return address. I'll investigate that as a follow-up to this fix.

> In addition, RealLog.java:385 does not contain VMErrorSubstitutions::doShutdown.

    383     @Override
    384     public Log autoflush(boolean onOrOff) {
    385         autoflush = onOrOff;
    386         return this;
    387     }

I think you are reading that the wrong way round. The frame info indicates that this is a frame for compiled method VMErrorSubstitutions::doShutdown and that the current code address was derived from source code in file RealLog.java. So, that means the current code address is inlined code and you need to look for the call to autoflush in (or below) method VMErrorSubstitutions::doShutdown. And indeed at line 100 of VMErrorSubstitutions.java we have

    100    private static void doShutdown(String msg, Throwable ex) {
    101        try {
    102            Log log = Log.log();
    103            log.autoflush(true);

So, this appears to be correct.

olpaw:
> @adinn thanks a lot. I will feed it to our CI.

Gates all pass. Except for that style issue:
```
1. ERROR in /b/b/e/main/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java (at line 215)
	private int findLo(LinkedList<PrimaryEntry> classPrimaryEntries, boolean isDeoptTargetCU) {
	            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
The method findLo(LinkedList<PrimaryEntry>, boolean) from the type DwarfInfoSectionImpl can be declared as static
----------
2. ERROR in /b/b/e/main/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java (at line 233)
	private int findHi(LinkedList<PrimaryEntry> classPrimaryEntries, boolean includesDeoptTarget, boolean isDeoptTargetCU) {
	            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
The method findHi(LinkedList<PrimaryEntry>, boolean, boolean) from the type DwarfInfoSectionImpl can be declared as static
----------
2 problems (2 errors)
```
Please fix.

adinn:
<sigh/> Should be ok now :-)

olpaw:
With https://github.com/oracle/graal/pull/2612/commits/6c5420a33a06c782d709e48165bf6b89cc442981 `mx eclipseformat` fails
```
Changes:
--- a/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
+++ b/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
@@ -55,7 +55,8 @@
     /*
      * Create a primary range.
      */
-    public Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line, boolean isDeoptTarget) {
+    public Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line,
+                    boolean isDeoptTarget) {
         this(fileName, filePath, className, methodName, paramNames, returnTypeName, stringTable, lo, hi, line, isDeoptTarget, null);
     }
 
@@ -65,10 +66,12 @@
     public Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line, Range primary) {
         this(fileName, filePath, className, methodName, paramNames, returnTypeName, stringTable, lo, hi, line, false, primary);
     }
+
     /*
      * Create a primary or secondary range.
      */
-    private Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line, boolean isDeoptTarget, Range primary) {
+    private Range(String fileName, Path filePath, String className, String methodName, String paramNames, String returnTypeName, StringTable stringTable, int lo, int hi, int line,
+                    boolean isDeoptTarget, Range primary) {
         /*
          * Currently file name and full method name need to go into the debug_str section other
          * strings just need to be deduplicated to save space.
```

adinn:
Hmm, that didn't happen with my local eclipseformat. I am using 4.14version

adinn:
Hi @YaSuenag

I have identified why stack frame #3 has not been correctly unwound. It's a problem in the frame data being passed to through to the debug info generator which will need a compiler change to fix it. I'll raise a separate issue for it.

olpaw:
> Hmm, that didn't happen with my local eclipseformat. I am using 4.14. Is the build using a later one? I'll correct this by hand.

Our `gate-vm-style-linux-amd64` shows
```
> 2020-06-26 14:41:11(+00:09) [INFO] Set environment variable ECLIPSE_EXE=/sc/bb/downloads/eclipse-4.14.0-linux-amd64_8aabce73d3/eclipse/eclipse
```
in its logs.

PR #2614, opened 2020-06-25
Prints JDK build number along with  java.version

zakkak:
Since graal is sensitive to JDK changes printing the build number along
with the java version, makes issue reporting and debugging a bit easier.

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
@@ -82,7 +82,7 @@ class DefaultOptionHandler extends NativeImage.OptionHandler<NativeImage> {
                 if (!NativeImage.graalvmConfig.isEmpty()) {
                     message += " " + NativeImage.graalvmConfig;
                 }
-                message += " (Java Version " + System.getProperty("java.version") + ")";
+                message += " (Java Version " + System.getProperty("java.runtime.version") + ")";
                 nativeImage.showMessage(message);
                 System.exit(0);
                 return true;


substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
```diff
@@ -82,7 +82,7 @@ public boolean consume(Queue<String> args) {
                 if (!NativeImage.graalvmConfig.isEmpty()) {
                     message += " " + NativeImage.graalvmConfig;
                 }
-                message += " (Java Version " + System.getProperty("java.version") + ")";
```
dougxc, line 85: `java.version` and `java.vm.version` are 2 different pieces of info:
```
java -XshowSettings:properties  2>&1| grep 'java.*version'
    java.class.version = 52.0
    java.runtime.version = 1.8.0_251-b08
    java.specification.version = 1.8
    java.version = 1.8.0_251
    java.vm.specification.version = 1.8
    java.vm.version = 25.251-b08-jvmci-20.2-b01
```
People may already be depending on the current output format and semantics. This is API and evolving it should be done carefully.

zakkak: You are right. I only checked with java 11.
Changing this to a draft PR for the time being.
Thanks @dougxc

zakkak:
Would something like @jerboaa 's suggestion here https://github.com/graalvm/mandrel/issues/68#issue-645562223 be acceptable?

jerboaa:
> Would something like @jerboaa 's suggestion here [graalvm#68 (comment)](https://github.com/graalvm/mandrel/issues/68#issue-645562223) be acceptable?

I don't think this wouldn't work, since it would break the build on JDK 8 (Runtime.version() is 11+ API). Perhaps creating an overlay for the build number and using that could work, a la:
https://github.com/oracle/graal/pull/2403

zakkak:
I have edited this PR to only append the build number when not using JDK 8. However to make it work I had to add two `--add-exports` in the native-image launcher. I am not sure if this is the right way or where else I should add such exports. Any help/feedback is welcome.

Thank you.

jaikiran:
Hello @zakkak,

> Since graal is sensitive to JDK changes printing the build number along
with the java version, makes issue reporting and debugging a bit easier.

Do build numbers change for (released) Java versions? AFAIK, they don't. So `11.0.6` Java version (for example) will always have the same build number. Printing that additionally won't add additional value, isn't it?

If this is about EA releases of Java, then the version there will be printed something like `15-ea` and I can understand requiring to print build numbers in such cases, but Graal (released) versions won't use EA JDKs anyway. So I'm not sure what cases will printing the build number help.

zakkak:
Hello @jaikiran ,

> 
> Do build numbers change for (released) Java versions? AFAIK, they don't. So `11.0.6` Java version (for example) will always have the same build number. Printing that additionally won't add additional value, isn't it?
> 

I might be wrong but I don't see why the build number cannot change for a Java release.
Additionally, I see that labs-openjdk shipped as 11.0.6+9 with jvmci 20.1-b01 (https://github.com/graalvm/labs-openjdk-11/releases/tag/jvmci-20.1-b01) while adopt-openjdk shipped as 11.0.6+10 (https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/tag/jdk-11.0.6%2B10).

> If this is about EA releases of Java, then the version there will be printed something like `15-ea` and I can understand requiring to print build numbers in such cases, but Graal (released) versions won't use EA JDKs anyway. So I'm not sure what cases will printing the build number help.

When using an EA release you actually get something like `11.0.6-ea` not just `11-ea`.
Regarding cases where printing the build number helps see the Mandrel distribution of Graal (https://github.com/graalvm/mandrel) that uses upstream OpenJDK instead of the labs-OpenJDK and right now requires EA releases to build. So in the Mandrel case you might get an Alpha1 release built using `11.0.8-ea+5` and then an Alpha2 built using `11.0.8-ea+8` etc.

jaikiran:
Hello @zakkak, the formal specification of these version numbers currently resides at https://openjdk.java.net/jeps/223. Based on the specification there and some background from here https://github.com/graalvm/mandrel/issues/68, I think what you are looking for, is the `java.runtime.version` system property value, which as per the JEP will be:

```
Name                            Syntax
------------------------------  --------------
...
java.runtime.version            $VSTR
```
The definition of `$VSTR` is provided in the section titled "Version Strings" in that JEP. So if the `java.runtime.version` gets used here, then you should start seeing your expected output like `11.0.8-ea+8`

jaikiran:
> So if the java.runtime.version gets used here, then you should start seeing your expected output like 11.0.8-ea+8

What that would mean is, your patch/PR can be as simple as this:

```
diff --git a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
index ebe30e712a4..a4d14123659 100644
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
@@ -82,7 +82,7 @@ class DefaultOptionHandler extends NativeImage.OptionHandler<NativeImage> {
                 if (!NativeImage.graalvmConfig.isEmpty()) {
                     message += " " + NativeImage.graalvmConfig;
                 }
-                message += " (Java Version " + System.getProperty("java.version") + ")";
+                message += " (Java Version " + System.getProperty("java.runtime.version") + ")";
                 nativeImage.showMessage(message);
                 System.exit(0);
                 return true;
```

zakkak:
Thanks a lot @jaikiran that's exactly what we need.

I tried it with:
```
$ more Version.java  
public class Version {
       System.out.println(System.getProperty("java.runtime.version"));
   } 
}
```

and it yields

```
$ java Version
1.8.0_252-b09

$ /opt/jvms/openjdk-11.0.8_8/bin/java Version  
11.0.8-ea+8

$ /opt/jvms/labsjdk-ce-11.0.7-jvmci-20.2-b02/bin/java Version
11.0.7+10-jvmci-20.2-b02

$ /opt/jvms/jdk1.7.0_25/bin/java Version      
1.7.0_25-b15
```

@dougxc would using `java.runtime.version` be OK?

dougxc:
Yes, I think `java.runtime.version` provides the right amount of info.

zakkak:
Nice, thanks @dougxc

jerboaa:
@dougxc @olpaw Is this waiting on further input from us or would this be ready to be merged?

olpaw (approved):

olpaw:
> @dougxc @olpaw Is this waiting on further input from us or would this be ready to be merged?

Thanks for the reminder. It's running in our CI now.

jerboaa:
Thanks!

Issue #2624, opened 2020-06-29
Debug info frame size changes are not being recorded and processed correctly


adinn:
The compiler adds frame-event marks to a compilation result at 6 distinct points: prologue entry, prologue stack grow, prologue exit, epilogue entry, epilogue stack shrink and epilogue exit. These events are passed on to debug info generation code as frame change events. There is an error in the generation of events and an error in the way they are converted to  frame change events.

1) The compiler currently marks epilogue end events at the point where the frame has been torn down i.e. the same point as the epilogue stack shrink. This is incorrect because the epilogue should included all code up to and including any ensuing return (or, in special cases, up to a block end via a branch out of the method). The range of code up to that final transfer of control at epilogue exit needs to be marked as having an empty stack. Furthermore, the epilogue exit mark is needed in order to identify that subsequent code generated in the method -- code which can only be reached via a jump past the epilogue -- should be understood as having a non-empty stack i.e. the epilogue exit mark serves to re-establish the presence of a stack frame for that next segment of generated method code.

2) The debug info frame size change notifications notify the initial frame growth during the prologue and notfiy a frame shrink during each epilogue. However, they do not re-notify stack growth for generated method code that follows an epilogue exit. Clearly, doing so wihtout a remedy for the previous error is not possible since the point at which the epilogue exit is generated cannot currently be used to decide if the epilogue is actually at the end of a method.

How to reproduce:

Build libpolgloty.so with debug info enabled as per instructions in issue #2587
build and run a.out as per instructions in issue #2587
Debug the core in gdb as per instructions in issue #2587 
Observe the following stack trace
   (gdb) bt
    #0  com.oracle.svm.core.jdk.VMErrorSubstitutions::doShutdown(java.lang.String, java.lang.Throwable)(void) () at com/oracle/svm/core/log/RealLog.java:385
    #1  0x00007f78baf03879 in com.oracle.svm.core.util.VMError::shouldNotReachHere(java.lang.String, java.lang.Throwable)(void) () at com/oracle/svm/core/threadlocal/FastThreadLocalInt.java:51
    #2  0x00007f78baf0380c in com.oracle.svm.core.util.VMError::guarantee(boolean, java.lang.String)(void) () at com/oracle/svm/core/util/VMError.java:88
    #3  0x0000000000000000 in ?? ()
    (gdb) p/x $rsp

Use objdump to dump info and frame sections so that code addresses and frame sizes can be identified

[...]

PR #2625, opened 2020-06-29
Debugframesize

adinn:
This PR fixes issue #2624. The changes in the compiler code ensure that a Frame.returned() callback occurs at the end of any basic block that also includes a call to Frame.leave(). This is needed to ensure that marks inserted by the Substrate implementors of Frame are correctly located. Substrate Frame.leave implementations still add marks for EPILOGUE_START and EPILOGUE_INCD_RSP. Generation of EPILOGUE_END occurs in Substrate Frame.returned implementations i.e. at the end of the basic block terminated by the return (or, in certain cases, a method exiting jump to a handler). The current code places the EPILOGUE_END mark before the terminating return (or jump). In deed, in some cases it also precedes safepoint code or other code that operates with the empty frame.

The result of making this compiler change is to allow the debug info frame record generator to decrement the stack frame size at EPILOGUE_INCD_RSP but then restore it again at EPILOGUE_END if it finds there is code that follows the EPILOGUE_END.

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot.aarch64/src/org/graalvm/compiler/hotspot/aarch64/AArch64HotSpotBackend.java
+++ b/compiler/src/org.graalvm.compiler.hotspot.aarch64/src/org/graalvm/compiler/hotspot/aarch64/AArch64HotSpotBackend.java
@@ -253,6 +253,11 @@ public class AArch64HotSpotBackend extends HotSpotHostBackend implements LIRGene
         }
 
         @Override
+        public void returned(CompilationResultBuilder crb) {
+            // nothing to do
+        }
+
+        @Override
         public boolean hasFrame() {
             return true;
         }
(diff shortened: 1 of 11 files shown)


adinn:
@olpaw @tkrodriguez Any chance of a review for this small patch?

tkrodriguez:
It looks ok to me.  I'll run it through our internal gates today.

adinn:
Thanks Tom!

olpaw:
@adinn thanks for the fix. Unfortunately this breaks debug_frame generation of EE debuginfo. I haved to fix that first & then I can merge your PR. Sorry for the inconvenience.

adinn:
@olpaw No problem. Thanks for letting me know. I hope the compiler changes offer a better/simpler approach for the EE code to generate frame info.

olpaw:
Merged on master: https://github.com/oracle/graal/commit/7b174f1d5fc56cd3dfd8677c00ae54f1ac02c9d1

BTW, the test that failed and made me aware that I need to adjust EE debuginfo is a good one. You might want to add something similar on CE. It builds a simple image (that has methods with multiple exit points) and then runs that image with GDB in "assembly single stepping mode" (si) until it reaches the breakpoint for `exit@plt / exit`. After each `si` it dumps the stack with `bt` and searches for `in ??` in the stacktrace. If it finds such a string the test fails (because it found a code location where getting a stack trace does not work).

adinn:
> BTW, the test that failed and made me aware that I need to adjust EE debuginfo is a good one. You might want to add something similar on CE. It builds a simple image (that has methods with multiple exit points) and then runs that image with GDB in "assembly single stepping mode" (si) until it reaches the breakpoint for exit@plt / exit. After each si it dumps the stack with bt and searches for in ?? in the stacktrace. If it finds such a string the test fails (because it found a code location where getting a stack trace does not work).

Hmm, that's a very good idea! I'll develop a test along those lines.

PR #2631, opened 2020-06-30
Fix the documentation in PolyglotNativeAPI

jyukutyo:
It seems that the documentation value of the `poly_value_remove_array_element` method is copy of the `poly_value_set_array_element` method. That is a remove method but its comment is "Sets the value".

I found the relationship between the PolyglotNativeAPI class and org.graalvm.polyglot.Value class, so I got the correct comment of the `poly_value_remove_array_element` method from the `removeArrayElement` of the Value class.


vjovanov:
Thank you, very much!

PR #2638, opened 2020-07-01
Make sure SSL logging (configured using javax.net.debug) works in native application

jaikiran:
Consider the following very trivial Java program:
```
import java.net.URL;
import java.io.InputStream;
import java.net.*;
import java.io.*;

public class HelloWorld {
		System.out.println("Hello World"); 
		System.getProperties().list(System.out);
     	final URL targetURL = new URL("https://google.com/");
        HttpURLConnection responseConnection = (HttpURLConnection) targetURL.openConnection();
        System.out.println("Connected to target");
        try (final InputStream is = responseConnection.getInputStream()) {
        	is.read();
        }
        System.out.println("Done");

	} 
}
```
which just connects using `HTTPS` to some website and reads the stream.

Let's assume that we generate a native image out of this application

The https://github.com/oracle/graal/blob/master/substratevm/PROPERTIES.md documentation states:

> Passing -D<key>=<value> to an image execution affects properties seen at image run-time

Running the generated image with:
```
./helloworld -Djavax.net.debug=all
```
doesn't generate any SSL logging at all. It starts generating these logs only if at the  native-image generation _build time_ was passed a value for `-Djavax.net.debug`. 

This is because the SSL logging infrastructure in the JDK classes (`sun.security.ssl.Debug` in Java 8 and `sun.security.ssl.SSLLogger` in Java 11 and above) use class static blocks to decide and store the decision of whether or not to generate log messages. It's almost impossible to try and prevent these classes from being initialized at build time (I tried), so as a result, their state ends up getting set during image build time and as a result the `-Djavax.net.debug` system property has no effect on the logging.

The commit in this PR uses the existing `RuntimeClassInitializationSupport` infrastructure to reinitialize these classes at runtime (if accessible) so that the logging decision is re-evaluated at runtime. With this commit, I am now able to see the SSL log messages if `-Djavax.net.debug` system property is set, when launching the generated image.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
@@ -144,6 +144,19 @@ public class SecurityServicesFeature extends JNIRegistrationUtil implements Feat
         ImageSingletons.lookup(RuntimeClassInitializationSupport.class).rerunInitialization(clazz(access, "sun.security.ssl.SSLContextImpl$DefaultManagersHolder"),
                         "for reading properties at run time");
 
+        /*
+         * SSL debug logging enabled by javax.net.debug system property is setup during the class
+         * initialization of sun.security.ssl.Debug (in Java 8) or sun.security.ssl.SSLLogger (in
+         * Java 11). We cannot avoid these classes from being dragged in during image build time, so
+         * we have to reinitialize these classes at runtime to allow honouring runtime passed
+         * configuration for the javax.net.debug system property.
+         */
+        if (JavaVersionUtil.JAVA_SPEC == 8) {
+            ImageSingletons.lookup(RuntimeClassInitializationSupport.class).rerunInitialization(clazz(access, "sun.security.ssl.Debug"), "for substitutions");
+        } else if (JavaVersionUtil.JAVA_SPEC >= 11) {
+            ImageSingletons.lookup(RuntimeClassInitializationSupport.class).rerunInitialization(clazz(access, "sun.security.ssl.SSLLogger"), "for substitutions");
+        }
+
         if (SubstrateOptions.EnableAllSecurityServices.getValue()) {
             /* Prepare SunEC native library access. */
             prepareSunEC();


jaikiran:
Hello @cstancu, any inputs/reviews on this one?

jaikiran:
Hello @dougxc, is there anything I can do to help move this forward? I was hoping this would make it to the 20.2 release scheduled for August 18 2020.

dougxc:
@peter-hofer could you please help @cstancu with this.

cstancu:
@jaikiran thanks for the contribution and sorry for the late reply. I'll merge this in and make sure it will be included in the 20.2 release.

cstancu:
`sun.security.ssl.Debug` has been replaced by `sun.security.ssl.SSLLogger` in https://hg.openjdk.java.net/jdk/jdk11/rev/68fa3d4026ea (see https://bugs.openjdk.java.net/browse/JDK-8196584). This has also been backported to JDK8 `8u261` by https://bugs.openjdk.java.net/browse/JDK-8234225. I'll modify this PR accordingly. This also means that trying to build with a JDK8 prior to `8u261` will fail since `sun.security.ssl.SSLLogger` would be missing. cc: @dougxc

jaikiran:
Hello @cstancu,

> @jaikiran thanks for the contribution and sorry for the late reply. I'll merge this in and make sure it will be included in the 20.2 release.

Thank you very much for the review and help.

> sun.security.ssl.Debug has been replaced by sun.security.ssl.SSLLogger in https://hg.openjdk.java.net/jdk/jdk11/rev/68fa3d4026ea (see https://bugs.openjdk.java.net/browse/JDK-8196584). 

Correct. That's why I included the Java version check.

> This has also been backported to JDK8 8u261 by https://bugs.openjdk.java.net/browse/JDK-8234225.

Didn't know that. So thank you for pointing me to it.

> I'll modify this PR accordingly. This also means that trying to build with a JDK8 prior to 8u261 will fail since sun.security.ssl.SSLLogger would be missing

You are right. Should we perhaps do this logic in the `beforeAnalysis` of the `SecurityServicesFeature` instead of during it in the `duringSetup`? That way we could perhaps use `BeforeAnalysis#registerReachabilityHandler` for these classes to make sure this logic gets applied only if the class is reachable? But I'm just guessing that non-availability of such classes won't cause errors when `registerReachabilityHandler` is used. Is that right? If yes, then I think we can even get rid of the Java version check and blindly add the reachability handlers, which enroll them for rerunning initialization at runtime, for these 2 classes?

cstancu:
It gets more complicated than that. The change has only been backported to OracleJDK (8u261) but not to OpenJDK(8u262), so the proper solution is to condtitonally load the classes, regardless of platform:
```
optionalClazz(access, "sun.security.ssl.Debug").ifPresent(c -> rci.rerunInitialization(c, "for reading properties at run time"));
optionalClazz(access, "sun.security.ssl.SSLLogger").ifPresent(c -> rci.rerunInitialization(c, "for reading properties at run time"));
```
The analysis reachability handler wouldn't work for this purpose. The reachability handler assumes that the element (class, method, field) is present and checks if it is reachable from an analysis point-of-view. In this case the class is missing alltogether.

jaikiran:
Thank you for all these inputs and explanation @cstancu. It helps.

I won't be updating this PR since you noted that you will be doing the changes. Do let me know if you need me to do it though.

Issue #2639, opened 2020-07-01
Text relocations cause crash on Android.


lazar-mitrovic:
Since https://github.com/oracle/graal/commit/fdd84a00ef4f9ef5d537a92ee4783f678ffac674 compiling for Android with `-H:+UseOnlyWritableBootImageHeap` flag set, produces following runtime error:
```
java.lang.UnsatisfiedLinkError: dlopen failed: "libsubstrate.so" has text relocations
```
This is due to Android API level > 23 enforcing that shared libraries [cannot use text relocations](https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md#text-relocations-enforced-for-api-level-23).

Fix requires that relocations are moved to RW ELF segment.

PR #2640, opened 2020-07-01
Remove text relocations if UseOnlyWritableBootImageHeap flag is set

lazar-mitrovic:
Fixes https://github.com/oracle/graal/issues/2639


johanvos:
I can confirm this PR fixes the issue.

lazar-mitrovic:
Using proposed previous fix would move text section to writable segment, so as a workaround I added a linker flag to separate it. 
Note that this flag requires `ld.gold` linker.

PR #2641, opened 2020-07-02
Split tests in arm64 to avoid timeouts in Travis CI

zakkak:


.travis.yml
```diff
@@ -55,7 +55,10 @@ matrix:
   - env: JDK="jdk11" GATE="build,bootstraplite" PRIMARY="compiler"
   - os: linux
     arch: arm64
-    env: JDK="jdk11" GATE="fullbuild,test" PRIMARY="compiler"
+    env: JDK="jdk11" GATE="fullbuild,test:1:100" PRIMARY="compiler"
```
zakkak, line 58: This should be
```suggestion
    env: JDK="jdk11" GATE="fullbuild,test:1" PRIMARY="compiler"
```
but depends on https://github.com/graalvm/mx/pull/222

dougxc:
@zakkak this PR is currently being merged in it's [original form](https://github.com/oracle/graal/commit/99079f8aba4586a55041abb0a79a67470bdd6ed5). If that merge fails for some transient reason, we can adopt your rebased changes. In practice, I don't think it matters as we will hit other scalability problems should we ever have more than 100 separate "style" gate tasks.

boris-spas:
This was merged in it's original form as cedbc92ab75efe5bfec4c3fd72879f6b1d673f34.

PR #2646, opened 2020-07-02
Keep image zone rules cache separate #2637

### LINKED ISSUE #2637: ZoneRules.lastRuleCache side effects during source cache build fail native image generation
This is a bug in native image debug symbol generation. In particular, it's an issue with the way source cache generation is handled.

When `ImageClassLoader` walks the jars to find the classes, it inadvertently fills up the `ZoneRules.lastRulesCache` with time zone offset transitions for the years for which the classes were last modified. Typically, this would be filled up with years 2020 and 2019. See a [typical stacktrace](https://gist.github.com/galderz/27558be54bc0e3c6004187f32ac80c10) of this.

After the heap is creating, debug symbols generates the source cache, and at that port it can lookup classes in `*-sources.jar` files. This lookup can end up touching the same `ZoneRules.lastRuleCache` (see [stacktrace](https://gist.github.com/galderz/d282e44307e2aef0ca61760aebc143f1)), and if by chance, it adds a year that was not present after looking up classes, e.g. 2014, it will pollute the heap after snapshotting the heap. Native image detects this and fails with an error like:

```
[sourcecache-client/target/sourcecache:6632]        image:   6,911.30 ms,  3.25 GB
Fatal error:com.oracle.svm.core.util.VMError$HostedError: com.oracle.svm.core.util.UserError$UserException: Static field or an object referenced from a static field changed during native image generation?
  object:2017=[Ljava.time.zone.ZoneOffsetTransition;@5944088  of class: java.util.concurrent.ConcurrentHashMap$Node
  reachable through:
    object: [Ljava.util.concurrent.ConcurrentHashMap$Node;@ed99e14  of class: java.util.concurrent.ConcurrentHashMap$Node[]
    object: {2016=[Ljava.time.zone.ZoneOffsetTransition;@4194c6d5, 2017=[Ljava.time.zone.ZoneOffsetTransition;@5944088, 2019=[Ljava.time.zone.ZoneOffsetTransition;@118349c1, 2020=[Ljava.time.zone.ZoneOffsetTransition;@18a9bfe8, 2014=[Ljava.time.zone.ZoneOffsetTransition;@3258aabf, 2015=[Ljava.time.zone.ZoneOffsetTransition;@c4123b3}  of class: java.util.concurrent.ConcurrentHashMap
    object: ZoneRules[currentStandardOffset=Z]  of class: java.time.zone.ZoneRules
    object: Europe/London  of class: java.time.ZoneRegion
    object: sun.util.calendar.ZoneInfo[id="Europe/London",offset=0,dstSavings=3600000,useDaylight=true,transitions=242,lastRule=java.util.SimpleTimeZone[id=Europe/London,offset=0,dstSavings=3600000,useDaylight=true,startYear=0,startMode=2,startMonth=2,startDay=-1,startDayOfWeek=1,startTime=3600000,startTimeMode=2,endMode=2,endMonth=9,endDay=-1,endDayOfWeek=1,endTime=3600000,endTimeMode=2]]  of class: sun.util.calendar.ZoneInfo
    object: 0000-12-30T16:47:04.192Z  of class: sun.util.calendar.ImmutableGregorianDate
    object: [CommonEra (C.E.) since 0000-12-30T16:47:04.192 local time]  of class: sun.util.calendar.Era
    object: [Lsun.util.calendar.Era;@3bb95304  of class: sun.util.calendar.Era[]
    root: sun.util.calendar.JulianCalendar$Date.setNormalizedYear(int)

	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:647)
	at com.oracle.svm.hosted.NativeImageGenerator.lambda$run$0(NativeImageGenerator.java:451)
```

I've built a reproducer for this issue [here](https://github.com/galderz/jawa/tree/master/substratevm/sourcecache).

Some observations:
* `Files.isRegularFile()` cannot be substituted by `File.isDirectory` because the Path is a Path within a Zip FileSystem (e.g. a jar), and method is unsupported for that type of FS.
* This is a niche scenario, but the JDK tries to read all attributes (including time based ones), even when you only want to check if a file is a directory or not. Discussed with @adinn and this is fine.
* I've tried to do substitutions of `zipfs.ZipUtils` but obviously those don't have an effect during the heap build time.

I'm unsure yet about a fix here:
* One possibility would be to eagerly fill the ZoneUtils.lastRuleCache for the last N years, before snapshotting the heap. There's only one TZ in play here, which is `ZoneId.systemDefault()`.
* Another option would be to use some reflection magic to clear `lastRulesCache` after navigating the classes, and clear it again after navigating the source cache (before writing the heap).

galderz:
Closes #2637 

* Before this patch, the image would contain zone rules cache from the JVM computing the image.
* Side effects here, such as class or source cache lookups, were affecting its contents and leading to error when writing image.
* With this change, we guarantee that the zone rules cache in the image is a brand new instance, separate from the one in the JVM computing the image.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_time_zone_ZoneRules.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_time_zone_ZoneRules.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2020, 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.core.jdk;
+
+import com.oracle.svm.core.annotate.Alias;
+import com.oracle.svm.core.annotate.RecomputeFieldValue;
+import com.oracle.svm.core.annotate.TargetClass;
+
+import java.time.zone.ZoneOffsetTransition;
+import java.time.zone.ZoneRules;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+@TargetClass(ZoneRules.class)
+final class Target_java_time_zone_ZoneRules {
+
+    @SuppressWarnings("unused") //
+    @Alias @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.NewInstance, declClass = ConcurrentHashMap.class) //
+    private transient ConcurrentMap<Integer, ZoneOffsetTransition[]> lastRulesCache;
+
+}


galderz:
Can anyone review this? @zakkak @adinn? We need the fix to be able to add third party libraries with older sources to the source cache.

olpaw:
> With this change, we guarantee that the zone rules cache in the image is a brand new instance, separate from the one in the JVM computing the image.

Sounds like the right approach.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeSubstitutions.java
```diff
@@ -0,0 +1,23 @@
+package com.oracle.svm.core.jdk;
```
olpaw, line 1: > LGTM but there are some styling issues and it's missing the copyright header.

@galderz please take care of the styling issues. Ensure that `mx eclipseformat` and `mx checkstyle` pass without issues.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/TimeSubstitutions.java
```diff
@@ -0,0 +1,23 @@
...
+import com.oracle.svm.core.annotate.RecomputeFieldValue;
+import com.oracle.svm.core.annotate.TargetClass;
+
+import java.time.zone.ZoneOffsetTransition;
+import java.time.zone.ZoneRules;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+@TargetClass(ZoneRules.class)
+final class Target_java_time_zone_ZoneRules {
```
christianwimmer, line 13: For all new code we are switching to the model where each `Target_...` class is in its own .java file. In your case this avoids the awkward `class TimeSubstitutions`.

galderz:
> As you point in #2637. I wonder if we should address reading all file properties in the JDK as well, that way eventually we can remove this substitution, thoughts?

Discussed this with @adinn, and he didn't think it was such a good idea. I can't find the details of the chat right now, but IIRC he said that more often than not you're reading all the file properties.

galderz:
@olpaw I've fixed the style issues.

@christianwimmer Moved the substitution to its own file.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_time_zone_ZoneRules.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.time.zone.ZoneOffsetTransition;
+import java.time.zone.ZoneRules;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+@TargetClass(ZoneRules.class)
+final class Target_java_time_zone_ZoneRules {
+
+    @SuppressWarnings("unused") @Alias @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.NewInstance, declClass = ConcurrentHashMap.class) private transient ConcurrentMap<Integer, ZoneOffsetTransition[]> lastRulesCache;
```
zakkak, line 39: I think the following is a bit more readable :)

```suggestion
    @SuppressWarnings("unused") //
    @Alias @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.NewInstance, declClass = ConcurrentHashMap.class) //
    private transient ConcurrentMap<Integer, ZoneOffsetTransition[]> lastRulesCache;
```

adinn:
@eginez I don't think it is worth pursuing a change in the OpenJDK behaviour. The Files API abstracts over various different underlying file systems, Windows, Unix, Linux, Zip, Jar and the range of cases can be (and has been) extended by application code. There is a baked in assumption that checking for a directory requires reading the corresponding file's attributes and testing the result for the relevant attribute. Changing this assumption for some or all implementations would not be easy to get agreement for and any change might well risk breaking existing clients and/or extensions. Also, it would very likely only be changed for newer JDK releases and not get backported to jdk11.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_time_zone_ZoneRules.java
galderz: @zakkak Blame it on `mx eclipseformat`...

zakkak: I guessed so. Note though that `eclipseformat` and `checkstyle` should be happy (didn't check it but I see other classes following this style) with the above change.

olpaw: @galderz you can make `mx eclipseformat` happy with the following trick
```
@SuppressWarnings("unused") //
@Alias //
@RecomputeFieldValue(kind = RecomputeFieldValue.Kind.NewInstance, declClass = ConcurrentHashMap.class) //
private transient ConcurrentMap<Integer, ZoneOffsetTransition[]> lastRulesCache;
```

galderz: Fixed and pushed.

olpaw:
@galderz running in our CI now.

PR #2649, opened 2020-07-03
Add intrinsics for get/put unsafe unaligned methods on AArch64

SwapnilGaikwad:
Intrinsics replace the bytewise read/write operations by necessary word sized chunks while performing unaligned memory accesses using the unsafe API. Covers char, short, int and long types.

**Microbenchmarking results:**
Relevant snippet of the microbenchmark is:
```
@Benchmark
    public void jmhTimeUnsafeGetPutInt() {
        int res = 0;
        int pos;
            pos = (i * intSize);
            res += unsafe.getIntUnaligned(intArr, arrBaseOffset + pos + 1);
            unsafe.putIntUnaligned(intArr, arrBaseOffset + pos + 1, res);
        }
    res_int = res;
}
```

Performance results on an A72 AArch64 machine are:
Before:
```
Benchmark                 Mode  Cnt     Score   Error  Units
jmhTimeUnsafeGetPutChar   avgt   10   367.850 ± 0.058  ns/op
jmhTimeUnsafeGetPutShort  avgt   10   321.127 ± 1.524  ns/op
jmhTimeUnsafeGetPutInt    avgt   10   533.823 ± 0.075  ns/op
jmhTimeUnsafeGetPutLong   avgt   10  1718.885 ± 0.193  ns/op
```

After:
```
Benchmark                 Mode  Cnt     Score   Error  Units
jmhTimeUnsafeGetPutChar   avgt   10  225.210 ± 0.043  ns/op
jmhTimeUnsafeGetPutShort  avgt   10  225.435 ± 0.045  ns/op
jmhTimeUnsafeGetPutInt    avgt   10  206.207 ± 0.137  ns/op
jmhTimeUnsafeGetPutLong   avgt   10  205.953 ± 0.135  ns/op
```

### DIFF
--- a/compiler/src/org.graalvm.compiler.replacements.aarch64/src/org/graalvm/compiler/replacements/aarch64/AArch64GraphBuilderPlugins.java
+++ b/compiler/src/org.graalvm.compiler.replacements.aarch64/src/org/graalvm/compiler/replacements/aarch64/AArch64GraphBuilderPlugins.java
@@ -295,9 +295,18 @@ public class AArch64GraphBuilderPlugins implements TargetGraphBuilderPlugins {
         registerUnsafePlugins(new Registration(plugins, Unsafe.class), explicitUnsafeNullChecks,
                         new JavaKind[]{JavaKind.Int, JavaKind.Long, JavaKind.Object}, "Object");
         if (JavaVersionUtil.JAVA_SPEC > 8) {
-            registerUnsafePlugins(new Registration(plugins, "jdk.internal.misc.Unsafe", replacements), explicitUnsafeNullChecks,
-                            new JavaKind[]{JavaKind.Int, JavaKind.Long, JavaKind.Object},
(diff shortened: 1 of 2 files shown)


teshull:
Hi @SwapnilGaikwad. Thank you for the PR. Just so you know, I have a couple other things on my queue ahead of this PR, but I should be able to look at this in two weeks.

teshull:
I merged this yesterday, but didn't see your new commit before merging. Also, I changed the other initializations within `AArch64GraphBuilderPlugins.registerUnsafePlugins` to use UnsafeAcessPlugin.

Looking at your new commit, I agree that the unaligned plugins should be registered in a different method. Can you rebase that commit on the current master, and also rename it to `registerUnsafeUnalignedPlugins` or something like that (I don't care for the registerInternalUnsafeSpecificPlugins name)?

PR #2651, opened 2020-07-03
Reads around unsafe volatile read should not be commoned

rwestrel:
The following sequence:

v1 = Read of field
v2 = Unsafe volatile read
v3 = Read of field
// use of v1 & v3

is erroneously transformed to:
v1 = Read of field
v2 = Unsafe volatile read
// use of v1 only

by PEReadEliminationClosure
that is first and last reads of field are commoned.

### DIFF
--- a/compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/VolatileAccessReadEliminationTest.java
+++ b/compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/VolatileAccessReadEliminationTest.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, Red Hat Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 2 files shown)


compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ReadAroundUnsafeVolatileReadShouldNotCommon.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    public void test2() {
+        final StructuredGraph graph = getFinalGraph("testMethod2");
+        Assert.assertEquals(2, getFieldReads(graph));
+    }
+
+    public static int testMethod3(int offset) {
+        int v = field;
+
```
davleopo, line 107: Any particular reason for those newlines? If not pls remove them

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ReadAroundUnsafeVolatileReadShouldNotCommon.java
```diff
@@ -0,0 +1,125 @@
...
+import org.graalvm.compiler.core.common.type.IntegerStamp;
+import org.graalvm.compiler.core.common.type.Stamp;
+import org.graalvm.compiler.nodes.NodeView;
+import org.graalvm.compiler.nodes.StructuredGraph;
+import org.graalvm.compiler.nodes.memory.ReadNode;
+import org.junit.Assert;
+import org.junit.Test;
+
+// See https://bugs.openjdk.java.net/browse/JDK-8248598
+public class ReadAroundUnsafeVolatileReadShouldNotCommon extends GraalCompilerTest {
```
davleopo, line 39: Could you rename that to have the word "test" in its title somehow?

rwestrel:
@davleopo Thanks for reviewing this. I just pushed an updated patch that includes your comments.

davleopo:
@rwestrel  thanks, I will let your first version finish internal testing and come back to you if any problems arise.

PR #2661, opened 2020-07-08
Add compilation directory to debug info

zakkak:
This PR adds the DW_AT_comp_dir attribute to debug info, making it
point to the appropriate source file cache. This removes the need of
using `set directores sources/jdk:sources/graal:sources/src` in gdb.

At the moment there are 3 directories that might contain sources:
1. `sources/jdk`
2. `sources/graal`
3. `sources/src`

This patch detects the corresponding path for each compilation unit and adds it to the `DW_AT_comp_dir` attribute.
For instance for `java.base/java/lang/Shutdown.java` it sets `DW_AT_comp_dir` to `sources/jdk`, and for `com/oracle/svm/core/RegisterDumper.java` it sets it to `sources/graal`.

### DIFF
--- a/substratevm/DEBUGINFO.md
+++ b/substratevm/DEBUGINFO.md
@@ -174,31 +174,38 @@ configure extra source search paths (see below).
 Configuring source paths in gdb
 -------------------------------
 
-In order for gdb to be able to locate the source files for your app
-classes, Graal classes and JDK runtime classes you need to provide gdb
-with a list of source root dirs using the 'set directories' command:
+By default gdb will employ the three local directory roots
+`sources/{jdk,graal,src}` to locate the source files for your app classes, Graal
+classes and JDK runtime classes. If the sources cache is not located in the
+directory in which you run gdb you can configure the required paths using the
+following command:
 
     (gdb) set directories /path/to/sources/jdk:/path/to/sources/graal:/path/to/sources/src
 
-Directory .../sources/jdk should contain source files for all JDK runtime
+Directory `/path/to/sources/jdk` should contain source files for all JDK runtime
 classes referenced from debug records.
 
-Directory .../sources/graal should contain source files for all GraalVM
+Directory `/path/to/sources/graal` should contain source files for all GraalVM
 classes referenced from debug records. Note that the current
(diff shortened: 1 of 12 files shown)


substratevm/DEBUGINFO.md
```diff
@@ -174,37 +174,41 @@ configure extra source search paths (see below).
 Configuring source paths in gdb
 -------------------------------
 
-In order for gdb to be able to locate the source files for your app
-classes, Graal classes and JDK runtime classes you need to provide gdb
-with a list of source root dirs using the 'set directories' command:
+By default gdb will look into `/path/to/sources/{jdk,graal,src}` directories to
```
adinn, line 177: I think this would be clearer:

    By default will look employ the three local directory roots `sources/{jdk,graal,src}` to locate the source files for your app classes, Graal classes and JDK runtime classes. If the sources cache is not located in the directory in which you run gdb you can configure the required paths using the following command:

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
```diff
@@ -36,6 +36,7 @@
  */
 
 public class Range {
+    private final String compilationDirectory;
```
adinn, line 39: This name seems wrong as the value is not actually anything to do with compilation. I understand that it is used to populate the comp_dir DWARF property but that use is incidental to its origin -- which I think it is more helpful to signal. So I'd prefer to use field name cachePath and modifying the getter accordingly.

substratevm/DEBUGINFO.md
zakkak: This should now be resolved

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
zakkak: I agree. This should now be fixed all the way to `com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl#writeCU`.

PR #2664, opened 2020-07-10
Add more re-associations for associative binary operations

XiaohongGong:
It has supported the re-association to associate the values which satisfy the specified criterion like `"(a + 1) + 2 => a + (1 + 2)"`.
However, it needs one of the original operands satisfy the criterion. It cannot handle patterns like `"(a + 1) + (b + 2)"`, which could be optimized to `"(a + b) + (1 + 2)"`.

This patch adds another kind of re-association to expose the matched values as the operand. For example with the constantness criterion, `"(a + 1) + b"` is re-associated to `"(a + b) + 1"`. Used together with the existed re-association, it could create more opportunities for other optimizations like constant-fold and loop-invariant elimination.

Change-Id: If32b49a14dd08ac1ab903878da0f459f452da221


davleopo:
Hi @XiaohongGong , thanks for your contribution.

I have some general questions: do you have any particular benchmarks/workloads in mind that may profit from the additional re-association here?
And with that, did you measure the compile time impact of the transformations?

Could you elaborate a bit more on the use cases you had in mind, looking at the failing test cases it seems the re-association should
not always be done as it interferes with Graal's counted loop detection with is essential for many optimizations.

There are significant tests failing, the CountedLoopTests.
Your additional re-association seems to trigger for loop induction variables effectively destroying the code patterns we except for induction variable detection with is necessary for loops to be detected as counted.

XiaohongGong:
> Could you elaborate a bit more on the use cases you had in mind, looking at the failing test cases it seems the re-association should
> not always be done as it interferes with Graal's counted loop detection with is essential for many optimizations.
> 
> There are significant tests failing, the CountedLoopTests.
> Your additional re-association seems to trigger for loop induction variables effectively destroying the code patterns we except for induction variable detection with is necessary for loops to be detected as counted.

Hi @davleopo  Thanks for looking at this PR.  The main reason for me to create this re-association is to make more opportunities  for  the loop "ReassociateInvariantPhase" which tries to hoist the loop-invariant expressions. 

And I have noticed the failing case of `"CountedLoopTest"` as you mentioned above, and the possible influence to the counted loop optimization.  I have already created a fixing for it, that is just skipping the re-association for patterns like `i = i + var` and `i = i - var`.  The new patch will be updated soon. Thanks again!

XiaohongGong:
> I have some general questions: do you have any particular benchmarks/workloads in mind that may profit from the additional re-association here?
> And with that, did you measure the compile time impact of the transformations?

I will have a testing for the performance and compilation impact. Thanks again!

XiaohongGong:
Hi @davleopo , here is the compilation and performance test results.

**Compilation**
I got the compilation time by respectively running bootstrap with `"-XX:+BootstrapJVMCI"` and the `CTW` with `mx gate -o -t 'CTW:hosted'`.  And the final value is the average of the first 2000 compilation results.

```
                   without this patch       with this patch      delta
Bootstrap              423.505ms               429.539ms         -0.014 
CTW                    108.001ms               105.412ms         +0.025
```
So I think the compilation impact is tiny enough to be ignore.

**Performance**
Here is a jmh benchmark:
```
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Benchmark)
@Fork(value = 3)
@Measurement(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
@Warmup(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
public class LoopInvariant {

    @Param("1024")
    public int length;

    private int[] ia;
    private long[] la;

    private Random random = new Random(0);
    private int startIndex = 4;
    private int field1 = 4;
    private int field2 = 5;
    private long fieldl1 = 4L;
    private long fieldl2 = 5L;

    /**
     * Initialize. New array objects and set initial values.
     */
    @Setup(Level.Trial)
    public void init() {
        ia = new int[length];
        la = new long[length];
        for (int i = 0; i < length; i++) {
            ia[i] = random.nextInt();
            la[i] = random.nextLong();
        }
    }

    @Benchmark
    public void loopInvariantInt() {
        for (int i = startIndex; i < length; i++) {
            int m = i * field1;
            int n = (i - 1) * field2;
            ia[i] = m * n;
        }
    }

    @Benchmark
    public void loopInvariantLong() {
        for (int i = startIndex; i < length; i++) {
            long m = i * fieldl1;
            long n = (i - 1) * fieldl2;
            la[i] = m * n;
        }
    }
}
```

And the results on x86_64:
```
before 
  Benchmark                        (length)  Mode  Cnt     Score     Error  Units
  LoopInvariant.loopInvariantInt       1024  avgt   15  1503.548 ± 114.386  ns/op
  LoopInvariant.loopInvariantLong      1024  avgt   15  1479.330 ±  17.618  ns/op
after  
  Benchmark                        (length)  Mode  Cnt     Score     Error  Units
  LoopInvariant.loopInvariantInt       1024  avgt   15  1127.063 ± 102.551  ns/op
  LoopInvariant.loopInvariantLong      1024  avgt   15  1291.498 ±  44.286  ns/op
```

Results on AArch64:
```
before
  Benchmark                        (length)  Mode  Cnt     Score   Error  Units
  LoopInvariant.loopInvariantInt       1024  avgt   15  1568.473 ± 0.235  ns/op
  LoopInvariant.loopInvariantLong      1024  avgt   15  1536.475 ± 0.548  ns/op

after   
  Benchmark                        (length)  Mode  Cnt     Score   Error  Units
  LoopInvariant.loopInvariantInt       1024  avgt   15  1182.658 ± 2.619  ns/op
  LoopInvariant.loopInvariantLong      1024  avgt   15  1171.550 ± 0.282  ns/op
```

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ReassociateAndCanonicalTest2.java
```diff
@@ -0,0 +1,362 @@
...
+ */
+package org.graalvm.compiler.core.test;
+
+import org.graalvm.compiler.nodes.StructuredGraph;
+import org.graalvm.compiler.nodes.StructuredGraph.AllowAssumptions;
+import org.junit.Test;
+
+public class ReassociateAndCanonicalTest2 extends GraalCompilerTest {
+
+    private static int rnd1 = (int) (Math.random() * 100);
```
davleopo, line 34: please use a constant seed for tests (e.g. a prime number)

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ReassociateAndCanonicalTest2.java
```diff
@@ -0,0 +1,362 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.graalvm.compiler.core.test;
+
+import org.graalvm.compiler.nodes.StructuredGraph;
+import org.graalvm.compiler.nodes.StructuredGraph.AllowAssumptions;
+import org.junit.Test;
+
+public class ReassociateAndCanonicalTest2 extends GraalCompilerTest {
```
davleopo, line 32: Could you also please provide test cases for loops? some we are sure we do not want to regress any more in the future

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/ReassociateInvariantPhase.java
```diff
@@ -31,25 +31,36 @@
...
+        int iterations = 0;
+        try (DebugContext.Scope s = debug.scope(scope)) {
             boolean changed = true;
             while (changed) {
-                changed = false;
                 final LoopsData dataReassociate = new LoopsData(graph);
-                for (LoopEx loop : dataReassociate.loops()) {
-                    changed |= loop.reassociateInvariants();
-                }
+                changed = dataReassociate.loops().stream().anyMatch(predicate);
```
davleopo, line 63: I would prefer a regular loop here if possible

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/BinaryArithmeticNode.java
```diff
@@ -235,6 +235,165 @@ private static ReassociateMatch findReassociate(BinaryNode binary, NodePredicate
...
+    public static ValueNode reassociateUnmatchedValues(BinaryArithmeticNode<?> node, NodePredicate criterion, NodeView view, boolean skipLoopPhiValues) {
+        ValueNode forX = node.getX();
+        ValueNode forY = node.getY();
+        assert node.getOp(forX, forY).isAssociative();
+
+        // Skip re-associations for expressions like "i = i + var". This may influence the counted
+        // loop detection. For example, if the loop counter is re-associated from "i += (var + 1)"
+        // to "i = (i + var) + 1", the counted loop will not be detected as expected.
+        if (skipLoopPhiValues) {
+            LoopPhiPredicate phiCriterion = new LoopPhiPredicate(node);
```
davleopo, line 302: is the phi predicate class & allocation necessary here? I might miss something but it seems it does not escape and the logic of apply is very simple.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/BinaryArithmeticNode.java
```diff
@@ -235,6 +235,165 @@ private static ReassociateMatch findReassociate(BinaryNode binary, NodePredicate
...
+            if (phiCriterion.apply(forX) || phiCriterion.apply(forY)) {
```
davleopo, line 303: also I am not sure this is restrictive enough. Have you tried with derived induction variables as well and what happens for re-associations of mul nodes?

XiaohongGong: Regarding to the counted loop detection, I think re-associations for `mul ` nodes might have no influence to it, since the loop trip node should be a base induction variable, isn't it?
And except for the counted loop, do you mean it might have any other influence for re-associations like `a = 5 * i * 6  -> a = 30 * i`?

XiaohongGong:
> While re-association inside loops can gain performance, as you outlined, I am not sure about the involved compile time impact
> doing it in regular canonicalizations, that is why we limited it so far.
> Doing it multi level in canonical methods unconditionally seems a potential compile time problem.
> Can you run with -Dgraal.Time (e.g. a bootstrap, ctw and some dacapo benchmarks) so we see what the phase local time there is both to canonicalization and re-association phase.

Hi @davleopo sure! I'v got the compile time by running the bootstrap, ctw and dacapo benchmarks. 
Here is the tests command the results:

Bootstrap: 
```
mx vm -server -XX:+UnlockExperimentalVMOptions -XX:+EnableJVMCI  -Dgraal.CompilationWatchDogStartDelay=60.0D -esa -Dgraal.DetailedAsserts=true -Dgraal.VerifyGraalGraphs=true -Dgraal.VerifyGraalGraphEdges=true -Dgraal.VerifyGraalPhasesSize=true -Dgraal.VerifyPhases=true -Dgraal.CompilationFailureAction=ExitVM -Dgraal.Time=FrontEnd:4 -XX:+UseJVMCICompiler -XX:-TieredCompilation -XX:+BootstrapJVMCI -version
```

```
FrontEnd_Accm                   606756.1 ms
CanonicalizerPhase_Accm         100658.9 ms
ReassociateInvariantPhase_Accm  5737.3 ms
```              
CTW:
```
mx gate --extra-vm-argument=-Dgraal.Time=FrontEnd:4 -o -t 'CTW:hosted'
```
```
FrontEnd_Accm                   1648485.7 ms
CanonicalizerPhase_Accm         312858.3 ms
ReassociateInvariantPhase_Accm  17276.4 ms

```

Dacapo:
```
mx benchmark dacapo:* -- -Dgraal.Time=FrontEnd:4
```
```
                                lusearch     jython     h2      pmd     avrora   luindex   fop     xalan    sunflow  unit

FrontEnd_Accm                   1648485.7   298992.0  48478.4  66268.3  20574.2  24677.6  47170.0  55162.3  23199.4  ms
CanonicalizerPhase_Accm         312858.3    38963.8   8566.1   11983.6  3289.0   4099.0   8450.9   9830.8   4019.6   ms
ReassociateInvariantPhase_Accm  17276.4     3122.9    506.3    1098.0   193.9    421.5    441.9    415.3    279.8    ms
```

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/BinaryArithmeticNode.java
XiaohongGong: Good catch! We could use a static function to check instead of `NodePredicate` here. Thanks!

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ReassociateAndCanonicalTest2.java
XiaohongGong: OK, thanks for the advice! I will update it soon!

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/ReassociateInvariantPhase.java
XiaohongGong: Yeah, a regular loop better. I will update it later. Thanks for the advice!

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/BinaryArithmeticNode.java
XiaohongGong: The main concern to check the derived induction variables is that we need to get the `LoopsData` and detect the counted loop each time when canonicalizaion which I think influences the compile time more.

davleopo: no, I think we can come back there if necessary later. I was thinking of a case where we have a none re-associated basic induction variable and then a derived one and if the derived one can be destroyed if it is subject to re-association, but maybe that cannot happen, I just wanted to think it through...

davleopo:
Hi @XiaohongGong 

Thanks for the compile time numbers, but can you also capture them for master branch without your changes and compare it as for the micro like master bootstrap canonicalization time vs your branch bootstrap canonicalization time and so on?

I put your PR through some internal benchmarking and several regressions showed up.
It seems after a successful re-association of operations inside loops your canonicalizer changes can undo them again (I have not investigated what the final problem is but it seems the re-association phase optimizes them correctly but one phase later during canonicalization they are destroyed again).
Please consider the following artificial micro I created and check the graph after re-association and then the follow-up canonicalization
and compare to master branch. The 3 additions with the m1 constant inside the loop are properly re-associated to have the param + -1 float out of the loop, while with your branch they seem somewhat interleaved. You can simplify the benchmark by removing param1 & param2 additions to have just one operation subject to the re-association out of the loop but for a clearer perf impact picture I made 3 of them. 

```
@BenchmarkMode(Mode.Throughput)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@State(Scope.Benchmark)
@Fork(value = 1)
@Measurement(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
@Warmup(iterations = 5, time = 1000, timeUnit = TimeUnit.MILLISECONDS)
public class ReassociationBenchmark {
    @State(Scope.Benchmark)
    public static class ThreadState {
        int iterations = 10000;
        int param = 12;
        int param1 = 13;
        int param2 = 14;
        int sideEffect;
    }
    @Benchmark
    @Fork(jvmArgsAppend = {"-Dgraal.PartialUnroll=false", "-Dgraal.LoopPeeling=false", "-XX:JVMCIThreads=1", "-XX:-UseOnStackReplacement"})
    public int reassociate(ThreadState state) {
        int result = 0;
        int param = state.param;
        int param1 = state.param1;
        int param2 = state.param2;
        for (int i = 0; i < state.iterations; i++) {
            result += i + param + -1;
            result += i + param1 + -1;
            result += i + param2 + -1;
            state.sideEffect += result;
        }
        return result;
    }
}
```


Results master (mode is throughput so ops/ms, higher is better)
```
Benchmark                            Mode  Cnt   Score   Error   Units
ReassociationBenchmark.reassociate  thrpt    5  91.635 ± 5.087  ops/ms
```

Results your branch
```
Benchmark                            Mode  Cnt   Score   Error   Units
ReassociationBenchmark.reassociate  thrpt    5  70.540 ± 1.558  ops/ms
```

Also with your changes the re-association loop in the phase needs 7 iterations instead of 3 which seems like a how overhead.

I think we should proceed by understanding the reverse impact of the follow up canonicalization. I still think doing your proposed changes only in the re-associate invariants phase is a better idea than doing them unconditionally in the canonicalizer. But once you fixed the problem here we can do another round of benchmarks and continue discussions from there :-)

XiaohongGong:
Hi @davleopo Thanks for your micro benchmark and the regressions your pointed. Regarding to the potential influence to loop optimization and the increased compile time, I'm also considering the benfit to apply this re-association unconditionally in the canonicalization.  And I think it may be not so valuable for canonicalization. So do you think if it's better to only apply this to the `ReassociateInvariantPhase` ?

XiaohongGong:
> Thanks for the compile time numbers, but can you also capture them for master branch without your changes and compare it as for the micro like master bootstrap canonicalization time vs your branch bootstrap canonicalization time and so on?

I'v updated the compile time here: https://github.com/oracle/graal/pull/2664#issuecomment-661634680

davleopo:
Hi @XiaohongGong 

thanks for the extensive compile-time numbers. They look interesting and a bit un-expected. I expected the canonicalizer phase impact to be higher but it seems neglectable in most cases.
However, the ReassociateInvariantsPhase time impact is enormous with up to 3x than without the patch, while this mostly is not that visible in the overall phase times, we are very concerned about compile time impacts on a phase granularity if its that high.

Before we consider further steps I think we should understand why the canonicalizer phase destroys the re-associations made by your patch later again. Did you have time to analyze the regression case above?

I think we want a solution that captures the case you presented as a micro, i.e., inside loops being optimized in the ReassociateInvariants phase but without the compile time regressions. I think the patterns in the canonicalizer are mostly not important (also because the beneficial loop cases will be handled by the ReassociateInvariantsPhase). 
But I think we should really first care about getting rid of the regression cases and then continue from there :-)

XiaohongGong:
> Did you have time to analyze the regression case above?

Sure! Applying `"Canonicalizer"` after `ReassociateInvariantPhase` might cause this regression. 
Here is the reason to make this happen:

The origin pattern inside loop:
```
 result += i + param + -1;
 result += i + param1 + -1;
 result += i + param2 + -1;
```

After `ReassociateInvariantPhase`, it is optimized to:
```
result += i + i + i + ((param + param1 + param2) + (-3))
```
So ` ((param + param1 + param2) + (-3))` can be moved outside of the loop as expected.

However, if `Canonicalizer` is applied after then, the pattern is re-associated back to:
```
result += (i + i + i + (param + param1 + param2)) + (-3)
```
Since it uses the `isConstantPredicate` while not `InvariantPredicate`  for `CanonicalizationPhase`, and this will separate out the constant value which is `-3` in this case.

One fix for this regression is to prevent the `re-association` after `ReassociateInvariantPhase`. However, I don't think this is a good idea. So any better idea to address it? Thanks!

davleopo:
> One fix for this regression is to prevent the re-association after ReassociateInvariantPhase. However, I don't think this is a good idea. So any better idea to address it? Thanks!

Will this re-association in the canoncalizer after also apply for the original cases? I.e. previously to your canonical() changes to be done for all associative nodes ?

XiaohongGong:
> > One fix for this regression is to prevent the re-association after ReassociateInvariantPhase. However, I don't think this is a good idea. So any better idea to address it? Thanks!
> 
> Will this re-association in the canoncalizer after also apply for the original cases? I.e. previously to your canonical() changes to be done for all associative nodes ?

`"canoncalizer"` is unconditionally applied to all cases I think. If the original cases have been re-associated to the expected patterns, this re-association will do nothing to them.  However, for the case above, `i + i + i + ((param + param1 + param2) + (-3))` is the new created node in `ReassociateInvariantPhase`, which can be re-associated in the subsequent `"canoncalizer"`. 

BTW, I'm thinking about another way to fix it, that is to limit the `re-association` in "`canoncalizer`" working inside a basic block like the match rule. So that the regression for "`ReassociateInvariantPhase`" will not exist.  But I still need more investigation for it. And we should also evaluate the whole benefits.  And to make it simple, I still think giving up this re-association in "canonicalizer" is the optimal choice.

davleopo:
> BTW, I'm thinking about another way to fix it, that is to limit the re-association in "canoncalizer" working inside a basic block like the match rule. So that the regression for "ReassociateInvariantPhase" will not exist. But I still need more investigation for it. And we should also evaluate the whole benefits. And to make it simple, I still think giving up this re-association in "canonicalizer" is the optimal choice.

Sounds good, I think a solution that captures the pattern from your original micro done in the ReassociateInvariantPhase should give the best benefits without most of the negative compile time implications.

XiaohongGong:
> > BTW, I'm thinking about another way to fix it, that is to limit the re-association in "canoncalizer" working inside a basic block like the match rule. So that the regression for "ReassociateInvariantPhase" will not exist. But I still need more investigation for it. And we should also evaluate the whole benefits. And to make it simple, I still think giving up this re-association in "canonicalizer" is the optimal choice.
> 
> Sounds good, I think a solution that captures the pattern from your original micro done in the ReassociateInvariantPhase should give the best benefits without most of the negative compile time implications.

Hi @davleopo , considering there are other regressions of this re-association. For example, considering the original codes inside a loop (inv1, inv2 are loop invariants):
```
 x = (var1 + inv1) + inv2
 y = x + var2
```
They are expected to be re-associated to:
```
 x = var1 + (inv1 + inv2)
 y = x + var
```
which has two additions inside the loop. 

However, if applying the re-association I added, it will be transformed to: 
```
 x = var1 + (inv1 + inv2)
 y = (var1 + var2) + (inv1 + inv2) 
```
It has three additions inside the loop. And different from the above regression above, it's not caused by the `canonical`.  So I'm considering a more sensible way to fix it. Since this new added re-association (extract the matched operand) is used as an assistant for the existed re-association that combines the matched operands, I think it's better to add some conditions for running it. For example, we can check whether the expressions are expected to be used for the existed re-association. 

Here is the codes that I want to add:
```
private static boolean worthReassociation(BinaryArithmeticNode<?> node, NodePredicate criterion) {
        // If the node is an expression like "c = (a + 1) + b", and it is single used to "c + 2" or
        // "c + (d + 2)", it is worth to be re-associated to "c = (a + b) + 1". Then its usage can
        // be further optimized to "(a + b) + 3" or "(a + b + d) + 3".
        if (node.hasExactlyOneUsage()) {
            Node usage = node.usages().first();
            if (usage instanceof BinaryArithmeticNode) {
                BinaryArithmeticNode<?> binary = (BinaryArithmeticNode<?>) usage;
                if (isReassociative(binary, node)) {
                    ValueNode otherNode = binary.getX() == node ? binary.getY() : binary.getX();
                    if (criterion.apply(otherNode) || isReassociative(binary, otherNode, criterion)) {
                        return true;
                    }
                }
            }
        }

        // If the node is an expression like "(a + 1) + (b + 2)", it's also worth exposing the
        // matched operand. Then it can be further re-associated to "(a + b) + 3".
        return isReassociative(node, node.getX(), criterion) && isReassociative(node, node.getY(), criterion);
 }
```
This could fix the regressions that we met. Currently the new patch is our internal testing. If it is ok for you, I will update the patch as soon as possible. Thanks!

For the increasing compile time to `ReassociateInvariantPhase`,  I'm trying to reduce the time by fine tuning the codes. Thanks!

davleopo:
Hi @XiaohongGong ,

Thanks for the explanations.

> It has three additions inside the loop. And different from the above regression above, it's not caused by the canonical

Good catch!

> Since this new added re-association (extract the matched operand) is used as an assistant for the existed re-association that combines the matched operands, I think it's better to add some conditions for running it

Interesting idea! 

> This could fix the regressions that we met. Currently the new patch is our internal testing. If it is ok for you, I will update the patch as soon as possible. Thanks!

I am looking forward to the updated version :-) I will put it through internal testing and benchmarking then as well so we get more data points for the performance impacts.

XiaohongGong:
Hi @davleopo , I'v updated the patch to the new one. Here is the main changes compared to before:
1. Rename the original "ReassociateInvariantPhase" to "ReassociationPhase".  It not only contains the loop invariant re-associations, but also the constant re-associations (only the constant push down part "(a * 2) * b" => "(a * b) * 2").   To avoid regressions to loop invariant, the constant part is applied after loop invariant re-association, and only applied to expressions outside of the loop.
2. For "ReassociationPhase", to avoid collecting the "loopDatas" each time the nodes in loops are changed (it may increase the compile time), the new added loop expressions are saved and appended to the original loop nodes list.
3. Canonicalization for "multiply constant" ("m * 2"  => "m << 1") is deferred after the "ReassociationPhase", to make sure expressions like "(m * 4) * (n * 2)" could be reassociated to "(m * n) * 8", and then to "(m * n) << 3" finally.

I'v tested our internal benchmarks and didn't find the significant performance regressions.

The new compile time compared with master is:
```
Bootstrap:
                                              Before        After       unit
FrontEnd_Accm                                 490614.7     493758.8      ms
PhaseTime_CanonicalizerPhase_Accm             36398.6      33296.6       ms
PhaseTime_CanonicalizerPhase_Instance_Accm    47496.3      45102.7       ms
PhaseTime_ReassociateInvariantPhase_Accm      2751.7       5411.4        ms

CTW:
                                              Before        After       unit
FrontEnd_Accm                                 1480048.7    1478955.4     ms
PhaseTime_CanonicalizerPhase_Accm             112048.9     110754.5      ms
PhaseTime_CanonicalizerPhase_Instance_Accm    183582.1     182383.9      ms
PhaseTime_ReassociateInvariantPhase_Accm       9641.1      18727.1       ms

Dacapo:

Without this patch:
                                  lusearch     jython     h2      pmd     avrora   luindex   fop     xalan    sunflow  unit

FrontEnd_Accm                     38414.6     344545.8  51026.5  94873.1  19802.0  29773.0  62950.5  68017.5  42132.7   ms
CanonicalizerPhase_Accm           3125.5      24713.7   4630.2   8525.2   1657.3   2661.0   5714.3   5931.6   3975.1    ms
CanonicalizerPhase_Instance_Accm  3848.3      29215.3   5379.9   9438.6   1944.8   2763.8   6663.2   7584.1   4698.5    ms
ReassociateInvariantPhase_Accm    171.3       2502.4    295.3    662.9    75.6     304.9    410.4    333.1    161.2     ms

With this patch:
                                  lusearch     jython     h2      pmd     avrora   luindex   fop     xalan    sunflow  unit

FrontEnd_Accm                     34402.9    293843.7   46611.3  67609.8   18022.4  23882.6  42556.1  45907.5  27409.5  ms
CanonicalizerPhase_Accm           2722.3     18357.7    4347.9   6339.2    1504.2   2078.8   3853.1   3957.8   2472.2   ms
CanonicalizerPhase_Instance_Accm  3590.4     22290.1    4753.0   6645.4    1773.8   2330.4   4478.3   4999.7   2889.7   ms
ReassociationPhase_Accm           312.6      3367.5     421.6    799.5     151.7    391.5    404.8    391.7    215.3    ms
```

The time for the single phase `ReassociationPhase` increased, however  the whole compile time can be reduced in a way. So I think the compile time is ok. 

Regarding to your feedback! Thanks!

davleopo:
Hi @XiaohongGong , thanks for the updated PR, I will put it through our benchmarking and testing and will work through the changes in detail and come back with a review!

XiaohongGong:
> Hi @XiaohongGong , thanks for the updated PR, I will put it through our benchmarking and testing and will work through the changes in detail and come back with a review!

Thanks so much for your update!

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ReassociationPhase.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Re-associate loop invariant so that invariant parts of the expression can move outside of the
+     * loop.
+     */
+    @SuppressWarnings("try")
+    private static void reassociateInvariant(StructuredGraph graph, DebugContext debug) {
+        LoopsData loopsData = new LoopsData(graph);
+        int iterations = 0;
+        try (DebugContext.Scope s = debug.scope("ReassociateInvariants")) {
```
davleopo, line 65: we had problems with while(true) loops in the past and thus impose a policy where every while true loop must have an upper bound of iterations that should be a sanity filter.
While it is true that this loop should terminate, in favor of defensive programming please add a constant upper number of iterations (idk 32? or what was the highest number of iterations y have seen during testing?) and abort the loop if the upper bound is reached.

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ReassociationPhase.java
```diff
@@ -0,0 +1,121 @@
...
+            boolean changed = true;
+            while (changed) {
+                changed = false;
+                for (LoopEx loop : loopsData.loops()) {
+                    changed |= loop.reassociateInvariants();
+                }
+                loopsData.deleteUnusedNodes();
+                iterations++;
+                debug.dump(DebugContext.VERBOSE_LEVEL, graph, "after iteration %d", iterations);
```
davleopo, line 74: pls make that a VERY_DETAILED level and add the string "reassociation:" to the message

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ReassociationPhase.java
```diff
@@ -0,0 +1,121 @@
...
+            reassociateConstant(graph, debug);
+        } catch (Throwable e) {
+            throw debug.handle(e);
+        }
+        graph.setAfterReassociation();
+    }
+
+    /**
+     * Re-associate loop invariant so that invariant parts of the expression can move outside of the
+     * loop.
```
davleopo, line 59: maybe you could also add a simple java doc example here pls?

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ReassociationPhase.java
```diff
@@ -0,0 +1,121 @@
...
+ * Rearrange {@link BinaryArithmeticNode#isAssociative() associative binary operations} for loop
+ * invariants and constants.
+ */
+public class ReassociationPhase extends Phase {
+
+    @SuppressWarnings("try")
+    @Override
+    protected void run(StructuredGraph graph) {
+        DebugContext debug = graph.getDebug();
+        try (DebugContext.Scope s = debug.scope("ReassociationPhase")) {
```
davleopo, line 48: every phase has its own scope why is this one necessary?

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ReassociationPhase.java
```diff
@@ -0,0 +1,121 @@
...
+import org.graalvm.compiler.nodes.NodeView;
+import org.graalvm.compiler.nodes.StructuredGraph;
+import org.graalvm.compiler.nodes.ValueNode;
+import org.graalvm.compiler.nodes.calc.BinaryArithmeticNode;
+import org.graalvm.compiler.nodes.util.GraphUtil;
+import org.graalvm.compiler.phases.Phase;
+
+/**
+ * Rearrange {@link BinaryArithmeticNode#isAssociative() associative binary operations} for loop
+ * invariants and constants.
```
davleopo, line 40: please add a simple code sample here that shows such an operation

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/MulNode.java
```diff
@@ -92,27 +93,37 @@ public ValueNode canonical(CanonicalizerTool tool, ValueNode forX, ValueNode for
...
             }
 
-            if (c instanceof PrimitiveConstant && ((PrimitiveConstant) c).getJavaKind().isNumericInteger()) {
-                long i = ((PrimitiveConstant) c).asLong();
-                ValueNode result = canonical(stamp, forX, i, view);
-                if (result != null) {
-                    return result;
+            if (op.isAssociative()) {
+                // Canonicalize expressions like "(a * 2) * 4" => "(a * 8)"
+                ValueNode reassociated = reassociate1(self != null ? self : (MulNode) new MulNode(forX, forY).maybeCommuteInputs(), ValueNode.isConstantPredicate(), forX, forY, view);
```
davleopo, line 108: We try to always use factory methods of nodes so any immediate folding can occur during construction already

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/BinaryArithmeticNode.java
```diff
@@ -261,7 +393,7 @@ private static ReassociateMatch findReassociate(BinaryNode binary, NodePredicate
      * @param forY
      * @param forX
      */
-    public static ValueNode reassociate(BinaryArithmeticNode<?> node, NodePredicate criterion, ValueNode forX, ValueNode forY, NodeView view) {
+    public static ValueNode reassociate1(BinaryArithmeticNode<?> node, NodePredicate criterion, ValueNode forX, ValueNode forY, NodeView view) {
```
davleopo, line 396: please give this method and the reassociate2 a better descriptive name that makes clear what is done in each method

compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/GraalOptions.java
```diff
@@ -94,7 +94,7 @@
     public static final OptionKey<Boolean> LoopPeeling = new OptionKey<>(true);
 
     @Option(help = "", type = OptionType.Debug)
-    public static final OptionKey<Boolean> ReassociateInvariants = new OptionKey<>(true);
+    public static final OptionKey<Boolean> Reassociation = new OptionKey<>(true);
```
davleopo, line 97: I think this name alone is not descriptive enough, maybe ReassociateLoopInvariants? Also, please add a small sentence in the help text above in the annotation

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/StructuredGraph.java
```diff
@@ -978,6 +980,14 @@ public void setAfterExpandLogic() {
         isAfterExpandLogic = true;
     }
 
+    public boolean isAfterReassociation() {
+        return isAfterReassociation;
+    }
+
+    public void setAfterReassociation() {
```
davleopo, line 987: In the course of my suggestion for the mul node this can be removed in favor of the shift to mul transformation

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/MulNode.java
```diff
@@ -92,27 +93,37 @@ public ValueNode canonical(CanonicalizerTool tool, ValueNode forX, ValueNode for
...
+                if (reassociated != self) {
+                    return reassociated;
                 }
             }
 
-            if (op.isAssociative()) {
-                // canonicalize expressions like "(a * 1) * 2"
-                return reassociate(self != null ? self : (MulNode) new MulNode(forX, forY).maybeCommuteInputs(), ValueNode.isConstantPredicate(), forX, forY, view);
+            // Defer the lowering to make sure the constant values have been re-associated
```
davleopo, line 114: See my general review comments about the shift to mul rewrite.

davleopo (commented):
Hi @XiaohongGong  thanks for the updated PR I think we are in a good shape I added some minor comments. 
I have however, one remaining concern: your solution to defer mul to shift opt after reassociation. While this works this is not a very good solution for other parts of our, internal, code base. There are other code parts that expect mul nodes to be optimized early on. I played around with a different solution that is equally good I think and works nicely with all parts of the code base. 

To make the respective re-associations for mul nodes working I would suggest to rewrite shift nodes with constants back to mul nodes for the sake of re-association and a subsequent canonicalization can clean them up again. This allows us to get rid of the isAfterReassociation flag and simplifies code in the mul node and the PR altogether.

If you like I can take the PR from here (after you fixed the minor comments), or you can implement my suggestion, as you wish :-) .

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ReassociationPhase.java
XiaohongGong: Since it has the "ReassociateInvariants" debug scope, is it really needed to add the "reassociation" to the message?

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ReassociationPhase.java
XiaohongGong: Since the code sample will be added for `"ReassociateInvatiant"` and `"reassociateConstant"` respectively, does it still need to add the code sample here? Thanks!

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/MulNode.java
XiaohongGong: Yes, it's better to use the `create` function normally. However, we cannot use the factory method here, since it is inside the `canonical` function which will also be called by "create". Using `create` here will create a dead loop, which looks like `create` -> `canonical` -> `create` -> `canonical` -> ...

XiaohongGong: Yes, it's better to use the `create` normally. However, it cannot be used here, since this is inside of the `canonical` which is called by `create`. Using `create` here will make it a dead loop which looks like `create` -> `canonical` -> `create` -> `canonical` -> ...

And this is the log detail:
```
        at jdk.internal.vm.compiler/org.graalvm.compiler.nodes.calc.MulNode.create(MulNode.java:70)
        at jdk.internal.vm.compiler/org.graalvm.compiler.nodes.calc.MulNode.canonical(MulNode.java:108)
        at jdk.internal.vm.compiler/org.graalvm.compiler.nodes.calc.MulNode.create(MulNode.java:70)
        at jdk.internal.vm.compiler/org.graalvm.compiler.nodes.calc.MulNode.canonical(MulNode.java:108)
        at jdk.internal.vm.compiler/org.graalvm.compiler.nodes.calc.MulNode.create(MulNode.java:70)
        at jdk.internal.vm.compiler/org.graalvm.compiler.nodes.calc.MulNode.canonical(MulNode.java:108)
        at jdk.internal.vm.compiler/org.graalvm.compiler.nodes.calc.MulNode.create(MulNode.java:70)
```

compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/GraalOptions.java
XiaohongGong: `Reassociation` is not only used for loop invariants, but also the constant reassociation. So do you have any better idea for it? Thanks!

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ReassociationPhase.java
davleopo: No, right, adding it in the other places should be sufficient.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/MulNode.java
davleopo: ah, yes overlooked that ;-) never mind

compiler/src/org.graalvm.compiler.phases.common/src/org/graalvm/compiler/phases/common/ReassociationPhase.java
XiaohongGong: Got it! Thanks again!

compiler/src/org.graalvm.compiler.core.common/src/org/graalvm/compiler/core/common/GraalOptions.java
davleopo: > not only used for loop invariants, but also the constant reassociation
I know, but the important use case is loops, thus the suggestion. But you are right. I think the name of the phase is fine, though. Maybe sth like ReassociateExpressions?

XiaohongGong: Yeah, make sense! I will change the name in the new patch. Thanks!

davleopo:
Hi,

> Yes, I agree with you that some other parts might expect mul nodes to be optimized early on.

The main problem there is that this may transitively enable other transformations be done, thus prohibiting the transformation early on is not ideal.

> Rewriting shift with constants back to mul can only handle the pow-of-two constant multiplication. Multiplying with a non-pow-of-two constant like a * 3 will be lowered to "a << 1 + a". So we need to not only consider the shift but also the "add"/"sub" if applying the reverse transformation. I think this might make things more complex. Do you have a better idea for it?

Thats true. Initially I thought loosing some of those constants is not crucial, but  if we decide we need to cover all mul cases I believe the rewrite is not too complex. its one level deeper after the muls, should be do-able.
But I can take that over, as mentioned above, and polish the details there. Please just make the suggested changes and I will fix this issue internally :-) .

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ReassociationTest.java
```diff
@@ -0,0 +1,474 @@
...
+    public static long testOverflow4Snippet() {
+        return rndL1 * (Long.MIN_VALUE * rndL2) * (-3L);
+    }
+
+    public static long refOverflow4Snippet() {
+        return (rndL1 * rndL2) * (Long.MIN_VALUE * (-3L));
+    }
+
+    private void testReassociateConstant(String testMethod, String refMethod) {
+        test(testMethod);
```
davleopo, line 350: that's something I overlooked. While this is a nice way of testing the pure functionality of the phase I think we also want to test the application of the phase embedded in the regular compilation pipeline.
For the test cases above can you please add assertions for the most important node counts, e.g. for the mul test ensure after a regular compilation the final low tier graph before backend contains one mul node with constant 15 and so on.
We want to especially test the mul stuff we are discussing about and for that we need to test the final graph result after all phases.

XiaohongGong:
> Hi,
> 
> > Yes, I agree with you that some other parts might expect mul nodes to be optimized early on.
> 
> The main problem there is that this may transitively enable other transformations be done, thus prohibiting the transformation early on is not ideal.
> 
> > Rewriting shift with constants back to mul can only handle the pow-of-two constant multiplication. Multiplying with a non-pow-of-two constant like a * 3 will be lowered to "a << 1 + a". So we need to not only consider the shift but also the "add"/"sub" if applying the reverse transformation. I think this might make things more complex. Do you have a better idea for it?
> 
> Thats true. Initially I thought loosing some of those constants is not crucial, but if we decide we need to cover all mul cases I believe the rewrite is not too complex. its one level deeper after the muls, should be do-able.
> But I can take that over, as mentioned above, and polish the details there. Please just make the suggested changes and I will fix this issue internally :-) .

Great, thanks so much about this! And I will also take attention for it. Thanks again!

XiaohongGong:
Hi @davleopo ,  the new version is updated. Could you please take a look at it again? Thanks so much!

compiler/src/org.graalvm.compiler.core.test/src/org/graalvm/compiler/core/test/ReassociationTest.java
davleopo: I think the final graph y added should be sufficient, thx. Verifying the node counts by hand would just guarantee that if we start doing something categorically wrong both inside the test and the ref snippet we would still catch it but thats not necessary for now.

davleopo:
Hi @XiaohongGong ,

thanks for the fast update :-) . I will take over and do the remaining changes and then go through internal benchmark and testing again. I will come back to you if anything major blocking comes up, else I will do remaining changes/fixes.

XiaohongGong:
Hi @davleopo , thanks so much! Waiting for your testing and fixing! Thanks again!

davleopo:
@XiaohongGong  I finished the remaining fixes internally and started integration, should appear here soon.

XiaohongGong:
> @XiaohongGong I finished the remaining fixes internally and started integration, should appear here soon.

Hi @davleopo , thanks so much for all your help！

PR #2668, opened 2020-07-11
Fix SimpleLanguage readMember implementation

lucaswerkmeister:
As far as I can tell, `readMember` implementations must assume that they may be called with invalid keys (i. e., the caller is not responsible for calling `isMemberReadable` first), and should themselves throw an `UnknownIdentifierException` if necessary rather than return `null`.

The specific pattern used here – return the value if it is not `null`, unconditional `throw` at the end – is based on the `readMember` implementations of [`WasmModule`](https://github.com/oracle/graal/blob/5cdf18908cf604f2d69a3bb523dd4123602cd30e/wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/WasmModule.java#L109) and [`PythonAbstractObject`](https://github.com/graalvm/graalpython/blob/f9e72166e24221cf23d51547923ce620eb57882b/graalpython/com.oracle.graal.python/src/com/oracle/graal/python/builtins/objects/PythonAbstractObject.java#L233) (though both of them have more than one potential return before the `throw`).

---

I’m not sure how to test this – improvements to the pull request are welcome.

### DIFF
--- a/truffle/src/com.oracle.truffle.sl/src/com/oracle/truffle/sl/runtime/FunctionsObject.java
+++ b/truffle/src/com.oracle.truffle.sl/src/com/oracle/truffle/sl/runtime/FunctionsObject.java
@@ -49,6 +49,7 @@ import com.oracle.truffle.api.dsl.Cached;
 import com.oracle.truffle.api.interop.InteropLibrary;
 import com.oracle.truffle.api.interop.InvalidArrayIndexException;
 import com.oracle.truffle.api.interop.TruffleObject;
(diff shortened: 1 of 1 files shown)


fniephaus:
LGTM. Since the specialization is behing a boundary, there's no point to use a profile (now that InteropExceptions are no longer slow path exceptions).

boris-spas:
Merged in merge commit e66f81e1a327af3eb60294e205121d3b9c0c797a.

PR #2673, opened 2020-07-14
native-image - Register String.getBytes() for JNI access at runtime

jaikiran:
`String.getBytes()` gets used through JNI, for example[1]. The commit in this PR registers it for JNI runtime access to avoid issues like the one reported in Quarkus here[2]

[1] https://github.com/openjdk/jdk/blob/master/src/java.base/share/native/libjava/jni_util.c#L831
[2] https://github.com/quarkusio/quarkus/issues/10682


jaikiran:
Hello @cstancu, do you think this change looks OK? We just had one another report in the Quarkus project with this same issue (in a different code path) https://github.com/quarkusio/quarkus/issues/12495

PR #2682, opened 2020-07-16
UnsafeSubstitutionsTest stores object to int field

nick-arm:
Running this test on AArch64 with -XX:-UseCompressedOops crashes the VM
with a SIGBUS. This is caused by a call to unsafePutObjectVolatile with
a non-64-bit-aligned offset (12). The VM should throw an exception here
rather than crash, and that needs to be fixed separately (it's a variant
of JDK-8246051), but we should also fix the test to use the offset of
the correct field.


teshull:
Hi @nick-arm. Do you or someone else from Arm plan to add in the support for exception throwing?

nick-arm:
> Hi @nick-arm. Do you or someone else from Arm plan to add in the support for exception throwing?

Yes I'll send a patch to the JDK list. We just need to add `GuardUnsafeAccess` to some additional unsafe implementations in `src/hotspot/share/prims/unsafe.cpp`.

PR #2683, opened 2020-07-16
Allow NULL for result in poly_context_eval / poly_value_execute

YaSuenag:
We have to set `poly_value` for result when we call `poly_context_eval()` or `poly_value_execute()`. However we don't need to set it when the callee does not return any values as below:

```
#define SIMPLE_CALL "console.log('SIMPLE_CALL')"
#define ARROW_FUNC  "() => { console.log('ARROW_FUNC') }"
```

Currently if we pass `NULL` as result, we will encounter crash due to segmentation fault.  
If this PR is merged, we can pass `NULL` to them as below:

```
  poly_context_eval(thread, context, "js", "SIMPLE_CALL", SIMPLE_CALL, NULL);
```
```
  poly_value func;
  poly_context_eval(thread, context, "js", "ARROW_FUNC", ARROW_FUNC, &func);
  poly_value_execute(thread, func, NULL, 0, NULL);
```


substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
```diff
@@ -481,7 +481,10 @@ public static PolyglotStatus poly_context_eval(PolyglotIsolateThread thread, Pol
             String jCode = CTypeConversion.toJavaString(source_utf8);
 
             Source sourceCode = Source.newBuilder(languageName, jCode, jName).build();
-            result.write(createHandle(c.eval(sourceCode)));
+            Value evalResult = c.eval(sourceCode);
+            if (result.isNonNull()) {
```
vjovanov, line 485: Why wouldn't we write the `null` polyglot value? Is it for performance reasons?

YaSuenag: `null` is different from "return value is nothing", so it should not set in this case.
This PR proposes to allow null value as the result variable when the caller does not need it.

vjovanov: Does `c.eval(sourceCode)` ever return `null`? Do you have an example where this makes a difference? The native API should really be exactly the same as the original unless there are strong performance reasons.

YaSuenag: It would not return `null`, it is described in [Javadoc](https://www.graalvm.org/sdk/javadoc/org/graalvm/polyglot/Context.html#eval-org.graalvm.polyglot.Source-).

If we write the code in Java, we can ignore the result because `eval` returns it as return value. However native API needs an argument to store the result even if it is not needed in caller. This PR does not inconsists Polyglot API - it equivalent to ignore the result.

vjovanov: I don't fully understand. Do you have an example of the code that can't be done with the current way? If not, do you have an example where the code is faster or shorter with this change?

Do you want to avoid the `isNonNull` call in the native part?

YaSuenag: I think following example would earn benefits of this PR.

```
  const char *LANG = "js";
  poly_isolate isolate;
  poly_thread thread;
  poly_context context;

  poly_create_isolate(NULL, &isolate, &thread);
  poly_create_context(thread, &LANG, 1, &context);

/*
  poly_value value;
  poly_context_eval(thread, context, "js", "SIMPLE_CALL", "console.log('SIMPLE_CALL')", &value);
*/
  poly_context_eval(thread, context, "js", "SIMPLE_CALL", "console.log('SIMPLE_CALL')", NULL); // CRASH!
```

`console.log()` would [return _undefined_ to caller](https://console.spec.whatwg.org/#logger), so we can ignore it. However if we pass `NULL` to 5th argument of `poly_context_eval()` (it means the result of `value`), GraalVM would crash.

Of course we can avoid the crash when we set the value to 5th argument and ignore it, but I think it is easy to understand if we don't need to pass unneeded value.

YaSuenag: In addition, comments on function declaration for both `poly_value_execute()` and `poly_context_eval()` does not say `result` is mandatory. However if the user does not set it, it causes critical problem (crash).

YaSuenag:
I updated the description for the return value. Please tell me if it is not enough or you have any questions.

substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
vjovanov: This now explains it very well and is a valuable addition. Thank you very much!

vjovanov (approved):
LGTM, we can merge this.

vjovanov:
This change has made it into the repo, but the PR was not closed for some reason. Thanks again!

Issue #2693, opened 2020-07-20
NativeImageDebugInfoProvider should publish consistent class and file


adinn:
Class NativeImageDebugInfoProvider provides the source cache manager and debug info generator with details of type names and paths plus associated file names for the various types compiled into a native image. With the current implementation the type name and file name may be mismatched when the details of the type are derived from a HostedType or an AnalysisType. As an example, The Analysis type for java.lang.Class reports its name as java.lang.Class and its file as DynamicHub.java.

The implementation should dereference from HostedType to AnalysisType and from AnalysisType to ResolvedJavaType before constructing the Java class name and FileName. That will ensure that the class name and filename match.

**Steps to reproduce the issue**
Compile HelloWorld.java with flags `-H:GenerateDebugInfo=1 -DLog=DwarfCodeInfo -Dgraal.LogFile=out.log`
Search for text `'Failed to find'` in file `out.log`
Amongst the errors listed will be the message `Failed to find source file for class java.lang.Class`

This is latest Graal head, any arch on jdk11.

PR #2695, opened 2020-07-20
Enforce consistent lookup of class and file names

adinn:
This patch fixes issue  #2693. It ensures that the NativeImageDebugInfoProvider hands out consistent source file names and class names when providing info about types compiled into a native image.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
@@ -99,7 +99,12 @@ class NativeImageDebugInfoProvider implements DebugInfoProvider {
             this.method = method;
             HostedType declaringClass = method.getDeclaringClass();
             Class<?> clazz = declaringClass.getJavaClass();
-            this.javaType = declaringClass.getWrapped();
+            /*
+             * HostedType wraps an AnalysisType and both HostedType and AnalysisType punt calls to
+             * getSourceFilename to the wrapped class so for consistency we need to do the path
+             * lookup relative to the doubly unwrapped HostedType.
+             */
+            this.javaType = declaringClass.getWrapped().getWrapped();
             this.compilation = compilation;
             SourceManager sourceManager = ImageSingletons.lookup(SourceManager.class);
             try (DebugContext.Scope s = debugContext.scope("DebugCodeInfo", declaringClass)) {
@@ -309,8 +314,9 @@ class NativeImageDebugInfoProvider implements DebugInfoProvider {
                 clazz = ((OriginalClassProvider) declaringClass).getJavaClass();
             }
             /*
-             * HostedType and AnalysisType punt calls to getSourceFilename to the wrapped class so
-             * for consistency we need to do the path lookup relative to the wrapped class.
+             * HostedType wraps an AnalysisType and both HostedType and AnalysisType punt calls to
+             * getSourceFilename to the wrapped class so for consistency we need to do the path
+             * lookup relative to the doubly unwrapped HostedType or singly unwrapped AnalysisType.
              */
             if (declaringClass instanceof HostedType) {
                 declaringClass = ((HostedType) declaringClass).getWrapped();
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceManager.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceManager.java
@@ -110,6 +110,9 @@ public class SourceManager {
      * @return the file name or null if it the class cannot be associated with a source file
      */
     private static String computeBaseName(ResolvedJavaType resolvedType) {
+        if (resolvedType.isPrimitive()) {
+            return null;
+        }
         String fileName = resolvedType.getSourceFileName();
         if (fileName == null) {
             /* ok, try to construct it from the class name */


adinn:
The gate tests appear to be falling foul of a problem that relates to use of gdb 7.11. The debuginfo gate tests for both jdk8 and jdk11 see gdb crash when it tries to read the debug info for hello. That doesn't happen when I run locally with with a gdb 8/9 release. I will investigate what happens locally with 7.11 and report back when I find out more.

adinn:
The gate problem is documented in issue #2697. I patched that issue on top of this fix and verified that the 7.11 crash is avoided (n.b. I also checked the info output and ensured that the method which CU which caused the crash was still present but with the corrected variant format).

I have raised a separate PR for #2697 so it can be integrated first then used to ensure that this patch passes the gate tests.

adinn:
This is still failing the debuginfo gate test on AArch64. I noticed that the failing run prints an interesting message when the gdb session starts:

    [hello.hello:8465]        write:   9,313.04 ms,  5.20 GB
    [hello.hello:8465]      [total]: 214,857.79 ms,  5.20 GB
    GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
    . . .
    Reading symbols from /home/travis/build/oracle/graal/substratevm/svmbuild/hello.hello...done.
    Fatal error: Failed to create a new Isolate. (code 7)
    (gdb) set listsize 1
    (gdb) set pagination off
    (gdb) break hello.Hello::main
    Breakpoint 1 at 0x4a4b58: file hello/Hello.java, line 67.
    
    (gdb) run
    (gdb) list
    Checker list bp 1: match 0 failed at line 0 1	/*

    Checker list bp 1 {
      re.compile('67[ \t]+Greeter greeter = Greeter\\.greeter\\(args\\);')
    }

    1	/*

The interesting and unexpected output is `Fatal error: Failed to create a new Isolate. (code 7)`. This precedes the run command so I assume it is being printed out of order.

As fr the rest of the output: the list command is listing text `1	/*` while the regexp is expecting to see line 67 which is the first line in Hello::main. So this indicates that the source file has been but there is no line info for the break address -- that means it defaults to line 1.

I have no idea why this happens on the Travis machines but not on my machine.

olpaw:
> The interesting and unexpected output is Fatal error: Failed to create a new Isolate. (code 7).
> This precedes the run command so I assume it is being printed out of order.

@teshull @peter-hofer it seems isolate creation does not always work reliably on AArch64. Have you seen this before? Could it be that running the image from inside gdb causes the problem? Maybe some strict security setting causes trouble?

peter-hofer:
@adinn this is a problem on `master`, where I've recently merged changes to the image heap that require it to be 1M-aligned. On ARM64, the image heap loaded by the dynamic linker does not satisfy that alignment, so the code tries to create a suitable mapping by looking at `/proc/self/maps`. The issue then is that we get a path of `/var/snap/lxd/common/lxd/storage-pools/instances/containers/travis-job-oracle-graal-712188659/rootfs/home/travis/build/oracle/graal/substratevm/svmbuild/hello.hello`, but our environment is a `chroot`'ed one where we need `/home/travis/build/oracle/graal/substratevm/svmbuild/hello.hello`. I suppose we would have the same issue on AMD64, but don't run into this because the image heap has a suitable alignment.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -309,8 +314,9 @@ private void computeFullFilePathAndCachePath() {
...
             }
             /*
-             * HostedType and AnalysisType punt calls to getSourceFilename to the wrapped class so
-             * for consistency we need to do the path lookup relative to the wrapped class.
+             * HostedType wraps an AnalysisType and both HostedType and AnalysisType punt calls to
+             * getSourceFilename to the wrapped class so for consistency we need to do the path
+             * lookup relative to the doubly unwrapped HostedType.
              */
             if (declaringClass instanceof HostedType) {
                 declaringClass = ((HostedType) declaringClass).getWrapped();
```
zakkak, line 322: Shouldn't this be:

```suggestion
                declaringClass = ((HostedType) declaringClass).getWrapped().getWrapped();
```
?

adinn: No, I don't think so. The result of that single getWrapped() call is an AnalysisType. The next if detects that and unwraps a second time. That second if also deals with the case where an AnalysisType comes in direct.

zakkak: Indeed. So a potential optimization would be to change the following `if` in an `else if` since the result of `HostedType#getWrapped()` returns an `AnalysisType` as you point out.

I might be nit-picking here :)

zakkak: On a second thought, I think this should be fixed, because as it stands it only works because the if statements are in the correct order, making it easy to break things with a simple reordering without any clear indication that these two  statements should not be reordered.

adinn: The if cascade is correct and the comment provides the information which explains why. However, I agree it does not spell it out. I will change the comment to make this easier to infer:

            /*
             * HostedType wraps an AnalysisType and both HostedType and AnalysisType punt calls to
             * getSourceFilename to the wrapped class so for consistency we need to do the path
             * lookup relative to the doubly unwrapped HostedType or singly unwrapped AnalysisType.
             */
I'm not sure we need to protect any further against programmers who transpose conditional operations that side-effect the elements of their conditions. Instead I recommend we just reject their contributions. ;-)

adinn:
@olpaw any chance of getting this reviewed and merged?

olpaw:
> @olpaw any chance of getting this reviewed and merged?

LGTM, now let's see if the CI gods are with us :crossed_fingers:

adinn:
@olpaw Thanks Paul!

Issue #2697, opened 2020-07-21
gdb 7.11 crashes reading debuginfo


adinn:
gdb sometimes crashes reading debuginfo from a Linux native image. The problem occurs when a compile unit (CU) for a method has no associated file and hence has no line entries in the debug_line section. The current info section record for the CU always includes a stmt_list field which holds the offset of the line records in the  debug_line section. When no fields are present it gets written as -1. gdb8/9 treat this as indicating an empty list but gdb7 crashes. Strictly there is no 'empty' value for this field. The CU info record should omit the stmt_list field if the CU contains no associated line records.

n.b. This problem is not straightforward to replicate. It requires at least one method for a synthetic class (say a proxy) to be compiled as a top level method rather than being inlined. The problem showed up when integrating the fix for issue #2693 and the proposed fix (generating a CU info record without a stmt_list field) avoided the crash.

The problem happens on both jdk8 or jdk11+ but only on Linux.

PR #2698, opened 2020-07-21
Omit stmt_list from CU info record when it has no associated line records

adinn:
This patch fixes issue #2697. It was tested by applying it on top of the patch for #2695 where it resolved the gdb 7.11 crash that is stopping that latter patch pass the debuginfotest gate. The fix allows for two variants of the info Compile Unit record for a top level compiled method, one with a stmt_list and one without. The latter  is used when there are no line records associated with a method.

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
@@ -29,7 +29,8 @@ package com.oracle.objectfile.elf.dwarf;
 import com.oracle.objectfile.LayoutDecision;
 import org.graalvm.compiler.debug.DebugContext;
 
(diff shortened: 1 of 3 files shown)


olpaw:
Looks good. Please change your commit message to start with an uppercase letter.

adinn:
This seems to be failing on the AArch64 debuginfotest gate. It looks like the line listing at the 1st breakpoint is showing line 1 instead of line 67 (which is weirdly reminiscent of as earlier AArch64 bug). I think that may well unrelated to this fix. Since this fix corrects the observed problem and it passes on x86_64 I propose to investigate the AArch64 problem in a follow-up issue.

adinn:
Well, I built the latest master (which includes this patch) on AArch64 then ran the debuginfotest gate task with gdb 7.11.1 in my path. The task completed without a problem.

I'm not sure what caused the AArch64 gate test to fail earlier but without being able to reproduce it I certainly can't fix it. I will keep an eye out to see if it ever fails again.

zakkak:
@adinn it looks like it is still failing in travis CI
https://travis-ci.org/github/oracle/graal/jobs/710834126

PR #2700, opened 2020-07-21
Fixes inlined code's cache paths for DWARF generation

zakkak:
Since inlined methods may be defined in a different file than the one
defining a compiled method, the cache paths of all the source files
contributing code to the compiled method need to be set as the
compilation directory to make gdb work.

E.g. If:
1. `foo` invokes `bar` and `baz`
2. `bar` and `baz` get inlined in `foo`
3. `foo`'s source file is cached in `sources/src`
4. `bar`'s source file is cached in `sources/jdk`
5. `baz`'s source file is cached in `sources/graal`

then the compilation directory in `foo`'s compilation unit needs to be
set to `sources/src:sources/jdk:sources/graal`

### DIFF
--- a/substratevm/mx.substratevm/mx_substratevm.py
+++ b/substratevm/mx.substratevm/mx_substratevm.py
@@ -727,7 +727,7 @@ def _debuginfotest(native_image, path, build_only, args):
                          '-H:+VerifyNamingConventions',
                          '-cp', classpath('com.oracle.svm.test'),
                          '-Dgraal.LogFile=graal.log',
-                         '-H:GenerateDebugInfo=1',
+                         '-g',
                          '-H:DebugInfoSourceSearchPath=' + sourcepath,
                          '-H:DebugInfoSourceCacheRoot=' + join(path, 'sources'),
                          'hello.Hello'] + args
--- a/substratevm/mx.substratevm/testhello.py
+++ b/substratevm/mx.substratevm/testhello.py
@@ -30,7 +30,7 @@
 # Assumes you have already executed
 #
 # $ javac hello/Hello.java
-# $ mx native-image -H:GenerateDebugInfo=1 hello.hello
+# $ mx native-image -g hello.hello
 #
 # Run test
 #
--- a/substratevm/mx.substratevm/testhello.py
+++ b/substratevm/mx.substratevm/testhello.py
@@ -34,7 +34,7 @@
 #
 # Run test
 #
-# gdb -d /path/to/sources/src -d /path/to/sources/graal -d /path/to/sources/jdk -x testhello.py /path/to/hello
+# gdb -x testhello.py /path/to/hello
 #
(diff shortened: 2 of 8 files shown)


zakkak:
@adinn could you please review this ?

adinn:
Hi Foivos. I don't think this patch is modelling the information correctly.

The merged cache path you are trying to construct is used to populate a field of a compilation unit (CU). With C++ info records a CU corresponds to a source file. However, in the current Java info model a CU corresponds to a class rather than a file. This is explained in the comment in DebugInfoBase but I'll repeat it here.

debuginfo section CUs need to be ordered by increasing address range. However, the Graal compiler does not guarantee that methods of classes sharing a common file will be contiguous. If, say, Java file F contains class C and Java file F' contains classes C' and C'' then the Graal compiler does compile all methods of C into a single, continuous range, ditto for C' and C''. Unfortunately, the compiler may still interleave the methods of C between those of C' and C''.

In consequence the merged cache path should really be a property of a ClassEntry not a FileEntry. In the above example, C' and C'' may be app classes but C' might only inline Graal code and C'' might only inline jdk code. C' should used merged path `"sources/src:sources/graal"` while C'' should use path `"sources/src:sources/jdk"`. Your code in ClassEntry.addSubrange that adds elements to the merged set needs to be changed accordingly.

Also, there are a few other things that I think need correcting.
You need to be careful not to merge an empty classpath ("") into the merge set.

When constructing the merged set string in FileEntry.getCachePathsString() you append  ":" after each merge set element. That leaves an extra unnecesary ":" at the end.

Finally, when you move FileEntry.getCachePathsString() to ClassEntry please correct the spelling (missing 'r' in String).

zakkak:
Thanks for the review @adinn 

I somehow got carried away by initial approach of having a cache path per FileEntry (which did make sense then) and didn't think about it. Now that the cache paths are in ClassEntry the code makes much more sense (and also got cleaner).

I added checks for the empty cache path case and also used `String.join()` which take cares of ':' at the end, and is cleaner at the same time.

Oh the typo is also fixed now :)

adinn:
Ok, that looks good modulo the unused import and maybe further style fixes that turn up.

n.b. you can run `mx checkstyle` and `mx eclipseformat -e /path/to/eclipse --primary` to check for such problems in advance

zakkak:
@adinn Can you be more specific about the "unused import"? I don't see any :/ (nor does the IDE)

adinn:
Ah sorry it showed up as a failure in the Travis substrate fullstyle task (first one) but that is now being rerun. It flagged an unnecessary include of HashSet.

zakkak:
Thanks again @adinn

zakkak:
Hi @olpaw, could we get this reviewed and merged please?

zakkak:
Thank you @olpaw !

PR #2722, opened 2020-07-29
enable C++/LLVM interoperability calls of (class) methods and constructors

pichristoph:
These changes enable calling methods of classes and constructors of classes that have been written in C++ and have been compiled to LLVM code. This is done via Sulong and the Truffle Interoperability framework. 

This PR is also for discussion with the Sulong team.


sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
```diff
@@ -116,6 +138,48 @@ static boolean isMemberModifiable(LLVMPointerImpl receiver, String ident,
...
+        for (int i = 0; i < arguments.length; i++) {
+            newArguments[i + 1] = arguments[i];
+        }
+        LLVMInteropType.Clazz clazz = (LLVMInteropType.Clazz) receiver.getExportType();
+
+        Method method = clazz.findMethod(member, newArguments);
+        if (method == null) {
+            throw UnknownIdentifierException.create(member);
+        }
+        LLVMFunction llvmFunction = LLVMLanguage.getContext().getGlobalScope().getFunction(method.getLinkageName());
```
rschatz, line 170: You can use a `@CachedContext(LLVMLanguage.class) context` parameter to get the context in a more efficient way.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
```diff
@@ -116,6 +138,48 @@ static boolean isMemberModifiable(LLVMPointerImpl receiver, String ident,
...
+        if (llvmFunction == null) {
+            CompilerDirectives.transferToInterpreter();
+            final String clazzName = clazz.toString().startsWith("class ") ? clazz.toString().substring(6) : clazz.toString();
+            final String msg = String.format("No implementation of declared method %s::%s (%s) found", clazzName, method.getName(), method.getLinkageName());
+            throw new LLVMLinkerException(msg);
+        }
+
+        LLVMFunctionDescriptor fn = LLVMLanguage.getContext().createFunctionDescriptor(llvmFunction);
+
+        return InteropLibrary.getUncached().execute(fn, newArguments);
```
rschatz, line 180: You can use `@CachedLibrary InteropLibrary interop` to get a more efficient version of the interop library.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
rschatz, line 170: Also, e should probably do some caching here, since this is quite inefficient to do every time we call a method. If both `clazz` and `member`, and the argument array length are all the same, then `llvmFunction` should also be the same, so we can cache it. At compile time, we'd only have the comparisons, and then a direct call, instead of an expensive lookup and an indirect call.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
```diff
@@ -116,6 +138,48 @@ static boolean isMemberModifiable(LLVMPointerImpl receiver, String ident,
...
+        }
+        LLVMFunction llvmFunction = LLVMLanguage.getContext().getGlobalScope().getFunction(method.getLinkageName());
+        if (llvmFunction == null) {
+            CompilerDirectives.transferToInterpreter();
+            final String clazzName = clazz.toString().startsWith("class ") ? clazz.toString().substring(6) : clazz.toString();
+            final String msg = String.format("No implementation of declared method %s::%s (%s) found", clazzName, method.getName(), method.getLinkageName());
+            throw new LLVMLinkerException(msg);
+        }
+
+        LLVMFunctionDescriptor fn = LLVMLanguage.getContext().createFunctionDescriptor(llvmFunction);
```
rschatz, line 178: I don't think we should create a new function descriptor every time we call the function. I think you actually want to use an `LLVMAccessSymbolNode` here to fetch the function pointer that points to this method.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
```diff
@@ -232,6 +296,46 @@ Object readArrayElement(long idx,
...
+
+        @ExportMessage
+        Object readArrayElement(long idx,
+                        @Cached BranchProfile exception) throws InvalidArrayIndexException {
+            try {
+                int index = (int) idx;
+                if (index < type.getMemberCount()) {
+                    return type.getMember(index).getName();
+                } else {
+                    return type.getMethod(index - type.getMemberCount());
```
rschatz, line 330: Should probably be `type.getMethod(...).getName()`.

sulong/tests/com.oracle.truffle.llvm.tests/src/com/oracle/truffle/llvm/tests/interop/CxxMethodsTest.java
```diff
@@ -0,0 +1,199 @@
...
+        } finally {
+            freePoint.execute(point);
+        }
+    }
+
+    @Test
+    public void testConstructor() {
+        Value point = allocPoint.execute();
+        try {
+            constructor.execute(point);
```
rschatz, line 192: Not sure I understand this test. You're passing a point to the constructor? Shouldn't the constructor have zero arguments, and return a `Point`?

rschatz (changes requested):
Thanks, looks good in general!

I've put some comments inline in the code, most are about general Truffle stuff and style issues.

sulong/tests/com.oracle.truffle.llvm.tests/src/com/oracle/truffle/llvm/tests/interop/CxxMethodsTest.java
pichristoph: True, it does indeed not make sense to pass the parameter here. I will change it to `Value point = constructor.newInstance()` and also remove the `execute` call of `allocPoint`.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
```diff
@@ -116,6 +142,84 @@ static boolean isMemberModifiable(LLVMPointerImpl receiver, String ident,
...
+        @SuppressWarnings("unused")
+        @Specialization
+        static Object doResolve(LLVMPointerImpl receiver, String member, Object[] arguments,
+                        @CachedContext(LLVMLanguage.class) LLVMContext context, @CachedLibrary(limit = "5") InteropLibrary interop,
+                        @Cached(value = "asClazz(receiver)", allowUncached = true) LLVMInteropType.Clazz clazz,
+                        @Cached(value = "clazz.findMethodByArguments(receiver, member, arguments)", allowUncached = true) Method method,
+                        @Cached(value = "arguments.length", allowUncached = true) int argCount,
+                        @Cached(value = "method.getName()", allowUncached = true) String methodName,
+                        @Cached(value = "getLLVMFunction(context, method, clazz, member)", allowUncached = true) LLVMFunction llvmFunction,
+                        @Cached(value = "create(llvmFunction)", allowUncached = true) LLVMAccessSymbolNode accessSymbolNode)
```
rschatz, line 181: These `@Cached` values are all unused. I don't think they are necessary, since this is the generic fallback specialization that does the resolving on each call.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
```diff
@@ -116,6 +142,84 @@ static boolean isMemberModifiable(LLVMPointerImpl receiver, String ident,
...
+                        throws UnsupportedMessageException, ArityException, UnsupportedTypeException {
+            Object[] newArguments = addSelfObject(receiver, arguments);
+            return interop.execute(accessSymbolNode.execute(), newArguments);
+        }
+
+        @SuppressWarnings("unused")
+        @Specialization
```
rschatz, line 173: Should replace `doCached`. It usually doesn't pay off to keep the cached versions after the inline cache overflows.

As a nice side effect, when this replaces `doCached`, the `allowUncached = true` arguments in `doCached` shouldn't be needed anymore.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
```diff
@@ -116,6 +142,84 @@ static boolean isMemberModifiable(LLVMPointerImpl receiver, String ident,
...
+                        @Cached(value = "arguments.length", allowUncached = true) int argCount,
+                        @Cached(value = "method.getName()", allowUncached = true) String methodName,
+                        @Cached(value = "getLLVMFunction(context, method, clazz, member)", allowUncached = true) LLVMFunction llvmFunction,
+                        @Cached(value = "create(llvmFunction)", allowUncached = true) LLVMAccessSymbolNode accessSymbolNode)
+                        throws UnsupportedMessageException, ArityException, UnsupportedTypeException, UnknownIdentifierException {
+            Object[] newArguments = addSelfObject(receiver, arguments);
+            LLVMInteropType.Clazz newClazz = asClazz(receiver);
+            Method newMethod = newClazz.findMethodByArguments(receiver, member, arguments);
+            LLVMFunction newLLVMFunction = getLLVMFunction(context, newMethod, newClazz, member);
+            Object newReceiver = context.createFunctionDescriptor(newLLVMFunction);
```
rschatz, line 187: We shouldn't create a new function descriptor here. I wonder whether we need a variant of `LLVMAccessSymbolNode` that can take a dynamic symbol as argument... /cc @Palez

rschatz (commented):
Looks almost ready to merge now, just a few minor things in CommonPointerLibraries.
@Palez you reworked the symbol resolution code recently. Do we have a way for dynamically looking up functions, when the `LLVMFunction` is not compile time constant?

@pichristoph Can you please resolve the merge conflicts? Then I can let it run through our internal gate.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
Palez: Currently the `LLVMAccessSymbolNode` must take a symbol in order to be able to access it. We could just simple create a `LLVMDynAccessSymbolNode` for this use case, something like:

`public abstract class LLVMDynAccessSymbolNode extends LLVMExpressionNode {
   public LLVMDynAccessSymbolNode() {
    }
    public abstract LLVMPointer execute();
    @Specialization
    LLVMPointer doAccess(LLVMSymbol symbol,
                    @CachedContext(LLVMLanguage.class) LLVMContext context) {
        CompilerAsserts.partialEvaluationConstant(symbol);
        if (symbol.hasValidIndexAndID()) {
            int bitcodeID = symbol.getBitcodeID(false);
            if (context.symbolTableExists(bitcodeID)) {
                AssumedValue<LLVMPointer>[] symbols = context.findSymbolTable(bitcodeID);
                int index = symbol.getSymbolIndex(false);
                AssumedValue<LLVMPointer> assumedValue = symbols[index];
                if (assumedValue != null) {
                    return assumedValue.get();
                }
            }
        }
        CompilerDirectives.transferToInterpreter();
        throw new LLVMLinkerException(this, String.format("External %s %s cannot be found.", symbol.getKind(), symbol.getName()));
    }
}`

@pichristoph Would you like me to add this class onto the repo or would you like to do it?

pichristoph: @Palez if it is fine for you, I will add it. Thank you for the suggestion/solution!

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/nodes/others/LLVMDynAccessSymbolNode.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public LLVMDynAccessSymbolNode() {
+    }
+
+    public abstract LLVMPointer execute(LLVMSymbol symbol);
+
+    @Specialization
+    LLVMPointer doAccess(LLVMSymbol symbol,
+                    @CachedContext(LLVMLanguage.class) LLVMContext context) {
+        CompilerAsserts.partialEvaluationConstant(symbol);
```
rschatz, line 55: I don't think this assertion is always correct. I'd just remove it.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
```diff
@@ -116,6 +143,79 @@ static boolean isMemberModifiable(LLVMPointerImpl receiver, String ident,
...
+                        throws UnsupportedMessageException, ArityException, UnsupportedTypeException {
+            Object[] newArguments = addSelfObject(receiver, arguments);
+            return interop.execute(accessSymbolNode.execute(), newArguments);
+        }
+
+        @SuppressWarnings("unused")
+        @Specialization(replaces = "doCached")
+        static Object doResolve(LLVMPointerImpl receiver, String member, Object[] arguments,
+                        @CachedContext(LLVMLanguage.class) LLVMContext context, @CachedLibrary(limit = "5") InteropLibrary interop,
+                        @Cached(value = "create()", allowUncached = true) LLVMDynAccessSymbolNode dynAccessSymbolNode)
```
rschatz, line 177: `allowUncached` for nodes is wrong, that would create a new node for every call when used through an uncached library. Put the `@GenerateUncached` annotation on the `LLVMDynAccessSymbolNode`, then it should work without `allowUncached = true`. Also `value = "create()"` shouldn't be necessary, since that's the default (but not wrong either).

rschatz (approved):

rschatz:
Looks good now, thanks a lot!

I started our internal gate, I'll report if that catches additional non-trivial issues.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/interop/LLVMForeignCallNode.java
```diff
@@ -188,19 +190,20 @@ public static PackForeignArgumentsNode create(FunctionType bitcodeFunctionType,
...
 
-    public LLVMForeignCallNode(LLVMLanguage language, LLVMFunctionDescriptor function, LLVMInteropType interopType, LLVMSourceFunctionType sourceType) {
+    protected LLVMForeignCallNode(LLVMLanguage language, LLVMFunctionDescriptor function, LLVMInteropType interopType, LLVMSourceFunctionType sourceType, Structured returnBaseType, Type escapeType) {
         super(language);
-        this.returnBaseType = getReturnBaseType(interopType);
+        this.returnBaseType = returnBaseType;
         this.getStack = LLVMGetStackFromThreadNode.create();
         this.callNode = DirectCallNode.create(getCallTarget(function));
-        this.prepareValueForEscape = LLVMDataEscapeNode.create(function.getLLVMFunction().getType().getReturnType());
+        this.callNode.forceInlining();
```
rschatz, line 205: I think this is a wrong merge conflicting with e8b0489b4e69108949049e887145d45293eff0e5, which removed the `forceInlining`.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/pointer/CommonPointerLibraries.java
```diff
@@ -119,6 +146,92 @@ static boolean isMemberModifiable(LLVMPointerImpl receiver, String ident,
...
+            LLVMFunction newLLVMFunction = getLLVMFunction(context, newMethod, newClazz, member);
+            Object newReceiver = dynAccessSymbolNode.execute(newLLVMFunction);
+            return interop.execute(newReceiver, newArguments);
+        }
+    }
+
+    static LLVMInteropType.Clazz asClazz(LLVMPointerImpl receiver) throws UnsupportedTypeException {
+        LLVMInteropType type = receiver.getExportType();
+        if (!(type instanceof LLVMInteropType.Clazz)) {
+            throw UnsupportedTypeException.create(new Object[]{receiver}, receiver + " cannot be casted to LLVMInteropType.Clazz");
```
rschatz, line 207: String concatenation is not safe on the fast path (internally this creates quite a lot of code). In this particular case, I'd just use a simple string (i.e. literally `"Receiver cannot be cast to LLVMInteropType.Clazz"`, or maybe better `"Receiver type is not a class."`).

rschatz: (found by our internal gate)

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/interop/LLVMForeignCallNode.java
pichristoph: Indeed, thanks!

PR #2738, opened 2020-08-07
Fix code that assumes suites are under a vcs repository

zakkak:
As of mx 5.270.0 (https://github.com/graalvm/mx/pull/221), mx projects can be built even outside a vcs repository.
This patch makes the necessary changes to allow graal to be built outside a vcs repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,8 @@ matrix:
   - os: linux
     arch: arm64
     env: JDK="jdk11" GATE="build,bootstraplite" PRIMARY="compiler"
+  - os: linux
+    env: JDK="jdk11" WITHOUT_VCS=true
 # GR-16977
 #  - env: JDK="jdk11" GATE="build,test,helloworld" PRIMARY="substratevm"
 
@@ -135,7 +137,14 @@ install:
 script:
   - echo ${JAVA_HOME}
   - ${JAVA_HOME}/bin/java -version
-  - mx --primary-suite-path ${TRAVIS_BUILD_DIR}/${PRIMARY} --J @"-Xmx2g" --java-home=${JAVA_HOME} gate --strict-mode --tags ${GATE}
+  - |
+      if [[ ${WITHOUT_VCS} ]]
+      then
+        rm -rf .git
+        mx --primary-suite-path ${TRAVIS_BUILD_DIR}/substratevm native-image --help
+      else
+        mx --primary-suite-path ${TRAVIS_BUILD_DIR}/${PRIMARY} --J @"-Xmx2g" --java-home=${JAVA_HOME} gate --strict-mode --tags ${GATE}
+      fi
 after_failure:
   - cat hs_err*
 notifications:
--- a/compiler/mx.compiler/mx_compiler.py
+++ b/compiler/mx.compiler/mx_compiler.py
@@ -1273,7 +1273,10 @@ def _update_graaljdk(src_jdk, dst_jdk_dir=None, root_module_names=None, export_t
 
         vm_name = 'Server VM Graal'
         for d in _graal_config().jvmci_dists:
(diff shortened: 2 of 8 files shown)


dougxc:
Is there some way we can test this in Travis? Maybe `rm -rf .git` before running some tests?

dougxc:
Don't we need to place a `.mx_vcs_root` file in the root directory?

zakkak:
> Is there some way we can test this in Travis? Maybe `rm -rf .git` before running some tests?

There is this https://github.com/graalvm/mx/pull/216

> Don't we need to place a `.mx_vcs_root` file in the root directory?

No it works both with `.mx_vcs_root` and `ci.hocon`

dougxc:
> There is this graalvm/mx#216

I would suggest moving/copying that gate task to the graal repo itself so that it is run for every Graal change, not just every mx change (which are less frequent).

> No it works both with .mx_vcs_root and ci.hocon

At some point, I think we want to break the unnatural coupling to the existence of `ci.hocon` so would prefer that Graal has its own `.mx_vcs_root` file.

zakkak:
@dougxc thanks for the feedback I fixed the comments in https://github.com/oracle/graal/pull/2738/commits/e9615754550a382061af490068257fa0453137bd (not a travis user so please let me know if there is a better way to achieve this)

.travis.yml
```diff
@@ -135,7 +137,14 @@ install:
 script:
   - echo ${JAVA_HOME}
   - ${JAVA_HOME}/bin/java -version
-  - mx --primary-suite-path ${TRAVIS_BUILD_DIR}/${PRIMARY} --J @"-Xmx2g" --java-home=${JAVA_HOME} gate --strict-mode --tags ${GATE}
+  - |
+      if [[ ${RM_GIT} ]]
+      then
+        rm -rf .git
+        mx --primary-suite-path ${TRAVIS_BUILD_DIR}/substratevm native-image --version
```
dougxc, line 144: Why not run the same testing as in the `else` branch?
Also, I would rename `RM_GIT` to `WITHOUT_VCS` as it better documents the intention of this case.

zakkak: > Why not run the same testing as in the `else` branch?

That's a good question and kind of show why this kind of tests might not be that future proof.
Looking at the changes in https://github.com/oracle/graal/pull/2738/commits/05d53b05ae15cab2e6831161ae6a77b7e7213a5d one sees that only 3 suites are affected (sdk, compiler, substratevm).
Now in these suites only a subset of the commands are affected. For instance in substratevm the only affected commmands are `native-image` and `native-image-configure` (the ones invoking `_ensure_vm_built`. As a result, 
`mx --primary-suite-path substratevm gate --tags build` works fine even without this PR.

In order to test all the changes of this PR one needs to run:
1. `mx --primary-suite-path substratevm native-image -version`
2. `mx --primary-suite-path compiler makegraaljdk /tmp/graaljdk`
3. `mx --primary-suite-path sdk javadoc`

but this still allows people to perform changes that could break mx commands performed outside a vcs repository.
E.g.  `mx --primary-suite-path substratevm build`

Unfortunately we cannot (it would take too long) test every mx command of graal, so I just settled with `mx native-image`.
WDYT?

> Also, I would rename `RM_GIT` to `WITHOUT_VCS` as it better documents the intention of this change.

:+1: I'll take care of that.

dougxc: Ok, then we'll start with the testing you've added and expand as necessary (i.e. when we find other commands that need to be fixed to work without VCS).

pjkendall982: Thank u

zakkak: Thanks @dougxc!

dougxc: Have you looked into the Travis failures on this PR?

zakkak: Travis looks happy now :)

PR #2747, opened 2020-08-11
Feature/cppfields: enabled inherited class field access via Sulong's interoperability library

pichristoph:
Enable accessing inherited fields directly via the interoperability of Sulong/Truffle, such that `object.inheritedField` is also possible now (in contrast to the (only) former possibility: `object.superclass.inheritedField`).
To be discussed (with Sulong team) before merging.


pichristoph:
This PR is outdated. The new pull request (that also contains the changes of this PR) can be found at PR #2932 .

PR #2755, opened 2020-08-14
Fix #1788 when using native-image-agent to generate config.

### LINKED ISSUE #1788: [native-image] native-image-agent may miss some configuration data for getMethod and getField
`Class.getMethod` and `Class.getField` may return a `java.lang.reflect.Method` or `java.lang.reflect.Field` of some super class of the given `Class`. To make `getMethod` and `getField` work properly at runtime, some necessary reflection data should be constructed for the sub class in the current implementation. However, the native-image-agent does not handle this case and may miss some configuration data.

Here is an example.

```java
import java.lang.reflect.Method;
public class Super {

    static class Sub extends Super {}
    static class Sub2 extends Super {
        public static void test() {
            System.out.println("sub2");
        }
    }
    public static void test() {
        System.out.println("super");
    }

        Class<?> clazz = args.length == 0 ? Sub.class : Sub2.class;
        Method m = clazz.getMethod("test");
        m.invoke(null);
    }
}
```
In the above example, `Sub2` is added to prevent native-image from changing the reflection call to direct call. Let us suppose that we will never pass any argument to start the program. So that we can run the program with the native-image-agent to record the call of `getMethod("test")` and use the generated `reflect-config.json` to guide the compilation of `native-image`.

The generated `reflect-config.json` will only contain a single entry. However, `Sub.class.getMethod("test")` requires that class `Sub` itself should be configured. And a `NoSuchMethodException` will be thrown at runtime. 

```json
[
{
  "name":"Super",
  "methods":[{"name":"test","parameterTypes":[] }]
}
]
```

Here is a workaround.

```diff
diff --git a/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/ReflectionProcessor.java b/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/ReflectionProcessor.java
index e7844ca043..ddbe9781cf 100644
--- a/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/ReflectionProcessor.java
+++ b/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/ReflectionProcessor.java
@@ -133,6 +133,7 @@ class ReflectionProcessor extends AbstractProcessor {
                 // fall through
             case "getField": {
                 configuration.getOrCreateType(clazzOrDeclaringClass).addField(singleElement(args), memberKind, unsafeAccess);
+                if (!clazzOrDeclaringClass.equals(clazz)) configuration.getOrCreateType(clazz).setAllPublicFields();
                 break;
             }
 
@@ -147,6 +148,7 @@ class ReflectionProcessor extends AbstractProcessor {
                     parameterTypes = Collections.emptyList();
                 }
                 configuration.getOrCreateType(clazzOrDeclaringClass).addMethod(name, SignatureUtil.toInternalSignature(parameterTypes), memberKind);
+                if (!clazzOrDeclaringClass.equals(clazz)) configuration.getOrCreateType(clazz).setAllPublicMethods();
                 break;
             }
 
```

tianxiaogu:
1. Record the name of the receiver of `getMethod/getField` rather than
`getDeclaringClass` of `Method`/`Field` in `ReflectionProcessor`
2. Use `getMethod` to register method in `ReflectionRegistryAdapter`

### DIFF
--- a/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/ReflectionProcessor.java
+++ b/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/ReflectionProcessor.java
@@ -145,7 +145,11 @@ class ReflectionProcessor extends AbstractProcessor {
                 memberKind = ConfigurationMemberKind.DECLARED;
                 // fall through
             case "getField": {
-                configuration.getOrCreateType(clazzOrDeclaringClass).addField(singleElement(args), memberKind, false, unsafeAccess);
+                String theClass = clazzOrDeclaringClass;
+                if (memberKind != ConfigurationMemberKind.DECLARED && !clazzOrDeclaringClass.equals(clazz)) {
+                    theClass = clazz;
+                }
+                configuration.getOrCreateType(theClass).addField(singleElement(args), memberKind, false, unsafeAccess);
                 break;
             }
 
@@ -159,7 +163,11 @@ class ReflectionProcessor extends AbstractProcessor {
                 if (parameterTypes == null) { // tolerated and equivalent to no parameter types
                     parameterTypes = Collections.emptyList();
                 }
-                configuration.getOrCreateType(clazzOrDeclaringClass).addMethod(name, SignatureUtil.toInternalSignature(parameterTypes), memberKind);
+                String theClass = clazzOrDeclaringClass;
+                if (memberKind != ConfigurationMemberKind.DECLARED && !clazzOrDeclaringClass.equals(clazz)) {
+                    theClass = clazz;
+                }
+                configuration.getOrCreateType(theClass).addMethod(name, SignatureUtil.toInternalSignature(parameterTypes), memberKind);
                 break;
             }
 
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/config/ReflectionRegistryAdapter.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/config/ReflectionRegistryAdapter.java
@@ -25,6 +25,7 @@
 package com.oracle.svm.hosted.config;
 
 import java.lang.reflect.Executable;
+import java.lang.reflect.Field;
 import java.util.List;
 
 import org.graalvm.nativeimage.impl.ReflectionRegistry;
@@ -100,7 +101,11 @@ public class ReflectionRegistryAdapter implements ReflectionConfigurationParserD
 
     @Override
     public void registerField(Class<?> type, String fieldName, boolean allowWrite, boolean allowUnsafeAccess) throws NoSuchFieldException {
-        registry.register(allowWrite, allowUnsafeAccess, type.getDeclaredField(fieldName));
+        try {
+            registry.register(allowWrite, allowUnsafeAccess, type.getDeclaredField(fieldName));
+        } catch (NoSuchFieldException e) {
+            registry.register(allowWrite, allowUnsafeAccess, type.getField(fieldName));
+        }
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


peter-hofer:
Thank you for your contribution @tianxiaogu and apologies that your original report was left unanswered.
I believe there is a misunderstanding. The agent and configuration files always deal with declaring classes. When intercepting `getField` and `getMethod`, the agent determines the declaring class of the member that was looked up (see code references below). In the code from your original report #1788, you have two static methods which are unrelated to each other except that they are in the same class hierarchy. However, the agent will record the method that is called correctly. Because in your program, you call only one of the two methods in a single execution, you will need to run the agent for both cases and use the `config-merge-dir` argument to get a configuration that supports both. I tried it for your example and got the correct reflection configuration:
```
[
{
  "name":"Super",
  "methods":[{"name":"test","parameterTypes":[] }]
},
{
  "name":"Super$Sub2",
  "methods":[{"name":"test","parameterTypes":[] }]
}
]

```

https://github.com/oracle/graal/blob/5c19690e729a8f492a60c15481e6486100563ec7/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java#L302-L308

https://github.com/oracle/graal/blob/5c19690e729a8f492a60c15481e6486100563ec7/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java#L449-L455

tianxiaogu:
I am an employee of Alibaba and Alibaba has signed the OCA. Would you please help me check the OCA status of Alibaba?

@peter-hofer Thanks a lot for your reply. The problem is that if we did not create any entry for `Sub` in the configuration file (i.e., `reflect.json`), class `Sub` will not be able to be registered in the `ReflectionDataBuilder`. If `Sub` is not registered in `ReflectionDataBuilder`, it has no runtime reflection data. Thus calling `getMethod` on `Sub.class` will throw `NoSuchMethodException` at runtime, which should not happen.

peter-hofer:
I have registered your email address for our OCA checks, the status should soon update.

peter-hofer:
Thank you for clarifying @tianxiaogu . You're right, with the example in #1788, the call to `Sub.getMethod("test")` should succeed. However, it is only necessary to add an entry for `Sub` in the reflection configuration so that it is assigned runtime reflection data during the image build, so I think that is what the agent should do for `getMethod` and `getField` on subclasses.

peter-hofer:
On a related note, calling `Sub2.class.getMethod("test")` will fail if method `Sub2.test()` is not registered in the reflection configuration, even when method `Super.test()` and class `Sub2` itself are registered. This might seem like a bug, but it is deliberate because during the image build, we filter _shadowed_ methods such as `Super.test()` that are overridden in classes like `Sub2` and any of their subclasses. (This is actually implicit with the way reflection data is gathered, but more complex with JNI lookup information)

tianxiaogu:
> Thank you for clarifying @tianxiaogu . You're right, with the example in #1788, the call to `Sub.class.getMethod("test")` should succeed. However, it is only necessary to add an entry for `Sub` in the reflection configuration so that it is assigned runtime reflection data during the image build, so I think that is what the agent should do for `getMethod` and `getField` on subclasses. The change is as simple as adding:
> 
> ```
> if (!clazz.equals(clazzOrDeclaringClass)) {
>     // ensure that the queried class is assigned reflection data
>     configuration.getOrCreateType(clazz);
> }
> ```

Thank you @peter-hofer! Your solution is more concise and much better :) Would you like to make a patch to fix the issue so that I can close this pull request?

peter-hofer:
Fixed in 1b42c794087070d8fac505a581beca025d1043a8. Thanks!

PR #2763, opened 2020-08-18
Add intrinsics for unsafe atomics on AArch64

SwapnilGaikwad:
This PR contains two commits implementing intrinsics for different weak atomics that :

1) add intrinsic support for byte/short methods in unsafe API.
Update the tests for Int/Long atomic intrinsics to reflect their
existing support. Also, update the AtomicReadAndAddOp for AMD64 to
preserve the sign while operating on byte/word size operands.

This patch improves performance of Unsafe.getAndSet/Add on byte/short by
about 27.6% on A72 machine.

2) add intrinsic support for the following Unsafe API methods.
i) compareAndExchange: acquire, release
ii) compareAndSet: plain, acquire, release
for byte, boolean, char short, int, long types.

### DIFF
--- a/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
+++ b/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
@@ -35,6 +35,7 @@ import org.graalvm.compiler.asm.Label;
 import org.graalvm.compiler.asm.aarch64.AArch64Assembler;
 import org.graalvm.compiler.asm.aarch64.AArch64MacroAssembler;
 import org.graalvm.compiler.asm.aarch64.AArch64MacroAssembler.ScratchRegister;
+import org.graalvm.compiler.core.common.memory.MemoryOrderMode;
 import org.graalvm.compiler.lir.LIRInstructionClass;
 import org.graalvm.compiler.lir.LIRValueUtil;
 import org.graalvm.compiler.lir.Opcode;
@@ -60,7 +61,7 @@ public class AArch64AtomicMove {
         public static final LIRInstructionClass<CompareAndSwapOp> TYPE = LIRInstructionClass.create(CompareAndSwapOp.class);
 
         private final AArch64Kind accessKind;
-        private final int memoryBarrier;
+        private final MemoryOrderMode memoryOrder;
 
         @Def protected AllocatableValue resultValue;
         @Alive protected Value expectedValue;
@@ -69,7 +70,7 @@ public class AArch64AtomicMove {
         @Temp protected AllocatableValue scratchValue;
 
         public CompareAndSwapOp(AArch64Kind accessKind, AllocatableValue result, Value expectedValue, AllocatableValue newValue, AllocatableValue addressValue, AllocatableValue scratch,
-                        int memoryBarrier) {
+                                MemoryOrderMode memoryOrder) {
             super(TYPE);
             this.accessKind = accessKind;
             this.resultValue = result;
@@ -77,7 +78,7 @@ public class AArch64AtomicMove {
             this.newValue = newValue;
             this.addressValue = addressValue;
             this.scratchValue = scratch;
-            this.memoryBarrier = memoryBarrier;
+            this.memoryOrder = memoryOrder;
         }
 
         @Override
@@ -90,17 +91,11 @@ public class AArch64AtomicMove {
             Register newVal = asRegister(newValue);
             Register expected = asRegister(expectedValue);
 
-            // For the atomics with "acquire/release" suffixes, the corresponding memory barrier has
-            // been inserted after/before the CAS respectively. Thus, the CAS operation here does
-            // not need the "acquire" semantics. Contrarily, the strong variants, i.e., atomics
-            // without "acquire/release" suffixes, are not surrounded by memory barriers. Thus, the
-            // CAS operation needs to use the variant of instruction that has the "acquire" effect.
-            boolean acquire = memoryBarrier >= (LOAD_LOAD | LOAD_STORE | STORE_STORE);
-
-            // The exclusive store is always using the "release" semantic. That is required to
-            // ensure visibility to other threads of the exclusive write (assuming it succeeds)
-            // before that of any subsequent writes.
-            boolean release = true;
+            /**
+             * Determining whether acquire and/or release semantics are needed.
+             */
+            boolean acquire = (memoryOrder.postReadBarriers & (LOAD_LOAD | LOAD_STORE)) != 0;
+            boolean release = (memoryOrder.preWriteBarriers & (LOAD_STORE | STORE_STORE)) != 0;
 
             if (AArch64LIRFlagsVersioned.useLSE(masm.target.arch)) {
                 masm.mov(Math.max(size, 32), result, expected);
--- a/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
+++ b/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
@@ -105,14 +105,37 @@ public class AArch64AtomicMove {
                 // We could avoid using a scratch register here, by reusing resultValue for the
                 // stlxr success flag and issue a mov resultValue, expectedValue in case of success
                 // before returning.
+
+                /*
+                 * If both an acquire and release are requested, a full barrier is expected, and no
+                 * instructions should reordered across the atomic operations. Unfortunately, since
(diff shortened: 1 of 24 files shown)


compiler/src/org.graalvm.compiler.asm.amd64/src/org/graalvm/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -2520,6 +2520,10 @@ public final void movswl(Register dst, AMD64Address src) {
         AMD64RMOp.MOVSX.emit(this, DWORD, dst, src);
     }
 
+    public final void movswl(Register dst, Register src) {
```
teshull, line 2523: This looks to be no longer needed

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
```diff
@@ -56,34 +59,52 @@
...
+            // been inserted after/before the CAS respectively. Thus, the CAS operation here does
+            // not need the "acquire" semantics. Contrarily, the strong variants, i.e., atomics
+            // without "acquire/release" suffixes, are not surrounded by memory barriers. Thus, the
+            // CAS operation needs to use the variant of instruction that has the "acquire" effect.
+            boolean acquire = memoryBarrier >= (LOAD_LOAD | LOAD_STORE | STORE_STORE);
+
+            // The exclusive store is always using the "release" semantic. That is required to
+            // ensure visibility to other threads of the exclusive write (assuming it succeeds)
+            // before that of any subsequent writes.
+            boolean release = true;
```
teshull, line 103: FYI a release only affects instructions before the memory access, not after it. Subsequent memory operations can be hoisted. W.r.t. another thread, an acquire is needed to be issued by the other thread as well in order to have any ordering guarantees.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
```diff
@@ -133,25 +154,26 @@ public AtomicReadAndAddOp(AArch64Kind kind, AllocatableValue result, Allocatable
         @Override
         public void emitCode(CompilationResultBuilder crb, AArch64MacroAssembler masm) {
             assert accessKind.isInteger();
-            final int size = accessKind.getSizeInBytes() * Byte.SIZE;
+            final int srcSize = accessKind.getSizeInBytes() * Byte.SIZE;
```
teshull, line 157: Instead of srcSize and dstSize, can you use more descriptive names, like memAccessSize and addSize?

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
```diff
@@ -1171,24 +1140,56 @@ public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Rec
...
+            this.memoryBarriersKind = memoryBarriersKind;
+            this.accessKind = accessKind;
+            this.isLogic = isLogic;
+        }
+
+        @Override
+        public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver unsafe, ValueNode object, ValueNode offset, ValueNode expected, ValueNode newValue) {
+            // Emits a null-check for the otherwise unused receiver
+            unsafe.get();
+            if (memoryBarriersKind == MemoryBarriersKind.RELEASE) {
```
teshull, line 1182: Enforcing the orderings can be done directly within the CompareAndSwapNodes so that these memory barriers are not needed. See my other general comment

teshull (changes requested):

teshull:
In general for the different compareAndSwap variants, ordering needs to be enforced both at the software level (i.e. compiler cannot do illegal code movements) and at the hardware level (proper hw memory barriers and/or acquire/release instruction). 

From the software side, the MembarNodes are not needed as the CompareAnd(Swap|Exchange)Nodes implement `SingleMemoryKill` which prevents the compiler reorderings.

On the hardware side, I think it is best to pass any required HW orderings to AArch64AtomicMove::CompareAndSwapOp and then generate the appropriate instructions with the orderings instead of using the MembarNodes

I implemented a prototype solution which builds on your PR here: https://github.com/teshull/graal/tree/tshull/prototype_aarch64_atomics
I think my general concept is correct, but I did it quickly so it may be buggy.

Let me know what you think.

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64AtomicMove.java
```diff
@@ -56,61 +60,82 @@
...
+                 * Note that only dmb is needed to prevent instructions before the atomic operation
+                 * from executing too late. All instructions after the atomic operation are ordered
+                 * after the ldaxr and, if it occurs, will also see updated value since it is a
+                 * stxr.
+                 */
+                boolean fullBarrier = acquire && release;
+                if (fullBarrier) {
+                    masm.dmb(AArch64Assembler.BarrierKind.ANY_ANY);
+                }
+
```
XiaohongGong, line 130: Hi @teshull thanks for looking at the atomic patches.  For the fullbarrier mode, do you mean Graal will always insert the "dmb" before CAS, even it is generated for the Java Unsafe intrinisics?  I added the flag `useBarrier` to distinguish the JAVA atomics (should be` ldaxr+stlxr`) and vm internal atomics which needs conservative barriers (see https://github.com/oracle/graal/pull/2181).  
Regarding to the vm internal CAS, here is a related discussing thread: https://patchwork.kernel.org/project/linux-arm-kernel/patch/1391516953-14541-1-git-send-email-will.deacon@arm.com/

Regarding to this patch, it looks the final memory barriers are not the same with the above suggested ones.  WDYT? Please correct me if I have any misunderstanding. Thanks!

teshull: Hi @XiaohongGong. The code here is out of date. Here is what I am planning to merge:
```
/*
* Because the store is only conditionally emitted, a dmb is needed for performing a
* release.
*
* Furthermore, even if the stlxr is emitted, if both acquire and release semantics
* are required, then a dmb is anyways needed to ensure that the instruction
* sequence:
*
* A -> ldaxr -> stlxr -> B
*
* can not be executed as:
*
* ldaxr -> B -> A -> stlxr
*/
if (release) {
masm.dmb(AArch64Assembler.BarrierKind.ANY_ANY);
}

Register scratch = asRegister(scratchValue);
Label retry = new Label();
Label fail = new Label();
masm.bind(retry);
masm.loadExclusive(memAccessSize, result, address, acquire);
AArch64Compare.gpCompare(masm, resultValue, expectedValue);
masm.branchConditionally(AArch64Assembler.ConditionFlag.NE, fail);
masm.storeExclusive(memAccessSize, scratch, newVal, address, false);
// if scratch == 0 then write successful, else retry.
masm.cbnz(32, scratch, retry);
masm.bind(fail);
```

I realized that since the store is only conditionally emitted, it was not possible to tie any release semantics to it. Anyways, I think for a cas, since its is both an acquire & release, that a dmb is necessary due to the link you provided.

XiaohongGong: So to prevent the following reordering:
```
* A -> ldaxr -> stlxr -> B
*
* can not be executed as:
*
* ldaxr -> B -> A -> stlxr
```
Do you think a DMB after `stlxr` is needed? Also due to the discussions inside the thread I pointed, it seems that the final memory barrier mode should be: `ldxr, stlxr, dmb` , which is different from what you mentioned above `(dmb, ldaxr, stxr)`. Any others that I was missing or misunderstanding? Thanks!

teshull: I believe `dmb, ldaxr, stxr` should act a full barrier the same way as 'ldxr, stlxr, dmb`. However, we can't use `ldxr, stlxr, dmb`, as the store is only conditionally executed.

Within `dmb, ldaxr, stxr`, the dmb prevent all prior instructions from executing. Likewise, the ldaxr prevents all subsequent instructions from execution

XiaohongGong: Thanks for more explanation! It seems reasonable.

teshull:
This has been merged: 8acb8e9391b9383ee228a3972925b65c8d15d19f

PR #2766, opened 2020-08-19
Fixing AArch64 emitCompare and zr/sp issues

teshull:
Previously the AArch64 emitCompare was issuing many unnecessary loads.
In addition, some of the logic for using zr/sp within cmp, sub(s), and
add(s) needed to be fixed.

### DIFF
--- a/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
+++ b/compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
@@ -919,17 +919,25 @@ public class AArch64MacroAssembler extends AArch64Assembler {
         super.csinc(size, dst, zr, zr, condition.negate());
     }
 
+    private static ExtendType getLSLExtendType(int size) {
+        assert size == 32 || size == 64;
+        return size == 32 ? ExtendType.UXTW : ExtendType.UXTX;
+    }
+
     /**
      * dst = src1 + src2.
      *
      * @param size register size. Has to be 32 or 64.
-     * @param dst general purpose register. May not be null.
-     * @param src1 general purpose register. May not be null.
+     * @param dst general purpose register. May not be null. Can be zr if src1 != sp. Can be sp if
+     *            src1 != zr.
+     * @param src1 general purpose register. May not be null. Can be zr if dst != sp. Can be sp if
+     *            dst != zr.
      * @param src2 general purpose register. May not be null or stackpointer.
      */
     public void add(int size, Register dst, Register src1, Register src2) {
+        assert !(dst.equals(sp) && src1.equals(zr)) && !(dst.equals(zr) && src1.equals(sp));
         if (dst.equals(sp) || src1.equals(sp)) {
-            super.add(size, dst, src1, src2, ExtendType.UXTX, 0);
+            super.add(size, dst, src1, src2, getLSLExtendType(size), 0);
         } else {
             super.add(size, dst, src1, src2, ShiftType.LSL, 0);
         }
@@ -939,31 +947,15 @@ public class AArch64MacroAssembler extends AArch64Assembler {
      * dst = src1 + src2 and sets condition flags.
      *
      * @param size register size. Has to be 32 or 64.
-     * @param dst general purpose register. May not be null.
+     * @param dst general purpose register. May not be null or stackpointer.
      * @param src1 general purpose register. May not be null.
      * @param src2 general purpose register. May not be null or stackpointer.
      */
     public void adds(int size, Register dst, Register src1, Register src2) {
-        if (dst.equals(sp) || src1.equals(sp)) {
-            super.adds(size, dst, src1, src2, ExtendType.UXTX, 0);
+        if (src1.equals(sp)) {
+            super.adds(size, dst, src1, src2, getLSLExtendType(size), 0);
         } else {
             super.adds(size, dst, src1, src2, ShiftType.LSL, 0);
-        }
-    }
-
-    /**
-     * dst = src1 - src2 and sets condition flags.
-     *
-     * @param size register size. Has to be 32 or 64.
-     * @param dst general purpose register. May not be null.
-     * @param src1 general purpose register. May not be null.
-     * @param src2 general purpose register. May not be null or stackpointer.
-     */
-    public void subs(int size, Register dst, Register src1, Register src2) {
-        if (dst.equals(sp) || src1.equals(sp)) {
-            super.subs(size, dst, src1, src2, ExtendType.UXTX, 0);
-        } else {
-            super.subs(size, dst, src1, src2, ShiftType.LSL, 0);
         }
     }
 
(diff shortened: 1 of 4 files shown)


teshull:
@XiaohongGong @adinn @a74nh Would someone mind looking at this to double check that my changes make sense. It is passing all of our internal testing.

XiaohongGong:
Sure! I will take a look at it today. Thanks!

XiaohongGong:
Hi @teshull , this patch is ok to me. Thanks!

compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
```diff
@@ -919,17 +919,25 @@ public void cset(int size, Register dst, ConditionFlag condition) {
         super.csinc(size, dst, zr, zr, condition.negate());
     }
 
+    private static ExtendType getLSLExtendType(int size) {
```
a74nh, line 922: Probably drop the LSL from the method name. Graal's ExtendType enum doesn't mention LSL.

compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
```diff
@@ -1802,28 +1813,24 @@ public void fmadd(int size, Register dst, Register src1, Register src2, Register
...
      *
      * @param size register size. Has to be 32 or 64.
-     * @param x general purpose register. May not be null or stackpointer.
+     * @param x general purpose register. May not be null.
      * @param y general purpose register. May not be null or stackpointer.
      */
     public void cmp(int size, Register x, Register y) {
         assert size == 32 || size == 64;
-        super.subs(size, zr, x, y, ShiftType.LSL, 0);
+        subs(size, zr, x, y);
```
a74nh, line 1821: Why drop the super. here? Is it because x might be sp?

compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
```diff
@@ -1802,28 +1813,24 @@ public void fmadd(int size, Register dst, Register src1, Register src2, Register
...
     }
 
     /**
      * Compares x to y and sets condition flags.
      *
      * @param size register size. Has to be 32 or 64.
-     * @param x general purpose register. May not be null or stackpointer.
+     * @param x general purpose register. May not be null or zero-register.
      * @param y comparison immediate, {@link #isComparisonImmediate(long)} has to be true for it.
```
a74nh, line 1829: Do we want a assert(isComparisonImmediate(y)) check here?

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64LIRGenerator.java
```diff
@@ -384,31 +385,44 @@ protected boolean emitCompare(PlatformKind cmpKind, Value a, Value b, Condition
...
-                left = load(bExt);
+            /*
+             * The AArch64 comparison instruction can treat register 31 as the stack pointer
+             * register for the left operand, but not for the right operand.
+             */
+            boolean aIsStackPointer = ValueUtil.isRegister(aExt) && ValueUtil.asRegister(aExt).equals(AArch64.sp);
+            boolean bIsStackPointer = ValueUtil.isRegister(bExt) && ValueUtil.asRegister(bExt).equals(AArch64.sp);
+
+            if (aIsStackPointer && bIsStackPointer) {
+                // move b to a non-sp register
```
a74nh, line 396: Ok, this turns "cmp(sp,sp)" into "mov(r10,sp); cmp(r10,sp)"
For cosmetic reasons, I want to say that you should update the right so that it does "mov(r10,sp); cmp(r10,r10)". But it's not going to make any difference.

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64LIRGenerator.java
```diff
@@ -384,31 +385,44 @@ protected boolean emitCompare(PlatformKind cmpKind, Value a, Value b, Condition
...
                 right = loadNonConst(aExt);
                 mirrored = true;
             } else {
-                left = load(aExt);
+                left = aExt;
                 right = loadNonConst(bExt);
                 mirrored = false;
             }
-            append(new AArch64Compare.CompareOp(left, loadNonCompareConst(right)));
+            append(new AArch64Compare.CompareOp(loadReg(left), loadNonCompareConst(right)));
```
a74nh, line 410: All the changes in this patch make sense individually. But, I can't see what's actually removing the unnecessary loads. Is it the use of loadReg instead or load?

compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
teshull: Yes, I dropped the super so that it can call the aarch64macroassembler subs, which handles more combinations of xzr and sp

compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
teshull: I wanted to have LSL in the name, since that is what I'm trying to encode there (it just so happens which type is LSL depends on the instruction size)

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64LIRGenerator.java
teshull: Yes. Using load will introduce moves for values already assigned a registervalue. In general it would be a good idea to remove all unneeded `load`s in the code

compiler/src/org.graalvm.compiler.asm.aarch64/src/org/graalvm/compiler/asm/aarch64/AArch64MacroAssembler.java
teshull: Yeah, I agree. done

compiler/src/org.graalvm.compiler.core.aarch64/src/org/graalvm/compiler/core/aarch64/AArch64LIRGenerator.java
teshull: true. changed

PR #2775, opened 2020-08-21
Resolve #2740: LibraryFactory.createDispatched throw IllegalArgumentE…

### LINKED ISSUE #2740: LibraryFactory.createDispatched does not throw IllegalArgumentException for negative limit
 - GraalVM version or commit id if built from source: **20.1.0**
 - CE or EE: **CE**
 - JDK version: **JDK11**
 - OS and OS Version: **Arch Linux**
 - Architecture: **amd64**
 - The output of `java -Xinternalversion`:
```
OpenJDK 64-Bit Server VM (11.0.7+10-jvmci-20.1-b02) for linux-amd64 JRE (11.0.7+10-jvmci-20.1-b02), built on Apr 21 2020 13:02:22 by "buildslave" with gcc 7.3.0
```

No, but the relevant code still looks the same in current master (as of 807a399b51ed79f685796b8ca77d5e47ee9e0f0f).

**Describe the issue**
The javadoc for `LibraryFactory.createDispatched(int limit)` claims that supplying a negative limit will throw an `IllegalArgumentException`, but in fact it will return an uncached library just like for a limit of 0.

**Code snippet or code repository that reproduces the issue**
JUnit test snippet:
```java
	@Test(expected = IllegalArgumentException.class)
	public void myTest() {
		InteropLibrary.getFactory().createDispatched(-1);
	}
```
You can probably paste that into the JUnit test suite of some language implementation (I stuck it in my [BuiltinTest.java](https://github.com/lucaswerkmeister/graaleneyj/blob/8d4658c131c41b2dcdc79f56832d8bff07a7d11a/language/src/test/java/de/lucaswerkmeister/graaleneyj/test/BuiltinTest.java), and it “worked”, in that JUnit complained that the expected exception had not been thrown).

[This reproducer repository](https://github.com/lucaswerkmeister/repro-truffle-LibraryFactory-createDispatched-negative) is a fork of SimpleLanguage with the above JUnit snippet added. 

1. `git clone --depth 2 https://github.com/lucaswerkmeister/repro-truffle-LibraryFactory-createDispatched-negative repro`
2. `cd repro`
3. `mvn test`

**Expected behavior**
An `IllegalArgumentException` is thrown, as the javadoc claims. (That said, I think it would also be acceptable to resolve this issue by changing the javadoc and leaving the code the same.)

**Additional context**
<details><summary>Details</summary>

The code has looked more or less the same since its initial version of 3bba01a33bb8, where `createCachedDispatch(int limit)` looked like this:

```java
	/**
	 * Returns an cached and dispatched version of this library.
	 *
	 * @see Library#createCachedDispatch(Class, int) for further details.
	 * @since 1.0
	 */
	public final T createCachedDispatch(int limit) {
	    if (limit <= 0) {
	        return getUncachedDispatch();
	    } else {
	        return createCachedDispatchImpl(limit);
	    }
	}
```

The javadoc was added in 21e7d8325655, with no change to the code in that commit, so it now read:

```java
	/**
	 * Creates a new cached and automatically dispatched library given a limit. The limit specifies
	 * the number of cached instances that will be automatically dispatched until the dispatched
	 * library rewrites itself to an uncached version of the library. If the limit is zero then the
	 * library will use an uncached version from the start. Negative values will throw an
	 * {@link IllegalArgumentException}. It is discouraged to use {@link Integer#MAX_VALUE} as
	 * parameter to this method. Reasonable values for the limit range from zero to ten.
	 * <p>
	 * If possible it is recommended to not use this method manually but to use
	 * {@link CachedLibrary} instead.
	 * <p>
	 * Whenever the limit is reached for a node and the uncached version is in use, the current
	 * enclosing node will be available to the uncached library export of the library using
	 * {@link NodeUtil#getCurrentEncapsulatingNode()}.
	 *
	 * @see NodeUtil#getCurrentEncapsulatingNode()
	 * @see CachedLibrary
	 * @since 1.0
	 */
	@TruffleBoundary
	public final T createDispatched(int limit) {
	    if (limit <= 0) {
	        return getUncached();
	    } else {
	        return createDispatchImpl(limit);
	    }
	}
```

The method has stayed mostly unchanged since then.
</details>

pgillet:
…xception for negative limit

### DIFF
--- a/truffle/src/com.oracle.truffle.api.library.test/src/com/oracle/truffle/api/library/test/GenerateLibraryTest.java
+++ b/truffle/src/com.oracle.truffle.api.library.test/src/com/oracle/truffle/api/library/test/GenerateLibraryTest.java
@@ -162,8 +162,13 @@ public class GenerateLibraryTest extends AbstractLibraryTest {
             // expected
         }
 
-        // not really useful but shouldn't fail
-        createCachedDispatch(SampleLibrary.class, -1);
+        // Test create a dispatched library with negative limit
+        try {
+            createCachedDispatch(SampleLibrary.class, -1);
+            fail();
+        } catch (IllegalArgumentException e) {
+            // expected
+        }
 
         try {
             createCachedDispatch(InvalidLibrary.class, 0);
--- a/truffle/src/com.oracle.truffle.api.library/src/com/oracle/truffle/api/library/LibraryFactory.java
+++ b/truffle/src/com.oracle.truffle.api.library/src/com/oracle/truffle/api/library/LibraryFactory.java
@@ -266,7 +266,11 @@ public abstract class LibraryFactory<T extends Library> {
      */
     @TruffleBoundary
     public final T createDispatched(int limit) {
-        if (limit <= 0) {
+        if (limit < 0) {
+            throw new IllegalArgumentException(String.format("Negative limit: %d, expected greater than or equal to 0", limit));
+        }
+
+        if (limit == 0) {
             return getUncached();
         } else {
             ensureLibraryInitialized();


fniephaus (approved):

pgillet:
Regarding my OCA, please do not confuse pascalgillet -(at)- ymail -(dot)- com (Yahoo) with  pascalgillet -(at)- gmail -(dot)- com (gmail).

pgillet:
Hi, can this MR be merged or does it need further development/investigations? Thanks.

Issue #1610, opened 2019-08-12
[native-image] Generating random numbers


tomekw:
I wanted to get a random element from a sequence using `(rand-nth seq)` but I was always getting the same value. After digging into to the source code I figured out it uses `(. Math (random))` under the hood.

I've prepared a sample repo to reproduce the issue. It can be found here: https://github.com/tomekw/random

I had to initialize the main class at build time to compile the project. Maybe that's the cause?

**EDIT**: I'm on GraalVM 19.1.1 CE and on MacOS

tomekw:
Also I just created simple Java class:

```java
public class RandomNumber {
    public static void main(String[] args) {
        System.out.println(Math.random());
    }
}
```

and run

```bash
javac RandomNumber.java
native-image RandomNumber
```

This works:

```bash
java RandomNumber
```

but

```bash
./randomnumber
```

always returns the same value.

borkdude:
This does seem to work:

``` 
import java.util.Random;

public class RandomNumber {
    public static void main(String[] args) {
        Random x = new Random();
        System.out.println(x.nextDouble());
    }
}
``` 
so it might be a random seed issue. Probably the re-used `Random` object is seeded using the current compilation time, not the current run time.

jroper:
I just encountered exactly the same issue with `java.util.concurrent.ThreadLocalRandom`, this prorgam:

```java
public class Test {
    public static void main(String... args) {
        System.out.println(java.util.concurrent.ThreadLocalRandom.current().nextInt());
    }
}
```

Always prints out the same thing (per compilation).

jroper:
Ok, so this is interesting, I created this program:

```java
public class Test {
    public static void main(String... args) {
        if (args.length > 0) {
            System.out.println(java.util.concurrent.ThreadLocalRandom.current().nextInt());
        }
        new Thread(() -> {
            System.out.println(java.util.concurrent.ThreadLocalRandom.current().nextInt());
            System.exit(0);
        }).start();
    }
}
```

Whether you supply an argument to the program or not should have no impact on what's output from the thread that was started, since it's using thread local random, and so that random instance should be local to it. But as you can see, that's not the case, they're using a shared `Random` instance:

```
$ graal-test 
-347521900
$ graal-test 
-347521900
$ graal-test a
-347521900
1780548204
$ graal-test a
-347521900
1780548204
```

So it looks like `ThreadLocalRandom` is not even thread local in GraalVM native image, it's just using one shared `Random` instance.

[...]

PR #2790, opened 2020-08-25
Mark RandomNumbersFeature as an @AutomaticFeature

jaikiran:
As a fix for https://github.com/oracle/graal/issues/1610 the `RandomNumbersFeature` was introduced https://github.com/oracle/graal/commit/294c0b1ad97e5ef196165719f27e11a2087900f0. This feature however, right now, isn't automatically enabled. This means that users have to enable it themselves or frameworks have to enable it for them. Here's a recent issue reported in Quarkus https://github.com/quarkusio/quarkus/issues/11573.

Given that this is such a core feature/necessity of the Java random APIs, enabling this automatically, IMO, would be better.

The commit here adds the `@AutomaticFeature` to enable this.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/RandomNumbersFeature.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/RandomNumbersFeature.java
@@ -26,10 +26,12 @@ package com.oracle.svm.core.jdk;
 
 import java.util.concurrent.ThreadLocalRandom;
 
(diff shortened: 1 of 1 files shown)


cstancu:
I merged this in https://github.com/oracle/graal/commit/f364581654ad444f5b22e7154706289e5f5f8a9a. `ThreadLocalRandom` needed some special treatment, similar to `SplittableRandom`, to avoid pulling `SecureRandom` in even the simplest images.

jaikiran:
Thank you @cstancu

PR #2791, opened 2020-08-25
Fix activation code of VSCode extension

fniephaus:
The section is called ‘env.osx’, not ‘env.mac’ on macOS. Add a case for Windows. Also, throw a better error when `env` is not found.

Please assign to @dbalek.


fniephaus:
Bumping this PR as the extension otherwise does not work on macOS. Please integrate, thanks!

fniephaus:
Thanks!

PR #2801, opened 2020-08-27
[vscode] setWaitForClose() on client connect

fniephaus:
This avoids that a TruffleLSP server waits for a client disconnect if none is connected.

Please assign to @dbalek.


fniephaus:
This is still an issue on `master`. Please take a look

PR #2828, opened 2020-09-02
Unify source caches

zakkak:
This PR fixes https://github.com/graalvm/mandrel/issues/108.

As discussed in https://github.com/graalvm/mandrel/issues/108, the different packaging approaches between GraalVM CE and Mandrel results in sources being packed in different places.
The current implementation of the source caches relies on a specific mapping of packages to places, which is correct only for GraalVM CE.
Furthermore it requires hard-coding package prefixes to specific variables (see https://github.com/oracle/graal/blob/3cda6f3705113b8a513b02d4984b2f2f8c1b80e5/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceManager.java#L171-L203)

This patch refactors the source caches implementation to make it packaging-agnostic.
In the new implementation source roots are searched in the following order until the source is fount or the roots are exhausted:
1. JDK source roots
2. GraalVM source roots
3. Application source roots

That means that in order to detect an application source file we go through the JDK and GraalVM source roots.
As shown in https://github.com/graalvm/mandrel/issues/108#issuecomment-684897400 this doesn't seem to impact performance though (we may go through more sources but we avoid the string comparisons for finding the source cache type).


zakkak:
Investigating travis failure

zakkak:
@adinn can you please review this?

adinn (approved):
Hi Foivos,

This looks good to me.

zakkak:
Hi @olpaw could you please be the next to review this?

olpaw (approved):

olpaw:
@zakkak our CI detects issues with your code when running on JDK 15
```
/b/b/e/main/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceCache.java:123: error: reference to newFileSystem is ambiguous
            FileSystem srcFileSystem = FileSystems.newFileSystem(srcZipPath, null);
                                                  ^
  both method newFileSystem(Path,ClassLoader) in FileSystems and method newFileSystem(Path,Map<String,?>) in FileSystems match
Note: Some input files use or override a deprecated API that is marked for removal.
Note: Recompile with -Xlint:removal for details.
```
Please add a cast to the `null` that disambiguates the method that should be used.

zakkak:
Thanks @olpaw 

> Please add a cast to the `null`-argument that disambiguates the method that should be used.

Done.

zakkak:
That was fast, thanks a lot @olpaw

PR #2834, opened 2020-09-08
Report warning for unsupported system properties

ziyilin:
### Overview
The Java system properties are not consistent between traditional JDK version and native image version. Some system properties are missing and some are different. User may expect the same behavior between traditional and native image, or at least explicit message to warn them about the change. This patch checks the usage of such system properties and reports warning at both native image build time and run time.

### Features

- Check the usage of the missing system properties at native image build time and report warning messages as follows before writing image.
image
- Report using missing system property exception at run time

### System Property Differences
Here lists the detailed differences between traditional and native image version on JDK11. 

Property Name| Traditional JDK value | Native Image value
-- | -- | --
awt.toolkit | sun.awt.X11.XToolkit | N/A
file.encoding | UTF-8 | UTF-8
file.separator | / | /
java.awt.graphicsenv | sun.awt.X11GraphicsEnvironment | N/A
java.awt.printerjob | sun.print.PSPrinterJob | N/A
java.class.path | bin | EMPTY
java.class.version | 55 | 55
java.home | /home/cengfeng.lzy/tools/labsjdk-ce-11.0.8-jvmci-20.2-b03 | N/A
java.io.tmpdir | /tmp | /tmp
java.library.path | /usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib | EMPTY
java.runtime.name | OpenJDK Runtime Environment | N/A
java.runtime.version | 11.0.8+10-jvmci-20.2-b03 | N/A
java.specification.name | Java Platform API Specification | Java Platform API Specification
java.specification.vendor | Oracle Corporation | Oracle Corporation
java.specification.version | 11 | 11
java.vendor | GraalVM Community | Oracle Corporation
java.vendor.url | https://www.graalvm.org/ | https://www.graalvm.org/
java.vendor.url.bug | https://github.com/oracle/graal/issues | N/A
java.version | 11.0.8 | 11.0.8
java.version.date | 2020/7/14 | N/A
java.vm.compressedOopsMode | Zero based | N/A
java.vm.info | mixed mode | N/A
java.vm.name | OpenJDK 64-Bit Server VM | Substrate VM
java.vm.specification.name | Java Virtual Machine Specification | Java Virtual Machine Specification
java.vm.specification.vendor | Oracle Corporation | Oracle Corporation
java.vm.specification.version | 11 | 11
java.vm.vendor | GraalVM Community | Oracle Corporation
java.vm.version | 11.0.8+10-jvmci-20.2-b03 | GraalVM dev Java 11
jdk.debug | release | N/A
line.separator | \r\n | \r\n
os.arch | amd64 | amd64
os.name | Linux | Linux
os.version | 3.10.0-327.ali2012.alios7.x86_64 | 3.10.0-327.ali2012.alios7.x86_64
path.separator | : | :
sun.arch.data.model | 64 | 64
sun.boot.library.path | /home/cengfeng.lzy/tools/labsjdk-ce-11.0.8-jvmci-20.2-b03/lib | N/A
sun.cpu.endian | little | N/A
sun.cpu.isalist | EMPTY | N/A
sun.io.unicode.encoding | UnicodeLittle | N/A
sun.java.command | PrintAllProperties | N/A
sun.java.launcher | SUN_STANDARD | N/A
sun.jnu.encoding | UTF-8 | UTF-8
sun.management.compiler | HotSpot 64-Bit Tiered Compilers | N/A
sun.os.patch.level | unknown | N/A
user.country | US | N/A
user.dir | /home/cengfeng.lzy/svmTestsForCommunity/systemPropertyTest | /home/cengfeng.lzy/svmTestsForCommunity/systemPropertyTest
user.home | /home/cengfeng.lzy | /home/cengfeng.lzy
user.language | en | N/A
user.name | cengfeng.lzy | cengfeng.lzy
user.timezone | EMPTY | N/A

Legends:

- **EMPTY** stands for empty string
- **N/A** means not supported, calling `System.getProperty` with corresponding property name results **null**. 

The contents in this table are from the result of running following code:
```
        Properties p = System.getProperties();
        Enumeration keys = p.keys();
        while (keys.hasMoreElements()) {
            String key = (String) keys.nextElement();
            String value = (String) p.get(key);
            if(value.length()==0) value = "EMPTY";
            System.out.println(key + "=" + value);
        }
    }
```
### Limitation
This patch can only check the calling of `System.getProperty()` with string literals. Calling with string variables is missed.


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -141,7 +171,15 @@ public Properties getProperties() {
 
     protected String getProperty(String key) {
         initializeLazyValue(key);
-        return properties.getProperty(key);
+        String ret = properties.getProperty(key);
+        if (ret == null && isNotSupported(key)) {
+            System.out.println("Java system property " + key + " is not available in native image. Please explicitly assign the expected value via -D" + key + "=, or avoid using this property.");
```
christianwimmer, line 176: The VM must never itself print any unsolicited output. So I don't see a way to print a warning message like this.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -141,7 +171,15 @@ public Properties getProperties() {
...
+        }
+        return ret;
+    }
+
+    public static boolean isNotSupported(String key) {
+        return Arrays.stream(UNSUPPORTED_PROPERTIES).anyMatch(s -> s.equals(key));
```
christianwimmer, line 182: General comment: core parts of SVM cannot use streams, because that makes so much of the JDK reachable that the size of a "hello world" explodes.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/UnsupportedSystemPropertyErrors.java
```diff
@@ -0,0 +1,49 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.hosted;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class UnsupportedSystemPropertyErrors {
+    private final Set<String> systemPropertyCalls = new HashSet<>();
```
christianwimmer, line 32: needs to be a data structure that is thread safe, like `ConcurrentHashMap`

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SystemPropertyPlugin.java
```diff
@@ -0,0 +1,66 @@
...
+import org.graalvm.compiler.api.replacements.SnippetReflectionProvider;
+import org.graalvm.compiler.nodes.ValueNode;
+import org.graalvm.compiler.nodes.graphbuilderconf.GraphBuilderContext;
+import org.graalvm.compiler.nodes.graphbuilderconf.InvocationPlugin;
+import org.graalvm.compiler.nodes.graphbuilderconf.InvocationPlugins;
+import org.graalvm.nativeimage.ImageSingletons;
+
+public class SystemPropertyPlugin {
+    @SuppressWarnings("unused")
+    public static void registerInvocationPlugins(ImageClassLoader imageClassLoader, SnippetReflectionProvider snippetReflection, AnnotationSubstitutionProcessor annotationSubstitutions,
```
christianwimmer, line 44: Why do you pass in all these parameters such as `AnnotationSubstitutionProcessor` and `SVMHost`?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SystemPropertyFeature.java
```diff
@@ -0,0 +1,69 @@
...
+import org.graalvm.nativeimage.ImageSingletons;
+
+@AutomaticFeature
+public class SystemPropertyFeature implements GraalFeature {
+    private ImageClassLoader loader;
+    private SVMHost hostVM;
+    private AnnotationSubstitutionProcessor annotationSubstitutions;
+
+    public SystemPropertyFeature() {
+        ImageSingletons.add(UnsupportedSystemPropertyErrors.class, new UnsupportedSystemPropertyErrors());
```
christianwimmer, line 45: Since all of this checking code is running only during image generation, you 
1) don't need to split it up into the separate "Feature", "Errors", and "Plugins" classes
2) you don't need to register a separate singleton because the instance of `SystemPropertyFeature` is automatically available as an image singleton

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -65,7 +66,36 @@
...
                     "java.vm.specification.vendor",
-                    "java.vm.specification.version"
+                    "java.vm.specification.version",
+                    "sun.arch.data.model"
+    };
+
+    /**
+     * System properties are unsupported by default.
+     */
+    private static final String[] UNSUPPORTED_PROPERTIES = {
```
christianwimmer, line 76: "unsupported" is in general the wrong term for all of this PR. They are of course supported (since you can manually set them), they just are not set currently by default.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SystemPropertyPlugin.java
```diff
@@ -0,0 +1,66 @@
...
+                    if (SystemPropertiesSupport.isNotSupported(propertyKey) && ImageSingletons.contains(UnsupportedSystemPropertyErrors.class)) {
+                        UnsupportedSystemPropertyErrors unsupportedPropertyErrors = ImageSingletons.lookup(UnsupportedSystemPropertyErrors.class);
+                        unsupportedPropertyErrors.addUnsupportedInvoke("Unsupported property:" + propertyKey + " is used at " + b.getCode().asStackTraceElement(b.bci()));
+                    }
+                }
+                return false;
+            }
+
+            @Override
+            public boolean isDecorator() {
```
christianwimmer, line 61: No need for this, since the `apply` method always returns `false`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ConfigurationParser.java
```diff
@@ -71,4 +71,4 @@ protected static boolean asBoolean(Object value, String propertyName) {
         }
         throw new JSONParserException("Invalid boolean value '" + value + "' for element '" + propertyName + "'");
     }
-}
+}
```
christianwimmer, line 74: Unnecessary change (and also a checkstyle violation)

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -65,7 +66,36 @@
...
+                    "java.awt.printerjob",
+                    "java.home",
+                    "java.library.path",
+                    "java.runtime.name",
+                    "java.runtime.version",
+                    "java.vendor.url.bug",
+                    "java.vm.info",
+                    "sun.boot.class.path",
+                    "sun.boot.library.path",
+                    "sun.cpu.endian",
```
christianwimmer, line 89: that sounds like a property that can be inherited from image build time. Maybe some other too?

christianwimmer (commented):
Overall this is an interesting idea. Printing warning messages at image build time sounds helpful. It only works when `System.getProperties` is called, with a constant string argument. But that is probably a common use case.

I don't see a way to print a warning message at run time. A VM does not produce output on its own. One possibility would be to make a separate "report at run time option" that the user can manually turn on. In that case, the full stack trace should be printed too - just a warning message without any context information is not actionable.

ziyilin:
It's necessary to explicitly report the using of incompatible system property, otherwise user may get confused and it's difficult to debug the native image. Proper exception message will help the user to locate the problem.

I add a new option `-R:+ReportIncompatibleSystemPropertyError`  to throw `com.oracle.svm.core.jdk.IncompatibleSystemPropertyException` when a system property is null. User can manually turn it on.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ConfigurationParser.java
ziyilin: reverted

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
ziyilin: fixed by changing to INCOMPATIBLE_PROPERTIES

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SystemPropertyFeature.java
ziyilin: Merged UnsupportedSystemPropertyErrors and SystemPropertyPlugin classes into SystemPropertyFeature.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SystemPropertyPlugin.java
ziyilin: Don't need them, removed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/UnsupportedSystemPropertyErrors.java
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
ziyilin: Fixed by using exception here

Jsn2win (approved):

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SystemPropertyFeature.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public void registerInvocationPlugins(Providers providers, SnippetReflectionProvider snippetReflection, InvocationPlugins invocationPlugins, boolean analysis, boolean hosted) {
+        InvocationPlugins.Registration r = new InvocationPlugins.Registration(invocationPlugins, System.class);
+        r.register1("getProperty", String.class, new InvocationPlugin() {
+            @Override
+            public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver, ValueNode name) {
+                if (name.isConstant()) {
+                    String propertyKey = snippetReflection.asObject(String.class, name.asJavaConstant());
+                    if (SystemPropertiesSupport.isIncompatible(propertyKey)) {
+                        systemPropertyCalls.add("Incompatible property:" + propertyKey + " is used at " + b.getCode().asStackTraceElement(b.bci()));
```
christianwimmer, line 69: Minor: missing space after `:` in the message

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -65,7 +72,36 @@
...
                     "java.vm.specification.vendor",
-                    "java.vm.specification.version"
+                    "java.vm.specification.version",
+                    "sun.arch.data.model"
+    };
+
+    /**
+     * System properties are unsupported by default.
+     */
+    private static final String[] INCOMPATIBLE_PROPERTIES = {
```
christianwimmer, line 82: After thinking about the name for a while, my suggestion is to call it `UNDEFINED_PROPERTIES` (with the appropriate changes of the option and classes, fields, ...). That name avoids all sorts of legal hassles.

christianwimmer (changes requested):
Thanks for the cleanups, the code looks pretty good now. I added a few more comments.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/IncompatibleSystemPropertyException.java
```diff
@@ -0,0 +1,7 @@
+package com.oracle.svm.core.jdk;
```
christianwimmer, line 1: License header missing on this file

christianwimmer: You could also just use the `UnsupportedFeatureError` class that we already have.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -48,6 +50,11 @@
  */
 public abstract class SystemPropertiesSupport {
 
+    public static class SystemPropertyFeatureOptions{
+        @Option(help = "Report the usage of incompatible property as exception when it is not explicitly set)")//
+        public static final RuntimeOptionKey<Boolean> ReportIncompatibleSystemPropertyError = new RuntimeOptionKey<>(false);
```
vjovanov, line 55: I would name it `ReportUnsupportedSystemPropertyErrors`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -141,7 +177,20 @@ public Properties getProperties() {
 
     protected String getProperty(String key) {
         initializeLazyValue(key);
-        return properties.getProperty(key);
+        String ret = properties.getProperty(key);
+        if (ret == null && isIncompatible(key) && SystemPropertyFeatureOptions.ReportIncompatibleSystemPropertyError.getValue()) {
+            throw new IncompatibleSystemPropertyException("Java system property " + key + " is incompatible in native image. Please explicitly assign the expected value via -D" + key + "=, or avoid using this property.");
```
vjovanov, line 182: Here as well. I would use `UnsupportedSystemPropertyException`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SystemPropertyFeature.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
```
vjovanov, line 2: 2020

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -65,7 +72,36 @@
...
+                    "sun.arch.data.model"
+    };
+
+    /**
+     * System properties are unsupported by default.
+     */
+    private static final String[] INCOMPATIBLE_PROPERTIES = {
+                    "awt.toolkit",
+                    "file.encoding.pkg",
+                    "java.awt.graphicsenv",
```
vjovanov, line 85: @arodionov we should make sure to enable these once AWT is complete.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -65,7 +72,36 @@
...
+                    "java.awt.printerjob",
+                    "java.home",
+                    "java.library.path",
+                    "java.runtime.name",
+                    "java.runtime.version",
+                    "java.vendor.url.bug",
+                    "java.vm.info",
```
vjovanov, line 92: Can we put SubstrateVM or GraalVM Native Image here?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SystemPropertyFeature.java
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
ziyilin: > After thinking about the name for a while, my suggestion is to call it UNDEFINED_PROPERTIES (with the appropriate changes of the option and classes, fields, ...). That name avoids all sorts of legal hassles.

As Chris suggested, I think "undefined" would be a better name.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
ziyilin: fixed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SystemPropertyFeature.java
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -141,7 +177,20 @@ public Properties getProperties() {
...
-        return properties.getProperty(key);
+        String ret = properties.getProperty(key);
+        if (ret == null && isIncompatible(key) && SystemPropertyFeatureOptions.ReportIncompatibleSystemPropertyError.getValue()) {
+            throw new IncompatibleSystemPropertyException("Java system property " + key + " is incompatible in native image. Please explicitly assign the expected value via -D" + key + "=, or avoid using this property.");
+        }
+        return ret;
+    }
+
+    public static boolean isIncompatible(String key) {
+        for (int i = 0; i < INCOMPATIBLE_PROPERTIES.length; i++) {
```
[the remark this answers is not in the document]
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
christianwimmer: Correct, please use `undefined`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -48,25 +57,126 @@
...
+                    "java.home",
+                    "sun.java.command",
+                    "sun.java.launcher",
+                    "sun.os.patch.level",
+                    "user.country",
+                    "user.language",
+                    "user.timezone"};
+
+    /** Properties are removed or added since JDK8*/
+    private static final List<PropertyInfo> CHANGED_PROPERTIES_SINCE_8 = new ArrayList<>(
```
christianwimmer, line 131: I don't like this "delta" list, since it is difficult to maintain in the long run. Instead, we should change the whole mechansim we have right now (`HOSTED_PROPERTIES` and `initializeProperty()`)  to a `PropertyInfo` like you introduced - without the need for "added" and "deleted" but instead with a predicate `onlyWith` where you can specify the existing classes like `JDK8OrEarlier` or `JDK11OrLater`.
Note also that we do not need to support only JDK 8, 11, and 15 - and not any versions in between.

christianwimmer (changes requested):
Sorry, did not notice when you pushed the latest version. I think we are getting too complicated now, with the "added" / "removed" properties. Simplifying that is a bit of a refactoring of the existing code, but I think worth it.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
ziyilin: > Note also that we do not need to support only JDK 8, 11, and 15 - and not any versions in between.

Do you mean only support 8, 11 and 15, or all the versions in between? I think it should be the later. For example, `java.vm.compressedOopsMode` is introduced in JDK 9, its `onlyWith` should be `JDK9OrLater`.

christianwimmer: I mean "only support 8, 11 and 15". Many parts of native image don't support any of the versions in between already. Sticking to the long term support versions plus "latest" keeps the code base manageable.

ziyilin: Refactored.

Issue #2836, opened 2020-09-08
UserError.Abort can cause MissingFormatArgumentException


DeanWookey:
**Describe the issue**
Building a native image with a generated reflection configuration file as well as jni-config.json, proxy-config.json, and resource-config.json (as described here: https://github.com/oracle/graal/blob/master/substratevm/Configuration.md) may cause an error which causes the UserError.Abort method to fail, resulting in an unhelpful error message.

**Steps to reproduce the issue**
I'm unable to reproduce this since I don't know what's causing the error in my case due to the unhelpful MissingFormatArgumentException error message.

**Describe GraalVM and your environment:**
 - GraalVM version **graalvm-ce-java11-windows-amd64-20.2.0**
 - JDK major version: **11**
 - OS: **Windows 10**
 - Architecture: **AMD64**

```
Fatal error:java.util.MissingFormatArgumentException: Format specifier '%5c'
	at java.base/java.util.Formatter.format(Formatter.java:2672)
	at java.base/java.util.Formatter.format(Formatter.java:2609)
	at java.base/java.lang.String.format(String.java:2897)
	at com.oracle.svm.core.util.UserError.abort(UserError.java:68)
	at com.oracle.svm.hosted.config.ConfigurationParserUtils.doParseAndRegister(ConfigurationParserUtils.java:138)
	at com.oracle.svm.hosted.config.ConfigurationParserUtils.lambda$parseAndRegisterConfigurations$3(ConfigurationParserUtils.java:119)
	at java.base/java.util.stream.ReferencePipeline$4$1.accept(ReferencePipeline.java:212)
	at com.oracle.svm.hosted.config.ConfigurationParserUtils$1.tryAdvance(ConfigurationParserUtils.java:108)
	at java.base/java.util.Spliterator.forEachRemaining(Spliterator.java:326)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:658)
	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:274)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
[...]

PR #2838, opened 2020-09-10
Use the correct UserError.abort API

jaikiran:
Fixes https://github.com/oracle/graal/issues/2836

Hello @dougxc, @cstancu, the commit in this PR should fix the error reported in that linked issue. This now uses an API which doesn't consider the passed string as a "format specifier" and instead considers it a message.


peter-hofer:
Thank you for your contribution @jaikiran . I think we should look at all usages of `UserError.abort` and `UserError.guarantee` and ensure that they use formatting strings instead of string concatentation.

peter-hofer:
Done, thanks for your contribution!

jaikiran:
Thank you for doing this @peter-hofer. Sorry, I've been busy at work and didn't get  a chance to take into account your review input and redo the PR.

PR #2840, opened 2020-09-11
Don't use a container image to produce the native-images

zakkak:
Currently the workflow uses `-Dquarkus.native.container-build=true`
which uses a container image from
https://quay.io/repository/quarkus/ubi-quarkus-native-image?tab=tags
instead of using the build we just created.


galderz (approved):

zakkak:
This was merged in https://github.com/oracle/graal/commit/39416cb9b358bc6ac12601b15e294ef4905144cd

Thanks @dougxc

PR #2846, opened 2020-09-12
Rename HotSpotIntrinsicCandidate to IntrinsicCandidate

marschall:
JDK-8138732 renames HotSpotIntrinsicCandidate to IntrinsicCandidate and moves it to jdk.internal.vm.annotation.

It only seems to be referenced in comments. This change avoids the changes being overwritten the next time a "Update Graal" is done in the JDK.

See https://github.com/openjdk/jdk/pull/45


marschall:
I have signed the OCA, on the [OCA page](https://www.oracle.com/technical-resources/oracle-contributor-agreement.html#m) it says

> Philippe Marschall - BTrace - marschall

That is me

fniephaus (approved):

Issue #2502, opened 2020-05-25
native-image shell script requires a modded JDK to work.


zakkak:
**Describe the issue**
When building Graal bits for use with openJDK, `native-image` needs to pass the following arguments to the jvm `--upgrade-module-path ${location}/../../jvmci/graal.jar --add-modules "org.graalvm.truffle,org.graalvm.sdk" --module-path ${location}/../../truffle/truffle-api.jar:${location}/../../jvmci/graal-sdk.jar` and `-J--add-exports=jdk.internal.vm.ci/jdk.vm.ci.code=jdk.internal.vm.compiler` to native-image.

**Steps to reproduce the issue**
Please include both build steps as well as run steps:
```
curl -sL https://github.com/AdoptOpenJDK/openjdk11-upstream-binaries/releases/download/jdk-11.0.9%2B2/OpenJDK11U-jdk_x64_linux_11.0.9_2_ea.tar.gz -o jdk.tar.gz
curl -sL https://github.com/AdoptOpenJDK/openjdk11-upstream-binaries/releases/download/jdk-11.0.9%2B2/OpenJDK11U-static-libs_x64_linux_11.0.9_2_ea.tar.gz -o jdk-static-libs.tar.gz
mkdir -p openjdk
tar xf jdk.tar.gz -C openjdk --strip-components=1
tar xf jdk-static-libs.tar.gz -C openjdk --strip-components=1
export JAVA_HOME=$(pwd)/openjdk
git clone --depth 1 https://github.com/oracle/graal
pushd graal/substratevm
mx --components="Native Image" build
popd

export GRAAL_REPO=./graal
export MANDREL_JDK=./mandrel

cp -r ${JAVA_HOME} ${MANDREL_JDK}

mkdir ${MANDREL_JDK}/lib/svm
cp ${GRAAL_REPO}/substratevm/mxbuild/dists/jdk1.8/library-support.jar ${MANDREL_JDK}/lib/svm

mkdir ${MANDREL_JDK}/lib/svm/builder
cp ${GRAAL_REPO}/substratevm/mxbuild/dists/jdk11/{svm,pointsto}.jar ${MANDREL_JDK}/lib/svm/builder
cp ${GRAAL_REPO}/substratevm/mxbuild/dists/jdk1.8/objectfile.jar ${MANDREL_JDK}/lib/svm/builder

mkdir ${MANDREL_JDK}/languages
[...]

PR #2852, opened 2020-09-16
Introduce --system-libffi to use system-provided libffi instead of building libffi from source

zakkak:
Adding a --system-libffi flag as discussed in https://github.com/oracle/graal/issues/2502#issuecomment-680764625 (Re 2.)

### DIFF
--- a/sdk/mx.sdk/mx_sdk_vm_impl.py
+++ b/sdk/mx.sdk/mx_sdk_vm_impl.py
@@ -1727,6 +1727,9 @@ class GraalVmNativeImageBuildTask(_with_metaclass(ABCMeta, mx.ProjectBuildTask))
         return 'Building {}'.format(self.subject.name)
 
 
+def _system_libffi():
+    return (mx.get_opts().system_libffi or mx.env_var_to_bool('SYSTEM_LIBFFI'))
+
 class GraalVmBashLauncherBuildTask(GraalVmNativeImageBuildTask):
     def __init__(self, subject, args):
         """
@@ -1777,6 +1780,12 @@ class GraalVmBashLauncherBuildTask(GraalVmNativeImageBuildTask):
             image_config = self.subject.native_image_config
             return ' '.join(image_config.option_vars)
 
+        def _get_system_libffi():
+            if _system_libffi():
+                return '-H:+SystemLibFFI'
+            else:
+                return '-H:-SystemLibFFI'
+
         _template_subst = mx_subst.SubstitutionEngine(mx_subst.string_substitutions)
         _template_subst.register_no_arg('classpath', _get_classpath)
         _template_subst.register_no_arg('jre_bin', _get_jre_bin)
@@ -1784,6 +1793,7 @@ class GraalVmBashLauncherBuildTask(GraalVmNativeImageBuildTask):
         _template_subst.register_no_arg('extra_jvm_args', _get_extra_jvm_args)
         _template_subst.register_no_arg('macro_name', GraalVmNativeProperties.macro_name(self.subject.native_image_config))
         _template_subst.register_no_arg('option_vars', _get_option_vars)
+        _template_subst.register_no_arg('system_libffi', _get_system_libffi)
 
         with open(self._template_file(), 'r') as template, mx.SafeFileCreation(output_file) as sfc, open(sfc.tmpPath, 'w') as launcher:
             for line in template:
@@ -1905,6 +1915,8 @@ class GraalVmSVMNativeImageBuildTask(GraalVmNativeImageBuildTask):
             '--macro:' + GraalVmNativeProperties.macro_name(self.subject.native_image_config),
             '-H:NumberOfThreads=' + str(self.parallelism),
         ]
+        if _system_libffi():
+            build_args += ["-H:+SystemLibFFI"]
         if self.subject.native_image_config.is_polyglot:
             build_args += ["--macro:truffle", "--language:all"]
         return build_args
--- a/sdk/mx.sdk/vm/launcher_template.cmd
+++ b/sdk/mx.sdk/vm/launcher_template.cmd
@@ -52,7 +52,7 @@ for /f "delims=" %%i in ("%relcp:;=!newline!%") do (
 )
 
 set "jvm_args=-Dorg.graalvm.launcher.shell=true "-Dorg.graalvm.launcher.executablename=%executablename%""
-set "launcher_args="
+set "launcher_args=<system_libffi>"
 
 :: Check option-holding variables.
 :: Those can be specified as the `option_vars` argument of the LauncherConfig constructor.
--- a/sdk/mx.sdk/vm/launcher_template.sh
+++ b/sdk/mx.sdk/vm/launcher_template.sh
@@ -46,7 +46,7 @@ for e in "${relative_cp[@]}"; do
 done
 
 jvm_args=("-Dorg.graalvm.launcher.shell=true" "-Dorg.graalvm.launcher.executablename=$0")
-launcher_args=()
+launcher_args=("<system_libffi>")
 
 # Unfortunately, parsing of `--vm.*` arguments has to be done blind:
 # Maybe some of those arguments where not really intended for the launcher but were application arguments
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/c/function/CLibrary.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/c/function/CLibrary.java
@@ -44,6 +44,7 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.util.function.BooleanSupplier;
 
 /**
  * Denotes an external library that needs to be linked in. The annotation can be placed on any
@@ -67,7 +68,21 @@ public @interface CLibrary {
      *
      * @since 19.1.0
      */
-    boolean requireStatic() default false;
+    Class<?> requireStatic() default RequireStaticFalse.class;
+
+    final class RequireStaticFalse implements BooleanSupplier {
+        @Override
+        public boolean getAsBoolean() {
+            return false;
+        }
+    }
+
+    final class RequireStaticTrue implements BooleanSupplier {
+        @Override
+        public boolean getAsBoolean() {
+            return true;
+        }
+    }
 
     /**
      * Specifies the name of the libraries this library depends on.
--- a/substratevm/mx.substratevm/mx_substratevm.py
+++ b/substratevm/mx.substratevm/mx_substratevm.py
@@ -1479,6 +1479,26 @@ def _ensure_vm_built(config):
                 f.write(rev_value)
         build_native_image_image(config)
 
+def _build_libffi():
+    return not (mx.get_opts().system_libffi or mx.env_var_to_bool('SYSTEM_LIBFFI'))
+
+def mx_register_dynamic_suite_constituents(register_project, register_distribution):
+    dependencies = ['dependency:com.oracle.svm.native.libchelper/*',
+                         'dependency:com.oracle.svm.native.darwin/*',
+                         'dependency:com.oracle.svm.native.jvm.posix/*',
+                         'dependency:com.oracle.svm.native.jvm.windows/*']
+    if (_build_libffi()):
+        dependencies.append('extracted-dependency:truffle:LIBFFI_DIST')
+    layout = {
+        '<os>-<arch>/include/': ['extracted-dependency:truffle:TRUFFLE_NFI_NATIVE/include/*',
+                                 'file:src/com.oracle.svm.libffi/include/svm_libffi.h'],
+        '<os>-<arch>/': dependencies}
+    platforms = ["linux-amd64", "darwin-amd64", "windows-amd64"]
+    dist = mx.LayoutTARDistribution(suite, "SVM_HOSTED_NATIVE", [], layout, '', True, None,
+                                    plarforms = platforms, native = True, maven = True,
+                                    description = "SubstrateVM image builder native components")
+    register_distribution(dist)
+
 @mx.command(suite.name, 'native-image')
 def native_image_on_jvm(args, **kwargs):
     config = graalvm_jvm_config()
--- a/substratevm/mx.substratevm/suite.py
+++ b/substratevm/mx.substratevm/suite.py
@@ -1064,34 +1064,6 @@ suite = {
         },
 
         #
-        # Native Projects
-        #
-        "SVM_HOSTED_NATIVE": {
-            "native": True,
-            "platformDependent" : True,
-            "platforms" : [
-                "linux-amd64",
-                "darwin-amd64",
-                "windows-amd64",
-            ],
-            "layout": {
-                "<os>-<arch>/": [
-                    "dependency:com.oracle.svm.native.libchelper/*",
-                    "dependency:com.oracle.svm.native.darwin/*",
-                    "dependency:com.oracle.svm.native.jvm.posix/*",
-                    "dependency:com.oracle.svm.native.jvm.windows/*",
-                    "extracted-dependency:truffle:LIBFFI_DIST",
-                ],
-                "<os>-<arch>/include/": [
-                    "extracted-dependency:truffle:TRUFFLE_NFI_NATIVE/include/*",
-                    "file:src/com.oracle.svm.libffi/include/svm_libffi.h",
-                ]
-            },
-            "description" : "SubstrateVM image builder native components",
-            "maven": True
-        },
-
-        #
         # Internal Distributions
         #
         "SVM_DRIVER": {
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/darwin/DarwinSystemPropertiesSupport.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/darwin/DarwinSystemPropertiesSupport.java
@@ -40,7 +40,7 @@ import com.oracle.svm.core.posix.headers.Limits;
 import com.oracle.svm.core.posix.headers.Unistd;
 import com.oracle.svm.core.posix.headers.darwin.Foundation;
 
-@CLibrary(value = "darwin", requireStatic = true)
+@CLibrary(value = "darwin", requireStatic = CLibrary.RequireStaticTrue.class)
 public class DarwinSystemPropertiesSupport extends PosixSystemPropertiesSupport {
 
     @Override
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/CodeSynchronizationOperations.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/CodeSynchronizationOperations.java
@@ -29,7 +29,7 @@ import org.graalvm.nativeimage.c.function.CFunction;
 import org.graalvm.nativeimage.c.function.CFunction.Transition;
 import org.graalvm.nativeimage.c.function.CLibrary;
 
-@CLibrary(value = "libchelper", requireStatic = true)
+@CLibrary(value = "libchelper", requireStatic = CLibrary.RequireStaticTrue.class)
 public class CodeSynchronizationOperations {
 
     @CFunction(value = "codeSynchronization_clearCache", transition = Transition.NO_TRANSITION)
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
@@ -431,6 +431,14 @@ public class SubstrateOptions {
         return CodeAlignment.getValue();
     }
 
+    @Option(help = "Dynamically link native image to system libffi instead of statically linking to the static libffi shipped with GraalVM.")//
+    public static final HostedOptionKey<Boolean> SystemLibFFI = new HostedOptionKey<>(false);
+
+    @Fold
+    public static boolean useSystemLibFFI() {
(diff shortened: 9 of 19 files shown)


zakkak:
@gilles-duboscq can you please review this PR?

zakkak:
Hi @gilles-duboscq this is a kind reminder.

gilles-duboscq:
I think that looks reasonable.
What do you think @rschatz?

rschatz:
I'm not sure I understand the change fully. Doesn't this just remove the libffi dependency when the new flag is given? It's removing the statically linked `libffi.a` from `libtrufflenfi.so`, but doesn't it need to link `libtrufflenfi.so` to the system `libffi.so` instead?

Also, on the native-image side, it's just removing `libffi.a` from the `SVM_HOSTED_NATIVE` distribution. But native-image itself expects that to be there, it's linking native images statically to `libffi.a`, so I guess with that additional flag it should link them dynamically to the system `libffi.so` instead. Not sure how to do that though.

zakkak:
@rschatz thanks for the review and sorry for my carelessness.

> I'm not sure I understand the change fully. Doesn't this just remove the libffi dependency when the new flag is given? It's removing the statically linked `libffi.a` from `libtrufflenfi.so`, but doesn't it need to link `libtrufflenfi.so` to the system `libffi.so` instead?

Indeed. This is now fixed.

> Also, on the native-image side, it's just removing `libffi.a` from the `SVM_HOSTED_NATIVE` distribution. But native-image itself expects that to be there, it's linking native images statically to `libffi.a`, so I guess with that additional flag it should link them dynamically to the system `libffi.so` instead. Not sure how to do that though.

That's true as well. I am thinking about making `org.graalvm.nativeimage.c.function.CLibrary#requireStatic` a `java.util.function.Function` instead of `boolean`. This way I can use a function to check the value of the env variable `SYSTEM_LIBFFI` (or maybe a property that mx will set accordingly) as `requireStatic` in `com.oracle.svm.truffle.nfi.libffi.LibFFI`.

Suggestions are welcome :)

rschatz:
> > I'm not sure I understand the change fully. Doesn't this just remove the libffi dependency when the new flag is given? It's removing the statically linked `libffi.a` from `libtrufflenfi.so`, but doesn't it need to link `libtrufflenfi.so` to the system `libffi.so` instead?
> 
> Indeed. This is now fixed.

:+1: 

> 
> > Also, on the native-image side, it's just removing `libffi.a` from the `SVM_HOSTED_NATIVE` distribution. But native-image itself expects that to be there, it's linking native images statically to `libffi.a`, so I guess with that additional flag it should link them dynamically to the system `libffi.so` instead. Not sure how to do that though.
> 
> That's true as well. I am thinking about making `org.graalvm.nativeimage.c.function.CLibrary#requireStatic` a `java.util.function.Function` instead of `boolean`. This way I can use a function to check the value of the env variable `SYSTEM_LIBFFI` (or maybe a property that mx will set accordingly) as `requireStatic` in `com.oracle.svm.truffle.nfi.libffi.LibFFI`.
> 
> Suggestions are welcome :)

I think we should somehow build `native-image` in a way that it will always use the system libffi if the `--system-libffi` flag was used at build. Otherwise we'll get into problems if the flag is set differently during `mx build` and during the `native-image` run. Or maybe `native-image` should just detect the existence of the static library?

Of course we could also add a flag to native-image, because even if the static libffi is available, it's still possible to link against the system libffi, just the other way round is not possible.

cc @christianwimmer @peter-hofer (I don't know enough about native-image to be able to say what's best here and how to do that).

zakkak:
> I think we should somehow build `native-image` in a way that it will always use the system libffi if the `--system-libffi` flag was used at build. Otherwise we'll get into problems if the flag is set differently during `mx build` and during the `native-image` run. Or maybe `native-image` should just detect the existence of the static library?
> 
> Of course we could also add a flag to native-image, because even if the static libffi is available, it's still possible to link against the system libffi, just the other way round is not possible.

I like the idea of automatically detecting whether the static libffi is present and falling back to the system libffi if not. Regarding the additional `native-image` flag it might be nice to have for testing, but I don't see why would anyone use it otherwise.

To implement this I suggest refactoring the existing `requireStatic` approach and replacing it with an enum with the following options:
* `STATIC`: Uses static library (eventually fails if not present -- equivalent to current `requireStatic = true`)
* `OPTIONALLY_STATIC`: tries to link against static library and falls back to dynamic using something like:
```java
                Path libraryPath = getStaticLibraryPath(getAllStaticLibs(), lib.value());
                if (libraryPath == null) {
                    addDynamicNonJniLibrary(lib.value());
                } else {
                    addStaticNonJniLibrary(lib.value(), lib.dependsOn());
                }
```
(eventually fails if both static and dynamic libraries are missing)
* `DYNAMIC`: Uses dynamic library (eventually fails if not present -- equivalent to current `requireStatic = false`)

@christianwimmer @peter-hofer WDYT?

olpaw:
> To implement this I suggest refactoring the existing `requireStatic` approach and replacing it with an enum with the following 

There should be no need to make this more complex.

The linker command used to link the native image objectfile into and executable (or shared object) contains `-L<path-to>/lib/svm/clibraries/linux-amd64`. So if you just remove the `requireStatic = true` from `@CLibrary(value = "ffi",...` you should be fine. Then the linker will prefer the dynamic libffi if it is available on your system. Otherwise it will choose the the static library variant from `<path-to>/lib/svm/clibraries/linux-amd64`. (Unless you are building with `--static` where it is the other way around - i.e. the linker prefers the static library variant over the dynamic one)

olpaw:
> Otherwise it will choose the static library variant from `<path-to>/lib/svm/clibraries/linux-amd64`. (Unless you are building with `--static` where it is the other way around - i.e. the linker prefers the static library variant over the dynamic one)

Just think about linking `libffi` in the same fashion as `libz` (see com.oracle.svm.core.jdk.ZLib)

zakkak:
> So if you just remove the `requireStatic = true` from `@CLibrary(value = "ffi",...` you should be fine. Then the linker will prefer the dynamic libffi if it is available on your system. Otherwise it will choose the static library variant from `<path-to>/lib/svm/clibraries/linux-amd64`. (Unless you are building with `--static` where it is the other way around - i.e. the linker prefers the static library variant over the dynamic one)

Is that acceptable though? This will essentially change the behaviour of GraalVM CE even when not using `--system-libffi`, i.e. `native-image` will prefer the system libffi over the static one.

olpaw:
> Is that acceptable though? This will essentially change the behaviour of GraalVM CE even when not using `--system-libffi`, i.e., `native-image` will prefer the system libffi over the static one.

As far as native-image is concerned it makes sense to prefer linking against well established shared libraries if we can expect them to be installed on most systems (like we do with `libz`). Since even `systemd` depends on `libffi` it is probably a good candidate to have it dynamic per default. @christianwimmer do you see a problem with this approach?

olpaw:
A more conservative approach would be to change `org.graalvm.nativeimage.c.function.CLibrary#requireStatic` into
```java
    Class<?> requireStatic() default RequireStaticDefault.class;

    final class RequireStaticDefault implements BooleanSupplier {
        @Override
            return false;
        }
    }
```
This way you can provide a BooleanSupplier that makes requireStatic false for your builds (when `--system-libffi` is used to build graalvm)

rschatz:
I'm not sure systemd using it is a good indicator, that's Linux only, and typically only host Linux, not necessarily container base images. Also, what about MacOS and Windows?

Remember, this is not just about the systems we use to build native images, it's about the system where we deploy these images, that also includes slim container base images and so on. This is adding an additional dependency to all native images that include the Truffle NFI (which is many Truffle languages, either directly or via the LLVM runtime).

We could also just add `libffi.so` to the "language home" of the Truffle NFI language. This is not really different from e.g. a native-image that contains Ruby or Python or the LLVM runtime to have a dependency on the respective "language home" of these languages that contains e.g. Python runtime libraries. Currently the NFI is dependency-free in native-image...

If we decide that this is acceptable for native images, we might as well change it to link dynamically on Hotspot, too. We could still ship a `libffi.so` that we build ourselves with the official GraalVM builds to avoid the extra dependency there.

olpaw:
> If we decide that this is acceptable for native images, we might as well change it to link dynamically on Hotspot, too. We could still ship a `libffi.so` that we build ourselves with the official GraalVM builds to avoid the extra dependency there.

On Fedora that seems to be already the case
```
> dnf repoquery --installed --recursive --whatrequires libffi | grep java-
abrt-java-connector-0:1.2.0-5.fc33.x86_64
java-1.8.0-openjdk-1:1.8.0.272.b10-0.fc33.x86_64
java-1.8.0-openjdk-devel-1:1.8.0.272.b10-0.fc33.x86_64
java-1.8.0-openjdk-headless-1:1.8.0.272.b10-0.fc33.x86_64
java-11-openjdk-1:11.0.9.11-0.fc33.x86_64
java-11-openjdk-devel-1:11.0.9.11-0.fc33.x86_64
java-11-openjdk-headless-1:11.0.9.11-0.fc33.x86_64
java-latest-openjdk-1:15.0.0.36-1.rolling.fc33.x86_64
java-latest-openjdk-devel-1:15.0.0.36-1.rolling.fc33.x86_64
java-latest-openjdk-headless-1:15.0.0.36-1.rolling.fc33.x86_64
java-latest-openjdk-jmods-1:15.0.0.36-1.rolling.fc33.x86_64
java-latest-openjdk-src-1:15.0.0.36-1.rolling.fc33.x86_64
```

rschatz:
Yes, at least on Linux Hotspot already depends on libffi, so on Hotspot I really don't see any problem. This is really just about the deployment targets of native images.

olpaw:
> not necessarily container base images

If you build a native-image for containers nowadays you use muslc + `--static`. In this case the static libffi (if not found elsewhere, from `<path-to>/lib/svm/clibraries/linux-amd64`) is taken.

> Also, what about MacOS and Windows?

Here I'm not so sure. On Windows at least I would opt for `requireStatic` to be `true`. https://github.com/oracle/graal/pull/2852#issuecomment-718013349 would give us the flexibility to set this per platform.

zakkak:
To make sure we are on the same page :)

So @olpaw if I understand correctly you are suggesting switching to system libffi by default on Linux systems -- both for `libtrufflenfi.so` and native images. In that case `--system-libffi` would be a thing only for MacOS and Windows.

This, however, will result in native images depending on libffi being present on the deployment system as @rschatz pointed out.
To solve the latter @rschatz suggests adding `libffi.so` to GraalVM, but that seems to defeat the purpose of using dynamic linking. In order to keep the best of both words it I think that the shipped `libffi` should only by used as a fallback when libffi is not installed on the deployment system.

If we all agree on the above I can start working on the implementation.
Alternatively, I will try out the approach suggested in https://github.com/oracle/graal/pull/2852#issuecomment-718013349.

Thank you for all the feedback and the information :)

olpaw:
> If we all agree on the above I can start working on the implementation.
> Alternatively, I will try out the approach suggested in #2852 (comment).

To get your specific problem solved I'd say go for #2855 .

If we decide to also switch to dynamically linking libffi we can just tweak the `BooleanSupplier` to our needs later.

zakkak:
I implemented https://github.com/oracle/graal/pull/2852#issuecomment-718013349 in https://github.com/oracle/graal/commit/06681ae8aa8cca1605537d3aff8663598732d56f.

I propagate `--system-libffi` as needed to `native-image` but there is an issue with "folding" the flag in the native `native-image`.

Despite using `@Fold` in https://github.com/oracle/graal/commit/06681ae8aa8cca1605537d3aff8663598732d56f#diff-68aa9dff8095b3cafdca34f4f5ae0b69fbc20d19b6d628a8311e58d8fb34b482R496-R503 the `native-image` still gets the value at runtime. How can I overcome this?

olpaw:
> Using `-H:+PrintClassInitialization` I see that the class is not getting initialized at build-time so the above make sense.

Hi @zakkak 

Sorry for the late reply. I will look into this today.

olpaw:
@zakkak while I do have a working solution for a `Class<? extends RequireStaticSupplier> requireStaticSupplier()` in the `@CLibrary` annotation (see https://github.com/oracle/graal/compare/master...olpaw:paw/CLibrary.RequireStaticSupplier ) I want to take a step back and find out why we do have `libffi.a` on the linker command line when you are obviously not building truffle based images. The actual culprit is that we unconditionally have `libffi.a` on the linker command line even when `TruffleNFIFeature` not used at all. I will make changes that ensure that this will not happen anymore. This should then also solve your problem.

zakkak:
Sure, that would be even better :)

Thanks @olpaw

olpaw:
@zakkak I have merged https://github.com/oracle/graal/commit/a5d622b028c99bf4071d4c7036ae6300c052364e to main. This ensures `libffi.a` is only on the linker command line when `TruffleNFIFeature` is used. Please rebase / adjust your PR to those changes.

zakkak:
Hello @olpaw, I gave your branch https://github.com/olpaw/graal/tree/paw/CLibrary.RequireStaticSupplier a spin and it seems to have the same issue (https://github.com/oracle/graal/pull/2852#issuecomment-718256006) with mine.

More specifically:
1. I applied the following patch:
```diff
iff --git a/sdk/mx.sdk/vm/launcher_template.sh b/sdk/mx.sdk/vm/launcher_template.sh
index 28b4d3fff00..fd65930d127 100644
--- a/sdk/mx.sdk/vm/launcher_template.sh
+++ b/sdk/mx.sdk/vm/launcher_template.sh
@@ -46,7 +46,7 @@ for e in "${relative_cp[@]}"; do
 done
 
 jvm_args=("-Dorg.graalvm.launcher.shell=true" "-Dorg.graalvm.launcher.executablename=$0")
-launcher_args=()
+launcher_args=("-H:+SystemLibFFI")
 
 # Unfortunately, parsing of `--vm.*` arguments has to be done blind:
 # Maybe some of those arguments where not really intended for the launcher but were application arguments
diff --git a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java
index 1f03043b367..59beefa2bcc 100644
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/NativeLibraries.java
@@ -610,8 +610,10 @@ public final class NativeLibraries {
             }
 
             if (lib.requireStatic() || supplier.getAsBoolean()) {
+                System.out.println("Adding " + lib.value() + " as static");
                 addStaticNonJniLibrary(lib.value(), lib.dependsOn());
             } else {
+                System.out.println("Adding " + lib.value() + " as dynamic");
                 addDynamicNonJniLibrary(lib.value());
             }
         }
```

to force the bash script to always use `-H:+SystemLibFFI` and have `NativeLibraries` print how it links the different libraries.

2. I built the "Native Image" component along with a binary `native-image` using:
```
mx --primary-suite-path substratevm --native=native-image --components='Native Image' build
```

The part of the output from the build of the `native-image` binary was:
```
Building native-image.image... [dependency SVM_DRIVER updated]
[native-image:262233]    classlist:   2,256.80 ms,  0.96 GB
[native-image:262233]        (cap):     791.51 ms,  0.96 GB
[native-image:262233]        setup:   2,750.82 ms,  0.96 GB
Adding ffi as dynamic
Adding pthread as dynamic
Adding dl as dynamic
Adding libchelper as static
Adding jvm as static
Adding java as static
Adding z as dynamic
Adding libchelper as static
Adding rt as dynamic
[native-image:262233]     (clinit):     219.73 ms,  1.72 GB
[native-image:262233]   (typeflow):   9,135.67 ms,  1.72 GB
[native-image:262233]    (objects):   8,872.28 ms,  1.72 GB
[native-image:262233]   (features):     587.57 ms,  1.72 GB
[native-image:262233]     analysis:  19,369.49 ms,  1.72 GB
[native-image:262233]     universe:     467.80 ms,  1.72 GB
[native-image:262233]      (parse):   2,244.87 ms,  1.72 GB
[native-image:262233]     (inline):   5,075.91 ms,  2.24 GB
[native-image:262233]    (compile):  15,558.79 ms,  3.19 GB
[native-image:262233]      compile:  23,860.54 ms,  3.19 GB
[native-image:262233]        image:   2,357.99 ms,  3.19 GB
[native-image:262233]        write:     252.71 ms,  3.19 GB
[native-image:262233]      [total]:  51,620.72 ms,  3.19 GB
```

So we see that when building `native-image` the `-H:+SystemLibFFI` flag is respected (`Adding ffi as dynamic`).

3. I used the freshly built binary `native-image` to build a HelloWorld with:
```
./sdk/latest_graalvm_home/bin/native-image Hello
``` 

The output of the above is:
```
[hello:264143]    classlist:   5,244.81 ms,  1.21 GB
[hello:264143]        (cap):   1,578.31 ms,  1.21 GB
[hello:264143]        setup:   4,920.13 ms,  1.21 GB
Adding ffi as static
Adding pthread as dynamic
Adding dl as dynamic
Adding libchelper as static
Adding jvm as static
Adding java as static
Adding z as dynamic
Adding libchelper as static
Adding rt as dynamic
[hello:264143]     (clinit):     450.11 ms,  1.71 GB
[hello:264143]   (typeflow):   9,577.13 ms,  1.71 GB
[hello:264143]    (objects):  10,494.28 ms,  1.71 GB
[hello:264143]   (features):     689.26 ms,  1.71 GB
[hello:264143]     analysis:  21,822.37 ms,  1.71 GB
[hello:264143]     universe:   1,000.83 ms,  1.71 GB
[hello:264143]      (parse):   2,970.25 ms,  1.71 GB
[hello:264143]     (inline):   5,347.72 ms,  2.63 GB
[hello:264143]    (compile):  23,241.91 ms,  2.69 GB
[hello:264143]      compile:  33,560.03 ms,  2.69 GB
[hello:264143]        image:   4,888.65 ms,  3.81 GB
[hello:264143]        write:     342.39 ms,  3.81 GB
[hello:264143]      [total]:  72,303.17 ms,  3.81 GB
```

Showing that `-H:+SystemLibFFI` did not persist in the binary `native-image` (`Adding ffi as static`).

olpaw:
> Showing that `-H:+SystemLibFFI` did not persist in the binary `native-image` (`Adding ffi as static`).

How would it? You added `-H:+SystemLibFFI` to the `launcher_template.sh` that will be used to create the bash launcher. The bash launcher is only used when you run the native-image driver as JVM aplication. If you build the driver into an image (the `native-image`-image) the bash launcher is replaced by the image.

To bake some state into the native-image`-image you would use a mechanism like
https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java#L109..L110

The values of theses two system properties have at image build-time get baked into the image because they are assigned to static final fields and everything from `com.oracle.svm.driver` has its static initializers executed at image build time.

But anyway please do not use the changes from https://github.com/olpaw/graal/tree/paw/CLibrary.RequireStaticSupplier. Instead rebase to master and you will not have libffi on the linker command if you do not use the `TruffleNFIFeature`.

olpaw:
> To bake some state into the native-image`-image you would use a mechanism like
> https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java#L109..L110

Another alternative is to modify `SubstrateCompilerFlagsBuilder.compute_graal_compiler_flags_map` in `mx.substratevm/mx_substratevm.py`. If you add your custom flags to the `graal_compiler_flags_base array` each invocation of the image builder will have that flag. I.e. in your case you would like to have `-H:+SystemLibFFI` in `graal_compiler_flags_base array` if `mx.get_opts().system_libffi or mx.env_var_to_bool('SYSTEM_LIBFFI')` is `True`.

zakkak:
> How would it?

In my changes I expected that adding `@Fold` at https://github.com/oracle/graal/pull/2852/files#diff-68aa9dff8095b3cafdca34f4f5ae0b69fbc20d19b6d628a8311e58d8fb34b482R437 would result in persisting the "build-time" returned value in the binary `native-image`. Thanks for the pointers, I'll give it another go :) 

> Instead rebase to master and you will not have libffi on the linker command if you do not use the `TruffleNFIFeature`.

That fixes the "Mandrel issue", but seems orthogonal to me if we still want to introduce the `--system-libffi` flag since the latter should still work when `TruffleNFIFeature` is in use.

What I am trying to say is that to my understanding this PR doesn't really change after the rebase, the only difference is that now one needs to explicitly use `TruffleNFIFeature` to link with libffi (either statically or dynamically), but the rest remain the same.

christianwimmer:
For the Truffle NFI language, I think it is desirable to use a "known good" version of libffi that is shipped with GraalVM. Using an arbitrary older or newer version of libffi that we have never tested with sounds dangerous. Unless libffi itself is really careful to ensure both forward and backward compatibility.

olpaw:
> For the Truffle NFI language, I think it is desirable to use a "known good" version of libffi that is shipped with GraalVM. Using an arbitrary older or newer version of libffi that we have never tested with sounds dangerous. Unless libffi itself is really careful to ensure both forward and backward compatibility.

I fully agree. Also looking into `com.oracle.svm.truffle.nfi.libffi.LibFFIHeaderDirectives` shows that our query code compilation is performed against our libffi header file (`svm_libffi.h` from `lib/svm/clibraries/linux-amd64/include`). Thus not also linking against our libffi static library is a bad idea.

If you really want to link against the system provided libffi you would also need to perform LibFFIHeaderDirectives query code compilation with `ffi.h` header file from your host system. In other words you need provide a custom LibFFIHeaderDirectives for your use-case.

zakkak:
OK. Since the current state only requires FFI at build time, and doesn't require it to run `native-image` when not using `TruffleNFIFeature`, as far as Mandrel is concerned we are more than happy :)

So I am OK with closing this since you think it doesn't add any value (the contrary it appears :) ).

zakkak:
Thank you for all the help!

PR #2871, opened 2020-09-24
Fix errors manifested in renaissance-native-image:chi-square benchmark

d-kozak:
This pull request fixes errors that manifest in the **renaissance-native-image:chi-square** benchmark. 
Native Image Agent did not register com.sun.security.auth security classes. 

Two concrete changes are proposed here. 

1. java.util.** is removed from the com.oracle.svm.configure.trace.AccessAdvisor internalCallerFilter, because it was incorrectly filtering out relevant classes. In this case, those were javax.security.auth.spi.LoginModule implementations, in particular com.sun.security.auth.module.UnixLoginModule . These modules are loaded in javax.security.auth.login.LoginContext, line 687, using a java.util.ServiceLoader. And that's why the java.util.** was removed from the filter.

2. In this benchmark, com.sun.security.auth.UnixPrincipal is being loaded via ClassLoader.loadClass Unfortunately, analyzing the access in com.oracle.svm.agent.BreakpointInterceptor.isLoadClassInvocation causes a ConstantPoolException, because com.oracle.svm.jvmtiagentbase.ConstantPoolTool.seekEntry encounters a UTF-8 tagged entry with length 0, which results in the exception. After allowing the length 0, the UnixPrincipal class is registered correctly.


vjovanov:
This was merged.

PR #2906, opened 2020-10-08
[vscode] Add public API to register LSP notification handlers

fniephaus:
This PR lets GraalVM's extension for VS Code expose a public API that other extensions can use to register handlers for custom client notifications from the GraalLS backend. For this, extensions should depend on the GraalVM extension using `extensionDependencies` (see [here](https://code.visualstudio.com/api/references/vscode-api#extensions)).

Here's how the API can be used in a third-party extension:

```typescript
const graalVMExtension = vscode.extensions.getExtension('oracle-labs-graalvm.graalvm');
if (!graalVMExtension) {
    return console.error('Unable to find GraalVM extension.');
}
graalVMExtension.exports.onClientNotification('my_notification', (response: any) => console.log(response)).then((result: boolean) => {
    if (!result) {
        console.error('Failed to register 'my_notification' handler.');
    }
});
```

The current change is all we need for our Babylonian analysis use case on the client side at this point. In an earlier version, we exposed this functionality through a VS Code command, but exports seem to be the better solution. I am not sure there are any good alternatives, exposing the `languageClient` object to third-parties didn't seem like a good idea.

Please assign to @dbalek. /cc @MartinBalin

### DIFF
--- a/vscode/graalvm/src/extension.ts
+++ b/vscode/graalvm/src/extension.ts
@@ -10,7 +10,7 @@ import * as utils from './utils';
 import { toggleCodeCoverage, activeTextEditorChaged } from './graalVMCoverage';
 import { GraalVMConfigurationProvider, GraalVMDebugAdapterDescriptorFactory, GraalVMDebugAdapterTracker } from './graalVMDebug';
 import { installGraalVM, addExistingGraalVM, installGraalVMComponent, uninstallGraalVMComponent, selectInstalledGraalVM, findGraalVMs, InstallationNodeProvider, Component, Installation, setupProxy, removeGraalVMInstallation } from './graalVMInstall';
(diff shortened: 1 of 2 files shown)


fniephaus:
I've rebased the PR and move from a VS Code command to exposing a public API. This way, it is clear what functionality third party extensions are allowed to depend on. Please have a look and let me know what you think.

fniephaus:
An extension can also access the extension through the following type definition:

```typescript
import { GenericNotificationHandler } from 'vscode-languageclient';

export interface GraalVMExtension {
    onClientNotification(method: string, handler: GenericNotificationHandler): Promise<boolean>;
}
```

It then possible to use:
```typescript
const graalVMExtension = vscode.extensions.getExtension<GraalVMExtension>('oracle-labs-graalvm.graalvm');
```

PR #2912, opened 2020-10-11
Resolve #2347 Add -H:ExcludeResources option to native-image

### LINKED ISSUE #2347: [native-image] Add -H:ExcludeResources option
**Feature request**

`native-image` currently supports the `-H:IncludeResources` and `-H:ResourceConfigurationFiles` options documented in [RESOURCES.md](https://github.com/oracle/graal/blob/master/substratevm/RESOURCES.md#accessing-resources-in-substrate-vm-images )
These options allow to include resources via one or more regular expressions.

I'd like to propose to add a way to restrict the selected set via excludes.

Excludes is a list of regular expressions. With excludes in place, a resource is selected for inclusion if and only if its path matches some of the include patterns and it does not match any of the exclude patterns.

**What problem is this solving:**
Say I have many resources in my source tree and I want to include some subdirectories except for some filetypes. With excludes in place, I could pass something like

```
-H:IncludeResources=foo/.* -H:IncludeResources=bar/.* -H:IncludeResources=baz/.* -H:ExcludeResources=*.\.png -H:ExcludeResources=*.\.jpg -H:ExcludeResources=*.\.gif
```

Without excludes, I'd have to somehow hack everything into one expression using some negative lookaheads and alternations. I am not sure myself something like that is even possible.

**Describe the solution you'd like.**

I am primarily interested in getting `-H:ExcludeResources`. I have not thought of how excludes could be supported by `-H:ResourceConfigurationFiles`.

GraalVM users, and developers of libraries and frameworks which depend on GraalVM

Have a tool that scans the source tree, evaluates the includes and excludes and passes the selected paths one by one to `native-image`. We actually have something like that in Camel Quarkus: https://github.com/apache/camel-quarkus/pull/960/files but I believe a generic option like that belongs to `native-image` itself.

**Express whether you'd like to help contributing this feature**
I can try if there is no better candidate.

pgillet:
See https://github.com/oracle/graal/issues/2347 .

### DIFF
--- a/substratevm/Resources.md
+++ b/substratevm/Resources.md
@@ -5,21 +5,28 @@ To make calls such as `Class.getResource()` or `Class.getResourceAsStream()` (or
 
 ```json
 {
-  "resources": [
-    {"pattern": "<Java regexp that matches resource(s) to be included in the image>"},
-    {"pattern": "<another regexp>"},
-    ...
-  ]
+  "resources": {
+    "includes": [
+      {"pattern": "<Java regexp that matches resource(s) to be included in the image>"},
+      {"pattern": "<another regexp>"},
+      ...
+    ],
+    "excludes": [
+      {"pattern": "<Java regexp that matches resource(s) to be excluded from the image>"},
+      {"pattern": "<another regexp>"},
+      ...
+    ]
+  } 
 }
 ```
 
-The configuration file's path must be provided to `native-image` with `-H:ResourceConfigurationFiles=/path/to/resource-config.json`. Alternatively, individual resource paths can also be specified directly to `native-image`:
-```shell
+The configuration file's path must be provided to the native image builder with `-H:ResourceConfigurationFiles=/path/to/resource-config.json`. Alternatively, individual resource paths can also be specified directly to `native-image`:
+```
 native-image -H:IncludeResources=<Java regexp that matches resources to be included in the image> ...
 ```
-The `-H:IncludeResources` option can be passed several times to define more than one regexp to match resources.
+The `-H:IncludeResources` and `-H:ExcludeResources` options can be passed several times to define more than one regexp to match or exclude resources, respectively.
 
-To see which resources get included into the image, you can enable the related logging info with `-H:Log=registerResource:`.
+To see which resources get ultimately included into the image, you can enable the related logging info with `-H:Log=registerResource:`.
 
 ### Example Usage
 
@@ -41,6 +48,7 @@ Then:
 *  `Resource0.txt` can be loaded with `.*/Resource0.txt$`.
 *  `Resource0.txt` and `Resource1.txt` can be loaded with `.*/Resource0.txt$` and `.*/Resource1.txt$`
    (or alternatively with a single `.*/(Resource0|Resource1).txt$`).
+*  Also, if we want to include everything except the `Resource2.txt` file, we can simply exclude it with `-H:IncludeResources='.*/Resource.*txt$'` followed by `-H:ExcludeResources='.*/Resource2.txt$'`.
 
 See also the [guide on assisted configuration of Java resources and other dynamic features](Configuration.md#assisted-configuration-of-native-image-builds).
 
--- a/substratevm/src/com.oracle.svm.configure.test/src/com/oracle/svm/configure/test/config/ResourceConfigurationTest.java
+++ b/substratevm/src/com.oracle.svm.configure.test/src/com/oracle/svm/configure/test/config/ResourceConfigurationTest.java
@@ -1,0 +1,98 @@
+package com.oracle.svm.configure.test.config;
+
+import java.io.IOException;
+import java.io.PipedReader;
+import java.io.PipedWriter;
+import java.util.LinkedList;
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.oracle.svm.configure.config.ResourceConfiguration;
+import com.oracle.svm.configure.json.JsonWriter;
+import com.oracle.svm.core.configure.ResourceConfigurationParser;
+import com.oracle.svm.core.configure.ResourcesRegistry;
+
+public class ResourceConfigurationTest {
+
+    @Test
+    public void anyResourceMatches() {
+        ResourceConfiguration rc = new ResourceConfiguration();
+        rc.addResourcePattern(".*/Resource.*txt$");
+
+        Assert.assertTrue(rc.anyResourceMatches("com/my/app/Resource0.txt"));
+        Assert.assertTrue(rc.anyResourceMatches("com/my/app/Resource1.txt"));
+        Assert.assertTrue(rc.anyResourceMatches("/Resource2.txt"));
+        Assert.assertTrue(rc.anyResourceMatches("/Resource3.txt"));
+
+        rc.ignoreResourcePattern(".*/Resource2.txt$");
+
+        Assert.assertTrue(rc.anyResourceMatches("com/my/app/Resource0.txt"));
+        Assert.assertTrue(rc.anyResourceMatches("com/my/app/Resource1.txt"));
+        Assert.assertFalse(rc.anyResourceMatches("/Resource2.txt"));
+        Assert.assertTrue(rc.anyResourceMatches("/Resource3.txt"));
+    }
+
+    @Test
+    public void printJson() {
+        ResourceConfiguration rc = new ResourceConfiguration();
+        rc.addResourcePattern(".*/Resource.*txt$");
+        rc.ignoreResourcePattern(".*/Resource2.txt$");
+        PipedWriter pw = new PipedWriter();
+        JsonWriter jw = new JsonWriter(pw);
+
+        try (PipedReader pr = new PipedReader()) {
+            pr.connect(pw);
+
+            Thread writerThread = new Thread(new Runnable() {
+
+                @Override
+                public void run() {
+                    try {
+                        rc.printJson(jw);
+                    } catch (IOException e) {
+                        Assert.fail(e.getMessage());
+                    } finally {
+                        try {
+                            jw.close();
+                        } catch (IOException e) {
+                        }
+                    }
+                }
+            });
+
+            List<String> addedResources = new LinkedList<>();
+            List<String> ignoredResources = new LinkedList<>();
+
+            ResourcesRegistry registry = new ResourcesRegistry() {
+
+                @Override
+                public void addResources(String pattern) {
+                    addedResources.add(pattern);
+                }
+
+                @Override
+                public void ignoreResources(String pattern) {
+                    ignoredResources.add(pattern);
+                }
+
+                @Override
+                public void addResourceBundles(String name) {
+                }
+            };
+
+            ResourceConfigurationParser rcp = new ResourceConfigurationParser(registry);
+            writerThread.start();
+            rcp.parseAndRegister(pr);
+
+            writerThread.join();
+
+            Assert.assertTrue(addedResources.contains(".*/Resource.*txt$"));
+            Assert.assertTrue(ignoredResources.contains(".*/Resource2.txt$"));
+        } catch (IOException | InterruptedException e) {
+            e.printStackTrace();
+            Assert.fail(e.getMessage());
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


substratevm/Resources.md
```diff
@@ -5,21 +5,28 @@ To make calls such as `Class.getResource()` or `Class.getResourceAsStream()` (or
...
 }
 ```
 
 The configuration file's path must be provided to the native image builder with `-H:ResourceConfigurationFiles=/path/to/resource-config.json`. Alternatively, individual resource paths can also be specified directly to `native-image`:
 ```
-native-image -H:IncludeResources=<Java regexp that matches resources to be included in the image> ...
+native-image -H:IncludeResources=<Java regexp that matches resources to be included in the image> ... -H:ExcludeResources=<Java regexp that matches resources to be excluded from the image> ...
 ```
-The `-H:IncludeResources` option can be passed several times to define more than one regexp to match resources.
+The `-H:IncludeResources` and `-H:IncludeResources` options can be passed several times to define more than one regexp to match or exclude resources, respectively.
```
olpaw, line 27: and `-H:ExcludeResources` ?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ResourceConfigurationParser.java
```diff
@@ -48,21 +48,40 @@ public void parseAndRegister(Reader reader) throws IOException {
...
                 throw new JSONParserException("Unknown attribute '" + pair.getKey() + "' (supported attributes: name) in resource definition");
             }
         }
         if (resourcesObject != null) {
-            List<Object> resources = asList(resourcesObject, "Attribute 'resources' must be a list of resources");
-            for (Object object : resources) {
-                parseEntry(object, "pattern", registry::addResources, "resource descriptor object", "'resources' list");
+            Object includesObject = null;
+            Object excludesObject = null;
+            for (Map.Entry<String, Object> pair : resourcesObject.entrySet()) {
```
olpaw, line 65: That's a breaking change. We can't do that. Please also support old-style `"includes"`.

substratevm/Resources.md
pgillet: Oops!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ResourceConfigurationParser.java
pgillet: Yes, you are right. I changed the code to ensure backward compatibility.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ResourceConfigurationParser.java
```diff
@@ -60,9 +60,35 @@ private void parseTopLevelObject(Map<String, Object> obj) {
             }
         }
         if (resourcesObject != null) {
-            List<Object> resources = asList(resourcesObject, "Attribute 'resources' must be a list of resources");
-            for (Object object : resources) {
-                parseEntry(object, "pattern", registry::addResources, "resource descriptor object", "'resources' list");
+            if (resourcesObject instanceof Map) { // New format
+                Object includesObject = null;
+                Object excludesObject = null;
+                for (Map.Entry<String, Object> pair : ((Map<String, Object>) resourcesObject).entrySet()) {
```
olpaw, line 66: If you use `(((Map<String, Object>) resourcesObject).entrySet()).forEach((key, value) -> {` you avoid the `Map.Entry<String, Object>`-type orgy.

olpaw: Nah doesn't work because
```
    Object includesObject = null;
    Object excludesObject = null;
```
cannot be assigned to from the lambda.

olpaw: How about
```patch
--- substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ResourceConfigurationParser.java	(revision d2a6f30c05cb24f802e1d828f2eb03f7124dab8e)
+++ substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ResourceConfigurationParser.java	(date 1602691803931)
@@ -61,17 +61,9 @@
         }
         if (resourcesObject != null) {
             if (resourcesObject instanceof Map) { // New format
-                Object includesObject = null;
-                Object excludesObject = null;
-                for (Map.Entry<String, Object> pair : ((Map<String, Object>) resourcesObject).entrySet()) {
-                    if ("includes".equals(pair.getKey())) {
-                        includesObject = pair.getValue();
-                    } else if ("excludes".equals(pair.getKey())) {
-                        excludesObject = pair.getValue();
-                    } else {
-                        throw new JSONParserException("Unknown attribute '" + pair.getKey() + "' (supported attributes: name) in resource definition");
-                    }
-                }
+                Map<String, Object> map = (Map<String, Object>) resourcesObject;
+                Object includesObject = map.get("includes");
+                Object excludesObject = map.get("excludes");
 
                 List<Object> includes = asList(includesObject, "Attribute 'includes' must be a list of resources");
                 for (Object object : includes) {
```

pgillet: Yes, it's definitely neater but in this case, we do not check for extra unknown attributes anymore, like it is done just above:

```
for (Map.Entry<String, Object> pair : obj.entrySet()) {
            if ("resources".equals(pair.getKey())) {
                resourcesObject = pair.getValue();
            } else if ("bundles".equals(pair.getKey())) {
                bundlesObject = pair.getValue();
            } else {
                throw new JSONParserException("Unknown attribute '" + pair.getKey() + "' (supported attributes: name) in resource definition");
            }
        }
```

What do you think? I can also change both parts with your solution...

olpaw: > Yes, it's definitely neater but in this case, we do not check for extra unknown attributes anymore

Rrrrright. We do not want to break this strict checking behaviour for saving a few lines of code.
Sorry for the noise.

olpaw (approved):

substratevm/src/com.oracle.svm.configure.test/src/com/oracle/svm/configure/test/config/ResourceConfigurationTest.java
```diff
@@ -0,0 +1,98 @@
+package com.oracle.svm.configure.test.config;
```
olpaw, line 1: I ran your PR in our CI. 

It complains that this file is missing the license header.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ResourceConfigurationParser.java
```diff
@@ -47,6 +47,7 @@ public void parseAndRegister(Reader reader) throws IOException {
         parseTopLevelObject(asMap(json, "first level of document must be an object"));
     }
 
+    @SuppressWarnings("unchecked")
```
olpaw, line 50: Let's no use this. Revert to your original from where you had the cast to `Map<String, Object>`. While it is ugly its still better than suppress unchecked-warnings in the whole method body.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
```diff
@@ -72,12 +72,16 @@
 public final class ResourcesFeature implements Feature {
 
     public static class Options {
-        @Option(help = "Regexp to match names of resources to be included in the image.", type = OptionType.User)//
+        @Option(help = "Regexp to match names of resources to be included in the image.", type = OptionType.User)
```
olpaw, line 75: Please re-add the `//` at the end. Otherwise `mx eclipseformat` wants to put the annotation in the same line as the field.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
```diff
@@ -72,12 +72,16 @@
...
         public static final HostedOptionKey<String[]> IncludeResources = new HostedOptionKey<>(new String[0]);
+
+        @Option(help = "Regexp to match names of resources to be excluded from the image.", type = OptionType.User)
+        public static final HostedOptionKey<String[]> ExcludeResources = new HostedOptionKey<>(new String[0]);
     }
 
     private boolean sealed = false;
     private Set<String> newResources = Collections.newSetFromMap(new ConcurrentHashMap<>());
+    private Set<String> ignoredResources = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
olpaw, line 84: Same as https://github.com/oracle/graal/pull/2912/files#r506179629

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
```diff
@@ -72,12 +72,16 @@
 public final class ResourcesFeature implements Feature {
 
     public static class Options {
-        @Option(help = "Regexp to match names of resources to be included in the image.", type = OptionType.User)//
+        @Option(help = "Regexp to match names of resources to be included in the image.", type = OptionType.User)
         public static final HostedOptionKey<String[]> IncludeResources = new HostedOptionKey<>(new String[0]);
+
+        @Option(help = "Regexp to match names of resources to be excluded from the image.", type = OptionType.User)
+        public static final HostedOptionKey<String[]> ExcludeResources = new HostedOptionKey<>(new String[0]);
```
olpaw, line 79: Same as https://github.com/oracle/graal/pull/2912/files#r506179629

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
```diff
@@ -170,6 +178,14 @@ public void duringAnalysis(DuringAnalysisAccess access) {
         newResources.clear();
     }
 
+    private static Pattern[] compilePatterns(Set<String> patterns) {
```
olpaw, line 181: Another `mx eclipsformat` issue:
```patch
@@ -180,10 +180,10 @@
 
     private static Pattern[] compilePatterns(Set<String> patterns) {
         return patterns.stream()
-                .filter(s -> s.length() > 0)
-                .map(Pattern::compile)
-                .collect(Collectors.toList())
-                .toArray(new Pattern[]{});
+                        .filter(s -> s.length() > 0)
+                        .map(Pattern::compile)
+                        .collect(Collectors.toList())
+                        .toArray(new Pattern[]{});
     }
 
     @Override
```

olpaw (changes requested):
Our CI found a few minor issues that need fixing before I can merge your PR.
(You can run `mx eclipsformat` locally to verify you got the code formatting right)

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
pgillet: Re-added //

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
pgillet: Added //

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
pgillet: I am really not sure of the formatting. I use IntelliJ and I configured it to use the Eclipse code formatter as said in `compiler/docs/IDEs.md`...

olpaw (approved):

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ResourceConfigurationParser.java
pgillet: The CI seems to fail now because of the unchecked cast which is considered an error. That's why I put the annotation in the first place.

olpaw: Yea. I figured. Re-added the SuppressWarnings and fixed the merge conflict.

Because I fixed the merge conflict the PR did not close automatically when the changes got merged on master.
See: https://github.com/oracle/graal/commit/c1ccd5a2436a02f3d61a22bfeb0dbf5c3bd4d8be

olpaw:
Already on master: https://github.com/oracle/graal/commit/c1ccd5a2436a02f3d61a22bfeb0dbf5c3bd4d8be

PR #2918, opened 2020-10-13
[vscode] Add option to run GraalLS in debug mode

fniephaus:
and make connection process more robust.

Also, avoid failing connection if GraalLS process produces stderr output, forward errors to extension's `console` instead. Occasionally, the process would print the following to stderr which would always fail the connection:

> org.graalvm.shadowed.org.jline.utils.Log logr
> WARNING: Unable to create a system terminal, creating a dumb terminal

Please assign to @dbalek.

### DIFF
--- a/vscode/graalvm/package.json
+++ b/vscode/graalvm/package.json
@@ -55,6 +55,11 @@
 					"default": true,
 					"description": "Start GraalVM Language Server within processes being run or debugged"
 				},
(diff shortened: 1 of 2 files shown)


fniephaus:
The GraalVM extension `0.5.0` still fails sometimes because of this issue. The connection process should not be cancelled entirely if the process writes something to `stderr`.

eregon:
:+1: This would be quite useful

PR #2924, opened 2020-10-16
[lsp] Add support for LSPExtensions

fniephaus:
This adds the ability to extend the list of commands supported by the GraalLS through other instruments that provide the new `LSPExtension` service. See `LSPExtensionTestInstance` for an example.

Based on this change and #2906, we were able to implement Live Programming features on top of GraalLS and GraalVM's extension for VS Code.

Please assign to @dbalek and discuss! /cc @MartinBalin and @chumer

### DIFF
--- a/tools/src/org.graalvm.tools.api.lsp/src/org/graalvm/tools/api/lsp/LSPCommand.java
+++ b/tools/src/org.graalvm.tools.api.lsp/src/org/graalvm/tools/api/lsp/LSPCommand.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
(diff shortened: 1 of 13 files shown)


fniephaus:
It may be possible to move the two existing `GET_COVERAGE` and `DRY_RUN` commands into separate instruments as well. This may give us additional insights into what LSP extensions need to access. Please advise.

tools/src/org.graalvm.tools.api.lsp/src/org/graalvm/tools/api/lsp/LSPCommand.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Get the name of the command.
+     */
+    String getName();
+
+    /**
+     * Execute the command.
+     */
+    Object execute(LSPServerAccessor server, Env env, List<Object> arguments);
```
entlicher, line 44: With `Env` this is quite powerful. This allows an arbitrary instrumentation and can potentially be abused. How it's used? Would some restricted access be sufficient?

fniephaus: The `Env` object passed in is the `envInternal` and I exposed it because it *is* quite powerful. The idea of `LSPExtension`s is that they can parse, evaluate, and instrument code within the context of the GraalLS. Of course, we could hide some of `Env`'s functionality similar to how the `LSPServerAccessor` hides information. For our tools, we currently only need `#getInstrumenter()`, `#parse()`, and `#parseInline()`.

BTW: I wonder, can't other instruments just create `Context` objects through the Graal SDK and do arbitrary instrumentation? I mean all instruments must be trusted, right?

PR #2932, opened 2020-10-20
Feature/cppvirtual: C++/LLVM virtual method resolving and virtual field access via interop calls

pichristoph:
Enables resolving of virtual methods via Tuffle's InteropLibrary, and also accessing fields that are virtually inherited (C++). 
Also created for internal discussion with Sulong team.


sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/interop/access/LLVMInteropType.java
```diff
@@ -313,27 +391,124 @@ public Method findMethodByArgumentsWithSelf(String memberName, Object[] argument
...
+        @ExportMessage
+        Object readArrayElement(long index) throws UnsupportedMessageException {
+            try {
+                final String methodName = vtable.findMethod(index).getName();
+                try {
+                    Object readMember = InteropLibrary.getUncached(foreign).readMember(foreign, methodName);
+                    return new RemoveSelfArgument(readMember);
+                } catch (UnknownIdentifierException e) {
+                    final String msg = String.format("External method %s (identifier \"%s\") not found in type %s", methodName, e.getUnknownIdentifier(), foreign.getClass().getSimpleName());
+                    throw new IllegalStateException(msg);
```
pichristoph, line 467: Here, I am not sure whether this solution is correct, because from an interop view point, this is an UnknownIdentifierException (=i.e. the foreign object, e.g. JavaScript, does not have a specific field/method). However, from a C++ side, it is an array access by index.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/interop/access/LLVMInteropVtableAccessNode.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public static LLVMInteropVtableAccessNode create() {
+        return LLVMInteropVtableAccessNodeGen.create();
+    }
+
+    @Specialization
+    Object doPointer(LLVMPointer vtablePointer, long virtualIndex, Object[] arguments, @CachedLibrary(limit = "5") InteropLibrary interop)
+                    throws UnsupportedTypeException, ArityException, UnsupportedMessageException {
+
+        LLVMPointer vtableElementPointer = vtablePointer.increment(virtualIndex * 8);
```
pichristoph, line 55: Here, maybe there is a fast operation to get the pointer size (8 bytes) appropriately, if the '8' should not be hard-coded here.

sulong/tests/com.oracle.truffle.llvm.tests.interop.native/interop/vtableTest1.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+//-----------------------------------------test via polyglot API
+
+class A {
+public:
+    int k;
+    virtual int foo(int x);
+};
+
+int A::foo(int x) {
```
pichristoph, line 41: maybe dummy method can be removed if marking as abstract (`virtual int foo(int x)=0;`) works with gate tests.

sulong/tests/com.oracle.truffle.llvm.tests.interop.native/interop/vtableTest1.cpp
```diff
@@ -0,0 +1,89 @@
...
+    return 0 * x;
+} //dummy
```
pichristoph, line 43: Maybe there is a way to remove this method and make its declaration abstract (`virtual int foo(int x)=0`) if it works together with compilation of gate tests.

PR #2935, opened 2020-10-21
Exclude substituted class from build time initialization

ziyilin:
The class substituted with `@Substitute` or `@Delete` annotation should not be initialized at build time because it will not appear at run time. It makes no sense to put it in the native heap but causing initialization errors.

This patch fixes the issue https://github.com/oracle/graal/issues/2936

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/TypeInitializerGraph.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/TypeInitializerGraph.java
@@ -24,6 +24,7 @@
  */
 package com.oracle.svm.hosted.classinitialization;
 
+import java.lang.annotation.Annotation;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/TypeInitializerGraph.java
```diff
@@ -216,7 +221,18 @@ private boolean isInvokeUnsafeIterative(InvokeTypeFlow i) {
     }
 
     private void addInitializer(AnalysisType t) {
-        types.put(t, initialTypeInitializerSafety(t));
+        ResolvedJavaType rt = t.getWrappedWithoutResolve();
```
vjovanov, line 224: I would put this whole block in `initialTypeInitializerSafety`.

vjovanov:
Otherwise, LGTM.

ziyilin:
> Otherwise, LGTM.

What is LGTM short for?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/TypeInitializerGraph.java
ziyilin: fixed

zakkak:
> What is LGTM short for?

LGTM = Looks Good To Me

vjovanov:
This issue has been integrated.

PR #2945, opened 2020-10-27
Work around the absence of SpeculationReasonEncoding in OpenJDK11

### LINKED ISSUE #2933: "cannot find symbol SpeculationReasonEncoding" when building with vanilla OpenJDK
**Describe the issue**
As discussed in https://github.com/graalvm/mandrel/issues/146 after @2d5eee0 GraalVM no longer builds with vanilla OpenJDK.

> It looks like oracle@2d5eee0 depends on https://bugs.openjdk.java.net/browse/JDK-8220623 which got backported to labsJDK11 in graalvm/labs-openjdk-11@fbbdf5f

Unfortunately https://bugs.openjdk.java.net/browse/JDK-8220623 is rather big and includes many changes that are not all related to the issue at hand. I tried creating a minimal backport/patch for jdk11u by pulling in only the needed changes but I ended up pulling more and more dependencies (especially after entering the C++ code).  The main issue seems to be the way that failed speculations are gathered after https://bugs.openjdk.java.net/browse/JDK-8220623. I also tried pulling in only the necessary changes without changing the way that failed speculations are gathered, but then the `SpeculationReasonEncoding` is not actually used making this kind of patch kind of pointless.

So I am openning this issue to ask for directions on how to approach this and discuss possible alternatives.

Thanks! 

```
export JAVA_HOME=/tmp/openjdk11ea
curl -sL https://api.adoptopenjdk.net/v3/binary/latest/11/ea/linux/x64/jdk/hotspot/normal/openjdk -o jdk.tar.gz
curl -sL https://api.adoptopenjdk.net/v3/binary/latest/11/ea/linux/x64/staticlibs/hotspot/normal/openjdk -o jdk-static-libs.tar.gz
mkdir -p ${JAVA_HOME}
tar xf jdk.tar.gz -C ${JAVA_HOME} --strip-components=1
tar xf jdk-static-libs.tar.gz -C ${JAVA_HOME} --strip-components=1
git clone git@github.com:oracle/graal.git --depth 1
cd graal
mx --primary-suite-path substratevm build --projects=com.oracle.svm.graal
```

 - GraalVM version: **master branch (60faf95321609669bcf0a01e0e134c3b7906e435)**
 - JDK major version: **11**
 - OS: **Linux Fedora 32**
 - Architecture: **AMD64**

**More details**
```
$ mx --primary-suite-path substratevm build --projects=com.oracle.svm.graal         
Compiling com.oracle.svm.graal with javac-daemon(JDK 11)... [dependencies were added, removed or re-ordered]
/home/zakkak/code/graal/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationReason.java:41: error: cannot find symbol
    public SpeculationLog.SpeculationReasonEncoding encode(Supplier<SpeculationLog.SpeculationReasonEncoding> encodingSupplier) {
                                                                                  ^
  symbol:   class SpeculationReasonEncoding
  location: interface SpeculationLog
/home/zakkak/code/graal/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationReason.java:41: error: cannot find symbol
    public SpeculationLog.SpeculationReasonEncoding encode(Supplier<SpeculationLog.SpeculationReasonEncoding> encodingSupplier) {
                         ^
  symbol:   class SpeculationReasonEncoding
  location: interface SpeculationLog
/home/zakkak/code/graal/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationReasonEncoding.java:40: error: cannot find symbol
final class IsolatedSpeculationReasonEncoding extends ByteArrayOutputStream implements SpeculationLog.SpeculationReasonEncoding {
                                                                                                     ^
  symbol:   class SpeculationReasonEncoding
  location: interface SpeculationLog
3 errors

Compiling com.oracle.svm.graal with javac-daemon(JDK 11) failed
1 build tasks failed
```

zakkak:
Fixes #2933

Note that this PR fixes only the specific issue discussed in #2933, which is necessary to create a Mandrel build based on graalvm's master.  Even after merging this PR parts of GraalVM will fail to build using OpenJDK 11, for instance `mx --primary-suite-path substratevm --components='Native Image' build` will fail.

### DIFF
--- a/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedCodeInstallBridge.java
+++ b/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedCodeInstallBridge.java
@@ -60,16 +60,6 @@ public final class IsolatedCodeInstallBridge extends InstalledCode implements Op
                     " only acts as an accessor for cross-isolate data. None of the implemented methods may be called.";
 
     @Override
-    public long getAddress() {
-        throw VMError.shouldNotReachHere(DO_NOT_CALL_REASON);
-    }
-
-    @Override
-    public long getEntryPoint() {
-        throw VMError.shouldNotReachHere(DO_NOT_CALL_REASON);
-    }
-
-    @Override
     public String getName() {
         throw VMError.shouldNotReachHere(DO_NOT_CALL_REASON);
     }
--- a/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
+++ b/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
@@ -24,8 +24,13 @@
  */
 package com.oracle.svm.graal.isolated;
 
+import java.lang.reflect.Field;
 import java.nio.ByteBuffer;
 
+import jdk.vm.ci.meta.ResolvedJavaField;
+import jdk.vm.ci.meta.ResolvedJavaMethod;
+import jdk.vm.ci.meta.ResolvedJavaType;
+import org.graalvm.compiler.serviceprovider.UnencodedSpeculationReason;
 import org.graalvm.nativeimage.PinnedObject;
 import org.graalvm.nativeimage.c.function.CEntryPoint;
 import org.graalvm.nativeimage.c.type.CTypeConversion;
@@ -48,10 +53,87 @@ public final class IsolatedSpeculationLog extends IsolatedObjectProxy<Speculatio
         collectFailedSpeculations0(IsolatedCompileContext.get().getClient(), handle);
     }
 
+    private static byte[] encodeAsByteArray(SpeculationReason reason) {
+        int groupId = 0;
+        Object[] context = null;
+        try {
+            final Field groupIdField;
+            final Field contextField;
+            /*
+             * The following reflective accesses do not add runtime overhead as they are being
+             * converted to direct accesses during image building. Maintaining two separate paths
+             * that invoke {@code getDeclaredField} ensures that the analysis will be able to
+             * optimize the reflective accesses and convert them to direct ones.
+             */
+            if (reason instanceof UnencodedSpeculationReason) {
+                final Class<?> klass = UnencodedSpeculationReason.class;
+                groupIdField = klass.getDeclaredField("groupId");
+                contextField = klass.getDeclaredField("context");
+            } else {
+                final Class<?> klass = Class.forName("jdk.vm.ci.meta.EncodedSpeculationReason");
+                groupIdField = klass.getDeclaredField("groupId");
+                contextField = klass.getDeclaredField("context");
+            }
+            groupIdField.setAccessible(true);
+            groupId = groupIdField.getInt(reason);
+            contextField.setAccessible(true);
+            context = (Object[]) contextField.get(reason);
+        } catch (IllegalAccessException | NoSuchFieldException | ClassNotFoundException e) {
+            VMError.shouldNotReachHere("Failed to encode speculation reason", e);
+        }
+        IsolatedSpeculationReasonEncoding encoding = encode(groupId, context);
+        return encoding.getByteArray();
+    }
+
+    /** Adapted from {@code jdk.vm.ci.meta.EncodedSpeculationReason#encode}. */
+    private static IsolatedSpeculationReasonEncoding encode(int groupId, Object[] context) {
+        IsolatedSpeculationReasonEncoding encoding = new IsolatedSpeculationReasonEncoding();
+        encoding.addInt(groupId);
+        for (Object o : context) {
+            if (o == null) {
+                encoding.addInt(0);
+            } else {
+                addNonNullObject(encoding, o);
+            }
+        }
+        return encoding;
+    }
+
+    /** Copied from {@code jdk.vm.ci.meta.EncodedSpeculationReason#addNonNullObject}. */
+    private static void addNonNullObject(IsolatedSpeculationReasonEncoding encoding, Object o) {
+        Class<?> c = o.getClass();
+        if (c == String.class) {
+            encoding.addString((String) o);
+        } else if (c == Byte.class) {
+            encoding.addByte((Byte) o);
+        } else if (c == Short.class) {
+            encoding.addShort((Short) o);
+        } else if (c == Character.class) {
+            encoding.addShort((Character) o);
+        } else if (c == Integer.class) {
+            encoding.addInt((Integer) o);
+        } else if (c == Long.class) {
+            encoding.addLong((Long) o);
+        } else if (c == Float.class) {
+            encoding.addInt(Float.floatToRawIntBits((Float) o));
+        } else if (c == Double.class) {
+            encoding.addLong(Double.doubleToRawLongBits((Double) o));
+        } else if (o instanceof Enum) {
+            encoding.addInt(((Enum<?>) o).ordinal());
+        } else if (o instanceof ResolvedJavaMethod) {
+            encoding.addMethod((ResolvedJavaMethod) o);
+        } else if (o instanceof ResolvedJavaType) {
+            encoding.addType((ResolvedJavaType) o);
+        } else if (o instanceof ResolvedJavaField) {
+            encoding.addField((ResolvedJavaField) o);
+        } else {
+            throw new IllegalArgumentException("Unsupported type for encoding: " + c.getName());
+        }
+    }
+
     @Override
     public boolean maySpeculate(SpeculationReason reason) {
-        SpeculationReasonEncoding encoding = reason.encode(IsolatedSpeculationReasonEncoding::new);
-        byte[] bytes = ((IsolatedSpeculationReasonEncoding) encoding).getByteArray();
+        byte[] bytes = encodeAsByteArray(reason);
         try (PinnedObject pinnedBytes = PinnedObject.create(bytes)) {
             return maySpeculate0(IsolatedCompileContext.get().getClient(), handle, pinnedBytes.addressOfArrayElement(0), bytes.length);
         }
@@ -59,8 +141,7 @@ public final class IsolatedSpeculationLog extends IsolatedObjectProxy<Speculatio
 
     @Override
     public Speculation speculate(SpeculationReason reason) {
-        SpeculationReasonEncoding encoding = reason.encode(IsolatedSpeculationReasonEncoding::new);
-        byte[] bytes = ((IsolatedSpeculationReasonEncoding) encoding).getByteArray();
+        byte[] bytes = encodeAsByteArray(reason);
         ClientHandle<SpeculationReason> reasonHandle;
         try (PinnedObject pinnedBytes = PinnedObject.create(bytes)) {
             reasonHandle = speculate0(IsolatedCompileContext.get().getClient(), handle, pinnedBytes.addressOfArrayElement(0), bytes.length);
--- a/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/SpeculationReasonEncodingClone.java
+++ b/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/SpeculationReasonEncodingClone.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -24,12 +24,29 @@
  */
 package com.oracle.svm.graal.isolated;
 
-import jdk.vm.ci.meta.SpeculationLog.SpeculationReason;
+import jdk.vm.ci.meta.ResolvedJavaField;
+import jdk.vm.ci.meta.ResolvedJavaMethod;
+import jdk.vm.ci.meta.ResolvedJavaType;
 
-public final class IsolatedSpeculationReason extends IsolatedObjectProxy<SpeculationReason> implements SpeculationReason {
+/** Copied from {@code jdk.vm.ci.hotspot.HotSpotSpeculationEncoding.SpeculationReasonEncoding}. */
+public interface SpeculationReasonEncodingClone {
+    void addByte(int value);
 
-    IsolatedSpeculationReason(ClientHandle<SpeculationReason> reason) {
-        super(reason);
+    void addShort(int value);
+
+    void addInt(int value);
+
+    void addLong(long value);
+
+    void addMethod(ResolvedJavaMethod method);
+
+    void addType(ResolvedJavaType type);
+
+    void addString(String value);
+
+    default void addField(ResolvedJavaField field) {
+        this.addType(field.getDeclaringClass());
+        this.addInt(field.getModifiers());
+        this.addInt(field.getOffset());
     }
-
 }
--- a/compiler/src/org.graalvm.compiler.serviceprovider.jdk8/src/org/graalvm/compiler/serviceprovider/GraalServices.java
+++ b/compiler/src/org.graalvm.compiler.serviceprovider.jdk8/src/org/graalvm/compiler/serviceprovider/GraalServices.java
@@ -33,12 +33,10 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.math.BigDecimal;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ServiceConfigurationError;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -47,11 +45,9 @@ import jdk.vm.ci.code.VirtualObject;
 import jdk.vm.ci.code.site.Infopoint;
 import jdk.vm.ci.code.site.InfopointReason;
 import jdk.vm.ci.meta.ConstantPool;
+import jdk.vm.ci.meta.EncodedSpeculationReason;
 import jdk.vm.ci.meta.JavaType;
-import jdk.vm.ci.meta.ResolvedJavaField;
-import jdk.vm.ci.meta.ResolvedJavaMethod;
 import jdk.vm.ci.meta.ResolvedJavaType;
-import jdk.vm.ci.meta.SpeculationLog;
 import jdk.vm.ci.meta.SpeculationLog.SpeculationReason;
 import jdk.vm.ci.services.JVMCIPermission;
 import jdk.vm.ci.services.Services;
@@ -133,91 +129,6 @@ public final class GraalServices {
     public static boolean isToStringTrusted(Class<?> c) {
         ClassLoader cl = c.getClassLoader();
         return cl == null || cl == JVMCI_LOADER || cl == JVMCI_PARENT_LOADER;
-    }
-
-    /**
-     * An implementation of {@link SpeculationReason} based on encoded values.
-     */
-    public static class EncodedSpeculationReason implements SpeculationReason {
-        final int groupId;
-        final String groupName;
-        final Object[] context;
-        private SpeculationLog.SpeculationReasonEncoding encoding;
-
-        public EncodedSpeculationReason(int groupId, String groupName, Object[] context) {
-            this.groupId = groupId;
-            this.groupName = groupName;
-            this.context = context;
-        }
-
-        @Override
-        public boolean equals(Object obj) {
-            if (obj instanceof EncodedSpeculationReason) {
-                if (obj instanceof EncodedSpeculationReason) {
-                    EncodedSpeculationReason that = (EncodedSpeculationReason) obj;
-                    return this.groupId == that.groupId && Arrays.equals(this.context, that.context);
-                }
-                return false;
-            }
-            return false;
-        }
-
-        @Override
-        public SpeculationLog.SpeculationReasonEncoding encode(Supplier<SpeculationLog.SpeculationReasonEncoding> encodingSupplier) {
-            if (encoding == null) {
-                encoding = encodingSupplier.get();
-                encoding.addInt(groupId);
-                for (Object o : context) {
-                    if (o == null) {
-                        encoding.addInt(0);
-                    } else {
-                        addNonNullObject(encoding, o);
-                    }
-                }
-            }
-            return encoding;
-        }
-
-        static void addNonNullObject(SpeculationLog.SpeculationReasonEncoding encoding, Object o) {
-            Class<? extends Object> c = o.getClass();
-            if (c == String.class) {
-                encoding.addString((String) o);
-            } else if (c == Byte.class) {
-                encoding.addByte((Byte) o);
-            } else if (c == Short.class) {
-                encoding.addShort((Short) o);
-            } else if (c == Character.class) {
-                encoding.addShort((Character) o);
-            } else if (c == Integer.class) {
-                encoding.addInt((Integer) o);
-            } else if (c == Long.class) {
-                encoding.addLong((Long) o);
-            } else if (c == Float.class) {
-                encoding.addInt(Float.floatToRawIntBits((Float) o));
-            } else if (c == Double.class) {
-                encoding.addLong(Double.doubleToRawLongBits((Double) o));
-            } else if (o instanceof Enum) {
-                encoding.addInt(((Enum<?>) o).ordinal());
-            } else if (o instanceof ResolvedJavaMethod) {
-                encoding.addMethod((ResolvedJavaMethod) o);
-            } else if (o instanceof ResolvedJavaType) {
-                encoding.addType((ResolvedJavaType) o);
-            } else if (o instanceof ResolvedJavaField) {
-                encoding.addField((ResolvedJavaField) o);
-            } else {
-                throw new IllegalArgumentException("Unsupported type for encoding: " + c.getName());
-            }
-        }
-
-        @Override
-        public int hashCode() {
-            return groupId + Arrays.hashCode(this.context);
-        }
-
-        @Override
-        public String toString() {
-            return String.format("%s@%d%s", groupName, groupId, Arrays.toString(context));
-        }
     }
 
     static SpeculationReason createSpeculationReason(int groupId, String groupName, Object... context) {
--- a/substratevm/mx.substratevm/suite.py
+++ b/substratevm/mx.substratevm/suite.py
@@ -686,6 +686,27 @@ suite = {
             ],
         },
 
+        "com.oracle.svm.graal.test": {
+            "subDir": "src",
+            "sourceDirs": ["src"],
+            "dependencies": [
+                "mx:JUNIT_TOOL",
+                "sdk:GRAAL_SDK",
+                "com.oracle.svm.graal",
+            ],
+            "requiresConcealed" : {
+                "jdk.internal.vm.ci": ["jdk.vm.ci.meta"],
+            },
+            "checkstyle": "com.oracle.svm.core",
+            "workingSets": "SVM",
+            "annotationProcessors": [
+                "compiler:GRAAL_PROCESSOR",
+            ],
+            "javaCompliance": "8+",
+            "spotbugs": "false",
+            "testProject": True,
+        },
+
         "com.oracle.svm.thirdparty": {
             "subDir": "src",
             "sourceDirs": ["src"],
@@ -1171,6 +1192,7 @@ suite = {
             "com.oracle.svm.test",
             "com.oracle.svm.test.jdk11",
             "com.oracle.svm.configure.test",
+            "com.oracle.svm.graal.test",
           ],
           "distDependencies": [
             "mx:JUNIT_TOOL",
--- a/substratevm/src/com.oracle.svm.graal.test/src/com/oracle/svm/graal/test/IsolatedSpeculationLogEncodingNativeTest.java
+++ b/substratevm/src/com.oracle.svm.graal.test/src/com/oracle/svm/graal/test/IsolatedSpeculationLogEncodingNativeTest.java
@@ -29,6 +29,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
+import org.graalvm.compiler.serviceprovider.GraalServices;
 import org.graalvm.compiler.serviceprovider.UnencodedSpeculationReason;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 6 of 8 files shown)


substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedCodeInstallBridge.java
```diff
@@ -59,16 +59,6 @@ public IsolatedCodeInstallBridge(ClientHandle<? extends SubstrateInstalledCode.A
     private static final String DO_NOT_CALL_REASON = IsolatedCodeInstallBridge.class.getSimpleName() +
                     " only acts as an accessor for cross-isolate data. None of the implemented methods may be called.";
 
-    @Override
-    public long getAddress() {
```
zakkak, line 63: This and `getEntryPoint` are `final` in vanilla openjdk11

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +54,63 @@ public void collectFailedSpeculations() {
...
+        SpeculationReasonEncodingSupplier(Object speculationReasonEncodingProxy) {
+            this.speculationReasonEncodingProxy = speculationReasonEncodingProxy;
+        }
+
+        @Override
+        public Object get() {
+            try {
+                return speculationReasonEncodingProxy.getClass().getDeclaredConstructor().newInstance();
+            } catch (InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+                e.printStackTrace();
```
zakkak, line 76: This is still an early version to get some feedback. Error checking will be improved once we agree on the core changes.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +54,63 @@ public void collectFailedSpeculations() {
...
+    private byte[] getByteArray(SpeculationReason reason) {
+        byte[] bytes = null;
+        if (reason instanceof UnencodedSpeculationReason) {
+            Object encoding = ((UnencodedSpeculationReason) reason).encode();
+            assert encoding instanceof IsolatedSpeculationReasonEncoding;
+            bytes = ((IsolatedSpeculationReasonEncoding) encoding).getByteArray();
+        } else {
+            try {
+                Class<?> speculationReasonEncodingClass = Class.forName("jdk.vm.ci.meta.SpeculationLog$SpeculationReasonEncoding");
+                final Object speculationReasonEncodingProxy = Proxy.newProxyInstance(speculationReasonEncodingClass.getClassLoader(),
```
zakkak, line 91: Proxy and reflection accessors will eventually be cached in static variables as requested in https://github.com/oracle/graal/issues/2933#issuecomment-714362574

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +54,63 @@ public void collectFailedSpeculations() {
...
+        private Object speculationReasonEncodingProxy;
+
+        SpeculationReasonEncodingSupplier(Object speculationReasonEncodingProxy) {
+            this.speculationReasonEncodingProxy = speculationReasonEncodingProxy;
+        }
+
+        @Override
+        public Object get() {
+            try {
+                return speculationReasonEncodingProxy.getClass().getDeclaredConstructor().newInstance();
```
zakkak, line 74: I am not sure about this. Please make sure to review thoroughly.

compiler/src/org.graalvm.compiler.serviceprovider/src/org/graalvm/compiler/serviceprovider/UnencodedSpeculationReason.java
```diff
@@ -60,4 +66,66 @@ public int hashCode() {
     public String toString() {
         return String.format("%s@%d%s", groupName, groupId, Arrays.toString(context));
     }
+
+    /** Adapted from {@code jdk.vm.ci.meta.EncodedSpeculationReason#encode}. */
+    public Object encode() {
+        if (encoding == null) {
+            try {
+                Class<?> c = Class.forName("com.oracle.svm.graal.isolated.IsolatedSpeculationReasonEncoding");
```
dougxc, line 74: SVM code should not be referenced in non-SVM projects.

peter-hofer (changes requested):
What I had in mind was to access `UnencodedSpeculationReason` fields `context` and `groupId` reflectively from the code in `com.oracle.svm.graal.isolated`, and to call the `SpeculationReasonEncodingClone` methods directly. This has the advantage that the native image build can reliably turn the reflective accesses into direct field accesses.
Alternatively, you could move `SpeculationReasonEncodingClone` to `org.graalvm.compiler.serviceprovider` as `UnencodedSpeculationReasonEncoding` and use it just the same as `SpeculationReasonEncoding` in `DirectSpeculationReason`. That just leaves introducing a proxy needed to support both types of `SpeculationReason` depending on the JVMCI version.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +58,115 @@ public void collectFailedSpeculations() {
...
+            } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException | ClassNotFoundException e) {
+                e.printStackTrace();
+            }
+        }
+        return bytes;
+    }
+
+    /** Adapted from {@code jdk.vm.ci.meta.EncodedSpeculationReason#encode}. */
+    public IsolatedSpeculationReasonEncoding encode(UnencodedSpeculationReason reason) {
+        IsolatedSpeculationReasonEncoding encoding = null;
```
zakkak, line 108: Should we cache the encodings here?

zakkak:
> What I had in mind was to access `UnencodedSpeculationReason` fields `context` and `groupId` reflectively from the code in `com.oracle.svm.graal.isolated`, and to call the `SpeculationReasonEncodingClone` methods directly. This has the advantage that the native image build can reliably turn the reflective accesses into direct field accesses.

I followed this approach which made things a bit simpler. Thanks. Does the rest (other than error checking and caching) look OK?

zakkak:
Hi @peter-hofer could you please let me know if this is moving in the right direction?

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
peter-hofer: Looks like we can only do that using a `WeakIdentityHashMap`, but I wouldn't now unless it proves to have a performance impact.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
peter-hofer: Please do cache the proxy class `Proxy.newProxyInstance(...).getClass()` in a `static final` field. Reflection accessors do not need to be cached as long as its arguments are compile-time constants.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +58,115 @@ public void collectFailedSpeculations() {
...
+        private final Object speculationReasonEncodingProxy;
+
+        SpeculationReasonEncodingSupplier(Object speculationReasonEncodingProxy) {
+            this.speculationReasonEncodingProxy = speculationReasonEncodingProxy;
+        }
+
+        @Override
+        public Object get() {
+            try {
+                return speculationReasonEncodingProxy.getClass().getDeclaredConstructor().newInstance();
```
peter-hofer, line 78: Probably best to just do `PROXY_CLASS.getDeclaredConstructor(...).newInstance()` here so the native image build can fold the reflective instantiation.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +58,115 @@ public void collectFailedSpeculations() {
         collectFailedSpeculations0(IsolatedCompileContext.get().getClient(), handle);
     }
 
+    static class SpeculationReasonInvocationHandler implements InvocationHandler {
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+            return IsolatedSpeculationReasonEncoding.class.getDeclaredMethod(method.getName(), method.getParameterTypes()).invoke(args);
```
peter-hofer, line 64: The `invoke` call is missing a receiver of type `IsolatedSpeculationReasonEncoding`, which should first be instantiated in the supplier and passed to this invocation handler class via a constructor.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +58,115 @@ public void collectFailedSpeculations() {
...
+            try {
+                return speculationReasonEncodingProxy.getClass().getDeclaredConstructor().newInstance();
+            } catch (InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+                e.printStackTrace();
+            }
+            return null;
+        }
+    }
+
+    private byte[] getByteArray(SpeculationReason reason) {
```
peter-hofer, line 86: Please name this method `encodeAsByteArray`.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
peter-hofer: I doubt the image build can optimize the reflective accesses, and on top of repeatedly doing lookups at runtime, we will need to register the methods of `IsolatedSpeculationReasonEncoding` as reflectively accessible in a `Feature` during the image build.
So I'd rather avoid this proxy, but I don't see how we can have a class that directly implements `SpeculationReasonEncoding` when that interface cannot be referenced in the code.

peter-hofer (changes requested):

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
zakkak: @peter-hofer is there any way to inject this interface to `SpeculationLog` when using an openJDK which lacks it (using something like `com.oracle.svm.core.annotate.TargetClass#onlyWith`)

Alternatively, could we add the interface in the sources and compile it in only when needed? (possibly through mx).

zakkak: > The invoke call is missing a receiver of type IsolatedSpeculationReasonEncoding, which should first be instantiated in the supplier and passed to this invocation handler class via a constructor.

Shouldn't the said `IsolatedSpeculationReasonEncoding` be instantiated for each new encoding and as a result don't we need a new `SpeculationReasonInvocationHandler` for each new encoding and consequently a new proxy instance?

peter-hofer: > Shouldn't the said `IsolatedSpeculationReasonEncoding` be instantiated for each new encoding and as a result don't we need a new `SpeculationReasonInvocationHandler` for each new encoding and consequently a new proxy instance rendering https://github.com/oracle/graal/pull/2945/files#r519816278 impossible?

I agree, we do need a separate instance of each of these classes for each encoding operation. Not sure what that contradicts, I think you meant to link to a comment.

peter-hofer: > @peter-hofer is there any way to inject this interface to `SpeculationLog` when using an openJDK which lacks it (using something like `com.oracle.svm.core.annotate.TargetClass#onlyWith`)
> 
> Alternatively, could we add the interface in the sources and compile it in only when needed? (possibly through mx).

Using substitutions is a clever idea, but they don't support changes to the class hierarchy. What we could try is substituting methods of the dynamic proxy class to call the encoder directly instead. Substituting a generated class like a proxy sounds a bit fragile, though.

zakkak: > > Shouldn't the said `IsolatedSpeculationReasonEncoding` be instantiated for each new encoding and as a result don't we need a new `SpeculationReasonInvocationHandler` for each new encoding and consequently a new proxy instance rendering https://github.com/oracle/graal/pull/2945/files#r519816278 impossible?
> 
> I agree, we do need a separate instance of each of these classes for each encoding operation. Not sure what that contradicts, I think you meant to link to a comment.

It looks like after force pushing links to out-of-date discussions breaks :man_facepalming: 
New link after force pushing is https://github.com/oracle/graal/pull/2945#discussion_r519816278

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
zakkak: Not sure how to benchmark this. I will be happy to if you provide instructions though :)

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
zakkak: I cached the constructor as well.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
zakkak: > I doubt the image build can optimize the reflective accesses, and on top of repeatedly doing lookups at runtime, we will need to register the methods of `IsolatedSpeculationReasonEncoding` as reflectively accessible in a `Feature` during the image build.
So I'd rather avoid this proxy, but I don't see how we can have a class that directly implements SpeculationReasonEncoding when that interface cannot be referenced in the code.

@peter-hofer please have a look at https://github.com/zakkak/mandrel/commit/19398016bc4bf77061268db7f5395ebd9286ebb6, it removes the proxy and relies solely on our cloned interface.
If the original `encode` gets invoked elsewhere it will essentially generate the same encoding and cache it in the corresponding `EncodedSpeculationReason` instance.
If I am not mistaken this should work without the performance penalty.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +58,140 @@ public void collectFailedSpeculations() {
...
+        private static final Class<?> SPECULATION_REASON_ENCODING_CLASS;
+        private static final String SPECULATION_REASON_ENCODING_NAME = "jdk.vm.ci.meta.SpeculationLog$SpeculationReasonEncoding";
+
+        static {
+            Class<?> speculationReasonEncodingClass = null;
+            try {
+                speculationReasonEncodingClass = Class.forName(SPECULATION_REASON_ENCODING_NAME);
+            } catch (ClassNotFoundException e) {
+                VMError.shouldNotReachHere(SPECULATION_REASON_ENCODING_NAME +
+                                " not found in JDK. UnencodedSpeculationReason should have been used as SpeculationReason");
```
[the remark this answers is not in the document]
zakkak: OK, I changed to initialize everything to null in that case and catch the NullPointerException in `get()` to print a hopefully more helpful error message :)

```diff
diff --git a/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java b/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
index 75a3eca830f..eb5a4f915b6 100644
--- a/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
+++ b/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
@@ -78,31 +78,34 @@ public final class IsolatedSpeculationLog extends IsolatedObjectProxy<Speculatio
         private static final String CLASS_NAME = "jdk.vm.ci.meta.SpeculationLog$SpeculationReasonEncoding";
 
         static {
-            Class<?> speculationReasonEncodingClass = null;
+            Class<?> speculationReasonEncodingClass;
             try {
                 speculationReasonEncodingClass = Class.forName(CLASS_NAME);
             } catch (ClassNotFoundException e) {
-                VMError.shouldNotReachHere(CLASS_NAME +
-                                " not found in JDK. UnencodedSpeculationReason should have been used as SpeculationReason");
+                speculationReasonEncodingClass = null;
             }
             CLASS = speculationReasonEncodingClass;
-            assert CLASS != null;
-            PROXY_CLASS = Proxy.newProxyInstance(CLASS.getClassLoader(), new Class<?>[]{CLASS},
-                            new SpeculationReasonInvocationHandler(new IsolatedSpeculationReasonEncoding())).getClass();
-            Constructor<?> declaredConstructor = null;
-            try {
-                declaredConstructor = PROXY_CLASS.getDeclaredConstructor(CLASS);
-            } catch (NoSuchMethodException e) {
-                VMError.shouldNotReachHere("Failed to get constructor for " + PROXY_CLASS);
+            if (CLASS != null) {
+                PROXY_CLASS = Proxy.newProxyInstance(CLASS.getClassLoader(), new Class<?>[]{CLASS},
+                        new SpeculationReasonInvocationHandler(new IsolatedSpeculationReasonEncoding())).getClass();
+                Constructor<?> declaredConstructor = null;
+                try {
+                    declaredConstructor = PROXY_CLASS.getDeclaredConstructor(CLASS);
+                } catch (NoSuchMethodException e) {
+                    VMError.shouldNotReachHere("Failed to get constructor for " + PROXY_CLASS);
+                }
+                CONSTRUCTOR = declaredConstructor;
+            } else {
+                PROXY_CLASS = null;
+                CONSTRUCTOR = null;
             }
-            CONSTRUCTOR = declaredConstructor;
         }
 
         @Override
         public Object get() {
             try {
                 return CONSTRUCTOR.newInstance(new SpeculationReasonInvocationHandler(new IsolatedSpeculationReasonEncoding()));
-            } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
+            } catch (NullPointerException | InstantiationException | IllegalAccessException | InvocationTargetException e) {
                 VMError.shouldNotReachHere("Failed to instantiate " + CLASS_NAME + "through proxy", e);
             }
             return null;
```

zakkak:
(Pasting https://github.com/oracle/graal/pull/2945#discussion_r520948612 here as well just to make sure it didn't sleep through the many review comments)

> I doubt the image build can optimize the reflective accesses, and on top of repeatedly doing lookups at runtime, we will need to register the methods of `IsolatedSpeculationReasonEncoding` as reflectively accessible in a `Feature` during the image build.
So I'd rather avoid this proxy, but I don't see how we can have a class that directly implements SpeculationReasonEncoding when that interface cannot be referenced in the code.

@peter-hofer please have a look at https://github.com/zakkak/mandrel/commit/19398016bc4bf77061268db7f5395ebd9286ebb6 (which is not part of this PR), it removes the proxy and relies solely on our cloned interface.
If the original `encode` gets invoked elsewhere it will essentially generate the same encoding and cache it in the corresponding `EncodedSpeculationReason` instance.
If I am not mistaken this should work without the performance penalty.

FYI: The full diff of this approach against master can be seen [here](https://github.com/oracle/graal/compare/master...zakkak:no-speculationreasonencoding-for-openjdk11-v2)

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
peter-hofer: > @peter-hofer please have a look at [zakkak@1939801](https://github.com/zakkak/mandrel/commit/19398016bc4bf77061268db7f5395ebd9286ebb6), it removes the proxy and relies solely on our cloned interface.

Looks good. I did encounter problems when the proxy class is created at image build time, so that's another reason to avoid it.
However, I would like to have two separate code paths which directly call `UnencodedSpeculationReason.class.getDeclaredField("...").get*(reason)` and `Class.forName("jdk.vm.ci.meta.EncodedSpeculationReason").getDeclaredField("...").get*(reason)` and pass the values to a method that does the encoding. That way, we can be sure that these reflective accesses will become direct field accesses during the image build.
My only worry other than that is that `EncodedSpeculationReason` may be changed in the future and this code silently becomes broken, which would have been prevented by using a proxy class.

zakkak: > My only worry other than that is that `EncodedSpeculationReason` may be changed in the future and this code silently becomes broken, which would have been prevented by using a proxy class.

Yes I also thought about this, but even with the proxy in place this would still be the case for the `UnencodedSpeculationReason` path, no?
I guess we should implement some kind of test and add it to `mx gate` so that if/when `EncodedSpeculationReason` changes we get notified (the simplest I can think of is comparing the result of the copied methods with the result of the corresponding methods in `EncodedSpeculationReason`).

zakkak:
Converting to draft till I get the test working.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
zakkak: @peter-hofer the test is now added to the PR, I hope this is enough to detect potential future changes in `EncodedSpeculationReason` that could cause issues.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +53,89 @@ public void collectFailedSpeculations() {
         collectFailedSpeculations0(IsolatedCompileContext.get().getClient(), handle);
     }
 
+    private static byte[] encodeAsByteArray(SpeculationReason reason) {
+        int groupId = 0;
+        Object[] context = null;
+        try {
+            final Field groupIdField;
```
dougxc, line 60: If I understand the review conversation, these reflective accesses will be converted to non-reflective accesses during image building. A comment should be added somewhere here explaining that. Does the added unit test check that this happens?

zakkak: > Does the added unit test check that this happens?

Not at its current state. It cannot be compiled with native-image due to the invocation handler that is required to create the supplier that we pass to labsJDK's internal `encode`.
Would adding another test that only uses graalvm's implementations and running it with `native-unittest` be enough?

dougxc: If that tests the condition stated in the (to be added) comment, then yes.

zakkak: According to https://www.graalvm.org/reference-manual/native-image/Reflection/ 

>  If the target elements can be resolved, the calls are removed and instead the target elements are embedded in the code. If the target elements cannot be resolved, e.g., a class is not on the classpath or it does not declare a field/method/constructor, then the calls are replaced with a snippet that throws the appropriate exception at run time.

So to my understanding just compiling and running an image that covers the corresponding code segments should be enough.

Is there any better way to achieve this?

zakkak: [Comment](https://github.com/oracle/graal/pull/2945/files#diff-603f78c0054580a456da1e1f763e13bd03cd60958a2f5358fb25509c268d57f6R62-R67) added [test](https://github.com/oracle/graal/pull/2945/files#diff-69e104d7b1d63ef862b22d172b7e6727584403f02e35ff8250084fc6dae8f51f) added.

peter-hofer: Indeed, if the members cannot be resolved directly and are not registered by other means, usually a configuration file, using them will fail.

peter-hofer:
I'll run our tests on the current state and let you know.

peter-hofer:
Looks good overall, but there is another problem: for JDK 8, we have a class `GraalServices.EncodedSpeculationReason` which the current code tries to cast to `jdk.vm.ci.meta.EncodedSpeculationReason` and fails.

peter-hofer:
@dougxc , how come we need that extra class on JDK 8 rather than using `jdk.vm.ci.meta.EncodedSpeculationReason` or `UnencodedSpeculationReason` like on JDK 11?

dougxc:
I cannot recall the reasons. I think using `jdk.vm.ci.meta.EncodedSpeculationReason` should work.

zakkak:
Thank you @olpaw and @dougxc

I made `GraalServices.createSpeculationReason()` use  `jdk.vm.ci.meta.EncodedSpeculationReason` and removed `GraalServices.EncodedSpeculationReason`.
I changed the compliance level of the junit tests I added to "8+" so that we can use them for both JDK 8 and 11.
I also augmented the native junit test to test `GraalServices.createSpeculationReason()` as well.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/SpeculationReasonEncodingClone.java
```diff
@@ -0,0 +1,52 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.graal.isolated;
+
+import jdk.vm.ci.meta.ResolvedJavaField;
+import jdk.vm.ci.meta.ResolvedJavaMethod;
+import jdk.vm.ci.meta.ResolvedJavaType;
+
+/** Copied from {@code jdk.vm.ci.hotspot.HotSpotSpeculationEncoding.SpeculationReasonEncoding}. */
```
dougxc, line 31: This comment should also explain why copying is necessary so that no one is tempted to "uncopy" it.

substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/isolated/IsolatedSpeculationLog.java
```diff
@@ -48,19 +53,95 @@ public void collectFailedSpeculations() {
...
+                final Class<?> klass = UnencodedSpeculationReason.class;
+                groupIdField = klass.getDeclaredField("groupId");
+                contextField = klass.getDeclaredField("context");
+            } else {
+                final Class<?> klass = Class.forName("jdk.vm.ci.meta.EncodedSpeculationReason");
+                groupIdField = klass.getDeclaredField("groupId");
+                contextField = klass.getDeclaredField("context");
+            }
+            groupIdField.setAccessible(true);
+            groupId = groupIdField.getInt(reason);
```
peter-hofer, line 78: Feedback from internal review: 

> Instead of reflection, it is better to use a method handle returned by `MethodHandles.lookup().unreflectGetter` and stored in a static final field. The method handle is then really intrinsified to a direct field access.

zakkak: I tried this:
```java
if (reason instanceof UnencodedSpeculationReason) {
    final Class<?> klass = UnencodedSpeculationReason.class;
    final Field groupIdField = klass.getDeclaredField("groupId");
    final Field contextField = klass.getDeclaredField("context");
    final MethodHandle groupIdHandle = MethodHandles.lookup().unreflectGetter(groupIdField); // fails here
    final MethodHandle contextHandle = MethodHandles.lookup().unreflectGetter(contextField);
    groupId = (int) groupIdHandle.invoke(reason);
    context = (Object[]) contextHandle.invoke(reason);
} else {
    final Class<?> klass = Class.forName("jdk.vm.ci.meta.EncodedSpeculationReason");
    final Field groupIdField = klass.getDeclaredField("groupId");
    final Field contextField = klass.getDeclaredField("context");
    final MethodHandle groupIdHandle = MethodHandles.lookup().unreflectGetter(groupIdField);
    final MethodHandle contextHandle = MethodHandles.lookup().unreflectGetter(contextField);
    groupId = (int) groupIdHandle.invoke(reason);              // fails here
    context = (Object[]) contextHandle.invoke(reason);         // and here
}
```

but the if-case fails with:
```
Error: com.oracle.svm.hosted.substitute.DeletedElementException: Unsupported type java.lang.invoke.MemberName is reachable: All methods from java.lang.invoke should have been replaced during image building.
To diagnose the issue, you can add the option --report-unsupported-elements-at-runtime. The unsupported element is then reported at run time when it is accessed the first time.
```
for
```java
                final MethodHandle groupIdHandle = MethodHandles.lookup().unreflectGetter(groupIdField);
```

and the else-case with:
```
Error: com.oracle.graal.pointsto.constraints.UnsupportedFeatureException: Invoke with MethodHandle argument could not be reduced to at most a single call or single field access. The method handle must be a compile time constant, e.g., be loaded from a `static final` field. Method that contains the method handle invocation: java.lang.invoke.MethodHandle.invokeBasic(Object)
To enable method handles that do not require LambdaForm interpretation (e.g. because of a call to MethodHandle.bindTo()) or to diagnose the issue, you can add the option --report-unsupported-elements-at-runtime. The error is then reported at run time when the invoke is executed and the method handle has to be interpreted.
```
for
```java
                groupId = (int) groupIdHandle.invoke(reason);
                context = (Object[]) contextHandle.invoke(reason);
```
Which, if I am not mistaken, hints that the accesses were not beeing intrinsified before the change as well.

peter-hofer: Looks like the accesses cannot be intrinsified in the code, but what if you store them in a `static final` field? It should be possible to move the class detection logic to a `static` initializer for that.

zakkak: Tried that as well (see https://github.com/zakkak/mandrel/commit/8a6a7967324b1ab2ee73a5474b9245386b6e6dd3), but it still gives
```
Error: com.oracle.graal.pointsto.constraints.UnsupportedFeatureException: Invoke with MethodHandle argument could not be reduced to at most a single call or single field access. The method handle must be a compile time constant, e.g., be loaded from a `static final` field. Method that contains the method handle invocation: java.lang.invoke.LambdaForm$MH/658839328.invoke_MT(Object, Object, Object)
```

peter-hofer: I finally got around to look into this today, sorry for the delay. The issue is that calling the getter on the interface type (rather than a concrete subclass which is guaranteed to have the accessed fields) introduces a type-check, which makes the method handle logic too complex to inline.
I'll merge the reflection variant.

zakkak: Thanks you @peter-hofer!
Could you please outline the process you used to figure this out (if possible with tools available to the community)?

I tried creating bgv files from building the native `native-image` but I couldn't find the graph for `encodeAsByteArray`.

peter-hofer: From debugging the code that creates the error in `IntrinsifyMethodHandlesInvocationPlugin`, I could gather that the method handle graph contains an `If` node where it isn't supported. I then used the command line `mx native-unittest IsolatedSpeculationLog --build-args -H:Dump=:1 -H:MethodFilter=encodeAsByteArray` to get a bgv file with the method handle graph that cannot be inlined and saw that it has a type check.
I then consulted with @christianwimmer who told me to look at the parameter node's type. That turned out to be `SpeculationReason`, which is an interface that of course doesn't have the accessed field(s), so the type check is necessary. Since we use `Class.forName` and so cannot define a statically typed variable of the subclass with the fields, we cannot proceed.

zakkak: That helps a lot, thank you for the information @peter-hofer

PR #2952, opened 2020-11-01
Fix some warnings in Sulong

chrisseaton:
Fix some warnings in Sulong that I see on macOS.


rschatz (approved):
Huh, weird, I thought we just fixed all our warnings recently, but we probably didn't test on Mac.

The fix looks good, thanks!

PR #2957, opened 2020-11-02
Call tree analysis with Neo4j

galderz:
In order to improve debugging of analysis issues, I've been exploring producing neo4j graphs instead text based call graphs. This PR is an effort to do that and should be considered as a draft.

Functionality wise, it should be complete and equivalent to the existing call tree output. It includes:
* A VM entry point node.
* A `Method` node for each distinct direct, virtual and overriden method.
* Each `Method` includes the method name and the complete signature. 
* Entry relationships between VM and entry method nodes.
* Direct and virtual relationships between method nodes include bytecode index information.
* Bytecode index information expresses the possibility of a method calling another one from different locations within the origin method.
* Overriden relationships between virtual and override methods. These relationships have no bytecode index information.

The idea is to produce a cypher script that would load the equivalent of the existing call tree printer into a neo4j db. This adds not extra dependencies, it just produces an equivalent text file that can be loaded into neo4j via:

```
cat reports/call_tree_helloworld_20201102_115920.cypher | /opt/neo4j/bin/cypher-shell -u neo4j -p neo
```

It uses the techniques explained in [this article](https://medium.com/neo4j/efficient-neo4j-data-import-using-cypher-scripts-7d1268b0747) to efficiently batch data and take advantage of script caching. In spite of these efforts, loading data is not very fast right now. The hello world example has a total of ~8k method nodes (8869 to be precise, including virtual ones) and ~27k relationships. Loading that takes ~4m on my home server. For reference, my home server takes 12s to generate a hello world native image. I'm hoping @michael-simons or other neo4j experts (e.g. @DavideD) can help shape this into performing better.

Attaching a sample neo4j graph and a rough screenshot of the area it'd represent. You can find the full cypher output for hello world [here](https://www.dropbox.com/s/s4f8gf0nf8np6kg/call_tree_helloworld_20201102_115920.cypher?dl=0).

Implementation wise, it simply reuses the logic in `CallTreePrinter` to generate the graph and it uses the types referenced there.

TODO:
- [ ] Decide on granularity of elements on Methods. Right now it has name and signature, but the signature could be split into class name, parameters and returns, and each of the types there could be node themselves and link together. The cypher needs to process this faster to be usable.
- [ ] Improve performance.
- [ ] Decide on configuration options to get the output. Right now it just piggy backs on the print analysis call option.

Screen Shot 2020-11-02 at 12 06 28
Screen Shot 2020-11-02 at 12 06 10

### DIFF
--- a/substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/reports/CallTreeCypher.java
+++ b/substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/reports/CallTreeCypher.java
@@ -24,9 +24,9 @@ import java.util.stream.Stream;
 public class CallTreeCypher {
 
     // TODO temporarily use a system property to try different values
-    private static final int BATCH_SIZE = Integer.getInteger("cypher.batch.size", 256);
+    private static final int BATCH_SIZE = Integer.getInteger("cypher.batch.size", 512);
 
-    private static final String METHOD_INFO_FORMAT = "properties:{name:'%n', signature:'" + CallTreePrinter.METHOD_FORMAT + "'}";
+    private static final String METHOD_INFO_FORMAT = "name:'%n', signature:'" + CallTreePrinter.METHOD_FORMAT + "'";
 
     private static final AtomicInteger virtualNodeId = new AtomicInteger(-1);
 
@@ -55,7 +55,7 @@ public class CallTreeCypher {
         final String unwindMethod = unwindMethod();
         final String script = methodsBatches.stream()
                 .map(methodBatch -> methodBatch.stream()
-                        .map(method -> String.format("{_id:%d, %s}", method.id, method.method.format(METHOD_INFO_FORMAT)))
+                        .map(method -> String.format("{_id:%d, %s}", method.id, asProperties(method.method.format(METHOD_INFO_FORMAT))))
                         .collect(Collectors.joining(", ", ":param rows => [", "]"))
                 )
                 .collect(Collectors.joining(unwindMethod, "", unwindMethod));
@@ -63,16 +63,28 @@ public class CallTreeCypher {
         writer.print(script);
     }
 
+    private static String asProperties(String properties) {
+        return "properties:{" + properties + "}";
+    }
+
     private static void printVirtualNodes(Map<VirtualInvokeId, Integer> virtualNodes, PrintWriter writer) {
         final Collection<List<Map.Entry<VirtualInvokeId, Integer>>> virtualNodesBatches = batched(virtualNodes.entrySet());
-        final String unwindMethod = unwindMethod();
+        final String unwindMethod = unwindVirtualMethod();
         final String script = virtualNodesBatches.stream()
                 .map(virtualNodesBatch -> virtualNodesBatch.stream()
-                        .map(virtualNode -> String.format("{_id:%d, %s}", virtualNode.getValue(), virtualNode.getKey().methodInfo))
+                        .map(virtualNode -> String.format("{_id:%d, %s}", virtualNode.getValue(), virtualMethodProperties(virtualNode.getKey())))
                         .collect(Collectors.joining(", ", ":param rows => [", "]"))
                 )
                 .collect(Collectors.joining(unwindMethod, "", unwindMethod));
         writer.print(script);
+    }
+
+    private static String virtualMethodProperties(VirtualInvokeId virtualNode) {
+        final String linkedIndexes = virtualNode.bytecodeIndexes.stream()
+                .map(String::valueOf)
+                .collect(Collectors.joining("->"));
+        String properties = virtualNode.methodInfo + ", bci: '" + linkedIndexes + "'";
+        return asProperties(properties);
     }
 
     private static String vmEntryPoint() {
@@ -86,6 +98,14 @@ public class CallTreeCypher {
                 "UNWIND $rows as row\n" +
                 "CREATE (n:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row._id})\n" +
                 "  SET n += row.properties SET n:Method;\n" +
+                ":commit\n\n";
+    }
+
+    private static String unwindVirtualMethod() {
+        return "\n\n:begin\n" +
+                "UNWIND $rows as row\n" +
+                "CREATE (n:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row._id})\n" +
+                "  SET n += row.properties SET n:VirtualMethod;\n" +
                 ":commit\n\n";
     }
 
--- a/substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/reports/CallTreeCypher.java
+++ b/substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/reports/CallTreeCypher.java
@@ -61,7 +61,7 @@ public class CallTreeCypher {
     private static void printAll(Map<AnalysisMethod, MethodNode> methodToNode, PrintWriter writer) {
         Set<MethodNode> entryNodes = new HashSet<>();
         Set<MethodNode> nonVirtualNodes = new HashSet<>();
-        Map<VirtualInvokeId, Integer> virtualNodes = new HashMap<>();
+        Map<String, Integer> virtualNodes = new HashMap<>();
 
         Map<Integer, Set<BciEndEdge>> directEdges = new HashMap<>();
         Map<Integer, Set<BciEndEdge>> virtualEdges = new HashMap<>();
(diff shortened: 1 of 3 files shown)


galderz:
p.s. Just got passed a link to the article [Import 10M Stack Overflow Questions into Neo4j In Just 3 Minutes](https://neo4j.com/blog/import-10m-stack-overflow-questions). I'll look into that to see if import can be faster :)

michael-simons:
Thanks for pinging me! I'll have a look at your Cypher.

michael-simons:
So a couple of things: First of all, I find that use of so many labels quite irritating. I think the intention of @conker84 behind this is to make use of the implicit index on different labels. Maybe it is because the import he described in the medium post should be generic in the end. I don't know. Furthermore - and that's what is making your import dead slow - the label is by far not unique.

Have a close look at this piece:

```
:begin
UNWIND $rows as row
CREATE (n:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row._id})
  SET n += row.properties SET n:Method;
:commit
```

The label is *always* exactly `UNIQUE IMPORT LABEL`. You cannot create dynamic labels based on parameters in Neo4j, neither in Browser or in Shell.

So when creating the relationships after having imported all the nodes, you always do a full node scan when matching on the label and the `UNIQUE IMPORT ID` property as the label is not unique and for the property no index exists.

To fix this, you need to create a unique index 

```
CREATE CONSTRAINT unique_import_id ON (n:`UNIQUE IMPORT LABEL`) ASSERT n.`UNIQUE IMPORT ID` IS UNIQUE;
```

However, the import will fail then as it turns out that the used id is not unique.
If it describes the same method calls, it is safe to replace the following blocks

```
:begin
UNWIND $rows as row
CREATE (n:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row._id})
  SET n += row.properties SET n:Method;
:commit
```

with

```
:begin
UNWIND $rows as row
MERGE (n:`UNIQUE IMPORT LABEL`{`UNIQUE IMPORT ID`: row._id})
  SET n += row.properties SET n:Method;
:commit
```

This will create the nodes if they not yet exists with the given pattern, otherwise just match them and the properties will be overwritten. If this is the case - I don't know and cannot judge this here - this is the way forward.

With the changes in place, the script runs against 4.0 community edition on my machine in about 9 seconds.

If your use case is still fulfilled, I recommend not using the naming at all but do this:

1. Assuming this unique index. You could create a comment atop the file for example with the code in it.

```
CREATE CONSTRAINT unique_method_id ON (n:Method) ASSERT n._id IS UNIQUE;
```

2. Do this for the VM node

```
:param rows => [{_id:-1, properties:{name:'VM'}}]
:begin
UNWIND $rows as row
MERGE (n:Method{_id: row._id})
  SET n += row.properties SET n:VM;
:commit
```

3. Do this for all the method nodes

```
:begin
UNWIND $rows as row
MERGE (n:Method{_id: row._id})
  SET n += row.properties;
:commit
```

4. Do this for the relationships

```
:begin
UNWIND $rows as row
MATCH (start:Method {_id: row.start._id})
MATCH (end:Method {_id: row.end._id})
CREATE (start)-[r:OVERRIDDEN_BY]->(end);
:commit
```

If you need further help with the uniqueness of method nodes and their properties, just give me a sign.

galderz:
Thx for hints @michael-simons. Having read the link above about the import command with CSV entries, is that a better approach compared with the current cypher script approach for importing huge amounts of nodes and relationships?

michael-simons:
Hi. I don't think so. The major difference is in what I said about uniqueness of labels and indices. If you don't use any indices and do a full store scan, than both imports will be slow.

Probably the CSV one is a tad bit faster, but less convenient than just pipe it through Cypher shell. 

I would run your thing on a bigger example and see how the difference evolves.

zakkak:
> Probably the CSV one is a tad bit faster, but less convenient than just pipe it through Cypher shell.

On the other hand CSV is a universal format so exporting the data in CSV format would give users more flexibility, so it might still be worth considering it.

michael-simons:
If you like, attach a CSV file with the same rows and properties as your Cypher. I can help you with the necessary import statements.

galderz:
@zakkak @michael-simons and everyone else interested, I've started a discussion on CSV vs Cypher on the graalvm slack channel, see [here](https://graalvm.slack.com/archives/CN9KSFB40/p1604389148113800). That'll give us a more dynamic place to share ideas, experiment results without overloading this PR with that info.

conker84:
Hi everybody as @michael-simons said that article had the goal to show the performance improvements that we had in the `apoc.export.cypher` procedure, while here the code seems handwritten. IMHO Micheal covered any possible area of improvement of @galderz's script.

galderz:
@michael-simons @zakkak I've experimented with CSV and created a [branch](https://github.com/galderz/mandrel/tree/t_call_tree_cypher_v3_csv) with it. A sample report folder can be found [here](https://www.dropbox.com/s/z0s6adzg27wf3g4/reports-csv-1112.tgz?dl=0). I can see what @michael-simons meant with inconvenience with CSV. To load all that CSV, you have to copy the CSV files to the Neo4j installation's import folder, and then it can be loaded with the cypher in there. The result is a more cumbersome to load into neo4j compared to having it all in cypher and then just pipe it. You don't need to start copying files around.

Performance wise, it took ~3m30s with the CSV method, whereas my original approach was taking 4m. This still feels a lot for not so much: ~8k method entries, and 27k relationships. I've tried different heap sizes (up to 16gb) and didn't make any difference. I also tried different `USING PERIODIC COMMIT` values (up to 4k) and that didn't change much either. The neo4j blog post [here](https://neo4j.com/blog/import-10m-stack-overflow-questions/) barely has details on the env (all it says it's a regular laptop with SSD), but it seems to take roughly the same time to load several orders of magnitude more data (31m nodes, 77m relationships).

zakkak:
@galderz adding
```
CREATE CONSTRAINT unique_method_id ON (m:Method) ASSERT m.methodId IS UNIQUE;
```
(as suggested by @michael-simons in https://github.com/oracle/graal/pull/2957#issuecomment-720563773)
in the beginning of `cypher_call_tree_helloworld_20201211_112253.cypher` brought the import from 4m and 30s down to 7s for me.

galderz:
@zakkak Oh wow 😀 . I'll explore that option further in both approaches and see what it looks like.

Also @zakkak, any thoughts on preference of cypher or csv? I think the cypher, although less universal, it's much easier to use with neo4j compared with csv (see my previous comment). I'll tidy this up this branch, that uses cypher, in the next few days.

galderz:
@zakkak Oh and thx for trying things!

zakkak:
> Also @zakkak, any thoughts on preference of cypher or csv? I think the cypher, although less universal, it's much easier to use with neo4j compared with csv (see my previous comment). I'll tidy this up this branch, that uses cypher, in the next few days.

Although I agree that having to copy the files to neo4j's import folder is a bit cumbersome, I personally think that CSV is the right (more generic) approach to follow that makes it easier to:

1. manipulate the output (even with shell commands) before importing it to a visualization tool
2. import the data in different tools
3. extend GraalVM to support more tools for visualization/exploration

### Sidenote
In my experiments I use the neo4j container image and mount the directory with the csv files using `-v` to avoid copying files (and having to purge the database).

1. To start the server I use:
```shell
podman run --rm --name=graal-neo4j \
    --user $(id -u):$(id -u) --userns=keep-id \
    -p7474:7474 -p7687:7687 \ 
    -e NEO4J_AUTH=neo4j/s3cr3t \ 
    -v /home/zakkak/Downloads/reports:/var/lib/neo4j/import:z \
    neo4j
``` 

2. To import the data I use:
```shell
podman exec -ti graal-neo4j \
    bin/cypher-shell \
    -u neo4j -p s3cr3t \
    -f /var/lib/neo4j/import/cypher_call_tree_helloworld_20201211_112253.cypher
```

3. To explore the graph I visit http://localhost:7474/

galderz:
@zakkak Good points! Just added 2 constraints (for vm and method id) in the csv branch and the time went down to 2.5 secs 🚀🚀🚀

galderz:
I've done more testing of loading times:

* Quarkus getting started quickstart: 5.5 seconds
* Infinispan Quakurs server: 18 seconds (~150k nodes, 450k relationships)

I'm doing further improvements to the functionality:

1. Separate `package` into `package` and `className`
2. Add a display-friendly package name field. For say package `org.graalvm...` it'd show `o.g....`.
3. Then, that can be combined with class name and method name in the neo4j css to show all that info in the node right away.

Finally, the type of cypher queries that would be useful here would be:

1. Find a method name and all its incoming relationships. This is to find out what brings a given method into the universe:

```
match (m:Method {name: "cloneKey"}) <- [r] - (o) return *
```

2. Find a methods incoming relationships 2 layers deep:

```
match (m:Method {name: "cloneKey"}) <- [r*1..2] - (o) return *
```

3. Find a method with with its incoming relationships, filtered by only its `DIRECT` and `VIRTUAL` incoming relationships:

```
match (m:Method {name: "cloneKey"}) <- [:DIRECT|:VIRTUAL] - (o) return *
```

galderz:
And thx @michael-simons for neo4j browser and query tips!

michael-simons:
TBH, I only gave some tips. The browser is done by a couple of great colleagues.

galderz:
Closing this PR. A new version of this PR is available [here](https://github.com/oracle/graal/pull/3128).

PR #2960, opened 2020-11-03
--target parameter, Platform/LibC changes

lazar-mitrovic:
- Add `--target` parameter which can be used as:
    ```
    native-image --target=linux-amd64 HelloWorld
     ```
- Added initial Android/iOS Platform classes as part of initial mobile support.
- Load platforms and LibCs via ServiceLoader
- Add `os.name` system property support (as specified in JDK)

### DIFF
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
@@ -84,6 +84,31 @@ public interface Platform {
         return platformGroup.isInstance(ImageSingletons.lookup(Platform.class));
     }
 
+    /**
+     * Returns the string representing Platform's OS name.
+     * <p>
+     * This method should be implemented either in a final class or as default method in respective
+     * OS interface.
+     *
+     * @since 21.0
+     */
+    default String getOS() {
+        throw new UnsupportedOperationException("Platform doesn't implement getOS");
+    }
+
+    /**
+     * Returns the string representing Platform's architecture name. This value should be the same
+     * as desired os.arch system property.
+     * <p>
+     * This method should be implemented either in final class or as default method in respective
+     * architecture interface.
+     *
+     * @since 21.0
+     */
+    default String getArchitecture() {
+        throw new UnsupportedOperationException("Platform doesn't implement getArchitecture");
+    }
+
     /*
      * The standard architectures that are supported.
      */
@@ -94,6 +119,14 @@ public interface Platform {
      */
     interface AMD64 extends Platform {
 
+        /**
+         * Returns string representing AMD64 architecture.
+         *
+         * @since 21.0
+         */
+        default String getArchitecture() {
+            return "amd64";
+        }
     }
 
     /**
@@ -103,6 +136,14 @@ public interface Platform {
      */
     interface AARCH64 extends Platform {
 
+        /**
+         * Returns string representing AARCH64 architecture.
+         *
+         * @since 21.0
+         */
+        default String getArchitecture() {
+            return "aarch64";
+        }
     }
 
     /*
@@ -115,6 +156,31 @@ public interface Platform {
      */
     interface LINUX extends InternalPlatform.PLATFORM_JNI {
 
+        /**
+         * Returns string representing LINUX OS.
+         *
+         * @since 21.0
+         */
+        default String getOS() {
+            return "linux";
+        }
+    }
+
+    /**
+     * Supported operating system: Android.
+     *
+     * @since 21.0
+     */
+    interface ANDROID extends LINUX {
+
+        /**
+         * Returns string representing ANDROID OS.
+         *
+         * @since 21.0
+         */
+        default String getOS() {
+            return "android";
+        }
     }
 
     /**
@@ -124,6 +190,31 @@ public interface Platform {
      */
     interface DARWIN extends InternalPlatform.PLATFORM_JNI {
 
+        /**
+         * Returns string representing DARWIN OS.
+         *
+         * @since 21.0
+         */
+        default String getOS() {
+            return "darwin";
+        }
+    }
+
+    /**
+     * Supported operating system: iOS.
+     *
+     * @since 21.0
+     */
+    interface IOS extends DARWIN {
+
+        /**
+         * Returns string representing IOS OS.
+         *
+         * @since 21.0
+         */
+        default String getOS() {
+            return "ios";
+        }
     }
 
     /**
@@ -133,6 +224,14 @@ public interface Platform {
      */
     interface WINDOWS extends InternalPlatform.PLATFORM_JNI {
 
+        /**
+         * Returns string representing WINDOWS OS.
+         *
+         * @since 21.0
+         */
+        default String getOS() {
+            return "windows";
+        }
     }
 
     /*
@@ -173,11 +272,28 @@ public interface Platform {
     }
 
     /**
+     * Supported leaf platform: Android on AArch64 64-bit.
+     *
+     * @since 21.0
+     */
+    final class ANDROID_AARCH64 implements ANDROID, AARCH64 {
+
+        /**
+         * Instantiates a marker instance of this platform.
+         *
+         * @since 21.0
+         */
+        public ANDROID_AARCH64() {
+        }
+
+    }
+
+    /**
      * Supported leaf platform: Darwin (MacOS) on x86 64-bit.
      *
      * @since 19.0
      */
-    class DARWIN_AMD64 implements DARWIN, AMD64 {
+    final class DARWIN_AMD64 implements DARWIN, AMD64 {
 
         /**
          * Instantiates a marker instance of this platform.
@@ -205,11 +321,27 @@ public interface Platform {
     }
 
     /**
+     * Supported leaf platform: iOS on AArch 64-bit.
+     *
+     * @since 21.0
+     */
+    final class IOS_AARCH64 implements IOS, AARCH64 {
+
+        /**
+         * Instantiates a marker instance of this platform.
+         *
+         * @since 21.0
+         */
+        public IOS_AARCH64() {
+        }
+    }
+
+    /**
      * Supported leaf platform: Windows on x86 64-bit.
      *
      * @since 19.0
      */
-    class WINDOWS_AMD64 implements WINDOWS, AMD64 {
+    final class WINDOWS_AMD64 implements WINDOWS, AMD64 {
 
         /**
          * Instantiates a marker instance of this platform.
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
@@ -93,7 +93,7 @@ public interface Platform {
      * @since 21.0
      */
     default String getOS() {
-        throw new UnsupportedOperationException("Platform doesn't implement getOS");
+        throw new UnsupportedOperationException("Platform `" + this.getClass().getCanonicalName() + "`, doesn't implement getOS");
     }
 
     /**
@@ -106,7 +106,7 @@ public interface Platform {
      * @since 21.0
      */
     default String getArchitecture() {
-        throw new UnsupportedOperationException("Platform doesn't implement getArchitecture");
+        throw new UnsupportedOperationException("Platform `" + this.getClass().getCanonicalName() + "`, doesn't implement getArchitecture");
     }
 
     /*
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/macho/MachOObjectFile.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/macho/MachOObjectFile.java
@@ -52,6 +52,8 @@ import com.oracle.objectfile.ObjectFile;
 import com.oracle.objectfile.SymbolTable;
 import com.oracle.objectfile.io.AssemblyBuffer;
 import com.oracle.objectfile.io.OutputAssembler;
+import org.graalvm.nativeimage.ImageSingletons;
+import org.graalvm.nativeimage.Platform;
 
 /**
  * Models a Mach-O relocatable object file.
@@ -80,7 +82,7 @@ public final class MachOObjectFile extends ObjectFile {
      * Create an empty Mach-O object file.
      */
     public MachOObjectFile(int pageSize) {
-        this(pageSize, MachOCpuType.from(System.getProperty("svm.targetArch") == null ? System.getProperty("os.arch") : System.getProperty("svm.targetArch")));
+        this(pageSize, MachOCpuType.from(ImageSingletons.lookup(Platform.class).getArchitecture()));
     }
 
     public MachOObjectFile(int pageSize, MachOCpuType cpuType) {
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/libc/MuslLibC.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/libc/MuslLibC.java
@@ -34,15 +34,6 @@ import org.graalvm.compiler.serviceprovider.JavaVersionUtil;
 
 public class MuslLibC implements LibCBase {
 
-    public MuslLibC() {
-        if (!SubstrateOptions.StaticExecutable.getValue()) {
-            throw UserError.abort("Musl can only be used for statically linked executables.");
-        }
-        if (JavaVersionUtil.JAVA_SPEC != 11) {
-            throw UserError.abort("Musl can only be used with labsjdk 11.");
-        }
-    }
-
     public static final String NAME = "musl";
 
     @Override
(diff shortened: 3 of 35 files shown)


sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
```diff
@@ -84,6 +86,31 @@ static boolean includedIn(Class<? extends Platform> platformGroup) {
         return platformGroup.isInstance(ImageSingletons.lookup(Platform.class));
     }
 
+    /**
+     * Returns string representing Platform's OS name
+     * <p>
+     * This method should be implemented either in final class or as default in respective OS
+     * interface.
+     *
+     * @since 20.3
```
vjovanov, line 95: 21.0

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
```diff
@@ -84,6 +86,31 @@ static boolean includedIn(Class<? extends Platform> platformGroup) {
         return platformGroup.isInstance(ImageSingletons.lookup(Platform.class));
     }
 
+    /**
+     * Returns string representing Platform's OS name
```
vjovanov, line 90: the string representation

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
```diff
@@ -84,6 +86,31 @@ static boolean includedIn(Class<? extends Platform> platformGroup) {
...
+     * <p>
+     * This method should be implemented either in final class or as default in respective OS
```
vjovanov, line 92: a final class 
as default is unclear

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
```diff
@@ -102,7 +131,36 @@ static boolean includedIn(Class<? extends Platform> platformGroup) {
...
 
+    /*
+     * Operating system bases.
+     */
+    /**
+     * Interface encapsulating all linux-derived operating systems.
+     *
+     * @since 20.3
+     */
+    interface LINUX_BASE extends InternalPlatform.PLATFORM_JNI {
```
vjovanov, line 147: I would move this to a non-API part.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
```diff
@@ -113,26 +171,54 @@ static boolean includedIn(Class<? extends Platform> platformGroup) {
...
     }
 
     /**
      * Supported operating system: Windows.
      *
      * @since 19.0
      */
-    interface WINDOWS extends InternalPlatform.PLATFORM_JNI {
-
+    interface WINDOWS extends WINDOWS_BASE {
```
vjovanov, line 218: I don't think we need `WINDOWS_BASE`?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/macho/MachOObjectFile.java
```diff
@@ -80,7 +82,7 @@
      * Create an empty Mach-O object file.
      */
     public MachOObjectFile(int pageSize) {
-        this(pageSize, MachOCpuType.from(System.getProperty("svm.targetArch") == null ? System.getProperty("os.arch") : System.getProperty("svm.targetArch")));
```
vjovanov, line 83: We  should fix this function.

vjovanov (approved):
LGTM

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/libc/MuslLibC.java
```diff
@@ -70,4 +61,14 @@ public boolean hasIsolatedNamespaces() {
     public boolean requiresLibCSpecificStaticJDKLibraries() {
         return true;
     }
+
+    @Override
+    public void checkIfLibCSupported() {
+        if (!SubstrateOptions.StaticExecutable.getValue()) {
+            throw UserError.abort("Musl can only be used for statically linked executables.");
+        }
+        if (JavaVersionUtil.JAVA_SPEC != 11) {
```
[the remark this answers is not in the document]
gradinac: The change should work out of the box, but it hasn't been tested yet. I think we should do it as part of a separate PR.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/libc/MuslLibC.java
```diff
@@ -70,4 +61,14 @@ public boolean hasIsolatedNamespaces() {
     public boolean requiresLibCSpecificStaticJDKLibraries() {
         return true;
     }
+
+    @Override
+    public void checkIfLibCSupported() {
+        if (!SubstrateOptions.StaticExecutable.getValue()) {
```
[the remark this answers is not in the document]
gradinac: The users would be required to have a dynamic musl linker installed in order to be able to run both the C query code and the generated native-image. I remember hitting segfaults when I tried this out a while back, most likely due to the generated executable picking the wrong dynamic linker/libraries.
We haven't really had any use-cases for this functionality yet as well - those who do use musl have so far been only interested in creating static native-images. We can of course enable and support this if users request such a feature.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -140,6 +144,9 @@ public static void setDebugInfoValueUpdateHandler(ValueUpdateHandler updateHandl
     @Option(help = "Path passed to the linker as the -rpath (list of comma-separated directories)")//
     public static final HostedOptionKey<String[]> LinkerRPath = new HostedOptionKey<>(null);
 
+    @Option(help = "String which would be appended to linker call")//
```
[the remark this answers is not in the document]
lazar-mitrovic: Fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/VMInspection.java
```diff
@@ -97,7 +97,7 @@ public static boolean isEnabled() {
 
 class DumpAllStacks implements SignalHandler {
     static void install() {
-        Signal.handle(Platform.includedIn(WINDOWS.class) ? new Signal("BREAK") : new Signal("QUIT"), new DumpAllStacks());
+        Signal.handle(Platform.includedIn(Platform.WINDOWS.class) ? new Signal("BREAK") : new Signal("QUIT"), new DumpAllStacks());
```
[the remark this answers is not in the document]
lazar-mitrovic: Fixed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java
```diff
@@ -314,57 +315,64 @@ public NativeImageGenerator(ImageClassLoader loader, HostedOptionProvider option
...
+
+    public static Platform loadPlatform(String os, String arch) {
+        ServiceLoader<Platform> loader = ServiceLoader.load(Platform.class);
+        for (Platform platform : loader) {
+            if (platform.getOS().equals(os) && platform.getArchitecture().equals(arch)) {
+                return platform;
+            }
+        }
+        throw UserError.abort("Platform specified as " + os + "-" + arch +
+                        " isn't supported.");
```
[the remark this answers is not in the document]
lazar-mitrovic: Fixed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ServiceLoaderFeature.java
```diff
@@ -103,7 +103,9 @@
      */
     private static final Set<String> SERVICES_TO_SKIP = new HashSet<>(Arrays.asList(
                     "java.security.Provider",                       // see SecurityServicesFeature
-                    "sun.util.locale.provider.LocaleDataMetaInfo"   // see LocaleSubstitutions
+                    "sun.util.locale.provider.LocaleDataMetaInfo",  // see LocaleSubstitutions
+                    "org.graalvm.nativeimage.Platform"  // type shouldn't be reachable after
+                                                        // intrinsification
```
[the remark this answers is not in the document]
lazar-mitrovic: Fixed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/c/CAnnotationProcessorCache.java
```diff
@@ -66,7 +72,23 @@
 
     public static class Options {
         @Option(help = "Indicate the C Annotation Processor to use previously cached native information when generating C Type information.")//
-        public static final HostedOptionKey<Boolean> UseCAPCache = new HostedOptionKey<>(false);
+        public static final HostedOptionKey<Boolean> UseCAPCache = new HostedOptionKey<Boolean>(false) {
+            @Override
+            public Boolean getValueOrDefault(UnmodifiableEconomicMap<OptionKey<?>, Object> values) {
+                if (!values.containsKey(this)) {
```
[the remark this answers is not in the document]
lazar-mitrovic: Should be fixed.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/macho/MachOObjectFile.java
```diff
@@ -80,7 +82,7 @@
...
+        this(pageSize, MachOCpuType.from(System.getProperty("svm.targetArch") == null ? ImageSingletons.lookup(Platform.class).getArchitecture() : System.getProperty("svm.targetArch")));
```
christianwimmer, line 85: How is it ever possible that `ImageSingletons.lookup(Platform.class).getArchitecture()` does not return the correct architecture? 

I think a goal of this PR should be to remove the system properties `svm.targetArch` and `svm.targetName`

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
```diff
@@ -103,6 +136,33 @@ static boolean includedIn(Class<? extends Platform> platformGroup) {
...
+
+    /*
+     * Operating system bases.
+     */
+    /**
+     * Interface encapsulating all linux-derived operating systems.
+     *
+     * @since 21.0
+     */
+    interface LINUX_BASE extends InternalPlatform.PLATFORM_JNI {
```
christianwimmer, line 157: I don't think we should introduce these `BASE` interfaces. We get a too deep and confusing hierarchy. I would not know when to use `LINUX` vs. `LINUX_BASE` in a `Platforms` annotation.

substratevm/src/com.oracle.svm.core.windows/src/com/oracle/svm/core/windows/WindowsSystemPropertiesSupport.java
```diff
@@ -122,13 +132,194 @@ private static String toJavaString(WCharPointer wcString, UnsignedWord length) {
...
-        int majorVersion = versionBytes.get(3);
-        int minorVersion = versionBytes.get(2);
-        return majorVersion + "." + minorVersion;
+        if (cachedOsNameAndVersion == null) {
+            cachedOsNameAndVersion = getOsNameAndVersion();
+        }
+        return cachedOsNameAndVersion.getRight();
+    }
+
+    public Pair<String, String> getOsNameAndVersion() {
```
christianwimmer, line 153: @pejovica FYI

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
vjovanov: You would always use LINUX_BASE unless it is Android specific or Linux specific. What would be a more minimal approach to handle Android and iOS specific things? 

They are a separate platform and this hierarchy tends to reflect that in the most minimal way we could find.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/macho/MachOObjectFile.java
lazar-mitrovic: I was trying to make this PR additive without breaking any of existing use-cases (for example Gluon Substrate, or non public platforms). My reasoning was that after this PR is accepted, removal of these non-api ways to specify non standard `os.name` shouldn't be an issue.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
vjovanov: One idea would be to call it `LINUX_FAMILY` to make things clear?

substratevm/src/com.oracle.svm.core.windows/src/com/oracle/svm/core/windows/WindowsSystemPropertiesSupport.java
```diff
@@ -122,13 +132,194 @@ private static String toJavaString(WCharPointer wcString, UnsignedWord length) {
...
+        /*
+         * Reimplementation of code from java_props_md.c
+         */
+        SysinfoAPI.OSVERSIONINFOEXA ver = StackValue.get(SysinfoAPI.OSVERSIONINFOEXA.class);
+        ver.dwOSVersionInfoSize(SizeOf.get(SysinfoAPI.OSVERSIONINFOEXA.class));
+        SysinfoAPI.GetVersionExA(ver);
+
+        boolean is64bit = true; /* ATM we only support 64-bit Windows OS's */
```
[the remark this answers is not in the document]
lazar-mitrovic: I implemented a workaround similar to [this](https://stackoverflow.com/a/5940770), since it was the least painful fix to do.

In the improbable case that in the future there are supported Windows systems which are using 64 bit architecture which name doesn't end with "64", this class will require changes anyway in order to support that system's naming convention.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
lazar-mitrovic: Maybe better example for why this approach would be beneficial is regarding current Apple developments. ATM Darwin base/family contains 3 different operating systems on `Aarch64` architecture (iOS, iPadOS and MacOS). Without some kind of base platform, differences between those systems would need to be implemented by opting-out and bunch of checks throughout the code base.

As for Android/Linux, differences are largely based on LibC (glibc/bionic), however our current approach means that there would be a bunch of extra checks (for AWT for example, since Android doesn't include Xserver) in order to separate those two.

If we ever mean to add this kind of abstraction, it would be best that it is done now, since changes at a later date would probably be API breaking. Otherwise, all those changes are located in latest commit of this PR, so they are easily revertable.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/macho/MachOObjectFile.java
christianwimmer: I'm talking here about internal usages of these system properties. I'm fine if we keep them set in the image builder for another release. But all internal usages should go away immediately in this PR.

substratevm/src/com.oracle.svm.core.windows/src/com/oracle/svm/core/windows/WindowsSystemPropertiesSupport.java
lazar-mitrovic: Pejović was working with me on this commit

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
christianwimmer: Did you notice that you just said "use LINUX_BASE unless it is Linux specific" - which is exactly my point of confusion?

Is there really anything that cannot be modeled using `ANDROID_AARCH64` and the other leaf platforms for Linux? That would be best.

If we really need an intermediate platform, I would leave "LINUX" as it is to include every flavor including Android (the bulk of all usages) and introduce a `LINUX_SERVER` to distinguish it from `ANDROID`. That avoids all the renamings in this PR, and all future problems where people will write new `Platforms` annotations wrongly with `LINUX` that then breaks Android.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/macho/MachOObjectFile.java
lazar-mitrovic: Oh now I get it, you are right. I'll change this asap.

lazar-mitrovic: Should be fixed now.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/Platform.java
vjovanov: You are right, from Wikipedia: 
 - Linux is a family of open-source Unix-like operating systems based on the Linux kernel.
 - Darwin is an open-source Unix-like operating system first released by Apple Inc. Darwin forms the core set of components upon which macOS (previously OS X and Mac OS X), iOS, watchOS, tvOS, and iPadOS are based.

I think we can then use e.g., `macOS` and `iOS` to distinguish things on the Darwin side. And for Linux we could use another name (this can be done in a separate PR). 

I am not in favor of `LINUX_SERVER` because it can also be used on desktops. Maybe, 'LINUX_DISTRIBUTION' but we can cross that bridge when we get there. 

This commit was anyhow made last as a proposal.

lazar-mitrovic: Commit reverted.

PR #2961, opened 2020-11-03
Add java.runtime.version to hosted properties

zakkak:
As of https://github.com/oracle/graal/pull/2614/commits/75f9a50c4552e652558cdf9c63df8fc1aac6ce49 `java.runtime.version` is being used by `native-image --version` so it should be added to the hosted properties to avoid returning `null` as the Java version.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
@@ -53,6 +53,7 @@ public abstract class SystemPropertiesSupport {
     /** System properties that are taken from the VM hosting the image generator. */
     private static final String[] HOSTED_PROPERTIES = {
                     "java.version",
+                    "java.runtime.version",
                     ImageInfo.PROPERTY_IMAGE_KIND_KEY,
                     /*
                      * We do not support cross-compilation for now. Separator might also be cached


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
```diff
@@ -53,6 +53,7 @@
     /** System properties that are taken from the VM hosting the image generator. */
     private static final String[] HOSTED_PROPERTIES = {
                     "java.version",
+                    "java.runtime.version",
```
olpaw, line 56: The `java.runtime.version` value you get when asking for the property at image runtime should not be the same as the `java.runtime.version` value of the JVM that runs the image builder.

In the case of printing `--version` of the `native-image`-image it's ok to bake-in the value from the host system. (We want to show which version of GraalVM built the `native-image`-image and not what `java.runtime.version` runs the image.)

A good runtime value for `java.runtime.version` would contain information that identifies the actual VM running the code that executes the system property lookup. Maybe we should use something like

- `Java Version 11.0.9+10-svm-20.3-b04` or
- `Java Version 11.0.9+10-substratevm-20.3-b04`
- `Java Version 11.0.9+10-native-image-20.3-b04`
- `Java Version 11.0.9+10-native-image-vm-20.3-b04`

@christianwimmer do you have a favorite?

olpaw (changes requested):
See https://github.com/oracle/graal/pull/2961/files/4df358349e99d67bf16b6fef1276aa7c7991daf2#r521973500

olpaw (changes requested):
See https://github.com/oracle/graal/pull/2961/files/4df358349e99d67bf16b6fef1276aa7c7991daf2#r521973500

christianwimmer:
`java.runtime.version` is the version of the JDK core libraries that are used (in contrast to `java.vm.version`). So I would say `java.runtime.version` should inherit the value from the image builder.

gilles-duboscq:
I see that in some places in the compiler (in hotspot specific code) we extract the jvmci version from `java.vm.version`. I'm not sure about `java.runtime.version`. It might simply be because that's how the JDK build scripts work. We specify the jvmci version so the configure script and maybe that's just reflected in multiple system properties.
/cc @dougxc

dougxc:
As Gilles says, the only property we care about is `java.vm.version`. I cannot recall how `java.runtime.version` is constructed.

olpaw:
@zakkak considering the above I suggest: Use either something like https://github.com/oracle/graal/pull/2961#discussion_r521973500 or just `11.0.8.0.2+1-LTS` (without a specific `jvmci` or whatever postfix) or if in doubt lets keep it as it is (return `null` for native-images). cc @christianwimmer

zakkak:
@olpaw while testing the above I noticed that https://github.com/oracle/graal/commit/ee25d6c9344b3b2bda0fb0b54968b1944afb105c fixed the issue that this PR tries to fix, but at the same time conflicts with your preference of not showing the `jvmci` suffix:

```
$ ./sdk/latest_graalvm_home/bin/native-image --version                                    
GraalVM Version 21.0.0-dev (Java Version 11.0.9+10-jvmci-20.3-b03)
```

So it looks like what's needed is something like:
```diff
diff --git a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
index 5ed68d86c2e..2ba3de446a5 100644
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
@@ -60,7 +60,7 @@ class DefaultOptionHandler extends NativeImage.OptionHandler<NativeImage> {
         }
     }
 
-    private static String javaRuntimeVersion = System.getProperty("java.runtime.version");
+    private static String javaRuntimeVersion = System.getProperty("java.runtime.version").replaceAll("-jvmci.*", "");
 
     @Override
     public boolean consume(Queue<String> args) {
```

Note: not sure if that works for all labs-openJDK variants

olpaw:
> @olpaw while testing the above I noticed that [ee25d6c](https://github.com/oracle/graal/commit/ee25d6c9344b3b2bda0fb0b54968b1944afb105c) fixed the issue that this PR tries to fix, but at the same time conflicts with your preference of not showing the `jvmci` suffix:

See my previous comment: https://github.com/oracle/graal/pull/2961#discussion_r521973500

 > In the case of printing --version of the native-image-image it's ok to bake-in the value from the host system. (We want to show which version of GraalVM built the native-image-image and not what java.runtime.version runs the image.)

zakkak:
Uh that slipped away :facepalm:, I am really sorry! 
Since my main goal was to be able to get the info from `--version` I am closing this issue. If the need to access `java.runtime.version` at image runtime we can revisit (and make use of the discussions in this PR :) )

PR #2993, opened 2020-11-18
Disallow Random/SplittableRandom in the image heap and initialize relevant classes at runtime

pivovarit:
Disallowing `Random` and `SplittableRandom` in the image heap since it can result in undesired side-effects. 

The "failure" happens at run time but is caused by an unwanted build-time init of one class which results in unwanted caching.

----

Reproducer:
```
echo "public class Hello { public static void main(String[] args) { System.out.println(StrictMath.random()); }}" >> Hello.java
javac Hello.java
native-image Hello
```

and then the fun begins:
```
❯ ./hello  
0.7712305575721234
❯ ./hello  
0.7712305575721234
❯ ./hello  
0.7712305575721234
```


vjovanov:
@pivovarit thanks for fixing this! I am wondering does the failure happens at build-time or at runtime?

I think we should really detect this case at build-time.

vjovanov:
Could you please sign the OCA and we can merge?

pivovarit:
The "failure" happens at run time.

Reproducer:
```
echo "public class Hello { public static void main(String[] args) { System.out.println(StrictMath.random()); }}" >> Hello.java
javac Hello.java
native-image Hello
```

and then the fun begins:
```
❯ ./hello  
0.7712305575721234
❯ ./hello  
0.7712305575721234
❯ ./hello  
0.7712305575721234
```

Just signed the agreement.

vjovanov:
I can't think of case where one would like to have a `Random` initialized at build-time, but I can think of a million cases where one dosn't want to do that. 

Can we add a check to [DisallowedImageHeapObjects](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/image/DisallowedImageHeapObjects.java#L57) that verifies that RandomNumbers don't end up in the heap?

@christianwimmer @cstancu do you see a reason not to disallow `Random` in the image heap?

pivovarit:
The only reason I could think of was demoing the unwanted side-effects of build-time initialization.

As soon as I get a green light, I'll add it to disallowed objects.

christianwimmer:
Thanks for detecting this issue.

And yes, we should disallow instances of `Random` and `SplittableRandom` in the image heap. Better to fail at image build time than to have non-random values at run time.

pivovarit:
Well, that's unfortunate: https://github.com/openjdk/jdk/blob/65393a093c6a43a0bd8ff9f79c29c3bc5756104f/src/java.base/share/classes/sun/net/www/protocol/http/DigestAuthentication.java#L104-L109

Will take care of that separately.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/DigestAuthenticationFeature.java
```diff
@@ -0,0 +1,41 @@
...
+package com.oracle.svm.core.jdk;
+
+import org.graalvm.nativeimage.ImageSingletons;
+import org.graalvm.nativeimage.hosted.Feature;
+import org.graalvm.nativeimage.impl.RuntimeClassInitializationSupport;
+
+import com.oracle.svm.core.annotate.AutomaticFeature;
+
+@AutomaticFeature
+public class DigestAuthenticationFeature implements Feature {
```
vjovanov, line 34: This can go to `JDKInitializationFeature`. I think it is good to keep them together.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/KrbServiceLocatorFeature.java
```diff
@@ -0,0 +1,40 @@
...
+package com.oracle.svm.core.jdk;
+
+import org.graalvm.nativeimage.ImageSingletons;
+import org.graalvm.nativeimage.hosted.Feature;
+import org.graalvm.nativeimage.impl.RuntimeClassInitializationSupport;
+
+import com.oracle.svm.core.annotate.AutomaticFeature;
+
+@AutomaticFeature
+public class KrbServiceLocatorFeature implements Feature {
```
vjovanov, line 34: Same here, it can go to 'JDKInitializationFeature'.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/DigestAuthenticationFeature.java
pivovarit: Sure, will update soon

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/RandomNumbersFeature.java
```diff
@@ -40,5 +40,6 @@ public void duringSetup(DuringSetupAccess access) {
          * values properly. Otherwise the numbers generated will be fixed for each generated image.
```
vjovanov, line 40: I think this one can be fused with the `JDKInitializationFeature`

vjovanov:
Can't wait to run tests to see what other bugs we found :)

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/image/DisallowedImageHeapObjects.java
```diff
@@ -55,6 +58,13 @@
     }
 
     public static void check(Object obj, DisallowedObjectReporter reporter) {
+        /* Random/SplittableRandom can not be in the image heap. */
+        if (((obj instanceof Random) && !(obj instanceof ThreadLocalRandom)) || obj instanceof SplittableRandom) {
```
vjovanov, line 62: Thoughtfull--thanks!

vjovanov (approved):
LGTM. Running the tests now.

pivovarit:
OCA signed. Got a confirmation from Dalibor

pivovarit:
@vjovanov any results yet?

vjovanov:
Yes, it has discovered a few issues in the internal tests. I will merge this until the end of the year. Is that OK?

pivovarit:
Sure!

zakkak:
@vjovanov any chance we could get this backported to 20.3.1 as well (if ready on time)?

vjovanov:
Will merge it in the next day or two (due to failing tests in downstream projects). Will make a backport PR without the strict check to 23.1. Hope it makes it.

zakkak:
This has been merged in https://github.com/oracle/graal/commit/a975ef583fb2096ff29e50f8005af07064042440 and backported to 20.3.1 in https://github.com/oracle/graal/commit/19298b728629dd59e0142f4f5888811deb9b5bd9

jas88:
> I can't think of case where one would like to have a `Random` initialized at build-time, but I can think of a million cases where one dosn't want to do that.

I happened to come across this thread working on exactly such a case - I'm trying to build a text parsing tool which uses a library which happens to use a SplittableRandom somewhere in its internal data structures. No interest at all in making the values non-deterministic across runs - indeed probably slightly preferably not to - but very interested in being able to use the library from compiled code, and pre-building the lookup table rather than re-doing it each startup. Would it be possible to suppress this check for such cases please? --DisableRandomChecks or similar, if there isn't already an analogous option somewhere.

pivovarit:
@jas88 why would you ever want to do that? this is also how standard JVM behaves by default

jas88:
@pivovarit When did the standard JVM start precompiling to native executables at all, let alone with filtering whether or not you've used a pseudo-random number? What you described at the start of this thread as "undesired side-effects" is entirely harmless (arguably beneficial in some ways), unlike the compilation failure I'm dealing with as a result of this patch.

If the behaviour involved _Secure_Random (which implies the caller actually needs something genuinely non-predictable), or a situation where you genuinely want non-deterministic "randomness", fair enough, but I for one want code that actually compiles and works rather than a spurious error message.

Presumably the best workaround for my case in the mean time is to pin to building with Graal 21.3.0?

pivovarit:
Are you sure your code is not compiling due to this patch? At the time of implementation, it was just changing the runtime behaviour and not rejecting anything that uses SplittableRandom (and any other Random)

jas88:
I was getting a compilation error: `Error: Detected an instance of Random/SplittableRandom class in the image heap. Instances created during image generation have cached seed values and don't behave as expected.  To see how this object got instantiated use --trace-object-instantiation=java.util.Random. The object was probably created by a class initializer and is reachable from a static field. You can request class initialization at image runtime by using the option --initialize-at-run-time=<class-name>. Or you can write your own initialization methods and call them explicitly from your main entry point.`

IMO that should be a _warning_ rather than an error?

Reverting to 20.3.0 gets past this build stage at least, but fails in compiling some glue code in PosixDirectives.c which seems unrelated. I'll try building a newer GraalVM from source with this check removed at some point.

PR #3015, opened 2020-11-23
Quote arguments properly when storing them to an arguments file 

### LINKED ISSUE #3003: -H:IncludeResources stopped working in GraalVM 20.3.0
**Describe the issue**

`-H:IncludeResources` used to be a common way in GraalVM 20.2.0's `native-image` to include native resources. It does not seem to work anymore in 20.3.0

I have created a simple reproducer project in https://github.com/ppalaga/include-resources

```
git clone git@github.com:ppalaga/include-resources.git
cd include-resources

# Test with 20.2.0
export JAVA_HOME=~/bin/graalvm/graalvm-ce-java11-20.2.0

$JAVA_HOME/bin/java -version
openjdk version "11.0.8" 2020-07-14
OpenJDK Runtime Environment GraalVM CE 20.2.0 (build 11.0.8+10-jvmci-20.2-b03)
OpenJDK 64-Bit Server VM GraalVM CE 20.2.0 (build 11.0.8+10-jvmci-20.2-b03, mixed mode, sharing)

mvn clean package

./app/target/app # the output is as expected, both resources are there in the image
app/app-resource.txt: Hello from the app
module/module-resource.txt: Hello from the module

# Now test with 20.3.0
export JAVA_HOME=~/bin/graalvm/graalvm-ce-java11-20.3.0

$JAVA_HOME/bin/java -version
openjdk version "11.0.9" 2020-10-20
OpenJDK Runtime Environment GraalVM CE 20.3.0 (build 11.0.9+10-jvmci-20.3-b06)
OpenJDK 64-Bit Server VM GraalVM CE 20.3.0 (build 11.0.9+10-jvmci-20.3-b06, mixed mode, sharing)

mvn clean package

./app/target/app # the output is not OK, none of the resources is there in the image
app/app-resource.txt not available
module/module-resource.txt not available
```

 - GraalVM version: see above
 - JDK major version: see above
 - OS: Fedora 31
 - Architecture: AMD64

ppalaga:
Fixes #3003

I have a couple of tests for the new implementation of `NativeImage.quoteFileArg()` and I wonder where could I put them?

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -67,8 +67,6 @@ import java.util.function.Supplier;
 import java.util.jar.Attributes;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.graalvm.compiler.options.OptionKey;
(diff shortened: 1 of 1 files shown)


ppalaga:
I have sent the signed OCA yesterday.

ppalaga:
I have confirmed with @robilad that my contribution goes under the blanket Red Hat agreement.

ppalaga:
I still wonder where I should put the tests for `NativeImage.quoteFileArg()`?

gradinac:
Hey @ppalaga! Thank you very much for this contribution!
After looking at the new and previous changes, @olpaw and I had a discussion on how to best move forward on this. In order to reduce the complexity of this part of native-image, we've decided on taking a completely different approach to argument passing from the native-image driver to the builder JVM.
The new approach has been integrated in: https://github.com/oracle/graal/commit/e54f5cfb0dcc21c777bafec9bcba02be2208afd5
In the new system, we pass the JVM arguments through the command line, while passing the native-image arguments through a custom file. This greatly reduces the complexity of this part of Native Image, while also working for Java 8.
Sorry for putting you through the trouble of writing this PR! However, it has been very useful in helping us decide how to tackle this.
Please let me know if the new changes work for you, if you hit any errors or if you have any concerns/questions about the new implementation.

ppalaga:
No problem, @gradinac. Having a reliable args passing on Java 8 is certainly a valid concern.

Barely looking at e54f5cfb0dcc21c777bafec9bcba02be2208afd5 , I wonder whether passing multiline argument values would work. 

Here is an example: Have a simple class

```
import java.io.IOException;

public class App {
    static final String myProperty;
    static {
        myProperty = System.getProperty("my.property");
        System.out.println("myProperty = " + myProperty);
    }
    public static void main(String[] args) throws IOException {
        System.out.println("my.property = " + myProperty);
    }
}

```

Compile it with

```
javac App.java
native-image App.class --no-server --no-fallback --native-image-info --verbose --initialize-at-build-time=App "-Dmy.property=foo
bar" -H:Class=App -H:Name=app
```

Note the newline in `-Dmy.property=...` 

Would the compilation work properly and would the `app` print the following?

```
my.property = foo
bar
```

ppalaga:
Passing `-Dmy.property=...` that contains a newline works because `-D*` options do not seem to go via the args file. But I was able to construct a different example with `-H:FallbackExecutorJavaArg` that does not work. Let me file a new issue.

ppalaga:
I have filed https://github.com/oracle/graal/issues/3033

PR #3048, opened 2020-12-05
allow serialization config to contain java source style syntax classnames

kkriske:
Similar to how jni and reflection config files allow references like `org.example.ClassName[]` instead of `[org.example.ClassName;`, this should also be allowed for the serialization config.
The added code is copied from `com.oracle.svm.hosted.config.ReflectionRegistryAdapter#resolveTypeResult` which is where these cases for reflection/jni are handled.


olpaw (approved):

olpaw:
@kkriske Looks good. To get this faster to master I cherry-picked your commit onto another related PR I'm currently working on. See https://github.com/oracle/graal/pull/3060

olpaw:
Your commit made it on master: https://github.com/oracle/graal/commit/9e003860bdacc5a84333e9c0668a89986ae0056f

PR #3049, opened 2020-12-05
add lazy property initialization api to SystemPropertiesSupport

kkriske:
From custom features, it is possible to set default values of system properties using:
```java
SystemPropertiesSupport systemProperties = ImageSingletons.lookup(SystemPropertiesSupport.class);
systemProperties.initializeProperty("property.name", "property.value");
```

This suggestion adds a way to add lazily computed properties as well, so these don't run at startup time.
One possible reason to want this (like in my case), is if the custom property relies on one of the lazy property values, such as `user.dir`.

I'm actually not sure if this way of adding custom properties is supposed to be public api, as the question to add custom properties at build-time has risen a few times, but was never answered with a solution in the line of getting the SystemPropertiesSupport object.

This may slightly interfere with the changes of https://github.com/oracle/graal/pull/2834.


kkriske:
This needs some more work to properly provide the required feature API. But before I do, could I get some feedback on whether this is something we want to support or not?

kkriske:
closing as this is a can of worms not worth opening just yet, more info: https://github.com/oracle/graal/issues/5557#issuecomment-1335069305

PR #3055, opened 2020-12-10
ServiceLoaderFeature correctly supports comma-separated options

kkriske:
The service exclusion options documentation says it supports comma-separated arguments but it does not.

example command:
```
native-image \
    -H:ServiceLoaderFeatureExcludeServices=package1.Service1,package2.Service2 \
    -H:ServiceLoaderFeatureExcludeServices=package3.Service3 \
    -H:ServiceLoaderFeatureExcludeServiceProviders=package1.impl.ServiceImpl1,package2.impl.ServiceImpl2 \
    -H:ServiceLoaderFeatureExcludeServiceProviders=package3.impl.ServiceImpl3
```

This results in the exclusion sets to contain the entries without splitting the comma-separated values:
```
SERVICES_TO_SKIP = ["package1.Service1,package2.Service2", "package3.Service3"]
SERVICE_PROVIDERS_TO_SKIP = ["package1.impl.ServiceImpl1,package2.impl.ServiceImpl2", "package3.impl.ServiceImpl3"]
```

while they should be:
```
SERVICES_TO_SKIP = ["package1.Service1", "package2.Service2", "package3.Service3"]
SERVICE_PROVIDERS_TO_SKIP = ["package1.impl.ServiceImpl1", "package2.impl.ServiceImpl2", "package3.impl.ServiceImpl3"]
```

This has an easy workaround by just specifying the option multiple times with only a single class reference per option, but the comma-separated use-case should work as documented.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ServiceLoaderFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ServiceLoaderFeature.java
@@ -33,7 +33,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


kkriske:
I assume this was fixed by https://github.com/oracle/graal/commit/cb793452f5241c9bc3101e610dc1ad2d5917b4e0 and can be closed?

kkriske:
With the discovery of #3500, it is reasonable to assume this is also still an issue. I however did not verify this.

PR #3058, opened 2020-12-11
JNI - skip searching for hiding subclasses for constructors

kkriske:
`JNIAccessibleMethod#finishBeforeCompilation` searches for hiding subclasses for registered constructors/class initializers/methods, but only methods is relevant. The `anyMatchIgnoreReturnType` predicate checks against `sub.getDeclaredMethods` so it won't ever match with `<init>` or `<clinit>` anyway, so we can exit early.

While somewhat unrelated, this does partially fix https://github.com/gluonhq/client-maven-plugin/issues/200 as the issue there is a class that's reachable through service-loading that simply can't be loaded, so the `getDeclaredMethods` call fails. But adding `Object#getClass` for JNI will fail just as hard there.


loicottet:
This was already fixed in the code base, the check is now in `JNIAccessFeature.anyMethodMatchesIgnoreReturnType`, which only considers methods.

Issue #2745, opened 2020-08-10
GraalVM native image kerberos authentication issue: failed to store/retrieve AccessControlContext to/from stack  


zcui001:
GraalVM and environment:
* OS: Linux
* GraalVM version 20.1.0 (both CE and EE)
* JDK:  1.8

Issue:
The same java application code for kerberos authentication works fine in JVM, but failed to get GSSCredential in native image.  We noticed JVM and native image have different behavior when handling AccessControlContext. 

Steps to reproduce:
1) On Linux, run 'kinit -f' to retrieve valid kerberos ticket and store to the cache. 
2) build and run the blow java code:

import javax.security.auth.Subject;
import javax.security.auth.SubjectDomainCombiner;
import javax.security.auth.kerberos.KerberosTicket;
import javax.security.auth.login.AppConfigurationEntry;
import javax.security.auth.login.Configuration;
import javax.security.auth.login.LoginContext;
import javax.security.auth.login.LoginException;
import java.security.*;
import java.util.*;

    public static void kerberosTest() throws PrivilegedActionException, LoginException {

        class LoginConfiguration extends Configuration
        {
            private AppConfigurationEntry entry;
            private LoginConfiguration(Map<String, String> options)
            {
                entry = new AppConfigurationEntry(
                        "com.sun.security.auth.module.Krb5LoginModule",
                        AppConfigurationEntry.LoginModuleControlFlag.REQUIRED,
                        new HashMap<>(options)
                );
            }

            @Override
            public AppConfigurationEntry[] getAppConfigurationEntry(String name)
            {
                return new AppConfigurationEntry[] {entry};
            }
        }

        //Kerberos login and get subject
        Subject subject = null;
        Map<String, String> options = new HashMap<>();
        options.put("doNotPrompt", "true");
        options.put("useTicketCache", "true");
        Configuration configuration = new LoginConfiguration(options);
        LoginContext loginContext = new LoginContext("", null, null, configuration);
        loginContext.login();
        subject = loginContext.getSubject();

        Subject.doAs(subject, new PrivilegedExceptionAction<Boolean>() {
            @Override
            public Boolean run() throws Exception
            {
                final AccessControlContext acc = AccessController.getContext();

                DomainCombiner dc = acc.getDomainCombiner();

                //run the code in JVM will retrieve valid dc but native image will return null for dc
                if (dc == null ) {
                    System.out.println("dc is null");
                    return Boolean.FALSE;
                }
                System.out.println("dc is NOT null");

[...]

PR #3065, opened 2020-12-14
Implement getStackAccessControlContext

lazar-mitrovic:
Proposed fix for kerberos authentication github issue #2745


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
```diff
@@ -135,7 +142,10 @@
     @Substitute
     private static <T> T doPrivilegedWithCombiner(PrivilegedExceptionAction<T> action, AccessControlContext context, Permission... perms) throws Throwable {
         try {
-            return action.run();
+            AccessControllerUtil.accStack.get().push(context);
+            T result = action.run();
```
vjovanov, line 146: What happens when this throws an exception, we could make an infinite stack? I think we should us try with resources to make sure this is closed.

PR #3068, opened 2020-12-15
make awt invocation plugin conditional on linux os

### LINKED ISSUE #3064: Headless property is baked in, can no longer run heady
**Describe the issue**
The awt invocation plugin causes the headless property to be stored at build-time, which means you can no longer switch between headless and heady execution at runtime. Since on windows, `-Djava.awt.headless=true` is set by native-image itself, this breaks any heady application.

The plugin seems to be directed at unix environments, as that's also the only awt support that's already added, so it might make sense to only conditionally add the plugin if we're targetting unix.

Additionally I think it's not preferable to have to decide at build-time if the application will run headless or heady.

The invocation plugin this is about:
https://github.com/oracle/graal/blob/6fc5aaf330c09dc78bec4c662ede73c37364e17f/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java#L1033-L1048

- run any natively compiled windows app with `-Djava.awt.headless=false` and it will still be headless

 - GraalVM version: **72439a3924dfe0391764a3f326dfb7da30cda8b4** (as good as up to date with master)
 - JDK major version: **11**
 - OS: **Windows 10**
 - Architecture: **x64**

kkriske:
fixes #3064


pejovica (approved):
@kkriske Thanks for the fix! Should be merged soon.

Issue #3018, opened 2020-11-23
Add JDK Flight Recorder support for Java native images


adinn:
# Synopsis

This issue proposes to add JDK Flight Recorder (JFR) support for Java native images for monitoring and performance profiling of applications. 

A prototype implementation is in progress, and work is ongoing to finalize the initial feature set for contribution to Graal CE. We welcome feedback from Oracle’s GraalVM team and the GraalVM community on both this proposal and the prototype implementation (which we will soon publish as a draft PR).

The prototype preserves and reuses existing Java APIs (`jdk.jfr.*`) together with a new Java-based implementation of the underlying system (currently under `com.oracle.svm.core.jdk.jfr.*`). This implementation closely resembles the native implementation in OpenJDK HotSpot (`hotspot/share/jfr`). Event instrumentation is forced during the analysis stage and reused in the resulting native image.

## Proposed Initial Additions
 - Provide compile-time configuration to include or exclude the JFR infrastructure
 - Support existing OpenJDK API to start, stop and dump JFR v2.1 compliant recording files (`.jfr`) during native image run (`jdk.jfr.Recording` and related classes)
 - Allow command-line options to configure recordings on start-up, similar to the options in OpenJDK
 - Support build-time inclusion of application-defined events derived from `jdk.jfr.Event` and related classes
 - Emit OpenJDK Java specific events (Socket/File read/write, ExceptionThrown, etc.)

## Proposed Long-Term Additions
 - Emit SubstrateVM specific events for memory allocation, garbage collection and other SubstrateVM subsystems
 - Support JFR crash dumps
 - Support leak profiling
 - Add Thread and Stack Trace data for events
 - Provide remote API to start, stop and dump recordings

## Expected Benefits
 - Native image execution monitoring and tracing
 - Native image performance profiling via JFR file analysis
 - Existing Java users and applications can migrate to native image with JFR available

## Limitations
 - No dynamic events created at runtime

thegreystone:
Is support for the jdk.ExecutionSample event, or a true CPU profiling event, part of this effort?

jiekang:
@thegreystone Support for `jdk.ExecutionSample` or similar is part of the overall effort but for a currently unknown future date.

jerboaa:
The WIP PR for this is here: https://github.com/oracle/graal/pull/3070

jiekang:
The initial support has made it into upstream via https://github.com/oracle/graal/pull/3444

[...]

PR #3070, opened 2020-12-15
[WIP] Add JDK Flight Recorder support for Java native images

jiekang:
This PR adds infrastructure to SubstrateVM supporting JDK Flight Recorder for Java native images (https://github.com/oracle/graal/issues/3018). This is a work in progress and the intention is to welcome feedback from Oracle’s GraalVM team and the GraalVM community during the development process leading up to initial integration.

## Completed Features
* Provide compile-time configuration to include or exclude the JFR infrastructure (`-H:+FlightRecorder`)
* Support existing OpenJDK API to start, stop and dump JFR v2.1 compliant recording files (`.jfr`) during native image run (`jdk.jfr.Recording` and related classes)
* Support build-time inclusion of application-defined events derived from `jdk.jfr.Event` and related classes

## In Progress Features
* Emit OpenJDK Java specific events (Socket/File read/write, ExceptionThrown, etc.)
* Allow command-line options to configure recordings on start-up, similar to the options in OpenJDK

## Other Issues in progress

Along with the major features above, many minor issues that will be addressed are tracked here: https://github.com/rh-jmc-team/graal/issues Issues labelled “jfr-initial-release” are targeted for eventual completion in this PR. This issue tracker is temporary and used to cleanly separate work items for this integration in a publicly accessible location. Note, as this is WIP, new issues will surely appear and be tracked there as well. Once this PR is accepted, we will follow upstream practices for all future work. As well, more long-term additions are described in the original issue: https://github.com/oracle/graal/issues/3018

## Testing

This implementation has been tested against native images produced from sources in this repository: https://github.com/rh-jmc-team/jfr-tests At this time, tests are manually run. We intend to add automated test infrastructure to compile test sources into native images, run them, and verify the resulting JFR file. Tests against other applications have been and will continue to be run as well.

As an example from the jfr-tests repository, TestConcurrentEvents starts a recording to disk, creates 8 threads that emit 1024*1024 events each, stops the recording and prints the location of the produced .jfr file. See `substratevm/JFR.md` in this branch for instructions on building with FlightRecorder enabled.

## Contributors:

Andrew Dinn <adinn@redhat.com>
Jie Kang <jkang@redhat.com>
Roman Kennke <rkennke@redhat.com>
Josh Matsuoka <jmatsuok@redhat.com>
Tako Schotanus <tschotan@redhat.com>
Simon Tooke <stooke@redhat.com>

### DIFF
--- a/substratevm/JFR.md
+++ b/substratevm/JFR.md
@@ -1,0 +1,43 @@
+## JFR Implementation Notes
+
+Implementation is based off of jdk/jdk commit rev
+
+d2816784605b
+
+## JDK Matrix
+
+This branch builds with:
+labsjdk-ce-11.0.9-jvmci-20.3-b06
+
+Tarballs can be found at:
+https://github.com/graalvm/labs-openjdk-11/releases
+
+For example:
+https://github.com/graalvm/labs-openjdk-11/releases/download/jvmci-20.3-b06/labsjdk-ce-11.0.9+10-jvmci-20.3-b06-linux-amd64.tar.gz
+
+## Testing the JFR code
+
+Clone the graalvm-jfr-tests repo somewhere to be used in the -cp argument. Use a JDK 11 to compile into class files.
+
+```
+javac /path/to/graalvm-jfr-tests/flat/EventCommit.java
+
+
+mx clean
+mx build
+mx native-image --allow-incomplete-classpath "-J-XX:FlightRecorderOptions=retransform=false" --no-fallback -ea -cp /path/to/graalvm-jfr-tests/flat/ EventCommit
+```
+
+See `/tmp` for a `my-recording...` jfr file
+
+## Logging
+
+The JFR implementation supports log levels via runtime flag:
+```
+-XX:FlightRecorderLogging=<value>
+```
+
+Values include:
+```
+trace, debug, info, warning, error
+```
--- a/substratevm/JFR.md
+++ b/substratevm/JFR.md
@@ -15,24 +15,26 @@ https://github.com/graalvm/labs-openjdk-11/releases
 For example:
 https://github.com/graalvm/labs-openjdk-11/releases/download/jvmci-20.3-b06/labsjdk-ce-11.0.9+10-jvmci-20.3-b06-linux-amd64.tar.gz
 
-## Testing the JFR code
+## Building a native image with JFR
 
-Clone the graalvm-jfr-tests repo somewhere to be used in the -cp argument. Use a JDK 11 to compile into class files.
+The JFR implementation is included at compile time via the flag:
+`-H:+FlightRecorder`
 
+## Running tests from the jfr-tests repo
+Clone the https://github.com/rh-jmc-team/jfr-tests repo to be used in the -cp argument. Follow its readme to compile class files via Maven.
+
+Then compile the classes to a native image, for example:
 ```
-javac /path/to/graalvm-jfr-tests/flat/EventCommit.java
-
-
-mx clean
 mx build
-mx native-image --allow-incomplete-classpath "-J-XX:FlightRecorderOptions=retransform=false" --no-fallback -ea -cp /path/to/graalvm-jfr-tests/flat/ EventCommit
+mx native-image -H:+FlightRecorder --no-fallback -cp /path/to/jfr-tests/target/classes com.redhat.jfr.tests.event.TestConcurrentEvents
+./com.redhat.jfr.tests.event.testconcurrentevents
 ```
 
-See `/tmp` for a `my-recording...` jfr file
+See `/tmp` for the jfr file
 
 ## Logging
 
-The JFR implementation supports log levels via runtime flag:
+The JFR implementation supports log levels via the runtime flag:
 ```
 -XX:FlightRecorderLogging=<value>
 ```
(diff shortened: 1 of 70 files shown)


jerboaa:
Note that @jiekang is covered by the Red Hat OCA. He works for Red Hat (as do other contributors).

thegreystone:
Nice one guys! :)

christianwimmer:
Thanks for the PR, we will have a look at it. Due to the imminent winter break, holidays, vacation, ... a realistic ETA for the first round of comments on our side is middle of January.

I'm also glad that you explicitly mark the PR as "WIP". Still, it is a lot of work that you already did on it (about 9000 lines of code added). I do not think it is a sustainable approach to make such large contributions without discussing the overall architecture beforehand. Ideally, the architecture discussions should start before writing the first line of code. Slack and Zoom discussions upfront are far more time effective than after-the-fact reviews.

adinn:
Hi Christian,

Yes, this is a work in progress -- although I do feel we have made a lot of progress to arrive where we are now. I understand why you might have preferred us to discuss this PR in advance. However, we had two problems doing that before now. First, we felt we had to have something tangible to discuss. Second, we wanted to be able to enter any discussion from a position where we understood enough of the complexities of how to implement a solution to make that discussion realistic and not just end up offloading onto you work that we had chosen to undertake.

We have released this PR with a basic set of useful, fully working features, extra features being left for further work. So, this really constitutes a 'pilot' implementation i.e. one that tests the feasibility of a partial solution right through to completion and at the same time tries to establish an acceptable path forward to a full solution. We have no belief that this implementation is the only way or the right way to do this and certainly have no expectation that you will simply rubber-stamp our work.

That explains why we wanted to be sure our approach really did work for those basic cases before asking you to look at it and help us determine if there is a better approach. In getting to this point we have uncovered and resolved a lot of problems shoe-horning JFR into Graal native, many to do with the change from dynamic to static (build/runtime) operation, just as many to do with the need to uncouple the close linkage between JFR and Hotspot and identify/recreate an equivalent in GraalVM. We have also understood a lot about the problems we have not yet addressed. I really don't believe we could have had a useful discussion with you without working this solution through as far as we have.

So, please don't feel we are trying to push this through a hasty review with no desire to listen to your opinion or follow your advice. Of course, we want to get a working version released as soon as possible in order for our users to be able to benefit from having a monitoring solution. However, we also want to get it right, so that we can maintain the functionality we have implemented so far and add the missing capabilities that we also need. We very much welcome your feedback and hope you can help us steer this towards a better design and better implementation.

christianhaeubl:
Thanks for the PR. So far, I only had a very brief look at the code but here a couple high-level comments that will probably have a large impact on your work in progress:
- The current implementation assumes that it will only be used in Java code that can allocate Java objects. However, JFR events (e.g., allocation, safepoint, and GC related events) may be triggered in places where Native Image must not allocate Java objects and can only call uninterruptible code. So, a lot of the core JFR infrastructure will have to be uninterruptible.
- Similar to HotSpot, every thread will need a Java and a native buffer. Otherwise, events such as safepoints or GCs would destroy the Java-level buffer.
- Please note that VM operations on Native Image can allocate and cause a GC. This is a major difference to HotSpot and may cause issues as a GC may trigger JFR events. Similar to that, I saw that a method that allocates a Java object but that has the following comment: "Do not safepoint here".
- A few things are too close to the HotSpot sources where this was ported from, e.g., most parts related to class (un)loading won't be needed for Native Image.

As @christianwimmer already mentioned, I will have a closer look at the PR in January.

christianwimmer:
Hi all, now that all the holidays are over we have some discussions and decisions to make.

We have also worked on some initial JFR support in the last months, and we have the approvals to make it public. @christianhaeubl is working on that, and will have that ready soon. Our work is also heavy work-in-progress. Some parts are more complete than this PR (like avoiding problems with GC and safepoints), some parts are less complete.

Our hope would be to combine both implementations and work together on a combined implementation. As as starting point, I suggest that we schedule a Zoom with all contributors. To find a time for that, and also to have more interactive control, we created a dedicated Slack channel in the GraalVM slack workspace: #native-image-jfr. This link should get you directly to it: https://graalvm.slack.com/archives/C01K81PSVLM

adinn:
@christianwimmer Hi Christian. Thanks for the update and enormous thanks to the Oracle team for deciding to open up your work and combine forces with the rest of the community to ensure JFR works in CE releases. Red Hat very much appreciates your offer of help and welcomes the benefit of your expertise in areas like GC and safepoints where our experience is still weak. Looking forward to the Zoom call and following up on this effort on Slack.

christianhaeubl (commented):
This is only a partial and not very detailed review. I skipped many classes where I think that large changes will be necessary anyways due to the differences between the HotSpot and the Native Image safepoint and VM thread handling.

Major issues:
- With the current implementation, it won't be possible to support many of the native events (i.e., the ones that are defined in the metadata.xml file).
- In Native Image, every thread can trigger JFR events. So, even if a thread is at a safepoint and executes a VM operation, it can still trigger JFR events (e.g., because it allocates a Java object). This is a major difference to HotSpot and breaks some of the assumptions that the HotSpot code has.

jiekang:
Closing as the work has been done in https://github.com/oracle/graal/pull/3444

PR #3073, opened 2020-12-16
Remove deprecated instantiation of wrapper classes for primitives

SwapnilGaikwad:
Use of constructors from wrapper classes fails while building jdk17 + graal.
The failures could be avoided by not treating warnings as errors. However, the getDefault() method uses new way (.valueOf) to instantiate wrapper classes, so this change would make instantiation consistent.

### DIFF
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
```diff
@@ -391,13 +390,13 @@ public long getFieldAddress(String name, String cppType) {
             // overhead when running with assertions enabled.
             T sentinel;
             if (type == Boolean.class) {
-                sentinel = type.cast(new Boolean(false));
```
dougxc, line 394: The constructors were deliberately chosen here to get object identity for use in the `value != sentinel` test down on line 406. Obviously this will not be possible once the constructors are removed for good but we have until then to think of a solution that does the equivalent of the identity based test but without relying on identity.

tkrodriguez: If the flag exists then it must have a non-null value for any of the primitive boxes, so we why can't we use null as their sentinel?  We really only need a sentinel value for Strings so we can handle null.

dougxc: We don't know if the flag exists if we're in the "expect not present" branch. The requirement for a non-null sentinel is due to [this](https://github.com/graalvm/graal-jvmci-8/blob/master/jvmci/jdk.vm.ci.hotspot/src/jdk/vm/ci/hotspot/HotSpotVMConfigAccess.java#L345). If we cannot rely on box identities here, then we'll have to do something different. I think this would do it:
```
diff --git a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
index d68c8a81786..12dabac8830 100644
--- a/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
+++ b/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfigAccess.java
@@ -380,29 +380,21 @@ public class GraalHotSpotVMConfigAccess {
         if (expectPresent) {
             return getFlag(name, type);
         }
+        // Expecting flag not to be present
         if (Assertions.assertionsEnabled()) {
             // There's more overhead for checking unexpectedly
             // present flag values due to the fact that a VM call
-            // not exposed by JVMCI is needed to determine whether
+            // not exposed by JVMCI (i.e.,
+            // jdk.vm.ci.hotspot.CompilerToVM.getFlagValue)
+            // is needed to determine whether
             // a flag value is available. As such, only pay the
             // overhead when running with assertions enabled.
-            T sentinel;
-            if (type == Boolean.class) {
-                sentinel = type.cast(new Boolean(false));
-            } else if (type == Byte.class) {
-                sentinel = type.cast(new Byte((byte) 123));
-            } else if (type == Integer.class) {
-                sentinel = type.cast(new Integer(1234567890));
-            } else if (type == Long.class) {
-                sentinel = type.cast(new Long(1234567890987654321L));
-            } else if (type == String.class) {
-                sentinel = type.cast(new String("1234567890987654321"));
-            } else {
-                throw new JVMCIError("Unsupported flag type: " + type.getName());
-            }
-            T value = access.getFlag(name, type, sentinel);
-            if (value != sentinel) {
+            try {
+                T value = access.getFlag(name, type, null);
+                // Flag value present -> fail
                 recordError(name, unexpected, String.valueOf(value));
+            } catch (JVMCIError e) {
+                // Flag value not present -> pass
             }
         }
         return access.getFlag(name, type, notPresent);
```

SwapnilGaikwad: Hi @dougxc, thanks for your suggestion. I have just updated the patch accordingly. I have mentioned you as Co-author, but feel free to put a separate patch if that's not fine with you.

dougxc: No need for attribution but it would be helpful if you could test that it's correct somehow.

dougxc:
Fixed by https://github.com/oracle/graal/commit/07e742f0e56e02aa4ca0309cf4f7b015d535700b.

PR #3076, opened 2020-12-17
Fix file cache crash.

manuelgustavor3:
As observed in our R3 tests, specialSrcRoots needs to be checked for emptiness, otherwise, an exception will be thrown.


fniephaus:
Hi @manuelgustavor3, looks like your PR didn't get any attention, apologies! Is this still an issue in GraalVM for JDK 17 or later? The SourceCache code has changed a bit and the problem may have already been fixed. Could you take a quick look?

PR #3079, opened 2020-12-17
windows awt support

kkriske:
This is a very crude first implementation of awt support for windows, the main purpose of this PR is sharing this with the GraalVM team for feedback and to discuss possible next steps.

The most interesting part of this PR is that it _should_ contain all reachability handlers to cover registration of JNI requirements of the awt library on windows. It does however **not** cover any other libraries such as lcms or mlib_image.

### current issues:
* for now, awt cannot be registered as a built-in library yet, as it isn't statically build yet (`JNI_OnLoad_awt` not found exception on linking)
* a lot of classes are requested for runtime initialization, under the assumption that they have to run the native initializers at runtime. (mostly called `initIDs`, which fetch and store method/field pointers). In other library features this seems to be handled by marking them for rerun instead of runtime but I am not sure of the implications hereof.

I would appreciate some information on these 2 points, especially why and if classes should be marked for rerun, and if awt (and related other libraries) can be build statically or if there are some things preventing this for now. (and if maybe I can be of any help there?)

With some supervision, I am willing to bring this to a useable state.


kkriske:
Looking further into the current setup of the unix support, if I'm understanding it correctly, the value of `java.awt.headless` has to be decided at build-time to be able to skip a bunch of JNI registrations and keep the image size small in case of a headless application.

Since some use cases require the headless property to be decided at runtime (if both modes have to be supported) I would suggest introducing a hosted option to switch between both options, probably with the default on build-time setting to keep the default image size small. (Which has the added issue that windows cannot have a `headless=false` property baked in for now, due to it being added by native-image itself).

pejovica:
@kkriske, thanks for the PR! Looks like it might fit in nicely with #3084, but I'll have to take a closer look, which probably won't be until early next year.

smallshen:
@kkriske @pejovica  I think the cause of `JNI_OnLoad_awt` is `DEF_STATIC_JNI_OnLoad`

https://github.com/graalvm/labs-openjdk-17/blob/aa3270e709c58bef32ca10fa826f9db7e8c76f42/src/java.base/share/native/libjava/jni_util.h#L373

https://github.com/graalvm/labs-openjdk-17/blob/aa3270e709c58bef32ca10fa826f9db7e8c76f42/src/java.base/share/native/libjava/jni_util.h#L362


Do you have similar link error like `long __cdecl JNI_OnLoad_awt(struct JavaVM_ *,void *)` ?
I'm confused why it produces `long __cdecl `, and the linker can't find it. 


Reproduce on Windows:
```java
import com.oracle.svm.core.annotate.AutomaticFeature;
import com.oracle.svm.core.jdk.NativeLibrarySupport;
import com.oracle.svm.core.jdk.PlatformNativeLibrarySupport;
import com.oracle.svm.hosted.FeatureImpl;
import com.oracle.svm.hosted.c.NativeLibraries;
import org.graalvm.nativeimage.hosted.Feature;

@AutomaticFeature
public class AWTFeature implements Feature {
    @Override
        NativeLibraries nativeLibraries = ((FeatureImpl.BeforeAnalysisAccessImpl) access).getNativeLibraries();

        nativeLibraries.addDynamicNonJniLibrary("Gdi32");
        nativeLibraries.addDynamicNonJniLibrary("Imm32");
        nativeLibraries.addDynamicNonJniLibrary("Winspool");
        nativeLibraries.addDynamicNonJniLibrary("Comdlg32");
        nativeLibraries.addDynamicNonJniLibrary("Comctl32");
        nativeLibraries.addDynamicNonJniLibrary("Shlwapi");
        nativeLibraries.addDynamicNonJniLibrary("Winmm");

        NativeLibrarySupport.singleton().preregisterUninitializedBuiltinLibrary("awt");
        nativeLibraries.addStaticJniLibrary("awt");

        PlatformNativeLibrarySupport.singleton().addBuiltinPkgNativePrefix("java_awt");

    }
}
```

`native-image.cmd "-Djava.awt.headless=false" --add-exports org.graalvm.nativeimage.builder/com.oracle.svm.hosted.c=ALL-UNNAMED --add-exports org.graalvm.nativeimage.builder/com.oracle.svm.hosted=ALL-UNNAMED  --add-exports org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk=ALL-UNNAMED -jar .\some-jar-with-awtcode.jar`

PR #2831, opened 2020-09-06
Allow root certificates to be configured at run time of native image

jaikiran:
The commit in this PR introduces the "option#2" support that has been noted and requested in https://github.com/oracle/graal/issues/1999.
More specifically, this commit now allows users to (optionally) specify a path to root certificates using the (standard) `javax.net.ssl.trustStore` system property at run time, while launching the native application.

Details about this change:

- This commit continues to embed the build time root certificates into the native image. This is intentional so as to not force users to always specify the root certificate location at run time.
- The commit introduces logic that's run at image run time which checks for a non-empty value for `javax.net.ssl.trustStore` system property. If it finds one, then instead of using the embedded certs, it goes ahead and uses the (existing) JRE infrastructure to parse and use the root certificates pointed to by value of that system property.
- Before this commit, there is a piece of code which initializes the `sun.security.util.UntrustedCertificates` at build time. That class internally loads a `lib/security/blacklisted.certs` file, which is relative to the Java installation used at build time of the image. This file contains the blacklisted certificates and is used to validate the "trusted" certificate. The commit in this PR (intentionally) continues to allow this logic to stay and thus continues to use the build time blacklist file for decision making at run time. IMO, this should be fine because unlike new certificates getting added to a trust store, a blacklisted certificate rarely (is it ever?) is considered to be good again. So carrying over this information to the run time, IMO, is a good thing. However, there isn't a way to add new certificates to this blacklist at runtime and I don't think that's something that should be in the scope of this specific change.
- There's this (existing) code in this `TrustStoreManagerFeature` which I don't fully understand:
```
RuntimeClassInitialization.initializeAtBuildTime(org.jcp.xml.dsig.internal.dom.XMLDSigRI.class);
```
I have given a short glance at that `org.jcp.xml.dsig.internal.dom.XMLDSigRI` to see if the change in this commit impacts in any way, that specific class and whether anything should be done to handle it. With my limited knowledge of that class, I haven't been able to decide if anything is needed for that one. So please do let me know if something needs to be done there.

- Some javadoc and comment changes have been done to explain this change

Finally, I've done some manual testing with this change. I used the following Java class to build a native image from:

```
import java.net.URL;
import java.io.InputStream;
import java.net.*;
import java.io.*;
import java.lang.reflect.*;

public class HelloWorld {
	public static void main(String[] args) throws Exception { 
		final URL targetURL = new URL("https://google.com/");
        HttpURLConnection responseConnection = (HttpURLConnection) targetURL.openConnection();
[...]

Issue #1999, opened 2019-12-16
Load root certificates at image build time


christianwimmer:
The JDK ships a file with root certificates: lib/security/cacerts. The JDK TrustStoreManager is responsible for loading. If no system property is set explicitly, then the cacerts file of the JDK is loaded. But that can be overridden using system properties.

We have a few options how to handle that in a native image:
1) Do not embed any root certificates into a native image. That means that the application needs to be shipped with a cacerts file, or the system properties must be used to locate certificates already installed on the system. I think that is a bad option though because it is difficult for users, and no certificates are available out-of-the-box without additional configuration.
2) Embed the root certificates into the native image, but still at run time look at the system properties and allow certificates from an alternative location to be loaded.
3) Embed the root certificates into the native image, and do not allow any different certificates to be loaded at run time.

The difference between option 2) and 3) from a security and usability perspective: In option 2) we treat security as something that can be configured at image run time. In option 3), we consider security as something that is fixed at image build time and therefore cannot be modified (explicitly or accidentally) at run time.

We believe that "immutable security" is a valuable feature of Native Image and therefore implement solution 3)

dmlloyd:
I would prefer solution "1" or "2", if anything.  I'm not sure that "immutable security" is a feature per se; it's more of a consequence.

sberyozkin:
@christianwimmer It is not clear if 3) allows for the custom certificates be included at the build time ? 

Option 2) looks the best though, can you please clarify why 3) is preferred to letting the native image users customize the default certificates at run time ?

christianwimmer:
@sberyozkin Yes, custom certificates can be included at image build time. All system properties such as `javax.net.ssl.trustStore` can be specified at image buid time.

@dmlloyd "immutable security" is a guiding principle of Native Image. For example, not allowing dynamic class loading at run time makes the code immutable at run time, allowing to reason about security at image build time. Do you have any technical arguments or use cases why solution 3) would not work?

dmlloyd:
> @dmlloyd "immutable security" is a guiding principle of Native Image. For example, not allowing dynamic class loading at run time makes the code immutable at run time, allowing to reason about security at image build time.  Do you have any technical arguments or use cases why solution 3) would not work?

No, I think it will work, but I suspect we'll end up having to add a Quarkus workaround to allow the trust store to be rebuilt from cacerts at run time, which is not ideal.

sberyozkin:
@christianwimmer, thanks...
[...]

PR #3091, opened 2020-12-27
Allow root certificates to be configured at run time of native image

matthyx:
The commit in this PR introduces the "option#2" support that has been noted and requested in #1999.
This is a continuation of https://github.com/oracle/graal/pull/2831 upon authorization from @jaikiran 
For `getTrustedKeyStore` I found it easier to integrate inside `loadKeyStore` which is lifted from jdk16 and modified to include the embedded one.
For `getTrustedCerts` I simply merge sets.
/cc @christianwimmer 
Please be indulgent as this is my first try with all these annotations...

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_sun_security_ssl_TrustStoreManager.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_sun_security_ssl_TrustStoreManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -24,33 +24,46 @@
  */
 package com.oracle.svm.core.jdk;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.security.AccessController;
 import java.security.KeyStore;
 import java.security.cert.X509Certificate;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashSet;
 import java.util.Set;
 
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.hosted.Feature;
 import org.graalvm.nativeimage.hosted.RuntimeClassInitialization;
 
+import com.oracle.svm.core.annotate.Alias;
 import com.oracle.svm.core.annotate.AutomaticFeature;
-import com.oracle.svm.core.annotate.Delete;
+import com.oracle.svm.core.annotate.RecomputeFieldValue;
 import com.oracle.svm.core.annotate.Substitute;
 import com.oracle.svm.core.annotate.TargetClass;
 import com.oracle.svm.core.util.VMError;
 import com.oracle.svm.util.ReflectionUtil;
+// Checkstyle: stop
+import sun.security.action.OpenFileInputStreamAction;
+import sun.security.ssl.SSLLogger;
+// Checkstyle: resume
 
 /**
- * Native image uses the principle of "immutable security" for the root certificates: They are fixed
- * at image build time based on the the certificate configuration used for the image generator. This
- * avoids shipping a `cacerts` file or requiring to set a system property to set up root
- * certificates that are provided by the OS where the image runs.
- *
- * As a consequence, system properties such as `javax.net.ssl.trustStore` do not have an effect at
- * run time. They need to be provided at image build time.
- *
- * The implementation "freezes" the return values of TrustStoreManager managers by invoking them at
- * image build time (using reflection because the class is non-public) and returning the frozen
- * values using a substitution.
+ * Root certificates in native image are fixed/embedded into the image, at image build time, based
+ * on the the certificate configuration used for the image generator. This avoids shipping a
+ * `cacerts` file or requiring to set a system property to set up root certificates that are
+ * provided by the OS where the image runs.
+ * <p>
+ * However, users are allowed to override these root certificates at run time by setting the
+ * `javax.net.ssl.trustStore` system property to point to a file path containing the certificates.
+ * <p>
+ * For embedding the build time root certificates, the implementation "freezes" the return values of
+ * TrustStoreManager managers by invoking them at image build time (using reflection because the
+ * class is non-public) and returning the frozen values using a substitution, if at run time the
+ * `javax.net.ssl.trustStore` system property isn't set
  */
 @AutomaticFeature
 final class TrustStoreManagerFeature implements Feature {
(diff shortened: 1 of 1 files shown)


matthyx:

Please check with email matthias@matthiasbertschy.info or with my handle @matthyx
/cc @robilad

vsevel:
I am glad we are making progress on this. this is a very important use case for us.

matthyx:
@christianwimmer can you check this PR please?
@robilad can you check my OCA?

matthyx:
@christianwimmer ping

matthyx:
> @christianwimmer ping

@christianwimmer should I check this point with someone else from graal project?

vsevel:
hello @christianwimmer this feature is really valuable. any chance you could offer some advice on how to move it forward?
@jaikiran you have done some work on this subject. do you think @matthyx is on the right track?

matthyx:
hello @christianwimmer will you have time to review this PR? should I assign someone, I am not familiar with the process here...

matthyx:
@cstancu would someone have time to look at this PR please?

matthyx:
@sdeleuze could you help me getting this feature in a next version of GraalVM?

sdeleuze:
I have bring this one to @vjovanov attention.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_sun_security_ssl_TrustStoreManager.java
```diff
@@ -94,27 +109,95 @@ public void afterRegistration(AfterRegistrationAccess access) {
...
+        KeyStore embeddedKeystore = ImageSingletons.lookup(TrustStoreManagerSupport.class).trustedKeyStore;
+        if (!"NONE".equals(descriptor.storeName) &&
+                        descriptor.storeFile == null) {
+
+            // No file available, no KeyStore available.
+            if (SSLLogger.isOn && SSLLogger.isOn("trustmanager")) {
+                SSLLogger.fine("No available key store");
+            }
+
+            return embeddedKeystore;
```
christianwimmer, line 160: Without doing a full method diff, this seems to be the only change compared to the original implementation.
I don't really understand how that should work. The comment still says "no KeyStore available" but then something is returned?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_sun_security_ssl_TrustStoreManager.java
```diff
@@ -94,27 +109,95 @@ public void afterRegistration(AfterRegistrationAccess access) {
...
-        return ImageSingletons.lookup(TrustStoreManagerSupport.class).trustedKeyStore;
+        return Collections.unmodifiableSet(set);
     }
 }
 
-/*
- * The internal classes to describe and load root certificates must not be reachable at run time.
- */
-
-@Delete
```
christianwimmer, line 112: This class was deleted for a good reason. It has, e.g., the `java.home` directory of the image builder baked into the static fields `defaultStorePath` and `jsseDefaultStore`. I don't see anything in this PR that would modify usages of these fields

christianwimmer (commented):
Sorry for not responding to this PR for a long time.

Unfortunately, I do not see how the proposed changes work. This might be just a documentation issue - for substitutions of JDK code that is security critical like this we need a very detailed explanation how the modified code works, and in which cases the substitution needs to be updated.

Since I don't understand the solution yet, I commented on some things that seem to be the most obvious problems to me. So honestly I'm not too hopeful that the proposed approach can work.

Also, there are no tests for the new functionality to check that the hybrid of build time and run time configuration works.

matthyx:
@christianwimmer first of all thanks for having a look... it shows that I'm complete n00b on jvm programming (let alone native code).
The idea was to adapt the jks loading code from the traditional jre but combining it with certs that were embedded at build time.
Give me some time to prepare my follow ups, and bear with me as it looks like we're on for several ping pong exchanges.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_sun_security_ssl_TrustStoreManager.java
```diff
@@ -94,27 +109,95 @@ public void afterRegistration(AfterRegistrationAccess access) {
...
 final class Target_sun_security_ssl_TrustStoreManager {
 
+    @Alias @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.NewInstance, declClassName = TrustStoreManagerFeature.TRUST_STORE_MANAGER_CLASS_NAME +
+                    "$TrustAnchorManager") private static Target_sun_security_ssl_TrustStoreManager_TrustAnchorManager tam = null;
+
     @Substitute
     private static Set<X509Certificate> getTrustedCerts() throws Exception {
-        return ImageSingletons.lookup(TrustStoreManagerSupport.class).trustedCerts;
-    }
+        Set<X509Certificate> set = new HashSet<>(ImageSingletons.lookup(TrustStoreManagerSupport.class).trustedCerts);
```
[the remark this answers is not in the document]
matthyx: Well this could be debated, our use case is to include some corporate root CA at build time, because these are required to build and run in all our OpenShift clusters.
Then at run time we need to use some cluster specific certificates managed via cert manager to validate SSL services.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_sun_security_ssl_TrustStoreManager.java
matthyx: But the `TrustAnchorManager` needs a `TrustStoreDescriptor` in order to load the certificate... so maybe I should just invalidate the static fields?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_sun_security_ssl_TrustStoreManager.java
matthyx: Yup that's the only difference...

You're right, maybe I should change the SSLLogger message to say that we found no keystore at runtime and we fall back to the embedded one.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_sun_security_ssl_TrustStoreManager.java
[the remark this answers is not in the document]
matthyx: In that case it was obvious for me to merge both sets... it also makes it easy to handle when `tam.getTrustedCerts` returns an empty one.

matthyx:
@christianwimmer thanks a lot for your review comments, I have tried to address some and reply to the others.

Where should I document the "design" for this feature?

I will also see if I can make some CI tests for it, because at the moment I simply build a HelloWorld which doesn't cover truststore merging:
```public class HelloWorld {
        final URL targetURL = new URL("https://google.com/");
        HttpURLConnection responseConnection = (HttpURLConnection) targetURL.openConnection();
        responseConnection.setRequestMethod("GET");
        try (final InputStream is = responseConnection.getInputStream()) {
            is.read();
        }
        System.out.println("Successfully communicated with " + targetURL);
    }
}
```

vsevel:
hello. any news on this?

matthyx:
@christianwimmer I got an idea: would it make sense to change the behavior of "native" java and instead of relying on a jks, use certificates directly from the CA bundle like Golang does:
https://golang.org/src/crypto/x509/root_linux.go
(of course it has to be implemented for each target platform in a different manner)

Since native compilation creates an executable, probably the enduser would expect to have a "native" way of handling certificates instead of the platform independent JKS approach... what do you think?

christianwimmer:
I assigned @teshull at Oracle to work on this for the GraalVM 21.3 release

matthyx:
Awesome, thanks a lot. Let me know what I can do to help.

teshull:
Drawing from #2831 and this PR, I made a new PR with support for setting a certificate file for the default keystore at runtime: #3652. The description is in https://github.com/oracle/graal/blob/github/tshull/GR-32212_truststoremanager/docs/reference-manual/native-image/CertificateManagement.md

Please let me know what you think.

matthyx:
> Drawing from #2831 and this PR, I made a new PR with support for setting a certificate file for the default keystore at runtime: #3652. The description is in https://github.com/oracle/graal/blob/github/tshull/GR-32212_truststoremanager/docs/reference-manual/native-image/CertificateManagement.md
> 
> Please let me know what you think.

Excellent, I have commented in your PR, and I'm closing this one as it's superseded by yours.

PR #3098, opened 2020-12-31
Proper getStackAccessControlContext implementation

### LINKED ISSUE #2745: GraalVM native image kerberos authentication issue: failed to store/retrieve AccessControlContext to/from stack

lazar-mitrovic:
* Proper getStackAccessControlContext implementation and removed some security substitutions (fixes #2745). Initial fixes for better Security support.
* Force `apt update` in Github CI to avoid azure repository issues

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
@@ -40,6 +40,7 @@ import java.security.ProtectionDomain;
 import java.security.Provider;
 import java.security.SecureRandom;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -48,7 +49,7 @@ import java.util.function.Predicate;
 
 import jdk.vm.ci.meta.MetaAccessProvider;
 import jdk.vm.ci.meta.ResolvedJavaField;
-import org.graalvm.compiler.phases.common.LazyValue;
+import org.graalvm.collections.Pair;
 import org.graalvm.compiler.serviceprovider.JavaVersionUtil;
 import org.graalvm.nativeimage.Platform;
 import org.graalvm.nativeimage.Platforms;
@@ -160,7 +161,8 @@ final class Target_java_security_AccessController {
         }
 
         if (context != null && context.equals(AccessControllerUtil.NO_CONTEXT_SINGLETON)) {
-            VMError.shouldNotReachHere("Invoked AccessControlContext was replaced at build time but wasn't reinitialized at run time.");
+            VMError.shouldNotReachHere("Invoked AccessControlContext was replaced at build time but wasn't reinitialized at run time.\n" +
+                            "This might be an indicator of improper build time initialization, or of a non-compatible JDK version.");
         }
 
         AccessControllerUtil.PrivilegedStack.push(context, caller);
@@ -199,50 +201,6 @@ class AccessControllerUtil {
         }
     }
 
-    static class INNOCUOUS_ACC {
-        static LazyValue<AccessControlContext> acc = new LazyValue<>(() -> new AccessControlContext(new ProtectionDomain[]{new ProtectionDomain(null, null)}));
-
-        static AccessControlContext get() {
-            return acc.get();
-        }
-
-        static void set(AccessControlContext ctx) {
-        }
-    }
-
-    static class NO_PERMISSIONS_CONTEXT {
-        static LazyValue<AccessControlContext> acc = new LazyValue<>(() -> AccessControllerUtil.contextWithPermissions(new Permission[0]));
-
-        static AccessControlContext get() {
-            return acc.get();
-        }
-
-        static void set(AccessControlContext ctx) {
-        }
-    }
-
-    static class GET_CLASS_LOADER_CONTEXT {
-        static LazyValue<AccessControlContext> acc = new LazyValue<>(() -> AccessControllerUtil.contextWithPermissions(new RuntimePermission("getClassLoader")));
-
-        static AccessControlContext get() {
-            return acc.get();
-        }
-
-        static void set(AccessControlContext ctx) {
-        }
-    }
-
-    static class GET_LOOKUP_CONTEXT {
-        static LazyValue<AccessControlContext> acc = new LazyValue<>(() -> AccessControllerUtil.contextWithPermissions(new RuntimePermission("dynalink.getLookup")));
-
-        static AccessControlContext get() {
-            return acc.get();
-        }
-
-        static void set(AccessControlContext ctx) {
-        }
-    }
-
     public static class PrivilegedStack {
 
         public static class StackElement {
@@ -263,7 +221,7 @@ class AccessControllerUtil {
             }
         }
 
-        @SuppressWarnings("rawtypes") private static final FastThreadLocalObject<ArrayDeque> stack = FastThreadLocalFactory.createObject(ArrayDeque.class);
+        @SuppressWarnings("rawtypes") private static final FastThreadLocalObject<ArrayDeque> stack = FastThreadLocalFactory.createObject(ArrayDeque.class, "accStack");
 
         @SuppressWarnings("unchecked")
         private static ArrayDeque<StackElement> getStack() {
@@ -296,14 +254,6 @@ class AccessControllerUtil {
         }
     }
 
-    public static AccessControlContext contextWithPermissions(Permission... perms) {
-        Permissions permissions = new Permissions();
-        for (Permission perm : perms) {
-            permissions.add(perm);
-        }
-        return new AccessControlContext(new ProtectionDomain[]{new ProtectionDomain(null, permissions)});
-    }
-
     static Throwable wrapCheckedException(Throwable ex) {
         if (ex instanceof Exception && !(ex instanceof RuntimeException)) {
             return new PrivilegedActionException((Exception) ex);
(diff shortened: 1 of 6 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
```diff
@@ -77,102 +79,108 @@
...
 
-    @Substitute
-    private static AccessControlContext createWrapper(DomainCombiner combiner, Class<?> caller, AccessControlContext parent, AccessControlContext context, Permission[] perms) {
-        return AccessControllerUtil.NO_CONTEXT_SINGLETON;
+        try {
+            return action.run();
+        } catch (Throwable ex) {
+            throw wrapCheckedException(ex);
+        } finally {
+            if (context != null) {
```
christianwimmer, line 161: pushing the `context` unconditionally but only popping it when `!= null` does not sound correct

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
```diff
@@ -77,102 +79,108 @@
 final class Target_java_security_AccessController {
 
     @Substitute
-    private static <T> T doPrivileged(PrivilegedAction<T> action) throws Throwable {
-        try {
-            return action.run();
-        } catch (Throwable ex) {
-            throw AccessControllerUtil.wrapCheckedException(ex);
-        }
+    public static <T> T doPrivileged(PrivilegedAction<T> action) throws Throwable {
```
christianwimmer, line 82: Please mark as many substitutions as possible as `JDK11OrEarlier`, since JDK 16 already has these methods implemented in Java.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
```diff
@@ -77,102 +79,108 @@
...
 
-    @Substitute
-    private static <T> T doPrivilegedWithCombiner(PrivilegedExceptionAction<T> action, AccessControlContext context, Permission... perms) throws Throwable {
+}
+
+@InternalVMMethod
+class AccessControllerUtil {
+
+    static final AccessControlContext NO_CONTEXT_SINGLETON;
+    static final ThreadLocal<Deque<AccessControlContext>> accStack = ThreadLocal.withInitial(LinkedList::new);
```
christianwimmer, line 122: Probably performance critical enough to use a `FastThreadLocal`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
```diff
@@ -77,102 +79,108 @@
...
-    private static <T> T doPrivileged(PrivilegedExceptionAction<T> action, AccessControlContext context) throws Throwable {
-        try {
-            return action.run();
-        } catch (Throwable ex) {
-            throw AccessControllerUtil.wrapCheckedException(ex);
+    /* From JDK15's AccessController */
+    static AccessControlContext checkContext(AccessControlContext context,
+                    Class<?> caller) {
+        // check if caller is authorized to create context
+        if (System.getSecurityManager() != null &&
```
christianwimmer, line 136: Note that `System.getSecurityManager` is currently constant folded to return `null`. So this is dead code. Maybe change it to 
```
if (System.getSecurityManager() != null) {
  throw VMError.shouldNotReachHere("Needs to be implemented when SecurityManager is supported");
}
```
so that it is explicit?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/StackTraceUtils.java
```diff
@@ -274,3 +280,67 @@ public boolean visitFrame(final FrameInfoQueryResult frameInfo) {
...
+            isPrivileged = true;
+            privilegedContext = AccessControllerUtil.accStack.get().peek();
+        }
+
+        if ((previousProtectionDomain == null || !previousProtectionDomain.equals(protectionDomain)) && (protectionDomain != null)) {
+            localArray.add(protectionDomain);
+            previousProtectionDomain = protectionDomain;
+        }
+
+        FrameInfoQueryResult caller = frameInfo.getCaller();
```
christianwimmer, line 317: So here you are relying on the fact that some method inlining happened. That does not sound correct.
What are you trying to do here? I don't really understand the comment below.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
christianwimmer, line 122: Why `LinkedList` and not `ArrayDeque`?

christianwimmer (changes requested):

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
lazar-mitrovic: Fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/StackTraceUtils.java
lazar-mitrovic: Yes this was a mistake. It should be more clear now.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
lazar-mitrovic: Fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
lazar-mitrovic: Fixed.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
lazar-mitrovic: Problem is that JDK16's implementation assumes that `getStackAccessControlContext` can easily access method locals to extract `context` and `caller`. Relevant code from Hotspot's `jvm.cpp`:
```
      javaVFrame *priv = vfst.asJavaVFrame();       // executePrivileged
      StackValueCollection* locals = priv->locals();
      StackValue* ctx_sv = locals->at(1); // AccessControlContext context
      StackValue* clr_sv = locals->at(2); // Class<?> caller
```
In order to avoid that - since AFAIK there is no nice and easy way to achieve that - I opted for Espresso approach by implementing parallel stack which is `FastThreadLocal`. Negative side of that approach is that we need to substitute `doPrivileged` even on JDK15+.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
lazar-mitrovic: Fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
lazar-mitrovic: This was remnant from debugging some failing gate. Should be fixed.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SecuritySubstitutions.java
christianwimmer: The parallel stack approach is fine. But still, the `doPrivileged` methods themselves do not need to be substituted anymore on JDK 16.

lazar-mitrovic: This should be fixed now.

mcraj017:
@lazar-mitrovic thanks for the contribution, is there any update on the changes requested by the reviewer?

lazar-mitrovic:
> @lazar-mitrovic thanks for the contribution, is there any update on the changes requested by the reviewer?

I intend to work on this towards the end of this week (after all of the `native-build-tools` issues are tackled).

mcraj017:
@lazar-mitrovic ok noted with thanks

qvistgaard:
Are you getting any closer to merging this PR?

lazar-mitrovic:
Closed in favor of: https://github.com/oracle/graal/pull/3854

PR #3139, opened 2021-01-20
Child context creation from an internal language

kustosz:
Fixes a bug in a child `TruffleContext` builder, where it required the instantiating language to be public. 
See the newly added test case, which fails with `language il is not installed` without the fixes implemented.

### DIFF
--- a/truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/ChildContextTest.java
+++ b/truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/ChildContextTest.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or
+ * data (collectively the "Software"), free of charge and under any and all
+ * copyright rights in the Software, and any and all patent rights owned or
+ * freely licensable by each licensor hereunder covering either (i) the
+ * unmodified Software as contributed to or provided by such licensor, or (ii)
+ * the Larger Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ *
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ *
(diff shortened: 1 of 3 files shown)


truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/ChildContextTest.java
```diff
@@ -0,0 +1,100 @@
...
+        Assert.assertEquals("foo", context.eval(PublicLang.ID, "").toString());
+    }
+
+    public static class LanguageContext {
+        public TruffleLanguage.Env env = null;
+    }
+
+    @TruffleLanguage.Registration(name = PublicLang.NAME, id = PublicLang.ID, dependentLanguages = {InternalLang.ID})
+    public static class PublicLang extends TruffleLanguage<LanguageContext> {
+        public static final String ID = "pl";
```
chumer, line 65: Language ids are globally unique for the tests. With `pl` there is the potential that it is used multiple times in the tests by accident. Please a more qualified id. Something like "ChildContextTest.PublicLang"

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextImpl.java
```diff
@@ -932,6 +940,12 @@ public boolean initializeLanguage(String languageId) {
         return false;
     }
 
+    @Override
+    public boolean initializeLanguage(String languageId) {
+        PolyglotLanguage language = requirePublicLanguage(languageId);
+        return initializeLanguage(language);
```
chumer, line 946: Use this implementation instead. The PolyglotImpl.guestToHostException method should only be used for calls from the embedding API and not from the TruffleContext.

```
 @Override
    public boolean initializeLanguage(String languageId) {
        PolyglotLanguage language = requirePublicLanguage(languageId);
        PolyglotLanguageContext languageContext = getContext(language);
        Object prev = hostEnter(languageContext);
         try {
              return initializeLanguage(language);
         } catch (Throwable t) {
            throw PolyglotImpl.guestToHostException(languageContext, t, true);
        } finally {
            hostLeave(languageContext, prev);
        }
   }
```

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextImpl.java
```diff
@@ -913,9 +913,17 @@ private PolyglotLanguageContext getLanguageContextImpl(Class<? extends TruffleLa
...
-    public boolean initializeLanguage(String languageId) {
-        PolyglotLanguage language = requirePublicLanguage(languageId);
+    void initializeInternalLanguage(String languageId) {
+        PolyglotLanguage language = engine.idToLanguage.get(languageId);
+        if (language == null) {
+            engine.requirePublicLanguage(languageId);
+            assert false;
+            return;
+        }
+        initializeLanguage(language);
```
chumer, line 923: Use this to enter leave without the embedding error handling. Also the language must never be null so that should be an assertion.
```
void initializeInnerContextLanguage(String languageId) {
        PolyglotLanguage language = engine.idToLanguage.get(languageId);
        assert language != null : "language that created inner context not found";
        Object prev = engine.enterIfNeeded(this);
         try {
              return initializeLanguage(language);
         } finally {
           engine.leaveIfNeeded(this, prev)
        }
}
```

chumer (changes requested):

chumer (changes requested):
Looks good now. Thanks! Will take it from here and integrate.

Issue #2825, opened 2020-09-01
Support of the Alpine Linux distribution


fernando-valdez:
**Feature request**

**Is your feature request related to a problem? Please describe.**
Provide out-of-the-box support of the Alpine Linux distribution on GraalVM CE and EE  

**Describe the solution you'd like.**
This can be delivered as its own platform. For example, if someone desires to use GraalVM using Alpine, it should be a package like this:  graalvm-ce-javaN-**Alpine**-amd64-X.Y.Z.tar.gz

**Describe who do you think will benefit the most.**
GraalVM users that want to use the advantage of a lightweight Linux distribution based on musl libc and busybox.

**Describe alternatives you've considered.**
Currently, is possible to execute GraalVM on Alpine using the glibc library. But the purpose of using Alpine is to use **musl**.

**Additional context.**
GraalVM's Github repository can demonstrate that there is a demand for support on Alpine. Here is a list of issues regarding the usage of Alpine:
https://github.com/oracle/graal/issues?q=+alpine+

petermz:
I'm currently working on it. I've got both 8 and 11 versions build on Alpine, but the native-image tool does not work properly: it produces images that crash when run

petermz:
I'm able to build substratevm and native-image with the attached patch applied and the following prerequisites:
- gcc-9.3.1 from [musl.cc](http://musl.cc/x86_64-linux-musl-native.tgz) (no luck with gcc package provided by Alpine)
- symlink named `musl-gcc` somewhere in PATH and pointing to `gcc` from the package above
- [labsjdk-11](https://github.com/graalvm/labs-openjdk-11.git)
- `apk add bash cmake git make openssh python2 zlib-static`
- [ninja](https://github.com/ninja-build/ninja) built for Alpine

[alpine-musl-ni.patch.txt](https://github.com/oracle/graal/files/5274407/alpine-musl-ni.patch.txt)

petermz:
I've managed to build complete GraalVM CE 20.3.0 package (including js, graalpython, truffleruby, and fastr) on Alpine Linux 3.11 with the 3 patches attached, and
- [labsjdk-11](https://github.com/graalvm/labs-openjdk-11.git) as boot JDK
- `apk add bash build-base cmake ed gfortran git make openssh python2 sed xz binutils-dev bzip2-dev curl-dev libffi-dev libxml2-dev openssl-dev pcre-dev readline-dev zlib-static`
[...]

PR #3141, opened 2021-01-21
linux-musl-amd64 support

petermz:
Hi,

This patch enables linux-musl-amd64 support. It addresses this [Alpine Linux issue](https://github.com/oracle/graal/issues/2825)

### mx files
* The build system calls `mx.get_os_variant()` and stores the value in a few properties around

### SVM
* At image build time, one such property is checked to tell whether we're cross compiling glibc -> musl, or run on musl natively. (This is in contrast to the initial commit in this branch, where more verbose annotation-based mechanism was used). Quite a few things in `MuslLibC` depend on that, e.g. we need `-static` for cross compiling only.
* `uint` -> `unsigned int` :)

### Sulong
* I need to pass `LIBCXX_HAS_MUSL_LIBC` to `cmake` when building `libcxx`. Here I piggyback on the existing `{{}}` mechanism, i.e. put a placeholder in `suite.py` and expand it in Python code.

### Truffle
* Do not define `_GNU_SOURCE` for musl builds.

### SDK
* The LLVM archive mentioned in `suite.py` is just the `LLVM_ORG_SRC` source package built with `-DLIBCXX_HAS_MUSL_LIBC=ON` on a musl host (Alpine 3.11)

### Installer
* I'm extending the filter language to support os-variant-arch triplet, where variant is optional. Graal updater checks the `OS_VARIANT` property and comes up with selector keys such as `linux_amd64` or `linux_musl_amd64`
* `GenerateCatalog` gets similar update, plus a new command line parameter to force variant


sdk/mx.sdk/mx_sdk_vm_impl.py
```diff
@@ -1973,8 +1976,13 @@ def _gen_gu_manifest(components, formatter, bundled=False):
...
     manifest["Bundle-Symbolic-Name"] = main_component.installable_id
     manifest["Bundle-Version"] = version
-    capability_fmt = 'org.graalvm; filter:="(&(graalvm_version={version})(os_name={os})(os_arch={arch})(java_version={java_version}))"'
+    capability_fmt = 'org.graalvm; filter:="(&(graalvm_version={version})(os_name={os}){variant}(os_arch={arch})(java_version={java_version}))"'
+    if mx.get_os_variant():
+        variant_filter = "(os_variant={})".format(mx.get_os_variant())
+    else:
+        variant_filter = ""
     manifest["Bundle-RequireCapability"] = capability_fmt.format(os=get_graalvm_os(),
+                                                                 variant = variant_filter,
```
gilles-duboscq, line 1985: minor style, remove the spaces around `=` in the keyword arguments.

sulong/mx.sulong/mx_sulong_suite_constituents.py
```diff
@@ -572,7 +572,12 @@ class CMakeProject(AbstractSulongNativeProject):  # pylint: disable=too-many-anc
     def __init__(self, suite, name, deps, workingSets, subDir, results=None, output=None, **args):
         super(CMakeProject, self).__init__(suite, name, deps, workingSets, subDir, results=results, output=output, **args)
         cmake_config = args.pop('cmakeConfig', {})
-        self.cmake_config = lambda: ['-D{}={}'.format(k, mx_subst.path_substitutions.substitute(v).replace('{{}}', '$')) for k, v in sorted(cmake_config.items())]
+        self.cmake_config = lambda: [
+            '-D{}={}'.format(k,
+                             mx_subst.path_substitutions.substitute(v)
+                                .replace('{{dollarsign}}', '$')
```
gilles-duboscq, line 578: It seems a bit strange to add an additional substitution mechanism since there is already one. I know this PR is just following a trend of the local code but it might be the right time to fix it rather than extend it.
This seems to process much more than paths so it should probably use `mx_subst.string_substitutions` and new string substitutions should be registered for `$` (`<dollarsign>`) and musl (`<has_musl_libc>`?). You can look for `string_substitutions.register_no_arg` to find examples.

gilles-duboscq (commented):
Overall this looks good. See my comments.

sulong/mx.sulong/mx_sulong_suite_constituents.py
rschatz: We fixed this already. `"\\$ORIGIN"` should now just work in `suite.py` (the `\\` escapes on the python level, resulting in the string containing `\$`, which mx interprets as literal `$`).

sulong/mx.sulong/suite.py
```diff
@@ -711,10 +711,12 @@
         "LIBCXX_CXX_ABI_INCLUDE_PATHS" : "<path:sdk:LLVM_ORG_SRC>/libcxxabi/include",
         "LIBCXX_ENABLE_STATIC" : "NO",
         "LIBCXX_ENABLE_EXPERIMENTAL_LIBRARY" : "NO",
+        # '{{musl}}' is replaced with YES or NO in the CMakeProject.
+        "LIBCXX_HAS_MUSL_LIBC" : "{{musl}}",
```
rschatz, line 715: I think this should use the `os_arch` mechanism of mx for this (see also current master, where we already use this for windows specific stuff):
```
"os": {
   "linux-musl": {
      "cmakeConfig": {
         "LIBCXX_HAS_MUSL_LIBC": "YES",
      }
   }
}
```

rschatz (changes requested):

rschatz (approved):

sdk/mx.sdk/mx_sdk_vm_impl.py
```diff
@@ -1248,6 +1249,8 @@ def contents(self):
             elif isinstance(image_config, mx_sdk.LauncherConfig):
                 suffix = _exe_suffix
                 project_name_f = GraalVmLauncher.launcher_project_name
+                if mx.get_os() == 'linux' and mx.get_os_variant() == 'musl':
+                    build_args.append('-H:+StaticExecutableWithDynamicLibC')
```
gradinac, line 1253: Hmm, do we hit issues with the executables we produce without this option? If I understand right, this means that all native-image executables that we ship as part of Graal only depend on `musl` and statically link everything else

espresso/mx.espresso/suite.py
```diff
@@ -290,6 +295,11 @@
                         "toolchain": "sulong:SULONG_BOOTSTRAP_TOOLCHAIN",
                     },
                 },
+                "linux-musl": {
+                    "<others>": {
+                        "ignore": "GNU Linux-only",
```
gilles-duboscq, line 300: While i understand not using the eden library on musl, it can't work to not build mokapot.
If espresso is not supported on musl, i would prefer to keep that a build failure for now.

espresso/mx.espresso/suite.py
```diff
@@ -410,6 +420,25 @@
...
+                "linux-musl": {
+                    "<others>": {
+                        "layout": {
+                            "./": [
+                                "file:mx.espresso/native-image.properties",
+                                "file:mx.espresso/reflectconfig.json",
+                            ],
+                            "LICENSE_JAVAONTRUFFLE": "file:LICENSE",
+                            "lib/": [
+                                "dependency:espresso:com.oracle.truffle.espresso.eden/<lib:eden>",
```
gilles-duboscq, line 432: if eden is ignored on musl, how does that work?

sulong/mx.sulong/suite.py
petermz: Looks much better indeed.

espresso/mx.espresso/suite.py
petermz: I've removed changes to this file, so building espresso now fails. I'm going to quickly look into the failure and see if I can come up with a fix in a day or two. If not, then I'll produce a separate PR later

petermz:
I've merged in the actual master branch and removed some outdated stuff. I'm now going to quickly look into the espresso failure and see if I can come up with a fix in a day or two. If not, then I'll produce a separate PR later

sdk/mx.sdk/mx_sdk_vm_impl.py
```diff
@@ -781,6 +781,7 @@ def _get_metadata(suites, parent_release_file=None):
         _metadata_dict.setdefault('JAVA_VERSION', _src_jdk.version)
         _metadata_dict.setdefault('OS_NAME', get_graalvm_os())
         _metadata_dict.setdefault('OS_ARCH', mx.get_arch())
+        _metadata_dict.setdefault('OS_VARIANT', mx.get_os_variant())
```
gilles-duboscq, line 784: I am wondering if we should skip this when `mx.get_os_variant()` is an empty string

gilles-duboscq (approved):

sdk/mx.sdk/mx_sdk_vm_impl.py
gradinac: Ah, perfect - thank you! :)

gradinac (approved):
The native-image changes look good to me

petermz:
I've fixed espresso build in a few steps:
* Excluded libeden since musl does not have `dlmopen` and namespaces
* Added `-Wno-error=cpp` to suppress errors related to nonstandard headers such as `<sys/poll.h>
* Added a `uint` typedef

Note that for musl builds to work, an accompanying change to mx is required. I've started a PR: https://github.com/graalvm/mx/pull/273

gilles-duboscq:
Thanks @petermz, the new changes for the espresso suite look good.
Note that i still don't think espresso would work with musl in this state but that's OK.

gilles-duboscq:
Thanks @petermz, i'll start integrating that.

petermz:
I've bumped mx version following integration of https://github.com/graalvm/mx/pull/273
Also `OS_VARIANT` is no longer set when variant is empty

Issue #2535, opened 2020-06-03
[native-image] Override META-INF native image properties


fhanik:
**Feature request**

Please include the following information:

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. E.g. I'm always frustrated when [...]

**Describe the solution you'd like.**
Some libraries include meta files for the native-image generator inside the JAR file at the `META-INF/native-image/groupId/artifactId` location.

The files often have to include reflection and resource data that encompasses all use cases.
When using a small portion of the library, it may be desirable to override this configuration with a provided one, a smaller one, to optimize the size of the binary.

**Describe who do you think will benefit the most.**
GraalVM users that compile the applications to native images will benefit since they will have a greater control over what is included in the image.

**Describe alternatives you've considered.**
Creating a repackaging step during a build cycle to repackage 3rd party jars is an option. This becomes hard to maintain.

olpaw:
That would essentially boil down to support passing of a blacklist of `META-INF/native-image/**/*` locations. Should not be too hard to add. This would be a good candidate for a PR provided by the community.

eginez:
Hey @fhanik, I am not sure this falls under the responsibility of the native-image build process. The optimization of the size of an image might require a pre-processing step which could be responsible of the above, in fact it seems like we the spring team is already doing something like that. We, the native-image team are trying, to keep the build-process focused on its core competency, namely building native executables.

galderz:
I've found myself needing this today. Btw, not the first time asked, @Sanne already asked for this in #1884 last year.

vjovanov:
I would be reluctant to introduce this feature as I believe, it would make the progress of the community slower. If there is a problem in the library, the right approach is to submit a PR to that library. 

Also, as @eginez pointed out, this can be achieved via other means (e.g., preprocessing). I would like to understand why submitting a PR to a library is not possible? Also, what makes it hard to remove invalid properties from a JAR?

maxandersen:
> I would like to understand why submitting a PR to a library is not possible? Also, what makes it hard to remove invalid properties from a JAR?

1) libraries does not necessarily release with the speed you need
2) removing invalid properties means forking/modifiying which one can do (we know we will have to it already in quarkus as many libs don't apply these things properly) 

note we will properly already need to do the #2 processing but still find it interesting we can explicitly turn off graalvm/native-image processing of bundled content to just reduce chance of surprises.

galderz:
[...]

PR #3149, opened 2021-01-25
Add flat to skip configuration files inside jars

galderz:
Related to #2535. 

Add a global option, `-H:+SkipConfigurationFiles`, to skip processing of native image metadata inside jar files.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ConfigurationFiles.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ConfigurationFiles.java
@@ -96,6 +96,9 @@ public final class ConfigurationFiles {
 
         @Option(help = "Comma-separated list of file names with declarative substitutions", type = OptionType.User)//
         public static final HostedOptionKey<String[]> SubstitutionFiles = new HostedOptionKey<>(null);
+
+        @Option(help = "Skip processing configuration files", type = OptionType.Expert)
+        public static final HostedOptionKey<Boolean> SkipConfigurationFiles = new HostedOptionKey<>(false);
     }
 
     public static List<Path> findConfigurationFiles(String fileName) {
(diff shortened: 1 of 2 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ConfigurationFiles.java
```diff
@@ -96,6 +96,9 @@
 
         @Option(help = "Comma-separated list of file names with declarative substitutions", type = OptionType.User)//
         public static final HostedOptionKey<String[]> SubstitutionFiles = new HostedOptionKey<>(null);
+
+        @Option(help = "Skip processing configuration files", type = OptionType.Expert)
+        public static final HostedOptionKey<Boolean> SkipConfigurationFiles = new HostedOptionKey<>(false);
```
olpaw, line 101: `HostedOptions` are only used for option processing in the image-builder and not the driver executable. If you want some driver specific option you have to add to `com.oracle.svm.driver.DefaultOptionHandler#consume`.

Also what is the usecase of this option? cc @christianwimmer

galderz: > HostedOptions are only used for option processing in the image-builder and not the driver executable. If you want some driver specific option you have to add to com.oracle.svm.driver.DefaultOptionHandler#consume.

+1

> Also what is the usecase of this option? cc @christianwimmer

Spelled out in linked issue #2535

olpaw: But this patch is a all-or-nothing approach and does not allow to exclude native image properties processing on a finer granular scale as suggested in https://github.com/oracle/graal/issues/2535#issuecomment-638208260
Instead there could be an option that takes a regex and uses that to decide if a given native image properties resource location should be excluded or not.

vjovanov: I am OK to add this as an non-API option, but then one needs to be able to specify the jars that are ignored. I don't believe there is much general value if one skips all jars on the classpath.

christianwimmer: I agree that skipping all configuration files is not really useful. It does not solve the "compositing" problem.

According to https://github.com/quarkusio/quarkus/issues/1658#issuecomment-762742950 the reason for this PR is the Oracle JDBC driver containing `--initialize-at-run-time=oracle.jdbc.OracleDriver,java.sql.DriverManager`. Since class initialization at run time is now the default, it should not be necessary at all.

galderz: Thx @christianwimmer @vjovanov @olpaw for the feedback. I've started a [discussion thread]((https://graalvm.slack.com/archives/CN9KSFB40/p1611664256033900)) in the GraalVM slack channel, where we can hash out an improved solution.

galderz:
Closing this PR. I've created a follow up PR with a more fine grained approach. See #3179.

PR #3152, opened 2021-01-25
Print TraceClassInitialization without error

ziyilin:
`-H:+TraceClassInitialization` can only print out trace information when class initialization error happens, but sometimes we still need to check the reason why a class is initialized without error.
This patch helps to achieve this goal. The specified class' initialization stack trace is printed out when the option is set.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
@@ -260,6 +260,10 @@ public class ClassInitializationFeature implements GraalFeature {
                 reportSafeTypeInitiazliation(universe, initGraph, path, provenSafe);
                 reportMethodInitializationInfo(path);
             }
+
+            if (SubstrateOptions.TraceClassInitialization.hasBeenSet()) {
+                reportTrackedClassInitializationTraces(path);
+            }
         }
     }
 
@@ -292,6 +296,21 @@ public class ClassInitializationFeature implements GraalFeature {
                                             .map(Class::getTypeName)
                                             .sorted()
                                             .forEach(writer::println));
+        }
+    }
+
+    private static void reportTrackedClassInitializationTraces(String path) {
+        Map<Class<?>, StackTraceElement[]> initializedClasses = ConfigurableClassInitialization.getInitializedClasses();
+        int size = initializedClasses.size();
+        if (size > 0) {
+            ReportUtils.report(size + " classes are tracked for initialization", path, "tracked_class_initialization", "txt", writer -> {
+                initializedClasses.forEach((k, v) -> {
+                    writer.println(k.getName());
+                    writer.println("---------------------------------------------");
+                    writer.println(ConfigurableClassInitialization.getTraceString(v));
+                    writer.println();
+                });
+            });
         }
     }
 
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/agent/NativeImageBytecodeInstrumentationAgent.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/agent/NativeImageBytecodeInstrumentationAgent.java
@@ -58,7 +58,7 @@ public class NativeImageBytecodeInstrumentationAgent {
 
     private static byte[] applyInitializationTrackingTransformation(@SuppressWarnings("unused") String moduleName, @SuppressWarnings("unused") ClassLoader loader, String className,
                     byte[] classfileBuffer) {
-        if (advisor.shouldTraceClassInitialization(className)) {
+        if (advisor.shouldTraceClassInitialization(className.replace('/', '.'))) {
             ClassReader reader = new ClassReader(classfileBuffer);
             ClassWriter writer = new ClassWriter(reader, COMPUTE_FRAMES);
             ClinitGenerationVisitor visitor = new ClinitGenerationVisitor(writer);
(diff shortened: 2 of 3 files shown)


gradinac (changes requested):
Hey @ziyilin, thank you very much for your contribution! This will be a useful new addition.

I've left several comments on how we could improve the integration of this functionality.
One more note, we should also update the `PrintClassInitialization` option to reflect that, when used in combination with `TraceClassInitialization`, we will also get a report file with tracked class' initialization traces printed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
```diff
@@ -283,14 +285,22 @@ private static void reportSafeTypeInitiazliation(AnalysisUniverse universe, Type
      * that belong to it.
      */
     private void reportMethodInitializationInfo(String path) {
+        List<String> tracedInit=new ArrayList();
```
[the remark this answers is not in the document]
ziyilin: fixed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
```diff
@@ -283,14 +285,22 @@ private static void reportSafeTypeInitiazliation(AnalysisUniverse universe, Type
...
                                             .sorted()
                                             .forEach(writer::println));
+            for(Class c: classes){
+                String trace =ConfigurableClassInitialization.classInitializationTrace(c);
+                if(trace!=null){
+                    tracedInit.add(trace);
+                }
+            }
         }
+        tracedInit.forEach(s->System.out.println(s));
```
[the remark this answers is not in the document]
ziyilin: fixed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
```diff
@@ -283,14 +285,22 @@ private static void reportSafeTypeInitiazliation(AnalysisUniverse universe, Type
...
     private void reportMethodInitializationInfo(String path) {
+        List<String> tracedInit=new ArrayList();
         for (InitKind kind : InitKind.values()) {
             Set<Class<?>> classes = classInitializationSupport.classesWithKind(kind);
             ReportUtils.report(classes.size() + " classes of type " + kind, path, kind.toString().toLowerCase() + "_classes", "txt",
                             writer -> classes.stream()
                                             .map(Class::getTypeName)
                                             .sorted()
                                             .forEach(writer::println));
+            for(Class c: classes){
```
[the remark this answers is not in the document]
ziyilin: fixed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
```diff
@@ -295,6 +299,33 @@ private void reportMethodInitializationInfo(String path) {
         }
     }
 
+    private static void reportTrackedClassInitializationTraces(String path) {
+        Map<Class<?>, StackTraceElement[]> initializedClasses = ConfigurableClassInitialization.getInitializedClasses();
+        int size = initializedClasses.size();
+        if (size > 0) {
+            // Print to file if required
+            if (Options.PrintClassInitialization.getValue()) {
```
gradinac, line 307: Thinking about it more now, we could probably use Options.TraceClassInitialization.hasBeenSet() here instead. I think that would make it more usable as it wouldn't require two different options to be found and set.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
```diff
@@ -295,6 +299,33 @@ private void reportMethodInitializationInfo(String path) {
...
+                    initializedClasses.forEach((k, v) -> {
+                        writer.println(k.getName());
+                        writer.println("---------------------------------------------");
+                        writer.println(ConfigurableClassInitialization.getTraceString(v));
+                        writer.println();
+                    });
+                });
+            }
+
+            // Print to console
```
gradinac, line 318: This might be redundant as we're already printing the report to a file - we should not duplicate it to the standard output here.

gradinac (commented):
This is looking great so far! I've left a couple of more comments - once they've been addressed, we can integrate the changes

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
ziyilin: `TraceClassInitialization` option has been already checked before entering this `reportTrackedClassInitializationTraces` method. I checked `PrintClassInitialization` option here to print an additional message into file besides console.

If we don't need to print the trace at two places but only to file, this check can be simply removed together with console printing code below.

gradinac: That's a good point - I think we should only print the traces to a file and remove the leftover code. I'm of course open to discussion about that, though I think that would be the most uniform approach

ziyilin: fixed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java
ziyilin: fixed

gradinac (approved):
Looks great! I will integrate these changes to `master` today.

zakkak:
Hi @gradinac, do you have any updates on this?

PR #3179, opened 2021-02-02
Add exclude-config option

galderz:
This PR Replaces #3149.

It uses a more fine grained approach to exclude configuration from jars. A new option `--exclude-config` is added that takes as a value a comma separating a jar name and regular expression of resources to exclude from the given jar. Jar names can include path elements for situations where two jar names have the same name but different leading path.

Examples:
```
--exclude-config foo.jar,META-INF\/native-image\/.*.properties
--exclude-config bar.jar,META-INF\/native-image\/.*.json
--exclude-config foo/example.jar,META-INF\/native-image\/.*.json
--exclude-config bar/example.jar,META-INF\/native-image\/.*.json
```

The approach was agreed with @olpaw on [this slack chat](https://graalvm.slack.com/archives/CN9KSFB40/p1611664256033900).

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
@@ -29,6 +29,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import org.graalvm.compiler.options.OptionType;
 
@@ -150,6 +151,18 @@ class DefaultOptionHandler extends NativeImage.OptionHandler<NativeImage> {
             case verboseServerOption:
                 args.poll();
                 NativeImage.showWarning("Ignoring server-mode native-image argument " + headArg + ".");
+                return true;
+            case "--exclude-config":
+                args.poll();
+                String excludeJar = args.poll();
+                if (excludeJar == null) {
+                    NativeImage.showError(headArg + " requires two arguments: a jar regular expression and a resource regular expression");
+                }
+                String excludeConfig = args.poll();
+                if (excludeConfig == null) {
+                    NativeImage.showError(headArg + " requires resource regular expression");
+                }
+                nativeImage.addExcludeConfig(Pattern.compile(excludeJar), Pattern.compile(excludeConfig));
                 return true;
         }
 
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -263,6 +263,8 @@ public class NativeImage {
     final Registry optionRegistry;
     private LinkedHashSet<EnabledOption> enabledLanguages;
 
+    private final List<ExcludeConfig> excludedConfigs = new ArrayList<>();
+
     public static final String nativeImagePropertiesFilename = "native-image.properties";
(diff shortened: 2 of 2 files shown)


maxandersen:
Is there way to use pattern for the jar?

galderz:
> Is there way to use pattern for the jar?

Not at the moment, but could be added.

olpaw:
> > Is there way to use pattern for the jar?
> 
> Not at the moment, but could be added.

Be care how you add that so that the separating of the regex for the jarfile and the regex for the resource path is unambiguous. I.e. if the regex of the path would contain a `,` you might mistakenly interpret it as the separator of jarfile-regex and resourcepath-regex. (I know ... `,` is quite unlikely in a regex for a path, but you never know what users are up to). Maybe the easiest solution would be to make the option take two arguments ... like
```
--exclude-config foo/.*\.jar META-INF\/native-image\/.*.json
```

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
```diff
@@ -151,6 +152,18 @@ public boolean consume(ArgumentQueue args) {
                 args.poll();
                 NativeImage.showWarning("Ignoring server-mode native-image argument " + headArg + ".");
                 return true;
+            case "--exclude-config":
+                args.poll();
+                String excludeConfig = args.poll();
+                if (excludeConfig == null) {
+                    NativeImage.showError(headArg + " requires a jar and resource regular expression, comma separated");
+                }
+                final String[] split = excludeConfig.split(",");
```
olpaw, line 161: If your regex contains a `","` you will only use the first half of it as the second argument to nativeImage.addExcludeConfig. What you need here is
```java
String[] split = excludeConfig.split(",", 2);
```

olpaw:
> > Is there way to use pattern for the jar?
> 
> Not at the moment, but could be added.

@galderz are you going to change the PR to make the path matching also be based on a regex (as @maxandersen suggested) or do you consider this PR ready as is?

olpaw (changes requested):
See https://github.com/oracle/graal/pull/3179/files#r571867459

galderz:
I'm handling this today. I'm going to try to apply @maxandersen's suggestion, along with the split into 2 parameters that @olpaw suggested. That sounds like a good combination.

galderz:
Updated PR. Added a commit to switch to 2 parameters and hence enable jar files to also be regular expressions. I can squash commits when it's ready to go in.

olpaw (approved):
LGTM. Please squash the commit so that I can create an internal PR.

galderz:
Thx @olpaw. Squashed and pushed.

zakkak:
@galderz it looks like there is a [formatting issue](https://github.com/oracle/graal/runs/1917831543?check_suite_focus=true#step:8:1603)

olpaw:
> @galderz it looks like there is a [formatting issue](https://github.com/oracle/graal/runs/1917831543?check_suite_focus=true#step:8:1603)

@zakkak is right. This should be fixed before I try to push this to master.
@galderz please adjust the formatting accordingly.

galderz:
Looking into the formatting issue

galderz:
How reliable is this formatting automation? When I run `mx eclipseformat -e /opt/eclipse/Eclipse.app/Contents/MacOS/eclipse --primary` locally I see [all these files formatted](https://gist.github.com/galderz/7c00ad5ca67ad6b88afc6d08f335bd15).

galderz:
Pushed an updated. Formatting should be fixed.

olpaw:
Your PR passed our gates and is now in the merge queue. ETA Monday

PR #3198, opened 2021-02-09
[GR-39179] GraalVM Thermometer tool

chrisseaton:
This is a rework of #2091. Check there for context.

What I'd like to do initially in this PR is resolve the big technical questions, before we debate the UI and code style, as I need help with the technical questions first and it may change the UI and code and so invalidate reviews of those.

### DIFF
--- a/compiler/src/org.graalvm.compiler.truffle.runtime/src/org/graalvm/compiler/truffle/runtime/debug/CompilationStatusListener.java
+++ b/compiler/src/org.graalvm.compiler.truffle.runtime/src/org/graalvm/compiler/truffle/runtime/debug/CompilationStatusListener.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package org.graalvm.compiler.truffle.runtime.debug;
+
+import com.oracle.truffle.api.frame.Frame;
+import com.oracle.truffle.api.instrumentation.CompilationState;
+import com.oracle.truffle.api.instrumentation.CompilationStateBackdoor;
+import org.graalvm.compiler.truffle.common.TruffleCompilerListener.CompilationResultInfo;
+import org.graalvm.compiler.truffle.common.TruffleCompilerListener.GraphInfo;
+import org.graalvm.compiler.truffle.runtime.*;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public final class CompilationStatusListener extends AbstractGraalTruffleRuntimeListener {
+
+    private final AtomicInteger finished = new AtomicInteger();
+    private final AtomicInteger failures = new AtomicInteger();
+    private final AtomicInteger dequeues = new AtomicInteger();
+    private final AtomicInteger deoptimizations = new AtomicInteger();
+
+    private CompilationStatusListener(GraalTruffleRuntime runtime) {
+        super(runtime);
+        CompilationStateBackdoor.ACCESSOR = this::sampleCompilationState;
+    }
+
+    public static void install(GraalTruffleRuntime runtime) {
+        runtime.addListener(new CompilationStatusListener(runtime));
+    }
+
+    @Override
+    public synchronized void onCompilationDequeued(OptimizedCallTarget target, Object source, CharSequence reason, int tier) {
+        dequeues.getAndIncrement();
+    }
+
+    @Override
+    public synchronized void onCompilationInvalidated(OptimizedCallTarget target, Object source, CharSequence reason) {
+        // We count invalidations as deoptimizations
+        deoptimizations.getAndIncrement();
+    }
+
+    @Override
+    public synchronized void onCompilationSuccess(OptimizedCallTarget target, TruffleInlining inliningDecision, GraphInfo graph, CompilationResultInfo result, int tier) {
+        finished.getAndIncrement();
+    }
+
+    @Override
+    public synchronized void onCompilationFailed(OptimizedCallTarget target, String reason, boolean bailout, boolean permanentBailout, int tier) {
+        if (TraceCompilationListener.isPermanentFailure(bailout, permanentBailout)) {
+            failures.getAndIncrement();
+        }
+    }
+
+    @Override
+    public synchronized void onCompilationDeoptimized(OptimizedCallTarget target, Frame frame) {
+        deoptimizations.getAndIncrement();
+    }
+
+    public synchronized CompilationState sampleCompilationState() {
+        /*
+         * Thermometer TODO: I'd like to track queue size and running count by
+         * using the listener as normal, but I can't find a way to make that
+         * work - I think the key problem is that compilations dequeued may
+         * either be queued or running and I can't tell which. We'll work to
+         * improve that if this is to be merged.
+         */
+
+        return new CompilationStateImpl(
+                runtime.getCompilationQueueSize(),
+                runtime.getCompilationsRunning(),
+                finished.get(),
+                failures.get(),
+                dequeues.get(),
+                deoptimizations.get());
+    }
+
+}
--- a/tools/src/com.oracle.truffle.tools.thermometer/src/com/oracle/truffle/tools/thermometer/Thermometer.java
+++ b/tools/src/com.oracle.truffle.tools.thermometer/src/com/oracle/truffle/tools/thermometer/Thermometer.java
@@ -1,0 +1,217 @@
+/*
+ * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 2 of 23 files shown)


tools/src/com.oracle.truffle.tools.thermometer/src/com/oracle/truffle/tools/thermometer/Thermometer.java
```diff
@@ -0,0 +1,217 @@
...
+
+            if (index == -1) {
+                env.getLogger("").log(Level.WARNING, "thermometer iteration location could not be parsed - should be file:line");
+                hasIterationLocation = false;
+            } else {
+                final String file = config.getIterationLocation().substring(0, index);
+                final int line = Integer.parseInt(config.getIterationLocation().substring(index + 1));
+
+                final SourceSectionFilter iterationNodeFilter = SourceSectionFilter.newBuilder()
+                        .tagIs(StandardTags.RootTag.class) // Thermometer TODO: I want the root of compilation units, so I catch OSR methods
```
chrisseaton, line 100: I want a tag for compilation roots. So including OSR roots. Do we think we can do that?

compiler/src/org.graalvm.compiler.truffle.runtime/src/org/graalvm/compiler/truffle/runtime/debug/CompilationStatusListener.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public synchronized CompilationState sampleCompilationState() {
+        /*
+         * Thermometer TODO: I'd like to track queue size and running count by
+         * using the listener as normal, but I can't find a way to make that
+         * work - I think the key problem is that compilations dequeued may
+         * either be queued or running and I can't tell which. We'll work to
+         * improve that if this is to be merged.
+         */
```
chrisseaton, line 87: If we could change how the compilation listener worked we could use it more simply and not have the `getCompilationsRunning()` API.

truffle/src/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/Instrumenter.java
```diff
@@ -460,4 +460,7 @@ public void onLoad(LoadSourceSectionEvent event) {
      */
     public abstract ExecutionEventNode lookupExecutionEventNode(Node node, EventBinding<?> binding);
 
+    // XXXXXXXXX
+    public abstract CompilationState sampleCompilationState();
```
chrisseaton, line 464: Instead of this I'd possibly like to have a `CompilationEventListener` in the instrumentation API, like `GraalTruffleRuntimeListener`, which would be fixed to allow me to do all the work I want to track compilations.

chrisseaton:
@fniephaus @eregon

truffle/src/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/CompilationState.java
```diff
@@ -0,0 +1,78 @@
...
+
+    int getRunning();
+
+    int getFinished();
+
+    int getFailed();
+
+    int getDequeued();
+
+    int getDeoptimizations();
```
chumer, line 43: Is there a reason you are capturing deoptimizations and not invalidations here?

truffle/src/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/Instrumenter.java
chumer: Yes that sounds like a better idea than exposing a concrete sampling. 

Here is a sketch of what I have in mind for that API:

```java

class Instrumenter {

  EventBinding attachCompilationEventListener(CompilationEventListener)

}


static abstract class CompilationInfo {

  CompilationInfo(){}

  public long getId() {...}
  public String getRootSimpleName() {...}
  public String getRootQualifiedName() {...}
  public int getTier() {...}
  public boolean isClone();
  public int getTruffleNodeCount();

}

static final class CompilationSuccessInfo extends CompilationInfo{

  public long getCodeSize() {...}
  public long getGraalNodeCount() {...}

}

static final class CompilationAbortInfo extends CompilationInfo{

  public boolean isTemporary(); // other temp bailout
  public boolean isPermanent(); // permanent bailout
  public boolean isFatal(); // internal error
  public String getReason();

}


interface CompilationEventListener {

void onQueued(Compilationinfo info)
void onDequeued(CompilationAbortInfo info)
void onStart(Compilationinfo info)
void onSuccess(CompilationSuccessInfo info)
void onFailed(CompilationAbortInfo info)
void onInvalidate(CompilationAbortInfo info,)

}
```

Note I don't think we should track deoptimizations in general. They are not necessarily related to compilations. Maybe we should have a separate listener for those events?
Is this something you definitely need?


we can use the same guaranteed order as in the compilation listener to document this:
```
 ( ((queue . unqueue))*
    . queue . started
    . (success)
      | ( failed)
    . invalidate )*
```

I suggest that in the default runtime we allow attaching events, but there are just never any events triggered.

tools/src/com.oracle.truffle.tools.thermometer/src/com/oracle/truffle/tools/thermometer/Thermometer.java
chumer: Instrumentation does not really know about this, and probably shouldn't.
How about you instrument a StatementTag instead?
Then you could use CompilerDirectives.isCompilationRoot() that folds to a constant on the fast path?

compiler/src/org.graalvm.compiler.truffle.runtime/src/org/graalvm/compiler/truffle/runtime/debug/CompilationStatusListener.java
chumer: I suggested a compilation listener API in a different comment.

chumer (changes requested):
I think both Truffle API and tool can be developed in the same PR, but it might speed up the process if we split the changes up into two PRs ultimately. One with the needed compilation listener API changes and then with the thermometer.

If the compilation listener API ultimately goes over your head, or you time out, feel free to delegate some parts to me. I have a TODO for this, also for other reasons.

Let me know If I missed to answer some open questions here.

tools/src/com.oracle.truffle.tools.thermometer/src/com/oracle/truffle/tools/thermometer/Thermometer.java
eregon: `StatementTag` would trigger in so many more places, and have a lot more footprint overhead, it sounds suboptimal to me.

chrisseaton: It does work with a statement tag, yes (it doesn't work with the root tag unless you turn off OSR.) As Benoit says it adds quite a bit of overhead.

I don't know how to argue whether instrumentation 'should' know about it or not, apart from this is an instrument it does need to know about it to solve the problem optimally. Also, an event that's triggered once per compilation unit entry seems pretty general to me.

truffle/src/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/CompilationState.java
chrisseaton: I simplified both deoptimisations and invalidations into a general concept of 'code that had to be thrown away' and called that deoptimisations.

truffle/src/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/Instrumenter.java
chrisseaton: Ok I'll try something minimal like this and ask for further comment when I've got the basics working.

tools/src/com.oracle.truffle.tools.thermometer/src/com/oracle/truffle/tools/thermometer/Thermometer.java
chumer: I was assuming that the iteration is something that specified as a single line location so very little instrumentation nodes would get created no?

chrisseaton: Sorry there's some confusion here. There's two probe nodes.

One probes iterations per second, and there's only one instance in the program (multiple due to splitting.)

The other sets a flag to say whether we're in compiled code or not, which is then sampled every few ms to give a proportion of time spent in compiled code. This second probe is the one we're ideally like to have once per compilation unit.

truffle/src/com.oracle.truffle.api.instrumentation/src/com/oracle/truffle/api/instrumentation/CompilationState.java
chumer: if code is thrown away it is always invalidated and made non reentrant. Deoptimization can happen for other reasons too (e.g. CompilerDirectives.transferToInterpreter), but does not imply invalidation. Because of that I think that terminology is misleading.

tools/src/com.oracle.truffle.tools.thermometer/src/com/oracle/truffle/tools/thermometer/Thermometer.java
chumer: Ok. So no problem with counting the iteration then.

Detecting OSR is tricky. We could maybe do it like with a new API CompilerDirectives.getCompilationRoot():

```
if (CompilerDirectives.inCompilationRoot()) {
  boolean isMethod = instrumentedNode.getRootNode().getCallTarget() == CompilerDirectives.getCompilationRoot();
  CompilerAsserts.partialEvaluationConstant(isMethod);
  if (isMethod) {
    // not in OSR
  } else {
    // in OSR, but hard to know which one, probably needs another API.
    CompilerDirectives.getCompilationRoot().getRootNode().somehowGetToTheLoop();
  }
}
```

I suggest we skip OSR in the first version.

I think you have another problem here. You actually need to track whether in compiled code for each thread. You could use ContextThreadLocals for that purpose. It is not very efficient in multi-threaded mode atm, but I have plans to make it much more efficient.

Another idea is to leverage the CPUSampler#takeSample() which can provide you with the same info, but is currently more heavy on overhead as it collects more info.

petenorth:
Any update on this PR? It would be really useful and could be the basis for exposing things like the size of the compilation queue etc to third party monitoring tools.

Or can the compilation queue size etc be accessed already from a Truffle language i.e. TruffleRuby

chrisseaton:
Sorry - the never-ending PR!

I'd like to really switch it to the new stack sampling API that I think we have thanks to the new safe points.

Give me a list of your feature wishes and how you'd like to access them and I'll incorporate that into what I try to get working.

petenorth:
Being able to run TruffleRuby in native image mode with the Thermometer instrument enabled would be useful.

The other usecases are where Truffle language metrics would be pushed or pulled to a third party monitoring solution (Datadog, New Relic, Prometheus etc.)

Normally this would be done just by adding a `gem 'newrelic_rpm'` to your gemfile etc. . So as long as the the SubstrateVM metrics and the Truffle Language Framework metrics (such as the compilation queue) etc are available to a Truffle language then the third party monitoring vendor can implement their agent as they see fit.

An alternative would be to implement this in Java as an instrument. This would mean having an instrument per monitoring vendor or an instrument with all of the monitoring vendor libraries included in it. The former is easier to maintain but I think would mean rebuilding the TruffleRuby native image for each instrument so is a bit of a non-starter (but running in JVM mode I guess would be fine). An instrument with all the vendor libraries becomes a monolith and building the TruffleRuby native image would take a long time.

So just having access to the the metrics from within a Truffle language such as TruffleRuby is probably enough to allow vendors to 'fill the space' and provide rich monitoring solutions - in a way that is complementary to the Thermometer tool.

chrisseaton:
> Being able to run TruffleRuby in native image mode with the Thermometer instrument enabled would be useful.

You can just build with this branch right now to get that.

> The other usecases are where Truffle language metrics would be pushed or pulled to a third party monitoring solution

Yeah I'd like to use OpenTelemetry.

PR #3205, opened 2021-02-12
Fix DW_AT_byte_size of DW_TAG_pointer for direct reference access

olpaw:


adinn (approved):
Yes, this makes perfect sense. Thanks for sorting this out.

PR #3207, opened 2021-02-12
Omit DebugLineInfo originating from inlined methods for now

olpaw:
To improve the debugging experience with CE debuginfo we should not emit lineinfo for non-root SourceMappings (i.e. inlined code) for now. Only when we later add debuginfo that species PC ranges for inlined methods (with `DW_TAG_inlined_subroutine`) having lineinfo for inlined code will work reasonably well when stepping through code with GDB.

This change will improve the stepping through experience with the CE debuginfo
It gets rid of jumping in-and-out of `GenScavengeAllocationSnippets` while line-stepping through code.

### DIFF
--- a/substratevm/mx.substratevm/mx_substratevm.py
+++ b/substratevm/mx.substratevm/mx_substratevm.py
@@ -756,6 +756,7 @@ def _debuginfotest(native_image, path, build_only, args):
                          '-cp', classpath('com.oracle.svm.test'),
                          '-Dgraal.LogFile=graal.log',
                          '-g',
+                         '-H:-OmitInlinedMethodDebugLineInfo',
                          '-H:DebugInfoSourceSearchPath=' + sourcepath,
                          '-H:DebugInfoSourceCacheRoot=' + join(path, 'sources'),
                          'hello.Hello'] + args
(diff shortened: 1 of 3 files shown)


adinn (approved):
I realise the current behaviour makes it harder to understand what is going on for many users. However, it is very helpful for someone who understands what the compiler does and is trying to debug a case where a native image app does not behave the same as the same app hosted on the JVM. I see that as the *primary* use case for native debugging at the moment (since debugging the Java app can and really should be done on the JVM before switching to a native image). I'd prefer to have a temporary, 'unsupported' switch to allow the old behaviour to be configured (e.g. an environment var setting) until proper handling of inline methods is implemented.

I'm not sure this matters too much as Foivos (@zakkak) is already looking into fixing up his patch that adds inline method info (he was waiting for my type info patch to be integrated). However, if that doesn't make it into the next release I think it would be good to provide that switch.

olpaw:
> I'd prefer to have a temporary, 'unsupported' switch to allow the old behaviour to be configured (e.g. an environment var setting) until proper handling of inline methods is implemented.

Yes that makes sense. I will add such a switch. Thanks for the input!

adinn:
@olpaw Thanks for the extra switch. All still good!

PR #3237, opened 2021-02-22
[GR-20638] Implement guest language safepoints for Truffle.

graalvmbot:
Implementation of Truffle Guest Language Safepoints.

Best start reviewing by reading the tutorial:
https://github.com/oracle/graal/blob/ece07b3a2923900ac2dd9a483048688267b51f7d/truffle/docs/Safepoints.md

Notable classes to review:

TruffleSafepoint
TruffleThreadLocalAction
API in TruffleLanguage.Env
API in TruffleInstrument.Env
API in Context

Note this currently does not yet work on HotSpot unless you disable compilation with --engine.Compilation=false as it needs a new JVMCI version that is currently internal.

Intended use-cases:

- Interrupt, exit, cancel and pause of polyglot context execution without invalidation of code.
- Reading Stack Traces from other Threads
- Ruby exposes it as part of the language to implement concurrent storage strategies.
- Debugger can evaluate expressions even when the context is currently running without violating thread-safety.
- Better and faster stack sampling without shadow stack for the cpu sampler tool.
- Traverse all references of a polyglot context (including stacks) for memory limits.
- Detect which languages are active on the stack lazily. E.g. this is part of allowing to mix Java with multiple threads and JavaScript in the same context.


chrisseaton:
Copyright years seem off throughout - for example `TruffleSafepoint.java`.

truffle/docs/Safepoints.md
```diff
@@ -0,0 +1,178 @@
...
+For example, a single arithmetic expression completes within a constant number of CPU cycles. 
+However, a loop that summarizes values over an array uses a non-constant time dependent on the actual array size. 
+This typically means that safepoints are best polled at the end of loops and at the end of function or method calls to cover recursion. 
+In addition, any guest language code that blocks the execution, like guest language locks, need to use the  `TruffleSafepoint.setBlocked(Interrupter)` API to allow cooperative polling of safepoints while the thread is waiting.
+
+Please read more details what steps languages need to do in the [javadoc](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/TruffleSafepoint.html).
+
+## Thread Local Actions
+
+Languages and instruments can submit action using their environment.
```
chrisseaton, line 34: > can submit an action

or

> can submit actions

truffle/docs/Safepoints.md
```diff
@@ -0,0 +1,178 @@
...
+```java
+
+Env env; // language or instrument environment
+
+env.submitThreadLocal(null, new ThreadLocalAction(true /*side-effecting*/, true /*synchronous*/) {
+     @Override
+     protected void perform(Access access) {
+         assert access.getThread() == Thread.currentThread();
+     }
+});
```
chrisseaton, line 47: Stylistic, but I feel like I'd prefer if `submitThreadLocal` created the `ThreadLocalAction` and I could write `env.submitThreadLocal(null, true, true, access -> assert access.getThread() == Thread.currentThread())`.

truffle/docs/Safepoints.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Current Limitations
+
+There is currently no way to run thread local actions while the a thread is executing in boundary annotated methods unless the method cooperatively polls safepoints or uses the blocking API. 
+Unfortunately it is not always possible to cooperatively poll safepoints, for example, if the code currently executes third party native code. 
+A future improvement will allow to run code for other threads while they are blocked. 
+This is one of the reasons why it is recommended to use `ThreadLocalAction.Access.getThread()` instead of directly using `Thread.currentThread()`.
+When the native call returns it needs to wait for any thread local action that is currently executing for this thread. 
+This will allow, to collect guest language stack traces from other threads while they are blocked by uncooperative native code.
+Currently the safepoint will be invoked when the native code returns.
```
chrisseaton, line 61: Ruby does some magic with alarm signals and things to interrupt native blocking calls. Is this going to be compatible with that?

truffle/docs/Safepoints.md
```diff
@@ -0,0 +1,178 @@
...
+[engine] [tl]   perform-done         0  thread[pool-1-thread-410]
+[engine] [tl]   perform-start        0  thread[pool-1-thread-411]
+[engine] [tl]   perform-done         0  thread[pool-1-thread-412]
+[engine] [tl]   perform-done         0  thread[pool-1-thread-411]
+[engine] [tl] done                   0  thread[pool-1-thread-411]
+```
+
+### Printing guest and host stack frames every time interval.
+
+The option `--engine.TraceStackTraceInterval=1000` allows to set the time in milliseconds to print the current stack trace on the next safepoint.
```
chrisseaton, line 118: As in it prints the stack on safepoints, but only once per `n` milliseconds?

truffle/docs/Safepoints.md
```diff
@@ -0,0 +1,178 @@
...
+Safepoints are explicitly polled by invoking the `TruffleSafepoint.poll(Node)` method. 
+A Truffle guest language implementation must ensure that a safepoint is polled repeatedly within a constant time interval. 
+For example, a single arithmetic expression completes within a constant number of CPU cycles. 
+However, a loop that summarizes values over an array uses a non-constant time dependent on the actual array size. 
+This typically means that safepoints are best polled at the end of loops and at the end of function or method calls to cover recursion. 
+In addition, any guest language code that blocks the execution, like guest language locks, need to use the  `TruffleSafepoint.setBlocked(Interrupter)` API to allow cooperative polling of safepoints while the thread is waiting.
+
+Please read more details what steps languages need to do in the [javadoc](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/TruffleSafepoint.html).
+
+## Thread Local Actions
```
chrisseaton, line 32: The difference between a 'safepoint' and a 'thread local action' isn't made clear in this document - could be confusing.

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleSafepoint.java
```diff
@@ -0,0 +1,399 @@
...
+/**
+ * Truffle safepoints allow interrupting the guest language execution to execute thread-local
+ * actions submitted by a language or tool. A safepoint is a location during the guest language
+ * execution where the state is consistent, and other operations can read its state.
+ * <p>
+ * <h3>Supporting Safepoints in a Language</h3>
+ *
+ * Safepoints are explicitly polled by invoking the {@link #poll(Node)} method. Safepoints are
+ * {@link #poll(Node) polled} with loose location or with {@link #pollHere(Node) exact location}. A
+ * poll with a loose location is significantly more efficient than a poll with a precise location as
```
chrisseaton, line 65: 'loose location' sounds strange to me. How about 'relaxed' or 'floating'? I think those are both existing terminology.

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleSafepoint.java
```diff
@@ -0,0 +1,399 @@
...
+                    interruptable.apply(object);
+                    break;
+                } catch (InterruptedException e) {
+                    TruffleSafepoint.poll(location);
+                }
+            }
+        } finally {
+            safepoint.setBlocked(prev);
+        }
+    }
```
chrisseaton, line 269: I think this is where TruffleRuby does some things with signals to interrupt native blocking calls? @eregon

compiler/src/org.graalvm.compiler.loop.phases/src/org/graalvm/compiler/loop/phases/LoopSafepointEliminationPhase.java
```diff
@@ -78,7 +78,7 @@ protected void run(StructuredGraph graph, MidTierContext context) {
                             canDisableSafepoint = ((ForeignCall) node).isGuaranteedSafepoint();
                         }
                         if (canDisableSafepoint) {
-                            loopEnd.disableSafepoint();
+                            loopEnd.garanteeSafepoint();
```
chrisseaton, line 81: Is this a typo? 'guarantee'

truffle/docs/Safepoints.md
chumer: @eregon  is investigating that. Last time I talked with him, there was no know problems with supporting that.

truffle/docs/Safepoints.md
chumer: I had this in my first version. I'd prefer that normally too. The advantage of the current design is that it is easier to document and it is shared across languages and instruments.

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleSafepoint.java
eregon: The SIGVTALRM stuff is essentially just another `Interrupter` that we'd set around blocking NFI calls.
This method is basically equivalent to `ThreadManager#runUntilResult()`.

truffle/docs/Safepoints.md
eregon: Yes, in Ruby we can interrupt most system calls, notably blocking NFI calls like `read(2)` by sending SIGVTALRM to a given thread (using `pthread_kill()`), and then the thread checks for safepoints before retrying transparently. And C extensions can provide their own Interrupter (Unblocker in TruffleRuby terms), so blocking calls can also be interrupted for C extensions if they use `rb_thread_call_without_gvl()`.

truffle/docs/Safepoints.md
chumer: Clarified this a bit:

The option `--engine.TraceStackTraceInterval=1000` allows to set the time interval in milliseconds to repeatedly print the current stack trace.
Note that the stack trace is printed on the next safepoint poll and therefore might not be accurate.

fniephaus:
Looks like this (or a version of this) has been pushed to master without closing this PR. Commit: https://github.com/oracle/graal/commit/d895395d06edacbe1d3a307f392aa31526ce7b6a

chumer:
Yes this was merged.

PR #3238, opened 2021-02-23
Improve substitution conflict error message

### LINKED ISSUE #1715: [native-image] When two substitutions are present the error is not helpful
Currently `register` looks like:

```
   private static <T> void register(Map<T, T> substitutions, T annotated, T original, T target) {
        if (annotated != null) {
            guarantee(!substitutions.containsKey(annotated) || substitutions.get(annotated) == original || substitutions.get(annotated) == target, "Already registered: %s", annotated);
            substitutions.put(annotated, target);
        }
        ...
    }
```

The output message only indicates the substitution is already registered but not what is causing the conflict. This means the only way for the user to find out is to attach a debugger and set a breakpoint at this line.

The error message should include more information about the two substitutions that conflict so that the user can potentially take appropriate action to remove one of them.

Regan-Koopmans:
This PR resolves #1715 

Conflicts encountered in the registration of method substitutions are currently logged without the conflicting value. These changes add this information to the error message such that debugging these errors can be done more effectively.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/AnnotationSubstitutionProcessor.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/substitute/AnnotationSubstitutionProcessor.java
@@ -811,11 +811,11 @@ public class AnnotationSubstitutionProcessor extends SubstitutionProcessor {
 
     private static <T> void register(Map<T, T> substitutions, T annotated, T original, T target) {
         if (annotated != null) {
-            guarantee(!substitutions.containsKey(annotated) || substitutions.get(annotated) == original || substitutions.get(annotated) == target, "Already registered: %s", annotated);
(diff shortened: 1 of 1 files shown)


fniephaus:
Thanks! This was merged a while ago via https://github.com/oracle/graal/commit/5ced762b1974eef8ec6df7461cf71d3eddb3605d.

fniephaus:
Sorry, I thought this was merged but that's not the case.

Regan-Koopmans:
@fniephaus I have signed the OCA, but this oca-check does not appear to recognise it. Can this be merged?

fniephaus:
@Regan-Koopmans the OCA check passes now... I've pinged the native image team for a quick review, but otherwise this should be good to go. Thanks!

cstancu (approved):

fniephaus:
Looks like we can get this merged soon, sorry for the delay!

fniephaus:
Merged via https://github.com/oracle/graal/commit/a19006f595b95d6401cb6c0408bc5a5de6058812.

PR #3253, opened 2021-03-03
Fix formatting in help texts

fniephaus:
This replaces all `%n`s with `\n` in two help texts that aren't run through `String.format()`. As a consequence, the formatting was broken when running with `--help:internal`. Here's an example:

image


fniephaus:
[It looks like](https://github.com/oracle/graal/pull/3253/checks?check_run_id=2020707735#step:8:756) code generation doesn't like new lines. Maybe help text should always be processed by `String.format()`?
What'd be a better fix, @chumer?

fniephaus:
Fixed with https://github.com/oracle/graal/commit/0e73661abe28efdc9358c9f97c7be99c13100b74.

PR #3266, opened 2021-03-10
Make -H:+PrintPointsToStatistics option thread safe

ziyilin:
PrintPointsToStatistics has NullPointerException and thread safe issues, as described in the issue: https://github.com/oracle/graal/issues/3267
This PR can fix the issues.


cstancu:
Thanks for the fix!

Issue #2552, opened 2020-06-08
UnsatisfiedLinkError when using PKCS11 in native image generated executable


mttcr:
**Describe the issue**
I have a java program that read smart cards under linux. It uses sun.security.pkcs11.SunPKCS11.
I have the following error when using PKCS11:
```
java.lang.UnsatisfiedLinkError: sun.security.pkcs11.wrapper.PKCS11.connect(Ljava/lang/String;Ljava/lang/String;)V [symbol: Java_sun_security_pkcs11_wrapper_PKCS11_connect or Java_sun_security_pkcs11_wrapper_PKCS11_connect__Ljava_lang_String_2Ljava_lang_String_2]
	at com.oracle.svm.jni.access.JNINativeLinkage.getOrFindEntryPoint(JNINativeLinkage.java:145)
	at com.oracle.svm.jni.JNIGeneratedMethodSupport.nativeCallAddress(JNIGeneratedMethodSupport.java:57)
	at sun.security.pkcs11.wrapper.PKCS11.connect(PKCS11.java)
	at sun.security.pkcs11.wrapper.PKCS11.<init>(PKCS11.java:147)
	at sun.security.pkcs11.wrapper.PKCS11.getInstance(PKCS11.java:160)
	at sun.security.pkcs11.SunPKCS11.<init>(SunPKCS11.java:318)
	at sun.security.pkcs11.SunPKCS11$1.run(SunPKCS11.java:112)
	at sun.security.pkcs11.SunPKCS11$1.run(SunPKCS11.java:109)
	at java.security.AccessController.doPrivileged(AccessController.java:119)
	at sun.security.pkcs11.SunPKCS11.configure(SunPKCS11.java:109)
...
```

**Steps to reproduce the issue**

compile and run this snippet with graal vm
```
    public static void main(String[] args)
    {
        String config = "name=opensc\n" //
            + "library=/usr/lib/x86_64-linux-gnu/opensc-pkcs11.so\n" //
            + "slot=0\n";
        sun.security.pkcs11.SunPKCS11 p11 = (SunPKCS11) new sun.security.pkcs11.SunPKCS11().configure("--" + config);
    }
```

**Describe GraalVM and your environment:**
 - GraalVM version: graalvm-ce-java11-linux-20.1.0
 - JDK major version: 11
 - OS: Ubuntu 18.04.4 LTS
 - Architecture: 64

**More details**
I actually did some things to make it works.
[...]

Issue #1410, opened 2019-06-20
native image on windows : dynamically load sunmscapi


mttcr:
Hello,
I'm trying to build a windows executable that makes use of sunmscapi. the native image command line is : "native-image myProgram --enable-all-security-services --no-fallback"
I read JCA-SECURITY-SERVICES.md about the Native implementations paragraph.
But i can't build the program and dynamically link it later, I have the following error:
```
  Creating library c:\Users\aaa\Downloads\graalvm-ce-windows-amd64-19.0.2\test\myprogram.lib and object c:\Users\aaa\Downloads\graalvm-ce-windows-amd64-19.0.2\test\myprogram.exp
myprogram.obj : error LNK2001: unresolved external symbol Java_sun_security_ec_ECDSASignature_signDigest
myprogram.obj : error LNK2001: unresolved external symbol Java_sun_security_ec_ECDSASignature_verifySignedDigest
myprogram.obj : error LNK2001: unresolved external symbol Java_sun_security_mscapi_KeyStore_loadKeysOrCertificateChains
myprogram.obj : error LNK2001: unresolved external symbol Java_sun_security_mscapi_PRNG_generateSeed
myprogram.obj : error LNK2001: unresolved external symbol Java_sun_security_mscapi_RSASignature_generatePublicKeyBlob
myprogram.obj : error LNK2001: unresolved external symbol Java_sun_security_mscapi_RSASignature_importPublicKey
myprogram.obj : error LNK2001: unresolved external symbol Java_sun_security_smartcardio_PCSC_SCardEstablishContext
c:\Users\aaa\Downloads\graalvm-ce-windows-amd64-19.0.2\test\myprogram.exe : fatal error LNK1120: 7 unresolved externals

        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
        at java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:593)
        at java.util.concurrent.ForkJoinTask.get(ForkJoinTask.java:1005)
        at com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:458)
[...]

PR #3300, opened 2021-03-21
statically link sunmscapi

kkriske:
This PR adds support to statically link `sunmscapi`, required for the `sun_security_mscapi` package.

First of all, this still has a **critical issue** that needs to be taken care of before it would be ready to merge.
Additionally, some extra JNI configuration can be added, and the reachability handler can be refined.

The critical issue with this addition, happens when both `sunmscapi` and `sunec` are added:
`sunec.lib(ECC_JNI.obj) : error LNK2005: ThrowException already defined in sunmscapi.lib(security.obj)`

The offending definitions:
https://github.com/graalvm/labs-openjdk-11/blob/0cbcc0176b3404262b95c180ea716e9b0cdeccf6/src/jdk.crypto.mscapi/windows/native/libsunmscapi/security.cpp#L132
https://github.com/graalvm/labs-openjdk-11/blob/0cbcc0176b3404262b95c180ea716e9b0cdeccf6/src/jdk.crypto.ec/share/native/libsunec/ECC_JNI.cpp#L50

I can handle the refinement of the reachability hander and the JNI config, but I need some help for the linking error.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
@@ -57,6 +57,7 @@ import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
@@ -309,6 +310,17 @@ public class SecurityServicesFeature extends JNIRegistrationUtil implements Feat
             /* Resolve calls to com_sun_security_auth_module_UnixSystem* as builtIn. */
             PlatformNativeLibrarySupport.singleton().addBuiltinPkgNativePrefix("com_sun_security_auth_module_UnixSystem");
         }
+
+        if (isWindows()) {
+            access.registerReachabilityHandler(SecurityServicesFeature::registerSunMSCAPIConfig, clazz(access, "sun.security.mscapi.SunMSCAPI"));
+            // statically linking sunmscapi conflicts with sunEC
+            // after the removal of sunEC in jdk 16, we can statically link sunmscapi
+            if (JavaVersionUtil.JAVA_SPEC >= 16) {
+                access.registerReachabilityHandler(SecurityServicesFeature::linkSunMSCAPI, clazz(access, "sun.security.mscapi.SunMSCAPI"));
+                /* Resolve calls to sun_security_mscapi* as builtIn. */
+                PlatformNativeLibrarySupport.singleton().addBuiltinPkgNativePrefix("sun_security_mscapi");
+            }
+        }
     }
 
     private void addManuallyConfiguredUsedProviders(DuringSetupAccess access) {
@@ -382,6 +394,52 @@ public class SecurityServicesFeature extends JNIRegistrationUtil implements Feat
             /* Library sunec depends on stdc++ */
             nativeLibraries.addDynamicNonJniLibrary("stdc++");
         }
+    }
+
+    private static void linkSunMSCAPI(DuringAnalysisAccess a) {
+        NativeLibraries nativeLibraries = ((FeatureImpl.DuringAnalysisAccessImpl) a).getNativeLibraries();
+        /* We statically link sunmscapi thus we classify it as builtIn library */
+        NativeLibrarySupport.singleton().preregisterUninitializedBuiltinLibrary("sunmscapi");
+
+        nativeLibraries.addStaticJniLibrary("sunmscapi");
+        /* Library sunmscapi depends on ncrypt and crypt32 */
+        nativeLibraries.addDynamicNonJniLibrary("ncrypt");
+        nativeLibraries.addDynamicNonJniLibrary("crypt32");
+    }
+
+    private static void registerSunMSCAPIConfig(BeforeAnalysisAccess a) {
+        registerForThrowNew(a, "java.security.cert.CertificateParsingException", "java.security.InvalidKeyException",
+                        "java.security.KeyException", "java.security.KeyStoreException", "java.security.ProviderException",
+                        "java.security.SignatureException", "java.lang.OutOfMemoryError");
+
+        a.registerReachabilityHandler(SecurityServicesFeature::registerLoadKeysOrCertificateChains,
+                        method(a, "sun.security.mscapi.CKeyStore", "loadKeysOrCertificateChains", String.class));
+        a.registerReachabilityHandler(SecurityServicesFeature::registerGenerateCKeyPair,
+                        method(a, "sun.security.mscapi.CKeyPairGenerator$RSA", "generateCKeyPair", String.class, int.class, String.class));
+        a.registerReachabilityHandler(SecurityServicesFeature::registerCPrivateKeyOf,
+                        method(a, "sun.security.mscapi.CKeyStore", "storePrivateKey", String.class, byte[].class, String.class, int.class));
+        a.registerReachabilityHandler(SecurityServicesFeature::registerCPublicKeyOf,
+                        method(a, "sun.security.mscapi.CSignature", "importECPublicKey", String.class, byte[].class, int.class),
+                        method(a, "sun.security.mscapi.CSignature", "importPublicKey", String.class, byte[].class, int.class));
+    }
+
+    private static void registerLoadKeysOrCertificateChains(DuringAnalysisAccess a) {
+        JNIRuntimeAccess.register(constructor(a, "java.util.ArrayList"));
+        JNIRuntimeAccess.register(method(a, "sun.security.mscapi.CKeyStore", "generateCertificate", byte[].class, Collection.class));
+        JNIRuntimeAccess.register(method(a, "sun.security.mscapi.CKeyStore", "generateCertificateChain", String.class, Collection.class));
+        JNIRuntimeAccess.register(method(a, "sun.security.mscapi.CKeyStore", "generateKeyAndCertificateChain", boolean.class, String.class, long.class, long.class, int.class, Collection.class));
+    }
+
+    private static void registerGenerateCKeyPair(DuringAnalysisAccess a) {
+        JNIRuntimeAccess.register(constructor(a, "sun.security.mscapi.CKeyPair", String.class, long.class, long.class, int.class));
+    }
+
+    private static void registerCPrivateKeyOf(DuringAnalysisAccess a) {
+        JNIRuntimeAccess.register(method(a, "sun.security.mscapi.CPrivateKey", "of", String.class, long.class, long.class, int.class));
+    }
+
+    private static void registerCPublicKeyOf(DuringAnalysisAccess a) {
+        JNIRuntimeAccess.register(method(a, "sun.security.mscapi.CPublicKey", "of", String.class, long.class, long.class, int.class));
     }
 
     private static void linkJaas(DuringAnalysisAccess a) {


pejovica:
> The critical issue with this addition, happens when both `sunmscapi` and `sunec` are added:
`sunec.lib(ECC_JNI.obj) : error LNK2005: ThrowException already defined in sunmscapi.lib(security.obj)`

Unfortunately, this would have to be fixed in the JDK. However, `sunec` has been removed in JDK 16 (see [JDK-8235710](https://bugs.openjdk.java.net/browse/JDK-8235710)), so you could adjust the PR so that static linking is used only on JDK 16+, while on older JDKs we would continue to use dynamic linking (which will now work properly, see 5a34e716b111961e5a40e4c641ca3ba318d01ce5).

kkriske:
I added a bit more JNI stuff, not sure if that's desired or not. And I added the suggested check for jdk 16. I have yet to try it out locally as I can't seem to build a java 16 version of GraalVM locally.

My build fails on the generation of polybench.exe, which I don't think is required so I'm probalby building it wrong? What would be the correct mx command to get a simple native-image capable local build?

kkriske:
Managed to get the build working and was able to try it out. It should be ready for review now.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
```diff
@@ -384,6 +405,37 @@ private static void linkSunEC(DuringAnalysisAccess a) {
...
+        NativeLibrarySupport.singleton().preregisterUninitializedBuiltinLibrary("sunmscapi");
+
+        nativeLibraries.addStaticJniLibrary("sunmscapi");
+        /* Library sunmscapi depends on ncrypt and crypt32 */
+        nativeLibraries.addDynamicNonJniLibrary("ncrypt");
+        nativeLibraries.addDynamicNonJniLibrary("crypt32");
+    }
+
+    private static void registerLoadKeysOrCertificateChains(DuringAnalysisAccess a) {
+        JNIRuntimeAccess.register(clazz(a, "java.util.ArrayList"));
```
pejovica, line 420: Not necessary, because the constructor/method registration will also register the class.
```suggestion
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
```diff
@@ -384,6 +405,37 @@ private static void linkSunEC(DuringAnalysisAccess a) {
...
+        JNIRuntimeAccess.register(constructor(a, "java.util.ArrayList"));
+        JNIRuntimeAccess.register(clazz(a, "sun.security.mscapi.CKeyStore$MY"));
+        JNIRuntimeAccess.register(clazz(a, "sun.security.mscapi.CKeyStore$ROOT"));
+        JNIRuntimeAccess.register(method(a, "sun.security.mscapi.CKeyStore", "generateCertificate", byte[].class, Collection.class));
+        JNIRuntimeAccess.register(method(a, "sun.security.mscapi.CKeyStore", "generateCertificateChain", String.class, Collection.class));
+        JNIRuntimeAccess.register(method(a, "sun.security.mscapi.CKeyStore", "generateKeyAndCertificateChain", boolean.class, String.class, long.class, long.class, int.class, Collection.class));
+    }
+
+    private static void registerGenerateCKeyPair(DuringAnalysisAccess a) {
+        JNIRuntimeAccess.register(clazz(a, "sun.security.mscapi.CKeyPair"));
```
pejovica, line 430:
```suggestion
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
```diff
@@ -384,6 +405,37 @@ private static void linkSunEC(DuringAnalysisAccess a) {
...
+        JNIRuntimeAccess.register(constructor(a, "sun.security.mscapi.CKeyPair", String.class, long.class, long.class, int.class));
+    }
+
+    private static void registerCPrivateKeyOf(DuringAnalysisAccess a) {
+        JNIRuntimeAccess.register(clazz(a, "sun.security.mscapi.CPrivateKey"));
```
pejovica, line 435:
```suggestion
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
```diff
@@ -309,6 +310,26 @@ public void beforeAnalysis(BeforeAnalysisAccess access) {
...
+        if (isWindows()) {
+            // JNI config required by sunmscapi
+            access.registerReachabilityHandler(SecurityServicesFeature::registerLoadKeysOrCertificateChains,
+                            method(access, "sun.security.mscapi.CKeyStore", "loadKeysOrCertificateChains", String.class));
+            access.registerReachabilityHandler(SecurityServicesFeature::registerGenerateCKeyPair,
+                            method(access, "sun.security.mscapi.CKeyPairGenerator$RSA", "generateCKeyPair", String.class, int.class, String.class));
+            access.registerReachabilityHandler(SecurityServicesFeature::registerCPrivateKeyOf,
+                            method(access, "sun.security.mscapi.CKeyStore", "storePrivateKey", String.class, byte[].class, String.class, int.class),
+                            method(access, "sun.security.mscapi.CSignature", "importECPublicKey", String.class, byte[].class, int.class),
+                            method(access, "sun.security.mscapi.CSignature", "importPublicKey", String.class, byte[].class, int.class));
```
pejovica, line 323: Could you factor this out into a helper method, perhaps `registerSunMSCAPIReachabilityHandlers`.

pejovica (commented):
> I added a bit more JNI stuff, not sure if that's desired or not.

It's definitely desirable. 😁

Is this the entire JNI config? If not, would you be able to provide the rest?

kkriske:
> Is this the entire JNI config? If not, would you be able to provide the rest?

I _think_ this is everything based on the source code I found [here](https://github.com/graalvm/labs-openjdk-11/blob/master/src/jdk.crypto.mscapi/windows/native/libsunmscapi/security.cpp). The jdk 16 code seems to be identical as well.

The only thing that may be missing is [some exception classes](https://github.com/graalvm/labs-openjdk-11/blob/60e0ae5d1ada661540ca16388b3315580ebecc78/src/jdk.crypto.mscapi/windows/native/libsunmscapi/security.cpp#L52-L60). But I'm not sure what to do with these.
Do I just put them in one entry like [here](https://github.com/oracle/graal/blob/f07d5e3b858b2441e942958d8f329238ff0b0c87/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationJava.java#L62-L69)?

pejovica:
> > Is this the entire JNI config? If not, would you be able to provide the rest?
> 
> I _think_ this is everything based on the source code I found [here](https://github.com/graalvm/labs-openjdk-11/blob/master/src/jdk.crypto.mscapi/windows/native/libsunmscapi/security.cpp). The jdk 16 code seems to be identical as well.

Great, then we shouldn't need an agent to collect the `sunmscapi` config anymore.

> The only thing that may be missing is [some exception classes](https://github.com/graalvm/labs-openjdk-11/blob/60e0ae5d1ada661540ca16388b3315580ebecc78/src/jdk.crypto.mscapi/windows/native/libsunmscapi/security.cpp#L52-L60). But I'm not sure what to do with these.
> Do I just put them in one entry like [here](https://github.com/oracle/graal/blob/f07d5e3b858b2441e942958d8f329238ff0b0c87/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationJava.java#L62-L69)?

I think we should register them here in a reachability handler, say `registerSunMSCAPIConfig`, which is triggered when the class `sun.security.mscapi.SunMSCAPI` becomes reachable. In fact, you can even move the registration of other handlers (suggested in https://github.com/oracle/graal/pull/3300#discussion_r662244696) into this new handler instead of the helper method.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
kkriske: done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
kkriske: done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
kkriske: done

kkriske:
The config is now located in a `registerSunMSCAPIConfig` which is registered as a reachabillity handler.
I duplicated the reachabillity handler for the linking, which could have 2 remarks:
* It duplicates the reachabillity handler, but I think it's clearer if the JNI config and the library linking are in separate methods
* I'm not sure if `registerClassInitializerReachabilityHandler` is right, or if it should be a normal class reachabillity handler.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
```diff
@@ -309,6 +310,17 @@ public void beforeAnalysis(BeforeAnalysisAccess access) {
             /* Resolve calls to com_sun_security_auth_module_UnixSystem* as builtIn. */
             PlatformNativeLibrarySupport.singleton().addBuiltinPkgNativePrefix("com_sun_security_auth_module_UnixSystem");
         }
+
+        if (isWindows()) {
+            access.registerClassInitializerReachabilityHandler(SecurityServicesFeature::registerSunMSCAPIConfig, clazz(access, "sun.security.mscapi.SunMSCAPI"));
```
pejovica, line 315: This should be a normal class reachability handler because the `sun.security.mscapi.SunMSCAPI` class is initialized at build time.
```suggestion
            access.registerReachabilityHandler(SecurityServicesFeature::registerSunMSCAPIConfig, clazz(access, "sun.security.mscapi.SunMSCAPI"));
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
```diff
@@ -309,6 +310,17 @@ public void beforeAnalysis(BeforeAnalysisAccess access) {
...
+            // statically linking sunmscapi conflicts with sunEC
+            // after the removal of sunEC in jdk 16, we can statically link sunmscapi
+            if (JavaVersionUtil.JAVA_SPEC >= 16) {
+                access.registerClassInitializerReachabilityHandler(SecurityServicesFeature::linkSunMSCAPI, clazz(access, "sun.security.mscapi.SunMSCAPI"));
```
pejovica, line 319: Same here.
```suggestion
                access.registerReachabilityHandler(SecurityServicesFeature::linkSunMSCAPI, clazz(access, "sun.security.mscapi.SunMSCAPI"));
```

pejovica (commented):
> I duplicated the reachabillity handler for the linking...

This is perfectly fine.

kkriske:
@pejovica changes applied.

pejovica (commented):
Two more corrections are needed, but other than that, it looks good!

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
```diff
@@ -384,6 +396,49 @@ private static void linkSunEC(DuringAnalysisAccess a) {
...
+        a.registerReachabilityHandler(SecurityServicesFeature::registerCPrivateKeyOf,
+                        method(a, "sun.security.mscapi.CKeyStore", "storePrivateKey", String.class, byte[].class, String.class, int.class),
+                        method(a, "sun.security.mscapi.CSignature", "importECPublicKey", String.class, byte[].class, int.class),
+                        method(a, "sun.security.mscapi.CSignature", "importPublicKey", String.class, byte[].class, int.class));
+    }
+
+    private static void registerLoadKeysOrCertificateChains(DuringAnalysisAccess a) {
+        JNIRuntimeAccess.register(constructor(a, "java.util.ArrayList"));
+        JNIRuntimeAccess.register(clazz(a, "sun.security.mscapi.CKeyStore$MY"));
+        JNIRuntimeAccess.register(clazz(a, "sun.security.mscapi.CKeyStore$ROOT"));
```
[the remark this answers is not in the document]
kkriske: Interesting, I did not know that.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/SecurityServicesFeature.java
```diff
@@ -384,6 +396,49 @@ private static void linkSunEC(DuringAnalysisAccess a) {
...
+                        "java.security.SignatureException", "java.lang.OutOfMemoryError");
+
+        a.registerReachabilityHandler(SecurityServicesFeature::registerLoadKeysOrCertificateChains,
+                        method(a, "sun.security.mscapi.CKeyStore", "loadKeysOrCertificateChains", String.class));
+        a.registerReachabilityHandler(SecurityServicesFeature::registerGenerateCKeyPair,
+                        method(a, "sun.security.mscapi.CKeyPairGenerator$RSA", "generateCKeyPair", String.class, int.class, String.class));
+        a.registerReachabilityHandler(SecurityServicesFeature::registerCPrivateKeyOf,
+                        method(a, "sun.security.mscapi.CKeyStore", "storePrivateKey", String.class, byte[].class, String.class, int.class),
+                        method(a, "sun.security.mscapi.CSignature", "importECPublicKey", String.class, byte[].class, int.class),
+                        method(a, "sun.security.mscapi.CSignature", "importPublicKey", String.class, byte[].class, int.class));
```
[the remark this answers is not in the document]
kkriske: fixed

kkriske:
requested changes applied.

pejovica (approved):
Excellent! I'll start the integration.

mttcr:
Thank you for the feature, it was awaited for #1410 ! 
I'm not sure this is the right place to ask, but could something similar be done to statically link j2pkcs11 under linux? this would solve #2552

sgeeroms:
Hi, we have an issue the when no java is installed the test reverts to dynamic linking, but then it fails to link them off course. Is there a way to force static linking in case there is no java installed. Or do I maybe not really understand what static linking beholds?
See the error we get on a machine without java installed. It works fine on a pc with java 17 installed (and in the PATH).
image

PR #3304, opened 2021-03-23
Associate debug Ranges with MethodEntries

zakkak:
As discussed in https://github.com/oracle/graal/pull/3290 this PR ensures that every debug `Range` is associated with a `MethodEntry`. This PR will enable us to continue working on https://github.com/oracle/graal/pull/2880 and associate the debug info for nested inline methods with `MethodEntry` instances.

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -53,7 +53,7 @@ public class ClassEntry extends StructureTypeEntry {
     /**
      * Details of this class's interfaces.
      */
-    protected LinkedList<InterfaceClassEntry> interfaces;
+    protected List<InterfaceClassEntry> interfaces;
     /**
      * Details of the associated file.
      */
@@ -66,7 +66,7 @@ public class ClassEntry extends StructureTypeEntry {
      * A list recording details of all primary ranges included in this class sorted by ascending
      * address range.
      */
-    private LinkedList<PrimaryEntry> primaryEntries;
+    private List<PrimaryEntry> primaryEntries;
     /**
      * An index identifying primary ranges which have already been encountered.
      */
@@ -78,7 +78,7 @@ public class ClassEntry extends StructureTypeEntry {
     /**
      * A list of the same files.
      */
-    private LinkedList<FileEntry> localFiles;
+    private List<FileEntry> localFiles;
     /**
      * An index of all primary and secondary dirs referenced from this class's compilation unit.
      */
@@ -86,7 +86,7 @@ public class ClassEntry extends StructureTypeEntry {
     /**
      * A list of the same dirs.
      */
-    private LinkedList<DirEntry> localDirs;
+    private List<DirEntry> localDirs;
     /**
      * This flag is true iff the entry includes methods that are deopt targets.
      */
@@ -94,14 +94,16 @@ public class ClassEntry extends StructureTypeEntry {
 
     public ClassEntry(String className, FileEntry fileEntry, int size) {
         super(className, size);
-        this.interfaces = new LinkedList<>();
+        this.interfaces = new ArrayList<>();
         this.fileEntry = fileEntry;
+        // methods is a sorted list and we want to be able to add more elements to it while keeping it sorted,
+        // so a LinkedList seems more appropriate than an ArrayList. (see getMethodEntry)
         this.methods = new LinkedList<>();
-        this.primaryEntries = new LinkedList<>();
+        this.primaryEntries = new ArrayList<>();
         this.primaryIndex = new HashMap<>();
-        this.localFiles = new LinkedList<>();
+        this.localFiles = new ArrayList<>();
         this.localFilesIndex = new HashMap<>();
-        this.localDirs = new LinkedList<>();
+        this.localDirs = new ArrayList<>();
         this.localDirsIndex = new HashMap<>();
         if (fileEntry != null) {
             localFiles.add(fileEntry);
@@ -231,7 +233,7 @@ public class ClassEntry extends StructureTypeEntry {
         return fileEntry;
     }
 
-    public LinkedList<PrimaryEntry> getPrimaryEntries() {
+    public List<PrimaryEntry> getPrimaryEntries() {
         return primaryEntries;
     }
 
@@ -240,11 +242,11 @@ public class ClassEntry extends StructureTypeEntry {
         return primaryIndex.get(primaryRange);
     }
 
-    public LinkedList<DirEntry> getLocalDirs() {
+    public List<DirEntry> getLocalDirs() {
         return localDirs;
     }
 
-    public LinkedList<FileEntry> getLocalFiles() {
+    public List<FileEntry> getLocalFiles() {
         return localFiles;
     }
 
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -268,8 +268,8 @@ public class ClassEntry extends StructureTypeEntry {
     }
 
     protected MethodEntry processMethod(DebugMethodInfo debugMethodInfo, DebugInfoBase debugInfoBase, DebugContext debugContext) {
-        String methodName = debugInfoBase.uniqueDebugString(debugMethodInfo.methodName());
-        String resultTypeName = TypeEntry.canonicalize(debugMethodInfo.returnTypeName());
+        String methodName = debugInfoBase.uniqueDebugString(debugMethodInfo.name());
+        String resultTypeName = TypeEntry.canonicalize(debugMethodInfo.valueType());
         int modifiers = debugMethodInfo.modifiers();
         List<String> paramTypes = debugMethodInfo.paramTypes();
         List<String> paramNames = debugMethodInfo.paramNames();
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -337,25 +337,11 @@ public class ClassEntry extends StructureTypeEntry {
         return superClass;
     }
 
-    public Range makePrimaryRange(String methodName, String symbolName, String paramSignature, String returnTypeName, StringTable stringTable, MethodEntry method, int lo,
-                    int hi, int primaryLine) {
+    public Range makePrimaryRange(String symbolName, StringTable stringTable, MethodEntry method, int lo, int hi, int primaryLine) {
         FileEntry fileEntryToUse = method.fileEntry;
         if (fileEntryToUse == null) {
-            /*
-             * Search for a matching method to supply the file entry or failing that use the one
-             * from this class.
-             */
-            for (MethodEntry methodEntry : methods) {
-                if (methodEntry.match(methodName, paramSignature, returnTypeName)) {
-                    /* maybe the method's file entry */
-                    fileEntryToUse = methodEntry.getFileEntry();
-                    break;
-                }
-            }
-            if (fileEntryToUse == null) {
-                /* Last chance is the class's file entry. */
-                fileEntryToUse = this.fileEntry;
-            }
+            /* Last chance is the class's file entry. */
+            fileEntryToUse = this.fileEntry;
         }
         return new Range(symbolName, stringTable, method, fileEntryToUse, lo, hi, primaryLine);
     }
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2020, 2020, Red Hat Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
@@ -35,9 +35,11 @@ import com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugTypeInfo.DebugType
 import org.graalvm.compiler.debug.DebugContext;
 
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 
 /**
@@ -134,7 +136,9 @@ public class ClassEntry extends StructureTypeEntry {
         /* Add details of fields and field types */
         debugInstanceTypeInfo.fieldInfoProvider().forEach(debugFieldInfo -> this.processField(debugFieldInfo, debugInfoBase, debugContext));
         /* Add details of methods and method types */
-        debugInstanceTypeInfo.methodInfoProvider().forEach(methodFieldInfo -> this.processMethod(methodFieldInfo, debugInfoBase, debugContext));
+        debugInstanceTypeInfo.methodInfoProvider().forEach(methodFieldInfo -> this.methods.add(this.processMethod(methodFieldInfo, debugInfoBase, debugContext)));
+        /* Sort methods to improve lookup speed */
+        this.methods.sort(MethodEntry::compareTo);
     }
 
     public void indexPrimary(Range primary, List<DebugFrameSizeChange> frameSizeInfos, int frameSize) {
@@ -294,9 +298,7 @@ public class ClassEntry extends StructureTypeEntry {
          * substitution
          */
         FileEntry methodFileEntry = debugInfoBase.ensureFileEntry(fileName, filePath, cachePath);
-        final MethodEntry methodEntry = new MethodEntry(methodFileEntry, methodName, this, resultType, paramTypeArray, paramNameArray, modifiers, debugMethodInfo.isDeoptTarget());
-        methods.add(methodEntry);
-        return methodEntry;
+        return new MethodEntry(methodFileEntry, methodName, this, resultType, paramTypeArray, paramNameArray, modifiers, debugMethodInfo.isDeoptTarget());
     }
 
     @Override
@@ -346,16 +348,30 @@ public class ClassEntry extends StructureTypeEntry {
         return new Range(symbolName, stringTable, method, fileEntryToUse, lo, hi, primaryLine);
     }
 
-    public MethodEntry ensureMethodEntry(DebugMethodInfo debugMethodInfo, DebugInfoBase debugInfoBase, DebugContext debugContext) {
+    public MethodEntry getMethodEntry(DebugMethodInfo debugMethodInfo, DebugInfoBase debugInfoBase, DebugContext debugContext) {
+        assert listIsSorted(methods);
         String methodName = debugInfoBase.uniqueDebugString(debugMethodInfo.name());
         String paramSignature = debugMethodInfo.paramSignature();
         String returnTypeName = debugMethodInfo.valueType();
-        // TODO improve data structure to avoid loops...
-        for (MethodEntry methodEntry : methods) {
-            if (methodEntry.match(methodName, paramSignature, returnTypeName)) {
+        ListIterator<MethodEntry> methodIterator = methods.listIterator();
+        while (methodIterator.hasNext()) {
+            MethodEntry methodEntry = methodIterator.next();
+            int comparisonResult = methodEntry.compareTo(methodName, paramSignature, returnTypeName);
+            if (comparisonResult == 0) {
                 return methodEntry;
+            } else if (comparisonResult > 0) {
+                methodIterator.previous();
+                break;
             }
         }
-        return processMethod(debugMethodInfo, debugInfoBase, debugContext);
+        MethodEntry newMethodEntry = processMethod(debugMethodInfo, debugInfoBase, debugContext);
+        methodIterator.add(newMethodEntry);
+        return newMethodEntry;
+    }
+
+    private boolean listIsSorted(List<MethodEntry> list) {
+        List<MethodEntry> copy = new ArrayList<>(list);
+        copy.sort(MethodEntry::compareTo);
+        return list.equals(copy);
     }
 }
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -294,7 +294,7 @@ public class ClassEntry extends StructureTypeEntry {
          * substitution
          */
         FileEntry methodFileEntry = debugInfoBase.ensureFileEntry(fileName, filePath, cachePath);
-        final MethodEntry methodEntry = new MethodEntry(methodFileEntry, methodName, this, resultType, paramTypeArray, paramNameArray, modifiers);
+        final MethodEntry methodEntry = new MethodEntry(methodFileEntry, methodName, this, resultType, paramTypeArray, paramNameArray, modifiers, debugMethodInfo.isDeoptTarget());
         methods.add(methodEntry);
         return methodEntry;
     }
@@ -338,7 +338,7 @@ public class ClassEntry extends StructureTypeEntry {
     }
 
     public Range makePrimaryRange(String methodName, String symbolName, String paramSignature, String returnTypeName, StringTable stringTable, MethodEntry method, int lo,
-                                  int hi, int primaryLine, boolean isDeoptTarget) {
+                    int hi, int primaryLine) {
         FileEntry fileEntryToUse = method.fileEntry;
         if (fileEntryToUse == null) {
             /*
@@ -357,7 +357,7 @@ public class ClassEntry extends StructureTypeEntry {
                 fileEntryToUse = this.fileEntry;
             }
         }
-        return new Range(symbolName, stringTable, method, fileEntryToUse, lo, hi, primaryLine, isDeoptTarget);
+        return new Range(symbolName, stringTable, method, fileEntryToUse, lo, hi, primaryLine);
     }
 
     public MethodEntry ensureMethodEntry(DebugInfoProvider.DebugMethodInfo debugMethodInfo, DebugInfoBase debugInfoBase, DebugContext debugContext) {
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -267,9 +267,9 @@ public class ClassEntry extends StructureTypeEntry {
         interfaceClassEntry.addImplementor(this, debugContext);
     }
 
-    protected void processMethod(DebugMethodInfo debugMethodInfo, DebugInfoBase debugInfoBase, DebugContext debugContext) {
-        String methodName = debugInfoBase.uniqueDebugString(debugMethodInfo.name());
-        String resultTypeName = TypeEntry.canonicalize(debugMethodInfo.valueType());
+    protected MethodEntry processMethod(DebugMethodInfo debugMethodInfo, DebugInfoBase debugInfoBase, DebugContext debugContext) {
+        String methodName = debugInfoBase.uniqueDebugString(debugMethodInfo.methodName());
+        String resultTypeName = TypeEntry.canonicalize(debugMethodInfo.returnTypeName());
         int modifiers = debugMethodInfo.modifiers();
         List<String> paramTypes = debugMethodInfo.paramTypes();
         List<String> paramNames = debugMethodInfo.paramNames();
@@ -294,7 +294,9 @@ public class ClassEntry extends StructureTypeEntry {
          * substitution
          */
         FileEntry methodFileEntry = debugInfoBase.ensureFileEntry(fileName, filePath, cachePath);
-        methods.add(new MethodEntry(methodFileEntry, methodName, this, resultType, paramTypeArray, paramNameArray, modifiers));
+        final MethodEntry methodEntry = new MethodEntry(methodFileEntry, methodName, this, resultType, paramTypeArray, paramNameArray, modifiers);
+        methods.add(methodEntry);
+        return methodEntry;
     }
 
     @Override
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -369,7 +369,7 @@ public class ClassEntry extends StructureTypeEntry {
         return newMethodEntry;
     }
 
-    private boolean listIsSorted(List<MethodEntry> list) {
+    private static boolean listIsSorted(List<MethodEntry> list) {
         List<MethodEntry> copy = new ArrayList<>(list);
         copy.sort(MethodEntry::compareTo);
         return list.equals(copy);
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -26,7 +26,7 @@
 
 package com.oracle.objectfile.debugentry;
 
-import com.oracle.objectfile.debuginfo.DebugInfoProvider;
+import com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugFieldInfo;
 import com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugFrameSizeChange;
 import com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugMethodInfo;
 import com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugTypeInfo;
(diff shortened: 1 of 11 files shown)


zakkak:
@adinn please review

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
```diff
@@ -335,10 +337,9 @@ public ClassEntry getSuperClass() {
...
-    public Range makePrimaryRange(String methodName, String symbolName, String paramSignature, String returnTypeName, StringTable stringTable, FileEntry primaryFileEntry, int lo,
-                    int hi, int primaryLine,
-                    int modifiers, boolean isDeoptTarget) {
-        FileEntry fileEntryToUse = primaryFileEntry;
+    public Range makePrimaryRange(String methodName, String symbolName, String paramSignature, String returnTypeName, StringTable stringTable, MethodEntry method, int lo,
+                                  int hi, int primaryLine, boolean isDeoptTarget) {
+        FileEntry fileEntryToUse = method.fileEntry;
         if (fileEntryToUse == null) {
             /*
              * Search for a matching method to supply the file entry or failing that use the one
```
adinn, line 345: So this iteration through the methods is now no longer needed because you have already done it under ensureMethodEntry.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
```diff
@@ -236,40 +236,38 @@ public void installDebugInfo(DebugInfoProvider debugInfoProvider) {
...
+            String className = TypeEntry.canonicalize(debugCodeInfo.ownerType());
+            String methodName = debugCodeInfo.name();
             String symbolName = debugCodeInfo.symbolNameForMethod();
             String paramSignature = debugCodeInfo.paramSignature();
-            String returnTypeName = TypeEntry.canonicalize(debugCodeInfo.returnTypeName());
+            String returnTypeName = TypeEntry.canonicalize(debugCodeInfo.valueType());
             int lo = debugCodeInfo.addressLo();
             int hi = debugCodeInfo.addressHi();
             int primaryLine = debugCodeInfo.line();
             boolean isDeoptTarget = debugCodeInfo.isDeoptTarget();
```
adinn, line 247: I think we probably ought to migrate DebugCodeInfo::isDeoptTarget() to DebugMethodInfo and cache this value in the MethodEntry when we create it. The test simply checks that the hosted method name ends in "**" and coudl be done up front when processing the DebugMethodInfo. If we do that then there is no need to retrieve isDeoptTarget here and pass it into makePrimaryRange. The latter function can read it off the methodEntry.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
```diff
@@ -356,6 +357,19 @@ public Range makePrimaryRange(String methodName, String symbolName, String param
                 fileEntryToUse = this.fileEntry;
             }
         }
-        return new Range(this.typeName, methodName, symbolName, paramSignature, returnTypeName, stringTable, fileEntryToUse, lo, hi, primaryLine, modifiers, isDeoptTarget);
+        return new Range(symbolName, stringTable, method, fileEntryToUse, lo, hi, primaryLine, isDeoptTarget);
+    }
+
+    public MethodEntry ensureMethodEntry(DebugInfoProvider.DebugMethodInfo debugMethodInfo, DebugInfoBase debugInfoBase, DebugContext debugContext) {
+        String methodName = debugInfoBase.uniqueDebugString(debugMethodInfo.name());
```
adinn, line 364: Import DebugInfoProvider.DebugMethodInfo so you don't have to have a qualified type in the argument list.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
zakkak: Correct. My plan is to fix this with https://github.com/oracle/graal/pull/3290

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/MethodEntry.java
```diff
@@ -29,13 +29,18 @@
 public class MethodEntry extends MemberEntry {
     TypeEntry[] paramTypes;
     String[] paramNames;
+    String[] paramTypeNames;
```
adinn, line 32: I'm not sure why you are materialising array paramTypeNames here rather than retrieve the names from the respective element of paramTypes. It seems an unnecessary waste of space to avoid a single field indirection. Or is there a good reason for this chnage?

adinn: Ah ok, I see this is because you need to use it to implement Range::getParamTypeNames. Ok.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -724,18 +724,13 @@ private int writeMethodParameterDeclarations(DebugContext context, ClassEntry cl
...
-                if (fileEntry != null) {
-                    pos = writeMethodParameterDeclaration(context, paramName, paramTypeName, false, isSpecification, buffer, pos);
-                } else {
-                    pos = writeMethodParameterDeclaration(context, paramTypeName, paramTypeName, false, isSpecification, buffer, pos);
-                }
+        for (String paramTypeName : range.getParamTypeNames()) {
+            String paramName = uniqueDebugString("");
+            FileEntry fileEntry = range.getFileEntry();
+            if (fileEntry != null) {
+                pos = writeMethodParameterDeclaration(context, paramName, paramTypeName, false, isSpecification, buffer, pos);
```
adinn, line 731: Hmm, this if/else sequence is wrong -- but not because of anything you have done. Passing paramTypeName in place of paramName in the else branch makes no sense at all (paramName is ignored for now). The method call here should include a filename argument or null. This argument should drive the choice of abbrev code as either DW_ABBREV_CODE_method_parameter_declaration2 or DW_ABBREV_CODE_method_parameter_declaration3. It does not matter too much as we don't have a line or column number so a file name for the parameter declaration is not a lot of use. Still it needs fixing, especially as the current code makes no sense at all. I'll raise a separate issue for this.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -195,6 +196,17 @@ protected static ResolvedJavaType getJavaType(HostedType hostedType, boolean wan
         return hostedType.getWrapped().getWrapped();
     }
 
+    protected static ResolvedJavaType getJavaType(AnalysisType analysisType, boolean wantOriginal) {
```
adinn, line 199: Why do you need this? Where are you seeing Analysis types pop up? I think you ought only to be seeing HostedTypes.

zakkak: Hmmm, you are right. I guess I somehow thought I should include it like we do in `computeFullFilePath` (but there it's just a step after unwrapping a `HostedType`). Anyhow this is now removed.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/MethodEntry.java
zakkak: Me neither, I think at the point of adding it I was using it in more places. Removed.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
zakkak: The import was actually already present :/
Fixed.

zakkak:
Thanks for the review @adinn . I have now resolved the issues you pointed out.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
adinn: Well, you should really delete the loop and comment as part of this commit since your existign changes have already made it redundant.

adinn (commented):
This is looking good apart from the eclipse style check failure.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
zakkak: OK, done in https://github.com/oracle/graal/pull/3304/commits/55503f9cc20fe2a80ddbe5adee0f839450eea777

zakkak:
@olpaw could you please review this?

adinn:
@olpaw Any chance of this getting a review from you or someone else who can get it through the gate. I believe the latest changes are sound.

adinn:
@olpaw Any chance of this getting a review from you or someone else who can get it through the gate. I believe the latest changes are sound.

zakkak:
This is now rebased (cleanly) to latest `master`.

@adinn https://github.com/oracle/graal/pull/3304/commits/acfec7abc76e2d14b3a0cf576f2845d6f1dcad07 and https://github.com/oracle/graal/pull/3304/commits/753a5951a766ffd6a7820125ed2826b0d5c384be are new in case you want to have a look at them.

adinn:
I'm not happy with the change to remove the method notifications during presentation of the type info.

Of course, method details presented by NativeImageDebugInstanceTypeInfo::methodInfoProvider and method range details presented later by NativeImageDebugInfo::codeInfoProvider ought to match up 1-1. You should never see range info for a method that has not been advertised using exactly the same classname, method name and signature as belonging to a type and vice versa. However, that's not a reason to drop one stream of data in favour of the other.

The two streams of data are produced from two independent data sets created by the GraalVM image generator which are then transformed by the DebugInfoProvider code to report them using a consistent mapping of HostedXXX objects to their String identifier names. The data sets and the  mapping need to be checked for consistency. It took some time to ensure the right names were used and it would be very easy for a change in the closed world analysis or compilation phase to inadvertently break it.

If you are worried about the performance issues when it comes to linking range info to method info then you can always use sorting and/or hashing to speed up mapping a range to a method. The same approach should also make it quicker to verify the reverse constraint that all methods are accounted for. If that verify step is too expensive then do it in a method called under an assert so it will not get run in production.

zakkak:
Following @adinn 's recommendation https://github.com/oracle/graal/commit/753a5951a766ffd6a7820125ed2826b0d5c384be is now replaced by https://github.com/oracle/graal/pull/3304/commits/40b5e8d7155814ae152964d9ef84029a8eb8a026

zakkak:
CI fails with:

```
There was 1 failure:
1) com.oracle.truffle.api.test.polyglot.PolyglotCachingTest#testSourceFreeContextStrong
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.oracle.truffle.api.test.GCUtils.assertObjectsCollectible(GCUtils.java:88)
	at com.oracle.truffle.api.test.polyglot.PolyglotCachingTest.testSourceFreeContextStrong(PolyglotCachingTest.java:165)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.oracle.mxtool.junit.MxJUnitWrapper$1.schedule(MxJUnitWrapper.java:324)
	at com.oracle.mxtool.junit.MxJUnitWrapper.runRequest(MxJUnitWrapper.java:357)
	at com.oracle.mxtool.junit.MxJUnitWrapper.main(MxJUnitWrapper.java:222)

FAILURES!!!
Tests run: 3983,  Failures: 1
```

I do not see how the changes of this PR could be causing this :/. I think that the failure is irrelevant to the PR, if not please let me know.

adinn:
Hi Foivis,

I am still not clear this is correct. I said this earlier
> method details presented by NativeImageDebugInstanceTypeInfo::methodInfoProvider and method range details presented later by NativeImageDebugInfo::codeInfoProvider ought to match up 1-1

When I said method range I was talking about the *primary* ranges. Your patch seems to interpret my comment as suggesting that every *subrange* whose details are presented under NativeImageDebugInfo::codeInfoProvider will match up with a method presented via NativeImageDebugInstanceTypeInfo::methodInfoProvider. I don't believe that is true. Indeed, your experience with prior inline patches suggests to me that it is false.

So, in class DebugInfoBase at line 266 you have this call

    266             MethodEntry subMethodEntry = subClassEntry.getMethodEntry(debugLineInfo, this);

In the case where the method does not exist you will hit the assert in ClassEntry at line 365

    365         assert false : "Method " + returnTypeName + " " + methodName + "(" + paramSignature + ") not found in ClassEntry of " + getTypeName();

That is not going to be good enough.

In this situation I believe you will need to create a dummy MethodEntry to represent a method that only appears inline. Also, if there is no ClassEntry for the class named in the method details then you will need to create a dummy ClassEntry for a class whose methods only ever appear inline.

These will be needed later on when you want to generate inline method info as they will define the class and method to which the inline details correspond. These dummy entries will need to have DWARF info generated for them but it will be a reduced set compared to the DWARF record we see for classes and methods that are notified via NativeImageDebugInstanceTypeInfo.

zakkak:
> When I said method range I was talking about the _primary_ ranges. Your patch seems to interpret my comment as suggesting that every _subrange_ whose details are presented under NativeImageDebugInfo::codeInfoProvider will match up with a method presented via NativeImageDebugInstanceTypeInfo::methodInfoProvider. I don't believe that is true. 

You have mentioned this concern before in https://github.com/oracle/graal/pull/2880#discussion_r596058717 but I still believe that all the info we need are provided by `NativeImageDebugInstanceTypeInfo::methodInfoProvider`.

> Indeed, your experience with prior inline patches suggests to me that it is false.

Can you please point me to such an example? 
I recall telling you (in a chat) that I hit such a case, but if I am not mistaken that turned out to be due to a different issue.

In any case I prefer this patch going in as is, if you agree that it doesn't break anything at the moment, and we can revisit the _assertion vs insert_ in #2880 where we will be able to verify the assertions by running tests with inlined methods and assertions enabled.

adinn:
@zakkak The problem I have here is that you want to modify this to something that as far as I can tell might break.

That's not just 'might break' as in we'll find out soon enough and fix it. That's 'might break' as in we could release this and only later find out that some users are unable to include debug info because their apps invoke inlined methods which 1) have not been notified by NativeImageDebugInstanceTypeInfo::methodInfoProvideror or 2) belong to a class that has not been notified by NativeImageDebugInstanceTypeInfo::methodInfoProvidertypeInfoProvider. If that happens for an app deployed using this release that needs to be debugged to work out why it is broken then we have a problem.

You say

> we will be able to verify the assertions by running tests with inlined methods and assertions enabled

At the very least we would need to run a comprehensive suite of tests with many different apps (and probably kinds of app) before we can safely assume that. What have you tested this on?

Essentially, I need a good reason to trust this problem is not going to happen. I have studied the Analysis code but I cannot say for sure that whenever an inlined method appears in the CompiledCode a corresponding class and method will have been notified via the DebugTypeInfo interface. I would be happy to accept your patch if an Oracle GraalVM dev could confirm that you are right. Equally if you could explain to me how the Analysis and compilation processes engineer this situation. Otherwise I think you need to allow for missing methods and types.

zakkak:
> At the very least we would need to run a comprehensive suite of tests with many different apps (and probably kinds of app) before we can safely assume that. What have you tested this on?

My primary driver is Quarkus' Integration Tests, but for testing inlined methods I also plan to add explicitly inlined methods to our debug info test.

> Essentially, I need a good reason to trust this problem is not going to happen. I have studied the Analysis code but I cannot say for sure that whenever an inlined method appears in the CompiledCode a corresponding class and method will have been notified via the DebugTypeInfo interface. I would be happy to accept your patch if an Oracle GraalVM dev could confirm that you are right. Equally if you could explain to me how the Analysis and compilation processes engineer this situation. Otherwise I think you need to allow for missing methods and types.

ACK, since that would take me some time (and I am not sure when I will be able to allocate that time :) ) I'll bring back the "insertion" code (unless someone from Oracle happen to confirm my claims before doint it) and come back with a more convincing story :).

zakkak:
I rebased the branch and brought back the code that creates a new MethodEntry in case it fails to find one.
@adinn FYI the new changes are in  https://github.com/oracle/graal/pull/3304/commits/d708a1fc2f81a6ea77c57cefcfac5673fcbd9fc9 , you have already reviewed the rest of the commits (the rebase did not have any conflicts)

adinn (approved):
Ok, this looks good now.

olpaw:
> Ok, this looks good now.

@adinn, @zakkak, sorry I lost track of what was going on with this PR. I will take a look today and tomorrow.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
```diff
@@ -294,7 +294,9 @@ protected void processMethod(DebugMethodInfo debugMethodInfo, DebugInfoBase debu
          * substitution
          */
         FileEntry methodFileEntry = debugInfoBase.ensureFileEntry(fileName, filePath, cachePath);
-        methods.add(new MethodEntry(methodFileEntry, methodName, this, resultType, paramTypeArray, paramNameArray, modifiers));
+        final MethodEntry methodEntry = new MethodEntry(methodFileEntry, methodName, this, resultType, paramTypeArray, paramNameArray, modifiers);
```
olpaw: We generally do use `final` only for fields (on locals it's mostly verbose). But you can keep it if you like.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -958,11 +981,33 @@ public int getFrameSize() {
...
+            return name().endsWith(HostedMethod.METHOD_NAME_DEOPT_SUFFIX);
         }
 
         @Override
-        public int getModifiers() {
+        public List<String> paramTypes() {
+            LinkedList<String> paramTypes = new LinkedList<>();
+            Signature signature = hostedMethod.getSignature();
+            for (int i = 0; i < signature.getParameterCount(false); i++) {
+                final JavaType parameterType = signature.getParameterType(i, null);
```
olpaw, line 992: See https://github.com/oracle/graal/pull/3304#discussion_r627286481

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -958,11 +976,33 @@ public int getFrameSize() {
 
         @Override
         public boolean isDeoptTarget() {
-            return methodName().endsWith(HostedMethod.METHOD_NAME_DEOPT_SUFFIX);
+            return name().endsWith(HostedMethod.METHOD_NAME_DEOPT_SUFFIX);
+        }
+
+        @Override
+        public List<String> paramTypes() {
+            LinkedList<String> paramTypes = new LinkedList<>();
```
olpaw, line 989: What's the point of using linked list (here and elsewhere). Using linked list is almost always a bad idea.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -958,11 +976,33 @@ public int getFrameSize() {
...
+                final JavaType parameterType = signature.getParameterType(i, null);
+                paramTypes.add(toJavaName(parameterType));
+            }
+            return paramTypes;
         }
 
         @Override
-        public int getModifiers() {
+        public List<String> paramNames() {
+            /* Can only provide blank names for now. */
```
olpaw, line 1000: How about
```java
String[] paramNames = new String[signature.getParameterCount(false)] ;
Arrays.fill(paramNames, "");
return Arrays.asList(paramNames);
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1057,11 +1106,26 @@ public String methodName() {
...
+        }
+
         @Override
         public String symbolNameForMethod() {
             return NativeImage.localSymbolNameForMethod(method);
         }
 
+        @Override
+        public boolean isDeoptTarget() {
+            return name().endsWith(HostedMethod.METHOD_NAME_DEOPT_SUFFIX);
```
olpaw, line 1126: How about `hostedMethod.compilationInfo.isDeoptTarget()` instead?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1081,6 +1145,33 @@ public int line() {
...
+            return paramTypes;
+        }
+
+        @Override
+        public List<String> paramNames() {
+            /* Can only provide blank names for now. */
+            LinkedList<String> paramNames = new LinkedList<>();
+            Signature signature = method.getSignature();
+            for (int i = 0; i < signature.getParameterCount(false); i++) {
+                paramNames.add("");
```
olpaw, line 1165: See https://github.com/oracle/graal/pull/3304/files#r627316683

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
```diff
@@ -335,27 +338,52 @@ public ClassEntry getSuperClass() {
...
-            if (fileEntryToUse == null) {
-                /* Last chance is the class's file entry. */
-                fileEntryToUse = this.fileEntry;
+        }
+        MethodEntry newMethodEntry = processMethod(debugMethodInfo, debugInfoBase, debugContext);
+        methodIterator.add(newMethodEntry);
+        return newMethodEntry;
+    }
+
+    private boolean listIsSorted(List<MethodEntry> list) {
```
olpaw, line 371: Since this just for assertion checking how about:
```java
list.stream().sorted().collect(Collectors.toList()).equals(list);
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/MethodEntry.java
```diff
@@ -72,26 +74,51 @@ public String getParamName(int idx) {
         return paramNames[idx];
     }
 
-    public boolean match(String methodName, String paramSignature, String returnTypeName) {
-        if (!methodName.equals(this.memberName)) {
-            return false;
+    public int compareTo(String methodName, String paramSignature, String returnTypeName) {
+        if (!memberName.equals(methodName)) {
```
olpaw, line 78: I would build this up as
```java
int memberNameCompare = memberName.compareTo(methodName);
if (memberNameCompare != 0) {
    return memberNameCompare;
}
int valueTypeCompare = valueType.getTypeName().compareTo(returnTypeName);
if (valueTypeCompare != 0) {
... etc etc ...
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/MethodEntry.java
```diff
@@ -72,26 +74,51 @@ public String getParamName(int idx) {
...
+    }
+
+    public boolean isDeoptTarget() {
+        return isDeoptTarget;
+    }
+
+    @Override
+    public int compareTo(MethodEntry other) {
+        assert other != null;
+        int result = methodName().compareTo(other.methodName());
```
olpaw, line 107: This could also profit from the strategy shown in https://github.com/oracle/graal/pull/3304/files#r627333745

olpaw (changes requested):
See comments. Please also remove final from locals where it is obvious that the local is only assigned once.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/MethodEntry.java
zakkak: I personally prefer this over https://github.com/oracle/graal/pull/3304/files#r627333745

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
zakkak: There's no excuse for it... Thanks for catching it!

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
zakkak: Sure, but at some point `paramNames` will not be `""`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
zakkak: @adinn this was copied from your implementation, do you have any objections on changing it to `hostedMethod.isDeoptTarget()`?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/MethodEntry.java
zakkak: Refactored, but as:
```
int result = memberName.compareTo(methodName);
if (result == 0) {
    result = valueTypeCompare = valueType.getTypeName().compareTo(returnTypeName);
}
if (result == 0) {
    ... etc etc ...
return result;
```

instead of as suggested. I find this equally readable while a bit more compact.

zakkak:
Thanks for the review @olpaw ! I have addressed most of your comments (for the rest please see my replies).

Could you please also provide some info regarding the following comment?

> Your patch seems to interpret my comment as suggesting that every subrange whose details are presented under `NativeImageDebugInfo::codeInfoProvider` will match up with a method presented via `NativeImageDebugInstanceTypeInfo::methodInfoProvider`. I don't believe that is true.

Source: https://github.com/oracle/graal/pull/3304#issuecomment-816780087

The `NativeImageDebugInstanceTypeInfo::methodInfoProvider` essentially gets all types present in the "universe" through `heap.getUniverse().getTypes()` and for each one of them it then gets the declaredMethods through `hostedType.getAllDeclaredMethods()`.  While `NativeImageDebugInfo::codeInfoProvider` gets all the compilations through `codeCache.compilations.entrySet()` and for each one of them it then gets the methods embedded in it through `compilation.getSourceMappings()`.

To my understanding the set of methods returned by the `methodInfoProvider` is a superset of the set of methods returned by the `codeInfoProvider` but @adinn is skeptical about it. We would really appreciate it if you could help us understand what's the case (and why if possible).

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/MethodEntry.java
olpaw: Hmm, so if the first assignment to `result` returns a value `!=` `0` you will have have a couple more `if`s to go though before finally returning that value assigned in the beginning although it was clear from the beginning that this is the value that was going to be returned? :thinking:

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: > do you have any objections on changing it to hostedMethod.isDeoptTarget()

Not at all. I was not aware that such a method existed when I coded my version. It is far better to use the API method call than rely on the internal details of how deopt method names are encoded.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/MethodEntry.java
olpaw: Of course the compiled version is likely to have optimized early exits. But while it runs in the interpreter it does exactly what you ask it to do. Also I find it easier to reason about code where locals are all effectively final (which is not the case in your variant)

Anyway, if you prefer this style I won't prevent you from merging it like that. I'm just providing my 2-cents.

zakkak:
@olpaw this is rebased to master and now resolves all your comments (except the one about the empty parameter names).
The CI failure seems unrelated to the PR.
It should be ready for merge :)

olpaw:
@zakkak I just saw that there other places in the debuginfo sources that use `LinkedList` (e.g. `com.oracle.objectfile.debugentry.ClassEntry#ClassEntry`). Please go over it and replace them with `ArrayList` whereever possible.

olpaw:
> I will do some testing to see if I can find examples to support or disprove this theory that I can present tomorrow.

@zakkak if you are right then `com.oracle.svm.hosted.image.NativeImageDebugInfoProvider.NativeImageDebugLineInfo#ownerType` should always successfully cast to `HostedMethod` and the else case should actually be an assertion. Do you agree?
```java
@Override
public String ownerType() {
    if (method instanceof HostedMethod) {
        return getJavaType((HostedMethod) method, true).toJavaName();
    } else {
        assert false: "NodeSourcePosition.getMethod() should always be a HostedMethod"; // <<<<<<<<<<
        return method.getDeclaringClass().toJavaName();
    }
}
```

zakkak:
> @zakkak I just saw that there are other places in the debuginfo sources that use `LinkedList` (e.g. `com.oracle.objectfile.debugentry.ClassEntry#ClassEntry`). Please go over it and replace them with `ArrayList` whereever possible.

Resolved in https://github.com/oracle/graal/pull/3304/commits/8ed65a8bd1d6eab42d18b0e9c501b71183e5e3ac

> Actually assuming to be always a `HostedMethod` is incorrect. But assuming it to be `com.oracle.svm.core.meta.SharedMethod` would be correct (`SharedMethod` is what you see if you have Graal running at image-runtime).

I am not sure TBH. The way I tested my hypothesis was by adding an assertion in `com.oracle.objectfile.debugentry.ClassEntry#getMethodEntry` where we essentially search for a method entry (in the method entry set provided by the `methodInfoProvider`) while parsing the data provided by the `codeInfoProvider`. If the search fails and we need to add a new method (`com/oracle/objectfile/debugentry/ClassEntry.java:370`) then the assert should hit and disprove my hypothesis. My testing so far has failed to hit that assert, but I am not confident about the coverage of my testing :)

olpaw (changes requested):
See https://github.com/oracle/graal/pull/3304#discussion_r629947799

olpaw:
`mx eclipseformat` found another style issue. I will fix that on my side.

zakkak:
> That said, assuming that you will find all HostedMethods that exist after aot-compilation only by iterating the hosted universe (heap.getUniverse().getTypes() and then going though each type with hostedType.getAllDeclaredMethods()) is not valid. E.g. method specialization can result in having multiple versions of a HostedMethod where you would only see a single one if you just iterated through the hosted universe. Debuginfo generation needs to be able to handle that. E.g. method specialization can result in having multiple versions of a HostedMethod where you would only see a single one if you just iterated through the hosted universe.

That makes sense, but I think it doesn't answer the question we are interested in, possibly because I didn't ask properly so please let me rephrase.

Our goal here is to be able to map an address range to a debuginfo method-declaration entry where the debuginfo method-declaration entry essentially contains info about where (in the source code) the method is defined, what its type is, etc.    
The `SourceMapping`s help us do this by mapping address ranges (`sourceMapping.getStartOffset()`--`sourceMapping.getEndOffset()`) to `ResolvedJavaMethod`s. Then, through the `ResolvedJavaMethod`s we are able to get access to the corresponding source Java method and construct a debug method entry.

The question is if there is any chance to find a Java method through the `SourceMapping`s that we have not seen before through the `methodInfoProvider`. 

More practically, do you think there is any chance to reach https://github.com/zakkak/mandrel/blob/0ea21026b4507d2d29908093b548a9e47361347b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java#L370 since we create method entries for each method of each class in the "universe" in https://github.com/zakkak/mandrel/blob/0ea21026b4507d2d29908093b548a9e47361347b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java#L141 ?

> method specialization can result in having multiple versions of a HostedMethod where you would only see a single one if you just iterated through the hosted universe

This single one though should be able to provide me with enough info to generate the method-declaration entry, right?

> `mx eclipseformat` found another style issue. I will fix that on my side.

Oh the comment is too long...
Thanks for taking care of it.

olpaw:
> More practically, do you think there is any chance to reach https://github.com/zakkak/mandrel/blob/0ea21026b4507d2d29908093b548a9e47361347b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java#L370 since we create method entries for each method of each class in the "universe" in https://github.com/zakkak/mandrel/blob/0ea21026b4507d2d29908093b548a9e47361347b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java#L141 ?

Right, unless something dodgy is going on in the backend you should not hit the `newMethodEntry`-part of `com.oracle.objectfile.debugentry.ClassEntry#getMethodEntry`.

> 
> > method specialization can result in having multiple versions of a HostedMethod where you would only see a single one if you just iterated through the hosted universe
> 
> This single one though should be able to provide me with enough info to generate the method-declaration entry, right?

Yes.

olpaw:
@zakkak all gates but one are passing internally. And that failing gate has nothing to do with your PR. We are working on getting this sorted. In other words ... it might take a few days before you see your changes showing up on master.

zakkak:
> @zakkak all gates but one are passing internally. And that failing gate has nothing to do with your PR. We are working on getting this sorted. In other words ... it might take a few days before you see your changes showing up on master.

OK, thanks for the update Paul.

PR #3315, opened 2021-03-29
Improve support for resource registration

jovanstevanovic:
Task and high-level overview (#3326)

**PROGRESS**: 
- First step toward creating a custom filesystem with a TODO list is committed.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKVersionSpecificResourceBuilder.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKVersionSpecificResourceBuilder.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2021, 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.svm.core.jdk;
+
+import java.net.URL;
+import java.net.URLConnection;
+
+public class JDKVersionSpecificResourceBuilder {
+
+    @SuppressWarnings("unused")
+    public static Object buildResource(String name, URL url, URLConnection urlConnection) {
+        return null;
+    }
+}
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
@@ -29,11 +29,9 @@ import static com.oracle.svm.core.annotate.RecomputeFieldValue.Kind.Reset;
 import static com.oracle.svm.core.snippets.KnownIntrinsics.readHub;
 
 import java.io.File;
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
 import java.net.URL;
-import java.util.Enumeration;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
@@ -724,28 +722,6 @@ final class Target_jdk_internal_loader_BootLoader {
     @Substitute
     private static boolean hasClassPath() {
         return true;
-    }
-
-    @SuppressWarnings("unused")
-    @Substitute
-    private static URL findResource(String mn, String name) {
-        return ClassLoader.getSystemClassLoader().getResource(name);
-    }
-
-    @SuppressWarnings("unused")
-    @Substitute
-    private static InputStream findResourceAsStream(String mn, String name) {
-        return ClassLoader.getSystemClassLoader().getResourceAsStream(name);
-    }
-
-    @Substitute
-    private static URL findResource(String name) {
-        return ClassLoader.getSystemClassLoader().getResource(name);
-    }
-
-    @Substitute
-    private static Enumeration<URL> findResources(String name) throws IOException {
-        return ClassLoader.getSystemClassLoader().getResources(name);
     }
 
     /**
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_lang_ClassLoader.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_lang_ClassLoader.java
@@ -24,13 +24,13 @@
  */
 package com.oracle.svm.core.jdk;
 
-import java.io.ByteArrayInputStream;
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.net.URLConnection;
 import java.security.ProtectionDomain;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -58,8 +58,75 @@ import com.oracle.svm.core.util.VMError;
 import jdk.vm.ci.meta.MetaAccessProvider;
 import jdk.vm.ci.meta.ResolvedJavaField;
 
+@TargetClass(className = "jdk.internal.loader.Resource", onlyWith = JDK11OrLater.class)
+@SuppressWarnings("unused")
+final class Target_jdk_internal_loader_Resource_JDK11OrLater {
+
+}
+
+@TargetClass(className = "sun.misc.Resource", onlyWith = JDK8OrEarlier.class)
+@SuppressWarnings("unused")
+final class Target_sun_misc_Resource_JDK8OrEarlier {
+
+}
+
+@SuppressWarnings("unchecked")
+class ResourcesHelper {
+
+    private static <T> T urlToResource(String resourceName, URL url) {
+        try {
+            if (url == null) {
+                return null;
+            }
+            URLConnection urlConnection = url.openConnection();
+            Object resource = JDKVersionSpecificResourceBuilder.buildResource(resourceName, url, urlConnection);
+            VMError.guarantee(resource != null);
+            return (T) resource;
+        } catch (IOException e) {
+            return null;
+        } catch (ClassCastException classCastException) {
+            throw VMError.shouldNotReachHere(classCastException);
+        }
+    }
+
+    static <T> T nameToResource(String resourceName) {
+        return urlToResource(resourceName, nameToResourceURL(resourceName));
+    }
+
+    static <T> Enumeration<T> nameToResources(String resourceName) {
+        Enumeration<URL> urls = Resources.createURLs(resourceName);
+        List<T> resourceURLs = new ArrayList<>();
+        while (urls.hasMoreElements()) {
+            resourceURLs.add(urlToResource(resourceName, urls.nextElement()));
+        }
+        return Collections.enumeration(resourceURLs);
+    }
+
+    static URL nameToResourceURL(String resourceName) {
+        return Resources.createURL(resourceName);
+    }
+
+    static InputStream nameToResourceInputStream(String resourceName) throws IOException {
+        URL url = nameToResourceURL(resourceName);
+        return url != null ? url.openStream() : null;
+    }
+
+    static List<URL> nameToResourceListURLs(String resourcesName) {
+        Enumeration<URL> urls = Resources.createURLs(resourcesName);
+        List<URL> resourceURLs = new ArrayList<>();
+        while (urls.hasMoreElements()) {
+            resourceURLs.add(urls.nextElement());
+        }
+        return resourceURLs;
+    }
+
+    static Enumeration<URL> nameToResourceEnumerationURLs(String resourcesName) {
+        return Collections.enumeration(nameToResourceListURLs(resourcesName));
+    }
+}
+
 @TargetClass(classNameProvider = Package_jdk_internal_loader.class, className = "URLClassPath")
-@SuppressWarnings("static-method")
+@SuppressWarnings({"unused", "static-method"})
 final class Target_jdk_internal_loader_URLClassPath {
 
     /* Reset fields that can store a Zip file via sun.misc.URLClassPath$JarLoader.jar. */
@@ -91,18 +158,63 @@ final class Target_jdk_internal_loader_URLClassPath {
     @Delete
     @TargetElement(onlyWith = JDK8OrEarlier.class)
     private static native boolean knownToNotExist0(ClassLoader loader, String className);
+
+    @Substitute
+    public URL findResource(String name, boolean check) {
+        return Resources.createURL(name);
+    }
+
+    @Substitute
+    public Enumeration<URL> findResources(final String name, final boolean check) {
+        return Resources.createURLs(name);
+    }
+
+    @Substitute
+    @TargetElement(name = "getResource", onlyWith = JDK11OrLater.class)
+    public Target_jdk_internal_loader_Resource_JDK11OrLater getResourceJDK11OrLater(String name, boolean check) {
+        return ResourcesHelper.nameToResource(name);
+    }
+
+    @Substitute
+    @TargetElement(name = "getResources", onlyWith = JDK11OrLater.class)
+    public Enumeration<Target_jdk_internal_loader_Resource_JDK11OrLater> getResourcesJDK11OrLater(final String name,
+                    final boolean check) {
+        return ResourcesHelper.nameToResources(name);
+    }
+
+    @Substitute
+    @TargetElement(name = "getResource", onlyWith = JDK8OrEarlier.class)
+    public Target_sun_misc_Resource_JDK8OrEarlier getResourceJDK8OrEarlier(String name, boolean check) {
+        return ResourcesHelper.nameToResource(name);
+    }
+
+    @Substitute
+    @TargetElement(name = "getResources", onlyWith = JDK8OrEarlier.class)
+    public Enumeration<Target_sun_misc_Resource_JDK8OrEarlier> getResourcesJDK8OrEarlier(final String name,
+                    final boolean check) {
+        return ResourcesHelper.nameToResources(name);
+    }
 }
 
 @TargetClass(URLClassLoader.class)
-@SuppressWarnings("static-method")
+@SuppressWarnings({"unused", "static-method"})
 final class Target_java_net_URLClassLoader {
     @Alias @RecomputeFieldValue(kind = Kind.NewInstance, declClass = WeakHashMap.class)//
     private WeakHashMap<Closeable, Void> closeables;
 
     @Substitute
-    private InputStream getResourceAsStream(String name) {
-        List<byte[]> arr = Resources.get(name);
-        return arr == null ? null : new ByteArrayInputStream(arr.get(0));
+    public InputStream getResourceAsStream(String name) throws IOException {
+        return ResourcesHelper.nameToResourceInputStream(name);
+    }
+
+    @Substitute
+    public URL findResource(final String name) {
+        return ResourcesHelper.nameToResourceURL(name);
+    }
+
+    @Substitute
+    public Enumeration<URL> findResources(final String name) {
+        return ResourcesHelper.nameToResources(name);
     }
 
     @Substitute
@@ -113,38 +225,47 @@ final class Target_java_net_URLClassLoader {
 }
 
 @TargetClass(className = "jdk.internal.loader.BuiltinClassLoader", onlyWith = JDK11OrLater.class)
-@SuppressWarnings("static-method")
+@SuppressWarnings({"unused", "static-method"})
 final class Target_jdk_internal_loader_BuiltinClassLoader {
 
     @Substitute
-    public URL findResource(@SuppressWarnings("unused") String mn, String name) {
-        List<byte[]> arr = Resources.get(name);
-        return arr == null ? null : Resources.createURL(name, arr.get(0));
+    public URL findResource(String mn, String name) {
+        return ResourcesHelper.nameToResourceURL(name);
+    }
+
+    @Substitute
+    public InputStream findResourceAsStream(String mn, String name) throws IOException {
+        return ResourcesHelper.nameToResourceInputStream(name);
     }
 
     @Substitute
     public URL findResource(String name) {
-        List<byte[]> arr = Resources.get(name);
-        return arr == null ? null : Resources.createURL(name, arr.get(0));
-    }
-
-    @Substitute
-    public InputStream findResourceAsStream(@SuppressWarnings("unused") String mn, String name) {
-        List<byte[]> arr = Resources.get(name);
-        return arr == null ? null : new ByteArrayInputStream(arr.get(0));
+        return ResourcesHelper.nameToResourceURL(name);
     }
 
     @Substitute
     public Enumeration<URL> findResources(String name) {
-        List<byte[]> arr = Resources.get(name);
-        if (arr == null) {
-            return Collections.emptyEnumeration();
-        }
-        List<URL> res = new ArrayList<>(arr.size());
-        for (byte[] data : arr) {
-            res.add(Resources.createURL(name, data));
-        }
-        return Collections.enumeration(res);
+        return ResourcesHelper.nameToResourceEnumerationURLs(name);
+    }
+
+    @Substitute
+    private List<URL> findMiscResource(String name) {
+        return ResourcesHelper.nameToResourceListURLs(name);
+    }
+
+    @Substitute
+    private URL findResource(Target_java_lang_module_ModuleReference mref, String name) {
+        return ResourcesHelper.nameToResourceURL(name);
+    }
+
+    @Substitute
+    private URL findResourceOnClassPath(String name) {
+        return ResourcesHelper.nameToResourceURL(name);
+    }
+
+    @Substitute
+    private Enumeration<URL> findResourcesOnClassPath(String name) {
+        return ResourcesHelper.nameToResourceEnumerationURLs(name);
     }
 }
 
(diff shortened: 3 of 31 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
```diff
@@ -708,27 +709,17 @@ private static boolean hasClassPath() {
         return true;
     }
 
-    @SuppressWarnings("unused")
-    @Substitute
-    private static URL findResource(String mn, String name) {
-        return ClassLoader.getSystemClassLoader().getResource(name);
-    }
+    @AnnotateOriginal
```
christianwimmer, line 712: What is the `AnnotateOriginal` for? You are not adding any annotations, so this is not going to do anything

christianwimmer (commented):
So this PR removes some substitutions that are not necessary because the JDK code eventually calls a method that remains substituted. I'm curious - where is the actual "root" in the JDK that remains substituted?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributes.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public Object fileKey() {
+        return null;
+    }
+
+    public String toString() {
+        StringBuilder sb = new StringBuilder(1024);
+        Formatter fm = new Formatter(sb);
+        fm.format("    creationTime    : null%n");
```
vjovanov, line 63: What will Java return here?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributesView.java
```diff
@@ -0,0 +1,117 @@
...
+    public ResourceAttributesView(ResourcePath path, boolean isBasic) {
+        this.path = path;
+        this.isBasic = isBasic;
+    }
+
+    @SuppressWarnings("unchecked")
+    static <V extends FileAttributeView> V get(ResourcePath path, Class<V> type) {
+        if (type == null)
+            throw new NullPointerException();
+        if (type == BasicFileAttributeView.class)
```
vjovanov, line 38: Always use {}. This style is deprecated for a long time.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributesView.java
```diff
@@ -0,0 +1,117 @@
...
+
+    Map<String, Object> readAttributes(String attributes) throws IOException {
+        ResourceAttributes resourceAttributes = readAttributes();
+        LinkedHashMap<String, Object> map = new LinkedHashMap<>();
+        if ("*".equals(attributes)) {
+            for (AttributeID id : AttributeID.values()) {
+                try {
+                    map.put(id.name(), attribute(id, resourceAttributes));
+                } catch (IllegalArgumentException x) {
+                    // Silently handle this exception.
```
vjovanov, line 73: Hmmmm, this is almost always a bad idea. When does this happen?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributes.java
jovanstevanovic: In current implementation the creationTime, lastAccessTime, lastModifiedTime is not supported (returning null). We will need to support that as well, I'll add TODO.

vjovanov: yup, just do the same thing Java would return here.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributesView.java
```diff
@@ -0,0 +1,123 @@
...
+import java.nio.file.attribute.BasicFileAttributeView;
+import java.nio.file.attribute.FileAttributeView;
+import java.nio.file.attribute.FileTime;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class ResourceAttributesView implements BasicFileAttributeView {
+
+    private enum AttributeID
+    {
```
vjovanov, line 14: Please set up the formatting according to the IDEs.md doc.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributesView.java
vjovanov: Can we use something other than exceptions for control flow?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceFileSystemProvider.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * ResourcesFileSystemProvider is a core class that supports building a custom file system on top of
+ * resources in the native image.
+ *
+ * <p>
+ * Resources are collected in build time in {@link ResourcesFeature} and stored in a hash map in
```
vjovanov, line 32: at build time

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceFileSystemProvider.java
```diff
@@ -0,0 +1,212 @@
...
+ * {@link Resources} class. We are supporting only resources that are on the classpath.
```
vjovanov, line 33: Why only resources in the classpath?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceFileSystemProvider.java
```diff
@@ -0,0 +1,212 @@
...
+import java.nio.file.attribute.BasicFileAttributes;
+import java.nio.file.attribute.FileAttribute;
+import java.nio.file.attribute.FileAttributeView;
+import java.nio.file.spi.FileSystemProvider;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * ResourcesFileSystemProvider is a core class that supports building a custom file system on top of
```
vjovanov, line 28: Use doc links for ResourcesFileSystemProvider.

vjovanov: supports is superfluous, right? It always builds it.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceFileSystemProvider.java
jovanstevanovic: Now, we are scanning only the resource of the classpath. Eventually, when we finish the resource system, it will catch all the resources.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributesView.java
jovanstevanovic: We can try with an if-else statement instead of a try-catch. Sounds okay?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/FileSystemProviderSupport.java
```diff
@@ -132,6 +132,8 @@ public void afterRegistration(AfterRegistrationAccess access) {
              * during image generation.
              */
             installedProviders.addAll(FileSystemProvider.installedProviders());
+            // TODO: Register filesystem provider through ServiceLoader.
+            installedProviders.add(new ResourceFileSystemProvider());
```
jovanstevanovic, line 136: NativeImageResourceFileSystemProvider

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributes.java
```diff
@@ -0,0 +1,84 @@
...
+    public boolean isOther() {
+        return false;
+    }
+
+    @Override
+    public long size() {
+        return entry.size();
+    }
+
+    // TODO: Need further thinking.
```
jovanstevanovic, line 61: Null same as JarFSP.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/FileSystemProviderSupport.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributes.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceFileSystemProvider.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributes.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributesView.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributesView.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceFileSystemProvider.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/FileSystemProviderSupport.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributes.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceAttributesView.java
jovanstevanovic: We are now checking in an if-else statement if the specified attribute is in an attribute set instead of throwing an exception.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceStorageEntry.java
```diff
@@ -0,0 +1,31 @@
+package com.oracle.svm.core.jdk;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ResourceStorageEntry {
```
vjovanov, line 6: All of these should go into a separate package.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/ByteArrayChannel.java
```diff
@@ -0,0 +1,233 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.ClosedChannelException;
+import java.nio.channels.NonWritableChannelException;
+import java.nio.channels.SeekableByteChannel;
+import java.util.Arrays;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+public class ByteArrayChannel implements SeekableByteChannel {
```
vjovanov, line 12: Can't we use the JDK 11 class?

vjovanov: And move this into JDK8 so when we stop support, we can delete it.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
```diff
@@ -0,0 +1,1206 @@
+package com.oracle.svm.core.jdk.resources;
```
vjovanov, line 1: Copyright header is missing in all new files.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/NativeImageResourceFileSystemProviderTest.java
```diff
@@ -0,0 +1,302 @@
...
+                ByteBuffer byteBuffer = ByteBuffer.allocate((int) channel.size());
+                channel.read(byteBuffer);
+                String content = new String(byteBuffer.array());
+                Assert.assertTrue("Nothing has been read from new file!", content.length() > 0);
+            }
+        } catch (IOException ioException) {
+            Assert.fail("Exception occurs during copying file into new directory!");
+        }
+
+        // 6. Moving file to newly create directory.
```
vjovanov, line 178: It is better to have 6 different tests.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/ByteArrayChannel.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ResourceStorageEntry.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
```diff
@@ -68,6 +73,35 @@
...
+ * is open.
+ * </p>
+ *
+ * <p>
+ * Most of operations mentioned above aren't accessed directly i.e. by calling methods from this
+ * file system provider, however their are called indirectly from accessing method from
+ * {@link java.nio.file.Files}.
+ * </p>
+ *
+ * @author jovanstevanovic
```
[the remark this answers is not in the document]
jovanstevanovic: Removed.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/NativeImageResourceFileSystemProviderTest.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core.jdk8/src/com/oracle/svm/core/jdk/resources/ByteArrayChannel.java
```diff
@@ -0,0 +1,254 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.ClosedChannelException;
+import java.nio.channels.NonWritableChannelException;
+import java.nio.channels.SeekableByteChannel;
+import java.util.Arrays;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+public class ByteArrayChannel implements SeekableByteChannel {
```
vjovanov, line 37: Can we simply ignore the JDK 8 implementation? Would it still work for the set of programs that worked before this PR?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKVersionSpecificResourceBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.svm.core.jdk;
+
+import java.net.URL;
+import java.net.URLConnection;
+
+public class JDKVersionSpecificResourceBuilder {
```
vjovanov, line 31: Then we could skip all of this as well.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/ResourceStorageEntry.java
```diff
@@ -0,0 +1,56 @@
...
+    public ResourceStorageEntry(boolean isDirectory) {
+        this.isDirectory = isDirectory;
+        this.data = new ArrayList<>();
+    }
+
+    public boolean isDirectory() {
+        return isDirectory;
+    }
+
+    public void setDirectory(boolean directory) {
```
gradinac, line 45: This and `setData` don't seem to be used - I would remove them

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
```diff
@@ -68,6 +74,35 @@
...
+ * {@link Resources} class.
+ * </p>
+ *
+ * <p>
+ * {@link NativeImageResourceFileSystemProvider } is a core class for building a custom file system
+ * on top of resources in the native image.
+ * </p>
+ *
+ * <p>
+ * ResourceFileSystemProvider exposes methods for creating a new file system, getting existing one
```
gradinac, line 89: I would tweak this comment a bit:
```
The {@link NativeImageResourceFileSystemProvider} provides most of the functionality of a {@link FileSystem}. It is an in-memory file system that upon creation contains a copy of the resources included in the native-image.
Note that changes to files do not affect actual resources returned by resource manipulation methods like  `Class.getResource`. Upon being closed, all changes are discarded.
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
```diff
@@ -68,6 +74,35 @@
...
+ * or doing operations like getting file attributes, getting path appropriate for this file system,
+ * checking access rights, etc. Operations like moving files, copying files, deleting files, setting
+ * file attributes change the current state of file system and, these changes will persist till file
+ * system is open.
+ * </p>
+ *
+ * <p>
+ * Most of operations mentioned above aren't accessed directly i.e. by calling methods from this
```
gradinac, line 97: I would tweak this one as well:
```
As with other file system providers, these methods provide a low-level interface and are not meant for direct usage - see {@link java.nio.file.Files}
```

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
```diff
@@ -0,0 +1,1217 @@
...
+public class NativeImageResourceFileSystem extends FileSystem {
+
+    private static final String GLOB_SYNTAX = "glob";
+    private static final String REGEX_SYNTAX = "regex";
+
+    private static final Set<String> supportedFileAttributeViews = Collections.unmodifiableSet(
+                    new HashSet<>(Arrays.asList("basic", "resource")));
+
+    private final Set<InputStream> streams = Collections.synchronizedSet(new HashSet<>());
+    private final Set<Path> tmpPaths = Collections.synchronizedSet(new HashSet<Path>());
```
gradinac, line 106: `<Path>` is redundant here -> `<>`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
```diff
@@ -0,0 +1,1217 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.regex.Pattern;
+
+import org.graalvm.collections.MapCursor;
+
+/**
+ * <p>
+ * Some parts of code from this class are copied from jdk.nio.zipfs.ZipFileSystem with small tweaks.
+ * The main reason why we cannot reuse this class (without any code copying) is that we cannot
+ * extend jdk.nio.zipfs.ZipPath which is central class in Zip file system.
```
gradinac, line 94: is central class in Zip file system -> is the central class in the Zip file system

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileStore.java
```diff
@@ -0,0 +1,131 @@
...
+        throw new UnsupportedOperationException("Attribute isn't supported!");
+    }
+
+    private static class ResourceFileStoreAttributes {
+
+        private final FileStore fileStore;
+        private final long size;
+
+        ResourceFileStoreAttributes(NativeImageResourceFileStore fileStore) throws IOException {
+            Path path = FileSystems.getDefault().getPath(fileStore.name());
```
gradinac, line 114: If I understood correctly, a `NativeImageResourceFileStore` is bound to a path inside of the native-image file system. However, here we seem to be resolving that path in the default file system - I may have missed something, but it seems like this file isn't connected to the native-image file system and the information here, provided that such a path exists, most likely isn't what we're looking to return

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_lang_ClassLoader.java
```diff
@@ -229,46 +331,6 @@ public static ClassLoader getSystemClassLoader() {
     @Delete
     private static native void initSystemClassLoader();
 
-    @Substitute
-    private URL getResource(String name) {
```
gradinac, line 233: To make sure I understand these correctly, we've pushed these substitutions to the appropriate class loaders - we now respect the class loader hierarchy when looking up resources

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
```diff
@@ -0,0 +1,1217 @@
...
+            }
+            Entry eDst = getEntry(dst);
+            if (eDst != null) {
+                if (!hasReplace) {
+                    throw new FileAlreadyExistsException(getString(dst));
+                }
+            } else {
+                checkParents(dst);
+            }
+            Entry target = new Entry(eSrc, Entry.COPY);  // Copy eSrc entry.
```
gradinac, line 518: When we do a copy of the entry, we may either create a copy of the original entry's bytes, or, if the previous entry is getting deleted, take the original entry's bytes directly. In both cases, however, `copyOnWrite` remains false - if we attempt to write data to this entry (using `getBytes`), it seems like we will accidentally reload the entry - should we also flip the `copyOnWrite` flag to true here?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
```diff
@@ -0,0 +1,1217 @@
...
+    private final LinkedHashMap<IndexNode, IndexNode> inodes = new LinkedHashMap<>(10);
+
+    public NativeImageResourceFileSystem(NativeImageResourceFileSystemProvider provider, Path resourcePath, Map<String, ?> env) {
+        this.provider = provider;
+        this.resourcePath = resourcePath;
+        this.root = new NativeImageResourcePath(this, new byte[]{'/'});
+        if (!isTrue(env, "create")) {
+            throw new FileSystemNotFoundException(resourcePath.toString());
+        }
+        this.useTempFile = isTrue(env, "useTempFile");
```
gradinac, line 128: If I understood it correctly, this option allows us to create temporary files when modifying a file system entry? 
Would such a flag be required? This seems to only be used when attempting to create new output streams, but it may help reduce complexity if we remove it and keep everything in memory

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
```diff
@@ -0,0 +1,1217 @@
...
+            e.lastModifiedTime = System.currentTimeMillis();
+        }
+        OutputStream os;
+        if (useTempFile) {
+            e.file = getTempPathForEntry(null);
+            os = Files.newOutputStream(e.file, WRITE);
+        } else {
+            os = new ByteArrayOutputStream((e.size > 0) ? e.size : 8192);
+        }
+        return new EntryOutputStream(e, os);
```
gradinac, line 731: If we keep everything in memory, would it make sense to also close the output streams when closing the file system?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/ResourceStorageEntry.java
jovanstevanovic: Agree.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_lang_ClassLoader.java
jovanstevanovic: Correct. That was the idea at some point, but due to a lack of module support, I was forced to return almost all substitutions. Once we get full module support, we will surely push substitutions to appropriate class loaders.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
jovanstevanovic: Correct. No, it's not required, we can get rid of this flag and simplify the operation of creating a new output stream and keep everything in memory.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
jovanstevanovic: Agree. I overlooked it. We are keeping all information about opened input streams, so we should do the same thing for output streams.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystem.java
jovanstevanovic: Nice catch. That could cause a bug in the future.

PR #3336, opened 2021-04-10
Retrieve resource bytes from map as URLStreamHandler can be shared

### LINKED ISSUE #2291: Nonexistent relative classpath resource appears to exist in native image
# Background

Spring Framework provides a `Resource` abstraction for classpath resources that supports creation of relative resources. The `Resource` abstraction also provides support to determine if any given `Resource` _exists_ based on various techniques for detecting the presence of such a resource. For a Spring `UrlResource`, the `exists()` method is inherited from `AbstractFileResolvingResource` as can be seen below.

https://github.com/spring-projects/spring-framework/blob/9fb614a5c60bf6f269edee6b69e543d8ca6af1c3/spring-core/src/main/java/org/springframework/core/io/AbstractFileResolvingResource.java#L46-L86

The `NonexistentClasspathResource` test case provided in this issue reproduces a bug detected in the Spring Framework while running within a native image.

# Overview

If a _relative_ `URL` for a nonexistent resource is created based on an existing classpath resource within a native image, the nonexistent resource **appears to exist** within the native image even though it is not present.

For example, given the `NonexistentClasspathResource` test case below, the output when using a standard JDK results in the following.

```diff
+ SUCCESS: found classpath resource: NonexistentClasspathResource.class
+ getContentLengthLong() for file:/nonexistent_classpath_resource/NonexistentClasspathResource.class: 466
+ SUCCESS: classpath resource exists: file:/nonexistent_classpath_resource/NonexistentClasspathResource.class
+ SUCCESS: bogus classpath resource does not exist: file:/nonexistent_classpath_resource/Bogus.class
```

Whereas, if you compile the test case into a native image, the output is the following.

```diff
- FAILURE: could not find classpath resource: NonexistentClasspathResource.class
```

If you execute the test case with the GraalVM native image agent, it generates the following in `resource-config.json`:

```json
{
  "resources":[
    {"pattern":"\\QMETA-INF/services/jdk.vm.ci.hotspot.HotSpotJVMCIBackendFactory\\E"}, 
    {"pattern":"\\QMETA-INF/services/jdk.vm.ci.services.JVMCIServiceLocator\\E"}, 
    {"pattern":"\\QNonexistentClasspathResource.class\\E"}
  ],
  "bundles":[]
}
```

If you then build the native image with the configuration generated by the agent, the output is the following.

```diff
+ SUCCESS: found classpath resource: NonexistentClasspathResource.class
+ getContentLengthLong() for resource:NonexistentClasspathResource.class: 466
+ SUCCESS: classpath resource exists: resource:NonexistentClasspathResource.class
- getContentLengthLong() for resource:Bogus.class: 466
- FAILURE: bogus classpath resource appears to exist: resource:Bogus.class
```

The above **FAILURE** points out the bug within a native image.

The first resource is now found; however, the second "bogus" resource also appears to exist within the native image. Note that `java.net.URLConnection.getContentLengthLong()` returns a positive value for the nonexistent resource that is identical to the value returned for the existing resource (`466`).

See https://github.com/sbrannen/graalvm-playground/tree/master/nonexistent_classpath_resource for full details.

```java
import java.io.IOException;
import java.io.InputStream;
import java.net.URL;
import java.net.URLConnection;

public class NonexistentClasspathResource {

		new Tester().test();
	}


	static class Tester {

		void test() throws Exception {
			String name = "NonexistentClasspathResource.class";
			URL resource1 = getClass().getResource(name);
			if (resource1 == null) {
				System.err.println("FAILURE: could not find classpath resource: " + name);
				return;
			}
			System.err.println("SUCCESS: found classpath resource: " + name);

			if (!exists(resource1)) {
				System.err.println("FAILURE: classpath resource does not exist: " + resource1);
				return;
			}
			System.err.println("SUCCESS: classpath resource exists: " + resource1);

			URL resource2 = new URL(resource1, "Bogus.class");
			if (exists(resource2)) {
				System.err.println("FAILURE: bogus classpath resource appears to exist: " + resource2);
				return;
			}
			System.err.println("SUCCESS: bogus classpath resource does not exist: " + resource2);
		}

		private boolean exists(URL url) {
			try {
				// Try a URL connection content-length header
				URLConnection con = url.openConnection();
				if (con.getContentLengthLong() > 0) {
					System.err.format("getContentLengthLong() for %s: %d%n", url, con.getContentLengthLong());
					return true;
				}

				// Fall back to stream existence: can we open the stream?
				getInputStream(url).close();
				System.err.println("opened InputStream for " + url);
				return true;
			}
			catch (IOException ex) {
				return false;
			}
		}

		private InputStream getInputStream(URL url) throws IOException {
			return url.openConnection().getInputStream();
		}

	}

}
```

# Environment

- GraalVM version: CE 20.0
- JDK major version: 8
- OS: macOS 10.14.6
- Architecture: i386 / x86_64

jperedadnr:
This PR fixes #2291.

When this approach is used:

```
URL resource1 = getClass().getResource("some.name");
URL resource2 = new URL(resource1, "other.name");
```

the first resource is registered during build time, and the URL is created via `Resources::createURL` (on runtime), where a `new URLStreamHandler()` is created based on its `resourceBytes`.

However, for the second resource, the `URLStreamHandler` from the context (the first resource) is used, according to the `URL` [constructor](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/net/URL.java#L649).

If that is the case, the `resourceBytes` can't be taken from the first resource, and the `resources` map has to be used to retrieve such bytes, if such resource2 exists and was correctly registered during built time.

In the test case added to the issue, the `Bogus.class` resource is not found, so a FNFE is thrown, and the test succeeds.

Another use case of such approach can be found in the JavaFX `FXMLLoader` class, when an FXML file loads another FXML resource via `fx:include`, see [issue](https://github.com/gluonhq/substrate/issues/291).

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
@@ -25,6 +25,7 @@
 package com.oracle.svm.core.jdk;
 
 import java.io.ByteArrayInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
@@ -125,8 +126,26 @@ public final class Resources {
 
     public static URL createURL(String name, byte[] resourceBytes) {
         class Conn extends URLConnection {
+
+            private InputStream in;
+            private long length = -1L;
+
             Conn(URL url) {
                 super(url);
+
+                // remove "resource:" from url to get the resource name
+                String resName = url.toString().substring(1 + JavaNetSubstitutions.RESOURCE_PROTOCOL.length());
+                if (resName.equals(name)) {
+                    in = new ByteArrayInputStream(resourceBytes);
+                    length = resourceBytes.length;
+                } else {
+                    final List<byte[]> bytes = singleton().resources.get(resName);
+                    if (bytes != null && !bytes.isEmpty()) {
+                        byte[] resBytes = bytes.get(0);
+                        in = new ByteArrayInputStream(resBytes);
+                        length = resBytes.length;
+                    }
+                }
             }
 
             @Override
@@ -135,17 +154,20 @@ public final class Resources {
 
             @Override
             public InputStream getInputStream() throws IOException {
-                return new ByteArrayInputStream(resourceBytes);
+                if (in == null) {
+                    throw new FileNotFoundException(url.toString());
+                }
+                return in;
             }
 
             @Override
             public long getContentLengthLong() {
-                return resourceBytes.length;
+                return length;
             }
         }
 
         try {
-            return new URL("resource", null, -1, name, new URLStreamHandler() {
+            return new URL(JavaNetSubstitutions.RESOURCE_PROTOCOL, null, -1, name, new URLStreamHandler() {
                 @Override
                 protected URLConnection openConnection(URL u) throws IOException {
                     return new Conn(u);


johanvos:
Tested this locally, and it fixes the issue indeed.

jovanstevanovic:
It looks like we are dealing with the same issue. The only difference is that my changes are part of bigger resource refactoring. Take a look at changes in this PR #3315. If you have any suggestions let me know.

thomaswue:
@olpaw or @christianwimmer can you please look at this?

jperedadnr:
@jovanstevanovic Building GraalVM from head (so #3315 is included), works (at least related to the resources issues #2291 and https://github.com/gluonhq/substrate/issues/291), but only on Mac and Linux. Windows fails.

I've run these two tests:

- JavaFX/FXML [code](https://github.com/gluonhq/substrate/issues/291#issuecomment-577818121). On Windows it produces a NullPointerException:
```
[INFO] [SUB]  Caused by: java.lang.NullPointerException: Location is required.
[INFO] [SUB]   at javafx.fxml.FXMLLoader.loadImpl(FXMLLoader.java:3310)
[INFO] [SUB]   at javafx.fxml.FXMLLoader.loadImpl(FXMLLoader.java:3274)
```
- Java/ XML [code](https://github.com/gluonhq/substrate/issues/291#issuecomment-735370683), on Windows it fails with:
```
[INFO] [SUB] java.net.MalformedURLException: no protocol: Child.xml
[INFO] [SUB]   at java.net.URL.<init>(URL.java:645)
[INFO] [SUB]   at java.net.URL.<init>(URL.java:541)
[INFO] [SUB]   at hello.HelloWorld.main(HelloWorld.java:42)
```

jperedadnr:
I've also tested some dev builds on Windows:

- up until https://github.com/graalvm/graalvm-ce-dev-builds/releases/tag/21.2.0-dev-20210612_0244  "works", but still has #2291, as expected. A regular FXML [sample](https://github.com/gluonhq/gluon-samples/tree/master/HelloFXML) works fine.

- starting with https://github.com/graalvm/graalvm-ce-dev-builds/releases/tag/21.2.0-dev-20210615_0052 fails with same errors [reported](https://github.com/oracle/graal/pull/3336#issuecomment-864250902) for both App test and FXML app with `fx:include`, but also a new error shows up for the regular FXML [sample](https://github.com/gluonhq/gluon-samples/tree/master/HelloFXML):
```
[INFO] [SUB] java.util.MissingResourceException: Can't find bundle for base name hellofx.hello, locale en_US
[INFO] [SUB]   at java.util.ResourceBundle.throwMissingResourceException(ResourceBundle.java:2055)
```

olpaw:
@jovanstevanovic the findings from @jperedadnr indicate that the new resource implementation does not work as expected on Windows. Please fix this on master and also on the soon-to-created 21.2.0-branch. Resource support has to work properly for all our supported platforms.

olpaw:
@jperedadnr thanks for your testing https://github.com/oracle/graal/pull/3336#issuecomment-864250902 https://github.com/oracle/graal/pull/3336#issuecomment-864393491

Now that @jovanstevanovic s new resource support is on master the changes this PR provides are obsolete and the PR should be closed. Makes sense?

jperedadnr:
Sure, no problem, you can close this PR. 

However, and if it helps: since this PR fixes it for Windows, if it is useful for #3315, feel free to use all or parts of it.

olpaw:
> Sure, no problem, you can close this PR.
> 
> However, and if it helps: since this PR fixes it for Windows, if it is useful for #3315, feel free to use all or parts of it.

Thanks. I will create a new issue and make it refer to this PR for guidance.

gradinac:
Hey @jperedadnr! Thank you for reporting the issue and testing this out! It turned out to be a problem with the way we pass file paths to resource inclusion matchers. I've opened https://github.com/oracle/graal/pull/3498 to address this and I've tested out https://github.com/gluonhq/substrate/issues/291#issuecomment-735370683:
 - When `child.xml` isn't manually added to the config, we get an exception
 - When `child.xml` is manually added to the config, the test seems to work
 
The reason we need to add the resource manually is because the `native-image-agent` can't see this resource as accessed as the access happens through URL manipulation.

I've noticed that there were a couple of other Gluon samples, specifically https://github.com/gluonhq/gluon-samples/tree/master/HelloFXML, however, these seem to require a custom GraalVM build on Windows to test. Could you please try out the fix and see if it works, or is there another reproducer that I could run locally? I've tried running it without a custom GraalVM build but I was hitting linker errors due to missing symbols

jperedadnr:
I could try to apply the patch from #3498 to do a custom build and then run the tests again on Windows. I'll report back.

jperedadnr:
I've tested successfully this patch, the three tests reported that were failing now pass.

gradinac:
Thank you very much for verifying this! :)
I'll merge this PR to `master` and to the 21.2 release branch

PR #3367, opened 2021-04-21
Evaluate impact of including only invoked reflection proxies

loicottet:
This code allows limiting the inclusion of reflection proxies in an image based on a previous agent run which detected which methods are actually invoked through reflection.


loicottet:
These changes were merged as part of 3d7a8a7a281cf70ecfa8c201ead0736b8e145cdd

PR #3378, opened 2021-04-26
Add intrinsics for substring search and SIMD implementation for char search using String.indexOf on AAarch64

SwapnilGaikwad:
This PR contains two patches

i) The first patch intrinsifies three String methods to search a substring in a given target string:
- StringUTF16.indexOf([BI[BII)I  # both search and target string UTF16
- StringUTF16.indexOf([B[B)I     # both search and target string Latin1
- StringLatin1.indexOf([B[B)I    # search string is Latin1 and target string UTF16

This patch adds a boiler-plate structure for intrinsification of substring methods
that the subsequent patches could utilise for performance improvements. The patch
has not shown any noticeable performance regression on any of the benchmarks.

ii) The second patch adds SIMD implementation of String.indexOf char on AArch64
This patch replaces the chunk-based char search algorithm with its SIMD equivalent.
It uses chunks of size 32 bytes instead of 4 bytes. The patch repurposed the char 
search approach of Arm optimised routines [1].

The benefits of using SIMD implementation should increase for longer strings.
For long strings (~750 Latin1 chars), microbenchmarking shows an improvement of
8% (on A72) and 33% (on Neoverse N1).

[1] https://github.com/ARM-software/optimized-routines/blob/master/string/aarch64/strchr.S

### DIFF
--- a/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayIndexOfOp.java
+++ b/compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayIndexOfOp.java
@@ -50,6 +50,7 @@ public final class AArch64ArrayIndexOfOp extends AArch64LIRInstruction {
     public static final LIRInstructionClass<AArch64ArrayIndexOfOp> TYPE = LIRInstructionClass.create(AArch64ArrayIndexOfOp.class);
 
     private final boolean isUTF16;
+    private final boolean findTwoConsecutive;
     private final int arrayBaseOffset;
 
     @Def({REG}) protected Value resultValue;
@@ -62,11 +63,13 @@ public final class AArch64ArrayIndexOfOp extends AArch64LIRInstruction {
     @Temp({REG}) protected Value temp3;
     @Temp({REG}) protected Value temp4;
 
-    public AArch64ArrayIndexOfOp(int arrayBaseOffset, JavaKind valueKind, LIRGeneratorTool tool, Value result, Value arrayPtr, Value arrayLength, Value fromIndex, Value searchValue) {
+    public AArch64ArrayIndexOfOp(int arrayBaseOffset, JavaKind valueKind, boolean findTwoConsecutive, LIRGeneratorTool tool, Value result, Value arrayPtr, Value arrayLength, Value fromIndex,
+                    Value searchValue) {
         super(TYPE);
         assert byteMode(valueKind) || charMode(valueKind);
         this.arrayBaseOffset = arrayBaseOffset;
         this.isUTF16 = charMode(valueKind);
+        this.findTwoConsecutive = findTwoConsecutive;
         resultValue = result;
         arrayPtrValue = arrayPtr;
         arrayLengthValue = arrayLength;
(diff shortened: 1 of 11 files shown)


compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayIndexOfOp.java
```diff
@@ -87,180 +112,315 @@ private static boolean charMode(JavaKind kind) {
...
+            elementSize = ElementSize.HalfWord;
+            magicConstant = 0xc000_0300_00c0_0003L;
         }
-        masm.orr(64, searchChar, searchChar, searchChar, ShiftType.LSL, Byte.SIZE * 2);
-        masm.orr(64, searchChar, searchChar, searchChar, ShiftType.LSL, Byte.SIZE * 4);
 
-        // 2.1 Set end index at the starting position of the last chunk
-        masm.sub(64, endIndex, arrayLength, chunkSize / charSize);
-        masm.add(64, baseAddress, baseAddress, endIndex, ShiftType.LSL, shiftSize);
+        AArch64ASIMDMacroAssembler simdMasm = new AArch64ASIMDMacroAssembler(masm);
```
teshull, line 269: This can be accessed via `masm.neon`

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StringLatin1Substitutions.java
```diff
@@ -78,20 +78,12 @@ public static int indexOf(byte[] value, int ch, int origFromIndex) {
...
-        if (injectBranchProbability(UNLIKELY_PROBABILITY, fromIndex < 0)) {
-            fromIndex = 0;
-        }
-        if (injectBranchProbability(UNLIKELY_PROBABILITY, targetCount == 0)) {
-            // The empty string is in every string.
-            return fromIndex;
-        }
+    public static int indexOf(byte[] source, int sourceCount, byte[] target, int targetCount, int fromIndex) {
+        ReplacementsUtil.dynamicAssert(fromIndex >= 0, "StringLatin1.indexOf invalid args: fromIndex negative");
+        ReplacementsUtil.dynamicAssert(fromIndex < sourceCount, "StringLatin1.indexOf invalid args: fromIndex >= length(target)");
```
teshull, line 83: This assumption isn't valid.

teshull:
Initially running through the gates failed due to the assumption I mentioned above. I edited the code some (https://github.com/oracle/graal/pull/3701) and am running it through the gates again. I'll let you know if I run into more issues

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StringLatin1Substitutions.java
SwapnilGaikwad: You're right. If StringLatin1.indexOf is called from somewhere other than String.indexOf method, i.e., without going through the checks put by the indexOf method, then this assumption would be invalid.

teshull:
merged with 9183dff4665e561fea588f83ece80912e29f6bc2

PR #3382, opened 2021-04-28
[GR-30433] Make native-image driver buildable with --module-path.

olpaw:


olpaw:
Heads up: The remains of the previously to experimental demoted native-image build server mode will be removed soon.
With all the changes for module-support it is impossible to keep the server specific sources in a consistent state. See:

https://github.com/oracle/graal/pull/3382/commits/8130b753c9aa91b3b4f0b8a5d92d2aef96533114

PR #3397, opened 2021-05-10
Implement basic module substitutions

ivan-ristovic:
This PR implements basic methods of the `Module` class. 

Change list:
- Removed substitutions for methods such as `canRead()`, `canUse()` etc. because the JDK implementation can be reused
- Removed `@Delete` annotation from `ReflectionData`, because it is used by JDK
- Redefine the value of the `ReflectionData.uses` field, otherwise it will contain classes that were not nececarily discovered by analysis
- Added an alias for `java.lang.WeakHashMap`, in order to redefine the `ReflectionData.uses` field


ivan-ristovic:
I moved this PR to #3445 for better clarity.

PR #3430, opened 2021-05-24
Allow GraalVM CE to build with OpenJDK 8 and --no-jlinking

zakkak:
This patch allows GraalVM CE to build with OpenJDK 8 and the
--no-jlinking flag by not adding --upgrade-module-path and --add-exports
parameters when OpenJDK 8 is used.


zakkak:
@gilles-duboscq can you please have a look?

gilles-duboscq (approved):
Looks good. Thanks.
I'll pick it for integration.

PR #3431, opened 2021-05-25
CI: Cancel previous jobs for same PR or pendings ones for branches

zakkak:
The changes in this patch as well as the following commit message have
been copied from:
https://github.com/quarkusio/quarkus/commit/c97aa9dded0926a642680c27ecf859bf51f2e1e7

> See
> https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency
> for the documentation of this feature.
> 
> Essentially, this gives the following behavior:
> 
> * For pull requests, only the most recent HEAD gets built. Builds for
>   previous HEADs get cancelled automatically.
> * For branches (main, ...), we can only ever have two concurrent builds:
>   one running, and one (more recent) pending.
>   If a build is running and another one gets requested (because of a
>   push), the second one will be pending and wait for the first one to
>   finish.
>   If a third build gets requested before the first one finishes,
>   the second build gets cancelled and the third build will be pending
>   until the first build finishes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,10 @@ on:
       - '**.md'
   workflow_dispatch: []
 
+concurrency:
+  group: "workflow = ${{ github.workflow }}, ref = ${{ github.event.ref }}, pr = ${{ github.event.pull_request.id }}"
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 env:
   LANG: en_US.UTF-8
   MX_GIT_CACHE: refcache
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,15 @@ on:
       - '**.md'
   workflow_dispatch: []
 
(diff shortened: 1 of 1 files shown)


zakkak:
Hi @dougxc can you please have a look at this?

.github/workflows/main.yml
```diff
@@ -13,6 +13,10 @@ on:
       - '**.md'
   workflow_dispatch: []
 
+concurrency:
```
dougxc, line 16: Can you please add some comments here that paraphrase the text you put in the PR description.

zakkak: Done in https://github.com/oracle/graal/pull/3431/commits/f133a6407f422214ce7c797a74695d21feff9533

dougxc:
Should be resolved by https://github.com/oracle/graal/commit/21d500232f63a0f76a028aa6c5c3629d8067b38e - please re-open if not.

PR #3445, opened 2021-06-02
Implement Module/ModuleLayer substitutions

ivan-ristovic:
This PR implements basic methods of the `Module` class. 

Change list:
- Removed substitutions for methods such as `canRead()`, `canUse()` etc. because the JDK implementation can be reused
- Removed `@Delete` annotation from `ReflectionData`, because it is used from now on
- Redefine the value of the `ReflectionData.uses` field, otherwise it will contain classes that were not nececarily discovered by analysis
- Added an alias for `java.lang.ModuleLayer`, in order to access private fields
- Added an alias for `java.lang.WeakHashMap`, in order to redefine the `ReflectionData.uses` field
- Added a feature which synthesizes the boot module layer
- Replaced `System.bootLayer` module set with the synthesized module set


olpaw:
@ivan-ristovic please provide a before-after comparison based on a simple hello-world module-build sample so that it is clear how the runtime-representation of module info is changed by this PR.

substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/jdk/Target_java_lang_WeakPairMap.java
```diff
@@ -0,0 +1,53 @@
...
+import com.oracle.svm.core.annotate.TargetElement;
+
+import java.lang.ref.ReferenceQueue;
+import java.util.concurrent.ConcurrentHashMap;
+
+@TargetClass(className = "java.lang.WeakPairMap", onlyWith = JDK11OrLater.class)
+final class Target_java_lang_WeakPairMap<K1, K2, V> {
+
+    @Alias //
+    @TargetElement(name = "map")
```
olpaw, line 38: If field-name is already exactly like the one you are trying to alias in the `TargetClass` the `name = "map"` bit should be redundant.

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
```diff
@@ -0,0 +1,67 @@
...
+    public void beforeAnalysis(BeforeAnalysisAccess access) {
+        ImageSingletons.add(BootModuleLayerSupport.class, new BootModuleLayerSupport());
+    }
+
+    @Override
+    public void afterAnalysis(AfterAnalysisAccess access) {
+        FeatureImpl.AfterAnalysisAccessImpl accessImpl = (FeatureImpl.AfterAnalysisAccessImpl) access;
+        AnalysisUniverse universe = accessImpl.getUniverse();
+
+        Set<Object> moduleSet = universe.getTypes()
```
olpaw, line 58: Nice :-)

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
```diff
@@ -0,0 +1,67 @@
...
+                .stream()
+                .filter(t -> t.isReachable() && !t.isArray())
+                .map(t -> t.getJavaClass().getModule())
+                .filter(Module::isNamed)
```
olpaw, line 62: Rather than filtering out the unnamed module we should understand why we have it in the analysis results if we built a pure module-path based image. Pleas investigate what causes it to show up in the module-path based image usecase.

substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/jdk/Target_java_lang_WeakPairMap.java
ivan-ristovic: Yes, ofcourse, I plan to change that part anyway in order to support proper synthesizing of the `ReflectionData.uses` map.

ivan-ristovic:
> @ivan-ristovic please provide a before-after comparison based on a simple hello-world module-build sample so that it is clear how the runtime-representation of module info is changed by this PR.

I added a simple hello-module sample output to the PR description

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
ivan-ristovic: After investigating this a bit more, I discovered that in my case after analysis there are around 135 reachable classes in the unnamed module. All o135 classes seem to be synthesized from class methods contained in 
`com.oracle.core.svm.reflect`. 

I will keep updating this comment as I investigate this further.

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
```diff
@@ -0,0 +1,67 @@
...
+                .collect(Collectors.toSet());
+
+        BootModuleLayerSupport.instance().setReachableModules(moduleSet);
```
ivan-ristovic, line 65: @cstancu, I was advised to consult with you about the current feature implementation due to it's unique nature. For the boot module layer, substitutions depend on analysis results e.g. the boot module set should only consist of modules which contain a reachable class. Current implementation does the replacement of relevant fields after analysis, but all referenced classes should already be reachable. Do you think this might cause any issues (such as discovering a class not seen during analysis)?

substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/jdk/Target_java_lang_WeakPairMap.java
ivan-ristovic: Removed.

olpaw:
Hi @ivan-ristovic

- Please extend `mx hellomodule` to add testing that secures your improvements from future regressions. I.e. extend the existing example and add assertions that verify correct runtime-moduleinfo.
- Please also add a github gate runs `mx hellomodule` so that community PRs can see if they break the current state of module support without requiring us to run them in our internal CI. (You can look at https://github.com/oracle/graal/pull/3478 to get an idea which files you might need to change to add an new job to github CI)

olpaw:
Running your PR on top of https://github.com/oracle/graal/pull/3446 breaks `mx hellomodule`:
```
...
[moduletests.hello.app:74056]     (inline):     976.14 ms,  1.77 GB
[moduletests.hello.app:74056]    (compile):   4,475.44 ms,  3.29 GB
[moduletests.hello.app:74056]      compile:   6,686.27 ms,  3.29 GB
[moduletests.hello.app:74056]        image:   1,714.78 ms,  3.29 GB
[moduletests.hello.app:74056]        write:     166.64 ms,  3.29 GB
[moduletests.hello.app:74056]      [total]:  22,307.03 ms,  3.29 GB
# Printing build artifacts to: /home/pwoegere/OLabs/issues/GR-31840/graal/substratevm/svmbuild/hellomodule/moduletests.hello.app.build_artifacts.txt
Running image /home/pwoegere/OLabs/issues/GR-31840/graal/substratevm/svmbuild/hellomodule/moduletests.hello.app built from module:
Basic Module test involving module moduletests.hello.app and module moduletests.hello.lib
Hello Module World!
Now accessing package that is not exported in module moduletests.hello.lib
Seeing this requires --add-exports
Now accessing private method from not exported package in module moduletests.hello.lib
Exception in thread "main" java.lang.reflect.InaccessibleObjectException: Unable to make private static void hello.privateLib2.PrivateGreeter.greet() accessible: module moduletests.hello.lib does not "opens hello.privateLib2" to module moduletests.hello.app
	at java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:340)
	at java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:280)
	at java.lang.reflect.Method.checkCanSetAccessible(Method.java:198)
	at java.lang.reflect.Method.setAccessible(Method.java:192)
	at hello.Main.main(Main.java:53)
```

olpaw:
> When setting module runtime-state you must honor `--add-exports` and `--add-opens` that have been passed to native-image.
> https://github.com/olpaw/graal/blob/e044094cfb358a8b63399262aff461b08375145c/substratevm/mx.substratevm/mx_substratevm.py#L1095

Update. Forget what I said. Your changes actually fix the hellomodule test now correctly requiring `--add-opens` where previously only ``--add-exports` was required. Also now the commented out assert can be enabled in the test.

olpaw (changes requested):
See comments

substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/jdk/Target_java_lang_Module.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
```
olpaw, line 2: , 2012

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
olpaw: Do we already know what was causing methods from `com.oracle.core.svm.reflect` to have ties to the unnamed module?

ivan-ristovic:
> See comments [#3445 (comment)](https://github.com/oracle/graal/pull/3445#issuecomment-863225585) [#3445 (comment)](https://github.com/oracle/graal/pull/3445#issuecomment-863191898)

Done.

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
ivan-ristovic: Initially I thought that these classes are only related to `VarHandle`s but I was wrong. From what I can understand, they are generated for the runtime method reflection support - so that `Class.getMethod` will work properly. Now I am not sure if these generated classes belong in the unnamed module, I assume that is the case?

olpaw: I suspect these generated classes for runtime method reflection should be in the same module as the classes are they are generated for. This needs to be fixed in the runtime method reflection support.

olpaw:
Hi @ivan-ristovic
https://ol-bitbucket.us.oracle.com/projects/G/repos/graal/pull-requests/9053 is now on master.
Please rebase your PR so that we can get into merging your changes to master as well. Thanks!

ivan-ristovic:
> Hi @ivan-ristovic
> https://ol-bitbucket.us.oracle.com/projects/G/repos/graal/pull-requests/9053 is finally on master.
> Please rebase your PR so that we can get into merging your changes to master as well. Thanks!

Done.

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
ivan-ristovic: But in any case, the unnamed module should not be part of the boot layer, right? I am assuming it would be the best to keep this filter, just to be safe?

.github/workflows/main.yml
```diff
@@ -78,8 +78,9 @@ jobs:
               PRIMARY: "substratevm"
           - env:
               JDK: "labsjdk-ce-11"
-              GATE: "build,debuginfotest"
+              GATE: "build,debuginfotesti,hellomodule"
```
olpaw, line 81: You changed gate tag from `debuginfotest` to `debuginfotesti`. Since no `debuginfotesti` actually exists those debuginfo tests are not run anymore. I suggest adding a new gate instead on piggybacking on `GATE: "build,debuginfotest"` (also because we are modifying the environment with `USE_NATIVE_IMAGE_JAVA_PLATFORM_MODULE_SYSTEM`. Other gate tags might not like that.

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
olpaw: > But in any case, the unnamed module should not be part of the boot layer, right?

This is correct. To verify that it really is I created an example where a java application uses classes from the module path and the classpath at the same time (using an automatic-module as a bridge from modulepath to classpath) and then dumps the contents of `ModuleLayer.boot()`. As expected the involved unnamed module does **not** show up in `ModuleLayer.boot()`.
```
$ java -cp cp -p mp:mp-to-cp.jar -m trivialmodule/trivialpackage.HelloJava
Hello 𝓙𝓪𝓿𝓪 ☕
Hello from HelloJava class in module trivialmodule
Hello from ClassLoadedFromAutomaticModule class in module mp.to.cp
Hello From Class on Classpath unnamed module @7ac7a4e4
Modules in the boot layer:
module jdk.crypto.ec
module java.rmi
module java.management.rmi
module java.logging
module jdk.localedata
module jdk.security.jgss
module jdk.naming.rmi
module java.xml
module java.datatransfer
module java.desktop
module trivialmodule
module java.naming
module jdk.zipfs
module jdk.management.jfr
module jdk.jdeps
module java.xml.crypto
module jdk.charsets
module java.smartcardio
module java.prefs
module jdk.jfr
module java.management
module jdk.javadoc
module jdk.compiler
module java.security.jgss
module jdk.internal.opt
module jdk.unsupported.desktop
module java.compiler
module jdk.jartool
module jdk.crypto.cryptoki
module mp.to.cp
module jdk.security.auth
module jdk.naming.dns
module java.base
module jdk.management
module java.security.sasl
module jdk.jlink
```
I added the stuff i hacked together as attachment: [test-mix-mp-cp.zip](https://github.com/oracle/graal/files/6739944/test-mix-mp-cp.zip)

In other words, yes please keep the filter. It's actually required.

That said, we still have to address the issue of creating generated classes for runtime method reflection to be in the same module as the classes they are generated for.

.github/workflows/main.yml
ivan-ristovic: Corrected, it must have been a typo. I also agree a separate gate is better - I added it as well.

substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/jdk/BootModuleLayerSupport.java
```diff
@@ -0,0 +1,74 @@
...
+                .stream()
+                .map(o -> (Module) o)
+                .collect(Collectors.toSet()));
+        nameToModule.putAll(reachableModules.stream().collect(Collectors.toMap(Module::getName, m -> m)));
+    }
+
+    public Object getBootLayer() {
+        Target_java_lang_ModuleLayer originalLayer = SubstrateUtil.cast(bootLayer, Target_java_lang_ModuleLayer.class);
+        if (isAnalysisComplete) {
+            originalLayer.nameToModule = nameToModule;
```
olpaw, line 69: Having `@InjectAccessors` perform a side effect via `getBootLayer()` like that is a misuse of InjectAccessors.

Instead we have to declare `reachableModules` and `nameToModule` as `@UnknownObjectField`s and add a  `org.graalvm.nativeimage.hosted.Feature#beforeCompilation` handler to patch the data before compilation.

olpaw (changes requested):
See https://github.com/oracle/graal/pull/3445#discussion_r661380695

substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/jdk/BootModuleLayerSupport.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public final class BootModuleLayerSupport {
+
+    public static BootModuleLayerSupport instance() {
+        return ImageSingletons.lookup(BootModuleLayerSupport.class);
+    }
+
+    private final ModuleLayer bootLayer;
```
olpaw, line 44: The Support singleton should only contain fields that contain data that should end up in the image. Thus the `bootLayer` that you are saving for later so that you know in which ModuleLayer instance you need to replace the `nameToModule` and `modules` is better kept in the `ModuleLayerFeature`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/BootModuleLayerSupport.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.core.jdk;
+
+import java.util.Set;
+
+public final class BootModuleLayerSupport {
```
olpaw, line 29: You should not need a Java8 specific BootModuleLayerSupport class. Remove this class and instead add
```patch
diff --git a/substratevm/mx.substratevm/suite.py b/substratevm/mx.substratevm/suite.py
index cbde02f458..5ea0c85d0e 100644
--- a/substratevm/mx.substratevm/suite.py
+++ b/substratevm/mx.substratevm/suite.py
@@ -398,6 +398,7 @@ suite = {
             "sourceDirs": ["src"],
             "dependencies": [
                 "com.oracle.svm.hosted",
+                "com.oracle.svm.core.jdk11",
             ],
             "requires" : ["java.instrument"],
             "requiresConcealed" : {
```

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
olpaw: Since the way bootLayer for image runtime is now injected now uses afterAnalysis @cstancu s help is no more needed.

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
```diff
@@ -0,0 +1,185 @@
...
+
+    @Override
+    public void duringAnalysis(DuringAnalysisAccess access) {
+        FeatureImpl.DuringAnalysisAccessImpl accessImpl = (FeatureImpl.DuringAnalysisAccessImpl) access;
+        AnalysisUniverse universe = accessImpl.getUniverse();
+
+        Map<String, Module> reachableModules = universe.getTypes()
+                .stream()
+                .filter(t -> t.isReachable() && !t.isArray())
+                .map(t -> t.getJavaClass().getModule())
```
olpaw, line 125: Probably a good idea to add
```
.distinct()
```
after
```
.map(t -> t.getJavaClass().getModule())
```
(no need to look at any prev. looked module)

Not sure if it makes a measurable difference though (trading space for avoiding redundant recomputations).

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
```diff
@@ -0,0 +1,185 @@
...
+    public void beforeAnalysis(BeforeAnalysisAccess access) {
+        access.registerReachabilityHandler(
+                a -> a.registerAsUnsafeAccessed(ReflectionUtil.lookupField(ClassLoader.class, "classLoaderValueMap")),
+                ReflectionUtil.lookupMethod(ClassLoader.class, "trySetObjectField", String.class, Object.class)
+        );
+        RuntimeReflection.register(ReflectionUtil.lookupField(ClassLoader.class, "classLoaderValueMap"));
+    }
+
+    @Override
+    public void duringAnalysis(DuringAnalysisAccess access) {
```
olpaw, line 118: IMO, this should be
```diff
@@ -115,11 +115,11 @@
     }
 
     @Override
-    public void duringAnalysis(DuringAnalysisAccess access) {
-        FeatureImpl.DuringAnalysisAccessImpl accessImpl = (FeatureImpl.DuringAnalysisAccessImpl) access;
+    public void afterAnalysis(AfterAnalysisAccess access) {
+        FeatureImpl.AfterAnalysisAccessImpl accessImpl = (FeatureImpl.AfterAnalysisAccessImpl) access;
         AnalysisUniverse universe = accessImpl.getUniverse();
 
         Map<String, Module> reachableModules = universe.getTypes()
                 .stream()
                 .filter(t -> t.isReachable() && !t.isArray())
                 .map(t -> t.getJavaClass().getModule())
```

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
ivan-ristovic: Done. It is also a question if `distinct()` will perform more efficiently compared to `toMap` collector because `distinct()` can work in O(n) with a hash map as a helper data structure. In this case the map is actually the end product as well as a helper so it seemed natural to remove duplicates in that place. It is an interesting question and something to think about if we really wish this part of code to be as efficient as possible, I am not sure what is the best approach.

ivan-ristovic:
@olpaw About the module/package bookkeeping/lookup: at this point, it is nececary to check if the module with a given name has already been defined:
https://github.com/oracle/graal/blob/fda90e65bf3d81b9fc277ee1e310fcf28bf31623/substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/jdk/Target_java_lang_Module.java#L107

I was hoping that also the package lookup could also be more efficient as well (if the loader has already defined a package with a given name).

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
ivan-ristovic: Using `@UnknownObjectField` with `ModuleLayer` as a provided type does indeed allow for this. The only question that remains is what should be manually marked as reachable? I assume only the `ModuleLayer` and `Configuration` types, although if my reasoning is correct, they are always reachable anyway.

olpaw:
The following patch provides a solution to make setting the bootLayer once after analysis:
https://github.com/oracle/graal/commit/41c4c190c55216dd08b6994a6908a35b44c4a896

substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/jdk/Target_java_lang_Module.java
```diff
@@ -0,0 +1,219 @@
...
+                if (!ModuleUtil.isValidPackageName(pn)) {
+                    throw new IllegalArgumentException("Invalid package name: " + pn + " for module: java.base");
+                }
+            }
+
+            if (module.getClassLoader() != null) {
+                throw new IllegalArgumentException("Class loader must be the boot class loader");
+            }
+
+            synchronized (ModuleUtil.moduleLock) {
```
olpaw, line 76: Does `defineModule0` have to provide certain guarantees regarding synchronization that require you to have this?

olpaw (changes requested):
Look good overall. Time to merge the PR after final cleanup.

Please rebase the PR to master and squash the commits where it makes sense.

olpaw:
> I'm running your PR in our CI now to see if anything shows up on EE side.

@ivan-ristovic your PR breaks Java 8 compatibility. If you use JAVA_HOME with Java 8 and EXTRA_JAVA_HOMES with Java 11 you will see errors like:
```
Error: Class initialization of com.oracle.svm.core.hub.DynamicHub failed. Use the option --initialize-at-run-time=com.oracle.svm.core.hub.DynamicHub to explicitly request delayed initialization of this class.
com.oracle.svm.core.util.UserError$UserException: Class initialization of com.oracle.svm.core.hub.DynamicHub failed. Use the option --initialize-at-run-time=com.oracle.svm.core.hub.DynamicHub to explicitly request delayed initialization of this class.
	at com.oracle.svm.core.util.UserError.abort(UserError.java:82)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.reportInitializationError(ConfigurableClassInitialization.java:213)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.ensureClassInitialized(ConfigurableClassInitialization.java:194)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.forceInitializeHosted(ConfigurableClassInitialization.java:497)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.handleClass(AnnotationSubstitutionProcessor.java:297)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.init(AnnotationSubstitutionProcessor.java:266)
	at com.oracle.svm.hosted.NativeImageGenerator.createDeclarativeSubstitutionProcessor(NativeImageGenerator.java:941)
	at com.oracle.svm.hosted.NativeImageGenerator.setupNativeImage(NativeImageGenerator.java:873)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:531)
	at com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:492)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:400)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:566)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:122)
[js:9655]      [total]:   4,740.02 ms,  1.79 GB
Caused by: java.lang.BootstrapMethodError: java.lang.UnsupportedClassVersionError: com/oracle/svm/core/jdk/Target_java_lang_Module has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at com.oracle.svm.core.hub.DynamicHub.<clinit>(DynamicHub.java:348)
	at sun.misc.Unsafe.ensureClassInitialized(Native Method)
	at org.graalvm.compiler.serviceprovider.GraalUnsafeAccess.ensureClassInitialized(GraalUnsafeAccess.java:77)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.ensureClassInitialized(ConfigurableClassInitialization.java:178)
	... 10 more
Caused by: java.lang.UnsupportedClassVersionError: com/oracle/svm/core/jdk/Target_java_lang_Module has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:355)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	... 14 more
```

substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/jdk/Target_java_lang_Module.java
ivan-ristovic: Native `defineModule0` method uses mutexes when accessing module/package tables. If we do not synchronize, we might end up allowing the same package to be defined twice to the same classloader for example.

ivan-ristovic:
> @ivan-ristovic your PR breaks Java 8 compatibility.
> If you use JAVA_HOME with Java 8 and EXTRA_JAVA_HOMES with Java 11 you will see errors like:

Unfortunately, I do not have EE access yet. CE built locally with `mx --java-home <JDK8> --extra-java-homes <JDK11> build` passes without errors. I do not have the exact example as you do to test the launcher, however I can successfully build and run images of a few simple non-modular examples. This looks like overlay problem.

> Maybe a side effect of removing the `trySetObjectField` substitution.

`trySetObjectField` was annotated with `@OnlyWith(JDK11.class)` anyway, it should not be substituted on Java 8. As I understand, it was substituted due to JDK using `Unsafe.objectFieldOffset` which was not implemented on our side, however now it is so it looks like we can reuse the JDK implementation. We need `trySetObjectField` because `ModuleLayer.bindToLoader()` will transitively call this.

> Also `Target_java_lang_Module` and `Target_java_lang_module_ModuleReference` should be in `com.oracle.svm.core.jdk11` so that they are not seen when running on 8.

I agree, however that will cause problems as several other classes use these types (that is why I kept the overlays) for example:
https://github.com/oracle/graal/blob/11477c20416a40ff67ffbc6eab84f6df6b81f8da/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/DynamicHub.java#L348
I am not sure what is the best approach here, maybe you have some ideas?


On a side note, PR is rebased and commits are squashed. I have also added checks for already defined modules/packages.

olpaw:
> I agree, however that will cause problems as several other classes use these types (that is why I kept the overlays) for example:

That just means `singleModuleReference` and module specific substitutions like e.g. `com.oracle.svm.core.jdk.Target_java_lang_ClassLoader#getUnnamedModule()` need to be moved to `com.oracle.svm.core.jdk11` as well.

olpaw:
> Unfortunately, I do not have EE access yet. CE built locally with mx --java-home <JDK8> --extra-java-homes <JDK11> build passes without errors. I do not have the exact example as you do to test the launcher, however I can successfully build and run images of a few simple non-modular examples.

@ivan-ristovic the issue described in https://github.com/oracle/graal/pull/3445#issuecomment-897493823 can also easily reproduced on your side with.

1. export USE_NATIVE_IMAGE_JAVA_PLATFORM_MODULE_SYSTEM=false
2. export JAVA_HOME=/path/to/jdk-8
    export EXTRA_JAVA_HOMES=/path/to/jdk-11
3. export DYNAMIC_IMPORTS=/graal-js
4. // From graal/substratevm ...
    pushd ../..
    git clone https://github.com/oracle/graaljs.git
    popd
5. mx clean && mx build
6. mx native-image --macro:js-launcher -H:+ReportExceptionStackTraces

results in
```
[js:19332]    classlist:   1,664.54 ms,  1.60 GB
[js:19332]        setup:     364.64 ms,  2.23 GB
Error: Class initialization of com.oracle.svm.core.hub.DynamicHub failed. Use the option --initialize-at-run-time=com.oracle.svm.core.hub.DynamicHub to explicitly request delayed initialization of this class.
com.oracle.svm.core.util.UserError$UserException: Class initialization of com.oracle.svm.core.hub.DynamicHub failed. Use the option --initialize-at-run-time=com.oracle.svm.core.hub.DynamicHub to explicitly request delayed initialization of this class.
	at com.oracle.svm.core.util.UserError.abort(UserError.java:82)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.reportInitializationError(ConfigurableClassInitialization.java:213)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.ensureClassInitialized(ConfigurableClassInitialization.java:194)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.forceInitializeHosted(ConfigurableClassInitialization.java:497)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.handleClass(AnnotationSubstitutionProcessor.java:297)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.init(AnnotationSubstitutionProcessor.java:266)
	at com.oracle.svm.hosted.NativeImageGenerator.createDeclarativeSubstitutionProcessor(NativeImageGenerator.java:941)
	at com.oracle.svm.hosted.NativeImageGenerator.setupNativeImage(NativeImageGenerator.java:873)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:531)
	at com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:492)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:400)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:566)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:122)
Caused by: java.lang.BootstrapMethodError: java.lang.UnsupportedClassVersionError: com/oracle/svm/core/jdk/Target_java_lang_Module has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at com.oracle.svm.core.hub.DynamicHub.<clinit>(DynamicHub.java:348)
	at sun.misc.Unsafe.ensureClassInitialized(Native Method)
	at org.graalvm.compiler.serviceprovider.GraalUnsafeAccess.ensureClassInitialized(GraalUnsafeAccess.java:77)
	at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.ensureClassInitialized(ConfigurableClassInitialization.java:178)
	... 10 more
Caused by: java.lang.UnsupportedClassVersionError: com/oracle/svm/core/jdk/Target_java_lang_Module has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	... 14 more
[js:19332]      [total]:   2,089.44 ms,  2.23 GB
```

ivan-ristovic:
Locally this now passes, hope there are no issues on EE as well.

olpaw:
@Ivan, rerunning in out CI now ...

olpaw (approved):

olpaw:
@ivan-ristovic CI results look better now bit I see failing JDK16 gates. E.g.
```
[libpolyglot:26726]    classlist:   7,958.01 ms,  1.20 GB
[libpolyglot:26726]        setup:     977.05 ms,  1.20 GB
Error: Could not find target method: private static void com.oracle.svm.core.jdk11.Target_java_lang_Module_JDK11OrLater.defineModule0(java.lang.Module,boolean,java.lang.String,java.lang.String,java.lang.String[])
com.oracle.svm.core.util.UserError$UserException: Could not find target method: private static void com.oracle.svm.core.jdk11.Target_java_lang_Module_JDK11OrLater.defineModule0(java.lang.Module,boolean,java.lang.String,java.lang.String,java.lang.String[])
	at com.oracle.svm.core.util.UserError.abort(UserError.java:68)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.findOriginalMethod(AnnotationSubstitutionProcessor.java:736)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.handleMethodInAliasClass(AnnotationSubstitutionProcessor.java:368)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.handleAliasClass(AnnotationSubstitutionProcessor.java:340)
[libpolyglot:26726]      [total]:   9,025.46 ms,  1.20 GB
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.handleClass(AnnotationSubstitutionProcessor.java:310)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.init(AnnotationSubstitutionProcessor.java:266)
	at com.oracle.svm.hosted.NativeImageGenerator.createDeclarativeSubstitutionProcessor(NativeImageGenerator.java:944)
	at com.oracle.svm.hosted.NativeImageGenerator.setupNativeImage(NativeImageGenerator.java:873)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:531)
	at com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:492)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:400)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:566)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:122)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner$JDK9Plus.main(NativeImageGeneratorRunner.java:596)
# Printing build artifacts to: /b/b/e/graal/sdk/mxbuild/linux-amd64/libpolyglot.so.image/libpolyglot.build_artifacts.txt
```
You should be able to reproduce that easily by setting JAVA_HOME  to a JDK-16 and trying to build e.g. libpolyglot (use `mx native-image --macro:polyglot-library` in substratevm).

olpaw (approved):

olpaw:
@ivan-ristovic, to get rid of the `Fatal error:com.oracle.svm.util.ReflectionUtil$ReflectionUtilError: java.lang.NoSuchFieldException: classLoaderValueMap` errors I suggest adding
```patch
===================================================================
diff --git a/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java b/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
--- a/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java	(revision Staged)
+++ b/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java	(date 1629218976138)
@@ -95,7 +95,7 @@
 
     @Override
     public boolean isInConfiguration(IsInConfigurationAccess access) {
-        return new JDK11OrLater().getAsBoolean();
+        return new JDK11OrLater().getAsBoolean() && ModuleLayer.boot().findModule("org.graalvm.nativeimage.builder").isPresent();
     }
 
     @Override
```
so that the code in `ModuleLayerFeature#beforeAnalysis` that is problematic for JDK16 will never get run if we are not running the image builder on the module-path.

ivan-ristovic:
Excellent idea, added!

olpaw:
After @ivan-ristovic found another case of a missing `registerAsInHeap` we are switching to a different approach:

- Start analysis with prototype bootlayer then switch to real runtime-bootlayer after analysis.

olpaw:
... running in internal CI now. @ivan-ristovic @gradinac

Issue #3404, opened 2021-05-12
JDK bundles are not discovered and added to native image with JDK16


ilopmar:
**Describe the issue**

Bundles included in JDK 16 are not discovered and can't be included in the native image.

I have an application that declares the following in `resource-config.json`:

```json
{
  "resources": {
    "includes": []
  },
  "bundles": [
    {"name": "com.sun.org.apache.xerces.internal.impl.xpath.regex.message"},
    {"name": "com.sun.org.apache.xerces.internal.impl.msg.DatatypeMessages"},
    {"name": "com.sun.org.apache.xerces.internal.impl.msg.DOMMessages"},
    {"name": "com.sun.org.apache.xerces.internal.impl.msg.JAXPValidationMessages"},
    {"name": "com.sun.org.apache.xerces.internal.impl.msg.SAXMessages"},
    {"name": "com.sun.org.apache.xerces.internal.impl.msg.XIncludeMessages"},
    {"name": "com.sun.org.apache.xerces.internal.impl.msg.XMLMessages"},
    {"name": "com.sun.org.apache.xerces.internal.impl.msg.XMLSchemaMessages"},
    {"name": "com.sun.org.apache.xerces.internal.impl.msg.XMLSerializerMessages"},
    {"name": "com.sun.org.apache.xerces.internal.impl.msg.XPointerMessages"}
  ]
}
```

When building a native image with GraalVM JDK11 every works and all those bundles are included in the native image.

If I try to build the native image with GraalVM JDK16 (compiled from master branch) there are warnings about those bundles not being found, and thus, not being included in the native image.

The bundles are present in both JDK11 & 16:

JDK 11

![image](https://user-images.githubusercontent.com/559192/117996733-b15fef80-b342-11eb-8f1b-63e9d4ae890b.png)

JDK 16

![image](https://user-images.githubusercontent.com/559192/117996828-c5a3ec80-b342-11eb-8d6d-7b05a3f1b2c8.png)

**Steps to reproduce the issue**

I've created a very simple hello world app with the previous `resource-config.json` file to illustrate the problem. 

- `git clone https://github.com/ilopmar/graalvm-issue-jdk16-bundles`
- `cd graalvm-issue-jdk16-bundles`
- Use GraaVM JDK16 compiled from master branch
- `./build-native-image.sh`

During native image generation we can see the following warnings:

```
[my-app:169076]    classlist:     533.07 ms,  0.96 GB
[...]

PR #3446, opened 2021-06-02
[GR-31840] Allow native-image shared libs to be built on module-path.

olpaw:


olpaw:
@gilles-duboscq please have a look at the changes in `sdk/mx.sdk/mx_sdk_vm_impl.py` and `sdk/mx.sdk/mx_sdk_vm.py` if that makes sense to you.

sdk/mx.sdk/mx_sdk_vm.py
```diff
@@ -106,6 +106,7 @@ def __init__(self, destination, jar_distributions, build_args, use_modules=None,
         self.jar_distributions = jar_distributions
         self.build_args = build_args
         self.use_modules = use_modules
+        self.main_module = main_module
```
gilles-duboscq, line 109: Can a library have a "main module"? or should this rather be an attribute of `LauncherConfig`?

sdk/mx.sdk/mx_sdk_vm_impl.py
```diff
@@ -1113,8 +1114,8 @@ def __init__(self, component, image_config, **kw_args):
         :type component: mx_sdk.GraalVmComponent | None
         :type image_config: mx_sdk.AbstractNativeImageConfig
         """
-        deps = []
         self.image_config = image_config
+        deps = list(image_config.jar_distributions) if _get_svm_support().is_supported() else []
```
gilles-duboscq, line 1118: Why is this about SVM support? shouldn't it rather be about 8 vs 11+?

sdk/mx.sdk/mx_sdk_vm_impl.py
```diff
@@ -704,6 +704,7 @@ def _find_escaping_links(root_dir):
                     _add(layout, _component_base, 'dependency:{}/polyglot.config'.format(launcher_project), _component)
             for _library_config in sorted(_get_library_configs(_component), key=lambda c: c.destination):
                 graalvm_dists.update(_library_config.jar_distributions)
+                self.jimage_ignore_jars.update(_library_config.jar_distributions)
```
gilles-duboscq, line 707: I'm not sure i understand `jimage_ignore_jars` that seems a bit dangerous: what distribution is both on the jvmci/boot classpath and part of the jars of a launcher/library?

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java
```diff
@@ -0,0 +1,174 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.svm.hosted;
```
christianwimmer, line 26: Why is this not in a `.jdk11` package?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.oracle.svm.core.ClassLoaderSupport;
+import com.oracle.svm.core.annotate.AutomaticFeature;
+import com.oracle.svm.core.util.VMError;
+
+@AutomaticFeature
+public class ClassLoaderSupportFeature implements Feature {
+    @Override
+    public void afterRegistration(AfterRegistrationAccess a) {
+        VMError.guarantee(JavaVersionUtil.JAVA_SPEC <= 8);
```
christianwimmer, line 45: So looks like you are still relying on the multi-version-jar mechanism to distinguish JDK versions. I thought we agreed to not use that mechanism, ans instead do explicit version checks  in the feature before registration.

christianwimmer (changes requested):

sdk/mx.sdk/mx_sdk_vm.py
olpaw: While currently not implemented we could use that to only search for `@CEntryPoint`s in the specified module (if specified)

sdk/mx.sdk/mx_sdk_vm_impl.py
olpaw: I took inspiration from `mx_sdk_vm_impl.GraalVmNativeImage.__init__`
```python
svm_support = _get_svm_support()
if svm_support.is_supported():
    deps += self.native_image_jar_distributions
```
Previously I just had
```
deps = self.native_image_jar_distributions
```
here, but seeing the code from `mx_sdk_vm_impl.GraalVmNativeImage.__init__` made me suspect that `svm_support.is_supported()`should also be used here.

olpaw: The reason why I need to specify the dependencies is that I need to run the `NativePropertiesBuildTask` after the jars are made into proper modules so that using
```python
build_args += image_config.get_add_exports(set(export_deps_to_exclude))
```
in `mx_sdk_vm_impl.NativePropertiesBuildTask.contents` works reliably.

sdk/mx.sdk/mx_sdk_vm.py
olpaw: But you are right that it is better for now not to have it on `AbstractNativeImageConfig` level.

sdk/mx.sdk/mx_sdk_vm_impl.py
olpaw: This is needed so that we do not end up with a jlinked jdk that contains dummy module definitions of `org.graalvm.nativeimage.agent.jvmtibase`, `org.graalvm.nativeimage.agent.diagnostics`, `org.graalvm.nativeimage.agent.tracing`. Having them would break
```java
if (getBuilderModuleNames().contains(moduleName)) {
    /* Modules provided by the image-builder are not allowed on the -imagemp */
    String modulePathEntryStr = "module " + moduleName + " from " + modulePathEntry;
    showWarning("Ignoring " + modulePathEntryStr + " (implicitly provided by builder)");
    return;
}
``` in `com.oracle.svm.driver.NativeImage#addImageModulePath`

Comment (olpaw) [reply sdk/mx.sdk/mx_sdk_vm_impl.py:707]: > what distribution is both on the jvmci/boot classpath and part of the jars of a launcher/library?

Right, alternatively I could fix why we need this is the fist place.

Which is that currently the `native-image.properties` generated by `mx_sdk_vm_impl.NativePropertiesBuildTask.contents` have a `ImageClasspath/ImageModulePath` that contains all the transitive jars from the image builder itself. In case of using the module system this is not allowed and the jars from the image builder itself need to be filtered out (hence the filtering in `com.oracle.svm.driver.NativeImage#addImageModulePath`).

File: substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java (hunk 5 above)
Comment (olpaw) [reply substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java:45]: If I understand correctly the idea is to remove the jdk11 version of `com.oracle.svm.hosted.ClassLoaderSupportFeature` and to let `src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java` instead register 
`ImageSingletons.add(ClassLoaderSupport.class, new ClassLoaderSupportImpl11....)` for `JavaVersionUtil.JAVA_SPEC > 8` and `ImageSingletons.add(ClassLoaderSupport.class, new ClassLoaderSupportImpl8....)` otherwhise, right?

This should work if I add a dummy `ClassLoaderSupportImpl11` for 8 so that `src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java` compiles.

File: substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java (hunk 4 above)
Comment (olpaw) [reply substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java:26]: I will remove the jdk11 variant of `ClassLoaderSupportFeature.java`. See https://github.com/oracle/graal/pull/3446#discussion_r655974985

File: sdk/mx.sdk/mx_sdk_vm_impl.py (hunk 2 above)
Comment (gilles-duboscq) [reply sdk/mx.sdk/mx_sdk_vm_impl.py:1118]: I think it would be worth explaining why that dependency is added in a comment.

Regarding the check I think `GraalVmNativeImage` checks `_get_svm_support().is_supported()` because if that's true then the native image will be built which requires the input jars to exist. That's because those projects can either build native images or just shell launchers.
On the other end, `GraalVmNativeProperties` only exists if SVM is supported anyway so i don't think this check adds much. However a version check would remove the dependency when it's not necessary (on 8)

File: sdk/mx.sdk/mx_sdk_vm_impl.py (hunk 3 above)
Comment (gilles-duboscq) [reply sdk/mx.sdk/mx_sdk_vm_impl.py:707]: OK, thanks for the explanation. I'll let you decide what's the best approach there.

File: sdk/mx.sdk/mx_sdk_vm_impl.py (hunk 2 above)
Comment (olpaw) [reply sdk/mx.sdk/mx_sdk_vm_impl.py:1118]: > However a version check would remove the dependency when it's not necessary (on 8)

Makes sense. I will add that.

**Comment by olpaw** [comment]
Now that https://github.com/oracle/graal/commit/a461aa012519b8794f790368650fc3b842210dbd is on master we need https://github.com/oracle/graal/pull/3446/commits/45801420ab5ffac94f77b9180e1f82f4990726ce cc @gradinac

File: sdk/mx.sdk/mx_sdk_vm.py (hunk 1 above)
Comment (olpaw) [reply sdk/mx.sdk/mx_sdk_vm.py]: Fixed in https://github.com/oracle/graal/pull/3446/commits/91941f9adf5f9480491a2c07c7a1304008f4788f

File: substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java (hunk 5 above)
Comment (olpaw) [reply substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java:45]: Use same scheme as `RecordSupport`

Comment (olpaw) [reply substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ClassLoaderSupportFeature.java]: Fixed: https://github.com/oracle/graal/pull/3446/commits/89bb920793c0fdd774e467bb8729dc9f3c496bad

**Comment by olpaw** [comment]
https://github.com/oracle/graal/pull/3446/commits/7302e457bd786bbcddec2437aa3c742de393b16f  ... how silly of me, forgetting that not everything is in a module ... yet.

File: sdk/mx.sdk/mx_sdk_vm_impl.py [hunk 6]
```diff
@@ -1113,8 +1114,11 @@ def __init__(self, component, image_config, **kw_args):
         :type component: mx_sdk.GraalVmComponent | None
         :type image_config: mx_sdk.AbstractNativeImageConfig
         """
-        deps = []
         self.image_config = image_config
+        # With Java > 8 there are cases were image_config.get_add_exports is getting called in
```
d-kozak, line 1118: typo: there are cases 'where'

sdk/mx.sdk/mx_sdk_vm_impl.py
```diff
@@ -1898,7 +1906,7 @@ def _get_launcher_args():
             return ''
 
         def _get_add_exports():
-            res = self.subject.native_image_config.get_add_exports(_known_missing_jars)
+            res = ' '.join(self.subject.native_image_config.get_add_exports(_known_missing_jars))
```
d-kozak, line 1909: off-topic:
I am a little confused by the concept of known_missing_jars. Based on the comment above their definition it seems they are optional runtime dependencies. Shouldn't the variable then be called something like optional_runtime_depedencies, optional_runtime_deps or optional_svm_runtime_deps ? When I read known_missing_jars, it's not clear to me what the purpose is and it evokes the thought that there is probably some sort of implementation defect involved and this variable is a patch over it.

olpaw: The whole "concept" of `_known_missing_jars` seems very ad-hoc. I'm just using it here as it makes sense to get rid of `--add-exports` that would be hurtful if the respective jars are in fact missing. If there would be an optional jar that would actually need `--add-export`s we would be in trouble.

> it's not clear to me what the purpose is and it evokes the thought that there is probably some sort of implementation defect involved and this variable is a patch over it

I agree. but cleaning this up (introducing proper handling of optional jars) is beyond the scope of this PR. It would be an entire PR on its own ... seems someone is volunteering to address this :wink:

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/jdk11/ClassLoaderSupportImplJDK11OrLater.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    @Override
+    protected boolean isNativeImageClassLoaderImpl(ClassLoader loader) {
+        return loader == classLoaderSupport.getClassLoader() || loader instanceof NativeImageSystemClassLoader;
+    }
+
+    @Override
+    public List<ResourceBundle> getResourceBundle(String bundleSpec, Locale locale) {
+        String[] specParts = bundleSpec.split(":", 2);
```
d-kozak, line 72: Clever idea to allow the user to specify the module from which the bundle should be loaded. However, I would put the parsing into the caller and have the module as an optional argument. For the command-line options, it makes sense to have this parsing, but for the config file, a new key-value pair separate from the bundle name might be more expressive. And the classloader support shouldn't be aware of the nuances of locale configuration. However, this is more of a `todo` for me to implement it once this gets merged.

olpaw: > Clever idea to allow the user to specify the module from which the bundle should be loaded.

Thanks. The idea is actually from @ilopmar https://github.com/oracle/graal/issues/3404#issuecomment-839888673
This is what he naturally tried to use (expecting it to work). I just identified it as being worth to be implemented.

sdk/mx.sdk/mx_sdk_vm_impl.py
olpaw: Fixed https://github.com/oracle/graal/pull/3446/commits/f4899bc4fd706cbedab41a06f96039362fcc7ab7

sdk/mx.sdk/mx_sdk_vm_impl.py
olpaw: Fixed in fdae21edcc566b12bd00e989302c56d86217c478

substratevm/src/com.oracle.objectfile.jdk11/src/com/oracle/objectfile/ModuleAccess.java
```diff
@@ -31,11 +31,18 @@
     public static void openModuleByClass(Class<?> declaringClass, Class<?> accessingClass) {
```
d-kozak, line 31: Seems that there is exactly the same method defined in `ModuleSupport`. Could it be unified?

olpaw: Nope. Because the objectfile distribution does not depend on the svm distribution and the objectfile distribution also does not depend on the graal distribubtion. Thus there is no place where we could put this stuff that allow reuse from both sides.

d-kozak: Okay, I see.

d-kozak (approved):

substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/jdk11/ClassLoaderSupportImplJDK11OrLater.java
```diff
@@ -0,0 +1,181 @@
...
+        if (specParts.length > 1) {
+            moduleName = specParts[0];
+            bundleName = specParts[1];
+        } else {
+            moduleName = null;
+            bundleName = specParts[0];
+        }
+        String packageName = packageName(bundleName);
+        if (packageName == null) {
+            throw new MissingResourceException("ResourceBundle does not seem to be a fully qualified class name.", bundleName, locale.toLanguageTag());
```
zakkak, line 84: Hi @olpaw ,

With this check in place jar files including resource bundles in .properties file format at their root are not supported. Such an example is [org.eclipse.yasson](https://github.com/eclipse-ee4j/yasson/tree/master/src/main/resources).

Going through [ResourceBundle's Javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ResourceBundle.html) I was not able to find any mention/guidance on avoiding adding ResourceBundles to the root of jar files, so I am wondering if this is something we should relax on GraalVM.

Thanks

olpaw: Yes we should probably add fallback code for that to support that. Please create an issue for me.

zakkak: Hmm I forgot to mention the issue here.
FWIW the created issue was https://github.com/oracle/graal/issues/3535 and it got fixed by https://github.com/oracle/graal/pull/3540

PR #3458, opened 2021-06-10
Override toString() in HostException

fniephaus:
Throwable#toString() displays the class name and optionally the message. While the original message is forwarded, the class name is not. This makes toString() less useful, especially when the message is null. NullPointerExceptions, for example, do not have a message. toString() for an HostException wrapping a NPE returns "com.oracle.truffle.host.HostException".
This commit ensures that the original class name is included in HostException#toString(). For an NPE, that would be "com.oracle.truffle.host.HostException: java.lang.NullPointerException".

Please assign to @chumer

/cc @janehmueller


oubidar-Abderrahim:
Hi @fniephaus 

Thank you for taking the time to provide a fix for this, we will take a look at it shortly.
Since the changes in the PR are small, we probably will close this one and integrate it with a bigger PR to avoid running the gates multiple times.

Thank you again for contributing to GraalVM.

Issue #3423, opened 2021-05-20
GraalVM gate CI workflow failing due to a pylint check in mx_sulong_suite_constituents.py


zakkak:
**Describe the issue**
GraalVM gate CI workflow is failing with:

```
************* Module mx_sulong_suite_constituents
I: 49, 0: Locally disabling E0602 (locally-disabled)
I:191, 0: Locally disabling R0901 (locally-disabled)
E:133,15: Access to member 'results' before its definition line 134 (access-member-before-definition)

Traceback (most recent call last):
  File "/home/runner/work/graal/graal/mx/mx_gate.py", line 422, in gate
gate: 20 May 2021 01:16:37(+00:16) END:   Pylint [0:00:16.217429] [disk (free/total): 18.7GB/83.2GB]
    _run_gate(cleanArgs, args, tasks)
  File "/home/runner/work/graal/graal/mx/mx_gate.py", line 569, in _run_gate
    if mx.command_function('pylint')(['--primary']) != 0:
  File "/home/runner/work/graal/graal/mx/mx_commands.py", line 147, in __call__
    return self.command_function(*args, **kwargs)
  File "/home/runner/work/graal/graal/mx/mx.py", line 14469, in pylint
    run([pylint_exe, '--reports=n', '--rcfile=' + rcfile, pyfile] + additional_options, env=env)
  File "/home/runner/work/graal/graal/mx/mx.py", line 13035, in run
gate: 20 May 2021 01:16:37(+00:16) ABORT: Gate [0:00:16.954330] [disk (free/total): 18.7GB/83.2GB]
    abort(retcode)
  File "/home/runner/work/graal/graal/mx/mx.py", line 4085, in abort
    raise SystemExit(error_code)
SystemExit: 2

The sequence of mx commands that were executed until the failure follows:

mx --primary-suite-path sulong --J @-Xmx2g --java-home=/home/runner/work/graal/graal/jdk version --oneline
mx --primary-suite-path sulong --J @-Xmx2g --java-home=/home/runner/work/graal/graal/jdk sversions
mx --primary-suite-path sulong --J @-Xmx2g --java-home=/home/runner/work/graal/graal/jdk verifymultireleaseprojects
mx --primary-suite-path sulong --J @-Xmx2g --java-home=/home/runner/work/graal/graal/jdk pylint --primary

If the previous sequence is incomplete or some commands were executed programmatically use:

[...]

Issue #3433, opened 2021-05-26
GraalVM gate CI workflow failing to execute `mx --primary-suite-path substratevm native-image --help`


zakkak:
**Describe the issue**
It looks like https://github.com/oracle/graal/commit/6eee91e4e2229e53ced239617e504c519723ab8d broke the "GraalVM gate" workflow which is now failing with:

```
Can not find /home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_CE0C781D19_JAVA11/graalvm-ce0c781d19-java11-21.2.0-dev/bin/native-image
Did you forget to build? Try `mx build`
```
 
**Steps to reproduce the issue**

To reproduce locally:
```
cd /tmp
git clone git@github.com:graalvm/mx.git --depth 1
git clone git@github.com:oracle/graal.git --depth 1
cd graal
../mx/mx --primary-suite-path substratevm native-image --help
```

**Describe GraalVM and your environment:**
 - GraalVM version: **2b3fc59e72e18b9d1bc01ce6abbab2a51d0db8ac**
 - JDK major version: **11**
 - OS: **Fedora 34**
 - Architecture: **AMD64**

olpaw:
Hi @zakkak do you have a link to recent gate failing because of the changes in https://github.com/oracle/graal/commit/6eee91e4e2229e53ced239617e504c519723ab8d ?

olpaw:
Also note that in https://github.com/oracle/graal/blob/master/substratevm/SubstrateVM.md#building-images we explicitly state that 
`mx build` needs to be run before you can use `mx native-image`. So if `mx native-image` accidentally worked without is was just a coincidence you should not rely on.

rschatz:
I'm currently looking into the other failure and I just stumbled over it: https://github.com/oracle/graal/runs/2830086066

Not sure what this check is actually trying to do. If `WITHOUT_VCS` is set, instead of the gate command, it's just running `native-image --help`, without building anything? That looks fishy.

@zakkak do you remember why you introduced that? Shouldn't we still just run the normal `mx gate` command after the `rm -rf .git`? https://github.com/oracle/graal/commit/fc6ef771b442f1271da3037f40e03bd75ee3bb43

rschatz:
I have added an attempted fix to my PR: https://github.com/oracle/graal/pull/3478

zakkak:
> So if mx native-image accidentally worked without, it was just a coincidence you should not rely on.

[...]

PR #3478, opened 2021-06-15
Fix github actions failures.

rschatz:
Trying to fix https://github.com/oracle/graal/issues/3423

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -124,8 +124,9 @@ jobs:
       if: ${{ contains(matrix.env.GATE, 'style') }}
       run: |
         sudo apt install python-pip python-setuptools 
-        sudo pip install astroid==1.1.0
-        sudo pip install pylint==1.1.0
+        cat common.json |
+          jq -r '.deps.common.packages | to_entries[] | select(.key | startswith("pip:")) | (.key | split(":")[1]) + .value' |
+          xargs sudo pip install
     - name: Build GraalVM and run gate
       env: ${{ matrix.env }}
       run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,7 +94,9 @@ jobs:
           - env:
               JDK: "labsjdk-ce-11"
               GATE: "build"
-              PRIMARY: "substratevm"
+              PRIMARY: "vm"
+              DYNAMIC_IMPORTS: "/tools,/substratevm,/sulong"
+              NATIVE_IMAGES: "polyglot"
               WITHOUT_VCS: true
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -93,6 +93,8 @@ jobs:
               PRIMARY: "compiler"
           - env:
               JDK: "labsjdk-ce-11"
+              GATE: "build"
+              PRIMARY: "substratevm"
               WITHOUT_VCS: true
     steps:
     - uses: actions/checkout@v2
@@ -156,7 +158,6 @@ jobs:
         if [[ ${WITHOUT_VCS} ]]
         then
           rm -rf .git
-          ${MX_PATH}/mx --primary-suite-path substratevm native-image --help
-        else
-          ${MX_PATH}/mx --primary-suite-path ${PRIMARY} --J @"-Xmx2g" --java-home=${JAVA_HOME} gate --strict-mode --tags ${GATE}
         fi
+
+        ${MX_PATH}/mx --primary-suite-path ${PRIMARY} --J @"-Xmx2g" --java-home=${JAVA_HOME} gate --strict-mode --tags ${GATE}
--- a/common.json
+++ b/common.json
@@ -32,7 +32,7 @@
         "MX_PYTHON": "python3"
       },
       "packages": {
-        "pip:logilab-common ": "==1.4.4",
+        "pip:logilab-common": "==1.4.4",
         "pip:pylint": "==1.9.3",
         "pip:ninja_syntax": "==1.7.2"
       }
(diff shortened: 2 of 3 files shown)


rschatz:
Also added an attempted fix for https://github.com/oracle/graal/issues/3433

Basically, we're just doing `rm -rf .git` and then run `mx gate --tags build` (which is roughly equivalent to `mx clean && mx build`) in the `substratevm` directory.

@zakkak can you verify this is what the `WITHOUT_VCS=true` gate is supposed to test? If that's ok, I'll also add a test to our internal gate that does the same thing, to prevent future breakage (or do we have that already? cc @olpaw )

.github/workflows/main.yml
```diff
@@ -155,7 +158,6 @@ jobs:
         if [[ ${WITHOUT_VCS} ]]
         then
           rm -rf .git
-          ${MX_PATH}/mx --primary-suite-path substratevm native-image --help
-        else
-          ${MX_PATH}/mx --primary-suite-path ${PRIMARY} --J @"-Xmx2g" --java-home=${JAVA_HOME} gate --strict-mode --tags ${GATE}
         fi
+
+        ${MX_PATH}/mx --primary-suite-path ${PRIMARY} --J @"-Xmx2g" --java-home=${JAVA_HOME} gate --strict-mode --tags ${GATE}
```
olpaw, line 163: This will build a GraalVM release that contains the `/substratevm` suite.
After running this
`$(mx --primary-suite-path ${PRIMARY} graalvm-home)` contains a working `bin/native-image` bash-launcher.
i.e. this looks good.

olpaw (approved):

zakkak:
> Also added an attempted fix for #3433
> 
> Basically, we're just doing `rm -rf .git` and then run `mx gate --tags build` (which is roughly equivalent to `mx clean && mx build`) in the `substratevm` directory.
> 
> @zakkak can you verify this is what the `WITHOUT_VCS=true` gate is supposed to test? If that's ok, I'll also add a test to our internal gate that does the same thing, to prevent future breakage (or do we have that already? cc @olpaw )

Unfortunately, as stated in https://github.com/oracle/graal/pull/2738#discussion_r468109433 `mx gate --tags build` does not suffice.

> Looking at the changes in 05d53b0 one sees that only 3 suites are affected (sdk, compiler, substratevm).
> Now in these suites only a subset of the commands are affected. For instance in substratevm the only affected commmands are native-image and native-image-configure (the ones invoking `_ensure_vm_built`. As a result,
> `mx --primary-suite-path substratevm gate --tags build` works fine even without this PR.
> 
> In order to test all the changes of this PR one needs to run:
> 
>     mx --primary-suite-path substratevm native-image -version
>     mx --primary-suite-path compiler makegraaljdk /tmp/graaljdk
>     mx --primary-suite-path sdk javadoc
> 
> but this still allows people to perform changes that could break mx commands performed outside a vcs repository.
> E.g. `mx --primary-suite-path substratevm build`

.github/workflows/main.yml
zakkak: Unfortunately this is not enough to test if things work outside a VCS repo, please see https://github.com/oracle/graal/pull/3478#issuecomment-861811175 for more details.

olpaw: Since https://github.com/oracle/graal/commit/6eee91e4e2229e53ced239617e504c519723ab8d got merged almost all substrate-related mx commands (like e.g. `mx native-image`) are just calling artifacts that got built as part of running `mx build` in the substratevm directory. In other words if `mx build` (or `mx gate --tags build`) succeeded you can safely assume that `mx native-image --help` also works since its just calling the native-image bash launcher that is located at `$(mx --primary-suite-path ${PRIMARY} graalvm-home)/bin/native-image` after `mx build` made sure it gets created there.

olpaw:
> Unfortunately, as stated in [#2738 (comment)](https://github.com/oracle/graal/pull/2738#discussion_r468109433) `mx gate --tags build` does not suffice.

This is not true anymore. `_ensure_vm_built` got removed as part committing https://github.com/oracle/graal/commit/6eee91e4e2229e53ced239617e504c519723ab8d

.github/workflows/main.yml
rschatz: I guess we could also do this:
```
PRIMARY=vm
DYNAMIC_IMPORTS=/substratevm,/sulong,/tools
NATIVE_IMAGES=polyglot # to save some time
GATE=build
```

It's not literally building everything, but it would exercise almost every different category of stuff there is to build (e.g. Java projects, native projects, a native-image launcher, ...).

zakkak: I see, I admit I didn't check the latest state of `mx_substratevm.py`, my mistake. Thanks for making this clear Paul.

zakkak:
I see, I admit I didn't check the latest state of mx_substratevm.py, my mistake. Thanks for making this clear Paul.

zakkak (approved):
Thanks for the fix @rschatz

rschatz:
> Unfortunately, as stated in [#2738 (comment)](https://github.com/oracle/graal/pull/2738#discussion_r468109433) `mx gate --tags build` does not suffice.

Ah, I see, that explains my confusion. Also good to hear that it's not necessary anymore ;)

.github/workflows/main.yml
olpaw: > It's not literally building everything, but it would exercise almost every different category of stuff there is to build (e.g. Java projects, native projects, a native-image launcher, ...).

Yup that would give pretty good coverage.

rschatz: Done, and seems to be working (at least on github, still waiting for the internal gate to finish).

zakkak:
Thanks @rschatz

PR #3490, opened 2021-06-18
Fix: neo4j-analytics benchmark.

jovanstevanovic:
The **neo4j-analytics** benchmark is not working due to a proxy initialization problem. 

Stacktrace is:

> java.lang.NoSuchMethodError: org.neo4j.kernel.impl.api.index.IndexingService$Monitor.awaitingPopulationOfRecoveredIndex(org.neo4j.storageengine.api.schema.StoreIndexDescriptor)
> 	at com.sun.proxy.$Proxy196.<clinit>(Unknown Source)
> The following benchmarks failed: neo4j-analytics
> 	at com.oracle.svm.core.classinitialization.ClassInitializationInfo.invokeClassInitializer(ClassInitializationInfo.java:375)
> 	at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:295)
> 	at java.lang.reflect.Constructor.newInstance(Constructor.java:490)
> 	at java.lang.reflect.Proxy.newProxyInstance(Proxy.java:1022)
> 	at java.lang.reflect.Proxy.newProxyInstance(Proxy.java:1008)
>     ....
	
Reason: 
Each Proxy class contains static fields as a reference toward methods, something like `private static Method m1 = Class.getMethod("method1")`. We are replacing these reflective calls with constants inside `ReflectionPlugin`. If`method1` is receiving an object (non-primitive type) we are adding `EnsureInitilizedNode ` (to check if a class is initialized) and in that case, this part of the code is not working:  https://github.com/oracle/graal/blob/1637cd77629fd4ffbb14295fc79b486824307fa8/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java#L373-L393 .

Command to reproduce:
`mx --env ni-ce benchmark renaissance-native-image:neo4j-analytics -- --jvm=native-image --jvm-config=default-ce`

Link to a simpler version of reproducer:
https://github.com/jovanstevanovic/ProxyException

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
@@ -365,6 +365,16 @@ public class SubstrateGraphBuilderPlugins {
              */
             List<Class<?>> classList = new ArrayList<>();
             FixedNode successor = unwrapNode(newArray.next());
+            /*
+             * In a case when we are creating a proxy, which contains a method with a class as a
+             * parameter, a successor node will not be StoreIndexNode, so we need to skip
+             * initialization nodes.
+             */
+            if (successor instanceof EnsureClassInitializedNode) {
+                AbstractBeginNode classInitializedNode = ((EnsureClassInitializedNode) successor).next();
+                VMError.guarantee(classInitializedNode != null);
+                successor = classInitializedNode.next();
+            }
             while (successor instanceof StoreIndexedNode) {
                 StoreIndexedNode store = (StoreIndexedNode) successor;
                 assert getDeoptProxyOriginalValue(store.array()).equals(newArray);
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
@@ -366,9 +366,9 @@ public class SubstrateGraphBuilderPlugins {
             List<Class<?>> classList = new ArrayList<>();
             FixedNode successor = unwrapNode(newArray.next());
             /*
-             * In a case when we are creating a proxy, which contains a method with a class as a
-             * parameter, a successor node will not be StoreIndexNode, so we need to skip
-             * initialization nodes.
+             * In a case when we are creating a proxy, which contains a method with a class
+             * (initialized at runtime) as a parameter, a successor node will not be StoreIndexNode,
+             * so we need to skip initialization nodes.
              */
             if (successor instanceof EnsureClassInitializedNode) {
                 AbstractBeginNode classInitializedNode = ((EnsureClassInitializedNode) successor).next();
--- a/substratevm/src/com.oracle.svm.test/src/META-INF/native-image/proxy-config.json
+++ b/substratevm/src/com.oracle.svm.test/src/META-INF/native-image/proxy-config.json
@@ -1,0 +1,3 @@
+[
+  ["com.oracle.svm.test.proxy.CustomProxy"]
+]
--- a/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/proxy/CustomInvocationHandler.java
+++ b/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/proxy/CustomInvocationHandler.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2021, 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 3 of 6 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
```diff
@@ -371,6 +371,13 @@ private static void interceptProxyInterfaces(GraphBuilderContext b, ResolvedJava
              */
             List<Class<?>> classList = new ArrayList<>();
             FixedNode successor = unwrapNode(newArray.next());
+            if (successor instanceof EnsureClassInitializedNode) {
+                /*
+                 * Skip EnsureClassInitialized and Begin nodes.
+                 */
+                EnsureClassInitializedNode classInitializedNode = ((EnsureClassInitializedNode) successor);
+                successor = classInitializedNode.next().next();
```
vjovanov, line 379: Instead of having a three-line comment, I would simply cast to a Begin node (potentially assert on it) and the do next. You say the same, but with code instead of the comment.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
```diff
@@ -371,6 +371,13 @@ private static void interceptProxyInterfaces(GraphBuilderContext b, ResolvedJava
              */
             List<Class<?>> classList = new ArrayList<>();
             FixedNode successor = unwrapNode(newArray.next());
+            if (successor instanceof EnsureClassInitializedNode) {
+                /*
+                 * Skip EnsureClassInitialized and Begin nodes.
```
vjovanov, line 376: If this comment stays, I would make it one line.

vjovanov:
Can you make a small reproducer and a small test.

vjovanov (approved):

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
```diff
@@ -371,6 +371,13 @@ private static void interceptProxyInterfaces(GraphBuilderContext b, ResolvedJava
              */
             List<Class<?>> classList = new ArrayList<>();
             FixedNode successor = unwrapNode(newArray.next());
+            if (successor instanceof EnsureClassInitializedNode) {
```
vjovanov, line 374: Maybe mention when this happens.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/SerializationSupport.java
```diff
@@ -80,12 +80,10 @@ private StubForAbstractClass() {
 
     private static final class SerializationLookupKey {
         private final Class<?> declaringClass;
-        private final Class<?> targetConstructorClass;
 
-        private SerializationLookupKey(Class<?> declaringClass, Class<?> targetConstructorClass) {
-            assert declaringClass != null && targetConstructorClass != null;
+        private SerializationLookupKey(Class<?> declaringClass) {
```
olpaw, line 84: As already mentioned in our meeting today. This **breaks the serialization feature** and will not pass our internal CI gates. The correct way to fix your serialization issues for the benchmarks is to provide custom configs for special cases of serialization used by these benchmarks. The format for the serialization config entries can be found in https://github.com/oracle/graal/blob/f07d5e3b858b2441e942958d8f329238ff0b0c87/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/doc-files/SerializationConfigurationFilesHelp.txt#L14

olpaw: For such cases the serialization error message that you get at runtime contains all info that you need to create that custom bits of serialization config. https://github.com/oracle/graal/blob/89e4cfc7aeea69970b60c64cd075ceb2a104e864/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/SerializationSupport.java#L132

`declaringClass` provides the value for json key-`name` and customTargetConstructorClass provides the value for json key-`targetConstructorClass`

olpaw (changes requested):

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
jovanstevanovic: Done.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/SerializationSupport.java
jovanstevanovic: Undrestand. I'll try to do it in another way.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
jovanstevanovic: I think that one comment, before the if statement, is enough. So I removed this one.

vjovanov:
Please separate the neo4j fix from the serialization fix. My approval was only for neo4j.

PR #3501, opened 2021-06-21
Support byte arrays on platforms without unaligned access, depending on platform endiannes

koutheir:
Support byte arrays on platforms without unaligned access, depending on platform endiannes.

### DIFF
--- a/truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/memory/ByteArraySupports.java
+++ b/truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/memory/ByteArraySupports.java
@@ -80,7 +80,8 @@ final class ByteArraySupports {
     static {
         // We only use Unsafe for platforms that we know support it, and that support unaligned
         // accesses.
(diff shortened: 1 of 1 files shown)


koutheir:
@graalvmbot, I already signed the OCA on 2018-04-03.

koutheir:
The failed checks do not seem to be related to this patch.

gilles-duboscq:
Regarding the gate checks, they were fixed by #3478, i guess your PR stems from before that.

koutheir:
I rebased on top of the last HEAD, and the failed checks do not seem to be related to this patch.

gilles-duboscq:
Thanks. I think after those 2 tests got fixed, all the tests got broken for some other reason :frowning_face:

gilles-duboscq:
Could you describe a bit more the issue you are fixing here?
As far as I can tell, `SimpleByteArraySupport` always in a big-endian fashion so i'm not sure it's OK to use it for `LITTLE_ENDIAN`.

koutheir:
> As far as I can tell, `SimpleByteArraySupport` always in a big-endian fashion so i'm not sure it's OK to use it for `LITTLE_ENDIAN`.

I missed that. This makes this PR invalid. I'm closing it.

PR #3504, opened 2021-06-22
Support predefinition of classes with subtype and outer class relationships.

peter-hofer:


olpaw (approved):
Nicely done having `PredefinedClassesRegistryImpl` `add` and `defineClass` working together to ensure `defineClass` happens in the right order.

PR #3510, opened 2021-06-23
Fix `LLVMCreateEnumAttribute()` parameter to create an `EnumAttribute` instead of an `IntAttribute`

koutheir:
The two call sites to `LLVMCreateEnumAttribute()` fixed by this patch intend to create an LLVM `EnumAttribute`, because:
1. The attribute `kind` is retrieved through calling `LLVMGetEnumAttributeKindForName()`, and
2. The attribute being set has no integer value, as it is intended to be a boolean.

However, calling `LLVMCreateEnumAttribute()` with a non-zero value (e.g., `TRUE`) causes it to create an `IntAttribute`, instead of the intended `EnumAttribute`. This behavior is not documented, but going through the LLVM source code reveals this behavior:
- `LLVMCreateEnumAttribute()` [calls `Attribute::get()`](https://github.com/llvm/llvm-project/blob/c1a0a213378a458fbea1a5c77b315c7dce08fd05/llvm/lib/IR/Core.cpp#L151).
- `Attribute::get()` only creates an `EnumAttribute` [if the provided value is zero](https://github.com/llvm/llvm-project/blob/c1a0a213378a458fbea1a5c77b315c7dce08fd05/llvm/lib/IR/Attributes.cpp#L94).

### DIFF
--- a/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMIRBuilder.java
+++ b/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMIRBuilder.java
@@ -24,6 +24,7 @@
  */
 package com.oracle.svm.core.graal.llvm.util;
 
+import static com.oracle.svm.core.graal.llvm.util.LLVMUtils.ENUM_ATTRIBUTE_VALUE;
 import static com.oracle.svm.core.graal.llvm.util.LLVMUtils.FALSE;
 import static com.oracle.svm.core.graal.llvm.util.LLVMUtils.TRUE;
 import static com.oracle.svm.core.graal.llvm.util.LLVMUtils.dumpTypes;
@@ -168,7 +169,7 @@ public class LLVMIRBuilder implements AutoCloseable {
         int kind = LLVM.LLVMGetEnumAttributeKindForName(attribute.name, attribute.name.length());
         LLVMAttributeRef attr;
         if (kind != 0) {
-            attr = LLVM.LLVMCreateEnumAttribute(context, kind, TRUE);
+            attr = LLVM.LLVMCreateEnumAttribute(context, kind, ENUM_ATTRIBUTE_VALUE);
         } else {
             String value = "true";
             attr = LLVM.LLVMCreateStringAttribute(context, attribute.name, attribute.name.length(), value, value.length());
@@ -725,7 +726,7 @@ public class LLVMIRBuilder implements AutoCloseable {
         int kind = LLVM.LLVMGetEnumAttributeKindForName(attribute.name, attribute.name.length());
         LLVMAttributeRef attr;
         if (kind != 0) {
-            attr = LLVM.LLVMCreateEnumAttribute(context, kind, TRUE);
+            attr = LLVM.LLVMCreateEnumAttribute(context, kind, ENUM_ATTRIBUTE_VALUE);
             LLVM.LLVMAddCallSiteAttribute(call, (int) LLVM.LLVMAttributeFunctionIndex, attr);
         } else {
             setCallSiteAttribute(call, attribute, "true");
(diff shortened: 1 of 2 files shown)


koutheir:
The build failures are unrelated to this PR.

loicottet:
From what I can see both approaches have the attribute included in the bitcode as an enum attribute, did you run into problems linked with that behaviour? If not, I would keep the original code since its intent is clearer.

I believe the enum attribute is created [there](https://github.com/llvm/llvm-project/blob/c1a0a213378a458fbea1a5c77b315c7dce08fd05/llvm/lib/IR/Attributes.cpp#L88), so the null check on the value is not reached in this case, as we check whether the enum attribute exists beforehand with `LLVMGetEnumAttributeKindForName`.

koutheir:
> From what I can see both approaches have the attribute included in the bitcode as an enum attribute,...

No, one approach includes the attribute as an enum attribute, and the other approach includes it as an integer attribute.

> ..., did you run into problems linked with that behaviour?

Yes I did.

> If not, I would keep the original code since its intent is clearer.

I agree that `TRUE` delivers a clearer intent, but due to the poor API design (having to provide an integer even for enum attributes), the `TRUE` argument is just wrong. I suggest adding a comment to clarify the intent in this situation, or maybe a helper function with a long name identifying the intent.

> I believe the enum attribute is created there, so the null check on the value is not reached in this case, as we check whether the enum attribute exists beforehand with LLVMGetEnumAttributeKindForName.

No, an integer is still associated with the created attribute [if the argument is non-zero](https://github.com/llvm/llvm-project/blob/c1a0a213378a458fbea1a5c77b315c7dce08fd05/llvm/lib/IR/Attributes.cpp#L85), irrespective of the provided attribute kind.

loicottet:
> No, one approach includes the attribute as an enum attribute, and the other approach includes it as an integer attribute.

Sorry, I misspoke, I meant when disassembling the resulting `.bc` file, the attribute is specified in the same way in both cases.

> Yes I did.

Can you elaborate on those? From what I could see the program's behaviour would not be different between the two attribute types.

> I agree that `TRUE` delivers a clearer intent, but due to the poor API design (having to provide an integer even for enum attributes), the `TRUE` argument is just wrong. I suggest adding a comment to clarify the intent in this situation, or maybe a helper function with a long name identifying the intent.

In that case I would suggest creating a `NULL` value in `LLVMUtils` which would convey that meaning.

> No, an integer is still associated with the created attribute [if the argument is non-zero](https://github.com/llvm/llvm-project/blob/c1a0a213378a458fbea1a5c77b315c7dce08fd05/llvm/lib/IR/Attributes.cpp#L85), irrespective of the provided attribute kind.

You're right, if the attributes are not pregenerated somehow this will end up creating an `IntAttribute`, my bad.

koutheir:
> In that case I would suggest creating a `NULL` value in `LLVMUtils` which would convey that meaning.

I would not call it `NULL` because it needs to be an integer, whereas `NULL` usually represents a sentinel for a pointer. Maybe `ENUM_ATTRIBUTE_VALUE` would better convey the meaning, making the call:
```java
attr = LLVM.LLVMCreateEnumAttribute(context, kind, ENUM_ATTRIBUTE_VALUE);
```
instead of:
```java
attr = LLVM.LLVMCreateEnumAttribute(context, kind, TRUE);
```
where `ENUM_ATTRIBUTE_VALUE` would be defined in `LLVMUtils` as:
```java
static final long ENUM_ATTRIBUTE_VALUE = 0L;
```

@loicottet, do you agree to this change?

loicottet:
Looks good to me 👍

loicottet (approved):

koutheir:
@loicottet, I implemented the changes you requested. I hope this commit can be merged now.

koutheir:
@cstancu, can you please merge this PR?

loicottet:
I'm in the process of merging the PR, should be done shortly

koutheir:
@loicottet, the PR status says this patch was rejected. Has it been merged in another way?

loicottet:
Yes, the PR was merged, no problem there. I just had to create a new internal PR with your commit which means it wasn't recognised as exactly this PR by Github.

PR #3515, opened 2021-06-25
[GR-30765] Extend Native Image diagnostic output.

graalvmbot:


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoAccess.java
```diff
@@ -142,6 +142,25 @@ public static int getState(CodeInfo info) {
...
     }
 
+    public static String stateToString(int codeInfoState) {
+        switch (codeInfoState) {
+            case CodeInfo.STATE_CREATED:
+                return "created";
+            case CodeInfo.STATE_CODE_CONSTANTS_LIVE:
+                return "code constants live";
+            case CodeInfo.STATE_NON_ENTRANT:
+                return "non entrant";
```
peter-hofer, line 152: `non-entrant`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/snippets/StackOverflowCheckImpl.java
```diff
@@ -218,7 +199,7 @@ public void disableStackOverflowChecksForFatalError() {
          * ensures that any future calls to protectYellowZone() do not modify the stack boundary
          * again.
          */
-        yellowZoneStateTL.set(0xfefefefe);
+        yellowZoneStateTL.set(0x7EFEFEFE);
```
peter-hofer, line 202: Is that change because the value was negative and the guarantees failed?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/locks/SingleThreadedVMLockSupport.java
```diff
@@ -37,8 +38,16 @@
...
+final class SingleThreadedVMLockSupport extends VMLockSupport {
+    @Override
+    public VMMutex[] getMutexes() {
+        return null;
+    }
+
+    @Override
+    public VMCondition[] getConditions() {
+        return null;
+    }
```
peter-hofer, line 50: These should either return an empty array (with a check during printing to say something like `VM Mutexes: (none)`, or a comment on the superclass should say that `null` is a valid return value.

peter-hofer (approved):
This will be a lot more useful than our current diagnostic output. Thank you!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/snippets/StackOverflowCheckImpl.java
christianhaeubl: Yes.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxStackOverflowSupport.java
```diff
@@ -37,8 +37,23 @@
...
+    @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
+    @Override
+    public UnsignedWord lookupStackBase() {
+        Pthread.pthread_attr_t attr = StackValue.get(Pthread.pthread_attr_t.class);
+        PosixUtils.checkStatusIs0(Pthread.pthread_getattr_np(Pthread.pthread_self(), attr), "LinuxStackOverflowSupport: pthread_getattr_np");
+
+        WordPointer stackaddrPtr = StackValue.get(WordPointer.class);
+        WordPointer stacksizePtr = StackValue.get(WordPointer.class);
+        PosixUtils.checkStatusIs0(Pthread.pthread_attr_getstack(attr, stackaddrPtr, stacksizePtr), "LinuxStackOverflowSupport: pthread_attr_getstack");
+        PosixUtils.checkStatusIs0(Pthread.pthread_attr_destroy(attr), "LinuxStackOverflowSupport: pthread_attr_destroy");
```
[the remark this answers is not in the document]
christianhaeubl: That shouldn't be an issue as the stack boundaries are cached in thread-local values during thread startup. The segfault handler only accesses those thread-local values.

PR #3523, opened 2021-06-28
[GR-32146] Introduce an abstraction around the BigBang/Inflation.

graalvmbot:


substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/AnalysisPolicy.java
```diff
@@ -143,13 +143,13 @@ public abstract AbstractSpecialInvokeTypeFlow createSpecialInvokeTypeFlow(Byteco
                     TypeFlow<?>[] actualParameters, ActualReturnTypeFlow actualReturn, BytecodeLocation location);
 
     @SuppressWarnings("unused")
-    public int makePoperties(BigBang bb, AnalysisObject... objects) {
+    public int makeProperties(StaticAnalysisEngine analysis, AnalysisObject... objects) {
```
christianwimmer, line 146: I don't like that variables/parameters/fields are now called `analysis` instead of `bb`. `analysis` is a very generic term. But `staticAnalysisEngine` everywhere is of course annoyingly long.

So why not do the renaming "the other way round" and keep the name `BigBang` / variable name `bb` for what you now call `StaticAnalysisEngine`? The implementation classes can then be properly named `ReachabilityAnalysis` and `PointsToAnalysis` - but there are very few cases where you actually need to cast `BigBang` to `PointsToAnalysis`.

To minimize renamings, you can also keep the variable name `bb` when the type is actually already cast to `PointsToAnalysis`.

christianwimmer (changes requested):

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/meta/AnalysisMetaAccess.java
```diff
@@ -47,6 +48,14 @@ public AnalysisType lookupJavaType(Class<?> clazz) {
         return (AnalysisType) super.lookupJavaType(clazz);
     }
 
+    public AnalysisType lookupJavaType(String className) {
```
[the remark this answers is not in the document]
d-kozak: I agree. I was very unhappy making the commit https://github.com/oracle/graal/pull/3523/commits/39374fc5a798c01a4b5174b24a3fb2503003f067
There is actually only one place where the method is used - the  [PermissionsFeature](https://github.com/oracle/graal/blob/39374fc5a798c01a4b5174b24a3fb2503003f067/substratevm/src/com.oracle.svm.truffle.tck/src/com/oracle/svm/truffle/tck/PermissionsFeature.java) in the *com.oracle.svm.truffle.tck* project. Unfortunately, the CheckStyle configuration of that project does not allow dynamic class loading. I was not sure whether I'm allowed to change the CheckStyle config, but if you think it is reasonable, I can do so. Or alternatively, move the utility method somewhere else. However, since there is only one use of it, I would be happier if it could stay in  *com.oracle.svm.truffle.tck*

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/AnalysisPolicy.java
d-kozak: I was feeling that the shift from `BigBang` to `StaticAnalysisEngine` was a good one because the `BigBang/Inflation` seemed more confusing than descriptive to me, especially when I was starting. But it is true that `analysis` is very generic and  `staticAnalysisEngine` annoyingly long, `sbe` also does not sound right. I'll spend some time thinking about a different naming strategy. Then I'll either share the results here for further discussion or apply the steps you proposed in case I can't find a different solution.

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/meta/AnalysisMetaAccess.java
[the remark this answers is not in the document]
d-kozak: I moved the class loading back to the [PermissionsFeature](https://github.com/oracle/graal/blob/39374fc5a798c01a4b5174b24a3fb2503003f067/substratevm/src/com.oracle.svm.truffle.tck/src/com/oracle/svm/truffle/tck/PermissionsFeature.java).

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/meta/AnalysisMethod.java
```diff
@@ -258,6 +259,28 @@ public void registerAsInlined() {
         return getInvokeLocations().stream().map(location -> (AnalysisMethod) location.getMethod()).collect(Collectors.toSet());
     }
 
+    public static class CallContext {
```
cstancu, line 262: Why is this class necessary?

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/reports/CallTreePrinter.java
```diff
@@ -166,29 +164,29 @@ public void buildCallTree() {
              * Process the method: iterate the invokes, for each invoke iterate the callees, if the
              * callee was not already processed add it to the tree and to the work list.
              */
-            node.method.getTypeFlow().getInvokes().stream()
-                            .sorted(invokeComparator)
-                            .forEach(invoke -> processInvoke(invoke, node, workList));
-
+            // todo test nothing has been broken here
```
cstancu, line 167: Please make sure there are no TODOs left behind. Why is it necessary to introduce `AnalysisMethod.getCallees()`?

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/typestate/MultiTypeState.java
```diff
@@ -332,8 +333,12 @@ public void noteMerge(BigBang bb) {
 
     @Override
     public boolean closeToAllInstantiated(BigBang bb) {
-        if (typesCount > 200 && bb != null) {
-            MultiTypeState allInstState = (MultiTypeState) bb.getAllInstantiatedTypeFlow().getState();
+        if (!(bb instanceof PointsToAnalysis)) {
+            return false;
```
cstancu, line 337: `MultiTypeState` is a `pointsto` class, can this condition ever be true?

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/AnalysisObjectScanner.java
```diff
@@ -36,8 +36,15 @@
...
-        super(bigbang, executor, scannedObjects);
+    /**
+     * As opposed to the other subclasses of Object scanner, this one is very points-to specific,
+     * therefore it is keeping a reference to BigBang.
+     */
+    private final PointsToAnalysis analysis;
+
+    public AnalysisObjectScanner(PointsToAnalysis bb, CompletionExecutor executor, ReusableSet scannedObjects) {
+        super(bb, executor, scannedObjects);
+        this.analysis = bb;
```
cstancu, line 47: This class already has access to `bb` from its super class. It just needs to be cast to `PointsToAnalysis` on some accesses; that could be wrapped in a method to keep the code compact.

cstancu (changes requested):
I see that there are a lot of places where variables/parameters were renamed from `bigbang` to `bb`. You can extract those in a separate PR and merge them to simplify the diff and make it easier to see significant changes.

d-kozak:
I'm extracting the `bigbang -> bb` renaming into a separate PR: https://github.com/oracle/graal/pull/3683

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/meta/AnalysisMethod.java
d-kozak: Currently,  `CallTreePrinter` uses `MethodTypeFlow` and `InvokeTypeFlow`, which are points-to specific classes. The CallContext was my first attempt to create an abstraction around it. However, I was not particularly happy about it, so I will be grateful for your feedback & suggestions. I can think of the following ways.

1. Create a helper method on `AnalysisMethod` providing the information necessary for building the call tree. This is essentially what I was trying to achieve with `CallContext` and `getCallees` - get the same kind of information, but without using points-to specific classes.
2. Move the whole call-tree building algorithm behind an abstraction, so that different types of analysis can construct it in different ways. We could for example have an abstraction to create the full call graph and then another algorithm to convert the call graph into a call tree. The call graph to call tree conversion could already be shared.
3. Leave it be for now and let the `CallTreePrinter` be points-to dependent. This would of course be the easiest option, but I will (hopefully soon) have to solve this problem somehow anyway, so I prefer to resolve it now then later.

What is your opinion? I believe the second approach could be the cleanest. It will require more work now, but having an API for creating call graphs might be useful for other purposes as well (finding paths to 'unwanted methods' or between different methods, some lightweight checks, code smell detection, etc) and I believe there are other places across the project that generate call graphs, the [PermissionsFeature](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.truffle.tck/src/com/oracle/svm/truffle/tck/PermissionsFeature.java) for example, and I could try to design the API in such a way that these clients can use it instead of having their own call graph creation algorithms.

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/typestate/MultiTypeState.java
d-kozak: I've been thinking about the possibility of reusing the `TypeState` classes even outside the points-to analysis. An abstraction for sets of types and operations on them sounds useful, but these classes currently contain quite a lot of points-to specific code for that. If the time comes, I can try to refactor it, but that is currently a non-goal, so I'll just remove the check for now.

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/AnalysisObjectScanner.java
d-kozak: Good point, I've refactored it.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/analysis/Inflation.java
```diff
@@ -24,718 +24,28 @@
...
-                    }
-                    hub.initEnumConstants(enumConstants);
-                }
-            }
-        }
-    }
+/**
+ * Extension of StaticAnalysisEngine providing access to more svm specific methods.
+ */
+public interface Inflation extends BigBang {
```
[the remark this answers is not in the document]
d-kozak: `Inflation` is an interface, whereas `NativeImagePointsToAnalysis ` is an actual implementation of it. Since there will be a different implementation of `BigBang`for the Reachability Analysis, I'm keeping the Inflation interface for that class (or its subclass) to implement as well. There are places where `Inflation` is used, not `BigBang` and once there are multiple implementations of it, we would have to reintroduce it anyway, if I am not mistaken.

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/meta/AnalysisMethod.java
d-kozak: After discussions with @vjovanov, we decided to follow option 3 and just leave it be for now if no one objects. When the time comes to implement the call tree printing for Reachability Analysis, we will know more about what the shared API could look like and therefore we should be able to make a better decision.

d-kozak:
I've implemented the proposed changes, rebased & resolved all conflicts. All gates are passing again, therefore I would like to ask for another review.

vjovanov:
I would squash the fixup commits into a single one.

vjovanov (approved):

d-kozak:
> I would squash the fixup commits into a single one.

Done. I've rewritten the last commits so that the PR can be more easily reviewed commit by commit. The history goes as follows (summary):
1. Introduction of the interfaces
2. Changing the codebase to use them.
3. Resolving non-trivial cases 1 by 1
4. Switching the naming convention - `Bigbang/Inflation` became interfaces (this may make reviewing changes in those classes harder)
5. rebase & resolve conflicts
6. applying changes from the review
7. rebase & resolve conflicts

Issue #3492, opened 2021-06-18
Add `--engine.ExcludeInline` flag to exclude method inlining 


wildmaples:
**Feature request**

Please include the following information:

**Is your feature request related to a problem? Please describe.**

We're looking into optimising keyword arguments in TruffleRuby and need a way to benchmark methods without the inlining mechanism. 

**Describe the solution you'd like.**

Looking for something we can use when running, e.g. `jt ruby --engine.ExcludeInline foo.rb`.

**Describe who do you think will benefit the most.**

Developer of libraries and frameworks depending on GraalVM. 

**Describe alternatives you've considered.**

We've considered creating a method within a benchmark that meets all the "requirements" to avoid the inlining mechanism, but because the [logic/policy for inlining is not trivial](https://github.com/oracle/graal/blob/master/compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/DefaultInliningPolicy.java), we cannot easily make a method avoid inlining without additional complexities. It's simpler to ask graal to avoid it at runtime. 

**Express whether you'd like to help contributing this feature**
If you'd like to contribute, please read the [contribution guide](https://www.graalvm.org/community/contributors/).

I'll be happy to work on this feature.

chrisseaton:
We can turn off inlining with `--engine.Inlining=false` - I think we're more specifically looking for a feature to disable inlining for specific methods to help us benchmark more easily. So `--engine.ExcludeInline=foo` like `--engine.CompileOnly=foo`.

chumer:
> I'll be happy to work on this feature.

Cool!

You can add new engine options in PolyglotCompilerOptions. Should be straight forward.

After that you probably needs some changes in `AgnosticInliningPhase`.

Accessing the option is straight forward:
`request.options.get(PolyglotCompilerOptions.ExcludeInlining)`

You can access the compilation unit name with: CallNode#truffleAST#getName()

As Chris pointed out, syntax should probably be similar to `--engine.CompileOnly`.

@boris-spas is probably the right person to shepherd this PR through the internal gates. (assigning him)

boris-spas:
@wildmaples Hey! Cool! Feel free to ping me if you need any info :)

shreelakshmijoshi:
Hi @wildmaples !
I am a new open source contributor and would like to solve this issue, could you help me understand the issue better by providing some extra references ? 
Is this is issue open to contribute ?

[...]

PR #3524, opened 2021-06-28
Add option to exclude inlining, `--engine.ExcludeInlining=`

wildmaples:
# Issue 

https://github.com/oracle/graal/issues/3492

Add an option to exclude a method `--engine.ExcludeInlining=bloooop`  

#  Changes

When the nodes are being checked in `DefaultInliningPolicy.java`, I added a check to see if the node is a method that was excluded. If it is, we avoid the inlining and `continue;`. 

I'm not sure if you'd like this change, but I've also added a new state, `CallNode.State.Excluded` to mark a node excluded. This makes the `--engine.TraceInlining` output more readable to the user (see the trace below). Without this change, the nodes would remain as `Expanded`. 

# Test

I'm testing this change locally using this Ruby script `test.rb`:

```ruby
def baz; end

def bar
  baz
end

# Call #bar 4 times
def foo 
  bar
  bar
  bar
  bar
  baz
end

loop do
  foo
end
```

And running

`jt ruby --experimental-options --engine.ExcludeInlining=bar --engine.TraceInlining --engine.TraceCompilation --engine.MultiTier=false --engine.OSR=false --engine.CompilationThreshold=10000000 test.rb` 

Which is essentially `jt ruby --experimental-options --engine.ExcludeInlining=bar test.rb` but with additional options to remove irrelevant noise. 

The output is then:

```
➜  jt ruby --experimental-options --engine.ExcludeInlining=bar --engine.TraceInlining --engine.TraceCompilation --engine.MultiTier=false --engine.OSR=false --engine.CompilationThreshold=10000000 test.rb
WARNING: jt is not running on MRI or TruffleRuby Native, startup is slow
  Consider using following bash alias to run on MRI.
  `alias jt=/path/to/mri/bin/ruby /path/to/truffleruby/tool/jt.rb`
Using TruffleRuby with Graal: mxbuild/truffleruby-jvm-ce
$ /Users/mapleong/src/github.com/Shopify/truffleruby/mxbuild/truffleruby-jvm-ce/languages/ruby/bin/ruby \
  --experimental-options \
  --core-load-path=/Users/mapleong/src/github.com/Shopify/truffleruby/src/main/ruby/truffleruby \
  --experimental-options \
  --engine.ExcludeInlining=bar \
  --engine.TraceInlining \
  --engine.TraceCompilation \
  --engine.MultiTier=false \
  --engine.OSR=false \
  --engine.CompilationThreshold=10000000 \
  test.rb
Warning: Option 'engine.CompilationThreshold' is deprecated and might be removed from future versions.

[To redirect Truffle log output to a file use one of the following options:
* '--log.file=<path>' if the option is passed using a guest language launcher.
* '-Dpolyglot.log.file=<path>' if the option is passed using the host Java launcher.
* Configure logging using the polyglot embedding API.]
[engine] Inline start Object#baz                                                  |call diff     0.00|Recursion Depth      0|Explore/inline ratio     1.00|IR Nodes     23|Frequency     1.00|Truffle Callees      0|Forced false|Depth      0
[engine] Inline done  Object#baz                                                  |call diff     0.00|Recursion Depth      0|Explore/inline ratio     1.00|IR Nodes     23|Frequency     1.00|Truffle Callees      0|Forced false|Depth      0
[engine] opt done     Object#baz                                                  |AST    6|Tier 2|Time   85(  76+9   )ms|Inlined   0Y   0N|IR    23/   20|CodeSize    174|Addr 0x11ec9ea90|Src test.rb:1
[engine] Inline start Object#bar                                                  |call diff     0.00|Recursion Depth      0|Explore/inline ratio     1.00|IR Nodes     55|Frequency     1.00|Truffle Callees      1|Forced false|Depth      0
[engine] Inlined        Object#baz                                                |call diff    -1.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     25|Frequency     1.00|Truffle Callees      0|Forced false|Depth      1
[engine] Inline done  Object#bar                                                  |call diff     0.00|Recursion Depth      0|Explore/inline ratio     1.00|IR Nodes     55|Frequency     1.00|Truffle Callees      1|Forced false|Depth      0
[engine] opt done     Object#bar                                                  |AST   14|Tier 2|Time   58(  56+2   )ms|Inlined   1Y   0N|IR    27/   28|CodeSize    214|Addr 0x11ed0ce90|Src test.rb:3
[engine] Inline start Object#foo                                                  |call diff     0.00|Recursion Depth      0|Explore/inline ratio     5.00|IR Nodes    165|Frequency     1.00|Truffle Callees      5|Forced false|Depth      0
[engine] Excluded       Object#bar                                                |call diff    -2.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     66|Frequency     1.00|Truffle Callees      1|Forced false|Depth      1
[engine] Excluded       Object#bar                                                |call diff    -2.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     66|Frequency     1.00|Truffle Callees      1|Forced false|Depth      1
[engine] Excluded       Object#bar                                                |call diff    -2.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     66|Frequency     1.00|Truffle Callees      1|Forced false|Depth      1
[engine] Excluded       Object#bar                                                |call diff    -2.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     66|Frequency     1.00|Truffle Callees      1|Forced false|Depth      1
[engine] Inlined        Object#baz                                                |call diff    -1.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     25|Frequency     1.00|Truffle Callees      0|Forced false|Depth      1
[engine] Inline start block in <main> <split-6c701503>                            |call diff     0.00|Recursion Depth      0|Explore/inline ratio     3.67|IR Nodes    214|Frequency     1.00|Truffle Callees      1|Forced false|Depth      0
[engine] Inline done  Object#foo                                                  |call diff     0.00|Recursion Depth      0|Explore/inline ratio     5.00|IR Nodes    165|Frequency     1.00|Truffle Callees      5|Forced false|Depth      0
[engine] Inlined        Object#foo                                                |call diff   -10.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes    160|Frequency     1.00|Truffle Callees      5|Forced false|Depth      1
[engine] Excluded         Object#bar                                              |call diff    -2.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     66|Frequency     1.00|Truffle Callees      1|Forced false|Depth      2
[engine] Excluded         Object#bar                                              |call diff    -2.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     66|Frequency     1.00|Truffle Callees      1|Forced false|Depth      2
[engine] Excluded         Object#bar                                              |call diff    -2.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     66|Frequency     1.00|Truffle Callees      1|Forced false|Depth      2
[engine] Excluded         Object#bar                                              |call diff    -2.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     66|Frequency     1.00|Truffle Callees      1|Forced false|Depth      2
[engine] Inlined          Object#baz                                              |call diff    -1.00|Recursion Depth      0|Explore/inline ratio      NaN|IR Nodes     25|Frequency     1.00|Truffle Callees      0|Forced false|Depth      2
[engine] Inline done  block in <main> <split-6c701503>                            |call diff     0.00|Recursion Depth      0|Explore/inline ratio     3.67|IR Nodes    214|Frequency     1.00|Truffle Callees      1|Forced false|Depth      0
[engine] opt done     Object#foo                                                  |AST   46|Tier 2|Time  235(  52+183 )ms|Inlined   1Y   4N|IR   134/  515|CodeSize   2066|Addr 0x11edd7810|Src test.rb:8
[engine] opt done     block in <main> <split-6c701503>                            |AST   12|Tier 2|Time  241(  58+183 )ms|Inlined   2Y   4N|IR   166/  544|CodeSize   2066|Addr 0x11eddba10|Src test.rb:16
^C[engine] opt deopt    block in <main> <split-6c701503>                            |AST   12|Src test.rb:16
```

You can see trace compilation line for `#foo` contains `Inlined   1Y   4N`, where I believe [`1Y` is the inlined call to `#baz`](https://github.com/oracle/graal/blob/c7c061b3230852e9582badf788b3dab74a809ca9/comp[…]vm/compiler/truffle/runtime/debug/TraceCompilationListener.java) and `4N` refers to the 4 calls to `#bar`. 

You can also see that the `#bar` calls are marked `Excluded` in the inlining trace. 

Is there a test file for these classes? 

# Questions 

* What do you think? Was this similar to what you had in mind? 
* Do I have to add the exclusion check to other InliningPolicies as well? For example, `TrivialOnlyInliningPolicy`?

### DIFF
--- a/compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/DefaultInliningPolicy.java
+++ b/compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/DefaultInliningPolicy.java
@@ -114,6 +114,9 @@ final class DefaultInliningPolicy implements InliningPolicy {
         final PriorityQueue<CallNode> inlineQueue = getQueue(tree, CallNode.State.Expanded);
         CallNode candidate;
         while ((candidate = inlineQueue.poll()) != null) {
+            if (!InliningPolicy.acceptForInline(candidate, options)) {
+                continue;
+            }
             if (candidate.isTrivial()) {
                 candidate.inline();
                 continue;
--- a/compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/TrivialOnlyInliningPolicy.java
+++ b/compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/TrivialOnlyInliningPolicy.java
@@ -24,10 +24,22 @@
  */
 package org.graalvm.compiler.truffle.compiler.phases.inlining;
 
+import org.graalvm.compiler.truffle.options.PolyglotCompilerOptions;
+import org.graalvm.options.OptionValues;
+
 final class TrivialOnlyInliningPolicy implements InliningPolicy {
+    private final OptionValues options;
+
+    TrivialOnlyInliningPolicy(OptionValues options) {
+        this.options = options;
+    }
+
     @Override
     public void run(CallTree tree) {
         for (CallNode child : tree.getRoot().getChildren()) {
+            if (!InliningPolicy.acceptForInline(child, options)) {
+                continue;
+            }
             if (child.isTrivial()) {
                 child.expand();
                 child.inline();
--- a/compiler/src/org.graalvm.compiler.truffle.options/src/org/graalvm/compiler/truffle/options/PolyglotCompilerOptions.java
+++ b/compiler/src/org.graalvm.compiler.truffle.options/src/org/graalvm/compiler/truffle/options/PolyglotCompilerOptions.java
@@ -385,6 +385,9 @@ public final class PolyglotCompilerOptions {
 
     // Inlining
 
+    @Option(help = "Restrict inlined methods to ','-separated list of includes (or excludes prefixed with '~').", category = OptionCategory.INTERNAL)
+    public static final OptionKey<String> InlineOnly = new OptionKey<>("", OptionType.defaultType(String.class));
+
     @Option(help = "Enable automatic inlining of guest language call targets.", category = OptionCategory.EXPERT)
     public static final OptionKey<Boolean> Inlining = new OptionKey<>(true);
 
--- a/compiler/src/org.graalvm.compiler.truffle.test/src/org/graalvm/compiler/truffle/test/inlining/InlineOnlyTest.java
+++ b/compiler/src/org.graalvm.compiler.truffle.test/src/org/graalvm/compiler/truffle/test/inlining/InlineOnlyTest.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package org.graalvm.compiler.truffle.test.inlining;
+
+import com.oracle.truffle.api.CallTarget;
+import com.oracle.truffle.api.CompilerAsserts;
+import com.oracle.truffle.api.RootCallTarget;
+import com.oracle.truffle.api.Truffle;
+import com.oracle.truffle.api.TruffleLanguage;
+import com.oracle.truffle.api.TruffleRuntime;
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.nodes.DirectCallNode;
+import com.oracle.truffle.api.nodes.RootNode;
+import com.oracle.truffle.api.test.polyglot.ProxyLanguage;
+import org.graalvm.polyglot.Context;
+import org.junit.Test;
+
+public class InlineOnlyTest {
+    private static final String COMPILATION_ROOT_NAME = "main";
+    private static final String METHOD_EXCLUDED_FROM_INLINING = "~should-not-be-inlined";
+    private static final String METHOD_INLINED = "should-be-inlined,another-inlined-method";
+
+    @Test
+    public void testNoInliningForMethodsWithPattern() throws Exception {
+        try (Context c = Context.newBuilder().allowExperimentalOptions(true).//
+                        option("engine.CompilationFailureAction", "Throw").//
+                        option("engine.CompileImmediately", "true").//
+                        option("engine.BackgroundCompilation", "false").//
+                        option("engine.CompileOnly", COMPILATION_ROOT_NAME).//
+                        option("engine.InlineOnly", METHOD_EXCLUDED_FROM_INLINING).//
+                        build()) {
+            c.eval(InlineOnlyTest.ExcludePatternTestLanguage.ID, "");
+            c.eval(InlineOnlyTest.ExcludePatternTestLanguage.ID, "");
+        }
+    }
+
+    @Test
+    public void testInlinedMethods() throws Exception {
+        try (Context c = Context.newBuilder().allowExperimentalOptions(true).//
+                        option("engine.CompilationFailureAction", "Throw").//
+                        option("engine.CompileImmediately", "true").//
+                        option("engine.BackgroundCompilation", "false").//
+                        option("engine.CompileOnly", COMPILATION_ROOT_NAME).//
+                        option("engine.InlineOnly", METHOD_INLINED).//
+                        build()) {
+            c.eval(InlineOnlyTest.InlineSelectedMethodsTestLanguage.ID, "");
+            c.eval(InlineOnlyTest.InlineSelectedMethodsTestLanguage.ID, "");
+        }
+    }
+
+    @TruffleLanguage.Registration(id = InlineOnlyTest.ExcludePatternTestLanguage.ID, name = "InlineOnlyTest: With Exclude Pattern Test Language", version = "1.0")
+    public static class ExcludePatternTestLanguage extends ProxyLanguage {
+
+        public static final String ID = "truffle-exclude-pattern-test-language";
+
+        @Override
+        protected CallTarget parse(ParsingRequest request) throws Exception {
+            final TruffleRuntime runtime = Truffle.getRuntime();
+            final RootCallTarget rootCallTarget = runtime.createCallTarget(new RootNode(this) {
+
+                @Override
+                public String toString() {
+                    return getName();
+                }
+
+                @Override
+                public Object execute(VirtualFrame frame) {
+                    CompilerAsserts.neverPartOfCompilation("This node should not be inlined");
+                    return 99;
+                }
+
+                @Override
+                public String getName() {
+                    return METHOD_EXCLUDED_FROM_INLINING;
+                }
+            });
+            return runtime.createCallTarget(new RootNode(this) {
+
+                @Child DirectCallNode callNode = runtime.createDirectCallNode(rootCallTarget);
+
+                @Override
+                public String toString() {
+                    return getName();
+                }
+
+                @Override
+                public String getName() {
+                    return COMPILATION_ROOT_NAME;
+                }
+
+                @Override
+                public Object execute(VirtualFrame frame) {
+                    return callNode.call();
+                }
+            });
+        }
+    }
+
+    @TruffleLanguage.Registration(id = InlineOnlyTest.InlineSelectedMethodsTestLanguage.ID, name = "InlineOnlyTest: Inline Methods", version = "1.0")
+    public static class InlineSelectedMethodsTestLanguage extends ProxyLanguage {
+
+        public static final String ID = "truffle-inline-selected-methods-test-language";
+
+        @Override
+        protected CallTarget parse(ParsingRequest request) throws Exception {
+            final TruffleRuntime runtime = Truffle.getRuntime();
+            final RootCallTarget rootCallTarget = runtime.createCallTarget(new RootNode(this) {
+
+                @Override
+                public String toString() {
+                    return getName();
+                }
+
+                @Override
+                public Object execute(VirtualFrame frame) {
+                    CompilerAsserts.compilationConstant(METHOD_INLINED);
+                    return 99;
+                }
+
+                @Override
+                public String getName() {
+                    return METHOD_INLINED;
+                }
+            });
+            return runtime.createCallTarget(new RootNode(this) {
+
+                @Child DirectCallNode callNode = runtime.createDirectCallNode(rootCallTarget);
+
+                @Override
+                public String toString() {
+                    return getName();
+                }
+
+                @Override
+                public String getName() {
+                    return COMPILATION_ROOT_NAME;
+                }
+
+                @Override
+                public Object execute(VirtualFrame frame) {
+                    return callNode.call();
+                }
+            });
+        }
+    }
+}
--- a/compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/InliningPolicy.java
+++ b/compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/InliningPolicy.java
@@ -24,6 +24,12 @@
  */
 package org.graalvm.compiler.truffle.compiler.phases.inlining;
 
+import org.graalvm.collections.Pair;
+import org.graalvm.compiler.truffle.options.PolyglotCompilerOptions;
+import org.graalvm.options.OptionValues;
+
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 
 @SuppressWarnings("unused")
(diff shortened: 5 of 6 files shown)


compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/CallNode.java
```diff
@@ -418,6 +426,7 @@ public void collectInlinedTargets(TruffleInliningData inliningPlan) {
 
     public enum State {
         Cutoff,
+        Excluded,
```
boris-spas, line 429: I'm not a fan of adding an additional state for this. While it might seem useful to have the `Excluded` in the inlining trace I think that the state of the of the node and reason for that state are orthogonal issues. The policy should just check if the targets was excluded and decide not to inline it, leaving it's state as Explored (or potentially Cutoff if we decide to not even explore excluded nodes).

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/DefaultInliningPolicy.java
```diff
@@ -111,9 +111,14 @@ public Object newCallNodeData(CallNode callNode) {
 
     private void inline(CallTree tree) {
         final int inliningBudget = options.get(PolyglotCompilerOptions.InliningInliningBudget);
+        final String excludedMethod = options.get(PolyglotCompilerOptions.ExcludeInlining);
         final PriorityQueue<CallNode> inlineQueue = getQueue(tree, CallNode.State.Expanded);
         CallNode candidate;
         while ((candidate = inlineQueue.poll()) != null) {
+            if (candidate.isExcluded(excludedMethod)) {
+                candidate.exclude();
```
boris-spas, line 119: As mentioned before, I'd remove this call.

compiler/src/org.graalvm.compiler.truffle.options/src/org/graalvm/compiler/truffle/options/PolyglotCompilerOptions.java
```diff
@@ -398,6 +398,9 @@ private String indent(int nameLength) {
     @Option(help = "Perform a set of optimizations on each explored call target during inlining.", category = OptionCategory.INTERNAL)
     public static final OptionKey<Boolean> InliningTruffleTierOnExpand = new OptionKey<>(true);
 
+    @Option(help = "Exclude a specific method from being inlined.", category = OptionCategory.INTERNAL)
+    public static final OptionKey<String> ExcludeInlining = new OptionKey<>("");
```
boris-spas, line 402: I recall that in the issue we discussed having `ExcludeInline` use a similar syntax to `CompileOnly` (https://github.com/oracle/graal/issues/3492#issuecomment-867026345) i.e. a comma separated list of things to exclude.

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/CallNode.java
```diff
@@ -358,6 +362,10 @@ public boolean isTrivial() {
         return trivial;
     }
 
+    public boolean isExcluded(String optionsExcludeInlining) {
+        return !optionsExcludeInlining.isEmpty() && getTruffleAST().getName().contains(optionsExcludeInlining);
```
boris-spas, line 366: The node should not be aware of the notion of exclusion, it's up to the policy (the CallNode already has a `getName` that exposes the `truffleAST.getName()`

boris-spas (commented):
Re: `TrivialOnlyInliningPolicy` it should not be too much trouble to add the exclusion logic to it as well.

Re: tests - Feel free to add a new test class similar to org.graalvm.compiler.truffle.test.inlining.InliningTest that tests the expected behaviour.

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/CallNode.java
wildmaples: Right now, if a node takes the excluded path (i.e. [gets continued without inlined](https://github.com/oracle/graal/pull/3524/files#diff-e7b7dbb1dab8bdbd05febd1974d7ee8c960438950b1f69932ea14df31bf0bc4bR120)) the status remains as `Expanded` and not `Explored` or `Cutoff`. Is that a status you'd expected? And am I to manually set it to something else?

boris-spas: Sorry, a bit of a terminology mixup here, Expanded and Explored are the same thing, but I should have used Expanded since that's in the codebase. I think it's fine to keep it as `Expanded`, that just means that we did the partial evaluation of that call target but decided not to inline it.

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/CallNode.java
wildmaples: Fixed! I removed this function from CallNode and did the check directly within the policy - https://github.com/oracle/graal/pull/3524/files#diff-e7b7dbb1dab8bdbd05febd1974d7ee8c960438950b1f69932ea14df31bf0bc4bR140

compiler/src/org.graalvm.compiler.truffle.options/src/org/graalvm/compiler/truffle/options/PolyglotCompilerOptions.java
wildmaples: I totally misunderstood the initial syntax comment - thanks for clarifying. Fixed though! https://github.com/oracle/graal/pull/3524/files#diff-b38f98833b5f5f1fed1b5fbf826cd34dd03990f2643a70375033422c6460878eR402

wildmaples:
@boris-spas I believe I've addressed all your comments. Let me know if there are any outstanding things I can do. Thanks! 🙏

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/DefaultInliningPolicy.java
```diff
@@ -128,6 +132,18 @@ private void inline(CallTree tree) {
         }
     }
 
+    private boolean isExcludedMethod(CallNode candidate, String[] excludedMethods) {
```
boris-spas, line 135: Can be static.

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/DefaultInliningPolicy.java
```diff
@@ -128,6 +132,18 @@ private void inline(CallTree tree) {
...
+        for (String method : excludedMethods) {
+            if (method.isEmpty()) {
+                continue;
+            }
+            if (candidate.getTruffleAST().getName().contains(method)) {
```
boris-spas, line 140: Hm... the `contains` here might be an issue as it's a bit too aggressive (e.g. `ExcludeInlining=init` would exclude `init` `initialize` `__init__`). 
The annoying part is that as of late `getName` also includes splitting info (see org.graalvm.compiler.truffle.runtime.OptimizedCallTarget#getName e.g. `foo <split-12312>` )  so using `equals` will miss any splits :/ 
The correct thing to do would be to expose the raw name of the root node through the call target but that might be overkill for this PR so maybe just explicitly handle (i.e. remove) the `<split-xxxxx>` pattern if appears in the name than use `equals`

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/TrivialOnlyInliningPolicy.java
```diff
@@ -24,14 +24,38 @@
...
 final class TrivialOnlyInliningPolicy implements InliningPolicy {
+    private final OptionValues options;
+    TrivialOnlyInliningPolicy(OptionValues options) {
+        this.options = options;
+    }
+
     @Override
     public void run(CallTree tree) {
         for (CallNode child : tree.getRoot().getChildren()) {
+            final String[] excludedMethods = options.get(PolyglotCompilerOptions.ExcludeInlining).split(",");
```
boris-spas, line 39: Pull this out of the loop, it's expensive to read the options every time.

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/TrivialOnlyInliningPolicy.java
```diff
@@ -24,14 +24,38 @@
...
             }
         }
     }
+
+    private boolean isExcludedMethod(CallNode candidate, String[] excludedMethods) {
+        for (String method : excludedMethods) {
+            if (method.isEmpty()) {
+                continue;
+            }
+            if (candidate.getTruffleAST().getName().contains(method)) {
```
boris-spas, line 55: Same as in the default policy

compiler/src/org.graalvm.compiler.truffle.options/src/org/graalvm/compiler/truffle/options/PolyglotCompilerOptions.java
```diff
@@ -391,6 +391,9 @@ private String indent(int nameLength) {
     @Option(help = "Maximum depth for recursive inlining.", category = OptionCategory.EXPERT)
     public static final OptionKey<Integer> InliningRecursionDepth = new OptionKey<>(2);
 
+    @Option(help = "Exclude from being inlined for ','-separated list of methods.", category = OptionCategory.INTERNAL)
```
boris-spas, line 394: Help message is a bit confusing, maybe `A  ','-separated list of method names that should be never be inlined`

compiler/src/org.graalvm.compiler.truffle.test/src/org/graalvm/compiler/truffle/test/inlining/ExcludeInliningTest.java
```diff
@@ -0,0 +1,106 @@
...
+    private static final String COMPILATION_ROOT_NAME = "main";
+
+    @Test
+    public void testNoInlineForExcludedMethods() throws Exception {
+        try (Context c = Context.newBuilder().allowExperimentalOptions(true).//
+                option("engine.CompilationFailureAction", "Throw").//
+                option("engine.CompileImmediately", "true").//
+                option("engine.BackgroundCompilation", "false").//
+                option("engine.CompileOnly", COMPILATION_ROOT_NAME).//
+                option("engine.ExcludeInlining", "should-not-be-inlined").//
```
boris-spas, line 51: extract `should-not-be-inlined` as a constant (like `COMPILATION_ROOT_NAME`) since it's used in several places.

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/DefaultInliningPolicy.java
wildmaples: Sounds good. I can remove the split pattern and then use `equals`, which then means having to pass in both the class and method, e.g. `Object#foo` instead of just the method `foo`.

If that's intended, should we update the description (or maybe give an example) or do we assume people have that knowledge already?

wildmaples: In any case... [`92ae6d2` (#3524)](https://github.com/oracle/graal/pull/3524/commits/92ae6d24558b9976540f19681cf35bb1e40de5b3)

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/DefaultInliningPolicy.java
```diff
@@ -132,12 +132,16 @@ private void inline(CallTree tree) {
         }
     }
 
-    private static boolean isExcludedMethod(CallNode candidate, String[] excludedMethods) {
+    static boolean isExcludedMethod(CallNode candidate, String[] excludedMethods) {
+        // Removes split information from name
+        //   e.g. "Object#foo <split-1234>" => "Object#foo"
+        final String candidateWithoutSplitInfo = candidate.getTruffleAST().getName()
+                .replaceAll(" \\<[^()]*\\>", "");
```
boris-spas, line 139: This regex seems a bit too aggressive no? It should be only if ` <split-1234>` is on the end of the string. I'd recommend something like ` <split-[0-9|a-f]*>$`

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/phases/inlining/DefaultInliningPolicy.java
boris-spas: I think it's good to provide the full name of the call target one want's to exclude. e.g. Consider the java `toString` method, running with `ExcludeInlining=toString` will very likely match more than one wants for the experiment while going `ExcludeInlining=MyExperimentClass#toString` (or whatever the syntax) would be correct.

boris-spas:
@wildmaples This looks pretty good to me, not sure why the checkstyle is complaining, would you mind running `mx eclipseformat` just to make sure and then I'll try to shepherd the PR through the internal gates.

wildmaples:
@boris-spas I ran `mx eclipseformat` - many files were affected. Perhaps I was using a different formatter config? Anyway, I only committed the relevant file to this PR.

boris-spas:
Hey @wildmaples 

After some internal discussion on this topic we agreed on a couple of improvements that need to happen for this feature: `ExcludeInlining` should be renamed to `InlineOnly`. This would match with `CompileOnly` - an already existing and familiar CLI - and ti should behave the same way, allowing one to specify a comma separated list of patterns to inline. The `CompileOnly` option allows for a negation of the pattern, which would in practice enable the same usage as with the `ExcludeInlining` approach.

This means the following for the PR:
- Copy the code that handles the pattern recognition for `CompileOnly` (org.graalvm.compiler.truffle.runtime.EngineData#acceptForCompilation) and place it as a static method in the `InliningPolicy` so that it's available for all policies. Ideally, we would reuse the code but it's non-trivial since the `EngineData` lives on the runtime side and the inlining policies are on the compiler side of the libgraal divide, so we'll have to live with a bit of duplication.
- Since this option will not be used in most cases, make the default value for it null and make sure to have an early return before any processing for that case just so we don't do anything we don't need to (even parsing the ','-separated values).

Sorry we didn't start of with this approach, I wanted to keep it limited to the exclusion use case but we figured out there are use cases where a user might want more control over what gets inlined and what doesn't. Hope it's not too much of a bother.

tomstuart:
@boris-spas This is ready for review whenever you’re ready.

chumer:
> @boris-spas This is ready for review whenever you’re ready.

@boris-spas  is on vacation this week. he will get back to you next week.

compiler/src/org.graalvm.compiler.truffle.test/src/org/graalvm/compiler/truffle/test/inlining/InlineOnlyTest.java
```diff
@@ -0,0 +1,169 @@
...
+            final RootCallTarget rootCallTarget = runtime.createCallTarget(new RootNode(this) {
+
+                @Override
+                public String toString() {
+                    return getName();
+                }
+
+                @Override
+                public Object execute(VirtualFrame frame) {
+                    CompilerAsserts.compilationConstant(METHOD_INLINED);
```
boris-spas, line 139: This does not really assert anything.  `CompilerAsserts.compilationConstant` will not fail in the interprer so tis test will pass even if we don't inline. 
You can add a boolean flag which you can set from outside that when activated asserts that com.oracle.truffle.api.CompilerDirectives#inCompiledCode and not com.oracle.truffle.api.CompilerDirectives#inCompilationRoot

boris-spas (commented):
Good stuff! few small remarks.

p.s. Sorry for the delay.

boris-spas:
Hey @wildmaples,
Any progress on this? Do you mind if I fix-up the comments I made and try to push it through CI to get merged?

tomstuart:
Hey @boris-spas. I don’t think @wildmaples has had a chance to pick this up again yet; it’s on her todo list, but I’m sure she’d be happy if you got to it first.

boris-spas:
Ok, I'll take a crack at it, should be just minor changes.

boris-spas:
@wildmaples Thanks a lot for the work!

Issue #3535, opened 2021-07-01
native-image no longer supports resource bundles in .properties file format that reside at JAR's root


zakkak:
**Describe the issue**
#3446 and specifically commit 3e5e1b8 breaks support of jar files including resource bundles in .properties file format at their root. Such an example is [org.eclipse.yasson](https://github.com/eclipse-ee4j/yasson/tree/master/src/main/resources) (seen failing in https://github.com/quarkusio/quarkus/issues/18305).

**Steps to reproduce the issue**
Please include both build steps as well as run steps
1. Build `native-image` from `master` branch
2. `git clone --branch yasson-resourcebundle https://github.com/zakkak/issue-reproducers /tmp/yasson-resourcebundle`
3. `cd /tmp/yasson-resourcebundle`
4. `mvn package`
5. `native-image -jar target/reproducer-1.0-SNAPSHOT.jar`

**Describe GraalVM and your environment:**
 - GraalVM version: **44d3ba029ef261aefaf3371e8fee9403c5e88ab4**
 - JDK major version: **11**
 - OS: **Fedora 34**
 - Architecture: **AMD64**

**More details**
```
Apply jar:file:///home/zakkak/code/tmp/yasson-reproducer/target/reproducer-1.0-SNAPSHOT.jar!/META-INF/native-image/org.eclipse/yasson/native-image.properties
Executing [
/home/zakkak/code/mandrel-packaging/mandrel-java11-21.3.0-devadb0569cd5b/bin/java \
-XX:+UseParallelGC \
-XX:+UnlockExperimentalVMOptions \
-XX:+EnableJVMCI \
-Dtruffle.TrustAllTruffleRuntimeProviders=true \
-Dtruffle.TruffleRuntime=com.oracle.truffle.api.impl.DefaultTruffleRuntime \
-Dgraalvm.ForcePolyglotInvalid=true \
-Dgraalvm.locatorDisabled=true \
-Dsubstratevm.IgnoreGraalVersionCheck=true \
--add-exports=java.base/com.sun.crypto.provider=ALL-UNNAMED \
--add-exports=java.base/jdk.internal.event=ALL-UNNAMED \
--add-exports=java.base/jdk.internal.loader=ALL-UNNAMED \
--add-exports=java.base/jdk.internal.logger=ALL-UNNAMED \
--add-exports=java.base/jdk.internal.misc=ALL-UNNAMED \
--add-exports=java.base/jdk.internal.module=ALL-UNNAMED \
[...]

PR #3537, opened 2021-07-01
Support for JAR files with ResourceBundles as .properties at their root

zakkak:
Fixes: gh-3535 
Closes: https://github.com/quarkusio/quarkus/issues/18305

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/jdk11/ClassLoaderSupportImplJDK11OrLater.java
+++ b/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/jdk11/ClassLoaderSupportImplJDK11OrLater.java
@@ -34,7 +34,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.MissingResourceException;
 import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


olpaw:
@zakkak thanks for suggesting a fix. To solve the problem correctly we need a different fix.

What we actually need is:
```patch
diff --git a/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/jdk11/ClassLoaderSupportImplJDK11OrLater.java b/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/jdk11/ClassLoaderSupportImplJDK11OrLater.java
index df9556d6504..ca0012e99af 100644
--- a/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/jdk11/ClassLoaderSupportImplJDK11OrLater.java
+++ b/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/jdk11/ClassLoaderSupportImplJDK11OrLater.java
@@ -80,9 +80,6 @@ public final class ClassLoaderSupportImplJDK11OrLater extends ClassLoaderSupport
             bundleName = specParts[0];
         }
         String packageName = packageName(bundleName);
-        if (packageName == null) {
-            throw new MissingResourceException("ResourceBundle does not seem to be a fully qualified class name.", bundleName, locale.toLanguageTag());
-        }
         Set<Module> modules;
         if (moduleName != null) {
             modules = classLoaderSupport.findModule(moduleName).stream().collect(Collectors.toSet());
@@ -107,8 +104,7 @@ public final class ClassLoaderSupportImplJDK11OrLater extends ClassLoaderSupport
     private static String packageName(String bundleName) {
         int classSep = bundleName.replace('/', '.').lastIndexOf('.');
         if (classSep == -1) {
-            /* The bundle is not specified via a java.class or java.properties format. */
-            return null;
+            return ""; /* unnamed package */
         }
         return bundleName.substring(0, classSep);
     }
```
This will make your reproducer from https://github.com/oracle/graal/issues/3535 work but in addition will be able to correctly open up modules that have bundles in their unnamed package. I will create a PR and add you as a reviewer.

zakkak:
Closing in favor of #3540

PR #3548, opened 2021-07-05
[GR-31114] Remove SubstrateArraysCopyOfSnippets.

graalvmbot:
The SubstrateVM specific intrinsic for Arrays.copyOf is no longer needed
as we can now deal with the graphs the compiler produces. There is no more
good reason to keep maintaining this code.


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/snippets/SubstrateGraphBuilderPlugins.java
```diff
@@ -724,46 +718,6 @@ public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Rec
         });
     }
 
-    private static void registerArraysPlugins(InvocationPlugins plugins) {
-
-        Registration r = new Registration(plugins, Arrays.class).setAllowOverwrite(true);
-
-        r.register2("copyOf", Object[].class, int.class, new InvocationPlugin() {
```
christianwimmer, line 731: For the record: the intrinsification of some Arrays.copyOf methods was part of the (failed) experiments with context sensitive analysis years ago. The slightly better static analysis precision due to the separate node is not worth keeping the complexity around.

christianwimmer (approved):

Issue #2720, opened 2020-07-28
Analysis fails because field of TypeInfoImpl has type incompatible with proxy of XmlSeeAlso


Adelrisk:
**Describe the issue**
While building a native image using quarkus the analysis stage fails with the reason:

> Error: Field com.sun.xml.bind.v2.model.impl.TypeInfoImpl.upstream has declared type com.sun.xml.bind.v2.model.annotation.Locatable which is incompatible with types in state: 1TypeMObject<[0x00005C0F00005C10:!S:XmlSeeAlso$$ProxyImpl]>

In my project, I can only find `com.sun.xml.bind.v2.model.impl.TypeInfoImpl` in the jar `org.glassfish.jaxb:jaxb-runtime:jar:2.3.3-b02:compile` in my dependencies, which is only pulled in within the following top-level dependency:

```
 +- io.quarkus:quarkus-hibernate-orm-panache:jar:1.6.1.Final:compile
 |  +- io.quarkus:quarkus-hibernate-orm:jar:1.6.1.Final:compile
 |  |  +- org.hibernate:hibernate-core:jar:5.4.18.Final:compile
 |  |  |  +- net.bytebuddy:byte-buddy:jar:1.10.10:compile
 |  |  |  +- antlr:antlr:jar:2.7.7:compile
 |  |  |  +- org.dom4j:dom4j:jar:2.1.3:compile
 |  |  |  \- org.hibernate.common:hibernate-commons-annotations:jar:5.1.0.Final:compile
 |  |  +- org.hibernate:hibernate-graalvm:jar:5.4.18.Final:compile
 |  |  +- org.glassfish.jaxb:jaxb-runtime:jar:2.3.3-b02:compile
 |  |  |  +- org.glassfish.jaxb:txw2:jar:2.3.3-b02:compile
 |  |  |  \- com.sun.istack:istack-commons-runtime:jar:3.0.10:compile
 |  |  +- org.jboss.spec.javax.xml.bind:jboss-jaxb-api_2.3_spec:jar:2.0.0.Final:compile
 |  |  +- jakarta.persistence:jakarta.persistence-api:jar:2.2.3:compile
 |  |  +- org.hibernate:quarkus-local-cache:jar:0.1.0:compile
 |  |  |  \- com.github.ben-manes.caffeine:caffeine:jar:2.8.4:compile
 |  |  \- io.quarkus:quarkus-caffeine:jar:1.6.1.Final:compile
[...]

PR #3550, opened 2021-07-05
Support class predefinition via class loaders and handle proxies.

peter-hofer:


substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/AccessAdvisor.java
```diff
@@ -33,6 +35,12 @@
  * {@code AccessVerifier} classes which accesses to ignore when the agent is in restriction mode.
  */
 public final class AccessAdvisor {
+    /**
+     * {@link java.lang.reflect.Proxy} generated classes can be put in arbitrary packages depending
+     * on the visibility and module of the interfaces they implement, so we can only match against
+     * their class name using this pattern (which we hope isn't used by anything else).
+     */
+    public static final Pattern PROXY_CLASS_NAME_PATTERN = Pattern.compile("^(.+[/.])?\\$Proxy[0-9]+$");
```
christianhaeubl, line 43: This will definitely result in false positives. What is the worst that can happen for a false positive?

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -997,6 +998,11 @@ private static void onClassFileLoadHook(@SuppressWarnings("unused") JvmtiEnv jvm
         if (loader.equal(nullHandle())) { // boot class loader
             return;
         }
+        String className = fromCString(name);
+        if (className != null && AccessAdvisor.PROXY_CLASS_NAME_PATTERN.matcher(className).matches()) {
```
christianhaeubl, line 1002: I think it would be good to introduce a helper method for that check to ensure that it is done consistently (you use the same check in a few places).

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/AccessAdvisor.java
peter-hofer: I don't think that it's particularly prone to false positives: the class name has to start with `$` and end with a sequence of digits so this does not match `Proxy` or `Proxy1` or inner classes like `Connection$Proxy1`. I can imagine that other class generators could use a similar naming scheme though. The worst that can happen is that these classes are omitted from the resulting agent configurations.

dufoli:
@peter-hofer can it fix the issue #2720 : 
 Error: Field com.sun.xml.bind.v2.model.impl.TypeInfoImpl.upstream has declared type 

`com.sun.xml.bind.v2.model.annotation.Locatable which is incompatible with types in state: 1TypeMObject<[0x00001E4B00001E4C:!S:XmlSeeAlso$$ProxyImpl]>`

because it seems that proxy do not inherite of one of interface which seems really strange ...
Btw, I do not know where come ProxyImpl because the regexp seems to point to Proxy0, Proxy1, Proxy2....

peter-hofer:
@dufoli I doubt it, it seems your issue lies with the analysis at image build time, which is unrelated to this feature.

dufoli:
@peter-hofer  thanks for answer ;-)

PR #3552, opened 2021-07-07
[GR-32505] Make NativeImageClassLoaderSupport#initAllClasses take jdk.module.upgrade.path into account.

graalvmbot:


olpaw (approved):

dougxc:
Please add a test that builds a native image with `MX_BUILD_EXPLODED=true` to prevent this regressing. Otherwise, looks good.

PR #3555, opened 2021-07-07
(LLVM) GetElementPointer symbol

pichristoph:
Parse, store and use an LLVM symbol which refers to a GetElementPointer instruction

### DIFF
--- a/sulong/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LLVMParser.java
+++ b/sulong/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LLVMParser.java
@@ -51,10 +51,11 @@ import com.oracle.truffle.llvm.runtime.LLVMSymbol;
 import com.oracle.truffle.llvm.runtime.datalayout.DataLayout;
 import com.oracle.truffle.llvm.runtime.except.LLVMLinkerException;
 import com.oracle.truffle.llvm.runtime.global.LLVMGlobal;
-import com.oracle.truffle.llvm.runtime.nodes.memory.LLVMGetElementPtrNode;
+import com.oracle.truffle.llvm.runtime.nodes.api.LLVMExpressionNode;
 import com.oracle.truffle.llvm.runtime.options.SulongEngineOption;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Supplier;
 
 public final class LLVMParser {
     private final Source source;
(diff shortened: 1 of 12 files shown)


oubidar-Abderrahim:
Thank you for contributing to GraalVM, we will take a look into your changes.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/nodes/others/LLVMAccessElemPtrSymbolNode.java
```diff
@@ -0,0 +1,79 @@
...
+    public abstract LLVMPointer executeGeneric(VirtualFrame frame);
+
+    @Override
+    public String toString() {
+        return getShortString("elemPtrSymbol");
+    }
+
+    private LLVMPointer checkNull(LLVMPointer result) {
+        if (result == null) {
+            CompilerDirectives.transferToInterpreter();
```
rschatz, line 62: Please use `transferToInterpreterAndInvalidate`.
`transferToInterpreter` without invalidate is almost always wrong. This will for example prevent the compiler from moving the null check around.

That said, can this even happen? Looking at the implementation of `LLVMgetElementPtrNode`, I don't think this value can ever be null.

So I think this check can be completely removed.

sulong/projects/com.oracle.truffle.llvm.runtime/src/com/oracle/truffle/llvm/runtime/LLVMElemPtrSymbol.java
```diff
@@ -0,0 +1,91 @@
...
+    public LLVMGlobal asGlobalVariable() {
+        throw new IllegalStateException("GetElementPointerConstant " + getName() + " has to be resolved and might not be a global variable.");
+    }
+
+    public final Type getType() {
+        return type;
+    }
+
+    public final LLVMGetElementPtrNode getElementPtrNode() {
+        return elementPtrNode;
```
rschatz, line 78: Storing the node here and just returning it is a problem. This function is used in an `@Cached` specification. That means whenever this is used anywhere in the AST, this node is added as a child node. But nodes can only be added as a child node at a single point in the AST.

What if this symbol is used in more than one place?

I think it would be better to have a `createElementPtrNode` function in this class, that re-creates the GEP node on the fly for every individual user.

sulong/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LLVMParser.java
```diff
@@ -152,17 +155,28 @@ private void defineAlias(String aliasName, boolean isAliasExported, SymbolImpl v
...
+            GetElementPointerConstant elementPointerConstant = (GetElementPointerConstant) value;
+            defineExpressionSymbol(aliasName, isAliasExported, elementPointerConstant, targetDataLayout);
         } else {
             throw new LLVMLinkerException("Unknown alias type: " + value.getClass());
         }
     }
 
+    private void defineExpressionSymbol(String aliasName, boolean isAliasExported, GetElementPointerConstant elementPointerConstant, DataLayout targetDataLayout) {
+        LLVMSymbol baseSymbol = runtime.getFileScope().get(elementPointerConstant.getBasePointer().toString());
+        LLVMGetElementPtrNode elementPtrNode = (LLVMGetElementPtrNode) elementPointerConstant.createNode(runtime, targetDataLayout, GetStackSpaceFactory.createAllocaFactory());
```
rschatz, line 174: (cont from other comment)

... so here, instead of creating the node, I'd create a factory object or a lambda that captures everything it needs and is able to create multiple GEP nodes on the fly.

rschatz (changes requested):
Thanks!

Design lgtm in general!

Just one implementation issue: You're re-using a single node in multiple places in the AST. That won't work. See inline comments.

rschatz:
Also please check the failing gate run. There is an unused import somewhere.

pichristoph:
Thanks for feedback and comments! I hope the changes solve the issues, otherwise just mention it please.

rschatz:
Thanks, looks good now! I'll integrate it.

rschatz (approved):

PR #3340, opened 2021-04-12
Native-image: Add support of iOS simulator (IOS_AMD64)

Noisyfox:
This change is pretty self-explanatory: adds the IOS_AMD64 platform to native-image.
I currently have a working hello world running in ios simulator, the idea is same as https://github.com/gluonhq/substrate, with static libs build for ios simulator.

PR #3579, opened 2021-07-13
Use correct Java system properties for target platform

koutheir:
Use the correct Java system properties to determine target platform operating system and architecture. These are `svm.targetPlatformOS` and `svm.targetPlatformArch`, instead of the old names `svm.targetName` and `svm.targetArch`.

This allows native images to resolve the Java system properties `os.name` and `os.arch` without the need to extract this information from the target system at runtime.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java
@@ -127,14 +127,14 @@ public abstract class SystemPropertiesSupport {
         lazyRuntimeValues.put("java.library.path", this::javaLibraryPath);
         lazyRuntimeValues.put("os.version", this::osVersionValue);
 
-        String targetName = System.getProperty("svm.targetName");
-        if (targetName != null) {
-            initializeProperty("os.name", targetName);
+        String targetOS = System.getProperty("svm.targetPlatformOS");
+        if (targetOS != null) {
+            initializeProperty("os.name", targetOS);
         } else {
             lazyRuntimeValues.put("os.name", this::osNameValue);
         }
 
-        String targetArch = System.getProperty("svm.targetArch");
+        String targetArch = System.getProperty("svm.targetPlatformArch");
         if (targetArch != null) {
             initializeProperty("os.arch", targetArch);
         } else {


lazar-mitrovic:
This change is plain wrong. 
Currently `os.name` is calculated at runtime in order to match Hotspot implementation (calling `uname` or similar construct). Calling `System.getProperty("os.name")` should return name of the current (running) OS, and that is never known in the image build time.

Assuming that `Platform.getName` is equal to `os.name` is ALWAYS wrong on Windows (and it WILL break a lot of other platforms). 

`os.arch` is hard-coded from appropriate `Platform.getArch` implementation. That is correct since it is also hard-coded into JVM.

If you want to specify a custom `os.name` / `os.arch`, you can set `-Dsvm.targetName=<name>` and `-Dsvm.targetArch=<arch>` during `native-image` invocation. 

If you want to support a new `Platform` you should implement a leaf `Platform` interface, and corresponding `SystemPropertiesSupport` class (similar to how it is done in #3340). 

Take a look at #2960, since it is the PR that introduced this part of code.

koutheir:
@lazar-mitrovic, I disagree on a few points:

> Calling System.getProperty("os.name") should return name of the current (running) OS, and that is never known in the image build time.

If `--target=<os>-<arch>` is specified, then the target operating system is actually known at build time.

> Assuming that Platform.getName is equal to os.name is ALWAYS wrong on Windows...

What about assuming that `os.name` equals `svm.targetPlatformOS`? That is actually what this change does.

> os.arch is hard-coded from appropriate Platform.getArch implementation. That is correct since it is also hard-coded into JVM.

I'm using a JVM running on Linux/AMD64 to cross-compile native images that run on Linux/AArch64. The target architechture is clearly not hard-coded in the JVM.

> If you want to specify a custom os.name / os.arch, you can set -Dsvm.targetName=<name> and -Dsvm.targetArch=<arch> during native-image invocation.

Isn't this what the `--target=<os>-<arch>` flag is for?

lazar-mitrovic:
> If `--target=<os>-<arch>` is specified, then the target operating system is actually known at build time.

No it isn't. Especially for Windows, `os.name` has a different value [for each Windows version](https://diarium.usal.es/pmgallardo/2020/10/12/possible-values-of-os-dependent-java-system-properties/). Also we cannot assume things like that since we are aiming to be 100% compatible with HotSpot.

> What about assuming that `os.name` equals `svm.targetPlatformOS`? That is actually what this change does.

`svm.targetPlatformOS` is an internal system property that is generated [when `--target` option is set](https://github.com/lazar-mitrovic/graal/blob/372b87b76f073b7aa60b41979a3a90d80e1b0ba2/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java#L1258-L1286) (reasoning for this is explained in the comment). This in turn selects `Platform` [that has the same `Platform.getOS` and `Platform.getArchitecture`](https://github.com/lazar-mitrovic/graal/blob/372b87b76f073b7aa60b41979a3a90d80e1b0ba2/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGenerator.java#L336-L344), so those two values are functionally identical. 

Take a look at #2960 for more info.

> I'm using a JVM running on Linux/AMD64 to cross-compile native images that run on Linux/AArch64. The target architechture is clearly not hard-coded in the JVM.

I'm talking about the way `os.arch` is implemented in JVM (HotSpot), not about how it is implemented in SubtrateVM. And it is hard-codded in the JVM as you can see [here](https://github.com/openjdk/jdk/blob/ca806ef31970e2d020e631d55b44cfe30e11c5ca/src/java.base/unix/native/libjava/java_props_md.c#L420). If you want to cross-compile to Linux Aarch64 you can pass `--target=linux-aarch64` and it will create `Platform.LINUX_AARCH64` which returns `aarch64` when `getArchitecture` is invoked.

> Isn't this what the `--target=<os>-<arch>` flag is for?
This option is meant to ease selection of appropriate `Platform` interface. Resulting `Platform` and `LibC` interfaces are then used internally to determine further automatic configuration.

Again, #2960 is the big one but changing stuff without understanding what this PR does WILL break things.

koutheir:
> No it isn't. Especially for Windows, os.name has a different value for each Windows version.

I understand, but I would like to notice that this issue seems to affect only Windows.

So, if I understand well, the `--target=<os>-<arch>` flag sets `svm.targetPlatformOS` and `svm.targetPlatformArch` which are used to select the compilation target, but the actual values reported by `os.name` and `os.arch` are, respectively, `svm.targetName` and `svm.targetArch` when they are specified, and runtime values otherwise. Is this overview correct?

lazar-mitrovic:
> I understand, but I would like to notice that this issue seems to affect only Windows.

Well in most cases yes. For example if someone uses a Linux system which returns a string that isn't `Linux` when [`uname` is invoked](https://github.com/oracle/graal/blob/3560abc2d365eb7166fa5bc0c0033e4e3fce5746/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSystemPropertiesSupport.java#L71-L78) all assumptions would break. 

Also the cost of calculating this value in runtime for linux is literally one `uname` syscall per application run (as we are [using LazyRuntimeValues](https://github.com/oracle/graal/blob/3560abc2d365eb7166fa5bc0c0033e4e3fce5746/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/SystemPropertiesSupport.java#L73-L135) to determine this value only when needed and cache results). So there is really no point in optimizing this further.

> So, if I understand well, the `--target=<os>-<arch>` flag sets `svm.targetPlatformOS` and `svm.targetPlatformArch` which are used to select the compilation target, but the actual values reported by `os.name` and `os.arch` are, respectively, `svm.targetName` and `svm.targetArch` when they are specified, and runtime values otherwise. Is this overview correct?

That is correct. Original idea was to remove `svm.targetName` and `svm.targetArch` completely and force users to use corresponding `Platform` implementations, however I decided against it, since it would break a lot of downstream projects (and we were nearing feature freeze).

PR #3580, opened 2021-07-13
Properly cast pointers to integers and vice-versa

koutheir:
Properly casting pointers to integers (and vice-versa) needs casting to `intptr_t` as an intermediary type.

### DIFF
--- a/truffle/src/com.oracle.truffle.nfi.native/src/api.c
+++ b/truffle/src/com.oracle.truffle.nfi.native/src/api.c
@@ -79,21 +79,21 @@ static void newClosureRef(TruffleEnv *tenv, void *closure) {
     struct __TruffleEnvInternal *ienv = (struct __TruffleEnvInternal *) tenv;
     struct __TruffleContextInternal *context = ienv->context;
     JNIEnv *env = ienv->jniEnv;
-    (*env)->CallVoidMethod(env, context->LibFFIContext, context->LibFFIContext_newClosureRef, (jlong) closure);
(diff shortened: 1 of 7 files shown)


rschatz (approved):
Lgtm, thanks! I'll integrate it.

Out of curiosity: Did you also try running it? Or just building?

I did schedule tests already, and it passes on all architectures we test internally, but they are all 64-bit.

This certainly makes everything build on 32-bit architectures. Does it actually work, too? I mean, at least without JIT compilation, it should work in theory, since libffi supports it, and the Truffle NFI gets all its size/offset/alignment information from libffi. But my usual experience is that everything without tests is probably broken ;)

koutheir:
> Out of curiosity: Did you also try running it? Or just building?

I only built it on a 32-bits architecture.

> This certainly makes everything build on 32-bit architectures. Does it actually work, too?

I didn't test Truffle features. I'm only interested in building native images of Java applications.

> But my usual experience is that everything without tests is probably broken ;)

My experience confirms yours. A lot of code in GraalVM assumes running on a machine with 64-bits general purpose registers, and hard-coded values and algorithms are programmed with that assumption.

rschatz:
For completeness: I did try it out on a 32-bit chroot. It builds fine, but doesn't actually work. It's failing 110 of 404 Truffle NFI unittests, failing various assertions. No segfaults though, looks like the assumptions on pointer size are at least properly guarded with assertions in the code ;)

But definitely a step in the right direction, thanks a lot!

PR #3582, opened 2021-07-14
[GR-28460] Introduce GraphBuilderContext.recursiveInliningDepth.

graalvmbot:


compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/graphbuilderconf/GraphBuilderContext.java
```diff
@@ -259,6 +259,20 @@ default int getDepth() {
         return result;
     }
 
+    /**
+     * Computes the recursive inlining depth of the provided method, i.e., counts how often the
+     * provided method is already in the {@link #getParent() parent chain} starting this context.
```
dougxc, line 264: Use `@linkplain` instead when you have anchor text.
Also "starting this context" -> "of this context".

dougxc (approved):

PR #3597, opened 2021-07-16
Added support to register/access resource with trailing slash.

jovanstevanovic:
**Problem**
In most cases, directories as a resource are registered with a trailing slash, and currently, it's not supported. Similarly, when the directories as a resource are accessed with a trailing slash, they won't be found because of the same reason. This PR tries to fix both problems.

**Linked issue** 
https://github.com/micronaut-projects/micronaut-core/issues/5765

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
@@ -144,7 +144,7 @@ public final class Resources {
         if (resourceName == null) {
             return null;
         }
-        Enumeration<URL> urls = createURLs(resourceName);
+        Enumeration<URL> urls = createURLs(removeTrailingSlash(resourceName));
         return urls.hasMoreElements() ? urls.nextElement() : null;
     }
 
@@ -153,7 +153,7 @@ public final class Resources {
         if (resourceName == null) {
             return null;
         }
-        ResourceStorageEntry entry = Resources.get(resourceName);
+        ResourceStorageEntry entry = Resources.get(removeTrailingSlash(resourceName));
         if (entry == null) {
             return null;
         }
@@ -165,7 +165,7 @@ public final class Resources {
         if (resourceName == null) {
             return null;
         }
-        ResourceStorageEntry entry = Resources.get(resourceName);
+        ResourceStorageEntry entry = Resources.get(removeTrailingSlash(resourceName));
         if (entry == null) {
             return Collections.emptyEnumeration();
         }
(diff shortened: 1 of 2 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
```diff
@@ -324,14 +324,15 @@ private static void scanJar(DebugContext debugContext, File root, Pattern[] incl
     }
 
     private static boolean matches(Pattern[] includePatterns, Pattern[] excludePatterns, String relativePath) {
+        String relativePathWithTrailingSlash = relativePath + "/";
```
vjovanov, line 327: What if we have 2 `/`s?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
```diff
@@ -176,6 +176,10 @@ public static InputStream createInputStream(String resourceName) {
         }
         return Collections.enumeration(resourcesURLs);
     }
+
+    private static String removeTrailingSlash(String resourceName) {
+        return !resourceName.endsWith("/") ? resourceName : resourceName.substring(0, resourceName.length() - 1);
```
zakkak, line 181: Would that work with Windows paths, i.e., path ending with `\`?

jovanstevanovic: "Unfortunately" yes, because Java is mixing separators on Windows. `File.separator` (system dependent variable) will return `'\'` on Windows, but when it comes to resources Java is using  `'/'` (see comment here https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java#L262-L266 )

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java
jovanstevanovic: If we call this method from `scanDirectory`, the relative path is constructed here https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java#L263-L268, and it will not have an extra slash at the end. If we call this method from `scanJar` the relative path is constructed here https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ResourcesFeature.java#L309-L309, and also there will be no extra slash. So in both variants, we are not accumulating slashes at the end.

zakkak: I think adding an assertion for that would be great maintenance-wise.  

Something like the following maybe:
```
assert Pattern.compile(".*\\w$").matcher(relativePath).matches();
```

jovanstevanovic: Sounds good. I'll add a check here.

zakkak (approved):
LGTM, Thanks!

vjovanov (approved):

PR #3611, opened 2021-07-20
Fix agent tracing for null values inside arrays

ivan-ristovic:
When using `trace-output` option, the agent currently does not handle `null` values inside arrays properly, i.e. a `NullPointerException` will be thrown from [TraceFileWriter.printValue](https://github.com/oracle/graal/blob/0bdddb159b77133ceb9ff1b6ff0e2d69531ce70f/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/tracing/TraceFileWriter.java#L103) in case `value` is `null`:
https://github.com/oracle/graal/blob/0bdddb159b77133ceb9ff1b6ff0e2d69531ce70f/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/tracing/TraceFileWriter.java#L103

This PR resolves this issue by replacing the `value.toString()` with `Objects.toString(value)` which properly handles `null` values.

### DIFF
--- a/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/tracing/TraceFileWriter.java
+++ b/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/tracing/TraceFileWriter.java
@@ -32,6 +32,7 @@ import java.nio.file.Path;
 import java.util.Base64;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 
 import com.oracle.svm.agent.tracing.core.Tracer;
 import com.oracle.svm.agent.tracing.core.TracingResultWriter;
@@ -100,7 +101,7 @@ public class TraceFileWriter extends Tracer implements TracingResultWriter {
         if (value instanceof byte[]) {
             s = Base64.getEncoder().encodeToString((byte[]) value);
         } else if (value != null) {
-            s = value.toString();
+            s = Objects.toString(value);
         }
         json.quote(s);
     }


substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/tracing/TraceFileWriter.java
```diff
@@ -100,7 +101,7 @@ private static void printValue(JsonWriter json, Object value) throws IOException
         if (value instanceof byte[]) {
             s = Base64.getEncoder().encodeToString((byte[]) value);
         } else if (value != null) {
-            s = value.toString();
+            s = Objects.toString(value);
```
peter-hofer, line 104: I don't quite understand what the problem is. Before I recently added the `value != null` check above, `NullPointerException` could be thrown here, but now `null` will be passed to `JsonWriter.quote` which prints a JSON `null`. `Objects.toString` would return the string `"null"`, which is worse because it might be a legitimate value.

ivan-ristovic: Originally I observed this on a branch that was based on commit d8d914648604df9fa83c13aab35c81a37b61c027, and after rebasing to master I somehow failed to notice that this case is now handled. Thank you for a quick response!

ivan-ristovic:
As @peter-hofer pointed out, this PR is obsolete.

PR #3612, opened 2021-07-20
[GR-32715] Update JVMCI version check for reserved fields to include OpenJDK 11 update 13.

graalvmbot:


jerboaa (commented):
This should only get merged once it's definitive that the fields will be available in that OpenJDK version. It's still in candidate at this point. See:
https://bugs.openjdk.java.net/browse/JDK-8269877

jerboaa:
@chumer This PR needs to get rebased and can then go in. Relevant OpenJDK 11 fixes are going to be in 11.0.13 (+4 should be the first build with the fixes).

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java
```diff
@@ -398,14 +398,18 @@ private int getJvmciJvmtiCapabilityOffset(String name) {
     {
         int offset;
         try {
-            offset = getFieldOffset("JavaThread::_jvmci_reserved0", Integer.class, "jlong", -1, JVMCI ? jvmciGE(JVMCI_21_1_b02) : JDK >= 17);
+            offset = getFieldOffset("JavaThread::_jvmci_reserved0", Integer.class, "jlong", -1, hasJVMCIReservedFields());
```
jerboaa, line 401: This depends on https://bugs.openjdk.java.net/browse/JDK-8272736 being in OpenJDK 11, which it currently isn't. Just FYI. It should do no harm, though, as it would fall back to using the `intptr_t*` type in the catch block.

compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/GraalHotSpotVMConfig.java
```diff
@@ -800,7 +804,8 @@ public int threadTlabTopOffset() {
     public final long newMultiArrayOrNullAddress = getAddress("JVMCIRuntime::new_multi_array_or_null", 0L, JVMCI || JDK >= 12 || (!IS_OPENJDK && JDK == 11 && JDK_UPDATE >= 7));
     public final long dynamicNewInstanceOrNullAddress = getAddress("JVMCIRuntime::dynamic_new_instance_or_null", 0L, JVMCI || JDK >= 12 || (!IS_OPENJDK && JDK == 11 && JDK_UPDATE >= 7));
 
-    public final long invokeJavaMethodAddress = getAddress("JVMCIRuntime::invoke_static_method_one_arg", 0L, JVMCI ? jvmciGE(JVMCI_21_1_b02) : JDK >= 17);
+    public final long invokeJavaMethodAddress = getAddress("JVMCIRuntime::invoke_static_method_one_arg", 0L,
+                    JVMCI ? jvmciGE(JVMCI_21_1_b02) : (JDK >= 17 || (IS_OPENJDK && JDK == 11 && JDK_UPDATE >= 13)));
```
jerboaa, line 808: Any particular reason why `hasJVMCIReservedFields()` isn't used here? The condition is the same, though, semantically the method name doesn't necessarily make sense here ;-)

PR #3614, opened 2021-07-21
More JDK classes require runtime initialization

koutheir:
More JDK classes require runtime initialization:

- `sun.net.www.protocol.http.ntlm.NTLMAuthentication`: Calls `NTLMAuthentication.isTrustedSiteAvailable()` on Windows.
- `sun.net.PortConfig`: Calls `PortConfig.getLower0()` and `PortConfig.getUpper0()`.
- `sun.nio.ch.DevPollArrayWrapper`: Calls `IOUtil.fdLimit()`.
- `sun.nio.ch.EPoll`: Calls `EPoll.eventSize()`, `EPoll.eventsOffset()` and `EPoll.dataOffset()`.
- `sun.nio.ch.EPollSelectorImpl`: Calls `IOUtil.fdLimit()`.
- `sun.nio.ch.EventPortSelectorImpl`: Calls `IOUtil.fdLimit()`.
- `sun.nio.fs.LinuxWatchService.Poller`: `LinuxWatchService.eventSize()` and `LinuxWatchService.eventOffsets()`.
- `sun.awt.shell.Win32ShellFolder2.KnownLibraries`: Calls `Win32ShellFolder2.loadKnownFolders()`.


koutheir:
The build CI error is:
```
2021-07-23T00:10:39.4610027Z Error: Classes that should be initialized at run time got initialized during image building:
2021-07-23T00:10:39.4613039Z  sun.net.PortConfig the class was requested to be initialized at run time (Calls PortConfig.getLower0() and PortConfig.getUpper0()). To see why sun.net.PortConfig got initialized use --trace-class-initialization=sun.net.PortConfig
```

Given that I cannot change the build CI configuration to include `--trace-class-initialization=sun.net.PortConfig`, can someone help with this?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKInitializationFeature.java
```diff
@@ -87,6 +96,8 @@ public void afterRegistration(AfterRegistrationAccess access) {
         rci.initializeAtBuildTime("java.awt.font.TextAttribute", "Required for sun.text.bidi.BidiBase.NumericShapings");
         rci.initializeAtBuildTime("java.awt.font.NumericShaper", "Required for sun.text.bidi.BidiBase.NumericShapings");
 
+        rci.initializeAtRunTime("sun.awt.shell.Win32ShellFolder2$KnownLibraries", "Calls Win32ShellFolder2.loadKnownFolders()");
```
vjovanov, line 99: Hmm, who initialized this class in the first place?

koutheir: I don't know who does that, but I know that class must not be initialized at build time. I didn't find these classes by debugging, but rather through reading of the OpenJDK code.

koutheir: @vjovanov Should I resolve this conversation? Or do you need further analysis?

vjovanov: This is not our job: Only the classes that we initialize we should also make exceptions for. That is why we can't put this line until we know who initialized it.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKInitializationFeature.java
```diff
@@ -75,8 +75,17 @@ public void afterRegistration(AfterRegistrationAccess access) {
         rci.initializeAtBuildTime("sun.nio", "Core JDK classes are initialized at build time");
         if (Platform.includedIn(Platform.WINDOWS.class)) {
             rci.rerunInitialization("sun.nio.ch.PipeImpl", "Contains SecureRandom reference, therefore can't be included in the image heap");
+            rci.initializeAtRunTime("sun.net.www.protocol.http.ntlm.NTLMAuthentication", "Calls NTLMAuthentication.isTrustedSiteAvailable() on Windows");
         }
 
+        rci.initializeAtRunTime("sun.net.PortConfig", "Calls PortConfig.getLower0() and PortConfig.getUpper0()");
+
+        rci.initializeAtRunTime("sun.nio.ch.DevPollArrayWrapper", "Calls IOUtil.fdLimit()");
```
vjovanov, line 83: I would also group them together with the place where we initialize at build time.

vjovanov:
Other than that LGTM.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKInitializationFeature.java
koutheir: Does that means that a class initialized by an application should not be marked as "initializeAtRunTime"?

If so, then how is an application developer supposed to know that their application led to initialization of `sun.awt.shell.Win32ShellFolder2$KnownLibraries`, for instance?

And is the application developer supposed to dig into the OpenJDK sources to figure out that they need to initialize this class at runtime?

Just to clarify, my intention here is to mark classes that, if needed by the application, need to be initialized at runtime. I also want to notice that I'm not using `rerun...()` functions because I could not find where their semantics are defined.

vjovanov: The policy is, if we initialize somethig at build-time we need to address all the dependencies. If someone else does, then it is their job. That is why I ask, who initialized this class.

koutheir: > The policy is, if we initialize somethig at build-time we need to address all the dependencies. If someone else does, then it is their job.

So if a developer builds a native image of their application, and somehow causes `sun.awt.shell.Win32ShellFolder2$KnownLibraries` to be initialized at build time (because that is **the default**), then the developer needs to figure this out on their own?! This is an unrealistic expectation, because a Java developer should **not** need to analyze the Java standard library sources or disassemble the generated native image in order to figure out this bug.

Standard library classes that should never be initialized at build time are **known** to the developers of the Java standard library, and they should be marked so. The policy you're mentioning does not make sense in the context of a Java developer generating a native image of their application.

> That is why I ask, who initialized this class.

Again, no one did, and no one should initialize these classes at build time, ever.

vjovanov: Yes they should, build-time initialization is a complex and dangerous feature; it should be used with caution. If one uses it, they should take full responsibility for it and initialize everything properly. 

By analogy, hundreds of classes in the JDK should be marked as run-time initialized. It is a great burden to mark all those classes, especially when there are changes in the JDK  between versions.

koutheir: > If one uses it...

How can one disable it, so that runtime initialization is the default? Are there any flags to control this behavior?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKInitializationFeature.java
koutheir: The way this file appears to be organized is with `initializeAtBuildTime()` calls grouped, and other calls grouped. I simply tried to keep this choice.

koutheir:
I removed the classes `sun.net.www.protocol.http.ntlm.NTLMAuthentication` and `sun.awt.shell.Win32ShellFolder2.KnownLibraries` from forced runtime initialization, even though they should never be initialized at build time.

@vjovanov, can you please review this patch again?

vjovanov (approved):
LGTM

vjovanov:
Starting the merge now. Please don't make the PR from master in the future as it makes it harder to make a branch in our CI.

koutheir:
> Please don't make the PR from master in the future as it makes it harder to make a branch in our CI.

From where should I make future pull requests?

vjovanov:
From a branch in your repo. 

Also, had to change calls in this PR to `rerun` as a ton of tests were failing.

koutheir:
I didn't use `rerun` methods because I didn't find any description of their behavior/semantics.

PR #3615, opened 2021-07-22
Added name mapping, added experimental option to forcefully enable multithreading (default off)

TudbuT:
- Added name mapping of fields, methods, and classes at runtime (useful for running scripts off of obfuscated host applications)
- Added experimental option to forcefully enable multithreading (this is to allow devs more freedom, multithreading CAN be done safely, so blocking it completely makes no sense to me)

Feel free to suggest changes before merging!


TudbuT:
I registered and agreed now.

TudbuT:
Conflicts have been resolved.

TudbuT:
Yes, I have already signed it. The problem here is a github bug that my Access token's commits use a different Email than my Github account for some reason. Both of that is me, regardless of my two Email-adresses. Please confirm to the checker that I have signed.

chumer:
@TudbuT Hi Daniel, thanks for your contribution.

>  multithreading CAN be done safely, so blocking it completely makes no sense to me

Unfortunately we won't be able to accept forcefully enabling multi-threading as is. We have written about the reasons for this in: https://medium.com/graalvm/multi-threaded-java-javascript-language-interoperability-in-graalvm-2f19c1f9c37b
But TLDR: Multi-threading semantics are not specified for JavaScript and some other languages. Just imposing the Java memory model for that exposes many implementation details, that we cannot maintain or ensure compatibility with in the future. The blog article discusses the reasons and possible workarounds in more detail.

Do you want to elaborate what you understand with doing multithreading safely for JavaScript? Why do you think your accesses are safe? We support accesses from multiple threads, but just not at the same time. Maybe you can motivate it with a small example? 

Re.: `RuntimeNameMapper`. Could you please start by filing a Github issue with some details of the use-cases you intend to use this for? Why do existing features not work for this? E.g. why can't you declare a custom HostClassLoader with Context.Buidler.hostClassLoader(ClassLoader)? If something is obfuscated it is likely that this was done intentional by the authors either for licensing reasons or API compatibility reasons, what motivates you to deobfuscate the code?
The implementation also has several problems, but I suggest we discuss the use-cases first.

TudbuT:
> If something is obfuscated it is likely that this was done intentional by the authors either for licensing reasons or API compatibility reasons, what motivates you to deobfuscate the code
I know. In my case, an application isn't intended to be modified or decompiled (fields and methods also obfuscated, so classloaders wont help much), but plugins are intended - and making plugins for an obfuscated application is quite tedious.

chumer:
> > If something is obfuscated it is likely that this was done intentional by the authors either for licensing reasons or API compatibility reasons, what motivates you to deobfuscate the code
> 
> I know. In my case, an application isn't intended to be modified or decompiled (fields and methods also obfuscated, so classloaders wont help much), but plugins are intended - and making plugins for an obfuscated application is quite tedious.

Could you please create the requested Github issue with ideally with a runnable example that motivates this?
Did you consider wrapping the Java host values into your own types that are non obfuscated?

TudbuT:
> Could you please create the requested Github issue ideally with a runnable example that motivates this?

Using my fork, i have already done that at TudbuT/polyfire - It's a minecraft mod that allows for custom modules to be added to the game

> Did you consider wrapping the Java host values into your own types that are non obfuscated?

In an application the size of minecraft's codebase, that's not possible outside of runtime because i dont want my mod to be 100MB big

TudbuT:
Another point about why i think multithreading should be an option: Even if it is extremely broken and crashes regularly, giving the developers an option to enable it, and maybe display a warning in console, gives the devs the ability to, depending on their situation, use multithreading despite its risk. If the dev knows the risk but also knows how to evade that happening, it will be just fine. If you dont want it because ECMAscript doesnt contain it, then a polyglot engine would be just as forbidden. This functionality would, in most cases, extend the functions of javascript, and not limit them. 

TL;DR Regardless of if its broken, let devs choose and let them solve any bugs it causes themselves if they choose to enable it. Overall, through making it optional, there is simply no way for any disadvantages to be.

chumer:
> TL;DR Regardless of if its broken, let devs choose and let them solve any bugs it causes themselves if they choose to enable it. Overall, through making it optional, there is simply no way for any disadvantages to be.

There is just no way you can use JS from multiple threads at the same time that is correct. Nashorn had a blog post explaining what needs to do so it is correct. I haven' seen a single use-case where users did the correct thing. We expect things that we ship to work. We are not willing to support half-baked unsafe solutions. Nobody stops you from forking Graal.js and enable multi-threading. But you are on your own if you do that (including maintaining all the bugs). I recommend you to use one of the solutions we suggest in the blog post.

> Using my fork, i have already done that at TudbuT/polyfire - It's a minecraft mod that allows for custom modules to be added to the game. Here's the RuntimeNameMapper implementation: https://github.com/TudbuT/polyfire/blob/master/src/main/java/tudbut/mod/polyfire/utils/JSFieldMapper.java

I already feared that this is about Minecraft. The license situation there is at least interesting. Anyway, the use-case is reasonably valid, so we can think of a design for this. 

Here are a few notes of what needs to be done to continue:

* Then name and location of the mapper can be improved. I also suggest a different name NameMapper.
* Documentation in general needs to be improved for the mapper class. Make sure it is clear which format of the names is used (binary regular class name format). Instead of getClass, I suggest mapClassName and mapInnerClassName. Instead of getField, mapFieldName, and instead of getMethod, mapMethodName. 
* The mapper needs to be configurable as part of the HostAccess class. I suggest put it as inner class of HostAccess.
* The specification misses a description on how often any method of name mapper is called. I think best would be to guarantee its called once per class and method/field. 
* We should cache the looked up names in HostClassDesc so repeated lookups don't lead to different names. Which would be pretty complex for host method overload resolution. There you can also access the HostAccess instance.
* What if two methods with different signature resolve to the same name, should they be treated as overloads? I'd say probably yes. This needs to be well tested.
* There needs to be test coverage. We typically aim to almost 100% for embedder features, as we guarantee long term support for them. Please make sure you test errors in the implementation as well.
* You need to regenerate the API signatures
* You need to update the SDK changelog.md file to include the change.
* Add @since tags with the version number to all the public APi.
* The regular gate tests need to pass. Run `mx gate` in `truffle` and `compiler` suite to verify.

TudbuT:
FIRST OFF: If you insist on not adding multithreading, thats fine, but regardless of that, it makes no sense to me personally.
I am not trying to be annoying. I just strongly believe in my opinions and try to communicate my reasons here.

---

Let's look at this example, which is what i think you were probably talking about: https://blogs.oracle.com/nashorn/nashorn-multithreading-and-mt-safety

The reason the test in the article fails so miserably is that the running code is using globals, and is using a random delay between the operations. Java will behave the same and it does not demonstrate any breaking behaviour. It can be solved by not modifying a global variable but a local one or not adding the delay. There may also be other synchronization issues (like two threads modifying a single object at the same time, resulting in one copy being discarded), but most languages have that problem anyway. Synchronizing those modifications isnt gonna be hard either.

The reason the second example works is that it runs the script with a new global, effectively separating it into a new context.

Even you guys have said that MT is important in java in that article:
> On the other hand, having a language on the JVM and not supporting multithreading is just plain daft.  Especially, since our target platforms include multi-core servers.

> Developers will naturally be drawn to using threads.  Many of the Java APIs require use of threads.  So, the best we can do is provide guidelines on how to not shoot yourself in the foot

---

Regarding the improvements on the NameMapper: I can do that, just give a bit.

chumer:
> There may also be other synchronization issues (like two threads modifying a single object at the same time, resulting in one copy being discarded), but most languages have that problem anyway. 

Let me explain a bit further: Synchronization issues don't just lead to discarded copies. They can lead to fatal errors and worse memory corruption (so the actual bug appears some time later). Then you will file an issue for the crash that we didn't want to fix in the first place and where we need to make choices how this behaves. Even worse, we don't know when we will actually have fixed the problem, because there are an almost infinite number of possible bugs with race conditions and rich JavaScript semantics. We want to ensure that whatever we put out there we evolve in a reasonably compatible way and it works without internal errors. Exposing multi-threading without checks, besides fixing an infinite number of bugs, locks us in forever on the unspecified behavior. Sometimes you need to protect the users from themselves. I understand that this decision is unpopular. btw.: Even basic function calls are not safe if the global object is not frozen. 

This does not mean that some kind of safe multi-threading can be exposed and features can be developed. For example it would be reasonable to wrap accesses to JavaScript objects and automatically lock on accesses. Downside of this approach is deadlocks from lock ordering. It is also reasonable to temporarily release (aka leave) a context from a thread in a Java host call.

> Regarding the improvements on the NameMapper: I can do that, just give a bit.

sure, take your time

Issue #3620, opened 2021-07-23
Compiler is failing to track source position for NewInstance nodes


adinn:
The current GraalVM head is manifesting a problem with tracking source positions for compiled Java methods that affects debugging in native images.

During compilation a NewInstance node is often replaced by a VirtualInstance node.The source position (NodeSourcePosition) associated with the NewInstance is not copied across to the VirtualInstance. As a consquence the caller information in the SourceMapping for the compiled method that relates to the inlined allocation instructions ends in a root call that has no associated file or line number.

Manifesting this problem without the latest in progress debuginfo update is complicated. It can only be done by using a debugger to inspect the SourceMapping of a CompilationResult for a method that includes a new operation. The mapping will include one or more NodeSourcePosition objects for the inlined allocation. These will have caller chains that have valid inline positions for methods like GenScavengeAllocationSnippets::readTlabTop, AllocationSnippets.allocateInstanceImpl etc but the inlined caller chains all end with a root NodeSourcePosiiton for the top level method that is marked as a placeholder and has bci = -1. This root called should actually identify the bci for the new operation in the outer compiled method.

The latest debug info code manifests the problem more immediately. At the point where a new is inlinedin the Hello debugger test the code produces this backtrace:

    #0  com.oracle.svm.core.genscavenge.graal.GenScavengeAllocationSnippets::readTlabTop () at com/oracle/svm/core/genscavenge/graal/GenScavengeAllocationSnippets.java:129
    #1  org.graalvm.compiler.replacements.AllocationSnippets::allocateInstanceImpl () at org/graalvm/compiler/replacements/AllocationSnippets.java:59
    #2  com.oracle.svm.core.graal.snippets.SubstrateAllocationSnippets::allocateInstance () at com/oracle/svm/core/graal/snippets/SubstrateAllocationSnippets.java:124
    #3  0x0000000000405087 in Hello$Greeter::greeter(java.lang.String[] *) ()
    #4  0x0000000000405593 in Hello::main(java.lang.String[] *) () at Hello.java:43
    #5  0x000000000041b9e5 in com.oracle.svm.core.JavaMainWrapper::runCore () at com/oracle/svm/core/JavaMainWrapper.java:146
    #6  com.oracle.svm.core.JavaMainWrapper::run () at com/oracle/svm/core/JavaMainWrapper.java:182
[...]

PR #3621, opened 2021-07-23
ensure source position for new instance node is tracked

adinn:
This is proposed as a simple fix for the problem reported in issue #3620.

### DIFF
--- a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/virtual/CommitAllocationNode.java
+++ b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/virtual/CommitAllocationNode.java
@@ -141,7 +141,8 @@ public final class CommitAllocationNode extends FixedWithNextNode implements Vir
         for (int i = 0; i < virtualObjects.size(); i++) {
             VirtualObjectNode virtualObject = virtualObjects.get(i);
             int entryCount = virtualObject.entryCount();
-            tool.createVirtualObject(virtualObject, values.subList(pos, pos + entryCount).toArray(new ValueNode[entryCount]), getLocks(i), ensureVirtual.get(i));
+            /* n.b. the node source position of virtualObject will have been set when it was created */
+            tool.createVirtualObject(virtualObject, values.subList(pos, pos + entryCount).toArray(new ValueNode[entryCount]), getLocks(i), null, ensureVirtual.get(i));
             pos += entryCount;
         }
         tool.delete();
--- a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/virtual/CommitAllocationNode.java
+++ b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/virtual/CommitAllocationNode.java
@@ -141,8 +141,11 @@ public final class CommitAllocationNode extends FixedWithNextNode implements Vir
         for (int i = 0; i < virtualObjects.size(); i++) {
             VirtualObjectNode virtualObject = virtualObjects.get(i);
             int entryCount = virtualObject.entryCount();
-            /* n.b. the node source position of virtualObject will have been set when it was created */
-            tool.createVirtualObject(virtualObject, values.subList(pos, pos + entryCount).toArray(new ValueNode[entryCount]), getLocks(i), null, ensureVirtual.get(i));
+            /*
+             * n.b. the node source position of virtualObject will have been set when it was
+             * created.
+             */
+            tool.createVirtualObject(virtualObject, values.subList(pos, pos + entryCount).toArray(new ValueNode[entryCount]), getLocks(i), virtualObject.getNodeSourcePosition(), ensureVirtual.get(i));
             pos += entryCount;
         }
         tool.delete();
--- a/compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/nodes/frame/NewFrameNode.java
+++ b/compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/nodes/frame/NewFrameNode.java
@@ -37,6 +37,7 @@ import org.graalvm.compiler.graph.IterableNodeType;
 import org.graalvm.compiler.graph.Node;
 import org.graalvm.compiler.graph.NodeClass;
 import org.graalvm.compiler.graph.NodeInputList;
+import org.graalvm.compiler.graph.NodeSourcePosition;
 import org.graalvm.compiler.nodes.spi.Canonicalizable;
 import org.graalvm.compiler.nodes.spi.CanonicalizerTool;
 import org.graalvm.compiler.nodeinfo.NodeInfo;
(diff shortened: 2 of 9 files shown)


tkrodriguez:
Why not apply a similar fix to other places where `VirtualnstanceNode` and `VirtualArrayNode` are allocated?  All the ones I can see have the same problem.  I don't think changing `VirtualizerTool.replaceWithVirtual` to set the position is the right fix since that is sometimes used virtual objects that already exist.

adinn:
@tkrodriguez Thanks for looking at this.

>  I don't think changing VirtualizerTool.replaceWithVirtual to set the position is the right fix since that is sometimes used virtual objects that already exist.

In that case would we not already have a source node position on the virtual object? i.e. can we not finesse this by setting the source node position only if the current one is either null or a place holder.

I'm assuming that this case only happens when the 'same' virtual object can be reused for multiple virtualizations? If so then leaving the original source node position would be a way to record something useful about the provenance of this 'same' object at all points of use even if it related that provenance to a source location where an original version of that object was created rather than some later source point where an equivalent copy would have been created.

tkrodriguez:
Yes `replaceWithVirtual` could only set the value if there wasn't already a value in place.  I'd kind of prefer a strategy that forces the creator of `VirtualObjectNode` to think about it though.  We could require a source position argument to the constructor which forces all users to think about it.  Or maybe just requiring that argument to `VirtualizerTool.createVirtualObject` would be less intrusive.  Anyway, something more systematic but not too onerous would be better and I'm not overly picky about the particular.

adinn:
@tkrodriguez Hi Tom.

I just pushed a new fix where I modified createVirtualObject to accept a source position and,if non-null, set it on the VirtualNode (in which case it asserts a position is not already set). I changed *all* callers to pass what I think is the appropriate source position. There were 2 special cases:

 - I had to modify BoxNode to rely on createVirtualObject rather than do the set locally.

 - CommitAllocationNode passes null when processing it's virtualObjectList because the position is already set by the constructor for AllocationNode.

Well, there is a third case, actually. I'm not 100% clear I have done the right thing in NewFrameNode. I'm assuming the NewFrameNode has a source position at the point where its virtual references are processed by calling createVirtualObject and propagating the NewFrameNode position to them. I don't really know enough about Truffle to know if this is appropriate or not.

adinn:
One more tricky case corrected -- ObjectClone should not try to reset the source position when it duplicates an existing virtual node.

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/virtual/CommitAllocationNode.java
```diff
@@ -141,7 +141,8 @@ public void virtualize(VirtualizerTool tool) {
         for (int i = 0; i < virtualObjects.size(); i++) {
             VirtualObjectNode virtualObject = virtualObjects.get(i);
             int entryCount = virtualObject.entryCount();
-            tool.createVirtualObject(virtualObject, values.subList(pos, pos + entryCount).toArray(new ValueNode[entryCount]), getLocks(i), ensureVirtual.get(i));
+            /* n.b. the node source position of virtualObject will have been set when it was created */
+            tool.createVirtualObject(virtualObject, values.subList(pos, pos + entryCount).toArray(new ValueNode[entryCount]), getLocks(i), null, ensureVirtual.get(i));
```
tkrodriguez, line 145: You could pass `virtualObject.getNodeSourcePosition()` instead of null to indicate where you except the value to come from.

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/nodes/frame/NewFrameNode.java
```diff
@@ -268,12 +269,13 @@ public void virtualize(VirtualizerTool tool) {
             }
         }
 
-        tool.createVirtualObject(virtualFrameObjectArray, objectArrayEntryState, Collections.<MonitorIdNode> emptyList(), false);
+        NodeSourcePosition sourcePosition = getNodeSourcePosition();
```
tkrodriguez, line 272: This seems like as good an answer as we can provide.

compiler/src/org.graalvm.compiler.virtual/src/org/graalvm/compiler/virtual/phases/ea/VirtualizerToolImpl.java
```diff
@@ -309,6 +310,10 @@ public void createVirtualObject(VirtualObjectNode virtualObject, ValueNode[] ent
         closure.addVirtualAlias(virtualObject, virtualObject);
         PartialEscapeClosure.COUNTER_ALLOCATION_REMOVED.increment(debug);
         effects.addVirtualizationDelta(1);
+        if (sourcePosition != null) {
+            assert virtualObject.getNodeSourcePosition() == null : "source pos already set!";
```
tkrodriguez, line 314: We might add `|| sourcePosition == virtualObject.getNodeSourcePosition()` if we decide to pass `virtualObject.getNodeSourcePosition()` instead of null.

tkrodriguez (approved):
What's here looks good to me.  I'm going to be away for a week and a half so I won't be able to give further feedback after today.

mcraj017:
@adinn CodeFormatCheck failed in GraalVM Gate / labsjdk-ce-11 style,fullbuild compiler (pull_request) , 

can you please  run "mx eclipseformat", check in changes and repush?

```
gate: 28 Jul 2021 16:10:55(+03:12) ABORT: CodeFormatCheck [0:00:25.080044] [disk (free/total): 36.3GB/83.2GB]
Formatter modified files - run "mx eclipseformat", check in changes and repush
```

compiler/src/org.graalvm.compiler.truffle.compiler/src/org/graalvm/compiler/truffle/compiler/nodes/frame/NewFrameNode.java
adinn: Ok, thanks for confirming.

compiler/src/org.graalvm.compiler.virtual/src/org/graalvm/compiler/virtual/phases/ea/VirtualizerToolImpl.java
adinn: Yes, will do.

adinn:
@mcraj017 Hi Munish. I've fixed the style errors and also made the corrections Tom suggested.

dougxc:
LGTM

dougxc (approved):

adinn:
@dougxc Thanks for the review!

adinn:
Hi @dougxc

Is this PR going through the internal gate or has it got stuck for some reason?

dougxc:
I’m on vacation for 2 weeks but believe @mcraj017 is in the process of merging it.

adinn:
@dougxc sorry to interrupt your hols. thanks for the update!

adinn:
@mcraj017 gentle ping: is this still waiting to be merged?

mcraj017:
@adinn Yes, still waiting for merge, it will get merged by this week

PR #3622, opened 2021-07-23
[GR-33359] Fix image name argument parsing when combined with -m option

ivan-ristovic:
## Summary
This PR adresses two problems related to `native-image` argument handling. Namely, it:
- fixes `NullPointerException` in cases when a class could not be found on the provided module path
- modifies argument handling so that when image name is combined with `--module` argument, the image name is not mistakenly interpreted as a class name

### Main class resolving
Currently, our main class resolving is done via:
https://github.com/oracle/graal/blob/8d70605a5ec73bb2f0f0aef9a84cf4d8d8816366/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGeneratorRunner.java#L340

There is a catch block present for `ClassNotFoundException`, however, looking at the call chain, it is possible that control will reach `java.lang.Class.forName(Module, String)`, and by looking at the documentation for this method the return value could be `null` - which is currently not covered. Therefore, the following line will produce a `NullPointerException`:
https://github.com/oracle/graal/blob/8d70605a5ec73bb2f0f0aef9a84cf4d8d8816366/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/NativeImageGeneratorRunner.java#L351

### Image name handling in the presence of `--module` option
Invoking `native-image` with the following arguments
```sh
$ native-image MyClass foo
```
will produce a binary with the name `foo`. However, combining this with the `--module` (`-m` for short) option, for example
```sh
$ native-image -m my.module/MyClass foo
```
will cause a `NullPointerException` in the same line as mentioned above. This is tied to class resolving which could return `null`, however it is a misinterpreted main class argument when `-m` option is present - the image name will be interpreted as an overrule for the class name.

## Implementation details
This PR solves this problem by reusing the same overrule exclusion code used when `-jar` option is present:
https://github.com/oracle/graal/blob/8d70605a5ec73bb2f0f0aef9a84cf4d8d8816366/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java#L1168
As for main class resolving, a check is added to ensure that the `mainClass` field is not `null` before proceeding onto entry method lookup.

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
@@ -133,6 +133,7 @@ class DefaultOptionHandler extends NativeImage.OptionHandler<NativeImage> {
                     nativeImage.addPlainImageBuilderArg(nativeImage.oHClass + mainClassModuleArgParts[1]);
                 }
                 nativeImage.addPlainImageBuilderArg(nativeImage.oHModule + mainClassModuleArgParts[0]);
+                nativeImage.setModuleOptionMode(true);
                 return true;
             case "--configurations-path":
                 args.poll();
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -1166,7 +1166,7 @@ public class NativeImage {
                 }
             }
 
-            if (!jarOptionMode && !moduleOptionMode) {
+            if (!jarOptionMode) {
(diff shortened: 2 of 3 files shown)


olpaw:
The reason is that now
```java
                            if (hasMainClassModule) {
                                imageBuilderArgs.add(oH(SubstrateOptions.Name, "image-name from module-name") + mainClassModule.toLowerCase());
                            } 
```
is not used anymore.

ivan-ristovic:
@olpaw Yes, I realise now that in fact we wish to do the same as before, just skip overruling - I moved the check, I think it should be fine now.

olpaw (approved):

olpaw:
Rerunning in internal CI now ...

olpaw:
PR is in merge queue

Issue #3566, opened 2021-07-09
Optimize reflection configuration in Native Image


loicottet:
## Goal

Reduce image size and configuration size by limiting useless reflection capabilities being generated for methods never invoked through reflection and enabling reflection queries without configuration, while maintaining the option to fine-tune the reflection metadata being included for image size-sensitive applications.

### Reflection operation types

The Native Image implementation of Java’s reflection API currently doesn’t distinguish between two types of reflective operations:

- Queries: only require metadata about class members (i.e. Class.getMethod(s))
- Accesses: require the member to actually be present in the image (i.e. Method.invoke)

### Shortcomings of the current implementation

This design decision leads to access capabilities being provided for members that are only queried but never accessed. This has an impact on the final image size mainly due to two factors:

- Those methods are wrongly marked as reachable, which can lead to more unreachable code being included in the image;
- A proxy class providing invocation capabilities is generated for all reflection-accessible methods but is never executed.

This only happens for methods and constructors, so the following proposal focuses solely on those.

### Example

Given the following code (lightly edited for space):
```
class Test {
  public static void main() {
    callMethodsWithAnnotation(Test.class, Tag.class);
  }

  static void callMethodsWithAnnotation(Class<?> clazz, Class<?> annotationClazz) {
    for (Method m : clazz.getDeclaredMethods())
      if (m.getAnnotation(annotationClazz) != null)
        m.invoke(null);
  }

  @Tag
  void taggedMethod() { ... }

  void deadCode() { ... }
}
```

The current reflection configuration for this program is as follows:
```
[{
  "name":"Test",
  "allDeclaredMethods":true
}]
```

As a consequence, useless reflection invocation capabilities are created for the `main`, `callMethodsWithAnnotation` and, more importantly, `deadCode`, which could force the inclusion of an arbitrary amount of unreachable code into the image.

### Protocol

We measured the impact of the following changes on various regular and microservice benchmarks:
- Image size reduction when removing invocation capabilities for queried-only methods;
- Image size increase when including metadata required for reflection queries in all reachable classes;
- Configuration size reduction when including only invoked methods.

### Results

![Image impact](https://user-images.githubusercontent.com/3634331/125071019-073de300-e0b9-11eb-8e63-12564fe11810.png)

[...]

PR #3637, opened 2021-07-30
[GR-21937] Distinguish between invoked and queried reflection methods

graalvmbot:


loicottet:
Introduces a distinction among reflection methods between those that are actually invoked and the others. In addition, by default, this PR adds metadata to the image containing information about all the included methods (this is configurable, see the specification here: https://github.com/oracle/graal/issues/3566).

The PR is split into two commits, respectively for configuration changes and the implementation of method metadata.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoAccess.java
```diff
@@ -293,25 +303,33 @@ public static void lookupCodeInfo(CodeInfo info, long ip, SimpleCodeInfoQueryRes
...
+    public static void setFrameInfo(CodeInfo info, NonmovableArray<Byte> encodings) {
         CodeInfoImpl impl = cast(info);
         impl.setFrameInfoEncodings(encodings);
-        impl.setFrameInfoObjectConstants(objectConstants);
-        impl.setFrameInfoSourceClasses(sourceClasses);
-        impl.setFrameInfoSourceMethodNames(sourceMethodNames);
-        impl.setFrameInfoNames(names);
         if (!SubstrateUtil.HOSTED) {
             // notify the GC about the frame metadata that is now live
             Heap.getHeap().getRuntimeCodeInfoGCSupport().registerFrameMetadata(impl);
```
loicottet, line 311: @christianhaeubl Is it fine to just leave this here? Or does the GC need some further changes to support moving the encodings outside of the frame info?

christianhaeubl: The GCs will need further changes (they need to know which native memory locations reference Java objects).

vjovanov:
Would be good to change the `Reflection.md` to describe your changes. Then, I would point from the description of this PR there and to the key changes in the encoding.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageAgent.java
```diff
@@ -121,6 +121,7 @@ protected int onLoadCallback(JNIJavaVM vm, JvmtiEnv jvmti, JvmtiEventCallbacks c
         boolean configurationWithOrigins = false;
         int configWritePeriod = -1; // in seconds
         int configWritePeriodInitialDelay = 1; // in seconds
+        boolean trackReflectionMetadata = false;
```
vjovanov, line 124: Any reason not to enable this by default?

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright (c) 2019, 2019, Oracle and/or its affiliates. All rights reserved.
```
vjovanov, line 2: 2021

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
```diff
@@ -0,0 +1,129 @@
...
+                return true;
+            }
+            return equals(other);
+        }
+    }
+
+    public enum ConfigurationAccessKind {
+        NONE,
+        QUERIED,
+        ACCESSED;
```
vjovanov, line 103: Would `INVOKED` fit better to what this value describes?

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
```diff
@@ -0,0 +1,129 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.configure.config;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class ConfigurationKind {
+    private static Map<ConfigurationMemberKind, Map<ConfigurationAccessKind, ConfigurationKind>> cache = new HashMap<>();
```
vjovanov, line 31: This cache seems like a memory leak as it stays here forever. Also, it takes a bit of time to understand how it works due to the map-in-a-map structure.

To avoid such confusion, I would propose to use a two-dimensional array here as both keys are enums. It will take less space, and it will immediately be obvious what it is.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -206,6 +313,83 @@ private void encodeReferenceMaps() {
...
+                    dataEncodingBuffer.putSV(exceptionClassIndex);
+                }
+
+                try {
+                    byte[] annotations = annotationEncoder.encodeAnnotations(GuardedAnnotationAccess.getDeclaredAnnotations(method));
+                    dataEncodingBuffer.putUV(annotations.length);
+                    for (byte b : annotations) {
+                        dataEncodingBuffer.putS1(b);
+                    }
+                    byte[] parameterAnnotations = annotationEncoder.encodeParameterAnnotations(method.getParameterAnnotations());
```
vjovanov, line 367: What about return type annotations?

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/target/Target_sun_reflect_annotation_AnnotationParser.java
```diff
@@ -0,0 +1,443 @@
...
+            case 's':
+                return constPool.getUTF8At(buf.getShort());
+            default:
+                throw new AnnotationFormatError(
+                                "Invalid member-value tag in annotation: " + tag);
+        }
+    }
+
+    @Substitute
+    private static Object parseByteArray(int length,
```
vjovanov, line 220: All of these are needed because we use an `int` for encoding and the JDK uses `short`?

This deserves a comment, both why we use `int` and why we substitute.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -97,6 +120,53 @@
...
+        final FrequencyEncoder<JavaConstant> objectConstants;
+        final FrequencyEncoder<Class<?>> sourceClasses;
+        final FrequencyEncoder<String> sourceMethodNames;
+        final FrequencyEncoder<String> names;
+
+        private Encoders() {
+            this.objectConstants = FrequencyEncoder.createEqualityEncoder();
+            this.sourceClasses = FrequencyEncoder.createEqualityEncoder();
+            this.sourceMethodNames = FrequencyEncoder.createEqualityEncoder();
+            sourceMethodNames.addObject("<init>");
```
christianwimmer, line 133: Adding `<init>` once here screws up the frequency of this name, i.e., it will get a high index even though it is encoded frequently. When you added in `prepareMetadataForMethod` for every constructor, then the frequency would be correct.

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
christianwimmer: If the `cache` remains, I think it needs to be all `ConcurrentHashMap` since it is filled lazily.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -393,6 +577,286 @@ public boolean verifyFrameInfo(CodeInfo info) {
...
+            } else if (type.isArray()) {
+                boolean supported = true;
+                Class<?> componentType = type.getComponentType();
+                if (!componentType.isPrimitive()) {
+                    for (Object val : (Object[]) value) {
+                        supported &= supportedValue(componentType, val, stringValues);
+                    }
+                } else if (componentType == boolean.class) {
+                    for (boolean val : (boolean[]) value) {
+                        supported &= supportedValue(componentType, val, stringValues);
```
christianwimmer, line 813: Why check all the elements of primitive arrays (for all the array types below)?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -393,6 +577,286 @@ public boolean verifyFrameInfo(CodeInfo info) {
...
+                return encoders.sourceClasses.contains((Class<?>) value);
+            } else if (type == String.class) {
+                if (stringValues != null) {
+                    stringValues.add((String) value);
+                }
+            } else if (type.isEnum()) {
+                if (stringValues != null) {
+                    stringValues.add(((Enum<?>) value).name());
+                }
+                return encoders.sourceClasses.contains(type);
```
christianwimmer, line 855: When this returns `false`, there is no point in adding to `stringValues` beforehand. But if I interpret the code correctly, then `stringValues` will be ignored anyway in that case?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -393,6 +577,286 @@ public boolean verifyFrameInfo(CodeInfo info) {
...
+            }
+            return filteredAnnotations.toArray(new Annotation[0]);
+        }
+
+        private void registerStrings(Annotation[] annotations) {
+            for (Annotation annotation : annotations) {
+                List<String> stringValues = new ArrayList<>();
+                if (supportedValue(annotation.annotationType(), annotation, stringValues)) {
+                    for (String stringValue : stringValues) {
+                        encoders.sourceMethodNames.addObject(stringValue);
```
christianwimmer, line 778: Interesting that you add to `encoders.sourceMethodNames` and not `encoders.names`. Is there even a reason to have two string tables? Both tables could contain the same strings, leading to duplication. Are some table indices much smaller when using two separate tables, so that we benefit from fewer bytes in the encoding?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -393,6 +577,286 @@ public boolean verifyFrameInfo(CodeInfo info) {
...
+                encodeValue(buf, valueAccessor.invoke(annotation), type.memberTypes().get(memberName));
+            }
+        }
+
+        void encodeValue(UnsafeArrayTypeWriter buf, Object value, Class<?> type) throws InvocationTargetException, IllegalAccessException {
+            buf.putU1(tag(type));
+            if (type.isAnnotation()) {
+                encodeAnnotation(buf, (Annotation) value);
+            } else if (type.isEnum()) {
+                buf.putU2(encoders.sourceClasses.getIndex(type));
```
christianwimmer, line 641: We can have more than 65000 classes in an image. And also more than 65000 entries in `sourceMethodNames` for the `String` case below. So all of these cannot use `U2` encoding.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -393,6 +577,286 @@ public boolean verifyFrameInfo(CodeInfo info) {
...
+                }
+            } else {
+                throw shouldNotReachHere();
+            }
+        }
+
+        void encodeArray(UnsafeArrayTypeWriter buf, Object value, Class<?> componentType) throws InvocationTargetException, IllegalAccessException {
+            if (!componentType.isPrimitive()) {
+                Object[] array = (Object[]) value;
+                buf.putU2(array.length);
```
christianwimmer, line 687: Who guarantees that these array lengths fit in an `U2`?

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageAgent.java
loicottet: Since only the reachable methods are included in the metadata for now, turning this on would include metadata that is not necessarily needed. Is it more important to get running programs from the get go (instead of having to flip this one switch if needed) or is image size in the general case more important here?

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
loicottet: I chose `ACCESSED` to have a general term that also applies to fields, since these values are attached to all member types, even if only trivially. I agree it's a bit unclear, but I think `INVOKED` is too specific here. Something like `USED` could be an alternative.

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
loicottet: Switched to a 2D-array

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
loicottet: Changed

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/target/Target_sun_reflect_annotation_AnnotationParser.java
loicottet: Added comment

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/target/Target_sun_reflect_annotation_AnnotationParser.java
```diff
@@ -0,0 +1,443 @@
...
+    private static Object parseShortArray(int length,
+                    ByteBuffer buf, Target_jdk_internal_reflect_ConstantPool constPool) {
+        short[] result = new short[length];
+        boolean typeMismatch = false;
+        int tag = 0;
+
+        for (int i = 0; i < length; i++) {
+            tag = buf.get();
+            if (tag == 'S') {
+                result[i] = buf.getShort();
```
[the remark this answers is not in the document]
loicottet: The array being returned is expected to be of a primitive type, so making it generic will cause problems.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -580,4 +580,8 @@ public ReportingSupport(String reportingPath) {
             this.reportsPath = reportingPath;
         }
     }
+
+    @APIOption(name = "configure-reflection-metadata")//
+    @Option(help = "Configure inclusion of reflection metadata instead of including it for all methods by default")//
```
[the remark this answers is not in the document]
loicottet: It does mean for all reachable methods, I've updated the comment.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -173,6 +251,33 @@ public void addMethod(SharedMethod method, CompilationResult compilation, int co
         ImageSingletons.lookup(Counters.class).codeSize.add(compilation.getTargetCodeSize());
     }
 
+    @SuppressWarnings("unchecked")
+    public void prepareMetadataForClass(Class<?> clazz) {
```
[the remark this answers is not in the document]
loicottet: Done

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -393,6 +577,286 @@ public boolean verifyFrameInfo(CodeInfo info) {
...
+     * Class and String values are represented by their index in the source metadata encoders
+     * instead of their constant pool indices. Additionally, Class objects are encoded directly
+     * instead of through their type signature. Primitive values are written directly into the
+     * encoding. This means that our encoding can be of a different length from the JDK one.
+     * 
+     * We use a modified version of the ConstantPool and AnnotationParser classes to decode the
+     * data, since those are not used in their original functions at runtime. (see
+     * {@link com.oracle.svm.core.jdk.Target_jdk_internal_reflect_ConstantPool})
+     */
+    class AnnotationEncoder {
```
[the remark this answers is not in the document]
loicottet: The code needs access to the string and class encoders, so I got rid of the inner class and included the methods in the main class.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
loicottet: Fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
loicottet: I indeed don't need to check those, I removed the checks

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
loicottet: Yes, `stringValues` collects strings during the check and only registers them at the end if it succeeds.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
loicottet: The distinction was present already in the code so I don't know the rationale behind it. It seems that `names` is meant for debugging information though so it felt more appropriate to add it to the encoder that is used for code that is actually required for correctness. The naming is perhaps not ideal then, that's true.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
loicottet: I ran into that problem on JCK gates, I'm in the process of fixing it.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
loicottet: The JDK decoder reads those as shorts so there must be an invariant in the encoding code requiring it.

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationType.java
```diff
@@ -278,40 +305,48 @@ public void setAllPublicFields() {
...
-        return allDeclaredMethods;
+    public ConfigurationAccessKind getAllDeclaredMethods() {
+        return allDeclaredMethodsAccess;
     }
 
-    public boolean haveAllPublicMethods() {
-        return allPublicMethods;
+    public ConfigurationAccessKind getAllPublicMethods() {
+        return allPublicMethodsAccess;
     }
```
peter-hofer, line 314: These methods seem to be called only in this class, you should delete them and access the fields directly.

peter-hofer (changes requested):

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -1245,24 +1306,18 @@ public static void onUnload() {
...
-                    brk("java/lang/Class", "getConstructor", "([Ljava/lang/Class;)Ljava/lang/reflect/Constructor;", BreakpointInterceptor::getConstructor),
-                    brk("java/lang/Class", "getDeclaredMethod", "(Ljava/lang/String;[Ljava/lang/Class;)Ljava/lang/reflect/Method;", BreakpointInterceptor::getDeclaredMethod),
-                    brk("java/lang/Class", "getDeclaredConstructor", "([Ljava/lang/Class;)Ljava/lang/reflect/Constructor;", BreakpointInterceptor::getConstructor),
 
                     brk("java/lang/Class", "getEnclosingMethod", "()Ljava/lang/reflect/Method;", BreakpointInterceptor::getEnclosingMethod),
                     brk("java/lang/Class", "getEnclosingConstructor", "()Ljava/lang/reflect/Constructor;", BreakpointInterceptor::getEnclosingMethod),
 
+                    brk("java/lang/reflect/Method", "invoke", "(Ljava/lang/Object;[Ljava/lang/Object;)Ljava/lang/Object;", BreakpointInterceptor::invokeMethod),
+                    brk("java/lang/reflect/Constructor", "newInstance", "([Ljava/lang/Object;)Ljava/lang/Object;", BreakpointInterceptor::invokeConstructor),
```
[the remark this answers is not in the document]
christianwimmer: HotSpot already has a slow/fast distinction: `NativeMethodAccessorImpl` for the first invokes of a method, and then generated bytecode for frequently invoked methods. So maybe setting the breakpoint in `NativeMethodAccessorImpl` instead of `Method` is sufficient to avoid any performance problems? It is not overly future-proof though, because HotSpot is going to change all of that as part of https://openjdk.java.net/jeps/416

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -206,6 +313,83 @@ private void encodeReferenceMaps() {
...
+        UnsafeArrayTypeWriter dataEncodingBuffer = UnsafeArrayTypeWriter.create(ByteArrayReader.supportsUnalignedMemoryAccess());
+        UnsafeArrayTypeWriter indexEncodingBuffer = UnsafeArrayTypeWriter.create(ByteArrayReader.supportsUnalignedMemoryAccess());
+        long lastTypeID = -1;
+        for (Map.Entry<SharedType, Set<Executable>> entry : methodData.entrySet()) {
+            SharedType declaringType = entry.getKey();
+            Set<Executable> methods = entry.getValue();
+            long typeID = declaringType.getHub().getTypeID();
+            assert typeID > lastTypeID;
+            lastTypeID++;
+            while (lastTypeID < typeID) {
```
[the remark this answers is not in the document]
loicottet: In this case we may still have holes if some types do not have any method metadata stored in the image.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageAgent.java
```diff
@@ -189,6 +190,8 @@ protected int onLoadCallback(JNIJavaVM vm, JvmtiEnv jvmti, JvmtiEventCallbacks c
                 build = Boolean.parseBoolean(getTokenValue(token));
             } else if (token.equals("experimental-configuration-with-origins")) {
                 configurationWithOrigins = true;
+            } else if (token.equals("track-reflection-metadata")) {
+                trackReflectionMetadata = true;
```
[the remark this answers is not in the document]
loicottet: Done

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/trace/ReflectionProcessor.java
```diff
@@ -116,6 +117,7 @@ public void processEntry(Map<String, ?> entry) {
             return;
         }
         ConfigurationMemberKind memberKind = ConfigurationMemberKind.PUBLIC;
+        ConfigurationAccessKind accessKind = ConfigurationAccessKind.ACCESSED;
```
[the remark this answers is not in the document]
loicottet: Fixed.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -1245,24 +1306,18 @@ public static void onUnload() {
...
                     brk("java/lang/Class", "getDeclaredField", "(Ljava/lang/String;)Ljava/lang/reflect/Field;", BreakpointInterceptor::getDeclaredField),
-                    brk("java/lang/Class", "getMethod", "(Ljava/lang/String;[Ljava/lang/Class;)Ljava/lang/reflect/Method;", BreakpointInterceptor::getMethod),
-                    brk("java/lang/Class", "getConstructor", "([Ljava/lang/Class;)Ljava/lang/reflect/Constructor;", BreakpointInterceptor::getConstructor),
-                    brk("java/lang/Class", "getDeclaredMethod", "(Ljava/lang/String;[Ljava/lang/Class;)Ljava/lang/reflect/Method;", BreakpointInterceptor::getDeclaredMethod),
-                    brk("java/lang/Class", "getDeclaredConstructor", "([Ljava/lang/Class;)Ljava/lang/reflect/Constructor;", BreakpointInterceptor::getConstructor),
 
                     brk("java/lang/Class", "getEnclosingMethod", "()Ljava/lang/reflect/Method;", BreakpointInterceptor::getEnclosingMethod),
                     brk("java/lang/Class", "getEnclosingConstructor", "()Ljava/lang/reflect/Constructor;", BreakpointInterceptor::getEnclosingMethod),
```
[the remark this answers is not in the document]
loicottet: We only need to move the method and constructor query methods, fields and classes are still queried the same way even without the flag active.

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
```diff
@@ -0,0 +1,131 @@
...
+            return equals(other);
+        }
+    }
+
+    public enum ConfigurationAccessKind {
+        NONE,
+        QUERIED,
+        ACCESSED;
+
+        public ConfigurationAccessKind combine(ConfigurationAccessKind other) {
```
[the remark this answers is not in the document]
loicottet: Done

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
```diff
@@ -157,7 +160,14 @@ private static void traceBreakpoint(JNIEnvironment env, JNIObjectHandle clazz, J
...
                             args);
-            guarantee(!testException(env));
+            JNIObjectHandle exception = handleException(env);
+            if (exception.notEqual(nullHandle())) {
+                /*
+                 * A stack overflow error happening during a breakpoint should be handled by the
+                 * program, not the agent.
+                 */
+                guarantee(jniFunctions().getIsInstanceOf().invoke(env, exception, agent.handles().javaLangStackOverflowError));
+            }
```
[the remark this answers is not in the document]
loicottet: This happens on a JCK test that causes a stack overflow through reflective calls. When that happens, the agent crashes because the stack overflow error happens during the breakpoint instead of during the program, so this simply aborts the current registration (useless in this case) and hands control back to the program to handle the exception. I have changed my implementation to not clear the exception on inspection.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
```diff
@@ -42,12 +42,17 @@
...
 
 public interface ReflectionRegistry {
     void register(Class<?>... classes);
 
-    void register(Executable... methods);
+    void register(boolean queriedOnly, Executable... methods);
 
     void register(boolean finalIsWritable, Field... fields);
 
+    default Set<Executable> getQueriedMethods() {
```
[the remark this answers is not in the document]
loicottet: Moved and renamed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -580,4 +580,8 @@ public ReportingSupport(String reportingPath) {
             this.reportsPath = reportingPath;
         }
     }
+
+    @APIOption(name = "configure-reflection-metadata")//
+    @Option(help = "Configure inclusion of reflection metadata instead of including it for all reachable methods by default")//
```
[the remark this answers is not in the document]
loicottet: Changed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
```diff
@@ -206,6 +314,83 @@ private void encodeReferenceMaps() {
         }
     }
 
+    public static final int NO_METHOD_METADATA = -1;
+
+    private void encodeMethodMetadata() {
```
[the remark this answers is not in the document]
loicottet: The sharing is quite important since most strings are method names which are also needed for stack walks. I refactored all the metadata-emitting logic into a new `MethodMetadataEncoder` class marked as `HOSTED_ONLY`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/RuntimeCodeInfoAccess.java
```diff
@@ -309,6 +309,8 @@ public static void forEachArray(CodeInfo info, NonmovableArrayAction action) {
         action.apply(impl.getCodeInfoIndex());
         action.apply(impl.getCodeInfoEncodings());
         action.apply(impl.getStackReferenceMapEncoding());
+        action.apply(impl.getMethodReflectionMetadataEncoding());
+        action.apply(impl.getMethodReflectionMetadataIndexEncoding());
```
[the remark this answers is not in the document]
loicottet: Removed

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/annotation/AnnotationTypeFeature.java
```diff
@@ -79,6 +80,22 @@ public void duringAnalysis(DuringAnalysisAccess access) {
         Stream<AnnotatedElement> allElements = Stream.concat(Stream.concat(universe.getFields().stream(), universe.getMethods().stream()), universe.getTypes().stream());
         Stream<AnnotatedElement> newElements = allElements.filter(visitedElements::add);
         newElements.forEach(this::reportAnnotation);
+
+        /*
+         * Parsing annotation data in reflection classes requires being able to instantiate all
+         * annotation types at runtime.
+         */
```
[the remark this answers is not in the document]
loicottet: Done

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ReflectionConfigurationParserDelegate.java
```diff
@@ -50,23 +50,23 @@ default T resolveType(String typeName) {
 
     void registerDeclaredFields(T type);
 
-    void registerPublicMethods(T type);
+    void registerPublicMethods(T type, boolean queriedOnly);
```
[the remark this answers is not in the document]
loicottet: Switched

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageCodeCache.java
```diff
@@ -219,6 +222,31 @@ public void buildRuntimeMetadata(CFunctionPointer firstMethod, UnsignedWord code
...
+        if (SubstrateOptions.ConfigureReflectionMetadata.getValue()) {
+            for (Executable queriedMethod : ImageSingletons.lookup(RuntimeReflectionSupport.class).getQueriedMethods()) {
+                System.out.println("Trying to add " + queriedMethod);
+                HostedMethod method = imageHeap.getMetaAccess().optionalLookupJavaMethod(queriedMethod);
+                if (method != null) {
+                    System.out.println("Adding metadata for " + method);
+                    codeInfoEncoder.prepareMetadataForMethod(method, queriedMethod);
+                } else {
+                    System.out.println("Not adding");
+                }
```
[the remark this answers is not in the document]
loicottet: Yeah those slipped through, they're already gone.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/hosted/ReflectionDataBuilder.java
```diff
@@ -118,10 +122,13 @@ public void register(Class<?>... classes) {
...
     @Override
-    public void register(Executable... methods) {
+    public void register(boolean queriedOnly, Executable... methods) {
         checkNotSealed();
+        if (queriedOnly && !SubstrateOptions.ConfigureReflectionMetadata.getValue()) {
+            throw UserError.abort("Found manual reflection metadata configuration. Please use --configure-reflection-metadata to enable this behavior.");
+        }
         for (Executable method : methods) {
-            if (reflectionMethods.add(method)) {
+            if (queriedOnly ? queriedMethods.add(method) : reflectionMethods.add(method)) {
```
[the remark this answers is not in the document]
loicottet: Done

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/target/RuntimeReflectionConstructorsImpl.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright (c) 2012, 2020, Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
loicottet: Done and checked

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/proxy/hosted/DynamicProxyFeature.java
```diff
@@ -93,4 +96,18 @@ public void beforeCompilation(BeforeCompilationAccess access) {
             throw proxyFallback;
         }
     }
+
+    @Override
+    public void duringAnalysis(DuringAnalysisAccess access) {
+        FeatureImpl.DuringAnalysisAccessImpl accessImpl = (FeatureImpl.DuringAnalysisAccessImpl) access;
+        for (AnalysisType type : accessImpl.getUniverse().getTypes()) {
```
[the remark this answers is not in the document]
loicottet: Moved

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/target/RuntimeReflectionConstructorsImpl.java
```diff
@@ -0,0 +1,65 @@
...
+        return SubstrateUtil.cast(method, Method.class);
+    }
+
+    @Override
+    public Constructor<?> newConstructor(Class<?> declaringClass, Class<?>[] parameterTypes, Class<?>[] checkedExceptions, int modifiers, byte[] annotations,
+                    byte[] parameterAnnotations) {
+        Target_java_lang_reflect_Constructor cons = new Target_java_lang_reflect_Constructor();
+        String signature = toInternalSignature(void.class, parameterTypes);
+        cons.constructor(declaringClass, parameterTypes, checkedExceptions, modifiers, -1, signature, annotations, parameterAnnotations);
+        return SubstrateUtil.cast(cons, Constructor.class);
```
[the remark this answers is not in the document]
loicottet: You can't have a `native` constructor so it's impossible to alias it directly, we have to use a regular method with `@TargetElement(name = CONSTRUCTOR_NAME)`

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/target/Target_java_lang_reflect_ReflectAccess.java
```diff
@@ -60,6 +63,20 @@ public Target_java_lang_reflect_Field copyField(Target_java_lang_reflect_Field f
...
         Target_java_lang_reflect_Constructor copy = constructor.copy();
         copy.genericInfo = constructor.genericInfo;
+        Util_java_lang_reflect_Method.copyExecutable(SubstrateUtil.cast(copy, Target_java_lang_reflect_Executable.class),
+                        SubstrateUtil.cast(constructor, Target_java_lang_reflect_Executable.class));
         return copy;
     }
 }
+
+class Util_java_lang_reflect_Method {
+    static void copyExecutable(Target_java_lang_reflect_Executable copy, Target_java_lang_reflect_Executable executable) {
```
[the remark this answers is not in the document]
loicottet: These methods were already present here, @cstancu was there a reason to not substitute `Method.copy` and `Constructor.copy` directly?

substratevm/src/com.oracle.svm.jni/src/com/oracle/svm/jni/access/JNIAccessFeature.java
```diff
@@ -134,7 +134,8 @@ public void register(Class<?>... classes) {
         }
 
         @Override
-        public void register(Executable... methods) {
+        public void register(boolean queriedOnly, Executable... methods) {
+            assert !queriedOnly;
```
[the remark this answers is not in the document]
loicottet: True, good catch!

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
```diff
@@ -0,0 +1,131 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.configure.config;
+
+public final class ConfigurationKind {
```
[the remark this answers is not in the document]
loicottet: Renamed

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationType.java
```diff
@@ -166,23 +172,26 @@ private void removeMethods(ConfigurationType other) {
...
-    private void setFlagsFromOtherUsingPredicate(ConfigurationType other, BiPredicate<Boolean, Boolean> predicate) {
-        allDeclaredClasses = predicate.test(allDeclaredClasses, other.allDeclaredClasses);
-        allPublicClasses = predicate.test(allPublicClasses, other.allPublicClasses);
-        allDeclaredFields = predicate.test(allDeclaredFields, other.allDeclaredFields);
-        allPublicFields = predicate.test(allPublicFields, other.allPublicFields);
-        allDeclaredMethods = predicate.test(allDeclaredMethods, other.allDeclaredMethods);
-        allPublicMethods = predicate.test(allPublicMethods, other.allPublicMethods);
-        allDeclaredConstructors = predicate.test(allDeclaredConstructors, other.allDeclaredConstructors);
-        allPublicConstructors = predicate.test(allPublicConstructors, other.allPublicConstructors);
+    private void setFlagsFromOtherUsingPredicate(ConfigurationType other, BiPredicate<Boolean, Boolean> flagPredicate,
```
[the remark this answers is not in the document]
loicottet: Renamed

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationType.java
loicottet: Refactored

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoEncoder.java
loicottet: Fixed

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
```diff
@@ -0,0 +1,131 @@
...
+        return accessKind;
+    }
+
+    public static ConfigurationKind get(ConfigurationMemberKind memberKind, ConfigurationAccessKind accessKind) {
+        int memberIndex = memberKind.ordinal();
+        int accessIndex = accessKind.ordinal();
+        if (cache[memberIndex][accessIndex] == null) {
+            cache[memberIndex][accessIndex] = new ConfigurationKind(memberKind, accessKind);
+        }
+        return cache[memberIndex][accessIndex];
```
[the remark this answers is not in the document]
loicottet: Done

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoAccess.java
loicottet: As you pointed out, this will be fine since the new encodings are only primitive arrays.

substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/NativeImageAgent.java
loicottet: Enabled by default

substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/ConfigurationKind.java
```diff
@@ -0,0 +1,131 @@
...
+            return ordinal() >= other.ordinal();
+        }
+    }
+
+    public ConfigurationKind intersect(ConfigurationKind other) {
+        return get(memberKind.intersect(other.memberKind), accessKind.combine(other.accessKind));
+    }
+
+    public ConfigurationKind union(ConfigurationKind other) {
+        return get(memberKind.union(other.memberKind), accessKind.combine(other.accessKind));
```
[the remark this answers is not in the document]
loicottet: I actually misunderstood the reason why `intersect` was used. I created `union` and `intersect` for `ConfigurationMemberAccessibility` as well.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/proxy/hosted/DynamicProxyFeature.java
```diff
@@ -93,4 +96,18 @@ public void beforeCompilation(BeforeCompilationAccess access) {
...
+            if (type.isAnnotation() && type.isReachable()) {
+                try {
+                    dynamicProxySupport.addProxyClass(type.getJavaClass());
+                } catch (Throwable t) {
+                    // TODO ignore for now
```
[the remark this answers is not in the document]
loicottet: Done

PR #3643, opened 2021-08-02
Fix infinite recursion.

oraluben:
I didn't run this code but this seems obvious and easy to fix.

Since this function didn't break any functionality, I guess deleting the function would be another option.


oraluben:
I re-commit with an email that signed the OCA, thanks.

PR #3646, opened 2021-08-04
Specialize SLDivNode for constant-at-runtime divisors

CAFxX:
With this change, that is really just meant as an example of an additional kind of use cases for `@Cached`, the following contrived example:

```
function main() {
  i = 0;
  j = 0;
  k = 1;
  while (j < 1000000000) {
    j = j + i/(100000000+k);
    i = i + 1;
    if (k == 1) {
      k = 3;
    } else {
      k = 1;
    }
  }
  return j;
}
```

runs on my machine in:
```
real    0m3.020s
user    0m2.241s
sys     0m0.107s
```
whereas before it took more than twice as long:
```
real    0m5.912s
user    0m5.175s
sys     0m0.082s
```

Incidentally, the resulting optimization capability (strength reduction of divisions with non-constant divisors at compile time) is AFAIK not available in any AOT compiler (gcc, clang, icc). There is a point to be made that such an optimization would be better implemented in graalvm itself, rather than sl.

Initially submitted at https://github.com/graalvm/simplelanguage/pull/108

### DIFF
--- a/truffle/src/com.oracle.truffle.sl/src/com/oracle/truffle/sl/nodes/expression/SLDivNode.java
+++ b/truffle/src/com.oracle.truffle.sl/src/com/oracle/truffle/sl/nodes/expression/SLDivNode.java
@@ -41,6 +41,7 @@
 package com.oracle.truffle.sl.nodes.expression;
 
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
+import com.oracle.truffle.api.dsl.Cached;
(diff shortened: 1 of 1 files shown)


CAFxX:
Already signed

CAFxX:
Ping?

CAFxX:
@graalvmbot ping

CAFxX:
Already signed (as already mentioned above)

CAFxX:
I'm closing this since there is seemingly no interest

PR #3668, opened 2021-08-11
3667: fix deprecated stat64 on Darwin

### LINKED ISSUE #3667: stat64 deprecated on Darwin
stat64 and related calls are deprecated on Darwin (e.g. see https://bugs.openjdk.java.net/browse/JDK-8021771)
GraalVM uses those, and this results into an unresolved _fstat64 symbol in the resulting executable.

While this is currently not a problem for running the executable, this is a stopper for acceptance on the AppStore. (I have a report of at least 1 app that was rejected because of this). Since stat64 is deprecated, we can expect more issues when using it in the future.

Compile the following code:

```import com.sun.management.OperatingSystemMXBean;
import com.sun.management.UnixOperatingSystemMXBean;
import java.lang.management.ManagementFactory;

...

OperatingSystemMXBean osMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
if (osMXBean instanceof UnixOperatingSystemMXBean) {
long openFileDescriptorCount = ((UnixOperatingSystemMXBean) osMXBean).getOpenFileDescriptorCount();
System.out.println("openFileDescriptorCount = " + openFileDescriptorCount);

```
`nm <output> |grep fstat64 `
will return
`U _fstat64`

```

HEAD version of graalvm, built with a labs-jdk16

### PR DESCRIPTION
**Pull request by johanvos**
This PR fixes #3667 by replacing stat64 symbols with stat symbols on Darwin.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/darwin/DarwinStat.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/darwin/DarwinStat.java
@@ -41,16 +41,16 @@ import com.oracle.svm.core.posix.headers.PosixDirectives;
 public class DarwinStat {
 
     @CStruct(addStructKeyword = true)
-    public interface stat64 extends PointerBase {
+    public interface stat extends PointerBase {
         @CField
         long st_size();
     }
 
-    @CFunction("fstat64")
-    public static native int fstat64(int fd, stat64 buf);
+    @CFunction("fstat")
+    public static native int fstat(int fd, stat buf);
 
     public static class NoTransitions {
         @CFunction(transition = CFunction.Transition.NO_TRANSITION)
-        public static native int fstat64(int fd, stat64 buf);
+        public static native int fstat(int fd, stat buf);
     }
 }
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixStat.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/PosixStat.java
@@ -44,8 +44,8 @@ public final class PosixStat {
             LinuxStat.stat64 stat = StackValue.get(LinuxStat.stat64.class);
             result = LinuxStat.fstat64(fd, stat);
         } else if (Platform.includedIn(Platform.DARWIN.class)) {
-            DarwinStat.stat64 stat = StackValue.get(DarwinStat.stat64.class);
-            result = DarwinStat.fstat64(fd, stat);
+            DarwinStat.stat stat = StackValue.get(DarwinStat.stat.class);
+            result = DarwinStat.fstat(fd, stat);
         } else {
             throw VMError.shouldNotReachHere("Unsupported platform");
         }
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by oubidar-Abderrahim** [comment]
Thank you for contributing to GraalVM, this Pr is tracked internally on GR-33126

**Comment by johanvos** [comment]
> Thank you for contributing to GraalVM, this PR is tracked internally on GR-33126

Is there a way to follow what is happening with this (in case it is done internally)?

File: substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/darwin/DarwinStat.java [hunk 1]
```diff
@@ -41,16 +41,16 @@
...
     @CStruct(addStructKeyword = true)
-    public interface stat64 extends PointerBase {
+    public interface stat extends PointerBase {
         @CField
         long st_size();
     }
 
-    @CFunction("fstat64")
-    public static native int fstat64(int fd, stat64 buf);
+    @CFunction("fstat")
```
peter-hofer, line 49: The situation is a bit more complicated than that: `_DARWIN_FEATURE_ONLY_64_BIT_INODE`, which is the default by now, changes the entire `struct stat` layout, which we notice because we get the struct layout from the header files (via generated C code, see `QueryCodeWriter`). But the header files also change C calls from `fstat` to `fstat$INODE64`, which we don't notice because what we specify in `@CFunction` is the actual symbol name. I have the following patch to address this (and explicitly define `_DARWIN_USE_64_BIT_INODE` to be safe). Can you please check that this works for the app store?
```patch
diff --git a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/PosixDirectives.java b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/PosixDirectives.java
index 14eb719e94f..8c877c6328e 100644
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/PosixDirectives.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/PosixDirectives.java
@@ -95,6 +95,6 @@ public class PosixDirectives implements CContext.Directives {
 
     @Override
     public List<String> getMacroDefinitions() {
-        return Arrays.asList("_GNU_SOURCE", "_LARGEFILE64_SOURCE");
+        return Arrays.asList("_GNU_SOURCE", "_LARGEFILE64_SOURCE", "_DARWIN_USE_64_BIT_INODE");
     }
 }
diff --git a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/darwin/DarwinStat.java b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/darwin/DarwinStat.java
index de1296e69e7..0890dfef134 100644
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/darwin/DarwinStat.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/darwin/DarwinStat.java
@@ -40,17 +40,23 @@ import com.oracle.svm.core.posix.headers.PosixDirectives;
 @CContext(PosixDirectives.class)
 public class DarwinStat {
 
+    /*
+     * NOTE that we set _DARWIN_USE_64_BIT_INODE in the C directives to force a layout for struct
+     * stat with a 64-bit st_ino, and we have to call functions with a $INODE64 suffix to match,
+     * such as fstat$INODE64.
+     */
+
     @CStruct(addStructKeyword = true)
     public interface stat extends PointerBase {
         @CField
         long st_size();
     }
 
-    @CFunction("fstat")
+    @CFunction("fstat$INODE64")
     public static native int fstat(int fd, stat buf);
 
     public static class NoTransitions {
-        @CFunction(transition = CFunction.Transition.NO_TRANSITION)
+        @CFunction(value = "fstat$INODE64", transition = CFunction.Transition.NO_TRANSITION)
         public static native int fstat(int fd, stat buf);
     }
 }
```

peter-hofer (commented):

PR #3672, opened 2021-08-13
Redict the initilization work to HostVM

ziyilin:
Different HostVM subclasses can have different initilization strategy.
The actual work should be implemented in the concrete HostVM subclasses.


substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/meta/AnalysisType.java
```diff
@@ -823,7 +823,7 @@ public final boolean isInitialized() {
     @Override
     public void initialize() {
         if (!wrapped.isInitialized()) {
-            throw GraalError.shouldNotReachHere("Classes can only be initialized using methods in ClassInitializationFeature: " + toClassName());
+            universe.hostVM.initializeType(this);
```
christianwimmer, line 826: I don't think this is a valid change. `HostVM.initializeType` is not about class initialization, it is a more general initialization method. And it is already called for every `AnalysisType` as part of the `initializationTask`.

PR #3681, opened 2021-08-16
[GR-33063] NativeImageResourceFileSystemProvider is used only in cases when Path/URI is from the resource system.

jovanstevanovic:
Before this PR, the implementation `NativeImageResourceFileSystemProvider#newFileSystem` was too greedy. 

If the NativeImageResourceFileSystemProvider is first in the list of `FileSystemProvider.installedProviders()` and the user uses `java.nio.file.FileSystems#newFileSystem(java.nio.file.Path, java.lang.ClassLoader)` it will be picked even when the requested filesystem is for a zip-file.

Other FileSystemProvider makes sure that they only return a FileSystem for Paths and URIs that they are responsible for. If they see a Path or URI that does not represent the file system they can provide they throw an UnsupportedOperationException to ensure the next FileSystemProvider in the list has a chance to construct a FileSystem for the given Path/URI.

Both variants of `NativeImageResourceFileSystemProvider#newFileSystem` are changed to only return a FileSystem if they see a Path/URI that represents a resource location that is part of the image that gets built.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystemProvider.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystemProvider.java
@@ -58,7 +58,8 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 public class NativeImageResourceFileSystemProvider extends FileSystemProvider {
 
-    private final String resourcePath = "file:/resources";
+    private final String resourcePath = "/resources";
+    private final String resourceUri = JavaNetSubstitutions.FILE_PROTOCOL + ":" + resourcePath;
     private NativeImageResourceFileSystem fileSystem;
     private final Lock writeLock;
     private final Lock readLock;
@@ -70,8 +71,8 @@ public class NativeImageResourceFileSystemProvider extends FileSystemProvider {
         }
 
         try {
-            // Syntax for URI resource is resource:file:resources!/{entry}
-            return Paths.get(new URI(resourcePath));
+            // Syntax for URI resource is resource:file:/resources/{entry}
+            return Paths.get(new URI(resourceUri));
         } catch (URISyntaxException e) {
             throw new IllegalArgumentException(e.getMessage(), e);
         }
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystemProvider.java
```diff
@@ -58,7 +58,8 @@
 
 public class NativeImageResourceFileSystemProvider extends FileSystemProvider {
 
-    private final String resourcePath = "file:/resources";
+    private final String resourceUri = "file:/resources";
```
olpaw, line 61:
```java
private final String resourceUri = "file:" + resourcePath;
```

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystemProvider.java
```diff
@@ -70,8 +71,8 @@ private Path uriToPath(URI uri) {
         }
 
         try {
-            // Syntax for URI resource is resource:file:resources!/{entry}
```
olpaw, line 73: So that comment was wrong all along, right?

jovanstevanovic: Unfortunately yes, I changed implementation slightly during the review period but forgot to update the comment accordingly.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/NativeImageResourceFileSystemProvider.java
jovanstevanovic: Done.

olpaw (approved):
I tested with building of the hellomodule test with debuginfo (-g uses ZipFileSystemProvider at image built-time) which is where I found the reported issue. I can confirm that this is now working as expected. Thanks for fixing.

olpaw:
Running in our CI now. Should be soon on master.

jovanstevanovic:
> I tested with building of the hellomodule test with debuginfo (-g uses ZipFileSystemProvider at image built-time) which is where I found the reported issue. I can confirm that this is now working as expected. Thanks for fixing.

Glad we solve a problem!

PR #3682, opened 2021-08-16
Add associative rule to add/sub node

kelthuzadx:
To mirror what C2 does: 

JDK-8270366: Add associative rule to add/sub node

### DIFF
--- a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/AddNode.java
+++ b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/AddNode.java
@@ -94,6 +94,40 @@ public class AddNode extends BinaryArithmeticNode<Add> implements NarrowableArit
                     return sub.getX();
                 }
             }
+            if (forX instanceof MulNode && forY instanceof MulNode) {
+                MulNode in1 = (MulNode) forX;
+                MulNode in2 = (MulNode) forY;
+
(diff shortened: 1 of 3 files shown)


compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/AddNode.java
```diff
@@ -94,6 +94,40 @@ private static ValueNode canonical(AddNode addNode, BinaryOp<Add> op, ValueNode
...
+                    addIn1 = in1.getX();
+                    addIn2 = in2.getY();
+                    mulIn = in1.getY();
+                } else if (in1.getY() == in2.getY()) {
+                    // Convert a*c+b*c into (a+b)*c
+                    addIn1 = in1.getX();
+                    addIn2 = in2.getX();
+                    mulIn = in1.getY();
+                } else if (in1.getX() == in2.getY()) {
+                    // Convert a*b+c*a into a*(b+c)
```
SwapnilGaikwad, line 120: Just curious, can ReassociationPhase avoid the need of checking all four combinations of inputs? e.g., if (c * a) is re-associated (a * c) then the expression would match  a * b + a * c on [Line 105.](https://github.com/oracle/graal/pull/3682/files#diff-8bf62d3a0789204759ca4e0e64d15acd93f222f5ff27cac6c89a23226c424f64R105) above.

kelthuzadx:
PING? Can I have a review for this PR?

PR #3692, opened 2021-08-18
Fix AArch64 Mach-O relocation

Noisyfox:
This PR adds basic support for encoding AArch64 relocations currently used by GraalVM in Mach-O file.

Mach-O requires to use `ARM64_RELOC_ADDEND` for encoding addend for certain relocation types:
https://opensource.apple.com/source/ld64/ld64-274.2/src/ld/parsers/macho_relocatable_file.cpp.auto.html


Noisyfox:
This patch is already used in the development of our iOS framework and it's working fine so far.

PR #3701, opened 2021-08-20
[GR-31039] More AArch64 IndexOf Support.

graalvmbot:


compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StringLatin1Substitutions.java
```diff
@@ -78,20 +78,11 @@ public static int indexOf(byte[] value, int ch, int origFromIndex) {
     }
 
     @MethodSubstitution
-    public static int indexOf(byte[] source, int sourceCount, byte[] target, int targetCount, int origFromIndex) {
-        int fromIndex = origFromIndex;
-        if (injectBranchProbability(UNLIKELY_PROBABILITY, fromIndex >= sourceCount)) {
-            return (targetCount == 0 ? sourceCount : -1);
-        }
```
SwapnilGaikwad, line 85: Should we restore this check? If sourceCount = fromIndex and targetCount = 0, we want to return sourceCount.
This check is already performed in [String.java](https://github.com/graalvm/labs-openjdk-11/blob/master/src/java.base/share/classes/java/lang/String.java#L1711-L1713). However, if this method is called from somewhere other than String.java, we may need to put this check.

teshull: I left in two of the dynamic asserts, fromIndex >=0 and targetCount > 0, which I think must be true; otherwise the code would throw array out of bounds errors (which I imagine it doesn't)

Given these two assumptions, I think fromIndex >= sourceCount is unnecessary, as the scenario you pointed out can't happen; I think (sourceCount - fromIndex < targetCount) should be good enough

teshull:
@SwapnilGaikwad I changed both the scalar and simd code a fair amount to hopefully be both a little smaller and faster. I haven't done any testing yet to make sure it works, but could you look over it to make sure things make sense?

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayIndexOfOp.java
```diff
@@ -77,190 +90,301 @@ public AArch64ArrayIndexOfOp(int arrayBaseOffset, JavaKind valueKind, LIRGenerat
...
         /*
-         * Add charSize to the curIndex, and retry if the end of the array has not yet been reached
-         * (i.e. the curIndex is still < 0).
+         * Add elementSize to the curIndex and retry if the end of the search region has not yet
+         * been reached, i.e., the curIndex is still < 0.
          */
-        masm.adds(64, curIndex, curIndex, charSize);
-        masm.branchConditionally(ConditionFlag.MI, searchByCharLoop);
+        masm.adds(64, curIndex, curIndex, elementByteSize);
+        masm.branchConditionally(ConditionFlag.MI, searchByElementLoop);
```
teshull, line 189: I'm now playing with the end index so that this check can (I think) be used in both the 1 and 2 element cases

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayIndexOfOp.java
```diff
@@ -77,190 +90,301 @@ public AArch64ArrayIndexOfOp(int arrayBaseOffset, JavaKind valueKind, LIRGenerat
...
+        /* Load address of first array element */
+        masm.add(64, baseAddress, asRegister(arrayPtrValue), arrayBaseOffset);
+
+        /*
+         * Search element-by-element for small arrays (with search space size of less than 32 bytes,
+         * i.e., 4 UTF-16 or 8 Latin1 elements) else search chunk-by-chunk.
+         */
+        masm.sub(64, searchLength, arrayLength, fromIndex);
+        if (!findTwoConsecutive) {
+            int chunkByteSize = 32;
```
teshull, line 377: You had this as 48. Is there a reason this isn't 32?

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayIndexOfOp.java
```diff
@@ -77,190 +90,301 @@ public AArch64ArrayIndexOfOp(int arrayBaseOffset, JavaKind valueKind, LIRGenerat
...
+        masm.bind(searchByChunkLoopHead);
+        masm.cmp(64, refAddress, chunkToReadAddress);
+        masm.branchConditionally(ConditionFlag.LE, processTail);
+        masm.bind(searchByChunkLoopTail);
+        masm.sub(64, currOffset, chunkToReadAddress, baseAddress);
+
+        masm.fldp(128, chunkPart1RegV, chunkPart2RegV, AArch64Address.createImmediateAddress(128, AArch64Address.AddressingMode.IMMEDIATE_PAIR_POST_INDEXED, chunkToReadAddress, 32));
+        /* 3. Find searchElement in the 32-byte chunk */
+        masm.neon.cmeqVVV(ASIMDSize.FullReg, eSize, chunkPart1RegV, chunkPart1RegV, searchElementRegV);
+        masm.neon.cmeqVVV(ASIMDSize.FullReg, eSize, chunkPart2RegV, chunkPart2RegV, searchElementRegV);
```
teshull, line 277: using `masm.neon.cmeqVVV` instruction instead of the prior trick you were doing

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayIndexOfOp.java
```diff
@@ -77,190 +90,301 @@ public AArch64ArrayIndexOfOp(int arrayBaseOffset, JavaKind valueKind, LIRGenerat
...
+        try (ScratchRegister scratchReg = masm.getScratchRegister()) {
+            Register magicConstantReg = scratchReg.getRegister();
+            masm.mov(magicConstantReg, magicConstant);
+            masm.neon.dupVG(ASIMDSize.FullReg, ElementSize.DoubleWord, tmpRegV1, magicConstantReg);
         }
+        masm.neon.andVVV(ASIMDSize.FullReg, chunkPart1RegV, chunkPart1RegV, tmpRegV1);
+        masm.neon.andVVV(ASIMDSize.FullReg, chunkPart2RegV, chunkPart2RegV, tmpRegV1);
+        /* 4.2 Convert 32-byte to 8-byte representation, 2 bits per byte. */
+        /* Reduce from 256 -> 128 bits. */
+        masm.neon.addpVVV(ASIMDSize.FullReg, ElementSize.Byte, chunkPart1RegV, chunkPart1RegV, chunkPart2RegV);
```
teshull, line 324: Made this code not dependent on element size; because it's finding the first matching byte I think this should be okay

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayIndexOfOp.java
```diff
@@ -77,190 +90,301 @@ public AArch64ArrayIndexOfOp(int arrayBaseOffset, JavaKind valueKind, LIRGenerat
...
+        masm.fldp(128, chunkPart1RegV, chunkPart2RegV, AArch64Address.createImmediateAddress(128, AArch64Address.AddressingMode.IMMEDIATE_PAIR_POST_INDEXED, chunkToReadAddress, 32));
+        /* 3. Find searchElement in the 32-byte chunk */
+        masm.neon.cmeqVVV(ASIMDSize.FullReg, eSize, chunkPart1RegV, chunkPart1RegV, searchElementRegV);
+        masm.neon.cmeqVVV(ASIMDSize.FullReg, eSize, chunkPart2RegV, chunkPart2RegV, searchElementRegV);
+        /* Determining if there was a match. */
+        /* Combine two registers into 1 register */
+        masm.neon.orrVVV(ASIMDSize.FullReg, tmpRegV1, chunkPart1RegV, chunkPart2RegV);
+        /* Reduce 128-bit value to 32/64-bit value */
+        if (elementByteSize == 1) {
+            masm.neon.umaxvSV(ASIMDSize.FullReg, ElementSize.Word, tmpRegV1, tmpRegV1);
```
teshull, line 283: On the neoverse timing charts I believe the Word variant is faster then the Byte variant

compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayIndexOfOp.java
SwapnilGaikwad: Oh yes, 48 turned out to be a sweet spot to switch to chunk-based approach in micro-benchmarking on N1 (about 4.5% improvement relative to the time it took with size 32). Other systems that I tested, including A72, also showed  improvements (relatively small though, less than 2% relative to time for 32) for longer strings. However, there is no single threshold that I found suitable across AArch64 machines to switch from char-by-char to chunk-based search. The thresholds mostly varied between 48 and 64. 
Well, micro-benchmarking numbers may not be enough to rely upon. I used them as a hint that the higher threshold of 48 would compensate for additional complexity of SIMD instructions and its approach of detecting a match. I'm fine with the threshold set to 32 as well.

SwapnilGaikwad: Let's keep it as 32. The difference between 32 and 48 as thresholds is now less than the noise (~1.6% with 32 being faster) on the synthetic microbenchmark. 
The newer version is now 66% faster relative to master on N1.

Issue #3707, opened 2021-08-23
Module `com.oracle.truffle.regex` does not read module `java.logging`


boris-petrov:
**Describe GraalVM and your environment :**
 - GraalVM version or commit id if built from source: **21.2.0**
 - CE or EE: **CE**
 - JDK version: **JDK 16**
 - OS and OS Version: **Arch Linux 5.13.12-arch1-1**
 - Architecture: **amd64**
 - The output of `java -Xinternalversion`:
```
OpenJDK 64-Bit Server VM (16.0.2+7) for linux-amd64 JRE (16.0.2+7), built on Jul 22 2021 03:33:00 by "builduser" with gcc 11.1.0
```

**Have you verified this issue still happens when using the latest snapshot?**
No.

**Describe the issue**
Putting Graal's JARs on the module-path and running a JS app that uses regexes blows up with:

```
Exception in thread "main" org.graalvm.polyglot.PolyglotException: java.lang.IllegalAccessError: class com.oracle.truffle.regex.tregex.TRegexCompiler (in module com.oracle.truffle.regex) cannot access class java.util.logging.Level (in module java.logging) because module com.oracle.truffle.regex does not read module java.logging
        at com.oracle.truffle.regex/com.oracle.truffle.regex.tregex.TRegexCompiler.shouldLogCompilationTime(TRegexCompiler.java:121)
        at com.oracle.truffle.regex/com.oracle.truffle.regex.tregex.TRegexCompiler.compile(TRegexCompiler.java:72)
        at com.oracle.truffle.regex/com.oracle.truffle.regex.RegexLanguage.createRegexObject(RegexLanguage.java:186)
        at com.oracle.truffle.regex/com.oracle.truffle.regex.RegexLanguage.parse(RegexLanguage.java:149)
        at org.graalvm.truffle/com.oracle.truffle.api.TruffleLanguage$ParsingRequest.parse(TruffleLanguage.java:863)
        at org.graalvm.truffle/com.oracle.truffle.api.TruffleLanguage.parse(TruffleLanguage.java:1517)
        at org.graalvm.truffle/com.oracle.truffle.api.LanguageAccessor$LanguageImpl.parse(LanguageAccessor.java:311)
        at org.graalvm.truffle/com.oracle.truffle.polyglot.PolyglotSourceCache.parseImpl(PolyglotSourceCache.java:94)
[...]

PR #3708, opened 2021-08-23
Require `java.logging` from `com.oracle.truffle.regex`

boris-petrov:
Fixes https://github.com/oracle/graal/issues/3707


chumer (approved):
Looks good will get this through the internal gate and file an internal issue for testing this.

PR #3734, opened 2021-08-27
Use `@argFiles` to prevent issues with long command line invocations

### LINKED ISSUE #3696: [native-image] Support for `@argfile` CLI option
As seen in issues https://github.com/oracle/graal/issues/2995 and https://github.com/graalvm/native-build-tools/issues/85, (mostly Windows) users are facing with issues when their `native-image` CLI invocation exceeds maximum allowed length. This is mostly due to large classpaths with long/absolute addresses.

In Windows this originates from a [documented Windows limitation](https://docs.microsoft.com/en-US/troubleshoot/windows-client/shell-experience/command-line-string-limitation).

There are [several possible workarounds](https://github.com/graalvm/native-build-tools ) that are commonly used for `java`/`javac` invocations:
* Shortening paths by using relative addresses and/or by moving dependencies to a known short path - `C:\\TMP` for example
* Using `fat jars`. This is approach that we are [currently pursuing](https://github.com/graalvm/native-build-tools/pull/126) for [native-build-tools](https://github.com/graalvm/native-build-tools ) as a stopgap solution for the `0.9.4` release. 
* Setting Classpath using a bootstrap jar with a custom manifest that contains the entire classpath. This is approach that we explored first, but it proved ineffective, since @melix discovered that the Native Image driver would extract the classpath from said jar and invoke the image builder through CLI (thus resulting in same error message).
* [Using `@argfile` option](https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javac.html#commandlineargfile). This option was introduced to `javac` executable in JDK7 and to the `java` executable [in JDK9](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8027634), and it is the recommended way to pass long CLI invocations. ATM we don't support this option.

Last two points are of importance for this proposal. In order to fully resolve this issue we must ensure that:
* Driver passes the command line options to the image builder using argument file instead of CLI.
* We expose the `@argfile` configuration option to the end user through `native-image` tool.
* Ensure that parameter parsing works the same way - `--verbose` should have the same output when invoked using `@optionFile` as when invoked using a standard CLI.

lazar-mitrovic:
This PR introduces 2 changes to the Native Image Driver (and its argument parsing mechanism):
1) Adds support for `@argFile` command line option to the `native-image` tool. This command is implemented in the [same way as in `java` tool](https://docs.oracle.com/en/java/javase/14/docs/specs/man/java.html#java-command-line-argument-files) (on JDK9+).
2) Configures Native Image driver to pass all of the JVM arguments through argument file when invoking `NativeImageGeneratorRunner` on JDK9+. This enables usage of `classpath jar` workaround when user encounters a problem with long command line invocation. _IMO this change is a bugfix, and should as such be backported to all of the supported versions_.

Closes #3696.

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/resources/Help.txt
+++ b/substratevm/src/com.oracle.svm.driver/resources/Help.txt
@@ -7,7 +7,10 @@ Usage: native-image [options] class [imagename] [options]
            (to build an image for a class)
    or  native-image [options] -jar jarfile [imagename] [options]
            (to build an image for a jar file)
+
 where options include:
+
+    @argument files       one or more argument files containing options
     -cp <class search path of directories and zip/jar files>
     -classpath <class search path of directories and zip/jar files>
     --class-path <class search path of directories and zip/jar files>
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -1372,21 +1372,24 @@ public class NativeImage {
         return result;
     }
 
+    protected static String createVMInvocationArgumentFile(List<String> arguments) {
+        try {
+            Path argsFile = Files.createTempFile("vminvocation", ".args");
+            String joinedOptions = String.join("\n", arguments);
+            Files.write(argsFile, joinedOptions.getBytes());
+            argsFile.toFile().deleteOnExit();
+            return "@" + argsFile;
+        } catch (IOException e) {
+            throw showError(e.getMessage());
+        }
+    }
+
     protected static String createImageBuilderArgumentFile(List<String> imageBuilderArguments) {
         try {
-            Path argsFile = Files.createTempFile("native-image", "args");
+            Path argsFile = Files.createTempFile("native-image", ".args");
             String joinedOptions = String.join("\0", imageBuilderArguments);
             Files.write(argsFile, joinedOptions.getBytes());
-            Runtime.getRuntime().addShutdownHook(new Thread() {
-                @Override
-                public void run() {
-                    try {
-                        Files.delete(argsFile);
-                    } catch (IOException e) {
-                        System.err.println("Failed to delete temporary image builder arguments file: " + argsFile.toString());
-                    }
-                }
-            });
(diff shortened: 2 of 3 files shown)


substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1373,9 +1373,30 @@ private void addTargetArguments() {
         return result;
     }
 
+    protected static String createVMInvocationArgumentFile(List<String> arguments) {
+        try {
+            Path argsFile = Files.createTempFile("vminvocation", ".args");
+            String joinedOptions = String.join("\n", arguments);
+            Files.write(argsFile, joinedOptions.getBytes());
+            Runtime.getRuntime().addShutdownHook(new Thread() {
```
olpaw, line 1381: Use `argsFile.toFile().deleteOnExit()` for more compactness.

olpaw (changes requested):
Looks good overall. Small cosmetic changes requested.

olpaw:
- [ ] Please test that this change works transitively with `native-image.properties` ?
I.e. native-image -cp foo.jar where foo.jar contains a META-INF/native-image/foo/native-image.properties file that contains
Args = @my_at_file

substratevm/src/com.oracle.svm.driver/resources/Help.txt
```diff
@@ -7,7 +7,10 @@ Usage: native-image [options] class [imagename] [options]
            (to build an image for a class)
    or  native-image [options] -jar jarfile [imagename] [options]
            (to build an image for a jar file)
+
 where options include:
+
+    @<argumentFile>       argument file containing options
```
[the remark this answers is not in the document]
lazar-mitrovic: Fixed

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
lazar-mitrovic: Fixed

lazar-mitrovic:
- [x] I can confirm that this change works transitively with `native-image.properties`.

olpaw (approved):

Issue #3743, opened 2021-08-31
Add JFR events for the garbage collector


jiekang:
As title says. It would be nice to have events to cover configuration, pauses, heap and metaspace data.

jiekang:
@christianhaeubl @christianwimmer 

Hi! We'll be looking to add these. However there is one significant blocker at the moment, the location of code:

JFR is in com.oracle.svm.jfr
GC is in com.oracle.svm.core.genscavenge

How do you envision JFR events are added to core functionality, including more than the GC? Do we move the 'native' bits into `com.oracle.svm.core.jfr`? Or maybe expose some kind of listener API for core pieces similar to `com.oracle.svm.core.thread.ThreadListenerSupport`?

christianwimmer:
@jiekang The GC projects (like `genscavenge`) should implement the JFR support directly without any further indirection. I think this can be achieved fairly easily by splitting the current `com.oracle.svm.jfr` project into a `com.oracle.svm.core.jfr` and `com.oracle.svm.hosted.jfr` part. The `core` part depends only on the `svm.core` project, and then `genscavenge` can depend on `svm.core.jfr`.

Looking at the source code (but without doing any refactoring myself), I think there are very few parts that need to be in `svm.hosted.jfr`. I see `JfrFeature` and `JfrEventSubstitution` have imports from `svm.hosted`, but nothing else.

fernando-valdez:
Native JFR event support changes merged from this PR https://github.com/oracle/graal/pull/3770

PR #3746, opened 2021-09-01
Refactor Jfr into core and hosted

zhengyu123:
JFR functionality should be available in core for implementing VM specific events, such as GC events (see https://github.com/oracle/graal/issues/3743 for details).

This patch splits current com.oracle.svm.jfr into com.oracle.svm.core.jfr and com.oracle.svm.hosted.jfr as @christianwimmer suggested.


christianhaeubl:
The basic change looks good but you also need to change the `suite.py` file (the projects are defined there). Please also make sure that you use the latest sources as there seem to be conflicts between master and this PR.

substratevm/mx.substratevm/suite.py
```diff
@@ -326,6 +326,43 @@
             "workingSets": "SVM",
         },
 
+        "com.oracle.svm.core.jfr": {
```
jiekang, line 329: Should there be an entry for `com.oracle.svm.hosted.jfr as well? (maybe similar to the entry for `com.oracle.svm.hosted.jdk11`)

substratevm/mx.substratevm/suite.py
```diff
@@ -1062,7 +1059,6 @@
                 "com.oracle.svm.core.genscavenge",
                 "com.oracle.svm.core.containers",
                 "com.oracle.svm.jni",
-                "com.oracle.svm.jfr",
```
jiekang, line 1065: Also here, should there be two additional entries for `core.jfr` and `hosted.jfr`?

substratevm/mx.substratevm/suite.py
zhengyu123: I am not sure here.  Adding com.oracle.svm.core.jfr entry here, because it has extra dependencies, etc.

I believe that com.oracle.svm.hosted.jfr is different from com.oracle.svm.hosted.jdk11. com.oracle.svm.hosted.jdk11 has its own directory. Also, I don't see entries for sub-packages of com.oracle.svm.hosted, e.g. com.oracle.svm.hosted.jdk, etc.
 
Honestly, I don't quite understand how suite.py works. I would appreciate if you can point me any doc/wiki, etc.

christianhaeubl: The documentation is located at https://github.com/graalvm/mx

You need to do the following steps:
- Add a project `com.oracle.svm.core.jfr` (analogue to `com.oracle.svm.jfr`) but with only one dependency on `com.oracle.svm.core`
- Add a project `com.oracle.svm.hosted.jfr` (analogue to `com.oracle.svm.jfr`) but with only one dependency on `com.oracle.svm.hosted`
- Add `com.oracle.svm.core.jfr` as a dependency to `com.oracle.svm.core.genscavenge`
- Add the projects that you added to the distribution `SVM`
- Remove the project `com.oracle.svm.jfr`
- Do `mx clean` and `mx ideclean`
- Test everything with `mx build` and `mx ideinit`. It should still be possible to build everything with `mx build` and the IDE should be happy as well (once you import the new project files generated by `mx ideinit`).

zhengyu123: @christianhaeubl Thanks for the instructions. 

Now, I am running into Java compliance issue, as com.oracle.svm.core.genscavenge has dependence on com.oracle.svm.core.jfr with higher Java compliance.

Should I move com.oracle.svm.core.jfr  under com.oracle.svm.core.jdk11 and create interface, e.g. JfrSupport to workaround jdk8?

Thanks.

christianhaeubl: I see, then the only option is to create a proper dummy implementation for JDK8. We would need at least 4 JFR projects then:
- `com.oracle.svm.core.jfr` (common infrastructure and interfaces)
- `com.oracle.svm.core.jfr.jdk8` (a dummy implementation that always acts as if JFR support is disabled)
- `com.oracle.svm.core.jfr.jdk11` (the current implementation from `com.oracle.svm.jfr`)
- `com.oracle.svm.hosted.jfr` (assuming that there is nothing JDK-specific in those classes)

`com.oracle.svm.core.genscavenge` would then only have a dependency on the JDK-independent part, i.e., `com.oracle.svm.core.jfr`.

zhengyu123:
Close in favor of [PR#3770](https://github.com/oracle/graal/pull/3770)

PR #3749, opened 2021-09-01
Process custom image classpath after parsing options

### LINKED ISSUE #3742: --exclude-config not working when provided after -jar
~`exclude-config` is not working when the metadata to be excluded is within classpath jar files referenced by the runnable jar (e.g. via the manifest). In this case, the exclusions are queried before `exclude-config` parameter has been parsed.~

Update: the real issue is that the order of of `--exclude-config` and `-jar` can cause exclude-config not to be applied. Checks for exclusion happen while processing `-jar`, so if `--exclude-config` is provided after that, it will be ignored. A temporary workaround is to provide `--exclude-config` before `-jar`, but a better solution is needed that is not relying on the order in which parameters are given.

Issue is present in the latest GraalVM CE release, 21.2.0.

Please assign this to me.

galderz:
Closes #3742

* Custom image classpath processing can be impacted by exclude-config.
* Moved the code out of -jar parsing, to avoid the need for specific parameter order.
* Without this change, --exclude-config needs to be passed before -jar for it to have an effect.
* Added some verbose messages for easier debugging of native-image binary.

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/DefaultOptionHandler.java
@@ -154,7 +154,9 @@ class DefaultOptionHandler extends NativeImage.OptionHandler<NativeImage> {
                 if (jarFilePathStr == null) {
                     NativeImage.showError(requireValidJarFileMessage);
                 }
-                handleJarFileArg(nativeImage.canonicalize(Paths.get(jarFilePathStr)));
+                final Path filePath = nativeImage.canonicalize(Paths.get(jarFilePathStr));
+                nativeImage.setJarFilePath(filePath);
+                handleJarFileArg(filePath);
                 nativeImage.setJarOptionMode(true);
                 return true;
             case verboseOption:
@@ -320,7 +322,6 @@ class DefaultOptionHandler extends NativeImage.OptionHandler<NativeImage> {
         if (!NativeImage.processJarManifestMainAttributes(filePath, nativeImage::handleMainClassAttribute)) {
             NativeImage.showError("No manifest in " + filePath);
         }
-        nativeImage.addCustomImageClasspath(filePath);
     }
 
     @Override
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -253,6 +253,7 @@ public class NativeImage {
     private boolean diagnostics = false;
     String diagnosticsDir;
     private boolean jarOptionMode = false;
+    private Path jarFilePath;
     private boolean moduleOptionMode = false;
     private boolean dryRun = false;
     private String printFlagsOptionQuery = null;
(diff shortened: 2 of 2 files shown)


mcraj017:
@galderz thanks for your contribution, I have assigned a reviewer to this PR

olpaw:
Hi @galderz ,

the fact that  `--exclude-config` needs to be passed before `-jar` is not a bug. It's well documented behaviour:

- See https://www.graalvm.org/reference-manual/native-image/Options _The options passed to native-image are evaluated left-to-right. For more information, see Native Image Build Configuration._
- and https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#order-of-arguments-evaluation

> The arguments passed to `native-image` are evaluated left-to-right. This also
> extends to arguments that get passed indirectly via `META-INF/native-image`
> based native image configuration. Suppose you have a JAR file that contains
> _native-image.properties_ with `Args = -H:Optimize=0`. Then by using the
> `-H:Optimize=2` option after `-cp <jar-file>` you can override the setting that
> comes from the JAR file.

We cannot modify the behaviour for `-jar` since this would lead to surprising behaviour that conflicts with our left-to-right evaluation policy. Suppose you use `-jar foo.jar` with a `foo.jar` that has a `META-INF/native-image/native-image.properties` embedded that contains
```
Args = -O0 -H:Name=customImageName
```
Due to our left-to-right evaluation users are able to override the args coming from that `native-image.properties` file.
Using `native-image -jar foo.jar -O1 -H:Name=lastArgumentWinsImageName` works as expected.

Your change would break that behaviour.

olpaw (commented):
See https://github.com/oracle/graal/pull/3749#issuecomment-913479146

mcraj017:
Closing this PR, If still there is any change is required, please raise a new PR

galderz:
@olpaw Thanks for the clarification. Are there any tests that verify the behaviour you mention?

olpaw:
There are not explicit test of this. But our whole way of building images (especially with mx) relies on that.

If you run e.g. `mx --extra-image-builder-argument=--verbose --env ce-complete build |& build.log` you can observe many instances of build options being overruled by subsequent options (based on the behaviour described in https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#order-of-arguments-evaluation).  Even the fact that `--extra-image-builder-argument` is useful at all relies on this behaviour.

Also the `native-image.properties` used around building truffle images relies heavily on this. I.e. overriding the default  `-Dtruffle.TruffleRuntime=com.oracle.truffle.api.impl.DefaultTruffleRuntime` with `-Dtruffle.TruffleRuntime=com.oracle.svm.truffle.api.SubstrateTruffleRuntime` in `lib/svm/macros/truffle/native-image.properties` but being able to override that again on command line with placing `-Dtruffle.TruffleRuntime=com.oracle.truffle.api.impl.DefaultTruffleRuntime` (after the truffle options) for building interpreter-only images.

sdeleuze:
@olpaw On Spring side, we need to use `--exclude-config` configured via `native-image.properties` in the classpath. Due to the ordering behavior, it seems it is not possible. Could you confirm?

If confirmed, I understand we can't change the global behavior of "ordering matters", but could we at least work on a fix for this specific issue to allow `--exclude-config` after `-cp` or `-jar` ?

cc @gradinac

olpaw:
> @olpaw On Spring side, we need to use --exclude-config configured via native-image.properties in the classpath. Due to the ordering behavior, it seems it is not possible. Could you confirm?

@sdeleuze, how about creating a subdir that contains `native-image.properties` and is put on the classpath **as the first entry**. I.e.
```
mkdir -p config-only-cp-entry/META-INF/native-image`
echo '--exclude-config ....' >  config-only-cp-entry/META-INF/native-image/native-image.properties
```
then use
```
-cp config-only-cp-entry:<other-classpath-entries ...>
```
Since config-only-cp-entry is the first classpath entry it's `native-image.properties` will be processed before all others.

olpaw:
> -cp config-only-cp-entry:<other-classpath-entries ...>

Possible also good to know that `native-image` (other than `java`) allow you to use `-cp` and `-jar` in the same command line invocation (or multiple `-cp` args in in the same invocation). Because of that you could also use:
```
-cp config-only-cp-entry -cp <other-classpath-entries ...>
```

olpaw:
@sdeleuze did the approach I suggested in https://github.com/oracle/graal/pull/3749#issuecomment-1008821958 work for you?

olpaw:
@galderz we could make `--exclude-config` order independent if we disallow the option to be used anywhere else then command line (i.e. embedding `--exclude-config` in `native-image.properties` in `META-INF/native-image` would be prohibited). 

Would this limitation be of with you? cc @sdeleuze

galderz:
@olpaw It's been a while since I worked on this, but the change here was no longer necessary once we understood the reason for the limitations. IIRC ordering parameters the expected way was enough.

sdeleuze:
@olpaw I think for now the plan is to workaround that at Native Build Tools level, see https://github.com/graalvm/native-build-tools/issues/185.

olpaw:
@sdeleuze FYI I fixed this on master last week (so it will in 22.2). Now if you use `--exclude-config ...` on the native-image command line it will affect all classpath & module-path entries used on the same command line **independent from having specified  `--exclude-config ...` before or after `-cp/p` arguments**. See https://github.com/oracle/graal/commit/006c61dccc163c1c95db47339a33464c191cc50f

sdeleuze:
@olpaw Awesome thanks a lot!

PR #3760, opened 2021-09-08
[GR-35887] Register serialization with associated classes

ziyilin:
When a class is registered as serialization target, all its associated
classes are registered as well.
The associated classes are:
 - Boxed type for primitive type
 - All super classes that implements Serializable interface
 - All fields' classes

Classes shall be excluded from registering:
 - Not effective final classes. It could be the given class' any subclass
   at runtime, so only register the base class is not sufficient. But to
   automatically explore all subclasses and their associated classes could be
   expensive. So such classes should be registered via configurations.
 - Class with "writeObject" method. We can't automatically find out what
   the customized serialization method does.

### DIFF
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeSerialization.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeSerialization.java
@@ -56,6 +56,27 @@ import org.graalvm.nativeimage.impl.RuntimeSerializationSupport;
 public final class RuntimeSerialization {
 
     /**
+     * Register the specified serialization target class itself and all associated classes.
+     * <p>
+     * According to the Java Object Serialization Specification, the associated classes include 1)
+     * all the target class' non-static and non-transient fields types and their associated classes;
+     * 2) other fields defined in the customised writeObject(ObjectOutputStream) and
+     * readObject(ObjectInputStream). This method can automatically explore all possible
+     * serialization target classes in the first scenario, but can't figure out the classes in the
+     * second scenario.
+     * <p>
+     * Another limitation is the specified {@code clazz} must have no subclasses (effectively
+     * final). Otherwise, the actual serialization target class could be any subclass of the
+     * specified class at runtime.
+     *
+     * @param clazz the serialization target class
+     * @Since 23.0
+     */
+    public static void registerIncludingAssociatedClasses(Class<?> clazz) {
+        ImageSingletons.lookup(RuntimeSerializationSupport.class).registerIncludingAssociatedClasses(ConfigurationCondition.alwaysTrue(), clazz);
+    }
+
+    /**
      * Makes the provided classes available for serialization at runtime.
      *
      * @since 21.3
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/RuntimeSerializationSupport.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/RuntimeSerializationSupport.java
@@ -42,6 +42,8 @@ package org.graalvm.nativeimage.impl;
 
 public interface RuntimeSerializationSupport {
 
+    void registerIncludingAssociatedClasses(ConfigurationCondition condition, Class<?> clazz);
+
     void register(ConfigurationCondition condition, Class<?>... classes);
 
     void registerWithTargetConstructorClass(ConfigurationCondition condition, Class<?> clazz, Class<?> customTargetConstructorClazz);
--- a/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
+++ b/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
@@ -28,18 +28,25 @@ package com.oracle.svm.reflect.serialize.hosted;
 // Checkstyle: allow reflection
 
 import static com.oracle.svm.reflect.serialize.hosted.SerializationFeature.println;
+import static com.oracle.svm.reflect.serialize.hosted.SerializationFeature.warn;
 
 import java.io.Externalizable;
+import java.io.ObjectOutputStream;
 import java.io.ObjectStreamClass;
+import java.io.ObjectStreamField;
 import java.io.Serializable;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
+import jdk.vm.ci.meta.JavaKind;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.hosted.Feature;
 import org.graalvm.nativeimage.hosted.RuntimeReflection;
@@ -83,7 +90,7 @@ public class SerializationFeature implements Feature {
         ImageClassLoader imageClassLoader = access.getImageClassLoader();
         ConfigurationTypeResolver typeResolver = new ConfigurationTypeResolver("serialization configuration", imageClassLoader, NativeImageOptions.AllowIncompleteClasspath.getValue());
         SerializationDenyRegistry serializationDenyRegistry = new SerializationDenyRegistry(typeResolver);
-        serializationBuilder = new SerializationBuilder(serializationDenyRegistry, typeResolver);
+        serializationBuilder = new SerializationBuilder(serializationDenyRegistry, access, typeResolver);
         ImageSingletons.add(RuntimeSerializationSupport.class, serializationBuilder);
 
         SerializationConfigurationParser denyCollectorParser = new SerializationConfigurationParser(serializationDenyRegistry, ConfigurationFiles.Options.StrictConfiguration.getValue());
@@ -127,6 +134,12 @@ public class SerializationFeature implements Feature {
         System.out.println(str);
         // Checkstyle: resume
     }
+
+    static void warn(String str){
+        // Checkstyle: stop
+        System.err.println("Warning:" + str);
+        // Checkstyle: resume
+    }
 }
 
 final class SerializationDenyRegistry implements RuntimeSerializationSupport {
@@ -136,6 +149,15 @@ final class SerializationDenyRegistry implements RuntimeSerializationSupport {
 
     SerializationDenyRegistry(ConfigurationTypeResolver typeResolver) {
         this.typeResolver = typeResolver;
+    }
+
+    /**
+     * No need to deny all associated classes, only the specified class itself is registered as
+     * denied.
+     */
+    @Override
+    public void registerIncludingAssociatedClasses(ConfigurationCondition condition, Class<?> clazz) {
+        register(condition, clazz);
     }
 
     @Override
@@ -173,14 +195,21 @@ final class SerializationBuilder extends ConditionalConfigurationRegistry implem
     private static final Method getExternalizableConstructorMethod = ReflectionUtil.lookupMethod(ObjectStreamClass.class, "getExternalizableConstructor", Class.class);
 
     private final Constructor<?> stubConstructor;
+    private final Field descField;
+    private final Method getDataLayoutMethod;
 
     private final SerializationSupport serializationSupport;
     private final SerializationDenyRegistry denyRegistry;
     private final ConfigurationTypeResolver typeResolver;
-
+    private final FeatureImpl.DuringSetupAccessImpl access;
     private boolean sealed;
 
-    SerializationBuilder(SerializationDenyRegistry serializationDenyRegistry, ConfigurationTypeResolver typeResolver) {
+
+    SerializationBuilder(SerializationDenyRegistry serializationDenyRegistry, FeatureImpl.DuringSetupAccessImpl access, ConfigurationTypeResolver typeResolver) {
+        this.access = access;
+        Class<?> classDataSlotClazz = access.findClassByName("java.io.ObjectStreamClass$ClassDataSlot");
+        descField = ReflectionUtil.lookupField(classDataSlotClazz, "desc");
+        getDataLayoutMethod = ReflectionUtil.lookupMethod(ObjectStreamClass.class, "getClassDataLayout");
         stubConstructor = newConstructorForSerialization(SerializationSupport.StubForAbstractClass.class, null);
         this.denyRegistry = serializationDenyRegistry;
         this.typeResolver = typeResolver;
@@ -191,6 +220,63 @@ final class SerializationBuilder extends ConditionalConfigurationRegistry implem
 
     private void abortIfSealed() {
         UserError.guarantee(!sealed, "Too late to add classes for serialization. Registration must happen in a Feature before the analysis has finished.");
+    }
+
+    @Override
+    public void registerIncludingAssociatedClasses(ConfigurationCondition condition, Class<?> clazz) {
+        registerIncludingAssociatedClasses(condition, clazz, new HashSet<>());
+    }
+
+    private void registerIncludingAssociatedClasses(ConfigurationCondition condition, Class<?> clazz, Set<Class<?>> alreadyVisited) {
+        if (alreadyVisited.contains(clazz)) {
+            return;
+        }
+        alreadyVisited.add(clazz);
+        String targetClassName = clazz.getName();
+        // If the serialization target is primitive, it needs to get boxed, because the target is
+        // always an Object.
+        if (clazz.isPrimitive()) {
+            Class<?> boxedType = JavaKind.fromJavaClass(clazz).toBoxedJavaClass();
+            registerIncludingAssociatedClasses(condition, boxedType, alreadyVisited);
+            return;
+        } else if (!Serializable.class.isAssignableFrom(clazz)) {
+            warn("Class " + targetClassName + " does not implement java.io.Serializable and was not registered for object serialization.\n");
+            return;
+        } else if (access.findSubclasses(clazz).size() > 1) {
+            // The classes returned from access.findSubclasses API including the base class itself
+            warn("Class " + targetClassName +
+                    " has subclasses. No classes were registered for object serialization.\n");
+            return;
+        }
+        try {
+            clazz.getDeclaredMethod("writeObject", ObjectOutputStream.class);
+            warn("Class " + targetClassName +
+                    " implements its own writeObject method for object serialization. Any serialization types it uses need to be explicitly registered.\n");
+            return;
+        } catch (NoSuchMethodException e) {
+            // Expected case. Do nothing
+        }
+        register(condition, clazz);
+
+        if (clazz.isArray()) {
+            registerIncludingAssociatedClasses(condition, clazz.getComponentType(), alreadyVisited);
+            return;
+        }
+        ObjectStreamClass osc = ObjectStreamClass.lookup(clazz);
+        try {
+            for (Object o : (Object[]) getDataLayoutMethod.invoke(osc)) {
+                ObjectStreamClass desc = (ObjectStreamClass) descField.get(o);
+                if (!desc.equals(osc) && !desc.equals(clazz)) {
+                    registerIncludingAssociatedClasses(condition, desc.forClass(), alreadyVisited);
+                }
+            }
+        } catch (ReflectiveOperationException e) {
+            VMError.shouldNotReachHere("Cannot register serialization classes due to", e);
+        }
+
+        for (ObjectStreamField field : osc.getFields()) {
+            registerIncludingAssociatedClasses(condition, field.getType(), alreadyVisited);
+        }
     }
 
     @Override
--- a/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/SerializationConfiguration.java
+++ b/substratevm/src/com.oracle.svm.configure/src/com/oracle/svm/configure/config/SerializationConfiguration.java
@@ -73,6 +73,11 @@ public class SerializationConfiguration implements ConfigurationBase, RuntimeSer
     }
 
     @Override
+    public void registerIncludingAssociatedClasses(ConfigurationCondition condition, Class<?> clazz) {
+        register(condition, clazz);
+    }
+
+    @Override
     public void register(ConfigurationCondition condition, Class<?>... classes) {
         for (Class<?> clazz : classes) {
             registerWithTargetConstructorClass(condition, clazz, null);
--- a/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/SerializationRegistrationTest.java
+++ b/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/SerializationRegistrationTest.java
@@ -68,32 +68,89 @@ public class SerializationRegistrationTest {
         }
     }
 
+    public static class AssociatedRegistrationTestClass implements Serializable {
+        private static final long serialVersionUID = 8498517244399174914L;
+        private int intVal;
+        private long[] longVals;
+
+        public AssociatedRegistrationTestClass(int intVal, long[] longVals) {
+            this.intVal = intVal;
+            this.longVals = longVals;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            AssociatedRegistrationTestClass that = (AssociatedRegistrationTestClass) o;
+            if (intVal != that.intVal) {
+                return false;
+            }
+            if (longVals.length != that.longVals.length) {
+                return false;
+            }
+            for (int i = 0; i < longVals.length; i++) {
+                if (longVals[i] != that.longVals[i]) {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(intVal, longVals);
+        }
+    }
+
     private static final byte[] serializedObject;
     private static final List<SerializableTestClass> list;
+    private static AssociatedRegistrationTestClass testClass;
+    private static final byte[] serializedAssociatedRegistrationTestClass;
 
     static {
         list = new ArrayList<>();
         list.add(new SerializableTestClass("Dummy"));
         list.add(new SerializableTestClass("Test"));
+        serializedObject = serializeObject(list);
 
+        testClass = new AssociatedRegistrationTestClass(101, new long[]{1L, 2L, 123L});
+        serializedAssociatedRegistrationTestClass = serializeObject(testClass);
+    }
+
+    private static byte[] serializeObject(Object obj) {
         ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
         try {
             ObjectOutputStream objectOutputStream = new ObjectOutputStream(byteArrayOutputStream);
-            objectOutputStream.writeObject(list);
+            objectOutputStream.writeObject(obj);
             objectOutputStream.flush();
         } catch (IOException e) {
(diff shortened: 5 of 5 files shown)


sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeSerialization.java
```diff
@@ -54,6 +54,10 @@
 @Platforms(Platform.HOSTED_ONLY.class)
 public final class RuntimeSerialization {
 
+    public static void registerAllAssociatedClasses(Class<?> clazz) {
```
zakkak, line 57: Please add Javadoc with `@since` tag.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/RuntimeSerializationSupport.java
```diff
@@ -41,6 +41,23 @@
...
+     * This method can automatically explore all possible serialization target classes
+     * in the first scenario, but can't figure out the classes in the second scenario.
+     * <p>
+     * Another limitation is the specified {@code clazz} must
+     * have no subclasses (effectively final). Otherwise, the actual serialization
+     * target class could be any subclass of the specified class at runtime.
+     *
+     * @param clazz the serialization target class
+     */
+    default void registerAllAssociatedClasses(Class<?> clazz){}
```
zakkak, line 60: Why are you providing an empty default implementation?

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -247,6 +252,92 @@ private void abortIfSealed() {
         UserError.guarantee(!sealed, "Too late to add classes for serialization. Registration must happen in a Feature before the analysis has finished.");
     }
 
+    @Override
+    public void registerAllAssociatedClasses(Class<?> clazz) {
+        if (!serializationSupport.isRegistered(clazz)) {
+            serializationSupport.addRegistered(clazz);
```
zakkak, line 258: I think this should be moved in `register()` to:
1. avoid adding classes that for some reason failed to get registered
2. ensure classes registered directly using `register()` get in the registered-set

zakkak (changes requested):
Thanks for the contribution @ziyilin, please see a couple of comments on how to improve this PR.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/RuntimeSerializationSupport.java
ziyilin: Because for `SerializationDenyRegistry`, it should not register all associated classes as denied. But it should be implemented in the `SerializationDenyRegistry` class. I have fixed this.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeSerialization.java
ziyilin: Since which version? 23.0?

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: This was added for two reasons:
1. Speed up
2. Avoid recursively registering the same class. For example, `int`'s boxed type `Integer`'s corrsonding `ObjectStreamClass` has a `int` field. We are looking into the the target class' `ObjectStreamClass` to register all fields, too. This could lead to recursively registering and stack overflow error.

I have moved this cache check to `register()` and adding a check when looking into ObjectStreamClass' fields.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeSerialization.java
zakkak: I would guess so, depending on when this gets merged as well. Let's see what the GraalVM team says about it. Thanks for addressing this!

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -137,6 +140,16 @@ static void println(String str) {
         this.typeResolver = typeResolver;
     }
 
+    /**
+     * No need to deny all associated classes, only the specified class itself is registered as
+     * denied. This method is the same as
+     * {@link SerializationDenyRegistry#registerAllAssociatedClasses(ConfigurationCondition, Class)}.
```
zakkak, line 146:
```suggestion
     * denied.
```
Seems like a left over from a previous version :)

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/SerializationSupport.java
```diff
@@ -113,10 +115,14 @@ public int hashCode() {
 
     private final Map<SerializationLookupKey, Object> constructorAccessors;
 
+    @Platforms(Platform.HOSTED_ONLY.class)
```
zakkak, line 118:
```suggestion
    @Platforms(Platform.HOSTED_ONLY.class) //
```
To make eclipseformat happy

zakkak (changes requested):
Thanks for addressing my previous comments @ziyilin.
It looks good to me now, except for an out of date javadoc and a styling issue.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: fixed

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/SerializationSupport.java
ziyilin: Fixed, also `run mx eclipseformat` to check other format issues.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -205,10 +219,100 @@ private void abortIfSealed() {
...
+        if (clazz.isArray()) {
+            registerAllAssociatedClasses(condition, clazz.getComponentType());
+            return;
+        }
+        ObjectStreamClass osc = ObjectStreamClass.lookup(clazz);
+        try {
+            Method getDataLayoutMethod = ReflectionUtil.lookupMethod(ObjectStreamClass.class, "getClassDataLayout");
+            Object[] dataLayout = (Object[]) getDataLayoutMethod.invoke(osc);
+            for (int i = 0; i < dataLayout.length; i++) {
+                Class<?> classDataSlotClazz = Class.forName("java.io.ObjectStreamClass$ClassDataSlot");
```
zakkak, line 289: It looks like dynamic class loading is not allowed here (not sure why since this is hosted code though):

```
/home/runner/work/graal/graal/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java:289: do not use dynamic class loading
```
source: https://github.com/oracle/graal/pull/3760/checks#step:9:1769

Judging by https://github.com/oracle/graal/pull/3760/files#diff-e7a9f15b46007ba6515c0e82b0d80801cbfcd01951b2ff13161bd3d8481fa807R200 I think the right way to do this is:
```suggestion
                Class<?> classDataSlotClazz = access.findClassByName("java.io.ObjectStreamClass$ClassDataSlot");
```

zakkak (changes requested):

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: Fixed as suggested

zakkak (approved):
LGTM, Thanks @ziyilin!

ziyilin:
@olpaw Will you have this PR a review please? It has been suspended for a while.

olpaw:
@ziyilin I should be able to take a look this week. Note that nonetheless this will not be in 22.0 as we are already in Truffle/Compiler/Native-Image Feature Freeze for GraalVM 22.0.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -205,10 +219,100 @@ private void abortIfSealed() {
...
+        if (clazz.isArray()) {
+            registerAllAssociatedClasses(condition, clazz.getComponentType());
+            return;
+        }
+        ObjectStreamClass osc = ObjectStreamClass.lookup(clazz);
+        try {
+            Method getDataLayoutMethod = ReflectionUtil.lookupMethod(ObjectStreamClass.class, "getClassDataLayout");
+            Object[] dataLayout = (Object[]) getDataLayoutMethod.invoke(osc);
+            for (int i = 0; i < dataLayout.length; i++) {
+                Class<?> classDataSlotClazz = access.findClassByName("java.io.ObjectStreamClass$ClassDataSlot");
```
olpaw, line 289: These refection lookups need to be hoisted out of the loop.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -205,10 +219,100 @@ private void abortIfSealed() {
...
     }
 
+    @Override
+    public void registerAllAssociatedClasses(ConfigurationCondition condition, Class<?> clazz) {
+        String targetClassName = clazz.getName();
+        // If the serialization target is primitive, it needs to get boxed, because the target is
+        // always an Object.
+        if (clazz.isPrimitive()) {
+            Class<?> boxedType = null;
+            switch (targetClassName) {
```
olpaw, line 229: Please use something like https://github.com/oracle/graal/blob/d08f7ad5797ac68431d8fc92b8ca3d4f6caf724b/truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostExecuteNode.java#L484..L505

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: Fixed

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: I found `jdk.vm.ci.meta.JavaKind` has proper methods to do the work:
https://github.com/graalvm/labs-openjdk-11/blob/master/src/jdk.internal.vm.ci/share/classes/jdk.vm.ci.meta/src/jdk/vm/ci/meta/JavaKind.java#L253-L274
and
https://github.com/graalvm/labs-openjdk-11/blob/master/src/jdk.internal.vm.ci/share/classes/jdk.vm.ci.meta/src/jdk/vm/ci/meta/JavaKind.java#L291-L293

olpaw: Yes that's even better.

olpaw:
Please add a testcase that makes use of this new API under package `com.oracle.svm.test`.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
olpaw: Please make `Field descField` a `private final` field in `SerializationBuilder` and initialize it in the constructor like we do for reflectively accessed methods. No need to get that same field over and over again with each call to `registerAllAssociatedClasses`

ziyilin: Fixed as suggested

ziyilin:
> 

@olpaw A new test case is added in `com.oracle.svm.test.SerializationRegistrationTest`.

olpaw:
> > 
> 
> @olpaw A new test case is added in `com.oracle.svm.test.SerializationRegistrationTest`.

The new tests causes gate style issues:
```
 /home/runner/work/graal/graal/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/SerializationRegistrationTest.java:71: warning: [overrides] Class AssociatedRegistrationTestClass overrides equals, but neither it nor any superclass overrides hashCode method
    public static class AssociatedRegistrationTestClass implements Serializable{
                  ^
```

ziyilin:
@olpaw The gate style issuss are fixed

olpaw:
> @olpaw The gate style issuss are fixed

Thanks. I should be able to look into your PR again tomorrow.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -205,10 +223,69 @@ private void abortIfSealed() {
...
+        }
+        register(condition, clazz);
+
+        if (clazz.isArray()) {
+            registerAllAssociatedClasses(condition, clazz.getComponentType());
+            return;
+        }
+        ObjectStreamClass osc = ObjectStreamClass.lookup(clazz);
+        try {
+            Method getDataLayoutMethod = ReflectionUtil.lookupMethod(ObjectStreamClass.class, "getClassDataLayout");
```
olpaw, line 260: Please also ensure this reflective method gets prepared in the SerializationBuilder constructor as you already did for `java.io.ObjectStreamClass$ClassDataSlot.desc`.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -205,10 +223,69 @@ private void abortIfSealed() {
...
+            Object[] dataLayout = (Object[]) getDataLayoutMethod.invoke(osc);
+            for (int i = 0; i < dataLayout.length; i++) {
```
olpaw, line 262: Replace with enhanced-for loop

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -205,10 +223,69 @@ private void abortIfSealed() {
...
+                if (!desc.equals(osc) && !desc.equals(clazz)) {
+                    registerAllAssociatedClasses(condition, desc.forClass());
+                }
+            }
+        } catch (ReflectiveOperationException e) {
+            VMError.shouldNotReachHere("Cannot register serialization classes due to", e);
+        }
+
+        ObjectStreamField[] fields = osc.getFields();
+        for (int i = 0; i < fields.length; i++) {
```
olpaw, line 273: Replace with enhanced-for loop

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -205,10 +223,69 @@ private void abortIfSealed() {
...
+        try {
+            clazz.getDeclaredMethod("writeObject", ObjectOutputStream.class);
+            println("Warning: The given class " + targetClassName +
+                            " has customized serialization method writeObject. " +
+                            "We can't figure out what should be registered. Skip.\n");
+            return;
+        } catch (NoSuchMethodException e) {
+            // Expected case. Do nothing
+        }
+        register(condition, clazz);
```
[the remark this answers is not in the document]
olpaw: Provide a private `registerAllAssociatedClasses` that takes a `Set<Class<?>>` parameter and make the `public void registerAllAssociatedClasses` call it with `registerAllAssociatedClasses(condition, clazz, new HashSet<>())`.
In the private `registerAllAssociatedClasses` use
```java
    private void registerAllAssociatedClasses0(ConfigurationCondition condition, Class<?> clazz, Set<Class<?>> alreadyVisited) {
        if (alreadyVisited.contains(clazz)) {
            return;
        }
        alreadyVisited.add(clazz);
        ... rest of impl.
    }
```

olpaw (changes requested):
See latest comments.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: Fixed

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: fixed

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: fixed

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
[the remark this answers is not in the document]
ziyilin: We already have a set `SerializationSupport.registeredTargets` to track all registered classes. It is checked in `register` method, so that all serialization target class registrations are checked, not only the ones through `registerAllAssociatedClasses`.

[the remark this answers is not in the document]
olpaw: Yes ... but
you are not using that to prevent redundant reentering of `registerAllAssociatedClasses` for the same class object.

Also note that it's a different thing knowing that you already have a class that was registered with `register` (i.e. having it in `SerializationSupport.registeredTargets` compared to knowing that you already ran `registerAllAssociatedClasses` for a given class. If you would reuse `SerializationSupport.registeredTargets`  for the above it can happen that previously you registered a class with `register` and
```java
if (isRegistered(clazz)) {
    return;
}
```
would prevent running `registerAllAssociatedClasses` although that hasn't happen before. So please implement this as I suggested in https://github.com/oracle/graal/pull/3760#discussion_r770612854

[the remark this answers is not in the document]
ziyilin: I see your point. Fixed as suggested.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -205,10 +227,70 @@ private void abortIfSealed() {
...
+        for (ObjectStreamField field : osc.getFields()) {
+            registerAllAssociatedClasses(condition, field.getType(), alreadyVisited);
+        }
+    }
+
     @Override
     public void register(ConfigurationCondition condition, Class<?>... classes) {
         for (Class<?> clazz : classes) {
-            registerWithTargetConstructorClass(condition, clazz, null);
+            if (!serializationSupport.isRegistered(clazz)) {
```
olpaw, line 290: This is not needed. Please revert with
```patch
--- a/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
+++ b/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
@@ -45,7 +45,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import jdk.vm.ci.meta.JavaKind;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.hosted.Feature;
 import org.graalvm.nativeimage.hosted.RuntimeReflection;
@@ -72,6 +71,8 @@ import com.oracle.svm.reflect.serialize.SerializationRegistry;
 import com.oracle.svm.reflect.serialize.SerializationSupport;
 import com.oracle.svm.util.ReflectionUtil;
 
+
 @AutomaticFeature
 public class SerializationFeature implements Feature {
     private SerializationBuilder serializationBuilder;
@@ -287,10 +288,7 @@ final class SerializationBuilder extends ConditionalConfigurationRegistry implem
     @Override
     public void register(ConfigurationCondition condition, Class<?>... classes) {
         for (Class<?> clazz : classes) {
-            if (!serializationSupport.isRegistered(clazz)) {
-                registerWithTargetConstructorClass(condition, clazz, null);
-                serializationSupport.addRegistered(clazz);
-            }
+            registerWithTargetConstructorClass(condition, clazz, null);
         }
     }
 
```

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/SerializationSupport.java
```diff
@@ -143,4 +149,15 @@ public Object getSerializationConstructorAccessor(Class<?> declaringClass, Class
                             " to serialization-config.json fixes the problem.");
         }
     }
+
+    @Platforms(Platform.HOSTED_ONLY.class)
```
olpaw, line 153: Same here please revert this with:
```patch
@@ -115,14 +111,10 @@ public class SerializationSupport implements SerializationRegistry {
 
     private final Map<SerializationLookupKey, Object> constructorAccessors;
 
-    @Platforms(Platform.HOSTED_ONLY.class) //
-    private final Set<Class<?>> registeredTargets;
-
     @Platforms(Platform.HOSTED_ONLY.class)
     public SerializationSupport(Constructor<?> stubConstructor) {
         constructorAccessors = new ConcurrentHashMap<>();
         this.stubConstructor = stubConstructor;
-        this.registeredTargets = new HashSet<>();
     }
 
     @Platforms(Platform.HOSTED_ONLY.class)
@@ -149,15 +141,4 @@ public class SerializationSupport implements SerializationRegistry {
                             " to serialization-config.json fixes the problem.");
         }
     }
-
-    @Platforms(Platform.HOSTED_ONLY.class)
-    public boolean isRegistered(Class<?> serializationCandidate) {
-        return registeredTargets.contains(serializationCandidate);
-    }
-
-    @Platforms(Platform.HOSTED_ONLY.class)
-    public void addRegistered(Class<?> serializationTarget) {
-        registeredTargets.add(serializationTarget);
-    }
-
 }
```

olpaw (changes requested):

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/SerializationSupport.java
ziyilin: Removed

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: Removed

olpaw (approved):

olpaw:
Running in our CI now ...

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeSerialization.java
```diff
@@ -55,6 +55,27 @@
...
+     * second scenario.
+     * <p>
+     * Another limitation is the specified {@code clazz} must have no subclasses (effectively
+     * final). Otherwise, the actual serialization target class could be any subclass of the
+     * specified class at runtime.
+     *
+     * @param clazz the serialization target class
+     * @Since 23.0
+     */
+    public static void registerAllAssociatedClasses(Class<?> clazz) {
```
peter-hofer, line 75: Can we call this `registerIncludingAssociatedClasses`? The name sounds like it might not register the class itself.

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
```diff
@@ -205,6 +227,63 @@ private void abortIfSealed() {
...
+            return;
+        }
+        try {
+            clazz.getDeclaredMethod("writeObject", ObjectOutputStream.class);
+            println("Warning: The given class " + targetClassName +
+                            " has customized serialization method writeObject. " +
+                            "We can't figure out what should be registered. Skip.\n");
+            return;
+        } catch (NoSuchMethodException e) {
+            // Expected case. Do nothing
```
peter-hofer, line 262: Aside from the `println` call which prints to `System.out` instead of `System.err` and could be renamed to `warn` as that's what it is used for (but I blame not having a proper facility for that anywhere), the error messages can be improved:
`Class foo.bar does not implement java.io.Serializable and was not registered for object serialization.`
`Class foo.bar has more than one subclass. No subclasses were registered for object serialization.`
`Class foo.bar implements its own writeObject method for object serialization. Any serialization types it uses need to be explicitly registered.`

peter-hofer (changes requested):

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeSerialization.java
ziyilin: 'registerIncludingAssociatedClasses' is too long, what about `registerWithAssociatedClasses`?

substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/serialize/hosted/SerializationFeature.java
ziyilin: Fixed as suggested. But for the subclass case, we set the checking threshold to 1 because the `access.findSubclasses` API returns a list of subclasses including the base class itself. We don't registered the base class and any of its subclasses in this case.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeSerialization.java
olpaw: I don't see that as a problem.
It's exactly as long as `registerWithTargetConstructorClass` a few lines below.

ziyilin: Renamed the method to registerIncludingAssociatedClasses as suggested.

PR #3763, opened 2021-09-08
[GR-33549] Run image-builder on modulepath when module options are used.

olpaw:
@ivan-ristovic please have a look. This will eliminate the need for using `USE_NATIVE_IMAGE_JAVA_PLATFORM_MODULE_SYSTEM=true`.
Whenever native-images gets passed `-p` or `-m` (or the equivalent long forms) the driver will switch to running the image builder on modulepath.  Note that `USE_NATIVE_IMAGE_JAVA_PLATFORM_MODULE_SYSTEM=true` can still be used to force running the image builder on modulepath even when no `-p` or `-m` is used.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -84,7 +84,6 @@ jobs:
               JDK: "labsjdk-ce-11"
               GATE: "hellomodule"
               PRIMARY: "substratevm"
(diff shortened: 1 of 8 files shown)


Jsn2win (commented):
Workflow/main

ivan-ristovic (approved):
LGTM. Tested locally and works as expected.

PR #3776, opened 2021-09-10
[GR-33504] Implement --add-reads for native-image.

graalvmbot:
Similar to add-opens and add-exports that we already have implemented we also need our own implementation of
```
--add-reads module=target-module(,target-module)*
       Updates module to read the target-module, regardless of the module declaration.  target-module can be all unnamed to read all unnamed modules.
```
I addition to requiring this for completeness it will also allow us to get rid of
```
com.oracle.svm.hosted.jdk11.NativeImageClassLoaderSupportJDK11OrLater#adjustLibrarySupportReadAllUnnamed
```

As part of this PR we also need to change the way we get access to our Hosted and RuntimeOptions.

Instead of using imageClassLoader.findSubclasses in HostedOptionParser we need to use ServiceLoader based access (as graal does) to collect the options so that we can run options processing before running
```
com.oracle.svm.hosted.AbstractNativeImageClassLoaderSupport#initAllClasses
```

This is needed so that options that affect initAllClasses can be processed before.


zapster (approved):
I very much like the clean-ups, thanks! LGTM

PR #3791, opened 2021-09-13
[GR-32616]

graalvmbot:


zakkak:
FWIW this PR appears to be a follow up to https://github.com/oracle/graal/pull/3589

PR #3793, opened 2021-09-13
[GR-33796] Non-jlinked-build fixes for breakage after cc11ae1

### LINKED ISSUE #3778: Build break with OpenJDK 11/17
A graal vm master build from source fails with OpenJDK 16 or (via mandrel) for OpenJDK 11. The offending commit seems to be cc11ae1d1fad550424b1fdc0a6cf7e9d6815874a which a) removed some flags configs for older JDKs which breaks the JDK 16 build. I suppose that's OK since JDK 17 GAs this week b) re-ordered `builderJavaArgs` so that `-XX-UseJVMCICompiler` comes first in the native image generator invocation.  That latter parts needs to have `-XX:+UnlockExperimentalVMOptions` as a preceeding arg. Using that, gets us further, but not quite done.

Failure for JDK 16 is:
```
[...]
berOfThreads=8 -H:Path=/disk/graal/upstream-sources/graalvm/sdk/mxbuild/linux-amd64/native-image.image
Apply jar:file:///disk/graal/upstream-sources/graalvm/sdk/mxbuild/linux-amd64/GRAALVM_76BF0784AB_JAVA16_STAGE1/graalvm-76bf0784ab-java16-21.3.0-dev/lib/graalvm/svm-driver.jar!/META-INF/native-image/com.oracle.substratevm/svm-driver/native-image.properties
Error: Image building not supported for Java version 16.0.2 in /disk/graal/upstream-sources/graalvm/sdk/mxbuild/linux-amd64/GRAALVM_76BF0784AB_JAVA16_STAGE1/graalvm-76bf0784ab-java16-21.3.0-dev with VM configuration "OpenJDK 64-Bit Server VM". Update SubstrateCompilerFlagsBuilder.compute_graal_compiler_flags_map() in mx_substratevm.py to add a configuration for a new Java version.

Building native-image.image: Failed due to error: 1
Building native-image.image failed
1 build tasks failed
```

Failure for JDK 17/11 is:

```
[...]
DEBUG [OperatingSystem] Execute [.\mandrel-java17-21.3.0-devc4f6af1e281\lib\svm\bin\native-image.cmd, --macro:native-image-agent-library] in C:\workspace\workspace\mandrel-maste---c3557c9d
Error: VM option 'UseJVMCICompiler' is experimental and must be enabled via -XX:+UnlockExperimentalVMOptions.
Error: The unlock option must precede 'UseJVMCICompiler'.
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
Exception in thread "main" java.lang.RuntimeException: java.lang.RuntimeException: Failed, exit code: 1
    at OperatingSystem.exec(build.java:1733)
    at build.buildAgents(build.java:188)
    at build.main(build.java:170)
Caused by: java.lang.RuntimeException: Failed, exit code: 1
    at OperatingSystem.exec(build.java:1718)
    at build.buildAgents(build.java:188)
    at build.main(build.java:170)
    at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
    at java.base/java.lang.reflect.Method.invoke(Method.java:568)
    at jdk.compiler/com.sun.tools.javac.launcher.Main.execute(Main.java:419)
    at jdk.compiler/com.sun.tools.javac.launcher.Main.run(Main.java:192)
    at jdk.compiler/com.sun.tools.javac.launcher.Main.main(Main.java:132)
```

jerboaa:
Add jdk.scripting.nashorn modules for module look-up on JDK 11.
Target_jdk_nashorn_api_scripting_ClassFilter substitution needs nashorn
target class loaded.

cc11ae1d1fad550424b1fdc0a6cf7e9d6815874a re-ordered arguments passed to
the native image generator. Add -XX:+UnlockExperimentalVMOptions before
actually using any one of those experimental options as that's needed
for some JVMs.

Add relevant modules to the upgraded module path list since for
non-jlinked-builds they are either not there or too old.

Closes #3778

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_jdk_nashorn_api_scripting_NashornScriptEngineFactory.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_jdk_nashorn_api_scripting_NashornScriptEngineFactory.java
@@ -28,9 +28,10 @@ import javax.script.ScriptEngine;
 
 import com.oracle.svm.core.annotate.Substitute;
 import com.oracle.svm.core.annotate.TargetClass;
+import com.oracle.svm.core.jdk.NashornSupport.NashornAvailable;
 import com.oracle.svm.core.util.VMError;
 
-@TargetClass(className = "jdk.nashorn.api.scripting.NashornScriptEngineFactory", onlyWith = JDK14OrEarlier.class)
+@TargetClass(className = "jdk.nashorn.api.scripting.NashornScriptEngineFactory", onlyWith = {JDK14OrEarlier.class, NashornAvailable.class})
 public final class Target_jdk_nashorn_api_scripting_NashornScriptEngineFactory {
     @Substitute
     @SuppressWarnings({"unused", "static-method"})
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -481,12 +481,6 @@ public class NativeImage {
                 }
             }
 
-            if (useJVMCINativeLibrary) {
-                builderJavaArgs.add("-XX:+UseJVMCINativeLibrary");
-            } else {
-                builderJavaArgs.add("-XX:-UseJVMCICompiler");
-            }
-
             String javaVersion = String.valueOf(JavaVersionUtil.JAVA_SPEC);
             String[] flagsForVersion = graalCompilerFlags.get(javaVersion);
             if (flagsForVersion == null) {
@@ -516,6 +510,12 @@ public class NativeImage {
                 }
             }
 
+            if (useJVMCINativeLibrary) {
+                builderJavaArgs.add("-XX:+UseJVMCINativeLibrary");
+            } else {
+                builderJavaArgs.add("-XX:-UseJVMCICompiler");
+            }
+
             return builderJavaArgs;
         }
 
@@ -524,11 +524,12 @@ public class NativeImage {
          */
         public List<Path> getBuilderModulePath() {
             List<Path> result = new ArrayList<>();
+            // Non-jlinked JDKs need truffle and graal-sdk on the module path since they
+            // don't have those modules as part of the JDK.
+            result.addAll(getJars(rootDir.resolve(Paths.get("lib", "jvmci")), "graal-sdk", "enterprise-graal"));
+            result.addAll(getJars(rootDir.resolve(Paths.get("lib", "truffle")), "truffle-api"));
             if (modulePathBuild) {
                 result.addAll(getJars(rootDir.resolve(Paths.get("lib", "svm", "builder"))));
-            } else {
-                result.addAll(getJars(rootDir.resolve(Paths.get("lib", "jvmci")), "graal-sdk", "enterprise-graal"));
-                result.addAll(getJars(rootDir.resolve(Paths.get("lib", "truffle")), "truffle-api"));
             }
             return result;
         }
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NashornSupport.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NashornSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -22,11 +22,30 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+
 package com.oracle.svm.core.jdk;
 
-import com.oracle.svm.core.annotate.TargetClass;
-import com.oracle.svm.core.jdk.NashornSupport.NashornAvailable;
+import java.util.function.BooleanSupplier;
 
-@TargetClass(className = "jdk.nashorn.api.scripting.ClassFilter", onlyWith = {JDK14OrEarlier.class, NashornAvailable.class})
-public final class Target_jdk_nashorn_api_scripting_ClassFilter {
+public final class NashornSupport {
+
+    // Determine if the jdk.scripting.nashorn module (or jar) is available. If it's not,
+    // the boolean supplier should return false.
+    static class NashornAvailable implements BooleanSupplier {
+
+        private static final String CLASSFILTER_NAME = "jdk.nashorn.api.scripting.ClassFilter";
+
+        @Override
+        public boolean getAsBoolean() {
+            try {
+                // Checkstyle: stop
+                Class.forName(CLASSFILTER_NAME);
+                // Checkstyle: resume
+                return true;
+            } catch (ClassNotFoundException e) {
+                return false;
+            }
+        }
+
+    }
 }


sdk/mx.sdk/mx_sdk_vm_impl.py
```diff
@@ -1932,8 +1932,10 @@ def _get_option_vars():
             return ' '.join(image_config.option_vars)
 
         def _get_launcher_args():
-            if not _jlink_libraries() and _src_jdk_version >= 9:
-                return '-J--add-exports=jdk.internal.vm.ci/jdk.vm.ci.code=jdk.internal.vm.compiler'
+            # Nashorn got removed in JDK 15, added exports for the JVMCI compiler got added
+            # with JDK-8223332 (JDK 13) and JDK-8273547 (11.0.14). Use the larger upper bound.
+            if not _jlink_libraries() and _src_jdk_version >= 9 and _src_jdk_version <= 14:
+                return '-J--add-modules=jdk.scripting.nashorn -J--add-exports=jdk.internal.vm.ci/jdk.vm.ci.code=jdk.internal.vm.compiler'
```
olpaw, line 1938: Instead of forcing `jdk.scripting.nashorn` onto the module path we should make sure that the substitutions that fail without are written in a way so that they can handle `jdk.scripting.nashorn` to be missing.

I.e. Define a new `BooleanSupplier` that is only true when module `jdk.scripting.nashorn` is available and add it to the `@TargetClass` annotations attribute value `onlyWith` e.g.
```java
@TargetClass(className = "jdk.nashorn.api.scripting.ClassFilter", onlyWith = {JDK14OrEarlier.class, hasModuleNashorn.class})
public final class Target_jdk_nashorn_api_scripting_ClassFilter {
}
```
and
```java
@TargetClass(className = "jdk.nashorn.api.scripting.NashornScriptEngineFactory", onlyWith = {JDK14OrEarlier.class, hasModuleNashorn.class})
public final class Target_jdk_nashorn_api_scripting_NashornScriptEngineFactory {
...
}
```

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -519,6 +523,44 @@ public Path getBuilderInspectServerPath() {
             return builderJavaArgs;
         }
 
+        private void setHasGraalModules() {
```
olpaw, line 526: I really wanted to get rid of stunts like this in the driver (as you can see in the removing of `getBuilderModuleNames` in https://github.com/oracle/graal/pull/3763).

Do we still need this if we solve the `jdk.scripting.nashorn` problems as suggested above?
Or do you have other means of knowing that this is not our GraalVM (i.e. different value for e.g. `com.oracle.svm.driver.NativeImage#graalvmVersion` or `com.oracle.svm.driver.NativeImage#graalvmConfig`)

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -484,6 +486,8 @@ public Path getBuilderInspectServerPath() {
             if (useJVMCINativeLibrary) {
                 builderJavaArgs.add("-XX:+UseJVMCINativeLibrary");
             } else {
+                // some JVMs need this unlocking
+                builderJavaArgs.add("-XX:+UnlockExperimentalVMOptions");
```
olpaw, line 490: Thanks for fixing this!

olpaw (changes requested):

sdk/mx.sdk/mx_sdk_vm_impl.py
jerboaa: OK, that seems a more elegant way of fixing it. Will fix.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
zakkak: @jerboaa the flag is actually present but not in the right place. With this change we essentially pass the option twice:

```Executing [
/home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_ADB1ABFC21_JAVA11/graalvm-adb1abfc21-java11-21.3.0-dev/bin/java \
-XX:+UnlockExperimentalVMOptions \
-XX:-UseJVMCICompiler \
-XX:+UseParallelGC \
-XX:+UnlockExperimentalVMOptions \
-XX:+EnableJVMCI \
...
```

I suggest applying the following instead:
```diff
diff --git a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
index 7d08509420c..8ef7d3e22dc 100644
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -481,12 +481,6 @@ public class NativeImage {
                 }
             }
 
-            if (useJVMCINativeLibrary) {
-                builderJavaArgs.add("-XX:+UseJVMCINativeLibrary");
-            } else {
-                builderJavaArgs.add("-XX:-UseJVMCICompiler");
-            }
-
             String javaVersion = String.valueOf(JavaVersionUtil.JAVA_SPEC);
             String[] flagsForVersion = graalCompilerFlags.get(javaVersion);
             if (flagsForVersion == null) {
@@ -516,6 +510,12 @@ public class NativeImage {
                 }
             }
 
+            if (useJVMCINativeLibrary) {
+                builderJavaArgs.add("-XX:+UseJVMCINativeLibrary");
+            } else {
+                builderJavaArgs.add("-XX:-UseJVMCICompiler");
+            }
+
             return builderJavaArgs;
         }
```

zakkak (changes requested):

sdk/mx.sdk/mx_sdk_vm_impl.py
jerboaa: Fixed in the updated patch.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
jerboaa: As we discussed on slack, this is no longer needed and we are using extra modules on the module path to get around the missing module error when building ni modules, like the native-image-agent.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
jerboaa: This would be a partial revert of https://github.com/oracle/graal/commit/cc11ae1d1fad550424b1fdc0a6cf7e9d6815874a. Not sure why the order was changed. @olpaw thoughts on that?

jerboaa: Happy to go either way, but I thought the order changes was intentional.

olpaw: There was some reordering required to allow us to switch to running the imagbuilder on module-path when module-related options are passed to the driver. But this particular change should not cause problems as the value of useJVMCINativeLibrary is not dependent on module-related options. Thus @zakkak suggestion should be followed here.

jerboaa: OK, fixed.

jerboaa:
I believe I've fixed the checkstyle warning now. Lets see what the gates tell us. It's now also rebased on latest master, which should fix the gate's failures on debuginfo tests.

olpaw (approved):

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_jdk_nashorn_api_scripting_NashornScriptEngineFactory.java
```diff
@@ -28,9 +28,10 @@
 
 import com.oracle.svm.core.annotate.Substitute;
 import com.oracle.svm.core.annotate.TargetClass;
+import com.oracle.svm.core.jdk.NashornSupport.NashornAvailable;
 import com.oracle.svm.core.util.VMError;
 
-@TargetClass(className = "jdk.nashorn.api.scripting.NashornScriptEngineFactory", onlyWith = JDK14OrEarlier.class)
+@TargetClass(className = "jdk.nashorn.api.scripting.NashornScriptEngineFactory", onlyWith = {JDK14OrEarlier.class, NashornAvailable.class})
```
zakkak, line 34: I fail to see why we need this check since we know that nashorn is available in JDK <= 14 and we already check for that.
What am I missing?

olpaw: As @jerboaa observed, there are JDK <= 14 where module `jdk.scripting.nashorn` is not per-default available as a module (running `java --list-modules` would not show it).

On our side it's not a problem. Our jlinked Java 11 has module `jdk.scripting.nashorn`.

zakkak: > As @jerboaa observed, there are JDK <= 14 where module `jdk.scripting.nashorn` is not per-default available as a module (i.e. running `java --list-modules` would not show it).

Judging by [@jerboaa's initial patch](https://github.com/oracle/graal/commit/5e4b8fa0c90444115930b09c39b1b999cdd5843f#diff-42623de4b000eeb25f4b626861ebe69fd6fbcb7e9480897d1d5c3ff79cb7eedd) it looks like the assumption was that module `jdk.scripting.nashorn` is indeed available in JDK <= 14

The [`--list-modules` part](https://github.com/oracle/graal/commit/5e4b8fa0c90444115930b09c39b1b999cdd5843f#diff-9e053bec9eca0799896bd471363fa6aa1df70269e454b89dff1ea5fb844a480cR540) was only meant to check for `org.graalvm` modules and not for `jdk.scripting.nashorn`.

I suggest we wait for @jerboaa's input on this before pushing it to the merge queue.

olpaw: > assumption was that module jdk.scripting.nashorn is indeed available in JDK <= 14

There is a difference between available and enabled as I tried to explain above.

olpaw: > I suggest we wait for @jerboaa's input on this before pushing it to the merge queue.

No problem.

jerboaa: For that reason it's not sufficient to only go by the JDK versions on the substitutions.

zakkak: OK, I see. I was also able to reproduce the issue with OpenJDK 11.0.12 and 11.0.13. Note, however, that the module appears in `java --list-modules`.

zakkak (approved):

olpaw:
dsd

olpaw:
@jerboaa I accidentally force-pushed to https://github.com/jerboaa/graal.git instead of our internal repo. Please ignore that. I add a comment once your PR ran though our CI.

jerboaa:
> @jerboaa I accidentally force-pushed to https://github.com/jerboaa/graal.git instead of our internal repo grimacing . Please ignore that (and the strange side effect of this PR getting closed). I add a comment once your PR ran though our CI.

OK, thanks.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_jdk_nashorn_api_scripting_NashornScriptEngineFactory.java
olpaw: @zakkak pure modular application (like e.g. `native-image-agent-library`) would need to specify `requires jdk.scripting.nashorn` in their `module-info.java` to have access to it. Otherwise `jdk.scripting.nashorn` is just not available for them.

Thus if you build an image on the module path (i.e. use `-p` `-m` options instead of `-cp` or `-jar` with native-image) it is not a given that `jdk.scripting.nashorn` will be available.

jerboaa:
https://github.com/jerboaa/graal/commit/c17a8f4148efcdd6ba13d3ffb5895ce56e09a28a is the commit which is intended to be pushed to graal master with this PR.

olpaw:
> [jerboaa@c17a8f4](https://github.com/jerboaa/graal/commit/c17a8f4148efcdd6ba13d3ffb5895ce56e09a28a) is the commit which is intended to be pushed to graal master with this PR.

Yes that's the one that is currently running in our CI.

olpaw:
The PR is in the merge queue.

PR #3803, opened 2021-09-16
[GR-32340] Introduce default resource-registration methods.

graalvmbot:


zakkak (approved):
LGMT, thanks @vjovanov

Issue #3852, opened 2021-09-30
Debug Info: ptype showing wrong modifiers?


zakkak:
**Describe the issue**
Using ptype to print information about a Java type/class results in some methods being listed with wrong modifiers. E.g. for `java.lang.Object` we see that `hashCode()` is listed as private.

```
(gdb) ptype 'java.lang.Object'
type = class java.lang.Object : public _objhdr {
  public:
    void Object();
    boolean equals(class java.lang.Object *);
  private:
    int hashCode();
  public:
    class java.lang.String * toString();
    void wait(long, int);
}
```

**Steps to reproduce the issue**
```
cat <<EOF > HelloWorld.java
public class HelloWorld {
    public static void main(String[] args) {
        System.out.println("Hello World!");
    }
}
EOF
javac HelloWorld.java
native-image -g HelloWorld
gdb helloworld
(gdb) ptype 'java.lang.Object'
type = class java.lang.Object : public _objhdr {
  public:
    void Object();
    boolean equals(class java.lang.Object *);
  private:
    int hashCode();
  public:
    class java.lang.String * toString();
    void wait(long, int);
}

**Describe GraalVM and your environment:**
 - GraalVM version: **d465c5885127caef6d551d417e6caef937fe225a**
 - JDK major version: **11**
 - OS: **Fedora 34**
 - Architecture: **AMD64**

cc @adinn

zakkak:
Update: It looks like we are peaking up the modifiers from [the substitute method](https://github.com/oracle/graal/blob/d465c5885127caef6d551d417e6caef937fe225a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java#L92-L96) instead of the original. I've got a fix and working on extending the `debuginfotest` to catch such issues both for methods and fields.

adinn:
Hmm, that's an interesting issue.

When we present the source code in the debugger we will actually be displaying the substituted method code. So, from the point of view of someone stepping through the implementation marking it as private matches the displayed view.

However, from the point of view of someone stepping into a call to the original method from a caller in some other class than the original (or, indeed, substituted) owner it really ought to look like a public method.

I think you are right that it would be clearest if the permissions displayed in the ptype listing reflected the original. A disparity when looking at a subsituted implementation is going to be understandable by virtue of the fact that the displayed method is clearly a replacement. A disparity at the point of call has no accompanying info in the source to explain the disparity.

PR #3853, opened 2021-09-30
Debug info: Get modifiers of original method/field for DW_AT_ACCESSIBILITY

zakkak:
Closes: #3852


zakkak:
@adinn can you please have a look?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -229,6 +230,16 @@ private static ResolvedJavaType getOriginal(HostedType hostedType) {
         return javaType;
     }
 
+    private int getOriginalModifiers(HostedMethod hostedMethod) {
+        ResolvedJavaMethod targetMethod = hostedMethod.getWrapped().getWrapped();
+        if (targetMethod instanceof SubstitutionMethod) {
+            targetMethod = ((SubstitutionMethod) targetMethod).getOriginal();
+        } else if (targetMethod instanceof CustomSubstitutionMethod) {
+            targetMethod = ((CustomSubstitutionMethod) targetMethod).getOriginal();
```
zakkak, line 238: I was not able to test this path with my tests. Recommendations are welcome!

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -608,7 +619,11 @@ public int size() {
 
             @Override
             public int modifiers() {
-                return field.getModifiers();
+                ResolvedJavaField targetField = field.wrapped.wrapped;
+                if (targetField instanceof SubstitutionField) {
+                    targetField = ((SubstitutionField) targetField).getOriginal();
```
zakkak, line 624: I was not able to test this path with my tests. Recommendations are welcome!

adinn: Me neither.

Looking at how AnnotationSubstituionProcessor is implemented it looks like you need both a target class and target field labelled with a Substitute annotation (also, probably a TargetElement annotation on the field). If you try using a Substitute annotation on the field when the class just has a TargetClass annotation then you get an error saying you need to use Inject, Delete or Alias on the field. However, I am not 100% clear what the difference is supposed to be between labelling a class with TargetClass vs labelling it with Substitute. I think it is something to do with whether fields that are not explicitly targeted get copied across.

adinn (approved):
This looks ok to me.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
zakkak: > However, I am not 100% clear what the difference is supposed to be between labelling a class with TargetClass vs labelling it with Substitute. I think it is something to do with whether fields that are not explicitly targeted get copied across.

This is covered in https://github.com/oracle/graal/blob/6a8914b68f1709e90c98f9dbd2134d854b2efbf5/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/annotate/TargetClass.java#L64-L69 seems to cover the last part.

The difference is essentially that in a Class annotated with `TargetClass` you may only `Substitute` the methods you want and the rest of the original methods will be available, on the other hand in a Class annotated with `Substitute` as well all the original methods are deleted and only methods annotated with `Substitute` or `KeepOriginal` are now available.

adinn: Yeah, that's the comment I was vaguely remembering having read. When I said fields above I really meant to say members.

PR #3857, opened 2021-10-01
Truncation in JVM_GetNanoTimeAdjustment() on 32-bits archs

koutheir:
The C `long` type is only 64-bits-wide on LP64 systems and the like. In 32-bits systems, `long` is usually 32-bits-wide. On such systems:
- The constant `0x0100000000L` cannot be represented on 32-bits, and should be `0x0100000000LL` instead, in order to explicitly mark its type as a `long long`.
- The Java `jlong` cannot be represented in a C `long` and it can cause unintentional data loss (truncation). `jlong` always corresponds to `int64_t`.


rodrigar-mx:
Thank you for contributing to GraalVM, we will take a look into your changes.

Issue #3636, opened 2021-07-30
native-image, JFR, buffers "[warn][jfr,system] Unable to commit. Requested size too large"


Karm:
Hello,

This issue, or perhaps a feature request if you will, concerns with handling buffers for larger JFR events when JFR is used with native-image built binaries. I will demonstrate what I mean on my homemade [JDK 11](https://github.com/openjdk/jdk11u-dev) build (HEAD f700d37d2b)  and [Mandrel master](https://github.com/graalvm/mandrel/tree/graal/master) build  (HEAD 6e6bbcacb3) and then show the behavior is also reproducible with [GraalVM CE 21.2.0](https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-21.2.0/graalvm-ce-java11-linux-amd64-21.2.0.tar.gz).

## printf debuging
I patched the  [JDK 11](https://github.com/openjdk/jdk11u-dev) build (HEAD f700d37d2b) as follows:

```patch
index 595635e41b..708effdeb1 100644
--- a/src/hotspot/share/jfr/recorder/storage/jfrStorage.cpp
+++ b/src/hotspot/share/jfr/recorder/storage/jfrStorage.cpp
@@ -562,6 +562,7 @@ static BufferPtr large_fail(BufferPtr cur, bool native, JfrStorage& storage_inst
 // even though it might be smaller than the requested size.
 // Caller needs to ensure if the size was successfully accommodated.
 BufferPtr JfrStorage::provision_large(BufferPtr cur, const u1* const cur_pos, size_t used, size_t req, bool native, Thread* t) {
+  log_info(jfr)("JfrStorage::provision_large called, used: " SIZE_FORMAT ", requested: " SIZE_FORMAT ".", used, req);
   debug_only(assert_provision_large_precondition(cur, used, req, t);)
   assert(t->jfr_thread_local()->shelved_buffer() != NULL, "invariant");
   BufferPtr const buffer = acquire_large(req, t);
diff --git a/src/jdk.jfr/share/classes/jdk/jfr/internal/EventWriter.java b/src/jdk.jfr/share/classes/jdk/jfr/internal/EventWriter.java
index a1d697bb9e..e5825eed21 100644
--- a/src/jdk.jfr/share/classes/jdk/jfr/internal/EventWriter.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/internal/EventWriter.java
@@ -195,7 +195,9 @@ public final class EventWriter {
             return false;
         }
[...]

PR #3865, opened 2021-10-05
native-image, JFR, buffers "[warn][jfr,system] Unable to commit. Requested size too large"

zhengyu123:
This patch address issue #3636

Current Jfr implementation can not handle events larger than default JfrBuffer size (8k), that results those events discarded.

The solution I purpose here, is to create temporary buffers to hold the large events, free them once they are flushed.

### DIFF
--- a/substratevm/src/com.oracle.svm.jfr/src/com/oracle/svm/jfr/JfrRecorderThread.java
+++ b/substratevm/src/com.oracle.svm.jfr/src/com/oracle/svm/jfr/JfrRecorderThread.java
@@ -107,6 +107,23 @@ public class JfrRecorderThread extends Thread {
                 }
             }
         }
+
+        // Flush temp buffers
+        buffers = globalMemory.getTempBuffers();
+        for (int i = 0; i < globalMemory.getBufferCount(); i++) {
+            JfrBuffer buffer = buffers.addressOf(i).read();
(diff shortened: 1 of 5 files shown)


substratevm/src/com.oracle.svm.jfr/src/com/oracle/svm/jfr/JfrRecorderThread.java
```diff
@@ -107,6 +107,23 @@ private void persistBuffers(JfrChunkWriter chunkWriter) {
...
             }
         }
+
+        // Flush temp buffers
+        buffers = globalMemory.getTempBuffers();
+        for (int i = 0; i < globalMemory.getBufferCount(); i++) {
+            JfrBuffer buffer = buffers.addressOf(i).read();
+            if (buffer.isNonNull()) {
+                assert buffer.isTempBuffer();
+                boolean shouldNotify = persistBuffer(chunkWriter, buffer);
```
christianhaeubl, line 117: Please move the common logic to a separate method.

zhengyu123:
Sorry for long delay. It appears to be a wrong solution after I rebased the patch to current tip. The real issue is that it should create large event for the case and large event support apparently broken right now. I close this PR, will purpose new PR to fix large event.

PR #3867, opened 2021-10-05
TRegex: Add support for simple boolean match checking.

nirvdrum:
The purpose of this PR is to add support for regex matches where there is no need for the match results beyond a boolean. In particular, it would make Ruby's `Regexp#match?` method avoid any match data allocation. Naturally, we can assure we don't create a Ruby `MatchData` from TruffleRuby, but without this PR we end up creating the resulting `int[]` in TRegex, which then goes unused.

I'm not sure I've done this the best we can with TRegex, but it does work. It's implemented as an additional, optional argument to the `exec` and `execBytes` interop calls. If the argument is not provided, the `simpleMatch` option is disabled and everything works the way it does today. If the `simpleMatch` argument is `true`, then a static instance of `RegexResult` is used on a match (similar to the way the `NO_MATCH`  works already).

With a corresponding set of changes I have for TruffleRuby to take advantage of this new option and using a simple expression like:

```ruby
1_000_000.times { /./.match?("abc") }
```

I end up with a decent code size savings.

Without PR:
```
[engine] opt done     Regexp#match?                                               |AST   30|Tier 2|Time 1176( 526+649 )ms|Inlined   9Y   0N|IR   521/  729|CodeSize   2024|Addr 0x7fd42e5d0890|Src resource:/truffleruby/core/regexp.rb:179
```

With PR:
```
[engine] opt done     Regexp#match?                                               |AST   30|Tier 2|Time  831( 402+430 )ms|Inlined   9Y   0N|IR   520/  590|CodeSize   1558|Addr 0x7f50da625810|Src resource:/truffleruby/core/regexp.rb:179
```

I'm currently seeing a 10 - 15% improvement on some benchmarks, but I need to iron those out. For the time being, I'm using a simple microbenchmark with a variety of regexps, both those that pass and fail, against a fixed size string.

I'm happy to rework the PR as necessary. This was more of a proof of concept to see if it's something the TRegex team would entertain adding.

### DIFF
--- a/regex/src/com.oracle.truffle.regex/src/com/oracle/truffle/regex/RegexExecNode.java
+++ b/regex/src/com.oracle.truffle.regex/src/com/oracle/truffle/regex/RegexExecNode.java
@@ -60,8 +60,9 @@ public abstract class RegexExecNode extends RegexBodyNode {
     @Override
     public final RegexResult execute(VirtualFrame frame) {
         Object[] args = frame.getArguments();
-        assert args.length == 2;
-        return executeDirect(args[0], (int) args[1]);
+        assert args.length == 2 || args.length == 3;
+        boolean simpleMatch = args.length == 3 ? (boolean) args[2] : false;
+        return executeDirect(args[0], (int) args[1], simpleMatch);
     }
 
     private int adjustFromIndex(int fromIndex, Object input) {
@@ -89,17 +90,17 @@ public abstract class RegexExecNode extends RegexBodyNode {
         return charAtNode.execute(input, i);
     }
 
-    public RegexResult executeDirect(Object input, int fromIndex) {
+    public RegexResult executeDirect(Object input, int fromIndex, boolean simpleMatch) {
         if (fromIndex < 0 || fromIndex > inputLength(input)) {
             CompilerDirectives.transferToInterpreterAndInvalidate();
             throw new IllegalArgumentException(String.format("got illegal fromIndex value: %d. fromIndex must be >= 0 and <= input length (%d)", fromIndex, inputLength(input)));
         }
-        return execute(input, adjustFromIndex(fromIndex, input));
+        return execute(input, adjustFromIndex(fromIndex, input), simpleMatch);
     }
 
     public boolean isBacktracking() {
         return false;
     }
 
-    protected abstract RegexResult execute(Object input, int fromIndex);
+    protected abstract RegexResult execute(Object input, int fromIndex, boolean simpleMatch);
 }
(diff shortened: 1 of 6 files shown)


nirvdrum:
@jirkamarsik @djoooooe @eregon You all might find this one interesting.

eregon:
Could you show the difference before/after with seafoam graphs?
Shouldn't the `int[]` be escape-analyzed and then it would be the same? Or maybe it escapes because the `match?` call sometimes matches and sometimes not?

djoooooe:
> Shouldn't the `int[]` be escape-analyzed and then it would be the same? Or maybe it escapes because the `match?` call sometimes matches and sometimes not?

Maybe the `int[]` escapes into the global value returned by `RegExp::last_mach`? JS had the exact same problem, which we alleviated a bit by adding an assumption that speculates on the global is never being used, and just stored the regexp match invocation parameters instead. If the assumption is invalidated, the last match can be re-calculated using the last parameters.

eregon:
> Maybe the `int[]` escapes into the global value returned by `RegExp::last_mach`?

No, `Regexp#match?` doesn't set `$~`/`Regexp.last_match` so that shouldn't be a concern.
And `$~`/`Regexp.last_match` are actually stored in the frame, it's not global.

The MatchData instance is already EA'd correctly if a regexp always matches.
But if it sometimes matches and sometimes not then I think the compiler has no choice but to allocate the MatchData (not sure about the TRegex result/`int[]` in that case).

nirvdrum:
Here are graphs of `Regexp#match?`, which returns a simple boolean value and does not need a `MatchData` for anything. I've tried to strip out some of the Ruby-specific logic in the graphs, but I couldn't get all of it. If you focus on the bottom of the two graphs, you'll see the difference in TRegex resulting from this PR.

[Graph without the simple boolean match](https://github.com/oracle/graal/files/7294097/graph-with-regex-result.pdf)
[Graph with the simple boolean match](https://github.com/oracle/graal/files/7294017/graph-without-regex-result.pdf)

djoooooe:
I'll reject this PR in favor of a more comprehensive solution via a regex compilation option or something equivalent, since matching without having to track any capture group boundaries allows the construction of much simpler DFAs:

At the moment, if you'd just want to know if a match exists in a given string, you'd run a DFA that just searches for the end index of a match, while ignoring all other capture groups. This DFA, however, currently simulates the behavior of a backtracking engine internally, such that the index reported is the same as one would get for the end of capture group 0 on a backtracker. If TRegex knows during DFA construction that this index is not needed, we can generate much simpler DFAs for these cases.

Example: matching `/ab|a/` on `"ab"` .
a backtracker would report "ab" as the full match, due to the order in which it processes the alternation. TRegex could stop immediately after the `"a"` in this case, because the DFA considers both branches at the same time, and we'd no longer be bound by backtracker semantics.

nirvdrum:
Okay. A more comprehensive solution sounds better, of course. I just hope it doesn't turn into a big architectural change that takes multiple releases to finish. The simplistic approach here has tangible benefits today.

Please let me know if you want me to try out any potential replacements.

eregon:
@nirvdrum @djoooooe just opened a TRegex PR to add support for the optimized boolean matching: https://github.com/oracle/graal/pull/4211
Would you like to work on adopting it in TruffleRuby?

nirvdrum:
Yeah. I'll take care of that.

PR #3927, opened 2021-10-20
Changed representation of WebAssembly Modules and Instances.

flohuemer:
Updated the internal representation of WebAssembly Modules and Instances. Changed the instantiation of WebAssembly Instances to reflect the [WebAssembly spec](https://webassembly.github.io/JS-BigInt-integration/core/exec/modules.html#instantiation).

- WasmModule and WasmInstance were simplified.
    - SymbolTable and RuntimeState were merged into WasmModule and WasmInstance.
- Simplified the representation of types (WasmInstance) in WasmNodes to only contain necessary data.
    - The types present are
        - FunctionTypes
        - FunctionInstances
        - Tables
        - Memories
        - Globals
    - Import and export information in now in the WasmModuleInstance.
    - WasmInstance no longer has a reference to WasmModule.
    - Globals are now represented by the WasmGlobal class.
- The BinaryParser now generates an intermediate representation of Wasm types (Memories, Globals, ...) which is later translated by the WasmInstantiator.
    - The intermediate representation is in the `...Definition` classes.
- Types (Memories, Globals, ...) are no longer stored in registries. They are now stored in their instances. 
- The Linker was merged into the WasmInstantiator.
- Late linking is no longer possible.
    - This is actually not defined in the WASM spec.
    - The LateLinkingSuite is no longer part of the Wasm tests. 
    - This also effects the `import_export_functions` tests in the LinkerSuite.
    - The tests in the LinkerSuite are now loaded in the order needed by the initialization phase.
- Fixed **TODO (GR-26734)**, where the first module of a context had to be renamed to `main` because of a WASI hack. They are now using their correct names.
    - This also effects internal benchmark names.
- These changes have no impact on the performance of compiled code.


axel22:
@flohuemer I'd prefer that we have a meeting before to discuss all the implications of a change of this magnitude.

flohuemer:
I will close this pull request and try to split it up into smaller ones.

Issue #3811, opened 2021-09-17
XLST causes the executable to crash at runtime


andirady:
**Describe the issue**
The following XLST doesn't work at runtime after compiled with native-image but works in JVM.

```xml
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
    <xsl:output omit-xml-declaration="yes" indent="yes"/>
    <xsl:strip-space elements="*"/>

    <xsl:template match="@*|node()">
        <xsl:copy>
            <xsl:apply-templates select="@*|node()"/>
        </xsl:copy>
    </xsl:template>

</xsl:stylesheet>
```

The problematic code portion is as follows:
```java
Transformer transformer;
try (var xlst = XmlFormatter.class.getClassLoader().getResourceAsStream("formatter.xlst")) {
    transformer = tf.newTransformer(new StreamSource(xlst)); // crashed presumably happened here.
}
```

**Steps to reproduce the issue**
Please include both build steps as well as run steps
1. Download zip from https://gist.github.com/andirady/82a8407523a45453541a7f9ef2802462
2. Extract the zip
3. ``cd`` into the extracted directory
2. ``make xmlformatter``
3. ``./xmlformatter test.xml``

**Describe GraalVM and your environment:**
 - GraalVM CE 21.2.0 (build 16.0.2+7-jvmci-21.2-b08)
 - openjdk version "16.0.2" 2021-07-20
 - Windows Subsystem for Linux (Debian 4.4.0-19041-Microsoft)
 - Architecture: x86_64

**More details**
```
Fatal error reported via JNI: JVM_DefineModule called:  Unimplemented

JavaFrameAnchor dump:

  Anchor 00007ffff2c40758 LastJavaSP 00007ffff2c40730 LastJavaIP 00007f803bbe23e8
  
TopFrame info:

  TotalFrameSize in CodeInfoTable 80
  
VMThreads info:

  VMThread 00007fffea0cf280  STATUS_IN_JAVA (safepoints disabled)  java.lang.Thread@0x7f803a645380
  
VM Thread State for current thread 00007fffea0cf280:

  0 (8 bytes): com.oracle.svm.jni.JNIThreadLocalEnvironment.jniFunctions = (bytes) 
    00007fffea0cf280: 00007f803a3df010
  
[...]

PR #3940, opened 2021-10-22
Substitute `ClassLoader.createOrGetClassLoaderValueMap` in order to make `Module.defineModule0` work with XLST

ivan-ristovic:
This PR addresses a part of #3811.

`ClassLoader.createOrGetClassLoaderValueMap` is invoked by `Module.defineModule0` in this case, and inititalizes the `classLoaderValueMap` in an atomic way. The JDK implementation uses `Unsafe` to get-or-set the field, so this PR introduces a substitution that will bypass a call to `Unsafe.objectFieldOffset`.

Our implementation of `Unsafe.objectFieldOffset` substitution currently uses reflection to fetch the offset of a field. If a field is filtered by JDK's reflection filters, that invocation will throw.


vjovanov:
I am OK with this PR if we open (and link from the code) the ticket that ill fix the problem in the right way.

vjovanov (approved):

ivan-ristovic:
This PR is now internally tracked as GR-36579.

ivan-ristovic:
This PR has been merged.

PR #3967, opened 2021-11-01
[GR-34184] Remove or replace jdk8 gates.

graalvmbot:


dougxc:
@zakkak since we're in the process of removing JDK 8 support, maybe you could help me by updating the `.github/workflows/main.yml` and `.github/workflows/quarkus.yml` to remove JDK 8 jobs. The general strategy is to replace them with JDK 17 jobs. I'm fine with you pushing a commit to this PR but it can be done in a separate PR as well.

zakkak:
@dougxc noted, I will work on it this week :)

compiler/mx.compiler/suite.py
```diff
@@ -1,5 +1,5 @@
 suite = {
-  "mxversion" : "5.309.2",
+  "mxversion" : "5.316.0",
```
zakkak, line 2: @dougxc I can't access 5.316.0, mx `master` seems to be on 5.315.2 https://github.com/graalvm/mx/blob/dd0a0c7a81c412722650f99e9123707a99e1f088/mx.py#L17796

dougxc: It's actually going to be easier if you test your changes in a separate branch. I can then later graft that onto my branch (or we can merge it independently).

PR #3980, opened 2021-11-02
Use reflection to invoke HotSpotJVMCIRuntime.getCurrentJavaThread

zakkak:
https://github.com/oracle/graal/commit/deb80d4bc6f7115c756c619c45d48375d8109fdf
relies on method `HotSpotJVMCIRuntime.getCurrentJavaThread()` which was
introduced by https://bugs.openjdk.java.net/browse/JDK-8231828, but was
not back-ported to upstream OpenJDK 11.

This patch uses reflection to access the method, like in
https://github.com/oracle/graal/commit/ab692b25f0d66df602c997ebafbdd0532d5e2008,
to avoid compilation errors when using upstream OpenJDK 11, aka Mandrel builds (which is
already known to not work with LibGraal due to JVMCI incompatibilities).

Closes: https://github.com/graalvm/mandrel/issues/310
Closes: https://github.com/graalvm/mandrel/issues/311

### DIFF
--- a/compiler/src/org.graalvm.libgraal.jni/src/org/graalvm/libgraal/jni/LibGraalJNIMethodScope.java
+++ b/compiler/src/org.graalvm.libgraal.jni/src/org/graalvm/libgraal/jni/LibGraalJNIMethodScope.java
@@ -32,6 +32,8 @@ import org.graalvm.nativeimage.c.type.CLongPointer;
 import org.graalvm.word.PointerBase;
 import org.graalvm.word.WordFactory;
 
+import java.lang.reflect.Method;
+
 import static org.graalvm.nativebridge.jni.JNIUtil.PopLocalFrame;
 import static org.graalvm.nativebridge.jni.JNIUtil.PushLocalFrame;
 
@@ -44,6 +46,17 @@ import static org.graalvm.nativebridge.jni.JNIUtil.PushLocalFrame;
  * created by libgraal will never be freed (i.e., a memory leak).
  */
 final class LibGraalJNIMethodScope extends JNIMethodScope {
+
+    private static final Method getCurrentJavaThreadMethod;
+    static {
+        Method m;
+        try {
+            m = HotSpotJVMCIRuntime.class.getMethod("getCurrentJavaThread");
+        } catch (NoSuchMethodException e) {
+            m = null;
+        }
+        getCurrentJavaThreadMethod = m;
+    }
 
     private static volatile int lastJavaPCOffset = -1;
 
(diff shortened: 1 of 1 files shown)


zakkak:
@oubidar-Abderrahim @mcraj017 could you please assign this to the right person? 
It is blocking us from building and testing Mandrel based on `master`.

Thank you in advance

tzezula:
But the `HotSpotJVMCIRuntime#getCurrentJavaThread` had been already called in the [Target_org_graalvm_compiler_hotspot_HotSpotGraalCompiler#compileMethod](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.graal.hotspot.libgraal/src/com/oracle/svm/graal/hotspot/libgraal/LibGraalFeature.java#L648) without reflection before the `LibGraalJNIMethodScope#open` was called.

dougxc:
> But the `HotSpotJVMCIRuntime#getCurrentJavaThread` had been already called in the [Target_org_graalvm_compiler_hotspot_HotSpotGraalCompiler#compileMethod](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.graal.hotspot.libgraal/src/com/oracle/svm/graal/hotspot/libgraal/LibGraalFeature.java#L648) without reflection before the `LibGraalJNIMethodScope#open` was called.

That was addressed by https://github.com/oracle/graal/blob/6ac7297ab9fd5f88d24a9a90a8eb2316a242dee8/substratevm/mx.substratevm/suite.py#L935.
@zakkak I would recommend just doing the same here: https://github.com/oracle/graal/blob/6ac7297ab9fd5f88d24a9a90a8eb2316a242dee8/compiler/mx.compiler/suite.py#L447 in the Mandrel fork.

zakkak:
> @zakkak I would recommend just doing the same here:
> 
> https://github.com/oracle/graal/blob/6ac7297ab9fd5f88d24a9a90a8eb2316a242dee8/compiler/mx.compiler/suite.py#L447
>
> in the Mandrel fork.

ACK, I have worked around this issue by removing the `org.graalvm.libgraal.jni` dependency through mandrel-packaging (see https://github.com/graalvm/mandrel-packaging/pull/197) like we do with a number of other packages that have caused issues in the past.

FWIW ideally we would like to be able to build `native-image` from `master` using upstream OpenJDK builds and `mx`. To achieve this we would need to restructure the dependencies in a way that allows us to skip building anything libgraal (or truffle etc.) related when the configuration does not rely on it. Unfortunately though this keeps getting pushed down on my priority list :/

Issue #3935, opened 2021-10-21
Improve module support for native-image building


olpaw:
https://github.com/oracle/graal/issues/1962 was implemented for 21.3. Basic support for building images from modules is available. But more work is needed.

- While all of the images coming from the substratevm suite (`native-image-launcher`, `native-image-agent-library`, ...) are already built on module-path our other GraalVM images from other suites are still built in the classpath. This needs to be fixed.
- Currently the image builder only runs on the module-path if an image should be built from the module-path. This needs to changes so that the image-builder always runs on the module-path.
- Module-info at image runtime is generated at image-buildtime. This is still incomplete. Hosted module access setup needs to be translated to runtime module-info access state so that module-related checks pass at image runtime as expected. One instance of this problem is explained in detail in https://github.com/oracle/graal/issues/3911
- Currently `StackTraceElement.getModuleName()` does not return module names. (see https://github.com/oracle/graal/issues/4008)

mcraj017:
@olpaw There is a request to add module name in stack traces of native image, please add that in this effort also
https://github.com/oracle/graal/issues/4008

PR #3991, opened 2021-11-05
[GR-34444] Replicate hosted module relations in runtime boot module layer

### LINKED ISSUE #3911: [GR-34444] Ensure java.lang.Module#isReflectivelyExportedOrOpen works as expected
Currently `java.lang.Module#isReflectivelyExportedOrOpen` simply returns true at image-runtime. This is needed to e.g. make e.g. ResourceBundle lookups at image runtime working. The ResourceBundle lookup code involves:
```java
        public ResourceBundle newBundle(String baseName, Locale locale, String format,
                                        ClassLoader loader, boolean reload)
                    throws IllegalAccessException, InstantiationException, IOException {
            /*
             * Legacy mechanism to locate resource bundle in unnamed module only
             * that is visible to the given loader and accessible to the given caller.
             */
            String bundleName = toBundleName(baseName, locale);
            ResourceBundle bundle = null;
            if (format.equals("java.class")) {
                try {
                    Class<?> c = loader.loadClass(bundleName);
                    // If the class isn't a ResourceBundle subclass, throw a
                    // ClassCastException.
                    if (ResourceBundle.class.isAssignableFrom(c)) {
                        Class<ResourceBundle> bundleClass = (Class<ResourceBundle>)c;
                        Module m = bundleClass.getModule();

                        // To access a resource bundle in a named module,
                        // either class-based or properties-based, the resource
                        // bundle must be opened unconditionally,
                        // same rule as accessing a resource file.
                        if (m.isNamed() && !m.isOpen(bundleClass.getPackageName())) {
                            throw new IllegalAccessException("unnamed module can't load " + // 🗲🗲🗲
                                bundleClass.getName() + " in " + m.toString());
                        }
                        ...
```
After https://github.com/oracle/graal/pull/3445 got merged this code can now fail with `IllegalAccessException` (see 🗲🗲🗲) due to moduleinfo no knowing that `m.isOpen(bundleClass.getPackageName())` should be true.

As a temporary workaround substitution:
```java
    @SuppressWarnings({"unused", "static-method"})
    @Substitute
    public boolean isReflectivelyExportedOrOpen(String pn, Module other, boolean open) {
        return true;
    }
```
was added to `Target_java_lang_Module_JDK11OrLater`.

A proper fix would remove that workaround and instead make sure that `m.isOpen(bundleClass.getPackageName())` gives the right answer. The hosted isOpen relationships between modules need to be carried over to module-info at runtime.

A good example that demonstrates if this is working correctly is using `ResourceBundle.getBundle("sun.security.util.Resources")` at image runtime. The image builder arguments always contains
```
--add-exports=java.base/sun.security.util=org.graalvm.nativeimage.builder
```
thus the bundle lookup at image built-time succeeds. But for the bundle lookup to also succeed at image runtime this opening up needs to be translated to the corresponding opening-up at image runtime.

ivan-ristovic:
This PR implements runtime relations between modules by transferring hosted module relations to runtime.

Fixes #3911 and addresses #3935

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
@@ -144,7 +144,7 @@ public final class ModuleLayerFeature implements Feature {
 
         Set<Module> analysisReachableSyntheticModules = analysisReachableNamedModules
                         .stream()
-                        .filter(m -> m.getDescriptor().modifiers().contains(ModuleDescriptor.Modifier.SYNTHETIC))
+                        .filter(ModuleLayerFeatureUtils::isModuleSynthetic)
                         .collect(Collectors.toSet());
 
         Set<String> allReachableModules = analysisReachableNamedModules
@@ -230,7 +230,7 @@ public final class ModuleLayerFeature implements Feature {
                         continue;
                     }
                     Module runtimeTo = e2.getValue().runtimeModule;
-                    if (hostedFrom.canRead(hostedTo)) {
+                    if (ModuleLayerFeatureUtils.isModuleSynthetic(hostedFrom) || hostedFrom.canRead(hostedTo)) {
                         moduleLayerFeatureUtils.addReads(runtimeFrom, runtimeTo);
                         if (hostedFrom == builderModule) {
                             for (Module appModule : applicationModules) {
@@ -239,7 +239,7 @@ public final class ModuleLayerFeature implements Feature {
                         }
                     }
                     for (String pn : runtimeFrom.getPackages()) {
-                        if (hostedFrom.isOpen(pn, hostedTo)) {
+                        if (ModuleLayerFeatureUtils.isModuleSynthetic(hostedFrom) || hostedFrom.isOpen(pn, hostedTo)) {
                             moduleLayerFeatureUtils.addOpens(runtimeFrom, pn, runtimeTo);
                             if (hostedTo == builderModule) {
                                 for (Module appModule : applicationModules) {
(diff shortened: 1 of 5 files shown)


substratevm/src/com.oracle.svm.hosted.jdk11/src/com/oracle/svm/hosted/ModuleLayerFeature.java
```diff
@@ -157,14 +188,79 @@ private ModuleLayer synthesizeRuntimeBootLayer(ImageClassLoader cl, Set<String>
...
     }
 
+    private void replicateVisibilityModifications(ModuleLayer runtimeBootLayer, ImageClassLoader cl, Set<Module> analysisReachableModules) {
+        List<Module> applicationModules = findApplicationModules(runtimeBootLayer, cl.applicationModulePath());
+
+        Map<String, HostedRuntimeModulePair> moduleLookupMap = analysisReachableModules
+                        .stream()
+                        .collect(Collectors.toMap(Module::getName, m -> new HostedRuntimeModulePair(m, runtimeBootLayer)));
+
+        Module builderModule = moduleLookupMap.get("org.graalvm.nativeimage.builder").hostedModule;
```
olpaw, line 206: Use `ModuleLayerFeature.class.getModule().getName()` instead of hardcoding `"org.graalvm.nativeimage.builder"`

olpaw: Or just `ModuleLayerFeature.class.getModule()` instead of `moduleLookupMap.get("org.graalvm.nativeimage.builder").hostedModule`

olpaw:
@ivan-ristovic I ran your PR in our CI and I see gate failures with `mx native-unittest --builder-on-modulepath ...`.
Steps to reproduce on your side:
```
cd substratevm
mx build
mx gate --omit-clean -t 'native unittests'
```
This will give you:
```
...
There were 5 failures:
1) com.oracle.svm.test.jdk11.jfr.TestClassEvent#test
java.lang.IllegalStateException: Can't create Flight Recorder. Unable to make field private static jdk.jfr.internal.handlers.EventHandler jdk.jfr.events.ActiveRecordingEvent.eventHandler accessible: module jdk.jfr does not "opens jdk.jfr.events" to module org.graalvm.nativeimage.builder
	at jdk.jfr.FlightRecorder.getFlightRecorder(FlightRecorder.java:183)
	at jdk.jfr.Recording.<init>(Recording.java:114)
	at jdk.jfr.Recording.<init>(Recording.java:138)
	at com.oracle.svm.test.jdk11.jfr.LocalJFR.startRecording(LocalJFR.java:41)
	at com.oracle.svm.test.jdk11.jfr.TestClassEvent.test(TestClassEvent.java:40)
...
FAILURES!!!
Tests run: 56,  Failures: 5
```
Please investigate.

PR #4007, opened 2021-11-11
[GR-34740] [GR-34739] Implement ResourcesFeature Module awareness.

graalvmbot:


olpaw:
@ivan-ristovic @jovanstevanovic FYI this PR will add the ability to have per-module resource inclusion.
E.g. having the same resource path but in two different modules is now working as expected.

TODO: native-image agent support for writing resource-config.json entries for `Module..getResourceAsStream()` calls

PR #4032, opened 2021-11-18
[GR-35117] Avoid pulling File and URL code in image via CodeSource.location.

graalvmbot:


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ProtectionDomainSupport.java
```diff
@@ -0,0 +1,143 @@
...
+ * {@link ProtectionDomainFeature#beforeAnalysis}.
+ * 
+ * Note that this still leads to observable differences in places where the location is used
+ * implicitly, like {@link CodeSource#toString} and {@link CodeSource#implies}. We accept that
+ * difference in behavior.
+ */
+public final class ProtectionDomainSupport {
+
+    public static class Options {
+        @Option(help = "Return a the application path as the Class.getProtectionDomain().getCodeSource().getLocation() for all classes that have no explicit ProtectionDomain")//
```
zakkak, line 72:
```suggestion
        @Option(help = "Return the application path as the Class.getProtectionDomain().getCodeSource().getLocation() for all classes that have no explicit ProtectionDomain")//
```

PR #4040, opened 2021-11-22
Fix typo in EncodedSnippets.java

eltociear:
orignal -> original


ezzarghili:
Thank you for the update, as this is a very small change and only a typo fix, we don't need a full PR integration for it, @olyagpl can take update the docs in an other PR, in the future opening an issue would be enough for team to add the change as part of an existing PR.

PR #4042, opened 2021-11-22
[GR-35274] Restore missing param type names in method signatures in debug log statements

adinn:
This PR fixes an error introduced during refactoring of the debug types info code (GR-31277). Full method names used in log statements are supposed to list the method param type names in the signature. The are currently listing the method parameter names (which at present are empty strings).

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
@@ -87,7 +87,7 @@ public class Range {
         }
         this.methodEntry = methodEntry;
         this.fullMethodName = isInline ? stringTable.uniqueDebugString(constructClassAndMethodName()) : stringTable.uniqueString(constructClassAndMethodName());
-        this.fullMethodNameWithParams = stringTable.uniqueString(constructClassAndMethodNameWithParams());
+        this.fullMethodNameWithParams = constructClassAndMethodNameWithParams();
         this.lo = lo;
         this.hi = hi;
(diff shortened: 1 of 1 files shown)


adinn:
This is a fix for issue #4043.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
```diff
@@ -87,7 +87,7 @@ public Range(StringTable stringTable, MethodEntry methodEntry, int lo, int hi, i
         }
         this.methodEntry = methodEntry;
         this.fullMethodName = isInline ? stringTable.uniqueDebugString(constructClassAndMethodName()) : stringTable.uniqueString(constructClassAndMethodName());
-        this.fullMethodNameWithParams = stringTable.uniqueString(constructClassAndMethodNameWithParams());
```
olpaw, line 90: So putting this into the string-table was unnecessasy?

adinn: Yes. The string table serves two purposes

1) deduplicate strings that may occur more than once so as to reduce the amount of memory used during debug info generation
2) optionally demarcate a subset of those strings as needing to be generated into the strings section and indexed.

In this case we know that the full method name with params string will only occur in this one instance and only be used for log messages. So, insertion in the string table is wasted effort.

olpaw (approved):

oubidar-Abderrahim:
Hi @olpaw 
is this PR ready to be mirrored to our internal repo?

PR #4051, opened 2021-11-24
[GR-21865] Introduce copy-on-write image heap provider on Windows.

pejovica:
This PR adds a copy-on-write image heap provider for Windows.

A few high-level notes on implementation:
- New image heaps are provided by memory mapping views of the loaded image heap.
- In case of failure to map a new image heap, the implementation falls back to copying from memory as before.
- For memory mapping to work, the OS must have support for memory placeholders. This is checked at runtime as not all Windows versions have the necessary support.
- The use of memory placeholders is transparent to the rest of the system, i.e., it is completely encapsulated in `WindowsVirtualMemoryProvider` .

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractImageHeapLayouter.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractImageHeapLayouter.java
@@ -77,6 +77,10 @@ public abstract class AbstractImageHeapLayouter<T extends AbstractImageHeapLayou
         return partitions;
     }
 
+    private T getLastPartition() {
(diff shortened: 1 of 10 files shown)


peter-hofer (approved):

joerg-wille:
This PR breaks compatibility with upx.
It is desirable to achieve minimum app size for which upx is commonly used to compress nativ Image.

Some ideas for a more proper fix:
- Compare the static parts of the in-memory image heap with the mapped file and fall back to `super.commitAndCopyMemory` when not matching
- Create a paging file backed (`INVALID_HANDLE_VALUE`) file mapping object, copy the uncompressed image heap to it and then that can be used and shared the same way as the mapped executable file

_Originally posted by @kornelpal in https://github.com/upx/upx/issues/559#issuecomment-1032047038_

PR #4053, opened 2021-11-24
Add `poly_engine_get_languages_size` to C API

### LINKED ISSUE #4052: `poly_engine_get_languages` in C ABI makes reference to `poly_engine_get_languages_size` that doesn't exist
**Describe the issue**

If you want to get the languages available using the C ABI, currently the code is a bit redundant and makes a reference to a function that I think was meant to be written, but then never was:

https://github.com/oracle/graal/blob/0550f758478993b1a0c011f4c8186915ee910d6a/substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java#L223-L245

Why this is a problem, is because it forces you to call the function twice.
Once to read the size, and then again to allocate the language elements inside of the array:

```cpp
// std::vector<>
size_t num_languages1;
poly_engine_get_languages(thread, engine, NULL, &num_languages1); // get the number of languages
std::vector<poly_language> languages(num_languages1);
poly_engine_get_languages(thread, engine, languages.data(), &num_languages1); // put the actual language pointers into the vector

// poly_language[]
size_t num_languages2;
poly_engine_get_languages(thread, engine, NULL, &num_languages2); // get the number of languages
poly_language languages_ptr[num_languages2];
poly_engine_get_languages(thread, engine, &languages_ptr, &num_languages2); // get the number of languages
```

**Expected behavior**

The function `poly_engine_get_languages` is implemented, which probably looks something like:

```java
 @CEntryPoint(name = "poly_engine_get_languages_size", documentation = { 
                 "Returns size of languages, used by {@link poly_engine_get_languages}", 
                 "", 
                 " @param engine for which languages are returned.", 
                 " @param size the number of languages in the engine.", 
                 " @since 19.0", 
 }) 
 public static PolyglotStatus poly_engine_get_languages_size(PolyglotIsolateThread thread, PolyglotEngine engine, SizeTPointer size) { 
     return withHandledErrors(() -> { 
         Engine jEngine = fetchHandle(engine); 
         UnsignedWord languagesSize = WordFactory.unsigned(jEngine.getLanguages().size()); 
         size.write(languagesSize); 
     }); 
 } 
```

Now you can write this:
```cpp
size_t num_languages;
poly_engine_get_languages_size(thread, engine, &num_languages);

poly_language languages[num_languages];
poly_engine_get_languages(thread, engine, languages, NULL);
```

GavinRay97:
Close #4052

### DIFF
--- a/substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
+++ b/substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
@@ -243,6 +243,21 @@ public final class PolyglotNativeAPI {
             }
         });
     }
+    
+     @CEntryPoint(name = "poly_engine_get_languages_size", documentation = { 
+                 "Returns size of languages, used by {@link poly_engine_get_languages}", 
+                 "", 
+                 " @param engine for which languages are returned.", 
+                 " @param size the number of languages in the engine.", 
+                 " @since 22.0", 
+     }) 
+     public static PolyglotStatus poly_engine_get_languages_size(PolyglotIsolateThread thread, PolyglotEngine engine, SizeTPointer size) { 
+         return withHandledErrors(() -> { 
+             Engine jEngine = fetchHandle(engine); 
+             UnsignedWord languagesSize = WordFactory.unsigned(jEngine.getLanguages().size()); 
+             size.write(languagesSize); 
+         }); 
+     }
 
     @CEntryPoint(name = "poly_create_context_builder", documentation = {
                     "Creates a context with a new engine polyglot engine with a list ",
--- a/substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
+++ b/substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
@@ -221,8 +221,24 @@ public final class PolyglotNativeAPI {
     }
 
     @CEntryPoint(name = "poly_engine_get_languages", documentation = {
-                    "Returns an array of size returned by {@link poly_engine_get_languages_size} where each element is a <code>poly_language<code> handle.",
+                    "Returns an array where each element is a <code>poly_language<code> handle.",
                     "",
+                    "To use, make two calls to this method:",
+                    "<p>",
+                    "   The first passing a size_t pointer to write the count of the languages to (IE: 'size_t poly_language_count')",
+                    "   The second passing the value of the written size_t length, and a pointer to a poly_language[poly_language_size]",
+                    "<code>",
+                    "size_t poly_language_count;",
+                    "",
+                    "// Get the number of languages",
+                    "poly_engine_get_languages(thread, engine, NULL, &poly_language_count);",
+                    "",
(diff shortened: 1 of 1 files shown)


substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
```diff
@@ -243,6 +243,21 @@ public static PolyglotStatus poly_engine_get_languages(PolyglotIsolateThread thr
             }
         });
     }
+    
+     @CEntryPoint(name = "poly_engine_get_languages_size", documentation = {
```
vjovanov, line 247: Adding this function would be redundant. The size of the array is already returned by the `SizeTPointer size`. I think only the documentation of that function needs to be updated.

GavinRay97: I agree it's redundant. I just saw the mention in the docs and thought it was an error.
However, this pattern for using it is also somewhat redundant and confused me at first:

```c

size_t num_languages;
poly_language languages_ptr[num_languages];

// get the number of languages
poly_engine_get_languages(thread, engine, NULL, &num_languages);

// put the actual language pointers into the array
poly_engine_get_languages(thread, engine, &languages_ptr, &num_languages); 
```

Maybe the best thing to do would be to explicitly state in the doc comment that to use it, it requires being called twice, with different combinations of parameters?

vjovanov: Yes, improving the docs is best. I know it is strange, I had the same argument with the C guy when we were designing this. Apparently, this is the standard thing in C.

GavinRay97: > I had the same argument with the C guy when we were designing this. Apparently, this is the standard thing in C.

🤣 Glad I was not the only one confused, haha.
Awesome -- I will remove the extra method and just update the existing docs accordingly 👍

GavinRay97:
Updated, requesting a review 👍

substratevm/src/org.graalvm.polyglot.nativeapi/src/org/graalvm/polyglot/nativeapi/PolyglotNativeAPI.java
```diff
@@ -221,8 +221,24 @@ public static PolyglotStatus poly_engine_close(PolyglotIsolateThread thread, Pol
     }
 
     @CEntryPoint(name = "poly_engine_get_languages", documentation = {
-                    "Returns an array of size returned by {@link poly_engine_get_languages_size} where each element is a <code>poly_language<code> handle.",
```
vjovanov, line 224: This will end up in the C API. Is `@link` of relevance there?

vjovanov: Answering to myself: https://www.doxygen.nl/manual/commands.html

vjovanov (approved):

vjovanov:
Closing as obsolete: the entire Polyglot Native API was removed in July 2024, so `poly_engine_get_languages_size` is no longer applicable.

Issue #4008, opened 2021-11-11
Show module name in stacktrace


Thihup:
Currently the stacktraces in native image doesn't show the module name, even when using a Java module.

One example is the following:
```
public class Main {

    public static void main(String[] args) {
        if (!Main.class.getModule().isNamed())
            throw new AssertionError("Must be in a named module");
        StackTraceElement stackTraceElement = new Exception().getStackTrace()[0];
        String moduleName = stackTraceElement.getModuleName();
        if (moduleName == null || "".equals(moduleName))
            throw new AssertionError("Must be in a named module");
        System.out.println("Hello from " + moduleName);
    }
}
```
When using the module-path, I guess it should return the module name.

mcraj017:
@Thihup thanks for reporting the issue, we will check it out and get back to you

mcraj017:
@Thihup I will add this request in the ongoing effort of improving module support on native-image

mcraj017:
please follow https://github.com/oracle/graal/issues/3935 for more updates, closing this issue now

olpaw:
@Thihup good point. This is currently missing and should be added. @ivan-ristovic this is something that would fit in your previous line of work.

PR #4056, opened 2021-11-26
Support showing module names in stack traces

ivan-ristovic:
Addresses #4008 and adds new features to #3935.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/code/SourceReferenceSupplier.java
+++ b/substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/code/SourceReferenceSupplier.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2021, 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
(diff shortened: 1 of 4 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/SourceReferenceSupplier.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.core.code;
+
+import com.oracle.svm.core.hub.DynamicHub;
+
+final class SourceReferenceSupplier {
```
olpaw, line 29: As already discussed with @ivan-ristovic, this uses the `overlayTarget` mechanism that we want to get rid of.

Let's wait with merging this until we have Java 8 removed from substratevm. Then is fallback class will become obsolete.

substratevm/src/com.oracle.svm.core.jdk11/src/com/oracle/svm/core/code/SourceReferenceSupplier.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Optional;
+
+final class SourceReferenceSupplier {
+
+    public static StackTraceElement createReference(Class<?> sourceClass, String sourceMethodName, int sourceLineNumber) {
+        if (sourceClass == null) {
+            return new StackTraceElement("", sourceMethodName, null, sourceLineNumber);
+        }
+
+        final String classLoaderName = sourceClass.getClassLoader() != null ? sourceClass.getClassLoader().getName() : null;
```
olpaw, line 39: No point in using final in locals.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/SourceReferenceSupplier.java
olpaw: Alternatively you can use `com.oracle.svm.util.ModuleSupport#getModuleName` to avoid introducing two variants of `com.oracle.svm.core.code.SourceReferenceSupplier`.

ivan-ristovic: That is true, however I could not remove the need for two separate constructors for the `StackTraceElement`, one of which is only present in JDK9+ (with module-related arguments).

PR #4074, opened 2021-12-02
Fix sulong build in github actions.

rschatz:
Looks like the update to CMake 3.22.0 broke the LLVM CMake build. Building libc++ and libc++abi individually still seems to work.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -146,15 +146,6 @@ jobs:
     - name: Build GraalVM and run gate
       env: ${{ matrix.env }}
       run: |
-        if [[ "${PRIMARY}" == "vm" || "${PRIMARY}" == "sulong" ]]
-        then
-          # workaround issue in cmake 3.22: downgrade to 3.21.4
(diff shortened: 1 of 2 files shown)


fniephaus (approved):

rschatz:
I accidentally broke the build on Mac, which our internal CI caught. But should be fixed on the revision I just pushed.

rschatz:
Still failing on Mac, at least in our CI. On my Macbook it works. No idea what the issue is here.

But anyway, according to https://libcxx.llvm.org/BuildingLibcxx.html#the-default-build, I think what I did in that PR is moving in the wrong direction anyway. Unfortunately, doing exactly what's described in that link still fails with CMake 3.22.

I'm asking for help on the LLVM mailing list, I need some help from someone more familiar with the LLVM build to debug this.
https://lists.llvm.org/pipermail/llvm-dev/2021-December/154140.html

In the meantime, I'm investigating whether we can downgrade the cmake version in Github Actions, to at least get these jobs green again until we can have a real solution for this issue.

rschatz:
I did merge a temporary manual downgrade of our CMake version as a quick fix for the failing gate. Still working on the "real" fix...

PR #4078, opened 2021-12-02
Fix library paths for native library reporterchelper

zakkak:
A new native shared library `reporterchelper` was introduced in https://github.com/oracle/graal/commit/e04a9cd21653bfc5c44c3dc2702b86085216de71 

However its location doesn't seem to be consistent with the location of other native libraries:

Under windows the location seems to be wrong, as it ends up under `/bin/`, and I would expect peakRSS to not work under windoes given that GraalVM [tries to load the library from `lib/svm/builder/lib/`](https://github.com/oracle/graal/blob/6fbf945cdd53fa0370f9c36a59058d912e3f6194/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/reporting/ProgressReporterCHelper.java#L48):
```
$ find -name "*chelper*" 
./lib/svm/clibraries/windows-amd64/libchelper.lib
./lib/svm/builder/clibraries/windows-amd64/libchelper.lib
./bin/svm/builder/lib/reporterchelper.dll
```

Under linux things look OK, but the library location is not platform dependent:
```
$ find -name "*chelper*"                    
./lib/svm/clibraries/linux-amd64/liblibchelper.a
./lib/svm/builder/clibraries/linux-amd64/liblibchelper.a
./lib/svm/builder/lib/libreporterchelper.so
```

### DIFF
--- a/substratevm/mx.substratevm/suite.py
+++ b/substratevm/mx.substratevm/suite.py
@@ -1556,8 +1556,13 @@ suite = {
             "native" : True,
             "platformDependent" : True,
             "description" : "SubstrateVM support libraries for the GraalVM",
+            "platforms" : [
+                "linux-amd64",
+                "darwin-amd64",
+                "windows-amd64",
+            ],
             "layout" : {
-                "svm/builder/lib/" : ["dependency:com.oracle.svm.native.reporterchelper"],
+                "builder/clibraries/<os>-<arch>/" : ["dependency:com.oracle.svm.native.reporterchelper"],
             },
             "maven" : False,
         },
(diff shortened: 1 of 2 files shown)


fniephaus:
Thanks for the PR, @zakkak!

>I would expect peakRSS to not work under windows

You are right, RSS reporting is currently broken on Windows. I'll have a look at this tomorrow.

fniephaus:
I have pushed a different fix for this with https://github.com/oracle/graal/commit/74b547b9812bf97c188c5130ca6612cac97643b6. Nonetheless, thanks for pointing out the issue and providing a possible fix!

PR #4090, opened 2021-12-06
Bump mx version for substratevm

zakkak:
4609007cba98c4d02738170401c55900b001c97f breaks builds with mx 5.310.0 due to
its dependency on
https://github.com/graalvm/mx/commit/2a364eec1736fb0c858703491bd30605526cb2ea

### DIFF
--- a/substratevm/mx.substratevm/suite.py
+++ b/substratevm/mx.substratevm/suite.py
@@ -1,6 +1,6 @@
 # pylint: disable=line-too-long
 suite = {
-    "mxversion": "5.310.0",
+    "mxversion": "5.315.0",
     "name": "substratevm",
     "version" : "22.0.0",
     "release" : False,


zakkak:
@rodrigar-mx @dougxc This is blocking one of our CI pipelines, could someone please have a look? Thanks

dougxc:
@rodrigar-mx please proceed to merge this.

rodrigar-mx:
ACK.

zakkak:
@ezzarghili @rodrigar-mx this needs to be backported to `release/graal-vm/22.0` as well. Thanks.

zakkak:
@gilles-duboscq FYI this didn't get backported to `release/graal-vm/22.0`.

zakkak:
> For release branches we use a fixed mx version that can be found in the `graal-common.json` file at the root of the repo.

The issue with that is that it points to `HEAD` for non-release branches, so there is not a single point of truth when it comes to automated testing.

> I can include this fix in the `release/graal-vm/22.0` branch but it will likely land only _after_ the 22.0.0 release.

It wouldn't be worth a 22.0.0.1 release (we work around it by applying the patch in our code base), but if you find other bugs and plan to do a bug release nevertheless then please include it.

PR #4115, opened 2021-12-12
Updated uq interpreter

utting:
An interpreter for GraalVM IR graphs.  
This can be useful for evaluating IR graphs during testing.

This implements 'interpret(InterpreterState)' methods in 30+ of the most common subclasses of FixedNode in order to interpret those control-flow nodes, and 'interpretExpr(InterpreterState)' methods in most of the common data flow nodes in order to interpret expressions.  The main entry point for the interpreter is the GraalInterpreter class and its executeGraph method.

The interpreter is based on a 2020-2021 UQ honours project by Leon Karstens plus further contributions by Kristian Thomassen and Mark Utting (UQ).  Some current limitations are that it focusses mostly on high-level nodes rather than lowered nodes, supports only simple objects whose fields are primitive values, it attempts to interpret methods in ALL classes that are called (an alternative might be to treat some java.lang.* classes as pre-defined), and uses reflection to initialize static fields (which means that it cannot find some private fields).


dougxc (commented):
I have done a cursory review of this PR and left a few comments.
While it's a good start, it will need substantial work before it might be integrated. Namely, all the TODO's should be addressed and other unimplemented functionality should be completed.

PR #4107, opened 2021-12-10
Fix fstat call in darwin-aarch64.

oraluben:
Looks like `fstat` and some other system calls don't have `$INODE64` suffix on aarch64 anymore, even if `_DARWIN_USE_64_BIT_INODE` is given, since they're only for a backward-compatible issue which only happens on Intel platform.

This PR is to fix the failure on darwin-aarch64.

<details>
  <summary>Failed test</summary>
  
  ```
# yyc @ Yichens-M1 in ~/IdeaProjects/graal/graal/substratevm on git:master o [17:08:52] 
$ mx native-unittest --regex ImageInfoTest 
========================================================================================================================
GraalVM Native Image: Generating 'com.oracle.svm.junit.svmjunitrunner'...
========================================================================================================================
[1/7] Initializing...                                                                                   (18.4s @ 0.21GB)
 Version info: 'GraalVM dev Java 17 CE'
 4 user-provided feature(s)
  - com.oracle.svm.configure.config.SignatureUtilFeature
  - com.oracle.svm.junit.JUnitFeature
  - com.oracle.svm.test.NativeImageResourceFileSystemProviderTest$RegisterResourceFeature
  - com.oracle.svm.test.SerializationRegistrationTestFeature
[2/7] Performing analysis...  [******]                                                                  (58.2s @ 0.69GB)
   3,823 (78.60%) of  4,864 classes reachable
   4,844 (61.95%) of  7,819 fields reachable
  17,933 (48.66%) of 36,851 methods reachable
      49 classes,    55 fields, and   622 methods registered for reflection
      58 classes,    60 fields, and    51 methods registered for JNI access
[3/7] Building universe...                                                                               (3.8s @ 1.12GB)
[4/7] Parsing methods...      [**]                                                                       (2.9s @ 2.62GB)
[5/7] Inlining methods...     [****]                                                                     (4.5s @ 1.37GB)
[6/7] Compiling methods...    [*****]                                                                   (31.3s @ 2.35GB)
[7/7] Creating image...Fatal error: java.lang.RuntimeException: There was an error linking the native image: Linker command exited with 1

Linker command executed:
[...]

PR #4136, opened 2021-12-17
Fix Python version in Github Actions.

oraluben:
The corresponding issue not only blocks running checks with Github Actions, but also triggers (spam) failure from Github Actions every time downstream repositories merge upstream graal.


oraluben:
Commited in #4107

PR #4180, opened 2022-01-06
[GR-36201] Avoid control chars in new build output to dumb terminals

adinn:
The new build output is wonderful except when you are running your build inside a dumb terminal (i.e. a shell inside emacs). The control sequences posted to the tty just make a mess of the output so it looks like this:

    $ mx debuginfotest
    . . .
    ================================================================================
    ^[]8;;https://www.graalvm.org/native-image^[\GraalVM Native Image^[]8;;^[\: Generating 'hello.hello'...
    ================================================================================
    [1/7] ^[]8;;https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/BuildOutput.md#stage-initializing^[\Initializing^[]8;;^[\...                                            (5.3s @ 0.27GB)
     ^[]8;;https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/BuildOutput.md#glossary-version-info^[\Version info^[]8;;^[\: 'GraalVM dev Java 11 CE'
    . . .

This patch modifies the native image build command definded in mx_substrate.py to detect a dumb terminal and disable the build options that generate control sequences, resulting in output like this:

    $ mx debuginfotest
    . . .
    ================================================================================
    GraalVM Native Image: Generating 'hello.hello'...
    ================================================================================
    [1/7] Initializing...                                            (5.8s @ 0.28GB)
     Version info: 'GraalVM dev Java 11 CE'
    . . .

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
@@ -77,8 +77,7 @@ public class ProgressReporter {
     private static final int CHARACTERS_PER_LINE;
     private static final int PROGRESS_BAR_START = 30;
     private static final boolean IS_CI = System.console() == null || System.getenv("CI") != null;
-    boolean IS_DUMB_TERM = "dumb".equals(System.getenv("TERM"));
-
+    private static final boolean IS_DUMB_TERM = isDumbTerm();
     private static final int MAX_NUM_FEATURES = 50;
     private static final int MAX_NUM_BREAKDOWN = 10;
     private static final String CODE_BREAKDOWN_TITLE = String.format("Top %d packages in code area:", MAX_NUM_BREAKDOWN);
@@ -135,6 +134,11 @@ public class ProgressReporter {
 
     static {
         CHARACTERS_PER_LINE = IS_CI ? ProgressReporterCHelper.MAX_CHARACTERS_PER_LINE : ProgressReporterCHelper.getTerminalWindowColumnsClamped();
+    }
+
+    private final static boolean isDumbTerm() {
+        String term = System.getenv("TERM");
+        return (term == null || term.equals("") || term.equals("dumb") || term.equals("unknown"));
     }
 
     public static ProgressReporter singleton() {
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
@@ -77,6 +77,7 @@ public class ProgressReporter {
     private static final int CHARACTERS_PER_LINE;
     private static final int PROGRESS_BAR_START = 30;
     private static final boolean IS_CI = System.console() == null || System.getenv("CI") != null;
+    boolean IS_DUMB_TERM = "dumb".equals(System.getenv("TERM"));
 
     private static final int MAX_NUM_FEATURES = 50;
     private static final int MAX_NUM_BREAKDOWN = 10;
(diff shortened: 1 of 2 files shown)


adinn:
n.b this PR fixes issue #4181

adinn:
@fniephaus could you please review this fix

fniephaus:
Thanks for the PR, @adinn. I was not aware of `$TERM=dumb`. Instead of fixing this on the mx level, should we check for `$TERM` being `dumb` directly in the reporter (somewhere [here](https://github.com/oracle/graal/blob/934e3e64f932f844f803273e51dafb6b96b843d5/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java#L151-L170) maybe). This way, the dump terminal logic will also work without going through mx.

adinn:
Hi @fniephaus

> I was not aware of $TERM=dumb

Use of TERM == dumb to indicate that the current tty should not be sent escape sequences is a convention that I think started with emacs but is, so I believe, now recognized by a variety of other software. It worked initially because there was no termcap entry for dumb (yes, I realise that last statement only makes sense if you have over a yard of beard). Any other unreconized name would have done just as well (e.g. not the usual suspects like annarbor, vt100, etc). Essentially, it meant that attempts to lookup in file /etc/termcap an escape sequence appropriate to some desired screen display/operation would produce an empty string.

> Instead of fixing this on the mx level, should we check for $TERM being dumb directly in the reporter (somewhere here maybe). This way, the dumb terminal logic will also work when building without mx.

You can build without using mx? :-O

Sure, I guess it makes sense for the check and disable to happen in the reporter. I have pushed a new version that implements this.

fniephaus:
> You can build without using mx? :-O

I didn't mean building Native Image itself without mx but all of our users use it directly (e.g. `native-image HelloWorld`), right?

> Sure, I guess it makes sense for the check and disable to happen in the reporter. I have pushed a new version that implements this.

Excellent, thanks!

fniephaus (approved):

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
```diff
@@ -77,6 +77,7 @@
     private static final int CHARACTERS_PER_LINE;
     private static final int PROGRESS_BAR_START = 30;
     private static final boolean IS_CI = System.console() == null || System.getenv("CI") != null;
+    boolean IS_DUMB_TERM = "dumb".equals(System.getenv("TERM"));
```
fniephaus, line 80: I'm going to append `private static final` to this for consistency.

adinn: Yes, apologies for not doing so in the first place.

adinn:
@fniephaus Is this proceeding through the internal gate or is it blocked on something?

fniephaus:
It's been blocked by me but I should be able to get this in this week. Sorry for the delay.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
olpaw, line 80: It seems `"unknown"` is another synonym for `"dumb"` that we should probably recognize here.
See https://stackoverflow.com/questions/39001517/features-obligatory-for-term-dumb-terminal

Also if `TERM` is not set at all (or empty) we should also interpret that as `IS_DUMB_TERM`.

adinn: Yes, I agree we should treat all 3 cases as IS_DUMB_TERM. Actually there are four cases because we need to allow for System.getenv returning null, "", "dumb" or "unknown". I have pushed an update.

olpaw (approved):

fniephaus:
This PR is now in the pipeline for merging (fixed a style issue on the way).

fniephaus:
Merged via https://github.com/oracle/graal/commit/909ae11ee18ea4b25666d341ca56d528b3123818.

Issue #4018, opened 2021-11-15
[GR-35118] debuginfotest fails on JDK17; assertion error on aarch64.


fniephaus:
After switching the vm-native gates from JDK 8 to JDK 17, the [debuginfotest started to fail on JDK17](https://github.com/oracle/graal/runs/4125253987?check_suite_focus=true). To debug the issue, I tried to reproduce it on an aarch64 machine. However, image generation crashed due to an `AssertionError` (see below).

@zakkak is that expected/a known issue? Is debuginfo supported on aarch64?

```
$ mx --primary-suite-path substratevm --J @-Xmx2g debuginfotest
parent=/home/foo/dev/graal/substratevm
sourcepath=/home/foo/dev/graal/substratevm/src/com.oracle.svm.test/src
sourcecache=/home/foo/dev/graal/substratevm/svmbuild/sources
[...]

PR #4183, opened 2022-01-06
Ensure old debuginfotest build products are not reused

adinn:
The mx debuginfotest command fails when building the test native image binary if an existing binary is present. The command builds and tests the binary twice, first with -H:-SpawnIsolates and then with -H:+SpawnIsolates. The failure  can happen at the first build if the command is being rerun without an intervening mx clean. If not it will occur at the second build because of the binary produced first time around. This patch modifies the python implementation to remove any existing build target before building.

### DIFF
--- a/substratevm/mx.substratevm/mx_substratevm.py
+++ b/substratevm/mx.substratevm/mx_substratevm.py
@@ -241,6 +241,9 @@ def native_image_context(common_args=None, hosted_assertions=True, native_image_
     common_args = [] if common_args is None else common_args
     base_args = ['--no-fallback', '-H:+EnforceMaxRuntimeCompileMethods']
     base_args += ['-H:Path=' + svmbuild_dir()]
+    # avoid escape sequences in output to dumb terminals (or emacs)
+    if os.getenv("TERM") == "dumb":
+        base_args += ["-H:-BuildOutputColorful", "-H:-BuildOutputProgress"]
     if mx.get_opts().verbose:
         base_args += ['--verbose']
     if mx.get_opts().very_verbose:
@@ -727,10 +730,18 @@ def _debuginfotest(native_image, path, build_only, args):
 
     # build with and without Isolates and check both work
 
+    # remove previously built native image binary
+    if os.path.exists(join(path, 'hello.hello')):
+        os.remove(join(path, 'hello.hello'))
+
     build_debug_test(['-H:+SpawnIsolates'])
     if mx.get_os() == 'linux' and not build_only:
         os.environ.update({'debuginfotest.isolates' : 'yes'})
         mx.run([os.environ.get('GDB_BIN', 'gdb'), '-ex', 'python "ISOLATES=True"', '-x', join(parent, 'mx.substratevm/testhello.py'), join(path, 'hello.hello')])
+
+    # remove previously built native image binary
+    if os.path.exists(join(path, 'hello.hello')):
+        os.remove(join(path, 'hello.hello'))
 
     build_debug_test(['-H:-SpawnIsolates'])
     if mx.get_os() == 'linux' and not build_only:


adinn:
@dougxc Could you please take a look at this patch.

I'm not sure if this fix is overkill. Perhaps mx can be coaxed into correctly identifying whether or not a rebuild is required?

adinn:
@dougxc Hmm, forget this PR,  there is something else going on here. Even with the removal of the previously created image the image generation sometimes fails at Creating image 

    ================================================================================
    GraalVM Native Image: Generating 'hello.hello'...
    ================================================================================
    [1/7] Initializing...                                            (5.4s @ 0.27GB)
     Version info: 'GraalVM dev Java 11 CE'
    . . .
    [6/7] Compiling methods...    [*****]                           (29.0s @ 1.50GB)
    [7/7] Creating image...--------------------------------------------------------------------------------
        4.0s (4.0% of total time) in 23 GCs | Peak RSS: 5.01GB | CPU load: 6.71
    --------------------------------------------------------------------------------
    Produced artifacts:
     /home/adinn/redhat/openjdk/graal/graal/substratevm/svmbuild/hello.hello.build_artifacts.txt
    ================================================================================

In a successful run we see

    [7/7] Creating image...                                         (12.9s @ 0.98GB)
       5.10MB ( 7.05%) for code area:    9,631 compilation units
       9.14MB (12.65%) for image heap:   2,130 classes and 133,600 objects
    . . .

I wonder if this may be something to do with the build output changes (@fniephaus any ideas?)

fniephaus:
What happens if you disable the new build output (via `-H:-BuildOutputUseNewStyle`)?

adinn:
> What happens if you disable the new build output (via -H:-BuildOutputUseNewStyle)?

Doh, I should have thought of that!

It still fails some of the time:

    $ mx debuginfotest
    . . .
    [hello.hello:89121]    (compile):  28,653.83 ms,  4.64 GB
    [hello.hello:89121]      compile:  40,724.32 ms,  4.64 GB

fniephaus:
Ok good, so it's unlikely to be related to the new output. Interesting that it fails without a proper error message.

adinn:
Hi @fniephaus.

I debugged this and it appears that the problem here is the one reported in #4018 i.e. the debug info generator cannot cope when a `JNICallTrampolineMethod` and a `JNINativeCallWrapperMethod` are generated with the same name. However, the difference here is that the assertion is not appearing in the printed output.

The exception propagates out of the debug info code and is in `NativeImageGenerator.main` in the catch block for `Throwable`

        . . .
        } catch (Throwable err) {
            reportFatalError(err);
            exitStatus = 1;
        } finally {
        . . .

the call to `reportFatalError` does the obvious

        protected static void reportFatalError(Throwable e) {
            System.err.print("Fatal error: ");
            e.printStackTrace();
        }

However, this does not result in any visible output, *even if* you specify `-H:-BuildOutputUseNewStyle` on the command line.

The reason for this is that the exception flow bypasses any code that might flush the wrapper PrintStream that is installed as the value of System.err. The exception happens under code that lies between the calls to `reporter.printCreationStart()` and `reporter.printCreationEnd` in `NativeGmageGenerator::doRun`. It actually pops out through the catch block for `Throwable where `VMError.shouldNotReachHere` gets called.

            . . .
            reporter.printCreationStart();
            Timer imageTimer = new Timer(imageName, "image");
            try (Indent indent = debug.logAndIndent("create native image")) {
                try (DebugContext.Scope buildScope = debug.scope("CreateImage", codeCacheProvider)) {
                    . . .
                } catch (Throwable e) {
                    throw VMError.shouldNotReachHere(e);
                }
            }
            Timer writeTimer = new Timer(imageName, "write");
            try (StopTimer t = writeTimer.start()) {
                 . . .
            }
            reporter.printCreationEnd(imageTimer, writeTimer, image.getImageSize(), bb.getUniverse(), heap.getObjectCount(), image.getImageHeapSize(), codeCache.getCodeCacheSize(),
                            codeCache.getCompilations().size(), image.getDebugInfoSize());
            if (SubstrateOptions.BuildOutputBreakdowns.getValue()) {
                ProgressReporter.singleton().printBreakdowns(compileQueue.getCompilationTasks(), image.getHeap().getObjects());
            }
            . . .

Of course, as the comment says, we are not supposed to arrive at the Throwable catch block. However, that does not mean the builder output code is off the hook.

During development we are definitely going to hit that catch block if any of the debuginfo generation code triggers an assert. Similarly, if the image write stage triggers an assert then exception flow will also bypass the `reporter.printCreationEnd` call. So in either case the assertion details will get swallowed by the wrapper when `NativeImageGenerator.reportFatalError` tries to print them.

I suspect the same problem can arise should an assert happen in one of the earlier stages. These all rely on calling PrintStageStart in a try block header so that the AutoClosable returned by PrintStageStart calls PrintStageEnd when the exception flows out of the try block. However, that is still too late to log the assertion. Printing of its details to stderr will only happen in the outer catch scope where `NativeImageGenerator.reportFatalError` gets called.

An obvious solution would be to flush the reporter inside that routine except that the reporter is not available in that outer block (it is local to method `doRun`). Perhaps a better solution would be to employ a finally block in doRun which asks the ProgressReporter to unwrap System.out and System.err to their original values at the point where it goes out of scope? Indeed this might even be best done using a try block AutoClosable?

fniephaus:
Thanks for digging into this. Looks like I was wrong and the reason you're not seeing the error is indeed related to the new build output. We recently introduced wrappers for stdout/stderr to control when content is printed (as your found out) and that not only can swallow errors but also causes issues with the old build output. It looks like we need to revert this and ensure that stdout/stderr is always printed immediately. I'm already working on a PR for that and hope we can get this in asap.

Nonetheless, this is independent from the issue you're addressing with this PR.

adinn:
@fniephaus Thanks for picking up the problem with swallowed exceptions. I'll keep an eye on the new PR. I am closing this PR as the problem that the build output change is masking is just a duplicate and is already being addressed by #4018.

olpaw:
@adinn, thanks for the groundwork on this issue. This is severe, as it compromises our users ability to provide meaningful bug reports. I have a PR that will be on master and on the 22.0 release branch soon. https://github.com/oracle/graal/pull/4196

PR #4211, opened 2022-01-14
[GR-36237] TRegex: boolean match results and capture group tracking code deduplication.

graalvmbot:
* TRegex: de-duplicate capture group updating code
* TRegex: add boolean match mode

Boolean match mode is used by calling `execBoolean` instead of `exec` on a `RegexObject`. This allows using the same `RegexObject` in both modes. By default, matchers/DFAs for `exec` are generated first, and `execBoolean` generates matchers lazily. This behavior can be inverted by setting the regex option `BooleanMatch=true` in the regex source.

This PR also fixes sticky flag matching, no matchers will produce results starting at the wrong indices anymore.


nirvdrum:
I've got this working in TruffleRuby. Thanks for all the work. For the non-boolean match cases, we have `exec` and `execBytes`. Should there be an equivalent `execBytesBoolean`?

eregon:
cc @djoooooe

djoooooe:
`byte[]` as an input string type is now deprecated in favor of `TruffleString`, which enables more optimizations inside TRegex. I'd strongly suggest to adopt `TruffleString` instead of keeping the `byte[]` based API.

Issue #4224, opened 2022-01-18
Static field offsets encoded in DWARF debug info are incorrect when isolates are enabled


adinn:
**Issue description**
Printing of static fields in gdb is no longer working correctly when debugging a native image created using -H:+SpawnIsolates. The offset of the static field from the heap register that is encoded into the debug records appears to be less than the correct value by 0x100000.

Note that this functionality was working previously but has not been regularly tested because of a problem in gdb 9 & 10. The problem can be manifested when using gdb 11.

Note also that the offset and subsequent printing are correct if the native image is compiled with -H:-SpawnIsolates. 

**Steps to reproduce the issue**
Ensure you have gdb 11 in your path
Clone the latest graal repo
Run the following command
```
$ (export GDB_CAN_PRINT=True ; mx debuginfotest )
  . . .
(gdb) print 'java.math.BigDecimal'::BIG_TEN_POWERS_TABLE
$3 = (_z_.java.math.BigInteger[] *) 0x0

(gdb) print 'java.math.BigDecimal'::BIG_TEN_POWERS_TABLE->data[3]->mag->data[0]
Traceback (most recent call last):
  File "/home/adinn/redhat/openjdk/graal/graal/substratevm/mx.substratevm/testhello.py", line 655, in <module>
    test()
  File "/home/adinn/redhat/openjdk/graal/graal/substratevm/mx.substratevm/testhello.py", line 278, in test
    exec_string = execute("print 'java.math.BigDecimal'::BIG_TEN_POWERS_TABLE->data[3]->mag->data[0]")
  File "/home/adinn/redhat/openjdk/graal/graal/substratevm/mx.substratevm/testhello.py", line 115, in execute
    return gdb.execute(command, to_string=True)
gdb.error: Location address is not set.
(gdb) 
```
At this point you need to type q to quit the failing scripted gdb session.

**Describe GraalVM and your environment:**
 - GraalVM version: latest head
 - JDK major version: 11
 - OS: Linux
 - Architecture: x86/aarch64

**Further info**

At the quit prompt above it is possible to identify the offset being used for field
`java.math.BigDecimal::BIG_TEN_POWERS_TABLE` as follows
```
(gdb) p &'java.math.BigDecimal'::BIG_TEN_POWERS_TABLE
$4 = (_z_.java.math.BigInteger[] **) 0x7ffff7758ce8
(gdb) p &'java.math.BigDecimal'::BIG_TEN_POWERS_TABLE - $r14
[...]

PR #4226, opened 2022-01-18
[GR-36446] Fix static field printing when isolates are enabled and test printing in debuginfotest with gdb 11.0+

adinn:
This is a fix for the problem reported in issue #4224.

The fix also enables printing of objects, including those reachable via a static field, under mx debuginfotest when gdb with version 11.0 or greater is in the path. Issue #4224 was not found until recently because the test was disabled due to a bug in gdb 10.x.

### DIFF
--- a/substratevm/mx.substratevm/testhello.py
+++ b/substratevm/mx.substratevm/testhello.py
@@ -149,8 +149,8 @@ def test():
     # range of major.minor versions
     # for now we use an env setting
     print("Found gdb version %s.%s"%(major, minor))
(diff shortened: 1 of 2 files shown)


olpaw (approved):

olpaw:
Running in internal CI now

adinn:
Great, thanks.

adinn:
@olpaw Any chance of this being pushed soon? Does it need a review from @pejovica?

olpaw:
Our CI is holding us back with transients. It's all approved. It should be on master soon.

adinn:
great thanks

PR #4228, opened 2022-01-18
[GR-36532] Improve debug info log output

adinn:
This trivial fix improves the quality of debug info log output

- Log info for compiled code subranges is moved from level VERBOSE to DETAILED. 
- Log info for heap object/primitive data is moved from level VERBOSE to DETAILED and generated under log context id DebugDataInfo instead of the current DebugCodeInfo.

Subrange messages constitute about 3/4 of the log output for a normal image build when log pattern DebugCodeInfo is passed. Data info messages produce much less output but separating them from code info messages is appropriate since info about heap objects is needed much less often than code info.


adinn:
@olpaw Could you please review this clean up of debug info log message generation.

adinn:
@olpaw @pejovica Could on eof you review this trivial fix to log output and then help get itinto master?

olpaw (approved):

PR #4241, opened 2022-01-19
[GR-36478] Fix detection of debuginfo size on Darwin.

graalvmbot:


zakkak (approved):
LGTM. Thanks @fniephaus

PR #4258, opened 2022-01-25
[GR-35701] Update Debug Info Feature doc.

graalvmbot:


adinn:
Regarding these suggest amendments:

> The first option, `-g`, instructs `native-image` to generate debug information, and `-O0` specifies that no compiler optimizations should be performed.
> Disable all optimizations is not required, but in general it makes the debugging experience better with fewer optimizations applied.

I think this advice is thoroughly misguided. No one in their right mind should be debugging a native image as part of development.That can and should be done when running on on the JVM where there will be a much better user experience.

The purpose of debug info in native images is to provide users some way of identifying what a deployed program is doing diffeently to the JVM version should behaviour of the native image ever diverge from the expected, correct behaviour established by testing and debugging on the JVM. In other words it is needed to debug problems with the way a native image has been generated.

In almost all cases that means users will need to debug the optimized binary used in deployment. In some cases building an unoptimized native image may help to pinpoint which specific aspect of the generation process has gone wrong but that will not be the common case. The 'better' debugging experience obtained by passing -O0 may make the debug session easier to follow but it will also render it pointless in almost all cases.

petermz:
Plus, I don't believe `-O0` actually means "no compiler optimizations". AFAICT it only affects inlining, all the other optimizations are applied as usual:
https://github.com/oracle/graal/blob/818b52c4a1a49a20e7329ae817b992e00a0394b0/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java#L133-L143

zakkak:
Closing this as superseded by https://github.com/oracle/graal/pull/4259

Issue #3929, opened 2021-10-20
Allow incomplete classpath configuration to be more granular


galderz:
**Is your feature request related to a problem? Please describe.**
Having 3rd party libraries define `--allow-incomplete-classpath` can be very problematic for frameworks such as Quarkus. 

When the flag is there, errors start to become more difficult to understand. It would even seem that the native image itself behaves differently enough, that seemingly unrelated errors appear out of nowhere.

Quarkus tries as hard as possible to avoid any 3rd party libraries setting it, but at time some of those are proprietary or closed source, so there's nothing we can do about it.

The worst part of the flag is the fact that it has a global implication.

**Describe the solution you'd like.**
It would be nice to have an optional way to make the flag not global. That is, find a way to trim the code that is affected by this flag.

For example, could we find an idiomatic way to say: allow these methods/classes/packages to reference bytecode for which classes are not in the classpath. For any other part of the code, throw an error.

**Describe who do you think will benefit the most.**
Frameworks which depend on GraalVM, e.g. Quarkus.

**Describe alternatives you've considered.**
We don't see any other alternatives.

**Additional context.**
Add any other context about the feature request here.
For example, link to the relevant projects, documentation, standards.

**Express whether you'd like to help contributing this feature**
I'd be happy to implement it once we've found an agreement.

galderz:
For reference, Sanne wrote up some more thoughts about this in [this discussion](https://github.com/oracle/graal/discussions/3932#discussioncomment-1544294).

christianwimmer:
PR with a proposed solution: https://github.com/oracle/graal/pull/4305

galderz:
@christianwimmer #4305 is not a solution for the problem specified above.

Using the nomenclature in #4305, Quarkus would be in the boat of using `--link-at-build-time` without arguments via the native-image, but we lack the ability to specify which classes/packages we'd want to be linked at run time. An alternative solution that Quarkus goes and specifies all packages+classes `--link-at-build-time` without mentioning the classes/packages that would need to be linked at run time is simply not scalable.

zakkak:
> An alternative solution that Quarkus goes and specifies all packages+classes --link-at-build-time without mentioning the classes/packages that would need to be linked at run time is simply not scalable.

It looks like Quarkus might end up doing something similar for build time initialization (as discussed in https://github.com/quarkusio/quarkus/discussions/42682 and https://github.com/quarkusio/quarkus/pull/48427) which would then enable us to use for `--link-at-build-time` as well.

Issue #2762, opened 2020-08-18
Remove some limitations of GraalVM Native Image


christianwimmer:
GraalVM Native Image has a few [limitations](https://github.com/oracle/graal/blob/master/substratevm/LIMITATIONS.md) that are preventing users from building native images from their code bases. This issue describe the areas where improvements are planned over the next releases. Actual work is done under separate issues, linked from this issue so that everyone who wants to work on it gets an overview of who is working on what.

### Support for the module system

The native image generator does not support a module path argument yet on JDK 11. The plan is to
* support all module path arguments in the same way as the the `java` launcher, and
* allow introspection of module information at run time using the respective JDK reflection APIs.

Only information about modules where at least one class is reachable will be available at run time.

Work to support the module path arguments is currently in progress: https://github.com/oracle/graal/issues/1962

### Support for method handles and the invokedynamic bytecode

Currently only method handles that are a compile time constant are supported. This is an important subset because it allows, for example, all Java lambdas and the invokedynamic-based string concatenation. Still, we want to support method handles also when they are not a compile time constant, i.e., when method handle chains are constructed and changed at run time.

The approach will be similar to reflection: methods that are available via a method handle must be registered in a configuration file at image build time.

Note that the performance of such method handles will be slower compared to the Java HotSpot VM because no dynamic compilation of method handle chains is possible.

Work is currently in progress: https://github.com/oracle/graal/issues/2761

### (Done) Support for Java serialization

Java serialization is a special form or reflection: during deserialization, classes are instantiated and fields are set in a reflective way without explicit bytecode. So the implementation approach will be similar to reflection: classes that are available for serialization must be registered in a configuration file at image build time.

(Update: support for serialization is finished and available starting with GraalVM 21.0)

### Improve support for resource bundles and locales

Currently, the locale is fixed at image build time, i.e., registered resource bundles are only included for a single locale. This single locale is of course also the default locale at runtime. This will be made more flexible, so that arbitrary locales can be included and the default locale is set correctly at run time.

Work is currently in progress: https://github.com/oracle/graal/issues/2982

### Improve support for resource registration

All resources that should be available at run time need to be listed at image build time. Currently, all listed resources are included in a non-structured form, i.e., it is not possible to navigate resources like a file system. Also resources are not separated by class loader, i.e., every class loader sees all resources.

The image heap must include a virtual file system for the included resources that allows navigation.

### Allow incomplete class path and all security services by default

[...]

PR #4305, opened 2022-02-09
[GR-36905] Implement --link-at-build-time and @<prop-values-file>.

graalvmbot:
Adds the following option to `native-image`:
```
    --link-at-build-time  require types to be fully defined at image build-time. If used
                          without args, all classes in scope of the option are required to
                          be fully defined.
```
Depending on where and how the option is used it behaves differently:
1) Using the option with args is always possible, e.g., `--link-at-build-time=foo.bar.Foobar,...`. The arguments are fully qualified class names, or package names (note: no package prefix names). 
  When used on the module path or class path (embedded in `native-image.properties` files) only classes and packages defined in the same .jar file can be specified.
  `@<prop-values-file>` can be used to redirect values for the option coming from a file relative to the place where the option is used.

2) When being used without args `--link-at-build-time` it depends on where it is used:
   * On the command line it means all classes should be seen as link-at-build-time classes
   * In modules (embedded in `native-image.properties` files) all classes defined in that module are treated as link-at-build-time classes
   * On classpath (embedded in `native-image.properties` files) an error message is show:
```
Error: Using '--link-at-build-time' without args only allowed on module-path. 'META-INF/native-image/org.mylibrary/native-image.properties' in 'file:///home/test/myapp/MyLibrary.jar' not part of module-path.
```

The PR also deprecates the option `--allow-incomplete-classpath`. Allowing an "incomplete classpath", i.e., link at image run time, is the new default and therefore does not need to be specified explicitly. The option is still allowed for compatibility reasons, but has no effect.

Motivation for this change (see also the discussions in https://github.com/oracle/graal/issues/2762#issuecomment-865784912, https://github.com/oracle/graal/discussions/3932#discussioncomment-1544294, and https://github.com/oracle/graal/issues/3929): Before this PR, support for an incomplete class path must be enabled by manually via `--allow-incomplete-classpath`. This ensures that no linking errors are thrown at run time. But since in practice many large projects needs to specify this option because optional dependencies are not always on the class path. So it is necessary to flip the default and enable the support by default.

Still, well-structured libraries should be able to do all linking at image build time, to avoid surprising errors at image run time. And well-structured applications should be able to do the same for all classes of the application. This is enabled by the new option `--link-at-build-time`. It is designed in a way so that libraries can only configure their own classes, i.e., avoid any side effects on other classes. The restrictions of the `--link-at-build-time` argument mentioned above ensure that.

When using the Java Platform Module System, it is easy for a library to link all its classes at image build time by specifying `--link-at-build-time` without listing classes and packages explicitly. But when using the class path, this is not possible because for .jar files on the class path there is no 1:1 mapping between packages and .jar files - multiple .jar files can contribute to the same package, and multiple .jar files can be merged into a single large .jar file. Therefore, packages for libraries used on the class path need to be listed explicitly. To make this process easier and allow, e.g., automatic generation of package lists, the `@<prop-values-file>` syntax can be used to provide the package list (or class list) in a separate file.


christianwimmer:
FYI @galderz @Sanne (I cannot add you as reviewers)

Sanne:
Many thanks for the huge progress!

Wondering though, is there any wayt to apply --link-at-build-time globally and yet allow specific exceptions?

I haven't had the chance to explore this in more depth yet, but from the description it seems I would need to set `--link-at-build-time` on the command line to have it apply the global effect, but not sure how I would then allow a particular library (and only that particular library) to relax the rule.

olpaw:
Hi @Sanne 

> Wondering though, is there any way to apply --link-at-build-time globally and yet allow specific exceptions?

That would not make sense. If you use `--link-at-build-time` on the command line (i.e. globally), every class the image builder sees during analysis will be required to be fully resolved. If you would like to only have individual classes or packages to be required to be fully resolved at build-time you have the following options:

1. Provide as many --link-at-build-time options **with specific arguments** on command line as you like. For example:
```
native-image --link-at-build-time=foo.bar,foobar.core.MyClass --link-at-build-time=my.other.RandomClass ...
```
2. If you are a library developer and you want to guarantee that none of the classes of your library will ever be unresolved at image-runtime (requiring all your library classes to be fully resolved at image build-time) you can add
```
--link-at-build-time=mylibrary.core,mylibrary.ext,mylibrary.aux,... etc
```
to the `Args = ` of the `native-image.properties` (in `META-INF/native-image/...`) that belongs to your library project.

Note that using `--link-at-build-time` **without arguments** in such context (as being part of a classpath entry) is **not** allowed. E.g. suppose you have `mylib.jar` and it contains `META-INF/native-image/org.example/mylib/native-image.properties` with `Args = --link-at-build-time` and you use it with `native-image -cp myapp:mylib.jar myapp.Main`, you would get the following error message:
```
Error: Using '--link-at-build-time' without args only allowed on module-path. 'META-INF/native-image/org.example/mylib/native-image.properties' in 'file:///home/pwoegere/mylib.jar' not part of module-path.
```
As the error messages indicates the image builder refuses jar files on classpath that use '--link-at-build-time' without args.

3. It is fine though to use  `--link-at-build-time` without args **with Java modules**. Assuming `mylib.jar` would be a Java module containing a module-info.class and it would be used only on the module-path. E.g. like this:
```
native-image --module-path mylib.jar:myapp.jar -m myapp
```
then the builder would accept that. In this case since `mylib.jar` is a Java module, '--link-at-build-time' is interpreted as: Require all classes in the module `mylib.jar` to be fully resolved at image build-time.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/LinkAtBuildTimeFeature.java
```diff
@@ -0,0 +1,169 @@
...
+import com.oracle.svm.core.option.SubstrateOptionsParser;
+import com.oracle.svm.core.util.UserError;
+import com.oracle.svm.hosted.jdk.ClassLoaderSupportFeatureJDK11OrLater;
+
+@AutomaticFeature
+public final class LinkAtBuildTimeFeature implements Feature {
+
+    static final class Options {
+        @APIOption(name = "link-at-build-time", defaultValue = "")//
+        @Option(help = "Require types to be fully defined at image build-time. If used without args, all classes in scope of the option are required to be fully defined.")//
```
vjovanov, line 61: I would somehow define the scope of the option. Maybe another sentence later.

olpaw:
@christianwimmer, re:

> When used on the module path or class path (embedded in native-image.properties files) only classes and packages defined in the same .jar file can be specified.

This is currently not enforced. If you use the `--link-at-build-time=foo.bar.Foobar,...` form embedded in `native-image.properties` files there is no checking if the given classes and packages are actually within the the jar where the `native-image.properties` is embedded in. Unfortunately it's not practical to require the above since in real-wolld configs you might have to specify `--link-at-build-time` with args for some library you depend on but that you do not have control over so that you could embed `--link-at-build-time` options in that library itself. Even this PR makes use of the fact that we do not enforce this. See https://github.com/oracle/graal/pull/4305/files#diff-98098d0b73f5a8109ce55af795a5cd911783234ccaadcb2346c9e6313c78fb50

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/LinkAtBuildTimeFeature.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private Object linkAtBuildTimeImpl(Class<?> clazz) {
+        if (requireCompleteAll) {
+            return reasonCommandLine;
+        }
+
+        if (clazz.isArray() || !classLoaderSupport.isNativeImageClassLoader(clazz.getClassLoader())) {
+            return "system default";
+        }
+        assert !clazz.isPrimitive() : "Primitive classes are not loaded via NativeImageClassLoader";
```
vjovanov, line 156: I would put this at the top of the method.

vjovanov (approved):
Elegant design. We can use the same approach for class initialization which will make it well-behaved.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/configure/ReflectionConfigurationParser.java
```diff
@@ -282,19 +280,15 @@ private String formatMethod(T clazz, String methodName, List<T> paramTypes) {
...
         String message = msg;
         if (cause != null) {
             message += " Reason: " + formatError(cause) + '.';
         }
-        if (allowIncompleteClasspath) {
-            System.err.println("Warning: " + message);
-        } else {
-            throw new JSONParserException(message + " To allow unresolvable reflection configuration, use option --allow-incomplete-classpath");
-        }
+        System.err.println("Warning: " + message);
```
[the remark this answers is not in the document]
olpaw: Since `allowIncompleteClasspath` doesn't exist anymore (i.e. is now `true` per default)
```java
if (allowIncompleteClasspath) {
System.err.println("Warning: " + message);
} else {
    throw new JSONParserException(message + " To allow unresolvable reflection configuration, use option --allow-incomplete-classpath");
}
```
the else branch became dead code.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/LinkAtBuildTimeFeature.java
olpaw: I can't. The `!classLoaderSupport.isNativeImageClassLoader(...)`-bit is the thing that ensures primitives are seen as `"system default"`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/LinkAtBuildTimeSupport.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public boolean linkAtBuildTime(Class<?> clazz) {
+        return feature.linkAtBuildTime(clazz);
+    }
+
+    public String errorMessageFor(ResolvedJavaType type) {
+        Class<?> clazz = ((OriginalClassProvider) type).getJavaClass();
+        if (clazz == null) {
+            return "This error is reported at image build time because class " + type.toJavaName(true) + " is registered for linking at image build time.";
```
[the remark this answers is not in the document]
olpaw: We can't. There is none.

It's just the fault behaviour that is responsible for that. I.e. if we have a `ResolvedJavaType` that has no corresponding Class instance we are dealing with a purely synthetic `ResolvedJavaType` created by the builder itself. Thus the only sane default here is to require such `ResolvedJavaType`s always to be fully resolved.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/SharedGraphBuilderPhase.java
```diff
@@ -189,9 +188,8 @@ protected void handleIllegalNewInstance(JavaType type) {
             if (allowIncompleteClassPath) {
```
zakkak, line 188: The comment above the `if` refers to the removed `--allow-incomplete-classpath` option

zakkak:
> Note that using `--link-at-build-time` without arguments in such context (as being part of a classpath entry) is not allowed. 

@olpaw my understanding is that this is by design in order to prevent libraries from forcing their consumers to link everything at build time. However, this was the default until now and it would be nice to have the option to enforce it even after this patch goes in. To prevent libraries from abusing it we could require it to be done programmatically instead.

>  That would not make sense. If you use `--link-at-build-time` on the command line (i.e. globally), every class that the image builder sees during analysis will be required to be fully resolved.

I think it makes sense if you want everything but a couple of classes/packages to be fully resolved at build time. Instead of having to create an exhaustive list of packages required to be resolved at build time, which might as well become incomplete in a future release, you would only need to define the classes/package that you don't require to be resolved at build time.

maxandersen:
Really happy to see movement on improving this! 

What I don't grok is how you foresee this change to be used in practice by frameworks like Quarkus, Micronaut, SpringNative etc. that has the need to extend/customize libraries that not yet have native-image configuration or for cases where a library config is not good for specific usecases.

With the proposed solution we now move from being able to explicit control to something where libraries/jars themself are the one to do customization and it is not possible to change that per library without modifying the jar or generate massive lists of classes/packages?

olpaw:
> I think it makes sense if you want everything but a couple of classes/packages to be fully resolved at build time. Instead of having to create an exhaustive list of packages required to be resolved at build time, which might as well become incomplete in a future release, you would only need to define the classes/package that you don't require to be resolved at build time.

That would be a full include/exclude design. The current PR does not have that.

olpaw:
> However, this was the default until now and it would be nice to have the option to enforce it even after this patch goes in.

`--link-at-build-time` on the command line gives you exactly that. Using it, makes native-image behave exactly as before this PR.

olpaw:
> With the proposed solution we now move from being able to explicit control to something where libraries/jars themself are the one to do customization and it is not possible to change that per library without modifying the jar or generate massive lists of classes/packages?

I could imagine a variant where we allow `--link-at-build-time` on command line with arguments where instead of package or class names we allow filenames. If filenames are passed that are used in either `-cp` or `-p`, The link-at-build-time logic would interpret it as if the classpath or modulepath entry would have specified `--link-at-build-time` for all it's containing classes.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/SharedGraphBuilderPhase.java
olpaw: Not really. Now the value of `allowIncompleteClassPath` here is determined per method that gets parsed via `!LinkAtBuildTimeSupport.singleton().linkAtBuildTime(method.getDeclaringClass())` in the `SharedBytecodeParser` constructor.

Sanne:
Ok, thanks for all clarifications - unfortunately it doesn't address what I was hoping for.

We probably have different context; which is totally fine but let me try challenge some points of the description for sake of avoiding misunderstandings:

[in context of previous flag `--allow-incomplete-classpath` this was said:]
>  But since in practice many large projects needs to specify this option because optional dependencies are not always on the class path. So it is necessary to flip the default and enable the support by default.

I would argue that Quarkus is a large project and our many users are building even larger projects; we regularly build non-trivial native-image applications and our CI systems are working non-stop with >1000 community libraries and 3rd party dependencies.

Of these thousands of libraries there is only ONE library which is abusing the `--allow-incomplete-classpath` flag by enforcing it, and I'm aware of ONE MORE library which makes it very tempting to resort to using this flag (as we're currently otherwise struggling to update it).
Honestly I'm not involved in all our extensions so it's possible that there are a couple more, but my point is that the actual need to allow incomplete classpath is not the norm, it rather is the exception.

In most cases in which the problem of "broken classpaths" manifests, there are typically better solutions - involving trivial refactoring in the code in question. 

The current problem is that when this exception materializes, it risks spreading: once the `--allow-incomplete-classpath` is used as workarounds for these extremely rare cases, it's hard to spot other regressions materializing.

[further]
> Still, well-structured libraries should be able to do all linking at image build time, to avoid surprising errors at image run time. And well-structured applications should be able to do the same for all classes of the application. 

I totally agree with this spirit. So we seem to agree that "good maintainers" should not need to use particular flags, if they only were somewhat interested in fixing such aspects they can achieve better effects with minor refactorings - then why introduce a flag?

The reason I brought the issue up is to allow integrating platforms to deal with technical debt, so to allow occasional one-off exceptions to the rule (of classpath must be fully defined) - but to do this effectively I would recommend:

 - allow exceptions to be defined by someone who did not package the jar
 - allow one-off exceptions on specific, tactical cases without affecting strict correctness of other code

The reasons should be clear: sometimes the maintainers don't get it, sometimes they are not interested / motivated, sometimes it's too late and the jar has shipped with an unintended integration issue. But we certainly want to keep encouraging towards designing "well-structured applications".

**Problem**
Beyond this patch not addressing my problem, I also have concerns: while some people had abused `--allow-incomplete-classpath` rather than fixing the problem, with the previous system we could workaround the issues at integration level. I am not sure we can actually maintain the benefits of not flipping the default in Quarkus and yet workaround the particular, exceptional offenders of well structured code w/o significant effort such as scanning and listing all classes of the application.

maxandersen:
to try and illustrate @Sanne 's concern (which I agree with btw) as how it would be nice to do is the following:

`--link-all-at-build-time -no-link-at-build-time oraclejdbc.jar`

i.e. enable linking by default and then specify the few exceptions.

That is the common case - we don't have tons of jars that needs incomplete classpath; it's the direct opposite.

Yes we could probably work around this by generating a file listing all the jars/classes/packages
but that seems quite inefficient and error-prone.

olpaw:
> I am not sure we can actually maintain the benefits of not flipping the default in Quarkus and yet workaround the particular, exceptional offenders of well structured code w/o significant effort such as scanning and listing all classes of the application.

@Sanne, if we implement suggestion https://github.com/oracle/graal/pull/4305#issuecomment-1057133814 on top of this PR, then this is just a matter of:
```java
void buildImage(List<Path> libs, Set<Path> libsThatRequireAllowIncompleteClassPath) {
    List<Path> linkAtBuildTimeLibs = new ArrayList(libs);
    linkAtBuildTimeLibs.removeAll(libsThatRequireAllowIncompleteClassPath);
    runNativeImage(
        "-cp", String.join(File.pathSeparator, libs),
        "--link-at-build-time", linkAtBuildTimeLibs,
        ....);
}
```
Do you consider that usable?

cc @christianwimmer

Sanne:
@olpaw yes I think something among those lines would be great. Thanks for the consideration!

But in terms of encouraging the wider community of users to structure their code optimally, wouldn't it be preferreable to make this easy to any GraalVM user?

christianwimmer:
@Sanne @maxandersen You have to factor in the reality of the Java specification here. It would be a very intrusive specification change in Java to mandate, or even allow, linking errors to be thrown before the first execution of the failing bytecode. So we must switch the default. That is unfortunately not negotiable.

Now the question is: how can we still get as much checking as possible at image build time, given this constraint. From all of our discussions, I got that you still want a way for Quarkus to preserve the old behavior, i.e., enforcing all linking at image build time. That is exactly what you get when specifying `--link-at-build-time` without any arguments on the command line. So in cases where Quarkus did not add `--allow-incomplete-classpath` (or removed `--allow-incomplete-classpath`from dependencies), that option can be specified.

Now Quarkus can, e.g., maintain a list of extensions for which it can inject a `--link-at-build-time` into a library, or even better encourage that library to specify it. But I don't think an option `--link-at-run-time` is useful, because it would only increase the mess and not reduce it. If  a library explicitly configures itself as `--link-at-build-time`  and then linking at build time does not work, I would consider that a real bug in that library. And it is a bug that is trivial to fix by (at least temporarily) removing the `--link-at-build-time` again.

maxandersen:
Maybe I'm missing something but how do you foresee Quarkus injects --link-at-runtime into a library ? By modifying the jar ?

I get you want to change the default but why make it so you as a Graalvm user must use a framework to get the behavior where buildtime init is easy to have happen ?

christianwimmer:
> how do you foresee Quarkus injects --link-at-runtime into a library?

There is no need to inject it into a library. Quarkus controls the `native-image` command line, so it can just put `--link-at-runtime=` for any packages or classes it wants on the command line.

maxandersen:
Okey, so I understood it right. That can be quite a long list to do and for sure it means users wanting the highest performance will be forced to use something like Quarkus to manage it. You can say that is great for Quarkus, but why not make this easy to manage when just using GraalVM ? 

I.e. make option to default to link at build for everything and then mark which jars cannot do it ?

olpaw:
> @olpaw yes I think something among those lines would be great. Thanks for the consideration!
> 
> But in terms of encouraging the wider community of users to structure their code optimally, wouldn't it be preferreable to make this easy to any GraalVM user?

Implemented as 3f1d5c89fd29f29c4d863c832ea745f4f2c9e988

christianwimmer:
@maxandersen 

As I already said making it the default to link at build time is not possible. And having two sets of options (link at build time and link at run time for packages and classes) is way too confusing.

Also, you mention "performance" in your comment. Note that there is no difference in any performance metric (peak performance, startup performance, memory footprint, ...) between linking at build time and linking at run time. Because "linking at run time" actually must means that direct calls to throw linkage errors are compiled in when an element is not available - there is still no performance overhead at run time.

maxandersen:
Yes sorry. My comment about performance was incomplete. What I mean is that if we delay failure to runtime Users will have to do a slow native build; then run to then see the failure...that is slow compared to eagerly fail and inform user of the issue. 

As said with the adjustments to the pr and comments here we can in quarkus "simulate" build time first by listing all jars. 

I just find it a big loss for the java ecosystem and graalvm end user developer usability In particular that everyone has to calculate that jar list rather than have an *option* to turn it on for all by default and be able to list exceptions.

n1hility:
> @Sanne @maxandersen You have to factor in the reality of the Java specification here. It would be a very intrusive specification change in Java to mandate, or even allow, linking errors to be thrown before the first execution of the failing bytecode. So we must switch the default. That is unfortunately not negotiable.
> 

@christianwimmer I can certainly appreciate the decision for a lazy default. The reason we prefer to implement a non-default approach, and enable eager processing using the integration points you are offering  is usability related. I argue, it's much better to catch problems up front than it is to blow up at runtime after some unlucky call path.  If it is an intentional lazy unreachable code block thats fine, however, this also defers actual errors (e.g. missing or mismatched dependency), which IMO are more common.

Regarding the specification, often code intentionally using an incomplete class path strategy (e.g. lazily reachable broken symbols) is actually non-portable since 5.4 allows for JVMs to utilize eager linking. IMO broken symbols is a practice that should be replaced with proper compliant decoupling (E.g. ServiceLoader).  That said, I do understand wanting to support it, since such usage works fine on OpenJDK, even if it is non-portable/compliant.

Also IMO, the elements that specify linkage errors being thrown at specific code points do not really make sense with a native compilation output. The spirit and intention was around dynamic loading of class bytecode which just doesn't exist in the case of a precompiled executable. Whenever we get a chance to look at a formal Leyden spec, I think this is one of those things we should fix. IMO GraalVM should be free to innovate and not tie its hands to outmoded concepts. Although, I understand that you can't rely on that happening. I just mention that for Java to evolve, we need to be open to better ideas.

PR #4316, opened 2022-02-11
Don't prepend resource URLs with '/'

zakkak:
In PR #4281 resource URLs started being prefixed with '/', which results
in failures to retrieve the resource in some cases.
E.g. https://github.com/quarkusio/quarkus/issues/23604

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
@@ -141,7 +141,7 @@ public final class Resources {
     private static URL createURL(String moduleName, String resourceName, int index) {
         try {
             String refPart = index != 0 ? '#' + Integer.toString(index) : "";
-            return new URL(JavaNetSubstitutions.RESOURCE_PROTOCOL, moduleName, -1, '/' + resourceName + refPart);
+            return new URL(JavaNetSubstitutions.RESOURCE_PROTOCOL, moduleName, -1, resourceName + refPart);
         } catch (MalformedURLException ex) {
             throw new IllegalStateException(ex);
         }
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/ResourceURLConnection.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/resources/ResourceURLConnection.java
@@ -44,6 +44,10 @@ public class ResourceURLConnection extends URLConnection {
         super(url);
     }
 
+    private static String resolveName(String resourceName) {
+        return resourceName.startsWith("/") ? resourceName.substring(1) : resourceName;
+    }
+
     @Override
     public void connect() {
         if (connected) {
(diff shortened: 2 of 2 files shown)


zakkak:
@olpaw can you please have a look at this?
It is causing the Quarkus CI to fail with 22.1-dev. Thank you.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
```diff
@@ -141,7 +141,7 @@ public static ResourceStorageEntry get(String moduleName, String resourceName) {
     private static URL createURL(String moduleName, String resourceName, int index) {
         try {
             String refPart = index != 0 ? '#' + Integer.toString(index) : "";
-            return new URL(JavaNetSubstitutions.RESOURCE_PROTOCOL, moduleName, -1, '/' + resourceName + refPart);
+            return new URL(JavaNetSubstitutions.RESOURCE_PROTOCOL, moduleName, -1, resourceName + refPart);
```
olpaw, line 144: @zakkak, we cannot simply omit the `'/'` separator. Otherwise we will end up with URLs like e.g.
```
resource://jdk.compilermodule-info.class
```
where it actually should be
```
resource://jdk.compiler/module-info.class
```
which will break the ability to create resource URLs from their string representations. Something that works fine with URLs created from resource access when running on the JVM. We therefore have to support that as well.

Consequently this change causes out gates to fail with
```
1) com.oracle.svm.test.NativeImageResourceFileSystemProviderTest#testURLExternalFormEquivalence
java.lang.AssertionError: Contents of original URL and one created from originals ExternalForm must be the same: java.io.FileNotFoundException: resource://jdk.compilermodule-info.class
	at com.oracle.svm.test.NativeImageResourceFileSystemProviderTest.testURLExternalFormEquivalence(NativeImageResourceFileSystemProviderTest.java:631)
	at java.lang.reflect.Method.invoke(Method.java:568)
	at com.oracle.mxtool.junit.MxJUnitWrapper.runRequest(MxJUnitWrapper.java:368)
	at com.oracle.svm.junit.SVMJUnitRunner.run(SVMJUnitRunner.java:217)
	at com.oracle.svm.junit.SVMJUnitRunner.main(SVMJUnitRunner.java:237)

FAILURES!!!
Tests run: 61,  Failures: 1
```
You can test this locally with
```
mx build
mx gate --omit-clean -t 'native unittests'
```
from the `substratevm` subdir.

Could it be that the failing quarkus tests need adjustment? Can you create a small reproducer of what actually fails on your side?

zakkak: Thanks for the detailed explanation @olpaw . Further investigation revealed that the breakage was indeed not caused by the leading slash (at least for one of the cases), but it was unveiled by it :).

I opened https://github.com/oracle/graal/issues/4326 for the underlying issue and prepared a draft PR for a fix (but I was not able to write a unit test for it yet, I have some trouble registering a directory as a resource) https://github.com/oracle/graal/pull/4327.

zakkak:
Closing as invalid fix.

PR #4327, opened 2022-02-15
[GR-37154] Fix handling of resources with trailing slashes

### LINKED ISSUE #4326: GraalVM's native-image substitute for `ClassLoader.getResource` normalizes resources' names while original doesn't
**Describe the issue**
GraalVM's native-image substitute for `ClassLoader.getResource` normalizes resources' names while the original does not.

Running
```
        URL url = cl.getResource("META-INF/");
        URL url2 = cl.getResource("META-INF/./");
        System.out.println("url = " + url);
        System.out.println("url2 = " + url2);
```

in JVM results in:
```
url = jar:file:/home/zakkak/code/tmp/issue-reproducers/target/reproducer-1.0-SNAPSHOT.jar!/META-INF/
url2 = null
```

while in native-image it results in:
```
url = resource:META-INF
url2 = resource:META-INF
```

I would expect it to return:
```
url = resource:META-INF/
url2 = null
```
instead.

cc @jovanstevanovic @vjovanov as it relates to https://github.com/oracle/graal/pull/3597

```
cd /tmp
git clone --branch resources-with-trailing-slash https://github.com/zakkak/issue-reproducers reproducers
cd reproducers
mvn package
export JAVA_HOME=/opt/jvms/graalvm-ce-java11-22.0.0.2
$JAVA_HOME/bin/java \
  -jar target/reproducer-1.0-SNAPSHOT.jar
$JAVA_HOME/bin/native-image \
  -jar target/reproducer-1.0-SNAPSHOT.jar
./reproducer-1.0-SNAPSHOT
```

 - GraalVM version: **9b25754d49ca0fec016b1f4c03999f09fdb07404**
 - JDK major version: **11**
 - OS: **Fedora 35**
 - Architecture: **AMD64**

zakkak:
Fix #4326

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
@@ -29,7 +29,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
@@ -43,7 +42,6 @@ import org.graalvm.nativeimage.Platforms;
 import org.graalvm.nativeimage.hosted.Feature;
 
 import com.oracle.svm.core.annotate.AutomaticFeature;
-import com.oracle.svm.core.jdk.resources.NativeImageResourcePath;
 import com.oracle.svm.core.jdk.resources.ResourceStorageEntry;
 import com.oracle.svm.core.util.ImageHeapMap;
 import com.oracle.svm.core.util.VMError;
@@ -85,9 +83,13 @@ public final class Resources {
         }
     }
 
+    private static String getResourceWithoutTrailingSlash(String name) {
+        return name.endsWith("/") ? name.substring(0, name.length() - 1) : name;
+    }
+
     private static void addEntry(String moduleName, String resourceName, boolean isDirectory, byte[] data) {
         Resources support = singleton();
-        Pair<String, String> key = Pair.create(moduleName, resourceName);
+        Pair<String, String> key = Pair.create(moduleName, getResourceWithoutTrailingSlash(resourceName));
         ResourceStorageEntry entry = support.resources.get(key);
         if (entry == null) {
             entry = new ResourceStorageEntry(isDirectory);
@@ -121,21 +123,17 @@ public final class Resources {
         addEntry(moduleName, resourceDirName, true, content.getBytes());
     }
 
-    /**
-     * Avoid pulling native file system by using {@link NativeImageResourcePath} implementation to
-     * convert <code>resourceName</code> to canonical variant.
-     */
-    public static String toCanonicalForm(String resourceName) {
-        NativeImageResourcePath path = new NativeImageResourcePath(null, resourceName.getBytes(StandardCharsets.UTF_8), true);
-        return new String(NativeImageResourcePath.getResolved(path));
-    }
-
     public static ResourceStorageEntry get(String name) {
-        return singleton().resources.get(Pair.createRight(name));
+        return get(null, name);
     }
 
     public static ResourceStorageEntry get(String moduleName, String resourceName) {
-        return singleton().resources.get(Pair.create(moduleName, resourceName));
+        ResourceStorageEntry resourceStorageEntry = singleton().resources.get(Pair.create(moduleName, getResourceWithoutTrailingSlash(resourceName)));
+        if (resourceStorageEntry != null && (resourceStorageEntry.isDirectory() || !resourceName.endsWith("/"))) {
+            return resourceStorageEntry;
+        } else {
+            return null;
+        }
     }
 
     private static URL createURL(String moduleName, String resourceName, int index) {
@@ -156,7 +154,7 @@ public final class Resources {
             return null;
         }
 
-        Enumeration<URL> urls = createURLs(moduleName, toCanonicalForm(resourceName));
+        Enumeration<URL> urls = createURLs(moduleName, resourceName);
         return urls.hasMoreElements() ? urls.nextElement() : null;
     }
 
@@ -170,7 +168,7 @@ public final class Resources {
             return null;
         }
 
-        ResourceStorageEntry entry = Resources.get(moduleName, toCanonicalForm(resourceName));
+        ResourceStorageEntry entry = Resources.get(moduleName, resourceName);
         if (entry == null) {
             return null;
         }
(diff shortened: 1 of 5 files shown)


zakkak:
Note: this draft allows resolving files from resourceNames with trailing slashes which is not desirable.

zakkak:
> I am keeping this as draft till I get some tests in place to ensure this doesn't break again in the future.

It appears that `mx` does not create `JarEntry`s for directories when generating the `svm-tests.jar` and thus I cannot register a directory as resource in order to implement the tests. @olpaw do you have any suggestions on how to work around this issue? Is there any way to force `mx` to create `JarEntry`s in the `svm-tests.jar`?

zakkak:
This is now ready for review. In order for the new tests to work `mx` needs to be patched with

```diff
diff --git a/mx_jardistribution.py b/mx_jardistribution.py
index 42e6c3f..c098fe6 100755
--- a/mx_jardistribution.py
+++ b/mx_jardistribution.py
@@ -998,6 +998,7 @@ class _ArchiveStager(object):
             def add_classes(archivePrefix, includeServices):
                 for root, _, files in os.walk(outputDir):
                     reldir = root[len(outputDir) + 1:]
+                    self.bin_archive.entries[reldir+'/'] = reldir+'/'
                     for f in files:
                         relpath = join(reldir, f)
                         self.add_file(dep, outputDir, relpath, archivePrefix, arcnameCheck=overlay_check, includeServices=includeServices)
```

I don't know if this change would be welcome on `mx` since it will affect all jars. Please advice on how to proceed.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
```diff
@@ -186,33 +184,36 @@ public static InputStream createInputStream(String moduleName, String resourceNa
...
 
         List<URL> resourcesURLs = new ArrayList<>();
 
-        /* If moduleName was unspecified we have to consider all modules in the image */
+        ResourceStorageEntry explicitEntry = Resources.get(moduleName, resourceName);
+        addURLEntries(resourcesURLs, explicitEntry, moduleName, resourceName);
+        /*
+         * If moduleName was unspecified we have to consider all modules in the image. We append
+         * those after the explicit entry in order to prioritize the explicit entry in case it
+         * exists
```
zakkak, line 195: @olpaw FYI

This change is a bit "out of scope" so I thought an explicit mention would help.

What it essentially does is to prioritize `resource:META-INF` over `resource:mymodule/META-INF` in case both are available and one requests `META-INF` without specifying a module.

olpaw: Thanks. I will have a look next week.

olpaw:
> @olpaw do you have any suggestions on how to work around this issue? Is there any way to force `mx` to create `JarEntry`s for directories in the `svm-tests.jar`?

After spending way too much time to find something appropriate I now think the best would be to add this to mx. But not like suggested in https://github.com/oracle/graal/pull/4327#issuecomment-1042317861

Instead i would make the mentioned code conditional of a special filename (e.g. `.empty_dir`). If we have such an entey the archiver should not add the file but instead the directory that contains file. @gilles-duboscq are you ok with such a change in mx?

gilles-duboscq:
i think that's fine

olpaw:
@zakkak please prepare the mx PR that we need for this PR and add me and @gilles-duboscq as reviewers.

zakkak:
Thanks for the input @olpaw !

> @zakkak please prepare the mx PR that we need for this PR and add me and @gilles-duboscq as reviewers.

Done. This PR is now blocked by https://github.com/graalvm/mx/pull/256

olpaw:
@zakkak your mx PR got merged. Now you need to update this PR to make substratevm depend on the version of mx that contains your change. I.e. adjust https://github.com/oracle/graal/blob/82be5fb5d182c2980c866e6929ac27b438fe130f/substratevm/mx.substratevm/suite.py#L3 accordingly.

zakkak:
@olpaw `mxversion` is now up to date.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Resources.java
olpaw: > What it essentially does is to prioritize resource:META-INF over resource:mymodule/META-INF in case both are available and one requests META-INF without specifying a module.

Have you tested if this is the same prioritization that would happen when running on the jvm?

zakkak: Good question! No I did not.

zakkak: According to my tests JVM follows the old prioritization.

So I guess we should revert this part.

Wold you like me to update the PR @olpaw ?

olpaw: Yes, please.

olpaw: @zakkak ideally you would also add a test that checks that the order is identical to the order on jvm.

zakkak: > Yes, please.

Done

> @zakkak ideally you would also add a test that checks that the order is identical to the order on jvm.

I'll give that a try. However, since implementing the tests might take some time and what they test is unrelated to this PR I would like you to proceed with reviewing/merging this PR while I work on them as https://github.com/oracle/graal/issues/4326 is causing our pipelines to fail.

Are there any tests doing similar checks, i.e. comparing the jvm behavior to that of svm (so that I don't reinvent the wheel)?

Thanks

olpaw: > Are there any tests doing similar checks, i.e. comparing the jvm behavior to that of svm (so that I don't reinvent the wheel)?

Yes the whole `NativeImageResourceFileSystemProviderTest` tests are like that. Either run them on the JVM with
```
mx unittest --very-verbose com.oracle.svm.test.NativeImageResourceFileSystemProviderTest
```
or run them as native-image via
```
mx native-unittest ...
```
that's how I verify that our implementations behave exactly like the jvm.

olpaw: > PR I would like you to proceed with reviewing/merging this PR while I work on them as https://github.com/oracle/graal/issues/4326 is causing our pipelines to fail.

OK.

olpaw (approved):

PR #4330, opened 2022-02-15
Implementation of Safepoint events

zhengyu123:
Add implementation of Safepoint event that provides safepoint latency information.

e.g. 

```
jdk.SafepointBegin {
  startTime = 15:43:26.488
  duration = 4.315 us
  safepointId = 2
  totalThreadCount = 5
  jniCriticalThreadCount = 0
  eventThread = "main" (javaThreadId = 1)
}

jdk.SafepointEnd {
  startTime = 15:43:26.490
  duration = 0.318 us
  safepointId = 2
  eventThread = "main" (javaThreadId = 1)
}


```

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Safepoint.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Safepoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -57,6 +57,9 @@ import com.oracle.svm.core.annotate.Uninterruptible;
 import com.oracle.svm.core.graal.nodes.KillMemoryNode;
 import com.oracle.svm.core.heap.Heap;
 import com.oracle.svm.core.jdk.UninterruptibleUtils;
+import com.oracle.svm.core.jfr.JfrTicks;
+import com.oracle.svm.core.jfr.events.SafepointBeginEvent;
+import com.oracle.svm.core.jfr.events.SafepointEndEvent;
 import com.oracle.svm.core.log.Log;
 import com.oracle.svm.core.nodes.CFunctionEpilogueNode;
 import com.oracle.svm.core.nodes.CFunctionPrologueNode;
@@ -612,6 +615,7 @@ public final class Safepoint {
         protected boolean freeze(String reason) {
             assert SubstrateOptions.MultiThreaded.getValue() : "Should only freeze for a safepoint when multi-threaded.";
             assert VMOperationControl.mayExecuteVmOperations();
+            long startTicks = JfrTicks.elapsedTicks();
 
             /* the current thread may already own the lock for non-safepoint reasons */
             boolean lock = !VMThreads.THREAD_MUTEX.isOwner();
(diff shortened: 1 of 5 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/TimedEvent.java
```diff
@@ -0,0 +1,38 @@
...
+package com.oracle.svm.core.jfr.events;
+
+import com.oracle.svm.core.annotate.Uninterruptible;
+import com.oracle.svm.core.jfr.HasJfrSupport;
+import com.oracle.svm.core.jfr.JfrTicks;
+
+public class TimedEvent {
+    @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
+    public static long getTicks() {
+        return HasJfrSupport.get() ? JfrTicks.elapsedTicks() : 0;
```
christianhaeubl, line 36: Seems that this pattern will be pretty common.
I think it makes sense to move the `HasJfrSupport.get()` check directly into `JfrTicks.elapsedTicks()` (`JfrTicks` is a helper class anyways). Then, it is possible to unconditionally call `JfrTicks.elapsedTicks()`, regardless if the JFR support is present in the image or not.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Safepoint.java
```diff
@@ -657,15 +662,17 @@ private static boolean isMyself(IsolateThread thread) {
...
             VMThreads.THREAD_MUTEX.assertIsOwner("Must hold mutex while requesting a safepoint.");
             final Log trace = Log.noopLog().string("[Safepoint.Master.requestSafepoints:  reason: ").string(reason);
+            int numOfThreads = 0;
 
             // Walk the threads list and ask each thread (except myself) to come to a safepoint.
             for (IsolateThread vmThread = VMThreads.firstThread(); vmThread.isNonNull(); vmThread = VMThreads.nextThread(vmThread)) {
                 if (isMyself(vmThread)) {
                     continue;
                 }
+                numOfThreads++;
```
christianhaeubl, line 675: Shouldn't this count also include the current thread? This code is always executed by the VM operation thread but the VM operation thread is a normal Java thread in Native Image. We support two different execution modes (see `VMOperationControl.useDedicatedVMOperationThread()`):
a. Without a dedicated VM operation thread, any normal Java thread can be temporarily promoted to the VM operation thread.
b. If a dedicated VM operation thread should be used, then a dedicated Java thread is started early on and only that thread may execute VM operations.

christianhaeubl (commented):
Thanks again for contributing, I added a small number of comments.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Safepoint.java
zhengyu123: I believe it means how many threads were brought to the safepoint. In hotspot, it does not count vmThread.

christianhaeubl: HotSpot reports the number of Java threads, so I think we should the same.
However, in our case that includes the VM operation thread because it is a Java thread as well. This is a major difference to HotSpot anyways and affects all JFR areas so I think it is best to consistently report the VM operation thread as a normal Java thread (e.g., the VM operation thread can allocate Java heap memory, so it will also show up in all allocation-related JFR events).

zhengyu123: Okay, fixed

christianhaeubl: Thanks, this PR will also be merged to Graal master in the next couple days.

PR #4333, opened 2022-02-16
Prerequisites for Windows debug typeInfo implementation

stooke:
This PR adds some methods to MethodEntry (getVtableOffset(), isFirstIntroduction()) and cleans up some of the Windows CodeView code.
The changes in this PR were previously in #3732, but that PR will be condensed to only contain specific typeInfo implementation.

This was requested by @pejovica , so I suggest he be the reviewer.


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSectionImpl.java
```diff
@@ -186,4 +186,27 @@ int getCurrentOffset() {
...
+            markRelocationSite(pos, ObjectFile.RelocationKind.SECREL_4, symbolName, offset);
+            pos += ObjectFile.RelocationKind.getRelocationSize(ObjectFile.RelocationKind.SECREL_4);
+            markRelocationSite(pos, ObjectFile.RelocationKind.SECTION_2, symbolName, 0);
+            pos += ObjectFile.RelocationKind.getRelocationSize(ObjectFile.RelocationKind.SECTION_2);
+        } else {
+            pos = CVUtil.putInt(0, buffer, pos);
+            pos = CVUtil.putShort((short) 0, buffer, pos);
+        }
+        return pos;
+    }
```
pejovica, line 211: I'd remove the `offset` parameter because it seems that we don't need it for now (we always pass `0`).

We could also simplify the implementation a bit using a helper method like this:
```suggestion
    public int markRelocationSite(byte[] buffer, int initialPos, String symbolName) {
        int pos = initialPos;
        pos += markRelocationSite(buffer, pos, symbolName, RelocationKind.SECREL_4);
        pos += markRelocationSite(buffer, pos, symbolName, RelocationKind.SECTION_2);
        return pos;
    }

    private int markRelocationSite(byte[] buffer, int initialPos, String symbolName, RelocationKind k) {
        if (buffer != null) {
            markRelocationSite(initialPos, k, symbolName, 0);
        }
        return initialPos + RelocationKind.getRelocationSize(k);
    }
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsection.java
```diff
@@ -66,7 +66,7 @@ protected int computeContents(byte[] buffer, int initialPos) {
     public void logContents(DebugContext debugContext) {
         CVSectionImpl section = cvDebugInfo.getCVSymbolSection();
         for (CVSymbolSubrecord subcmd : subcmds) {
-            section.log(debugContext, "     [0x%08x]  %s", subcmd.getPos(), subcmd.toString());
+            section.log("     [0x%08x]  %s", subcmd.getPos(), subcmd.toString());
```
pejovica, line 69: I think you can also drop `debugContext` from the method signature.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -55,9 +53,8 @@
      * The CodeView symbol section Prolog is also a CVSymbolSubsection, but it is not build in this
      * class.
      */
-    void build(DebugContext theDebugContext) {
-        this.debugContext = theDebugContext;
-        this.lineRecordBuilder = new CVLineRecordBuilder(debugContext, cvDebugInfo);
+    void build() {
+        this.lineRecordBuilder = new CVLineRecordBuilder(cvDebugInfo);
```
pejovica, line 57: Initialization could now be moved to the constructor.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1190,6 +1242,23 @@ private void computeFullFilePath() {
...
+        public boolean isFirstIntroduction() {
+            assert false;
+            return false;
+        }
+
+        @Override
+        public boolean isConstructor() {
+            assert false;
+            return false;
+        }
```
pejovica, line 1261: Perhaps you could make these implementations unnecessary by introducing a new superinterface, say `DebugMethodInfoBase`, for APIs shared by both `DebugMethodInfo` and `DebugRangeInfo`. Then the new APIs could be part of `DebugMethodInfo` only.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsection.java
stooke: Done.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
stooke: Done.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
stooke: I see where you're coming from, but this interface is under development by @adinn , and he is currently working on it for an upcoming PR.  Andrew Dinn, do you have any input here?  Putting in these dummy methods usually indicates a different abstraction is required.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSectionImpl.java
stooke: Done, though personally my taste is to keep the offset param.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: The change I am working on for adding local var info does include some [refactoring](https://github.com/adinn/graal/commit/29836dace5a3aee115e23c3bc147f96ddf2e8b72) of the debugInfo interfaces. I also refactored the implementation hierarchy to remove a lot of duplication and the implementation change did indeed  a common superclass. However, there was no reason to add a common super interface. The only real change is to fix interface DebugRangeInfo which was introduced rather hastily and got some things at the wrong level:

  - method ownerType() of DebugRangeInfo is moved up to interface DebugMethodInfo.
  - methods addressLo(), addressHi() and line() are moved up from DebugCodeInfo and DebugLocationInfo (was DebugLineInfo) to DebugRangeInfo

I am not clear that there is a need for a common superinterface. The rationale for the current interface organization proceeds as follows:

1. DebugMethodInfo provides information about a HostedMethod referenced from the type model (i.e. the list of HostedType objects needed in image)
2. DebugCodeInfo provides information about a top-level compiled method as derived from the compiled method base (i.e. the list of CompilationResult objects needed in image). Each compilation is associated with a HostedMethod and, therefore, can also provide all the information that DebugMethodInfo provides.
3. DebugLocation/LineInfo provides information about source positions referenced from the compiled method base each of which is either a top-level or inline compiled method and is associated with a ResolvedJavaMethod. That latter object may or may not be a HostedMethod. However, the DebugLocation/LineInfo  object can still provide all the information that DebugMethodInfo provides -- so long as it is careful not to conflate properties of the HostedMethod with properties of the underlying ResolvedJavaMethod (n.b. identity is established by unwrapping any HostedMethod to its original ResolvedJavaMethod).

There are a few important points to note:

Not every method presented via interface DebugMethodInfo will also be presented via interface DebugCodeInfo (but the converse holds).

Not every method presented via interface DebugLocation/LineInfo will have been presented via interface DebugMethodInfo (nor, by obvious inference, via DebugCodeInfo). This is why the generator occasionally has to add extra MethodEntry records during processing of Location/Line data.

So, the assumption behind the interface hierarchy is that all information provided via interface 1 should also be available via interfaces 2 or 3 and consistent between these different presentations. That is why a superinterface is not needed. Looking at Simon's additions to DebugMethodInfo I am not sure we need to revise that assumption:

 - isConstructor ought to be defined on DebugMethodInfo as it is a property of a method however it is manifested
  - vtableOffset ought to be defined on DebugCodeInfo as it only appropriate for methods which have been top level compiled. So, a MethodEntry ought to default to not having a vtable offset when it is created. This can then be updated during processing of DebugCodeInfo instances.
 - isFirstIntroduction also ought to be defined on DebugCodeInfo as it is only appropriate for methods which have been top level compiled (it is to do with accumulation of vtable entries) . As with the previous case associated the MethodEntry flag can default to clear and be set as needed during processing of DebugCodeInfo instances.

stooke: Thanks for the excellent explanation, @adinn !  I've taken your comments to heart and refactored this code - I hope the new version addresses @pejovica 's concerns.
A shame to press 'resolve conversation' as it will hide your comment.

adinn (approved):
Ok, with the latest change to the generic debug info model this looks fine to me.

pejovica:
@stooke As discussed, each `HostedMethod` has a list of implementations, which is essentially a list of overrides, so it is possible to construct a set of all overrides. For example, this could be done in the `NativeImageDebugInfoProvider` constructor like this:
```java
Set<HostedMethod> allOverrides = heap.getUniverse().getMethods().stream()
                .filter(HostedMethod::hasVTableIndex)
                .flatMap(m -> Arrays.stream(m.getImplementations())
                .collect(Collectors.toSet());
```
Which can then be used to implement `isOverride` as a membership test.

stooke:
@pejovica I've refactored the isOverride() code as discussed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
pejovica: @adinn  Thanks for the great explanation!

I've given some more thought to this and I don't think we can move the vtable APIs to the `DebugCodeInfo` interface. The problem is that we would then lack information for abstract methods because they are not presented through this interface but only through `DebugMethodInfo`.

In principle, we need accurate information for each `MethodEntry` in `ClassEntry.methods`, regardless of the interface through which `MethodEntry` was created. So I think we should move these APIs back to `DebugMethodInfo` and ensure proper implementations, which shouldn't be a problem except when `DebugLineInfo` isn't based on `HostedMethod`. However, in that case I think it's safe to assume that such methods cannot have information about vtable.

stooke: @pejovica I understand your concern, but I suggest we wait until we properly handle interfaces too, as they might have a similar issue as abstract functions.

adinn: @pejovica Ok, I see the problem and I agree that this means we need to migrate `vtableOffset` up to `DebugMethodInfo` and adjust the implementation methods accordingly.

I believe your assumption about the implications for `DebugLineInfo` is correct but your explanation actually misses an important nuance. If a `DebugLineInfo` record presents information about a ResolvedJavaMethod` that is not a `HostedMethod` this does not guarantee that there is no associated `HostedMethod`. We are still occasionally presented with cases where this happens. However, the important point to note is that in this situation we should never need to update the vtable index.

We only need a vtable index when the unwrapped original corresponds to a compiled method or an abstract parent method of some compiled method. In either of those cases we should already have seen that method via a `DebugMethodInfo` instance. Tht is because when a method is compiled in or is a parent of a method that gets compiled in then it must be part of the analysis closure. Its owner class must be in the type list that drives generation of `DebugTypeInfo` records. Since the method in question forms part of the live referenced method hierarchy the associated method cannot have been erased from the type and so will drive generation of a `DebugMethodInfo` instance.

So, the implementation of `DebugLineInfo` can simply return -1 for this case (or indeed for all cases) with the following proviso: every client of DebugLineInfo must recognises that a vtable index value of -1 returned does not override a prior index retruned for the same method presented via a DebugMethodInfo instance.

stooke: @pejovica , @adinn I'm happy to put back the vtableOffset() (and hasVtable()) methods to DebugMethodInfo, if there is consensus. There are also isContructor() and isOverload(); should those move back to DebugMethodInfo?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSectionImpl.java
stooke: @pejovica the use case for the offset parameter is when defining static member objects; they are defined as an offset from svm_heap.  I will keep the non-offset API (as it is the most common case) and add an API that has an offset parameter.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Regarding `isConstructor`, that was indeed what i said earlier:

>  - `isConstructor` ought to be defined on`DebugMethodInfo` as it is a property of a method however it is manifested

By the same reasoning I believe it would be appropriate to define `isOverload` on `DebugMethodInfo`.

stooke: @pejovica , @adinn I've returned isContructor(), isOverload(), hasVtable() and vtableOffset() to DebugMethodInfo, I kept them on DebugCodeInfo, since they should be correct there - let me know if that's an invalid assumption.  On DebugLineInfo, they will emit a GraalError.unimplemented(), giving us a chance to catch any incorrect usage.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSectionImpl.java
```diff
@@ -186,4 +186,42 @@ int getCurrentOffset() {
...
+     * @param buffer output buffer
+     * @param initialPos position of fixup in output buffer
+     * @param symbolName symbolname to reference
+     * @return new position in output buffer
+     */
+    public int markRelocationSite(byte[] buffer, int initialPos, String symbolName) {
+        int pos = initialPos;
+        pos = markRelocationSite(buffer, pos, symbolName, RelocationKind.SECREL_4, 0);
+        pos = markRelocationSite(buffer, pos, symbolName, RelocationKind.SECTION_2, 0);
+        return pos;
```
pejovica, line 202: Could be simplified a bit.
```suggestion
        return markRelocationSite(buffer, pos, symbolName, 0);
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/MethodEntry.java
```diff
@@ -153,6 +179,14 @@ public void updateRangeInfo(DebugInfoBase debugInfoBase, DebugMethodInfo debugMe
         }
     }
 
+    public boolean getIsVirtual() {
```
pejovica, line 182: I'd rename this to `isVirtual`.
```suggestion
    public boolean isVirtual() {
```

stooke: Done.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSectionImpl.java
stooke: Done.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1197,6 +1266,27 @@ private void computeFullFilePath() {
...
+
+        @Override
+        public boolean isVirtual() {
+            return method instanceof HostedMethod ? ((HostedMethod) method).hasVTableIndex() : false;
+        }
+
+        @Override
+        public boolean isOverride() {
+            GraalError.unimplemented("Use MethodInfo.isOverride() instead.");
+            return false;
```
pejovica, line 1282: We'll need this one too.
```suggestion
            return method instanceof HostedMethod && allOverrides.contains(method);
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1197,6 +1266,27 @@ private void computeFullFilePath() {
...
+        }
+
+        @Override
+        public int vtableOffset() {
+            /* TODO - convert index to offset (+ sizeof DynamicHub) */
+            return method instanceof HostedMethod ? ((HostedMethod) method).getVTableIndex() : -1;
```
pejovica, line 1288: We also need to check if there is a vtable index.
```suggestion
            return isVirtual() ? ((HostedMethod) method).getVTableIndex() : -1;
```

stooke: Done.

pejovica (approved):
@stooke Looks good! Can you please rebase the PR on master and squash commits where it makes sense, and I will then integrate it.

stooke:
@pejovica , squashed and rebased.  Let's see how the GitHub gate does...

pejovica:
@stooke Great, I'll start with integration.

PR #4346, opened 2022-02-21
Add detail phase pause events for scanning phase

zhengyu123:
This patch adds detail phase pause events for Serial GC's scanning phase.  The phase counts majority time of a GC cycle, it makes sense to further breakdown the phase into sub phases, gives a detailed insight of GC performance characters.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/GCImpl.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/GCImpl.java
@@ -534,7 +534,7 @@ public final class GCImpl implements GC {
                         cheneyScanFromRoots();
                     }
                 } finally {
-                    JfrGCEvents.emitGCPhasePauseEvent(getCollectionEpoch(), incremental ? "Incremental Scan Roots" : "Scan Roots", startTicks);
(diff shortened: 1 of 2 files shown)


substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/GCImpl.java
```diff
@@ -631,44 +631,61 @@ private void cleanRuntimeCodeCache() {
...
-             * objects are in fromSpace.
-             */
-            promoteChunksWithPinnedObjects();
+                /*
+                 * Make sure all chunks with pinned objects are in toSpace, and any formerly pinned
+                 * objects are in fromSpace.
+                 */
+                promoteChunksWithPinnedObjects();
+            } finally {
+                JfrGCEvents.emitGCPhasePauseEvent(getCollectionEpoch(), "Promote Objects", startTicks);
```
christianhaeubl, line 650: I don't think that "Promote Objects" is a good name and this unit of work seems pretty arbitrary:
- `prepareForPromotion(false);` performs a few initializations that need to be done before the GC can start with its work.
- `promoteChunksWithPinnedObjects();` promotes all chunks that have pinned objects.

christianhaeubl:
This PR will be merged to master in the next few days.

PR #4352, opened 2022-02-23
Typo fix "darvin" -> "darwin"

timmattison:
Fixed a typo in the MacOS documentation where "darvin" was used instead of "darwin"


olyagpl:
The corrections are online.
Thanks Tim!

PR #4358, opened 2022-02-25
[GR-37036] Virtual thread improvements.

graalvmbot:


oraluben:
Hi Peter @peter-hofer, I wonder is there any way to use this continuation mechanism on OpenJDK / Labs JDK? I notice Java 11 support is disabled but seems planned in the future.

peter-hofer:
Hi @oraluben , it works on 17 but is disabled on 11 because of a potential deadlock of the carrier thread with the virtual thread in `ForkJoinPool`.

oraluben:
Actually I would like to know how to generate native image with continuation enabled, I tried simply `-H:+SupportContinuations` but seems some other configurations are required.

peter-hofer:
`SupportContinuations` is all that's needed, if the build doesn't abort (with details why and instructions) you can use virtual threads via `Continuations.virtualThreadFactory`.

oraluben:
I see, I'm using `-H:+SpawnIsolates -H:+SupportContinuations --native-image-info --no-fallback` but got this: 
`Fatal error: com.oracle.svm.core.util.VMError$HostedError: Unexpected monitor field offset. Continuation support currently needs option -H:+SpawnIsolates and compressed 32-bit reference support.`

Did I miss some reference compression configuration?

peter-hofer:
@oraluben: the layout of the stored continuation objects (as you originally implemented it :wink:) currently depends on references being encoded in 32 bits, which is only supported with compressed references on GraalVM Enterprise Edition. I plan on changing that in the coming weeks/months.

peter-hofer:
@oraluben: with the merge of e861c5f46dc6bca0b8d31ec2587f2a584158cde6, continuations now also work with uncompressed references.

oraluben:
@peter-hofer got it, thanks! Now it's working smoothly: https://github.com/oraluben/loom-test/commit/b79533c70a97ea812636d48f0058b5433e726714

PR #4378, opened 2022-03-08
[GR-34717] AArch64 SIMD ArrayEquals

graalvmbot:
Fix edge cases and assertion errors in the SIMD implementation of String.indexOf() intrinsics on AArch64. It includes the following improvements.

    - search a single character in a string (ArrayIndexOfOp).
    - search two consecutive characters in a string (ArrayIndexOfOp).
    - compare two string regions of the same size (ArrayEqualsOp).

    Change-Id: Icade6d0f6ad4cbceab07891be3359b0da2fde163


compiler/src/org.graalvm.compiler.lir.aarch64/src/org/graalvm/compiler/lir/aarch64/AArch64ArrayEqualsOp.java
```diff
@@ -236,4 +272,105 @@ private void emitTailCompares(AArch64MacroAssembler masm, Register result, Regis
...
+        /* If there is mismatch, then no more searching is necessary */
+        masm.cbnz(64, hasMismatch, endLabel);
+
+        /* 5. No mismatch; jump to next loop iteration. */
+        // align array1Address to 32-byte boundary
+        // determine how much to subtract from array2Address to match aligned array1Address
+        Register array1Alignment = asRegister(resultValue);
+        masm.and(64, array1Alignment, array1Address, 31);
+        masm.sub(64, array2Address, array2Address, array1Alignment);
+        masm.bic(64, array1Address, array1Address, 31);
```
teshull, line 352: @SwapnilGaikwad I changed this to have array1 aligned after the first loop iteration. Can you check to see how this performs relative to having unaligned accesses the entire time?

SwapnilGaikwad: Sure

PR #4387, opened 2022-03-10
Emit Conditional Negate

AndyJGraham:
Emit conditional negate instructions when compiling the following Java code fragments:

“(a < 0) ? -a : a” "(a <= 0) ? -a : a"
"(a > 0) ?  a : -a" "(a >= 0) ? a : -a".

### DIFF
--- a/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/ConditionalNode.java
+++ b/compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/ConditionalNode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2009, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -49,6 +49,7 @@ import org.graalvm.compiler.nodes.spi.LIRLowerable;
 import org.graalvm.compiler.nodes.spi.NodeLIRBuilderTool;
 
 import jdk.vm.ci.meta.JavaConstant;
+import jdk.vm.ci.services.Services;
 
 /**
  * The {@code ConditionalNode} class represents a comparison that yields one of two (eagerly
@@ -58,6 +59,9 @@ import jdk.vm.ci.meta.JavaConstant;
 public final class ConditionalNode extends FloatingNode implements Canonicalizable, LIRLowerable {
 
     public static final NodeClass<ConditionalNode> TYPE = NodeClass.create(ConditionalNode.class);
+
+    private static boolean isAArch64 = false;
+
     @Input(InputType.Condition) LogicNode condition;
     @Input(InputType.Value) ValueNode trueValue;
     @Input(InputType.Value) ValueNode falseValue;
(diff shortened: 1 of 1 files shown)


teshull:
I cleaned up this PR here: #4409 

Some comments are inline

teshull:
@AndyJGraham do you have examples of perf improvements from this optimization?

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/ConditionalNode.java
```diff
@@ -76,6 +80,11 @@ public ConditionalNode(LogicNode condition, ValueNode trueValue, ValueNode false
         this.condition = condition;
         this.trueValue = trueValue;
         this.falseValue = falseValue;
+        // In order to be able to generate some platform-specific nodes,
+        // this Boolean variable records whether the platform is AArch64 or not.
+        if (Services.getSavedProperties().get("os.arch").equals("aarch64")) {
+            isAArch64 = true;
+        }
```
teshull, line 87: Now AMD64 & LLVM also support the AbsNode, so this can be applied to all architectures

The general way though to add architecture-specific traits is through `LoweringProvider`

compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/calc/ConditionalNode.java
```diff
@@ -244,6 +253,17 @@ public static ValueNode canonicalizeConditional(LogicNode condition, ValueNode t
                                 ValueNode and = new AndNode(shift, add.getY());
                                 return new AddNode(add.getX(), and);
                             }
+                        } else if (isAArch64 && trueValue instanceof NegateNode) {
+                            /*
+                             * Detect (x < 0) ? -x : x (and similar variants) and replace with an
+                             * "abs" node. The "absolute" node only applies to AArch64 machines.
+                             */
+                            if (lt.getY().isJavaConstant() && (lt.getY().asJavaConstant().asLong() == 0 || lt.getY().asJavaConstant().asLong() == 1)) {
```
teshull, line 261: The above check for `lt.getY().isDefaultConstant` is equivalent to `lt.getY().isJavaConstant() && lt.getY().asJavaConstant().asLong() == 0` so that check isn't needed here. 

Too I refactored the code so that `lt.getY().isJavaConstant() && lt.getY().asJavaConstant().asLong() == 1` can be triggered

AndyJGraham:
@teshull Please find attached a document describing the performance testing for this change. Originally, the performance testing showed a 34% performance improvement. However, this took place on a Cortex A72 machine, with an architecture that is now looking rather old. For this reason, the testing was repeated on three machines with differing architectures, all running Linux.

Within the attached document, one can see that the results are very different.

One can clearly see that the code change gives an improvement of between 11 and 30 percent on the Raspberry Pi 4. This is not surprising, as the architecture is like that of the machine on which the original measurements were taken.

However, on the Altra machine, a more recent architecture, one sees a performance degradation of between 13 and 30 percent.

Finally, on the Odroid machine, the change makes no discernible performance difference. (It is worth noting that this is a relatively new, “in order” core.)

With such varying results, especially poor on the most recent platform, it is difficult to justify this change and the feeling within Arm is that this patch should be abandoned.

[Perf Comparison.pdf](https://github.com/oracle/graal/files/8443984/Perf.Comparison.pdf)

teshull:
@AndyJGraham Thanks for this description. I'll read it next week and will think about it as well

PR #4412, opened 2022-03-21
[GR-37577] Change FallbackThreshold default to NoFallback

graalvmbot:


vjovanov:
No changes in the legal status for now.

Issue #3984, opened 2021-11-03
Native images unable to access declared annotations of Java records


zakkak:
**Describe the issue**
Native images unable to access declared annotations of Java records through `java.lang.reflect.RecordComponent.declaredAnnotations`.

A simple application like the following cannot compile into a native image:
```java
import java.lang.annotation.Annotation;
import java.lang.reflect.RecordComponent;

public class Main {

	public static void main(String[] args) {
		RecordComponent[] recordComponents = F.class.getRecordComponents();
		for (RecordComponent component: recordComponents) {
			System.out.println(component);
			Annotation[] annotations = component.getAnnotations();
			for (Annotation annotation : annotations) {
				System.out.println(annotation);
			}
		}

		F x = new F("x");
		System.out.println(x);
	}

	static record F(String name) {
	}
}
```

**Steps to reproduce the issue**
Please include both build steps as well as run steps
```
cd /tmp
git clone --branch record-annotations-21.3 https://github.com/zakkak/issue-reproducers reproducers
cd reproducers
mvn package
export JAVA_HOME=/opt/jvms/graalvm-ce-java17-21.3.0
$JAVA_HOME/bin/native-image --initialize-at-build-time --diagnostics-mode \
  --no-fallback -H:+ReportExceptionStackTraces \
  -H:ReflectionConfigurationFiles=./META-INF/native-image/reflect-config.json \
  -jar target/reproducer-1.0-SNAPSHOT.jar
./reproducer-1.0-SNAPSHOT
```

**Describe GraalVM and your environment:**
 - GraalVM version: **CE 21.3.0**
 - JDK major version: **17**
 - OS: **Fedora 34**
 - Architecture: **AMD64**

**More details**

The build fails with:
```
Error: com.oracle.svm.hosted.substitute.DeletedElementException: Unsupported method java.lang.Class.getConstantPool() is reachable: The declaring class of this element has been substituted, but this element is not present in the substitution class
To diagnose the issue, you can add the option --report-unsupported-elements-at-runtime. The unsupported element is then reported at run time when it is accessed the first time.
Detailed message:
Trace: 
	at parsing java.lang.System$2.getConstantPool(System.java:2262)
Call path from entry point to java.lang.System$2.getConstantPool(Class): 
	at java.lang.System$2.getConstantPool(System.java:2262)
	at java.lang.reflect.RecordComponent.declaredAnnotations(RecordComponent.java:198)
[...]

PR #4423, opened 2022-03-23
Add regression test for JDK 17 record annotations fix

jerboaa:
This adds a simple app using record annotations, for which a
native image gets generated. This is a regression test for
the fix described in issue #3984


jerboaa:
@loicottet @olpaw Could you please have a look at this? I'm not sure what the best practise for native image tests is. Thanks!

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -1212,6 +1217,35 @@ def helloworld(args, config=None):
...
+        module_path.append(join(proj_dir, 'target', 'recordannotations.jar'))
+        config = GraalVMConfig.build(native_images=['native-image', 'lib:native-image-agent', 'lib:native-image-diagnostics-agent'])
+        with native_image_context(hosted_assertions=False, config=config) as native_image:
+            module_path_sep = ';' if mx.is_windows() else ':'
+            run_args = [
+                '-ea',
+                '-p', module_path_sep.join(module_path), 'recordannotations.Main'
+            ]
+            # Build into native image
+            mx.log('Building image recordannotion: ' + str(module_path))
```
zakkak, line 1240:
```suggestion
            mx.log('Building image recordannotation: ' + str(module_path))
```

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -227,6 +227,7 @@ def __getattr__(self, name):
     'muslcbuild',
     'hellomodule',
     'condconfig',
+    'recordannotion',
```
zakkak, line 230:
```suggestion
    'recordannotation',
```

zakkak (changes requested):

zakkak:
@jerboaa I believe the test should better be placed under https://github.com/oracle/graal/tree/master/substratevm/src/com.oracle.svm.test and I don't think it needs a special gate tag, i.e., it's OK to run it with the `test` tag.

olpaw:
> @jerboaa I believe the test should better be placed under https://github.com/oracle/graal/tree/master/substratevm/src/com.oracle.svm.test and I don't think it needs a special gate tag, i.e., it's OK to run it with the `test` tag.

@zakkak is right. If we would add an `mx <command>` for every issue we are fixing, things would get out of hand rather soon.

@jerboaa please add a jdk17 specific test project `com.oracle.svm.test..jdk17` to `substratevm/mx.substratevm/suite.py`with `"javaCompliance": "17+"` and add it to distribution `"SVM_TESTS"`. Then add you test in there as a junit test case. No need to make any changes to `mx_substratevm.py`

jerboaa:
@olpaw @zakkak OK, thanks. Will do.

substratevm/src/com.oracle.svm.test/src/META-INF/native-image/reflect-config.json
```diff
@@ -0,0 +1,11 @@
+[
+  {
+    "name" : "com.oracle.svm.test.jdk17.recordannotations.F",
```
zakkak, line 3: @jerboaa shouldn't this configuration file be moved under `substratevm/src/com.oracle.svm.test.jdk17`?

substratevm/mx.substratevm/suite.py
```diff
@@ -683,6 +683,24 @@
...
+                "SVM",
+            ],
+            "checkstyle": "com.oracle.svm.test.jdk17",
+            "checkstyleVersion" : "8.36.1",
+            "workingSets": "SVM",
+            "annotationProcessors": [
+                "compiler:GRAAL_PROCESSOR",
+            ],
+            "javaCompliance": "17+",
+            "spotbugs": "false",
```
zakkak, line 701:
```suggestion
            "spotbugs": "false",
            "testProject": True,
```

You need this to get the CI happy :)

zakkak (changes requested):

substratevm/src/com.oracle.svm.test/src/META-INF/native-image/reflect-config.json
jerboaa: @zakkak Maybe. Originally I had it there, but the build merges this then and combines `native-image` configs of the `com.orgacle.svm.test.jdk17` and `com.orgacle.svm.test` projects when it builds it into a native image. So hypothetically, if a `reflect-config.json` existed in `com.orgacle.svm.test` it would override it or it would at least be confusing which one it would indeed use. I can move it to `substratevm/src/com.oracle.svm.test.jdk17`.

zakkak: Good point. @olpaw can you please advise on this?

substratevm/mx.substratevm/suite.py
jerboaa: Thanks, done!

zakkak (approved):
LGTM, thank you @jerboaa!

jerboaa:
If somebody could help getting this integrated, I'd appreciate it.

loicottet:
This flew under my radar, sorry. I have started the process to integrate this and it should be in soon.

jerboaa:
Thanks, @loicottet

PR #4428, opened 2022-03-24
Combine Windows debugger line number records.

stooke:
Currently when emitting CodeView debugger records, Graal emits several (binary address, source line) records per line.  While this works, the user experience is less than ideal; single-stepping may appear to stay on the same line, for example, and setting a breakpoint may actually set multiple breakpoints.

This patch will not emit a new line record if the source line is the same as the previous source line.
Besides improving the user experience, there is a 1% reduction in object file size.

This patch also modifies some debug() statements to improve readability.

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVLineRecord.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVLineRecord.java
@@ -124,13 +124,19 @@ final class CVLineRecord extends CVSymbolRecord {
 
         private final ArrayList<LineEntry> lineEntries = new ArrayList<>(DEFAULT_LINE_ENTRY_COUNT);
         private final int fileId;
+        private int previousLine;
 
         FileBlock(int fileId) {
             this.fileId = fileId;
+            this.previousLine = -2;
         }
 
         void addEntry(LineEntry le) {
-            lineEntries.add(le);
+            /* Only add an entry if the line number has changed. */
+            if (le.lineAndFlags != previousLine) {
+                lineEntries.add(le);
+                previousLine = le.lineAndFlags;
+            }
         }
 
         int computeContents(byte[] buffer, int initialPos) {
@@ -167,7 +173,7 @@ final class CVLineRecord extends CVSymbolRecord {
         static final int LINE_ENTRY_SIZE = 2 * Integer.BYTES;
 
         int addr;
-        int lineAndFLags;
+        int lineAndFlags;
 
         LineEntry(int addr, int line, int deltaEnd, boolean isStatement) {
             this.addr = addr;
@@ -175,7 +181,7 @@ final class CVLineRecord extends CVSymbolRecord {
             assert line >= 0;
             assert deltaEnd <= 0x7f;
             assert deltaEnd >= 0;
-            lineAndFLags = line | (deltaEnd << 24) | (isStatement ? 0x80000000 : 0);
+            lineAndFlags = line | (deltaEnd << 24) | (isStatement ? 0x80000000 : 0);
         }
 
         LineEntry(int addr, int line) {
@@ -185,7 +191,7 @@ final class CVLineRecord extends CVSymbolRecord {
         int computeContents(byte[] buffer, int initialPos) {
             int pos = initialPos;
             pos = CVUtil.putInt(addr, buffer, pos);
-            pos = CVUtil.putInt(lineAndFLags, buffer, pos);
+            pos = CVUtil.putInt(lineAndFlags, buffer, pos);
             return pos;
         }
     }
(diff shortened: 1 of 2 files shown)


adinn:
@oubidar-Abderrahim This is part of a long-term contribution of debug info support from Red Hat and @stooke is already registered as a contributor. As the main contributor of debug info to CE (and a project member, but not an official project reviewer) I am about to review this patch. It would be helpful if you could ensure that someone from Oracle also reviews it, perhaps @pejovica ?

adinn (approved):
This looks fine to me. We may need to revisit it depending once local var info gets included as that might actually remove the need for coalescing line records but I suspect that this may still be necessary even then.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVLineRecordBuilder.java
```diff
@@ -85,26 +85,26 @@ private void processRange(Range range) {
...
+            debug("  processRange: addNewFile: %s", file);
             lineRecord.addNewFile(fileId);
         }
 
         /* Add line record. */
         /* An optimization would be to merge adjacent line records. */
         int lineLoAddr = range.getLo() - primaryEntry.getPrimary().getLo();
         int line = Math.max(range.getLine(), 1);
-        debug("processRange:   addNewLine: 0x%05x %s\n", lineLoAddr, line);
+        debug("  processRange:   addNewLine: 0x%05x-0x%05x %s", lineLoAddr, range.getLo() - primaryEntry.getPrimary().getHi(), line);
```
pejovica, line 107: I'm not quite sure what `range.getLo() - primaryEntry.getPrimary().getHi()` is supposed to be. Did you perhaps mean `range.getHi() - primaryEntry.getPrimary().getLo()` instead?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVLineRecord.java
```diff
@@ -124,13 +124,19 @@ public String toString() {
...
+            this.previousLine = -2;
         }
 
         void addEntry(LineEntry le) {
-            lineEntries.add(le);
+            /* Only add an entry if the line number has changed. */
+            if (le.lineAndFlags != previousLine) {
+                lineEntries.add(le);
+                previousLine = le.lineAndFlags;
+            }
```
pejovica, line 139: I'd remove the `previousLine` field and use this instead:
```suggestion
            if (lineEntries.isEmpty() || lineEntries.get(lineEntries.size() - 1).lineAndFlags != le.lineAndFlags) {
                lineEntries.add(le);
            }
```

Otherwise, we need a comment explaining the choice of `-2` as the initial value. 😉

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVLineRecordBuilder.java
stooke: Yes, I did - good catch.  Fixed.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVLineRecord.java
stooke: Would you be okay with a clarifying comment?  I think the code is easier to read if I leave it as is.  Something like
```
/* Use -2 because it will never appears as a source line. -1 is used as 'no line. */
private static final int NO_PREVIOUS_LINE = -2;
```

pejovica: Well, that would be better, although technically `-2` is possible as this is a value of `lineAndFlags` (it's just that we currently always set high bits to `0`).

What do you think about changing the `previousLine` type to `Integer` and initializing the field to `null`?

stooke: I'd rather avoid the potential object allocation (even if a good compiler could remove it).  -2 is fairly unlikely, as the line number would have to be 16777214.  I could modify the assert so Graal complains if LineEntry of 16777214 is ever created.

pejovica: Yes, an assertion would suffice, and I'd add it here rather than in `LineEntry`:
```java
        void addEntry(LineEntry le) {
            assert le.lineAndFlags != NO_PREVIOUS_LINE;
            /* Only add an entry if the line number has changed. */
            ...
```

stooke: Agreed - done.

pejovica (approved):
@stooke Looks good now, thanks! I'll start with integration.

PR #4435, opened 2022-03-28
JFR: Fix JFR large events

zhengyu123:
Native JFR event can be large or small, depends on its size whether fits a byte or an encoded long, and it has to be specified manually when writing an event via `JfrNativeEventWriter.beginEventWrite` and  `JfrNativeEventWriter.endEventWrite`.

Current none of native events is emitted as a large event, that results some of events are silently discarded. #[3636](https://github.com/oracle/graal/issues/3636) is an example.

The patch fixes emitting large event.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrNativeEventWriter.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrNativeEventWriter.java
@@ -80,32 +80,29 @@ public final class JfrNativeEventWriter {
         UnsignedWord written = getUncommittedSize(data);
         if (large) {
             // Write a 4 byte size and commit the event if any payload was written.
-            if (written.aboveThan(Integer.BYTES)) {
-                assert written.belowOrEqual(MAX_PADDED_INT_VALUE);
+            if (written.belowOrEqual(MAX_PADDED_INT_VALUE) && written.aboveThan(Integer.BYTES)) {
                 Pointer currentPos = data.getCurrentPos();
                 data.setCurrentPos(data.getStartPos());
-                putInt(data, makePaddedInt((int) written.rawValue()));
+                putPaddedInt(data, (int) written.rawValue());
                 data.setCurrentPos(currentPos);
                 commitEvent(data);
+                return written;
             }
         } else {
-            // Abort if event size will not fit in one byte (compressed).
-            if (written.aboveThan(MAX_COMPRESSED_BYTE_VALUE)) {
-                reset(data);
-                written = WordFactory.unsigned(0);
-            } else {
-                // Write a 1 byte size and commit the event if any payload was written.
-                if (written.aboveThan(Byte.BYTES)) {
-                    assert written.belowOrEqual(MAX_COMPRESSED_BYTE_VALUE);
-                    Pointer currentPos = data.getCurrentPos();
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/EndChunkNativePeriodicEvents.java
```diff
@@ -66,14 +66,21 @@ private static void emitInitialEnvironmentVariables(StringEntry[] envs) {
...
-
+            long eventId = JfrEvent.InitialEnvironmentVariable.getId();
+            long ticks = JfrTicks.elapsedTicks();
+            int eventIdSize = JfrNativeEventWriter.getEncodedLongSize(eventId);
+            int ticksSize = JfrNativeEventWriter.getEncodedLongSize(ticks);
             for (StringEntry env : envs) {
-                JfrNativeEventWriter.beginEventWrite(data, false);
-                JfrNativeEventWriter.putLong(data, JfrEvent.InitialEnvironmentVariable.getId());
-                JfrNativeEventWriter.putLong(data, JfrTicks.elapsedTicks());
+                int encodedSize = JfrNativeEventWriter.getEncodedStringLength(env.key) +
```
christianhaeubl, line 74: I don't think that we should compute the size manually as this is error prone and slow. If I recall correctly, HotSpot uses a retry mechanism instead. So, I would suggest to do something similar:
- use `JfrNativeEventSetting.large` to decide if an event should be emitted as a small or as a large event.
- if we tried to emit the event as a small event, but if it turns out that it needs too much space: set `JfrNativeEventSetting.large` to `true` and retry emitting the event. All future events of this event type will then be emitted as large events.

christianhaeubl:
Thanks for the contribution, I added two comments.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/EndChunkNativePeriodicEvents.java
zhengyu123: Okay. I reduce the scope of this PR, only address the problem on emitting large event. Will take a look retry mechanism in hotspot.

Thanks.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrNativeEventWriter.java
```diff
@@ -80,11 +80,11 @@ public static UnsignedWord endEventWrite(JfrNativeEventWriterData data, boolean
         UnsignedWord written = getUncommittedSize(data);
         if (large) {
             // Write a 4 byte size and commit the event if any payload was written.
-            if (written.aboveThan(Integer.BYTES)) {
+            if (written.belowOrEqual(MAX_PADDED_INT_VALUE)) {
```
christianhaeubl, line 83: The logic in this method is in general not particularly consistent at the moment. I think it should be changed to the following:
- The event should only be committed if any event payload (i.e., anything besides the event size that is reserved early on) was written.
- execute `reset(data)` and return 0 if the event was not committed (e.g., because the event was too large)

christianhaeubl:
Thanks for the cleanups, this PR will get merged in the next few days.

PR #4469, opened 2022-03-31
JFR: emit a large event if fails to write as an small event 

zhengyu123:
Please review this patch that upgrades an event to an large event if it is failed to be emitted as a small event.

The implementation mirrors Hotspot implementation, through a retry mechanism.

I only apply retry to various length events, e.g. events contain `string`. For fixed length (uncompressed) events that obviously fit small events, I elect to add assert to ensure they are properly emitted.
 
This patch also fixes #[3636](https://github.com/oracle/graal/issues/3636)

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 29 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
```diff
@@ -495,6 +495,16 @@ public boolean isEnabled(JfrEvent event) {
         return eventSettings[(int) event.getId()].isEnabled();
     }
 
+    @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
+    public void setLarge(long eventTypeId, boolean large) {
```
christianhaeubl, line 499: `setLarge(JfrEvent event, boolean large)` is probably easier to use in the callers.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/EndChunkNativePeriodicEvents.java
```diff
@@ -66,35 +67,55 @@ private static void emitInitialEnvironmentVariables(StringEntry[] envs) {
...
-
+            boolean isLarge = SubstrateJVM.get().isLarge(JfrEvent.InitialSystemProperty);
             for (StringEntry systemProperty : systemProperties) {
-                JfrNativeEventWriter.beginEventWrite(data, false);
-                JfrNativeEventWriter.putLong(data, JfrEvent.InitialSystemProperty.getId());
-                JfrNativeEventWriter.putLong(data, JfrTicks.elapsedTicks());
-                JfrNativeEventWriter.putString(data, systemProperty.key);
-                JfrNativeEventWriter.putString(data, systemProperty.value);
-                JfrNativeEventWriter.endEventWrite(data, false);
+                if (!emitInitialSystemProperty(data, systemProperty, isLarge) && !isLarge) {
```
christianhaeubl, line 99: I think it is important to document this retry pattern and why it actually works (e.g., in the class-level JavaDoc of `JfrNativeEventWriter`): if the small event got cancelled (e.g., because the flushing fails), then the `JfrNativeEventWriterData` remains in the cancelled state and it is therefore guaranteed that retry fails as well. So, the retry can really only succeed if we really ran out of space while emitting the small event.

christianhaeubl (commented):
Thanks for the contribution, I added a few comments.

christianhaeubl:
Thanks, this PR will get merged to master in the next few days.

PR #4486, opened 2022-04-12
[GR-38821] Add Java version to new build output.

jerboaa:
This simple patch adds the precise Java version to the build output. Example:

```
[1/7] Initializing...                                                                                    (3.1s @ 0.15GB)
 Version info: 'GraalVM 22.2.0-dev CE Java 17'
 Java version info: '17.0.2+8'
 C compiler: gcc (redhat, x86_64, 11.2.1)
 Garbage collector: Serial GC
```

This would make the version info available via the build output (and which is also available by invoking `native-image --version`). Right now the precise Java version used is not clear from the output. Thoughts?

Example from `native-image --version` which does show the OpenJDK security update and build version, not just the major version:
```
GraalVM 22.0.0.2 Java 17 CE (Java Version 17.0.2+8-jvmci-22.0-b05)
```

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
@@ -240,6 +240,10 @@ public class ProgressReporter {
     public void printInitializeEnd(Collection<String> libraries) {
         stagePrinter.end(getTimer(TimerCollection.Registry.CLASSLIST).getTotalTime() + getTimer(TimerCollection.Registry.SETUP).getTotalTime());
         l().a(" ").doclink("Version info", "#glossary-version-info").a(": '").a(ImageSingletons.lookup(VM.class).version).a("'").println();
+        String javaVersion = System.getProperty("java.runtime.version");
+        if (javaVersion != null) {
+            l().a(" ").doclink("Java version info", "#glossary-java-version-info").a(": '").a(javaVersion).a("'").println();
+        }
         if (ImageSingletons.contains(CCompilerInvoker.class)) {
             l().a(" ").doclink("C compiler", "#glossary-ccompiler").a(": ").a(ImageSingletons.lookup(CCompilerInvoker.class).compilerInfo.getShortDescription()).println();
         }
--- a/docs/reference-manual/native-image/BuildOutput.md
+++ b/docs/reference-manual/native-image/BuildOutput.md
@@ -72,6 +72,10 @@ The version info of the Native Image process.
 This string is also used for the `java.vm.version` property within the generated image.
 Please report this version info when you [file issues][new_issue].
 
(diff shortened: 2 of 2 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
```diff
@@ -240,6 +240,10 @@ public void printUnsuccessfulInitializeEnd() {
     public void printInitializeEnd(Collection<String> libraries) {
         stagePrinter.end(getTimer(TimerCollection.Registry.CLASSLIST).getTotalTime() + getTimer(TimerCollection.Registry.SETUP).getTotalTime());
         l().a(" ").doclink("Version info", "#glossary-version-info").a(": '").a(ImageSingletons.lookup(VM.class).version).a("'").println();
+        String javaVersion = System.getProperty("java.runtime.version");
```
zakkak, line 243: I think you will need to store the value in a static final field instead, so that it gets baked in the native-image.

See https://github.com/oracle/graal/pull/2961#discussion_r521973500 and https://github.com/oracle/graal/commit/ee25d6c9344b3b2bda0fb0b54968b1944afb105c

jerboaa: @zakkak I'm not sure what you're getting at. Could you rephrase? In this patch it should specifically get the JVM version the native image builder runs on. It should never bake into the native image. A run from CI with this has, which I think is expected:

```
GraalVM Native Image: Generating 'javac' (executable)...
[1/7] Initializing...                                                                                    (8.5s @ 0.20GB)
 Version info: 'GraalVM dev Java 17 CE'
 Java version info: '17.0.3+4-jvmci-22.1-b03'
 C compiler: gcc (linux, x86_64, 7.5.0)
 Garbage collector: Serial GC
[...]
```

zakkak: I was mistakenly worried that the `ProgressReporter` would get the wrong `java.runtime.version` property when using a native launcher but apparently that's not the case according to my tests. Both the bash launcher and the native launcher report the correct Java version. Sorry for the noise.

jerboaa: OK, thanks for checking!

fniephaus (approved):
LGTM, thanks!

jerboaa:
Thanks for the review, @fniephaus

fniephaus:
Merged via https://github.com/oracle/graal/commit/c185c866bee639caec1daf9930f7b26f29e9aafb. (I had to rebase, that's why GitHub didn't realize this PR was merged).

jerboaa:
Thanks, @fniephaus

PR #4488, opened 2022-04-13
Add option to disable features

ziyilin:
Add `-H:DisableFeatures` option to disable the specified features. If a feature is specified in both `-H:Features` and `-H:DisableFeatures`, it still gets disabled.
This option is useful for extending GraalVM. When extending Substrate VM with new features, there are cases that some features defined in the substratevm sources become incompatible and need to be disabled.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/FeatureHandler.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/FeatureHandler.java
@@ -68,10 +68,15 @@ public class FeatureHandler {
         private static List<String> userEnabledFeatures() {
             return OptionUtils.flatten(",", Options.Features.getValue());
         }
+
+        @Option(help = "A comma-separated list of fully qualified Feature implementation classes that should be disabled")//
+        public static final HostedOptionKey<LocatableMultiOptionValue.Strings> DisableFeatures = new HostedOptionKey<>(new LocatableMultiOptionValue.Strings());
     }
 
     private final ArrayList<Feature> featureInstances = new ArrayList<>();
     private final HashSet<Class<?>> registeredFeatures = new HashSet<>();
+    private final HashSet<Class<?>> disabledFeatures = new HashSet<>();
+    private final ArrayList<Feature> disabledFeatureInstances = new ArrayList<>();
 
     public void forEachFeature(Consumer<Feature> consumer) {
         for (Feature feature : featureInstances) {
@@ -91,6 +96,21 @@ public class FeatureHandler {
     public void registerFeatures(ImageClassLoader loader, DebugContext debug) {
         IsInConfigurationAccessImpl access = new IsInConfigurationAccessImpl(this, loader, debug);
 
+        List<String> invalidDisables = new ArrayList<>();
+        OptionUtils.flatten(",", Options.DisableFeatures.getValue()).forEach(featureName -> {
+            try {
+                disabledFeatures.add(Class.forName(featureName, false, loader.getClassLoader()));
+            } catch (ClassNotFoundException e) {
+                invalidDisables.add(featureName);
+            }
+        });
+
+        if (!invalidDisables.isEmpty()) {
+            StringBuilder sb = new StringBuilder();
+            sb.append("The following features specified by -H:").append(Options.DisableFeatures.getName()).append(" are not available:\n");
+            sb.append(invalidDisables.stream().reduce((s1, s2) -> s1 + "," + s2).get());
+            VMError.shouldNotReachHere(sb.toString());
+        }
         LinkedHashSet<Class<?>> automaticFeatures = new LinkedHashSet<>(loader.findAnnotatedClasses(AutomaticFeature.class, true));
         Map<Class<?>, Class<?>> specificAutomaticFeatures = new HashMap<>();
         for (Class<?> automaticFeature : automaticFeatures) {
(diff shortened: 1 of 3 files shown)


oubidar-Abderrahim:
Hi @ziyilin, Thank you for contributing to GraalVM, could you please provide answers to the following questions?

Is your PR complete and ready for review?
Did you have a conversation about these changes with some team members and you would like for them to review this as they have more context? please mention them
Do your changes rely on other changes made in a separate open PR? if yes please link it to this one

ziyilin:
Hi, @oubidar-Abderrahim
> Is your PR complete and ready for review? 

Yes, it's ready for review.
> Did you have a conversation about these changes with some team members and you would like for them to review this as they have more context? please mention them

No, I don't have a specific reviewer candidate.
> Do your changes rely on other changes made in a separate open PR? if yes please link it to this one

oubidar-Abderrahim:
Hi @olpaw, could you please review this PR, Please let me know if you prefer to mirror this PR to our bitbucket first or keep the reviewing process on here. Thank you

ziyilin:
Hi @oubidar-Abderrahim and @olpaw . Is there any review update？

olpaw (changes requested):
If we want to add that, than should we not at least also extend `com.oracle.svm.hosted.ProgressReporter#printFeatures` to show which Features are not being used due to this option? cc @christianwimmer @fniephaus

olpaw:
@ziyilin can you provide a few concrete usecase-examples that make this option necessary?

ziyilin:
> @ziyilin can you provide a few concrete usecase-examples that make this option necessary?

@olpaw We have implemented a native image based framework to run Java application in TEE. The idea is similar with this paper: https://dl.acm.org/doi/10.1145/3464298.3493406. During the implementation, we found there are a few scenarios need to disable the existing features.
One is to substitute the usage of `/dev/random` and `/dev/urandom` file in `sun.security.provider.NativePRNG`. On the Linux platform, JDK reads the pseudo random number from two special devices,  `/dev/random` and `/dev/urandom`. But they are neither supported by Intel's TEE SDK nor Microsoft's Open Enclave(OE) SDK, resulting `IOException` when `NativePRNG` attempts to read them. Therefore, we use a native method call to get the pseudo random number instead of accessing `/dev/random` and `/dev/urandom`. With our modification, feature `com.oracle.svm.core.posix.NativeSecureRandomFilesCloser` becomes unnecessary, we would like to disable it to avoid unnecessary executions.
Another one is TEE and OE SDKs don't support `sysconf(_SC_PHYS_PAGES)` well. We need to disable the original `PhysicalMemoryFeature` and get the memory information in other ways.
Without the proposed option, we have to disable the feature in very hacking approach, such as reflectively removing `PhysicalMemorySupport` from `ImageSingletons`.

christianwimmer:
@ziyilin If there are features / substitutions that are unnecessary in certain use cases, then these features / substitutions themselves should be made conditional, i.e., there should be a `HostedOption` to disable them. That is cleaner than providing an option that disables a feature. We routinely refactor the internal structure, so you can certainly not assume that a class name like `com.oracle.svm.core.posix.NativeSecureRandomFilesCloser` remains stable even across minor releases.

ziyilin:
@christianwimmer From an extension's aspect, disabling an incompatible feature by option is much simpler than making them conditional. From GraalVM's aspect, one disabling feature option is more flexible than adding options to make features conditional. It's the extension's responsibility to make sure the features it would like to disable are still available. 

Taking our enclave framework for example. We were thinking about making the conflicted features conditional in the very beginning, and in fact that's we have done for the initial version. However, considering GraalVM is a general platform, but there is still no general standards in TEE area, different SDK vendor has different implementations, we decided to keep the vendor-specialized feature inside our framework. It gives us two benefits: 

1. The impact to GraalVM's source is minimized.  
2. It's more agile to respond to TEE SDK vendor's changes. For example, if the TEE SDK supports `/dev/random` in the future, we only need to remove our substitution class and modify the disable feature option.

By the way, what do you think of introducing enclave support in native image? I assume you have the background knowledge about running Java application in enclave as native image, because two of the authors of this [paper](https://dl.acm.org/doi/10.1145/3464298.3493406) are from Oracle Labs, Zurich. Our framework is similar in idea, but different in implementation. It has been running in our production. We are glad to contribute it to the upstream if you're interested.

ziyilin:
> If we want to add that, then should we not at least also extend `com.oracle.svm.hosted.ProgressReporter#printFeatures` to show which Features are not being used due to this option being used?

@olpaw Printing disabled features is added in the `com.oracle.svm.hosted.ProgressReporter#printFeatures`.

christianwimmer:
I am still not convinced that is a useful and maintainable option. Yes, of course it is easier to just disable a feature that does not work on some platform. But I do not think that justifies a broad option that can influence all aspects of the image generator. If you want to run on a non-standard platform (like your TEE) you will also likely need other modifications to the image generator and substitutions.

Note that there is a way to override the behavior of every `public` feature class: You can create a subclass of it, and also register the subclass as a feature on the command line. Your feature is then chosen instead of the base class feature since it is more specific. And in your subclass, you can overwrite methods and do nothing.

christianwimmer:
I am still not convinced that is a useful and maintainable option. Yes, of course it is easier to just disable a feature that does not work on some platform. But I do not think that justifies a broad option that can influence all aspects of the image generator. If you want to run on a non-standard platform (like your TEE) you will also likely need other modifications to the image generator and substitutions.

Note that there is a way to override the behavior of every `public` feature class: You can create a subclass of it, and also register the subclass as a feature on the command line. Your feature is then chosen instead of the base class feature since it is more specific. And in your subclass, you can overwrite methods and do nothing.

Issue #4504, opened 2022-04-22
Implement debug location info


adinn:
**Feature request**

Please include the following information:

Extend debug info generation to support information about the location of parameter and local variables

adinn:
@vvlevchenko I am extremely sceptical that there will be much benefit using this capability from a Java IDE that is oriented around a source line view of execution. That works with the JVM and JDWP because code executed under a debugger is deoptimized and run interpreted. This means that when you stop at a line and/or step to a new line you can be very sure that what gets executed is the precise code contained on that line and nothing else.

This is going to be very far from the case when you are executing compiled native image code, even code that is compiled using the economy compilation suite. Infopoints output by the compiler that are used to build debug info have a very approximate relationship to the location in compiled code that they are attached to.

So, when you are looking at some expression on line N of a method where the debugger has stopped you may find yourself in one of 4 main cases

1) none of the code on that line has actually been executed
2) some but not all of the code on that line has been executed (but not necessarily in some easy to follow depth first left to right order of evaluation)
3) all of the code on that line has already been executed
4) there is no code from that line to execute because it has all been optimized away (even though the compiler has still left the line number in the )

Actually, that belies the true complexity of the case because what you will frequently also find is that code from previous or succeeding lines may or may not already have been executed too. That can also include code that belongs to inlined methods which has been promoted before or delayed until after the current line's code. Clearly, the problem will be worse with full optimization but all the above things can happen *even with the economy optimization suite*.

You might wonder what is the point of the debug info then? Well, if a) you can view and step through the machine code and b) you understand something about how the compiler operates ,then the debug info view of the code does help enormously to work out how the machine code was generated and, as a result, isolate some potential native image bugs. Of course, that is only really possible if you debug using a low level debugger like gdb on Linux. However, it's of limited use and the pre-conditions I outlined above mean it is not an option for most developers.

It certainly does not provide a  comparable alternative for debugging problems with a Java app under development when compared with running and debugging on the JVM. So, the idea that devs will develop, debug and deploy on native image alone is, in my view, highly misguided. My belief has always been that debug info will only be of help in certain very specific use cases. Here are the two I have come up with so far:

1) Your app displays different behaviour when run as a native image vs when run on the JVM because of a bug in the GraalVM native toolchain.
2) Your app displays different behaviour when run as a native image vs when run on the JVM because you have introduced
substitutions and you need to check that they have been applied and are doing what you expect.

[...]

PR #4505, opened 2022-04-22
[GR-38416] Implement debug location info

adinn:
Implementation PR for  issue #4504

This includes location info for parameters and locals that ar ein registers or on the stack.

It does not yet include info for bindings to constants.

Updates to the reference manual will be provided as a separate PR.

### DIFF
--- a/substratevm/mx.substratevm/testhello.py
+++ b/substratevm/mx.substratevm/testhello.py
@@ -517,10 +517,7 @@ def test():
 
     # list inlineIs and inlineA and check that the listing maps to the inlined code instead of the actual code,
     # although not ideal this is how GDB treats inlined code in C/C++ as well
-    rexp = [r'file: "hello/Hello\.java", line number: 125, symbol: "hello\.Hello::inlineIs"',
-            r"125%sinlineA\(\);"%spaces_pattern,
-            r'file: "hello/Hello\.java", line number: 126, symbol: "hello\.Hello::inlineIs"',
-            r"126%s}"%spaces_pattern]
+    rexp = [r"125%sinlineA\(\);"%spaces_pattern]
     checker = Checker('list inlineIs', rexp)
     checker.check(execute("list inlineIs"))
     rexp = [r'file: "hello/Hello\.java", line number: 130, symbol: "hello\.Hello::inlineA"',
@@ -533,7 +530,7 @@ def test():
     execute("delete breakpoints")
     # Set breakpoint at inlined method and step through its nested inline methods
     exec_string = execute("break hello.Hello::inlineIs")
-    rexp = r"Breakpoint %s at %s: file hello/Hello\.java, line 130\."%(digits_pattern, address_pattern)
+    rexp = r"Breakpoint %s at %s: hello\.Hello::inlineIs\. \(2 locations\)"%(digits_pattern, address_pattern)
     checker = Checker('break inlineIs', rexp)
     checker.check(exec_string, skip_fails=False)
 
@@ -578,15 +575,12 @@ def test():
     execute("delete breakpoints")
     # Set breakpoint at method with inline and not-inlined invocation in same line
     exec_string = execute("break hello.Hello::inlineFrom")
-    rexp = r"Breakpoint %s at %s: hello\.Hello::inlineFrom\. \(4 locations\)"%(digits_pattern, address_pattern)
+    rexp = r"Breakpoint %s at %s: file hello/Hello\.java, line 141."%(digits_pattern, address_pattern)
     checker = Checker('break inlineFrom', rexp)
     checker.check(exec_string, skip_fails=False)
 
     exec_string = execute("info break 6")
-    rexp = [r"6.1%sy%s%s in hello\.Hello::inlineFrom at hello/Hello\.java:141"%(spaces_pattern, spaces_pattern, address_pattern),
-            r"6.2%sy%s%s in hello\.Hello::inlineFrom at hello/Hello\.java:179"%(spaces_pattern, spaces_pattern, address_pattern),
-            r"6.3%sy%s%s in hello\.Hello::inlineFrom at hello/Hello\.java:162"%(spaces_pattern, spaces_pattern, address_pattern),
-            r"6.4%sy%s%s in hello\.Hello::inlineFrom at hello/Hello\.java:179"%(spaces_pattern, spaces_pattern, address_pattern)]
+    rexp = [r"6%sbreakpoint%skeep%sy%s%s in hello\.Hello::inlineFrom at hello/Hello\.java:141"%(spaces_pattern, spaces_pattern, spaces_pattern, spaces_pattern, address_pattern)]
     checker = Checker('info break inlineFrom', rexp)
     checker.check(exec_string)
 
--- a/substratevm/mx.substratevm/testhello.py
+++ b/substratevm/mx.substratevm/testhello.py
@@ -465,7 +465,7 @@ def test():
     execute("continue")
 
     exec_string = execute("info args")
-    # we canto be sure whether "this" or the argument available
+    # we cannot be sure whether "this" or argument "x" are available
     # the call to println may get inlined in which case there is no
     # guarantee that the args won't be optimized away
     rexp = [r"this = %s"%(wildcard_pattern),
@@ -503,7 +503,7 @@ def test():
             r"%sprivate:"%spaces_pattern,
             r"%sstatic void noInlineFoo\(void\);"%spaces_pattern,
             r"%sstatic void noInlineHere\(int\);"%spaces_pattern,
-            r"%sstatic void noInlineManyArgs\(int, int, int, int, int, int, int, int, int, float, float, float, float, float, float, float, float, float\);"%spaces_pattern,
+            r"%sstatic void noInlineManyArgs\(int, int, int, int, boolean, int, int, long, int, long, float, float, float, float, double, float, float, float, float, double, boolean, float\);"%spaces_pattern,
             r"%sstatic void noInlineTest\(void\);"%spaces_pattern,
             r"%sstatic void noInlineThis\(void\);"%spaces_pattern,
             r"}"]
@@ -671,25 +671,29 @@ def test():
            r"i1 = 1",
            r"i2 = 2",
            r"i3 = 3",
-           r"i4 = 4",
+           r"b4 = true",
            r"i5 = 5",
            r"i6 = 6",
-           r"i7 = 7",
+           r"l7 = 7",
            r"i8 = 8",
+           r"l9 = 9",
            r"f0 = 0",
            r"f1 = 1.125",
            r"f2 = 2.25",
            r"f3 = 3.375",
-           r"f4 = 4.5",
+           r"d4 = 4.5",
            r"f5 = 5.625",
            r"f6 = 6.75",
            r"f7 = 7.875",
-           r"f8 = 9"]
+           r"f8 = 9",
+           r"d9 = 10.125",
+           r"b10 = false",
+           r"f11 = 12.375"]
     checker = Checker('info args', rexp)
     checker.check(exec_string)
     
     exec_string = execute("x/i $pc")
-    rexp = r"=> 0x542100 <hello.Hello::noInlineManyArgs(int, int, int, int, int, int, int, int, int, float, float, float, float, float, float, float, float, float)>:	sub    $0x68,%rsp"
+    rexp = r"=> 0x542100 <hello.Hello::noInlineManyArgs(int, int, int, int, boolean int, int, long, int, long, float, float, float, float, double, float, float, float, float, double, boolean, float)>:	sub    $0x68,%rsp"
     rexp = r".*sub %s\$0x%s,%%rsp"%(spaces_pattern, hex_digits_pattern)
     checker = Checker('x/i $pc', rexp)
     checker.check(exec_string)
@@ -700,20 +704,24 @@ def test():
            r"i1 = 1",
            r"i2 = 2",
            r"i3 = 3",
-           r"i4 = 4",
+           r"b4 = true",
            r"i5 = 5",
            r"i6 = 6",
-           r"i7 = 7",
+           r"l7 = 7",
            r"i8 = 8",
+           r"l9 = 9",
            r"f0 = 0",
            r"f1 = 1.125",
            r"f2 = 2.25",
            r"f3 = 3.375",
-           r"f4 = 4.5",
+           r"d4 = 4.5",
            r"f5 = 5.625",
            r"f6 = 6.75",
            r"f7 = 7.875",
-           r"f8 = 9"]
+           r"f8 = 9",
+           r"d9 = 10.125",
+           r"b10 = false",
+           r"f11 = 12.375"]
     checker = Checker('info args 2', rexp)
     checker.check(exec_string)
 
--- a/substratevm/mx.substratevm/testhello.py
+++ b/substratevm/mx.substratevm/testhello.py
@@ -465,50 +465,19 @@ def test():
     execute("continue")
 
     exec_string = execute("info args")
-    rexp = [r"this = %s"%(address_pattern),
-            r"%s = %s"%(varname_pattern, address_pattern)]
+    # we canto be sure whether "this" or the argument available
+    # the call to println may get inlined in which case there is no
+    # guarantee that the args won't be optimized away
+    rexp = [r"this = %s"%(wildcard_pattern),
+            r"%s = %s"%(varname_pattern, wildcard_pattern)]
     checker = Checker("info args println", rexp)
     checker.check(exec_string)
 
     exec_string = execute("ptype this");
+    # the debugger shoudl still know the type of "this"
     rexp = [r"type = class java\.io\.PrintStream : public java\.io\.FilterOutputStream {"]
     checker = Checker("ptype this", rexp)
     checker.check(exec_string);
-
-    if can_print_data:
-        # print the java.io.PrintStream instance and check its type
-        exec_string = execute("print /x *this")
-        rexp = [r"%s = {"%(wildcard_pattern),
-                r"%s<java.io.FilterOutputStream> = {"%(spaces_pattern),
-                r"%s<java.io.OutputStream> = {"%(spaces_pattern),
-                r"%s<java.lang.Object> = {"%(spaces_pattern),
-                r"%s<_objhdr> = {"%(spaces_pattern),
-                r"%shub = %s,"%(spaces_pattern, address_pattern),
-                r"%sidHash = %s"%(spaces_pattern, address_pattern),
-                r"%s}, <No data fields>}, <No data fields>},"%(spaces_pattern),
-                r"%smembers of java.io.FilterOutputStream:"%(spaces_pattern),
-                r"%sclosed = 0x0"%(spaces_pattern),
-                r"%sout = %s,"%(spaces_pattern, address_pattern),
-                r"%scloseLock = %s"%(spaces_pattern, address_pattern),
-                r"%s},"%(spaces_pattern),
-                r"%smembers of java.io.PrintStream:"%(spaces_pattern),
-                r"%stextOut = %s,"%(spaces_pattern, address_pattern),
-                r"%scharOut = %s,"%(spaces_pattern, address_pattern),
-                r"%sautoFlush = 0x1,"%(spaces_pattern),
-                r"%sclosing = 0x0"%(spaces_pattern),
-                r"}"]
-
-        checker = Checker("print DefaultGreeterSystem.out", rexp)
-
-        checker.check(exec_string, skip_fails=True)
-
-        # print the hub name field and check it is java.io.PrintStream
-        # n.b. the expected String text is not necessarily null terminated
-        # so we need a wild card before the final quote
-        exec_string = execute("x/s this->hub->name->value->data")
-        checker = Checker("print PrintStream hub name",
-                          r"%s:%s\"java.io.PrintStream.*\""%(address_pattern, spaces_pattern))
-        checker.check(exec_string, skip_fails=False)
 
     ###
     # Tests for inlined methods
--- a/substratevm/mx.substratevm/testhello.py
+++ b/substratevm/mx.substratevm/testhello.py
@@ -134,6 +134,7 @@ def test():
     package_file_pattern = '[a-zA-Z0-9_/]+\\.java'
     varname_pattern = '[a-zA-Z0-9_]+'
     wildcard_pattern = '.*'
+    arg_values_pattern = "(.*=.*)"
     # obtain the gdb version
     # n.b. we can only test printing in gdb 10.1 upwards
     exec_string=execute("show version")
@@ -548,7 +549,7 @@ def test():
     rexp = [r"#0%shello\.Hello::inlineA \(\) at hello/Hello\.java:130"%spaces_pattern,
             r"#1%shello\.Hello::inlineIs \(\) at hello/Hello\.java:125"%spaces_pattern,
             r"#2%shello\.Hello::noInlineThis\(void\) \(\) at hello/Hello\.java:120"%spaces_pattern,
-            r"#3%s%s in hello\.Hello::main\(java\.lang\.String\[\] \*\) \(\) at hello/Hello\.java:93"%(spaces_pattern, address_pattern)]
+            r"#3%s%s in hello\.Hello::main\(java\.lang\.String\[\] \*\) %s at hello/Hello\.java:93"%(spaces_pattern, address_pattern, arg_values_pattern)]
     checker = Checker('backtrace inlineMee', rexp)
     checker.check(exec_string, skip_fails=False)
 
@@ -568,7 +569,7 @@ def test():
             r"#1%s%s in hello\.Hello::inlineA \(\) at hello/Hello\.java:130"%(spaces_pattern, address_pattern),
             r"#2%shello\.Hello::inlineIs \(\) at hello/Hello\.java:125"%(spaces_pattern),
             r"#3%shello\.Hello::noInlineThis\(void\) \(\) at hello/Hello\.java:120"%(spaces_pattern),
-            r"#4%s%s in hello\.Hello::main\(java\.lang\.String\[\] \*\) \(\) at hello/Hello\.java:93"%(spaces_pattern, address_pattern)]
+            r"#4%s%s in hello\.Hello::main\(java\.lang\.String\[\] \*\) %s at hello/Hello\.java:93"%(spaces_pattern, address_pattern, arg_values_pattern)]
     checker = Checker('backtrace in inlineMethod', rexp)
     checker.check(exec_string, skip_fails=False)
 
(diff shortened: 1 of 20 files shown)


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
```diff
@@ -326,6 +334,56 @@ private int writeStackLocation(DebugContext context, int offset, byte[] buffer,
...
+            } else if (byteCount == 2) {
+                pos += putShort((short) constant.asInt(), scratch, 0);
+            } else if (byteCount == 4) {
+                int i = (kind == JavaKind.Int ? constant.asInt() : Float.floatToRawIntBits(constant.asFloat()));
+                pos += putInt(i, scratch, 0);
+            } else {
+                long l =  (kind == JavaKind.Long ? constant.asLong() : Double.doubleToRawLongBits(constant.asDouble()));
+                pos += putLong(l, scratch, 0);
+            }
+        } else {
```
olpaw, line 363: If we would patch all the `put*` methods in `com.oracle.objectfile.elf.dwarf.DwarfSectionImpl` with something like
```patch
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfSectionImpl.java	(revision Staged)
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfSectionImpl.java	(date 1651057782970)
@@ -147,10 +147,12 @@
      * Base level put methods that assume a non-null buffer.
      */
 
-    protected int putByte(byte b, byte[] buffer, int p) {
-        int pos = p;
+    protected int putByte(byte b, byte[] buf, int p) {
+        boolean useScratch = buf == null;
+        int pos = useScratch ? 0 : p;
+        byte[] buffer = useScratch ? scratch : buf;
         buffer[pos++] = b;
-        return pos;
+        return pos + (useScratch ? p : 0);
     }
 
     protected int putShort(short s, byte[] buffer, int p) {
```
we could get collapse these almost identical cases for `buffer == null` and `buffer != null`

substratevm/mx.substratevm/testhello.py
```diff
@@ -438,40 +464,20 @@ def test():
...
-                r"%scharOut = %s,"%(spaces_pattern, address_pattern),
-                r"%sautoFlush = 0x1,"%(spaces_pattern),
-                r"%sclosing = 0x0"%(spaces_pattern),
-                r"}"]
-
-        checker = Checker("print DefaultGreeterSystem.out", rexp)
-
-        checker.check(exec_string, skip_fails=True)
+    exec_string = execute("info args")
+    # we canto be sure whether "this" or the argument available
```
olpaw, line 468: s/canto/want to ?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
olpaw: > Yes, I agree that would be a very good idea, although I would prefer to do it in a follow-up PR.

Fine with me. I created https://github.com/oracle/graal/issues/4513 for that.

substratevm/src/com.oracle.svm.test/src/hello/Hello.java
```diff
@@ -178,4 +180,28 @@ private static void inlineTailRecursion(int n) {
         }
         inlineTailRecursion(n - 1);
     }
+
+    @NeverInline("For testing purposes")
+    private static void noInlineManyArgs(int i0, int i1, int i2, int i3, int i4, int i5, int i6, int i7, int i8,
+                    float f0, float f1, float f2, float f3, float f4, float f5, float f6, float f7, float f8) {
```
olpaw, line 186: What would currently happen with other primitive types like `boolean` or `double`?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfDebugInfo.java
```diff
@@ -753,4 +777,77 @@ public int getAbstractInlineMethodIndex(ClassEntry classEntry, String methodName
         assert abstractInlineMethodIndex.get(methodName) != null : classEntry.getTypeName() + methodName;
         return abstractInlineMethodIndex.get(methodName);
     }
+
+    /**
+     * A class used to associate properties with a specific param or local whether top level or
+     * inline.
+     */
+
+    static final class DwarfLocalProperties {
```
olpaw, line 786: Would it make sense to simply add an `int index` field in `NativeImageDebugLocalValueInfo` instead of managing those indexes externally?

substratevm/src/com.oracle.svm.test/src/hello/Hello.java
adinn: Nothing exciting I believe. All arguments occupy a register or an 8 byte stack slot.However, I could (or, more correctly, should) throw in some extra arguments of those types to check this works ok.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
```diff
@@ -245,110 +313,93 @@ public Range getSiblingCallee() {
...
+                switch (localValueInfo.localKind()) {
+                    case REGISTER:
+                    case STACKSLOT:
+                    case CONSTANT:
+                        List<Range> varRanges = varRangeMap.get(local);
+                        assert varRanges != null : "local not present in var to ranges map!";
+                        varRanges.add(subRange);
+                        break;
+                    case UNDEFINED:
+                        break;
```
olpaw, line 390:
```patch
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java	(revision Staged)
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java	(date 1651055443390)
@@ -377,7 +377,7 @@
         for (int i = 0; i < localValueCount; i++) {
             DebugLocalValueInfo localValueInfo = subRange.getLocalValue(i);
             DebugLocalInfo local = subRange.getLocal(i);
-            if (local != null && localValueInfo.localKind() != DebugLocalValueInfo.LocalKind.UNDEFINED) {
+            if (local != null ) {
                 switch (localValueInfo.localKind()) {
                     case REGISTER:
                     case STACKSLOT:
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
```diff
@@ -0,0 +1,718 @@
...
+        }
+
+        @SuppressWarnings("unused")
+        boolean shouldMerge(int otherLo, int otherHi, DebugLocalValueInfo otherValue) {
+            // ranges need to be contiguous to merge
+            if (hi != otherLo) {
+                return false;
+            }
+            // values need to be for the same line
+            if (value.line() != otherValue.line()) {
```
olpaw, line 408: I'm curious, does it really happen that when we have `hi == otherLo` that `value.line() != otherValue.line()` is ever true?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
```diff
@@ -0,0 +1,718 @@
...
+                return false;
+            }
+            // location kinds must match
+            if (value.localKind() != otherValue.localKind()) {
+                return false;
+            }
+            // locations must match
+            switch (value.localKind()) {
```
olpaw, line 416: Shouldn't all those checks on DebugLocalValueInfos be implemented by providing an `public boolean equals(Object obj)` on `NativeImageDebugLocalValueInfo`. Also there seems to be some redundancy between here and `com.oracle.svm.hosted.image.NativeImageDebugInfoProvider.NativeImageDebugLocalValueInfo#sameAs`

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
```diff
@@ -0,0 +1,718 @@
...
+    private static final int DWARF_REG_X86_64_R9 = 9;
+    private static final int DWARF_REG_X86_64_R10 = 10;
+    private static final int DWARF_REG_X86_64_R11 = 11;
+    private static final int DWARF_REG_X86_64_R12 = 12;
+    private static final int DWARF_REG_X86_64_R13 = 13;
+    private static final int DWARF_REG_X86_64_R14 = 14;
+    private static final int DWARF_REG_X86_64_R15 = 15;
+    private static final int DWARF_REG_X86_64_XMM0 = 17;
+
+    private static final int[] GRAAL_X86_64_TO_DWARF_REG_MAP = {
```
olpaw, line 684: Same as https://github.com/oracle/graal/pull/4505/files#r859968790

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
adinn: Yes, of course.

substratevm/src/com.oracle.svm.test/src/hello/Hello.java
olpaw: Yes please add other constants of primitive types here as well.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
adinn: Yes, it can happen. The line number here is not for the declaration of the local var but for the infopoint where a location was recorded for the local. So, we can have two successive infopoints with different line numbers that both provide values for the local.

However, your question raises a good point all the same. It is important not to merge Range instances with different lines because the Range objects are used to drive generation of debug line info. However, I am now wondering whether we need to care about the line difference when merging location extents. I don't think the debugger lookup for locations requires that the location ranges conform to the line number ranges. So, I think we can remove this check and rely on the subsequent value comparison to decide whether or not to merge  the location extents.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
adinn: Yes, I think so. However, note that the above check is happening in the generator code. So, as well as implementing `boolean equals(Object obj)` in class `NativeImageDebugLocalValueInfo` we would also need to expose it as a method of interface `DebugLocalValueInfo`.

olpaw: > we would also need to expose it as a method of interface DebugLocalValueInfo

Not sure why. Every Java object has `boolean equals(Object obj)` implicitly as part of its API. It's just that the default implementation is based of ref-equality instead of contents-equality. If you change this it an `equals` override in NativeImageDebugLocalValueInfo please also add `hashcode` for completeness (they should always come in pairs).

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
```diff
@@ -0,0 +1,718 @@
...
+        return pos;
+    }
+
+    private int writeConstantLocation(DebugContext context, JavaConstant constant, byte[] buffer, int p) {
+        int pos = p;
+        byte op = DW_OP_implicit_value;
+        JavaKind kind = constant.getJavaKind();
+        int byteCount = (kind == JavaKind.Object ? 8 : kind.getByteCount());
+        assert (kind != JavaKind.Object || constant.isNull()) : "only expecting null object constant!";
+        if (buffer == null) {
```
adinn, line 343: This needs fixing. The DWARF expression defining the constant needs to be preceded with a total byte count for the expression bytes. I accidentally deleted that before pushing the update for CONSTANT support.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
adinn: Doh! ... :-)

olpaw:
Testing with `svmbuild/hello.hello`shows that setting a breakpoint in static method `java.lang.Integer#stringSize` does not seem to work.  If fact this method is not even found with `info functions stringSize`. If I set a line breakpoint in it I do see something but it looks buggy:
```
Thread 1 "hello.hello" hit Breakpoint 1, hello.Hello::noInlineManyArgs(int, int, int, int, int, int, int, int, int, float, float, float, float, float, float, float, float, float) (i0=0, i1=1, i2=2, i3=3, i4=4, i5=5, i6=6, i7=7, i8=8, f0=0, f1=1.125, f2=2.25, f3=3.375, f4=4.5, f5=5.625, f6=6.75, f7=7.875, f8=9) at hello/Hello.java:187
187	        System.out.println("i0 = " + i0);
(gdb) s
java.lang.StringConcatHelper::mix(long, int) (lengthCoder=5, value=0) at java/lang/StringConcatHelper.java:112
112	        return checkOverflow(lengthCoder + Integer.stringSize(value));
(gdb) 
java.lang.Integer::stringSize (x=0) at java/lang/Integer.java:551
551	        int d = 1;
(gdb) info locals
d = <optimized out>
p = <error reading variable p (Unhandled dwarf expression opcode 0xff)>
i = <optimized out>
```
Here are the remaining results of `info locals` within `stringSize`
```
552	        if (x >= 0) {
d = <error reading variable d (Unhandled dwarf expression opcode 0x1)>
p = <error reading variable p (Unhandled dwarf expression opcode 0xff)>
i = <optimized out>
(gdb) 
553	            d = 0;
d = <optimized out>
p = <error reading variable p (Unhandled dwarf expression opcode 0xff)>
i = <optimized out>
(gdb) 
554	            x = -x;
d = <optimized out>
p = <error reading variable p (Unhandled dwarf expression opcode 0xff)>
i = <optimized out>
(gdb) 
556	        int p = -10;
d = <optimized out>
p = <error reading variable p (Unhandled dwarf expression opcode 0xff)>
i = <optimized out>
(gdb) 
557	        for (int i = 1; i < 10; i++) {
d = <optimized out>
p = <error reading variable p (Asked for position 0 of stack, stack only has 0 elements on it.)>
i = <optimized out>
(gdb) 
558	            if (x > p)
d = <optimized out>
p = <error reading variable p (Asked for position 0 of stack, stack only has 0 elements on it.)>
i = <error reading variable i (Unhandled dwarf expression opcode 0x1)>
(gdb) 
559	                return i + d;
d = <optimized out>
p = <error reading variable p (Unhandled dwarf expression opcode 0xff)>
i = <optimized out>
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -132,6 +153,7 @@ class NativeImageDebugInfoProvider implements DebugInfoProvider {
                         .flatMap(m -> Arrays.stream(m.getImplementations())
                                         .filter(Predicate.not(m::equals)))
                         .collect(Collectors.toSet());
+        wordBaseType = heap.getUniverse().getTypes().stream().filter(type -> type.getName().equals(GRAAL_WORDBASE_TYPE_NAME)).findFirst().get();
```
olpaw, line 156: If you use
```patch
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImage.java	(revision Staged)
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImage.java	(date 1651141290178)
@@ -471,7 +471,7 @@
                 Timer timer = TimerCollection.singleton().get(TimerCollection.Registry.DEBUG_INFO);
                 try (Timer.StopTimer t = timer.start()) {
                     ImageSingletons.add(SourceManager.class, new SourceManager());
-                    DebugInfoProvider provider = new NativeImageDebugInfoProvider(debug, codeCache, heap);
+                    DebugInfoProvider provider = new NativeImageDebugInfoProvider(debug, codeCache, heap, metaAccess);
                     objectFile.installDebugInfo(provider);
                 }
                 ProgressReporter.singleton().setDebugInfoTimer(timer);
```
then you have the HostedMetaAccess provider in the `NativeImageDebugInfoProvider` and you can get the `wordBaseType` via
```java
wordBaseType = metaAccess.lookupJavaType(WordBase.class);
```

olpaw: It's a good idea in general to use metaAccess to get the HostedType equivalent of a given class.

olpaw (changes requested):
See review comments. Please also tidy up the commit history (remember commit message start with uppercase letter) and fold cleanup commits into their predecessor commits where it makes sense.

This was just the first round of reviewing. I will have another look mid next week.

substratevm/mx.substratevm/testhello.py
adinn: Fixed

substratevm/src/com.oracle.svm.test/src/hello/Hello.java
adinn: Fixed

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/Range.java
adinn: Fixed

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
adinn: This got folded into the equals method where we do have to test for equality of line numbers.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
adinn: Fixed.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
```diff
@@ -0,0 +1,718 @@
...
+    private static final int DWARF_REG_AARCH64_V25 = 89;
+    private static final int DWARF_REG_AARCH64_V26 = 90;
+    private static final int DWARF_REG_AARCH64_V27 = 91;
+    private static final int DWARF_REG_AARCH64_V28 = 92;
+    private static final int DWARF_REG_AARCH64_V29 = 93;
+    private static final int DWARF_REG_AARCH64_V30 = 94;
+    private static final int DWARF_REG_AARCH64_V31 = 95;
+
+    // map from compiler register indices to corresponding dwarf register
+    private static final int[] GRAAL_AARCH64_TO_DWARF_REG_MAP = {
```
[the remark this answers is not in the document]
adinn: Fixed

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
adinn: I corrected the error and also added support for writing object constants. So, you can now see constant strings in the initial heap etc.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Fixed

adinn:
@olpaw I addressed all the issues form your original review. While i was fixing the error I introduced into writing of primitive and null constants I also added the code for writing constant object values (i.e. reference to initial heap objects). So, you can now see things like String constants, albeit as references -- you have to execute `(gdb) print stringvar->value` to see the actual byte array contents.

I'll squash the commits down and check for proper commit messages next.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
```diff
@@ -0,0 +1,756 @@
...
+
+        public final int encoding;
+
+        DwarfRegEncodingAMD64(int encoding) {
+            this.encoding = encoding;
+        }
+    }
+
+    // map from compiler X86_64 register indices to corresponding dwarf AMD64 register index
+    private static final int[] GRAAL_X86_64_TO_DWARF_REG_MAP = {
```
olpaw, line 722: Use the same pattern here as shown in https://github.com/oracle/graal/pull/4505/files#r865995904

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1211,79 +1729,497 @@ public int line() {
...
+    class ParamLocationProducer {
+        Register[] gpregs;
+        Register[] fregs;
+        int nextGPRegIdx;
+        int nextFPregIdx;
+        int nextStackIdx;
+        int stackParamCount;
+
+        ParamLocationProducer(ResolvedJavaMethod method) {
+            Architecture arch = ConfigurationValues.getTarget().arch;
```
olpaw, line 1858: Will this work across different OSes? If we have `OSX` or `Windows` instead of Linux?

olpaw (changes requested):
PR is now very close to merging. There are only a few remaining issues.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Doh! No, of course, this needs to employ an OS- and CPU-specific call protocol.

I actually wanted class ParamLocationProducer to employ the calling convention configuration data that drives the GraalVM compiler but I had no idea how to access it. So, instead I simply replicated the relevant info per CPU. Is there some simple way to obtain the compiler config as use it as an input to ParamLocationProducer?

For now I will fix this by multiplying cases in the implementation of ParamLocationProducer to respect both CPU and OS. However, it would be clearly be more 'future-proof' if the data that drives ParamLocationProducer was the same as that used by the compiler.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
```diff
@@ -0,0 +1,756 @@
...
+
+        public final int encoding;
+
+        DwarfRegEncodingAArch64(int encoding) {
+            this.encoding = encoding;
+        }
+    }
+
+    // map from compiler AArch64 register indices to corresponding dwarf AArch64 register index
+    private static final int[] GRAAL_AARCH64_TO_DWARF_REG_MAP = {
```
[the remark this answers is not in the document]
adinn: I'm not sure this is the right thing to do. See below for details.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
adinn: This doesn't actually work -- it does for AArch64 but only because that architecture has a more sensible naming scheme.

The AArch64 and X86_64 enum tags are sorted by the DWARF encoding. So, for x86_64 the register sequence for DWARF 0...7 is RAX(0),  RDX(1),  RCX(2), RBX(3), RSI(4), RDI(5), RBP(6), RSP(7).

The Graal register numbers do not follow the same order for their register  numbering. For numberings 0...7 the corresponding registers are RAX(0), RCX(1), RDX(2), RBX(3), RSP(4), RBP(5), RSI(6), RDI(7).

That different order is taken account of in the sequence in which the DwarfRegEncoding tags appear in the array initialization expression (notice how the first 8 entries in the following table are permuted). What it boils down to is that there are actually two translation steps at play in this iniitalization. The Graal register number n (range 0-31) indexes an array slot populated using a field expression that accesses the corresponding DwarfRegEncoding enum tag. The field retrieves the DWARF encoding for that DwarfRegEncoding.

This problem could be finessed by re-ordering the enum tags so that they are declared out of DWARF encoding order. If that were done then the iterative enumeration you are suggesting could be made to work. However, I think that would make this already fairly obscure translation process more difficult to follow.

I'd prefer to keep the current layout and add a comment to explain the rationale for this two specific step translation. As it stands, the code explicitly and independently displays each of the two steps in the translation via i) the argument passed to the enum tag and ii) the order in which entries appear in the initializer. Relying on the tag ordering to determine the permutation makes this step implicit and my view is that it would make the details of the mapping harder to understand and to document.

adinn: I have pushed a patch that retains the current array initilaizer and documents the steps in the translation. I hope that is ok.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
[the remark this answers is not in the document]
adinn: Fixed by documenting the current code as per below.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
olpaw: > Relying on the tag ordering to determine the permutation makes this step implicit and my view is that it would make the details of the mapping harder to understand and to document.

Alternatively you could just add another int field to the enum that holds the corresponding Graal register number explicitly for a given enum value (e.g. `regNumGraal`). Then again turning this into `GRAAL_X86_64_TO_DWARF_REG_MAP` is just a single line of code (similar to what I suggested before).

adinn: Yes, that would be much better. I'll push another update.

adinn: Ok, I did as you suggested.

n.b. To make this more bullet-proof I used the named Register instances {r0, ... v31} and {rax ... xmm15} from, respectively, AArch64 and AMD64 to specify the register numbers passed to the corresponding DwarfRegister enum tag. That required adding a dependency to the python suites exposing package jdk.interal.vm.ci/jdk.vm.ci.code to module com.oracle.objectfile . I hope that is ok.

adinn: Repushed with format fixes.

olpaw (approved):

olpaw:
@fniephaus this PR is now ready for internal CI.

fniephaus:
> @fniephaus this PR is now ready for internal CI.

 Great! I've opened an internal PR and am running the gates....

olpaw:
@adinn one of our internal CI gates found:
```
java.lang.AssertionError: Verification of jdk.vm.ci.meta.JavaMethod usage failed: Comparing 105|LoadField#method and 106|LoadField#method in HotSpotMethod<NativeImageDebugInfoProvider$NativeImageDebugLocationInfo.merge(NativeImageDebugInfoProvider$NativeImageDebugLocationInfo)> must use .equals() for object equality, not '==' or '!='
In com.oracle.svm.hosted.image.NativeImageDebugInfoProvider$NativeImageDebugLocationInfo.initLocalInfoList(NativeImageDebugInfoProvider.java:1646): parameter 4 of call to org.graalvm.compiler.debug.DebugContext.log(int, String, Object, Object) is a call to toString() which is redundant (the callee will do it) and forces unnecessary eager evaluation.
In com.oracle.svm.hosted.image.NativeImageDebugInfoProvider$NativeImageDebugLocationInfo.initSyntheticInfoList(NativeImageDebugInfoProvider.java:1679): parameter 4 of call to org.graalvm.compiler.debug.DebugContext.log(int, String, Object, Object) is a call to toString() which is redundant (the callee will do it) and forces unnecessary eager evaluation.
```

adinn:
> @adinn one of our internal CI gates found:
>
> > java.lang.AssertionError: Verification of jdk.vm.ci.meta.JavaMethod usage failed: Comparing 105|LoadField#method and 106|LoadField#method in HotSpotMethod<NativeImageDebugInfoProvider$NativeImageDebugLocationInfo.merge(NativeImageDebugInfoProvider$NativeImageDebugLocationInfo)> must use .equals() for object equality, not '==' or '!='
> > In com.oracle.svm.hosted.image.NativeImageDebugInfoProvider$NativeImageDebugLocationInfo.initLocalInfoList(NativeImageDebugInfoProvider.java:1646): parameter 4 of call to org.graalvm.compiler.debug.DebugContext.log(int, String, Object, Object) is a call to toString() which is redundant (the callee will do it) and forces unnecessary eager evaluation.
> > In com.oracle.svm.hosted.image.NativeImageDebugInfoProvider$NativeImageDebugLocationInfo.initSyntheticInfoList(NativeImageDebugInfoProvider.java:1679): parameter 4 of call to org.graalvm.compiler.debug.DebugContext.log(int, String, Object, Object) is a call to toString() which is redundant (the callee will do it) and forces unnecessary eager evaluation.

@olpaw I'll take a look at those errors and push a fix asap

teshull:
although it would probably be better to check `offsetsInSaveArea` instead of the list

PR #4517, opened 2022-04-28
[GR-38070] Share reflection invocation stubs with same signature.

graalvmbot:
This PR changes how reflective method invocations are done: instead of a generated stub that unpacks the parameters and then does a virtual/special/static call to invoke the actual target, now the invoke is always a function pointer call. This allows re-using the stubs for all reflective invocations with the same signature. To further increase sharing, the type check for the receiver is no longer done in the generated stub. This allows sharing of instance method invokes with different receiver classes.

For my usual Spring petclinic-jdbc benchmark, there are 1346 unique signatures for 7855 methods that are registered for reflection.

The PR also ensure that the generated graphs and the `SubstrateMethodAccessor` instances do not hold on to `java.lang.reflect.Method` objects. That removes many unnecessary `Method` objects from the image heap (at the cost of a little worse error message when invoking a method with wrong parameter types).

The first commit will be merged out separately by Codrut beforehand. The commit that introduces `KnownAccessors` is mostly a refactoring to make the vtable base offset easier available without the need to have a `RuntimeConfiguration` instance.


ChristophTF:
After commit 395e75b we observed a notable increase in entry points reported by `-H:+PrintAnalysisCallTree`.

The option considers every override of a registered virtual method as an entry point:
https://github.com/oracle/graal/blob/395e75be5428cc13c4af8fae8396863600417e45/substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/reports/CallTreePrinter.java#L189-L194

Since reflection-accessible methods are now registered as roots directly,
https://github.com/oracle/graal/blob/395e75be5428cc13c4af8fae8396863600417e45/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/hosted/ReflectionFeature.java#L148

and `Object`'s members are made reflection-accessible by default,
https://github.com/oracle/graal/blob/395e75be5428cc13c4af8fae8396863600417e45/substratevm/src/com.oracle.svm.reflect/src/com/oracle/svm/reflect/hosted/ReflectionFeature.java#L228

every derived `toString()` and `hashCode()` also appears as entry point.

For a minimal Hello World program the number of printed entry points therefore increased from _~900_ to _~1500_.

PR #4522, opened 2022-04-29
CI: Fix GH invalid workflow error "A sequence was not expected"

zakkak:
Fixes GH workflow which now fails with:

```
Invalid workflow file: .github/workflows/main.yml#L16
The workflow is not valid. .github/workflows/main.yml (Line: 16, Col: 22): A sequence was not expected
```

Source: https://github.com/oracle/graal/actions/runs/2244856633

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,9 @@ on:
       - '.travis.yml'
       - '.github/workflows/quarkus.yml'
       - '**.md'
-  workflow_dispatch: []
+  # Enable manual dispatch of the workflow
+  # see https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -13,7 +13,7 @@ on:
       - '.travis.yml'
       - '.github/workflows/quarkus.yml'
       - '**.md'
-  workflow_dispatch: []
+  workflow_dispatch:
```
dougxc, line 16: What's the point of an empty event?

dougxc: Thanks. A comment to explain that would be great.

zakkak: Done

dougxc:
I'll merge this as part of #4523.

PR #4559, opened 2022-05-09
[GR-38592] Avoid use of `BigInteger` in `CgroupSubsystemController`.

graalvmbot:


substratevm/src/com.oracle.svm.core.containers/src/com/oracle/svm/core/containers/CgroupSubsystemController.java
```diff
@@ -246,8 +246,7 @@ public static long convertStringToLong(String strval, long overflowRetval, long
         } catch (NumberFormatException e) {
             // For some properties (e.g. memory.limit_in_bytes, cgroups v1) we may overflow
             // the range of signed long. In this case, return overflowRetval
-            BigInteger b = new BigInteger(strval);
-            if (b.compareTo(BigInteger.valueOf(Long.MAX_VALUE)) > 0) {
+            if (strval.length() > 0 && strval.charAt(0) != '-') {
```
smarr, line 249: This doesn't preserve the semantics of checking whether the strval fits into a `Long`, no?

fniephaus: `Long.parseLong(strval)` does that already or am I missing something?
```
$ jshell
|  Welcome to JShell -- Version 11.0.15
|  For an introduction type: /help intro

jshell> Long.MAX_VALUE
$1 ==> 9223372036854775807

jshell> Long.parseLong("9223372036854775807")
$2 ==> 9223372036854775807

jshell> Long.parseLong("9223372036854775808")
|  Exception java.lang.NumberFormatException: For input string: "9223372036854775808"
|        at NumberFormatException.forInputString (NumberFormatException.java:65)
|        at Long.parseLong (Long.java:696)
|        at Long.parseLong (Long.java:817)
|        at (#3:1)
```

smarr: The Java doc says:

- "overflowRetval" is returned if the string representation exceeds the range of type long.
- "defaultRetval" is returned if another type of error occurred during conversion.

Just checking for `-` not being included doesn't get you the `defaultRetval` semantics, no?

```
jshell> Long.parseLong("9223372036854775808")
|  Exception java.lang.NumberFormatException: For input string: "9223372036854775808"
|        at NumberFormatException.forInputString (NumberFormatException.java:67)
|        at Long.parseLong (Long.java:715)
|        at Long.parseLong (Long.java:834)
|        at (#1:1)

jshell> Long.parseLong("92233720368547abcded75808")
|  Exception java.lang.NumberFormatException: For input string: "92233720368547abcded75808"
|        at NumberFormatException.forInputString (NumberFormatException.java:67)
|        at Long.parseLong (Long.java:711)
|        at Long.parseLong (Long.java:834)
|        at (#3:1)
```

It's not that I care about this issue really. I was just curious what the patch was to remove `BigInteger`, but when I saw it I though that does not really look equivalent.

olpaw (approved):

substratevm/src/com.oracle.svm.core.containers/src/com/oracle/svm/core/containers/CgroupSubsystemController.java
fniephaus: >Just checking for - not being included doesn't get you the `defaultRetval` semantics, no?

In this case, I think, it does. You may be missing one of the details:

- `NumberFormatException` is thrown if string exceeds the range of type long and in can it's not a number
- `new BigInteger(strval)` is used unconditionally and without a catch block, which indicates `strval` can only contain numbers.
- Given the behavior of `Long.parseLong()` and the use of `BigInteger`, we can simply check if the first character of the string is a `-` (minus) to check whether `strval` is greater than `Long.MAX_VALUE` or smaller than `Long.MIN_VALUE`

Let me know if you still think this is incorrect.

fniephaus: I think this is the thing you haven't considered:
```
jshell> new BigInteger("92233720368547abcded75808")
|  Exception java.lang.NumberFormatException: For input string: "0368547ab"
|        at NumberFormatException.forInputString (NumberFormatException.java:65)
|        at Integer.parseInt (Integer.java:652)
|        at BigInteger.<init> (BigInteger.java:543)
|        at BigInteger.<init> (BigInteger.java:672)
|        at (#1:1)
```

smarr: So, you're telling me that the previous implementation was already buggy, and you don't make it worse? fair enough :)

Issue #4542, opened 2022-05-05
Enabling Full Info Points on AArch64 causes compile failures


adinn:
** Describe the Problem**

The latest patch for debug info (PR  #4505) modifies the `GraphBuilderConfiguration` to enable full info point generation. 

    class SVMHost {
      ...
    @Override
    public GraphBuilderConfiguration updateGraphBuilderConfiguration(GraphBuilderConfiguration config, AnalysisMethod method) {
        return config.withRetainLocalVariables(retainLocalVariables())
                        .withUnresolvedIsError(linkAtBuildTimeSupport.linkAtBuildTime(method.getDeclaringClass()))
                        .withFullInfopoints(SubstrateOptions.GenerateDebugInfo.getValue() > 0);
    ...

This is not a problem on x86_64. However, it leads to multiple compile failures on AArch64. They all occur because `LocationMarkerPhase.processState()` has been passed a `FullInfopointNode` referencing a live value in the *link* register `lr` aka `r30`.

    java.lang.NullPointerException
	at com.oracle.svm.core.CalleeSavedRegisters.getOffsetInFrame(CalleeSavedRegisters.java:108)
	at com.oracle.svm.core.heap.SubstrateReferenceMapBuilder.addLiveValue(SubstrateReferenceMapBuilder.java:65)
	at jdk.internal.vm.compiler/org.graalvm.compiler.lir.dfa.RegStackValueSet$1.visitValue(RegStackValueSet.java:155)
	at jdk.internal.vm.compiler/org.graalvm.compiler.lir.ValueConsumer.visitValue(ValueConsumer.java:51)
	at jdk.internal.vm.compiler/org.graalvm.compiler.lir.util.IndexedValueMap.visitEach(IndexedValueMap.java:455)
	at jdk.internal.vm.compiler/org.graalvm.compiler.lir.dfa.RegStackValueSet.addLiveValues(RegStackValueSet.java:158)
	at jdk.internal.vm.compiler/org.graalvm.compiler.lir.dfa.LocationMarkerPhase$Marker.processState(LocationMarkerPhase.java:97)
        ...

The reference map builder assumes that `lr` is a callee saved register which is why it ends up calling `CalleeSavedRegisters.getOffsetInFrame`. Unfortunately, when the call to `getOffsetInFrame` looks up `lr` in the map stored in field `offsetsInSaveArea` the result is `null` leading to a NPE.

**Steps to reproduce the issue**

*Path 1:*
  1 Clone the repo from  #4505
[...]

PR #4565, opened 2022-05-11
[GR-38665] Avoid compile failures on AArch64 when full infopoints enabled

adinn:
This patch fixes the problem reported in issue #4542.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/CalleeSavedRegisters.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/CalleeSavedRegisters.java
@@ -100,6 +100,10 @@ public class CalleeSavedRegisters {
         VMError.guarantee(saveAreaOffsetInFrame == checkedSaveAreaOffsetInFrame, "Must have a single value for the callee save register area");
     }
 
+    public boolean calleeSaveable(Register register) {
+        return offsetsInSaveArea.get(register) != null;
+    }
+
     public int getSaveAreaSize() {
         return saveAreaSize;
     }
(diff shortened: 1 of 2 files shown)


adinn:
@teshull @olpaw Could one of you please review this patch and see it through the Oracle gate?

I tested this on AArch64 by modifying class SVMHost as described in the issue and it avoids the compile failures as expected.

Proper testing will need to wait for the debuginfo lcoal vars patch to be included at which point enabling debug info will also enable full infopoint generation.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/CalleeSavedRegisters.java
```diff
@@ -100,6 +100,9 @@ public void verifySaveAreaOffsetInFrame(int checkedSaveAreaOffsetInFrame) {
         VMError.guarantee(saveAreaOffsetInFrame == checkedSaveAreaOffsetInFrame, "Must have a single value for the callee save register area");
     }
 
+    public boolean calleeSaveable(Register register) {
+        return offsetsInSaveArea.get(register) != null;
+    }
     public int getSaveAreaSize() {
```
olpaw, line 106: Add a newline here. (see failing style gate)

adinn: Fixed.

olpaw (approved):

olpaw:
Running in internal CI now ...

adinn:
Thanks Paul.

PR #4581, opened 2022-05-18
[GR-38762] Correct AArch64 debug info location generation and update tests to ha…

adinn:
This patch corrects a few small errors in AArch64 debug info generation and also updates the debuginfo test to be aware of architecture specific differences and run the test accordingly.

The generator fixes include correcting a type where v1 was specified instead of r1 and adjusting the stack offsets for stack passed parameters to make allowances for how the pushed return address is included in the offset and the frame size.

The test fixes ensure that stack passed parameter locations are checked at the method's first instruction and at the point where the stack adjustment completes. This requires different case handling for AMD64 vs AArch64 because

1. gdb places a method break at the first instruction on AMD64 but at stack push completion on AArch64
2. the stack adjustment takes only one instruction step on AMD64 and takes two instruction steps on AArch64

The current patch leaves an issue outstanding on AArch64. Stack offsets for stack passed parameters are temporarily incorrect (out by two words) between the first and second of the two stack adjustment instructions. That will need to be addressed by a follow-up patch.

This PR fixes issue #4582

### DIFF
--- a/substratevm/mx.substratevm/testhello.py
+++ b/substratevm/mx.substratevm/testhello.py
@@ -159,6 +159,8 @@ def test():
     if os.environ.get('debuginfotest_isolates', 'no') == 'yes':
         isolates = True
         
+    arch = os.environ.get('debuginfotest_arch', 'amd64')
+        
     if isolates:
         print("Testing with isolates enabled!")
     else:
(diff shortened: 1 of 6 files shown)


substratevm/mx.substratevm/testhello.py
```diff
@@ -676,10 +678,28 @@ def test():
     checker = Checker('backtrace in recursive inlineTo', rexp)
     checker.check(exec_string, skip_fails=False)
 
-    exec_string = execute("break hello.Hello::noInlineManyArgs")
+    # on aarch64 the initial break occurs at the stack push
+    # but we need to check the args before and after the stack push
+    # so we need to use the examine command to identify the start
+    # address of the method and place an instruction break at that
+    # address to ensure we hiot the very first instruction
```
olpaw, line 685: have

substratevm/mx.substratevm/testhello.py
```diff
@@ -709,12 +729,28 @@ def test():
...
+        rexp = r"%ssub %s\$0x%s,%%rsp"%(wildcard_pattern, spaces_pattern, hex_digits_pattern)
     checker = Checker('x/i $pc', rexp)
     checker.check(exec_string)
 
-    execute("stepi")
+    if arch == 'aarch64':
+        exec_string = execute("stepi")
+        print(exec_string)
+        # n.b. stack param offsets will be wrong here because
+        # aarch64 creates the fram ein two steps, a sub of
```
olpaw, line 743: frame in

olpaw (approved):

substratevm/mx.substratevm/testhello.py
adinn: fixed

fniephaus:
Merged via https://github.com/oracle/graal/commit/ceea361a4d14f8cca11b0ab79ece69906110b040.

Issue #4513, opened 2022-04-27
Simplify DwarfLocSectionImpl#write*Locations implementations


olpaw:
The conversation https://github.com/oracle/graal/pull/4505#discussion_r859671752 shows that there is a way to simplify DwarfLocSectionImpl#write*Locations implementations by making all the put* methods in com.oracle.objectfile.elf.dwarf.DwarfSectionImpl a bit smarter. This should be implemented in a separate PR. See https://github.com/oracle/graal/pull/4505#discussion_r859844959

adinn:
@olpaw @fniephaus @zakkak  Could one of you please assign this to me?

adinn:
Thanks Fabio!

olpaw:
> @olpaw @fniephaus @zakkak Could one of you please assign this to me?

Sorry. For some reason I am not bale to find you in the list when I try to assign someone to an issue. I will try to get this sorted out.

olpaw:
> > @olpaw @fniephaus @zakkak Could one of you please assign this to me?
> 
> Sorry. For some reason I am not able to find you in the list when I try to assign someone to an issue. I will try to get this sorted out.

Interestingly I can assign issues to @zakkak but not to @adinn :exploding_head:

fniephaus:
>@olpaw @fniephaus @zakkak Could one of you please assign this to me?

>am I in danger of becoming a manager?

I'm not sure this is how it works, @adinn :wink:

fniephaus:
Fixed by #4592.

PR #4592, opened 2022-05-23
Simplify DWARF section buffer writeXXX methods

adinn:
This PR fixes issue #4513.

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfSectionImpl.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfSectionImpl.java
@@ -227,6 +227,43 @@ public abstract class DwarfSectionImpl extends BasicProgbitsSectionImpl {
         return pos;
     }
 
+    protected int putULEB(long val, byte[] buffer, int p) {
+        int pos = p;
+        long l = val;
+        for (int i = 0; i < 9; i++) {
+            byte b = (byte) (l & 0x7f);
+            l = l >>> 7;
+            boolean done = (l == 0);
+            if (!done) {
+                b = (byte) (b | 0x80);
+            }
+            pos = writeByte(b, buffer, pos);
+            if (done) {
+                break;
+            }
+        }
+        return pos;
+    }
+
+    protected int putSLEB(long val, byte[] buffer, int p) {
+        int pos = p;
+        long l = val;
+        for (int i = 0; i < 9; i++) {
+            byte b = (byte) (l & 0x7f);
+            l = l >> 7;
+            boolean bIsSigned = (b & 0x40) != 0;
+            boolean done = ((bIsSigned && l == -1) || (!bIsSigned && l == 0));
+            if (!done) {
+                b = (byte) (b | 0x80);
+            }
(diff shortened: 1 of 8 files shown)


adinn:
@olpaw @fniephaus The cleanup of the buffer write methods as promised. It actually cleans up the write methods for all the sections, not just the .debug_loc section. That's a good idea as ti makes all of the section generators a lot simpler and less susceptible to error.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfSectionImpl.java
```diff
@@ -300,105 +286,147 @@ protected int putUTF8StringBytes(String s, int startChar, byte[] buffer, int p)
...
 
-    protected int writeTag(long code, byte[] buffer, int pos) {
-        if (buffer == null) {
-            return pos + putSLEB(code, scratch, 0);
+    protected int writeInt(int i, byte[] buffer, int p) {
+        if (buffer != null) {
+            return putInt(i, buffer, p);
         } else {
-            return putSLEB(code, buffer, pos);
+            return p + 4;
```
fniephaus, line 309: Minor nit: wouldn't it be clearer to use those `BYTES` fields of `Short`, `Integer`, etc.?
```suggestion
            return p + Integer.BYTES;
```

olpaw (approved):
Very nice cleanup. Thanks!

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfSectionImpl.java
adinn: Yes, it would be a tad clearer. However, I don't think Integer.Bytes is ever going to be anything other than 4 and I suspect (but cannot guarantee) that every Java programmer knows that. So I'm not sure it matters too much.

Also if you look at the if branch for this else branch you will see that it calls `putInt` which is defined as

```
    protected int putInt(int i, byte[] buffer, int p) {
        int pos = p;
        if (littleEndian()) {
            buffer[pos++] = (byte) (i & 0xff);
            buffer[pos++] = (byte) ((i >> 8) & 0xff);
            buffer[pos++] = (byte) ((i >> 16) & 0xff);
            buffer[pos++] = (byte) ((i >> 24) & 0xff);
        } else {
            buffer[pos++] = (byte) ((i >> 24) & 0xff);
            buffer[pos++] = (byte) ((i >> 16) & 0xff);
            buffer[pos++] = (byte) ((i >> 8) & 0xff);
            buffer[pos++] = (byte) (i & 0xff);
        }
        return pos;
    }
```

So, the value 4 is *kind of hard-wired* into the code.

I suppose I could change `putInt` to iterate `for (i = 0; i < Integer.BYTES; i++)` and use `8 * i` or `32 - (8 * i)` as the shift operand and so on but ...  well, life is short ;-)

fniephaus: The PR already got integrated. Feel free to add some asserts as part of one of your next PRs :smiley:.

PR #4618, opened 2022-06-02
[GR-38378] Add `--add-opens` for `java.base/java.io`.

graalvmbot:


fniephaus:
It seems this is the wrong fix...

PR #4633, opened 2022-06-09
Removes import needed by javadoc to eliminate build-time dependency

zakkak:
Mandrel builds don't build/include `org.graalvm.compiler.truffle.compiler` and this import is causing them to fail.

Closes https://github.com/graalvm/mandrel/issues/387


peter-hofer:
In progress for 22.2.

PR #4635, opened 2022-06-09
[GR-39065] Enables native-image script to handle module path with spaces.

Karm:
The problem does not affect distributions that use native-image executable.

Mandrel distribution uses native-image as a launcher script, bash on Linux, cmd on Windows and it suffers when installed in a path with spaces in it.


Karm:
Hello @ezzarghili, @gilles-duboscq,

I would like this patch to make it in for the upcoming 22.2 release, would it be possible?

Cheers
K.

fniephaus:
>I would like this patch to make it in for the upcoming 22.2 release, would it be possible?

Yes. I'll take care of it.

fniephaus (approved):
LGTM

PR #4638, opened 2022-06-10
[GR-39144] Update reference manual now debug info includes local vars

adinn:
This PR includes proposed changes to the debug manual in the wake of support for parameter and local vars. The changes do not just document this new feature they also change the example sessions so that they employ the feature to reference data by name that were previously pulled out of memory using registers or raw addresses plucked out of thin air.

### DIFF
--- a/docs/reference-manual/native-image/DebugInfo.md
+++ b/docs/reference-manual/native-image/DebugInfo.md
@@ -17,6 +17,23 @@ The resulting image will contain debug records in a format the GNU Debugger (GDB
 Additionally, you can pass `-O0` to the builder which specifies that no compiler optimizations should be performed.
 Disabling all optimizations is not required, but in general it makes the debugging experience better.
 
+By default, debug info will only include details of some of the values of parameters and local variables.
+This means that the debugger will report many parameters and local variables as being undefined. If you pass `-O0` to the builder then full debug information will be included.
+If you
+want more parameter and local variable information to be included when employing higher
+levels of optimization (`-O1` or, the default, `-O2`) you need to pass an extra command
+line flag to the `native-image` command
+
+```shell
+native-image -g -H:+SourceLevelDebug Hello
+```
+
+Enabling debuginfo with flag `-H:GenerateDebugInfo=1` does not make any difference to how a generated
+native image is compiled and does not affect how fast it executes nor how much memory it uses at runtime.
+However, it can significantly increase the size of the generated image on disk. Enabling full parameter
+and local variable information by passing flag `-H:+SourceLevelDebug` can cause a program to be compiled
+slightly differently and for some applications this can slow down execution.
+
 > Note: Native Image debugging currently works on Linux with initial support for macOS. The feature is experimental.
 
 ## Source File Caching
@@ -96,15 +113,10 @@ The currently implemented features include:
   - casting/printing objects at different levels of generality
   - access through object networks via path expressions
   - reference by name to methods and static field data
+  - reference by name to values bound to parameter and local vars
 
 Note that single stepping within a compiled method includes file and line number info for inlined code, including inlined GraalVM methods.
 So, GDB may switch files even though you are still in the same compiled method.
-
-### Currently Missing Features
-
-  - reference by name to values bound to parameter and local vars
-
-This feature is scheduled for inclusion in a later release.
 
 ### Special considerations for debugging Java from GDB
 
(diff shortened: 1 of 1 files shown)


adinn:
@fniephaus Here are some suggested edits to the debug manual. Most of the changes are actually to use local/param names in the debugger commands in order to avoid having to use registers or raw addresses. However, I have also re-ordered things slightly. The ability to refer to data by name enables a more clear and coherent exposition where we show and explain the type mapping from Java to an equivalent C++ model that gdb understands before printing any actual values.

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -17,6 +17,23 @@ The resulting image will contain debug records in a format the GNU Debugger (GDB
...
+If you
+want more parameter and local variable information to be included when employing higher
+levels of optimization (`-O1` or, the default, `-O2`) you need to pass an extra command
+line flag to the `native-image` command
+
+```shell
+native-image -g -H:+SourceLevelDebug Hello
+```
+
+Enabling debuginfo with flag `-H:GenerateDebugInfo=1` does not make any difference to how a generated
```
olpaw, line 31: We should use the API variant of this option. I.e. wherever we have `-H:GenerateDebugInfo=1` in `DebugInfo.md` it should be replaced with `-g`.

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -142,25 +157,116 @@ type = class java.lang.String : public java.lang.Object {
 }
 ```
 
-The `print` command can be used to print the contents of a referenced object field by field.
-Note how a cast is used to convert a raw memory address to a reference for a specific Java type.
+The ptype command can also be used to identify the type of Java
```
olpaw, line 160: s/the type/the static type

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -142,25 +157,116 @@ type = class java.lang.String : public java.lang.Object {
...
+    int len;
+    java.lang.String *data[0];
+} *
+```
+
+There are a few details worth highlighting here. Firstly, the debugger
+sees a Java array reference as a pointer type, as it does every Java object
+reference.
+
+Secondly, the pointer points to a structure, actually  a C++ class,
```
olpaw, line 179: `s/  / `

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -523,6 +608,12 @@ This mapping includes entries for files and line numbers for inlined methods.
 
 The *rawline* segment provides details of how the line table is generated using DWARF state machine instructions that encode file, line, and address transitions.
 
+The *loc* section provides details of address ranges within
+which parameter and local variables declared in the info section
+are known to have a determinate value. The details identify where
+the value islocated, either in a machine register, on the stack or
```
olpaw, line 614: s/islocated/is located

olpaw (changes requested):
Very nice. Only small changes required.

docs/reference-manual/native-image/DebugInfo.md
adinn: Done (only two occurrences :-).

docs/reference-manual/native-image/DebugInfo.md
adinn: done

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -179,47 +285,41 @@ type = class java.lang.Object : public _objhdr {
     . . .
 ```
 
-Given an address that might be an object reference it is possible to verify that case and identify the object's type by printing the contents of the String referenced from the hub's name field.
+The fact that all objects have a common header pointing to a class
+makes it possible to perform a simple test to decide if an address
+is an object reference and, if so,  what the object's class is.
+Given a valid object reference it is always possible to print the
+contents of the `String` referenced from the hub's name field.
```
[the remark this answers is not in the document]
olpaw: > I would use the word downcast as we are goign down the type hierarchy.

Right ... that's what I mean't. `downcast` is the correct term.

> I think giving an example for the subtype as well is clearer

Fully agree.

docs/reference-manual/native-image/DebugInfo.md
adinn: Done

adinn:
@olpaw I have made all suggested corrections modulo a few tweaks to the extra note you suggested. Let me know if you need any further changes.

olpaw (approved):

PR #4651, opened 2022-06-16
[WIP] Add support for JFR monitor enter events

roberttoyonaga:
Add support for JFR jdk.JavaMonitorEnter events. This PR is marked as a work in progress because the jdk.JavaMonitorEnter event requires the address of the monitor object to be specified. In this event is supported in HotSpot by the address being obtained in native code, but within substrateVM it is not as easy to obtain addresses of Java objects.


oraluben:
> within substrateVM it is not as easy to obtain addresses of Java objects.

Have you tried `org.graalvm.compiler.word.Word#objectToTrackedPointer`?

adinn:
@christianhaeubl Could you please review this PR or suggest someone who could do so? Thanks!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/MultiThreadedMonitorSupport.java
```diff
@@ -255,8 +257,27 @@ private static void slowPathMonitorEnter(Object obj) {
     @RestrictHeapAccess(reason = NO_LONGER_UNINTERRUPTIBLE, access = Access.UNRESTRICTED)
     @Override
     public void monitorEnter(Object obj) {
+        long startTicks = com.oracle.svm.core.jfr.JfrTicks.elapsedTicks();
```
christianhaeubl, line 260: HotSpot only emits the event if there is a contention, i.e., no JFR event is emitted for the fast path. We need something similar.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/JavaMonitorEnterEvent.java
```diff
@@ -0,0 +1,48 @@
...
+    public static void emit(Object obj, org.graalvm.nativeimage.IsolateThread previousOwner,long startTicks) {
+        emit(obj, com.oracle.svm.core.jfr.SubstrateJVM.get().getThreadId(previousOwner), startTicks);
+    }
+    @Uninterruptible(reason = "Accesses a JFR buffer.")
+    public static void emit(Object obj, long previousOwner, long startTicks) {
+        if (SubstrateJVM.isRecording() && SubstrateJVM.get().isEnabled(JfrEvent.JavaMonitorEnter)) {
+            JfrNativeEventWriterData data = StackValue.get(JfrNativeEventWriterData.class);
+            JfrNativeEventWriterDataAccess.initializeThreadLocalNativeBuffer(data);
+
+            if (emit0(data, obj, previousOwner, startTicks, SubstrateJVM.get().isLarge(JfrEvent.JavaMonitorEnter)) == com.oracle.svm.core.jfr.JfrEventWriteStatus.RetryLarge) {
```
christianhaeubl, line 28: There shouldn't be any need for the retry mechanism as this event has a fixed size (it only contains primitive data). Search the code base for `JfrNativeEventWriter.beginSmallEvent(...)` to find examples where we emit JFR events that have a fixed size.

christianhaeubl: Please also use your Red Hat email address for all commits. Otherwise, we can't merge the changes to master easily.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/MultiThreadedMonitorSupport.java
roberttoyonaga: I've now implemented this behavior by only emitting events if tryLock() fails to get the lock

christianhaeubl: This new approach doesn't work:
1. You only fill `monitorOwners` when a lock contention is happening. So, you will usually keep track of an outdated value and not of the latest previous owner. One possible solution would be to introduce a custom `ReentrantLock` subclass that keeps track of the previous owner in a field. Writing the JFR thread id to that field should be cheap enough so that you can do it in the fast path.
2. There is no guarantee that your call to `lockObject.lock();` really runs into a contention. So, you will emit unnecessary JFR events from time to time. To solve that, you will probably need to add a method that is similar to `lock()` that returns if it was necessary to park the thread because of a contention.

roberttoyonaga: yes I see! Thank you Christian. I have used the solution you suggested for **1**. 

For **2**, wouldn't `tryLock()` returning false indicate that there is contention? Is the problem that the lock may be released in between the call to `tryLock()` and `lock()`? I think the Hotspot code makes one attempt to acquire the lock (+1 optional spin) and upon failure treats it as contention.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/OwnedReentrantLock.java
```diff
@@ -0,0 +1,60 @@
...
+    public long getPrevOwnerTid() {
+        return prevOwnerTid;
+    }
+
+    public OwnedReentrantLock() {
+        super();
+        currOwnerTid = SubstrateJVM.get().getThreadId(CurrentIsolate.getCurrentThread());
+    }
+
+    public boolean tryLockOwnedLock() {
```
christianhaeubl, line 47: You could rename this method to `enter` or `monitorenter`.
If you move the call to `JavaMonitorEnterEvent.emit(...);` into this method as well, then it should be sufficient to store the latest owner in a single field.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/MultiThreadedMonitorSupport.java
```diff
@@ -192,8 +193,10 @@ public class MultiThreadedMonitorSupport extends MonitorSupport {
      * Secondary storage for monitor slots. Synchronized to prevent concurrent access and
      * modification.
      */
-    private final Map<Object, ReentrantLock> additionalMonitors = new WeakIdentityHashMap<>();
+    private final Map<Object, OwnedReentrantLock> additionalMonitors = new WeakIdentityHashMap<>();
+    private final Map<Object, Long> monitorOwners = new WeakIdentityHashMap<>();
```
christianhaeubl, line 197: The fields `monitorOwners` and `monitorOwnersLock` are unused.

roberttoyonaga: removed!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/OwnedReentrantLock.java
roberttoyonaga: renamed and moved call to emit event

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/OwnedReentrantLock.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package com.oracle.svm.core.monitor;
+
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.graalvm.nativeimage.CurrentIsolate;
+import com.oracle.svm.core.jfr.SubstrateJVM;
+
+public class OwnedReentrantLock extends ReentrantLock {
```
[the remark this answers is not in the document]
roberttoyonaga: renamed!

christianhaeubl (approved):
Thanks, I will try to merge this change in the next couple days.

PR #4672, opened 2022-06-27
[GR-39483] Avoid implicitly setting class initialization reason for classes/packages

zakkak:
When marking a class `my.package.MyClass` for initialization for a given
reason before this patch `my` and `my.package` would also be assigned
the reason implicitly.  This resulted in reporting the wrong reason in
some cases.  For instance before this patch
`java.lang.AbstractStringBuilder` would be marked as initialized because
it's a "super type of sun.security.util.UntrustedCertificates" which is
clearly not the case.  This patch fixes the underlying problem and
reports "Core JDK classes are initialized at build time" as the reason
for initializing `AbstractStringBuilder` at `BUILD_TIME`.


fniephaus (approved):
LGTM

PR #4676, opened 2022-06-28
Add Support for JFR Java Monitor Wait and Thread Sleep events

roberttoyonaga:
With regard to the jdk.JavaMonitorWait event, I am using a queue to keep track of the TIDs of notifiers so that waiters have access to them when they resume and emit the jdk.JavaMonitorWait event. This takes takes advantage of `Condition.signal() ` signalling waiters in the order they started waiting.

I have created another PR https://github.com/oracle/graal/pull/4698, that uses substitutions more closely mimic the way this event is handled in Hotspot. This alternative approach is simpler, but requires maintenance of the substituted methods when the original methods change.


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/JavaMonitor.java
```diff
@@ -0,0 +1,104 @@
...
+
+public class JavaMonitor extends ReentrantLock {
+    private long ownerTid;
+
+    //Records the number of waiters that have not yet been notified
+    private int waitingThreads = 0;
+    /*
+     * This queue is used to record the TIDs of notifiers so that waiters have access to them when they eventually
+     * acquire the monitor lock and resume.
+     * Adding and removing TIDs from this queue in a FIFO fashion is correct because Condition.signal(), notifies waiters in a FIFO fashion.
```
christianhaeubl, line 45: I don't think that this assumption is correct because:
a. it is possible to specify a timeout (see `wait(long)`)
b. it is possible to interrupt a thread

christianhaeubl (changes requested):

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/JavaMonitorWaitEvent.java
```diff
@@ -0,0 +1,57 @@
...
+import com.oracle.svm.core.jfr.JfrEvent;
+import com.oracle.svm.core.jfr.JfrNativeEventWriter;
+import com.oracle.svm.core.jfr.JfrNativeEventWriterData;
+import com.oracle.svm.core.jfr.JfrNativeEventWriterDataAccess;
+import com.oracle.svm.core.jfr.JfrTicks;
+import com.oracle.svm.core.jfr.SubstrateJVM;
+
+public class JavaMonitorWaitEvent {
+    @Uninterruptible(reason = "Accesses a JFR buffer.")
+    public static void emit(long startTicks, Object obj, long notifier, long timeout, boolean timedOut) {
```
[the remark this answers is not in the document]
roberttoyonaga: I see. Done!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/JavaMonitor.java
roberttoyonaga: a. If a thread wakes up from a timeout then it will not pull anything off of the TID queue. It will instead just emit an event with the `timedout` flag set. [see here: ](https://github.com/oracle/graal/pull/4676/files#diff-323fb7546644875e073e4f7271cbc476ef672e237278eff7e21f7d8d47d928a8R383-R387)

b. If a waiting thread is interrupted it will also not affect the TID queue. 

I believe this means that a waiter will modify the notifier TID queue if and only if it has been notified.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/JavaMonitorWaitEvent.java
```diff
@@ -0,0 +1,57 @@
...
+        if (SubstrateJVM.isRecording() && SubstrateJVM.get().isEnabled(JfrEvent.JavaMonitorWait)) {
+            JfrNativeEventWriterData data = org.graalvm.nativeimage.StackValue.get(JfrNativeEventWriterData.class);
+            JfrNativeEventWriterDataAccess.initializeThreadLocalNativeBuffer(data);
+
+            JfrNativeEventWriter.beginSmallEvent(data, JfrEvent.JavaMonitorWait);
+            JfrNativeEventWriter.putLong(data, startTicks);
+            JfrNativeEventWriter.putLong(data, JfrTicks.elapsedTicks() - startTicks);
+            JfrNativeEventWriter.putEventThread(data);
+            JfrNativeEventWriter.putLong(data, SubstrateJVM.get().getStackTraceId(JfrEvent.JavaMonitorWait.getId(), 0));
```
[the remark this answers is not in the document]
roberttoyonaga: done!

christianhaeubl (changes requested):

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/JavaMonitor.java
roberttoyonaga: Thank you Christian. I have changed things so that we always clean up after a waiter resumes (through timeout, interrupt, or signal).  I have also changed things so that waiters are checked to still be waiting before we mark them as being notified (so that notifications don't get associated with waiters that have timed out or interrupted).  Some drawbacks of this approach  are that a linear scan of the maintained queue is required when waiters wake up. This is because the arbitrary order in which threads reacquire the lock. 

In Hotspot, a notifierTid field is added to the Node struct that sits on the wait and sync queues. This way, when a waiter wakes up, it can simply inspect that field. However, in substrateVM, the wait queue and sync queue is abstracted away in `Condition.signal() and Condition.await()` the solution in this PR tries to mirror  this with its own queue. I have created another PR [here](https://github.com/oracle/graal/pull/4698), that uses substitutions to copy the behaviour in Hotspot more closely. This alternative approach is much simpler, but requires maintenance of the substituted methods when the original methods change.

christianhaeubl (changes requested):
Thanks for working on that but the current solution still has major issues. Regarding https://github.com/oracle/graal/pull/4698: we can't use substitutions for this because those substitutions would affect the JDK as a whole (and therefore other usages of `ReentrantLock`) and not only the Java monitors. The other issue is that we would potentially need different substitutions for different JDK versions.

Maybe it is really time to move to a custom Java monitor implementation that is only loosely based on the `ReentrantLock` implementation but that no longer uses the JDK classes directly. This is something that we will need eventually anyways. Then, we can modify the internals of the locking code as needed.

Moving forward, I would suggest to split the work into the following PRs:
- Move the sleep event & all infrastructure changes that you currently made to a separate PR. We can merge this PR right away.
- Copy the JDK code for `ReentrantLock`, `NonfairSync`, and `ConditionObject` from the JDK19 sources. Strip the implementation down to the bare minimum that is needed by Native Image for implementing Java monitors and use that implementation instead of the JDK classes.
- Add support for the event `JavaMonitorWait` by modifying the new Native Image-specific Java monitor implementation.

roberttoyonaga:
@christianhaeubl Thank you Christian. I have factored out thread sleep events into another PR here https://github.com/oracle/graal/pull/4702. I'll begin working on the custom monitor implementation.

PR #4681, opened 2022-06-29
Removed unused constants.

koutheir:
Removed unused constants in `LLVMExceptionUnwind`.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/runtime/LLVMExceptionUnwind.java
+++ b/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/runtime/LLVMExceptionUnwind.java
@@ -152,24 +152,15 @@ public class LLVMExceptionUnwind {
 
     // Allow methods with non-standard names: Checkstyle: stop
 
+    // The following declarations are from <unwind.h>.
+    //
+    // See:
+    // - https://clang.llvm.org/doxygen/unwind_8h_source.html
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/runtime/LLVMExceptionUnwind.java
```diff
@@ -153,24 +153,9 @@ protected void customUnwindException(Pointer callerSP) {
     // Allow methods with non-standard names: Checkstyle: stop
 
     /* _Unwind_Reason_Code */
-    @CConstant
-    private static native int _URC_NO_REASON();
```
loicottet, line 157: I am fine with removing those constants. I think we should add a comment linking to the libunwind docs so we can find them back easily if we need them.

loicottet (approved):

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/runtime/LLVMExceptionUnwind.java
koutheir: Would a comment like this be fine?

```java
// The following declarations are from <unwind.h>.
//
// See:
// - https://clang.llvm.org/doxygen/unwind_8h_source.html
// - https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libgcc/unwind-generic.h
```

koutheir:

PR #4685, opened 2022-06-30
[GR-39497] Add support for JSON build output to GraalVM Native Image.

graalvmbot:


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ProgressReporter.java
```diff
@@ -447,12 +447,11 @@ public void printCreationEnd(int imageSize, int numHeapObjects, long imageHeapSi
                         .doclink("code area", "#glossary-code-area").a(":%,10d compilation units", numCompilations).println();
         EconomicMap<Pair<String, String>, ResourceStorageEntry> resources = Resources.singleton().resources();
         int numResources = resources.size();
-        recordJsonMetrics(ImageDetailKey.IMAGE_HEAP_RESOURCE_COUNT, numResources);
-        recordJsonResourceMetrics(resources);
```
jerboaa: @fniephaus by removing `recordJsonResourceMetrics` you no longer generate the `image_details -> image_heap -> resources -> bytes` value in the JSON output and the produced JSON would therefore no longer be a according to schema here:
https://github.com/oracle/graal/blob/fniephaus/GR-39497/json-build-output/docs/reference-manual/native-image/assets/build-output-schema-v1.json#L259..L262

So either, drop it from the schema too or keep the method.

fniephaus: I think you might have missed https://github.com/oracle/graal/pull/4685/files#diff-54ef73a23b10bd907d5869cc88b651fae7fef0467ccfaf8f50472cdd1e114eceR531 and https://github.com/oracle/graal/commit/5e652cded2e61a9326613fc37d4846fe2b4ec4e2 :)

PR #4702, opened 2022-07-05
Add thread sleep events and refactor

roberttoyonaga:
This adds JFR thread sleep events and adds method `getStackTraceId(JfrEvent,  int )` to SubstrateJVM.


christianhaeubl:
Thanks, I will merge this PR to master in the next few days.

PR #4705, opened 2022-07-07
TruffleDSL: Remove inInterpreter case from cached implicit type casts

smarr:
This is likely not beneficial any more now with hosted inlining.
In Ruby, it causes all long conversions to go through a call, and the compiler can’t use the context from previous checks to optimize.

There might be a small performance benefit: https://rebench.stefan-marr.de/compare/TruffleRuby/0b8869e410a4fc3104bfe6a0eddf732218c2008e/3eec1f52e35853535cf6ce3af13250c1edccf7a9#ruby-startup-TruffleRuby-native-interp

/cc @chumer


smarr:
The GH Actions fail somewhere with `No such file or directory: '.idea/modules.xml'`, which seems unrelated.

fniephaus:
>The GH Actions fail somewhere with No such file or directory: '.idea/modules.xml', which seems unrelated.

It is. We aware of this and will deploy a fix shortly.

PR #4715, opened 2022-07-11
Fix GetDirectBufferCapacity() implementation

koutheir:
[`GetDirectBufferCapacity()`](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#getdirectbuffercapacity) JNI must return `-1` if the given object is not a direct `java.nio.Buffer`.

### DIFF
--- a/sulong/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
+++ b/sulong/projects/com.oracle.truffle.llvm.parser/src/com/oracle/truffle/llvm/parser/LazyToTruffleConverterImpl.java
@@ -209,7 +209,7 @@ public class LazyToTruffleConverterImpl implements LazyToTruffleConverter {
             }
             LLVMStatementNode[] nodes = visitor.finish();
             info.setBlockDebugInfo(block.getBlockIndex(), visitor.getDebugInfo());
-            blockNodes[block.getBlockIndex()] = LLVMBasicBlockNode.createBasicBlockNode(options, nodes, visitor.getControlFlowNode(), block.getBlockIndex(), block.getName());
+            blockNodes[block.getBlockIndex()] = LLVMBasicBlockNode.createBasicBlockNode(nodes, visitor.getControlFlowNode(), block.getBlockIndex(), block.getName());
         }
 
         for (int j = 0; j < blockNodes.length; j++) {
@@ -226,7 +226,7 @@ public class LazyToTruffleConverterImpl implements LazyToTruffleConverter {
 
             if (cfg.isReducible() && cfg.getCFGLoops().size() > 0) {
                 loopSuccessorSlot = builder.addSlot(FrameSlotKind.Int, null, null);
-                resolveLoops(blockNodes, cfg, loopSuccessorSlot, exceptionSlot, info, options);
+                resolveLoops(blockNodes, cfg, loopSuccessorSlot, exceptionSlot, info);
             }
         }
 
@@ -267,7 +267,7 @@ public class LazyToTruffleConverterImpl implements LazyToTruffleConverter {
         return neededForDebug;
     }
 
-    private void resolveLoops(LLVMBasicBlockNode[] nodes, LLVMControlFlowGraph cfg, int loopSuccessorSlot, int exceptionSlot, LLVMRuntimeDebugInformation info, OptionValues options) {
+    private void resolveLoops(LLVMBasicBlockNode[] nodes, LLVMControlFlowGraph cfg, int loopSuccessorSlot, int exceptionSlot, LLVMRuntimeDebugInformation info) {
         // The original array is needed to access the frame nuller information for outgoing control
         // flow egdes
         LLVMBasicBlockNode[] originalBodyNodes = nodes.clone();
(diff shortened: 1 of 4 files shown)


koutheir:

oubidar-Abderrahim:
Hi, @koutheir, Thank you for contributing to GraalVM, could you please provide answers to the following questions?

Is your PR complete and ready for review?
Did you have a conversation about these changes with some team members and you would like for them to review this as they have more context? please mention them
Do your changes rely on other changes made in a separate open PR? if yes please link it to this one

koutheir:
> Is your PR complete and ready for review?

Yes.

> Did you have a conversation about these changes with some team members and you would like for them to review this as they have more context?

No.

> Do your changes rely on other changes made in a separate open PR?

No.

substratevm/src/com.oracle.svm.jni/src/com/oracle/svm/jni/functions/JNIFunctions.java
```diff
@@ -640,9 +640,9 @@ static WordPointer GetDirectBufferAddress(JNIEnvironment env, JNIObjectHandle ha
 
     @CEntryPoint(exceptionHandler = JNIExceptionHandlerReturnMinusOne.class, include = CEntryPoint.NotIncludedAutomatically.class, publishAs = Publish.NotPublished)
     @CEntryPointOptions(prologue = JNIEnvEnterPrologue.class, prologueBailout = ReturnMinusOneLong.class)
-    static long GetDirectBufferCapacity(JNIEnvironment env, JNIObjectHandle hbuf) {
-        Buffer buffer = JNIObjectHandles.getObject(hbuf);
-        return buffer.capacity();
+    static long GetDirectBufferCapacity(JNIEnvironment env, JNIObjectHandle handle) {
+        Object obj = JNIObjectHandles.getObject(handle);
+        return (obj instanceof Target_java_nio_Buffer) ? ((Target_java_nio_Buffer) obj).capacity : -1;
```
christianwimmer, line 645: The JNI documentation says
> Returns -1 if the given object is not a **direct** java.nio.Buffer

So I think there is also a check missing whether this is a direct buffer or not. `GetDirectBufferAddress` has the same issue, it also needs some check for a direct buffer first.

koutheir: > So I think there is also a check missing whether this is a direct buffer or not.

The `instanceof` operation **IS** the check:
```java
return (obj instanceof Target_java_nio_Buffer) ? ((Target_java_nio_Buffer) obj).capacity : -1;
```

> `GetDirectBufferAddress` has the same issue, it also needs some check for a direct buffer first.

No it doesn't. It also [performs the check](https://github.com/oracle/graal/blob/b49fe19ffe7f4cbd60ae2a040d19871d89fbd67c/substratevm/src/com.oracle.svm.jni/src/com/oracle/svm/jni/functions/JNIFunctions.java#L630):
```java
if (obj instanceof Target_java_nio_Buffer) {`
```

christianwimmer: No, `Buffer` can both be a direct buffer or a heap-based buffer. So that check is not sufficient.

koutheir: Then how is one supposed to check for these two types?

christianwimmer: The best thing is usually to look at the implementation in HotSpot, and do the same kind of checks.

peter-hofer: HotSpot checks for interface `sun.nio.ch.DirectBuffer`. You will need to introduce a `@TargetClass` as alias.

christianwimmer: @peter-hofer why a `TargetClass`? `DirectBuffer` is `public` so can be used directly for `instanceof` checks.

christianwimmer: And since `DirectBuffer` is unrelated in the class hierarchy to `Buffer`, we actually need to do a `instanceof` check for both types. Otherwise a malicious implementation of `DirectBuffer` that does not extend `Buffer` could lead to a `ClassCastException` again.

peter-hofer: True, a `@TargetClass` is actually not needed and checking for both `Buffer` and `DirectBuffer` is the safest variant. @koutheir , please also add such a check for `GetDirectBufferAddress`.

koutheir:
I might have done something incorrectly. I pulled changes and merged from the `master` branch, and applied the same changes again. Can we continue in #4838 ?

substratevm/src/com.oracle.svm.jni/src/com/oracle/svm/jni/functions/JNIFunctions.java
koutheir: I might have done something incorrectly. I pulled changes and merged from the master branch, and applied the same changes again in order to resolve the conflicts. Can we continue in https://github.com/oracle/graal/pull/4838 ?

peter-hofer:
Most likely you removed all commits in a push and that led the automation to believe that your changes were already integrated.

PR #4716, opened 2022-07-11
Align pthread_mutex_t and pthread_cond_t on word boundary

koutheir:
Align `pthread_mutex_t` and `pthread_cond_t` on word boundary, not object alignment.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/pthread/PthreadVMLockSupport.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/pthread/PthreadVMLockSupport.java
@@ -95,11 +95,26 @@ final class PthreadVMLockFeature implements Feature {
 
     @Override
     public void beforeCompilation(BeforeCompilationAccess access) {
-        ObjectLayout layout = ConfigurationValues.getObjectLayout();
-        int alignment = layout.getAlignment();
+        final int wordSize = ConfigurationValues.getTarget().wordSize;
 
-        int baseOffset = layout.getArrayBaseOffset(JavaKind.Byte);
-        /* padding if first element is not object aligned */
+        // `alignment` should actually be: `max(alignof(pthread_mutex_t), alignof(pthread_cond_t))`.
+        //
+        // Until `alignof()` can be queried from the C compiler, we hard-code this alignment to:
+        // - One word on 64-bit architectures.
+        // - Two words on 32-bit architectures.
+        //
+        // This split is arbitrary. Actual alignment requirements depend on the architecture,
+        // the Pthread library implementation, and the C compiler.
+        // These hard-coded values will need to be adjusted to higher values if we find out
+        // that `pthread_mutex_t` or `pthread_cond_t` have higher alignment requirements on some
+        // particular architecture.
+        assert wordSize == 8 || wordSize == 4 : "Unsupported architecture bit width";
+        final int alignment = (wordSize == 8) ? wordSize : (2 * wordSize);
+
+        ObjectLayout layout = ConfigurationValues.getObjectLayout();
+        final int baseOffset = layout.getArrayBaseOffset(JavaKind.Byte);
+
+        // Align the first element to word boundary.
         int nextIndex = NumUtil.roundUp(baseOffset, alignment) - baseOffset;
 
         PthreadVMMutex[] mutexes = mutexReplacer.getReplacements().toArray(new PthreadVMMutex[0]);


koutheir:

gilles-duboscq:
We have gone through a few iterations of that code in the initial fix.
What we really want here is `alignof(pthread_mutex_t)` and `alignof(pthread_cond_t)`.
For `pthread_mutex_t`, that seems to be `wordSize` on the few architectures i tried (amd64, riscv 32/64, armv7, aarch64, power, power64).
For `pthread_cond_t` that doesn't seem to hold as it requires an alignment of 8 bytes on `armv7`.

We tried to add support for `alignof` in the C query system but that added a dependency on C11 which we didn't want.
We settled on `layout.getAlignment()` as a strange way to get `8`...

We could use `wordSize` for `pthread_mutex_t` and just hard-code 8 for `pthread_cond_t`.

lewurm:
Thanks for your PR @koutheir. What problem does this PR fix for you? Could you elaborate a bit?

koutheir:
> We have gone through a few iterations of that code in the initial fix. What we really want here is `alignof(pthread_mutex_t)` and `alignof(pthread_cond_t)`. For `pthread_mutex_t`, that seems to be `wordSize` on the few architectures i tried (amd64, riscv 32/64, armv7, aarch64, power, power64). For `pthread_cond_t` that doesn't seem to hold as it requires an alignment of 8 bytes on `armv7`.
> 
> We tried to add support for `alignof` in the C query system but that added a dependency on C11 which we didn't want. We settled on `layout.getAlignment()` as a strange way to get `8`...
> 
> We could use `wordSize` for `pthread_mutex_t` and just hard-code 8 for `pthread_cond_t`.

In this case, I would rather use `2 * wordSize` to clearly indicate the alignment requirement, rather than link it to Java Object alignment. We could also indicate in a comment that an ideal solution would be `max(alignof(pthread_mutex_t), alignof(pthread_cond_t))`. What do you think, @gilles-duboscq ?

koutheir:
> What problem does this PR fix for you?

This fixes the meaning of the code, not its final effects.

> Could you elaborate a bit?

The method modified by this patch allocates a Java byte array that is used to store multiple C structures, namely `pthread_mutex_t` and `pthread_cond_t`. These structures are passed to `pthread_*()` functions. Most ABIs require these structures to be aligned on some boundary. Failing to do so results in `SIGSEGV`/`SIGBUS` signals when these structures are accessed (it is undefined behavior in C).

At runtime, depending on the actual address of the Java byte array, the method prepends as many padding bytes as needed in order to align the C structures. For this, the (exact or minimum) alignment of these structures must be known. Ideally, this should be communicated from the C compiler as:
```C
max(alignof(pthread_mutex_t), alignof(pthread_cond_t))
```

The previous version of the code uses `ConfigurationValues.getObjectLayout().getAlignment()` as the alignment, which is misleading, because the alignment of these structures has nothing to do with the alignment of Java objects.

My fix removes this wrong link, and replaces it with an alignment on a number of machine words. Given what @gilles-duboscq said, I suggest using an alignment of two words (8 (resp. 16) bytes on 32-bit (resp.  64-bit) architectures) in order to define a correct minimum alignment for these C structures, until Graal adds support for querying `alignof()` from the C compiler.

koutheir:
I don't understand the issues in the CI server build failures. Can someone help with this?

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/pthread/PthreadVMLockSupport.java
```diff
@@ -95,11 +95,14 @@ public void duringSetup(DuringSetupAccess access) {
 
     @Override
     public void beforeCompilation(BeforeCompilationAccess access) {
+        // This should actually be: `max(alignof(pthread_mutex_t), alignof(pthread_cond_t))`.
+        // Until `alignof()` can be queried from the C compiler, we hard-code this to two words.
+        final int alignment = 2 * ConfigurationValues.getTarget().wordSize;
```
lewurm, line 100: The comment is good.  Can we hard-code to `8` instead?  Then it wouldn't introduce any layout change on 64bit platforms.

lewurm:
> I don't understand the issues in the CI server build failures. Can someone help with this?

There was an internal change that broke our GitHub CI.  It should be fixed now, I rescheduled the failed jobs. Sorry about that.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/pthread/PthreadVMLockSupport.java
koutheir: I still don't want to hard-code this with numbers. Please take a look at my new changes.

lewurm (approved):
LGTM now, thank you for your patience @koutheir. I'll push it through our internal CI.

PR #4717, opened 2022-07-11
Fix bit shifting behavior in LLVM back-end, as per Java specifications

koutheir:
Fix bit shifting behavior in LLVM back-end, as per [Java specifications](https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.19).

### DIFF
--- a/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMGenerator.java
+++ b/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMGenerator.java
@@ -1469,24 +1469,36 @@ public class LLVMGenerator implements LIRGeneratorTool, SubstrateLIRGenerator {
             throw unimplemented("the LLVM backend only supports XOR of integers, vectors and floating point numbers");
         }
 
+        private LLVMValueRef actualShiftingDistance(LLVMValueRef a, LLVMValueRef b) {
+            // https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.19
+
+            LLVMTypeRef typeA = typeOf(a);
+            final int bitWidthA = LLVMIRBuilder.integerTypeWidth(typeA);
+            assert bitWidthA == 32 || bitWidthA == 64;
+
+            LLVMValueRef shiftDistanceBitMask = builder.constantInteger(bitWidthA - 1, bitWidthA);
+            LLVMValueRef valB = emitIntegerConvert(b, typeA);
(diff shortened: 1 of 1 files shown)


koutheir:

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMGenerator.java
```diff
@@ -1469,24 +1469,36 @@ public Value emitXorFP(Value a, Value b) {
...
         }
 
+        private LLVMValueRef actualShiftingDistance(LLVMValueRef a, LLVMValueRef b) {
+            // https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.19
+
+            LLVMTypeRef typeA = typeOf(a);
+            int bitWidthA = LLVMIRBuilder.integerTypeWidth(typeA);
+            assert bitWidthA == 32 || bitWidthA == 64;
+
+            LLVMValueRef shiftDistanceBitMask = builder.constantInteger((bitWidthA == 32) ? 0x1f : 0x3f, bitWidthA);
```
loicottet, line 1479: I'd prefer `bitWidthA - 1`, as it does not rely on the assertion being checked to be correct.

loicottet (approved):

loicottet (commented):
Do you have an example where the LLVM backend behaviour is different from the expected one? I've tried to reproduce this but the result is consistent with the Java spec, so I'd like to be sure that we actually need this before adding an instruction to every shift that's emitted

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMGenerator.java
koutheir: Done.

koutheir:
> Do you have an example where the LLVM backend behaviour is different from the expected one?

I encountered this issue in a JCK test. I don't remember which one it was, but I remember it was testing negative shifts behavior.

> I've tried to reproduce this but the result is consistent with the Java spec, so I'd like to be sure that we actually need this before adding an instruction to every shift that's emitted.

Notice that negative shifts are only well-defined in Java. [C designates shifts that are negative or higher than the number of bits as undefined behavior](https://en.cppreference.com/w/c/language/operator_arithmetic). LLVM does not support negative shifts, and [reinterprets any negative shifts as unsigned integers](https://llvm.org/docs/LangRef.html#shl-instruction), which is different from the Java behavior (no modulo).

PR #4718, opened 2022-07-11
Fix the implementation of INSTANCEOF byte code instruction

koutheir:
[`INSTANCEOF`](https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-6.html#jvms-6.5.instanceof) must not throw an exception when the object is `null`, even if the type cannot be resolved.


koutheir:

compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java
```diff
@@ -1237,11 +1237,30 @@ protected void handleUnresolvedCheckCast(JavaType type, ValueNode object) {
      */
     protected void handleUnresolvedInstanceOf(JavaType type, ValueNode object) {
         assert !graphBuilderConfig.unresolvedIsError();
-        AbstractBeginNode successor = graph.add(new BeginNode());
```
gilles-duboscq, line 1240: Was this change necessary? Or was it just for symmetry with the SVM code?

This structure will just end up being turned into a `FixedGuardNode`, i think should simplify this to just emitting a `FixedGuardNode` right away (like the checkcast case). Unless it fixes some issue i'm missing.

koutheir: > Was this change necessary? Or was it just for symmetry with the SVM code?

Only for symmetry.

> This structure will just end up being turned into a FixedGuardNode, i think should simplify this to just emitting a FixedGuardNode right away (like the checkcast case). Unless it fixes some issue i'm missing.

I'll try that instead.

koutheir: `FixedGuardNode` is undocumented, can you explain how I am expected to use it?

koutheir:
I have no idea what is leading to the failure of the test `org.graalvm.compiler.truffle.test.CompilationMemoryTest#testFieldsFreedAfterCompilation` in the CI builds. Can someone help with this?

compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java
gilles-duboscq: `FixedGuardNode` is a control-flow node that ensures its condition input is true and deoptimizes if it's false.
You don't need to specify a probability since the condition is expected to be true

gilles-duboscq:
Regarding `CompilationMemoryTest` it looks like a transient issue.
@chumer, it looks like the value is still held by a running compilation somehow. Could it be that somehow 2 compilations get started instead of 1?

compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java
koutheir: >  You don't need to specify a probability since the condition is expected to be true.

My expectation in this situation is that most references input to `instanceof` will be non-null, and that is the opposite of what `FixedGuardNode` assumes when `IsNullNode.create(object)` is provided as a condition.

gilles-duboscq: I general that's true but the fact that the type is not resolved indicates that either the type cannot be resolved or it was never necessary to resolve it (we always get null). We don't really support the first case beyond correctness (i.e., we always deopt), and in the second case the implicit probability of the fixed guard node is correct.

If you have to give anything more than a very low probability to a deopt, something is wrong. That can happen for some cases that we don't really support in compiled mode when the compiler is used as a JIT.

koutheir: I followed your suggestion to mirror the implementation of `checkcast`, but I kept comments.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/phases/SharedGraphBuilderPhase.java
```diff
@@ -211,7 +216,28 @@ protected void handleUnresolvedNewMultiArray(JavaType type, ValueNode[] dims) {
...
+            // The INSTANCEOF byte code refers to a type that could not be resolved.
+            // INSTANCEOF must not throw an exception if the object is null.
+            BeginNode nullObj = graph.add(new BeginNode());
+            BeginNode nonNullObj = graph.add(new BeginNode());
+            append(new IfNode(graph.addOrUniqueWithInputs(IsNullNode.create(object)),
+                            nullObj, nonNullObj, BranchProbabilityNode.NOT_FREQUENT_PROFILE));
+
+            // Case where the object is not null, and type could not be resolved: Throw an
+            // exception.
+            if (lastInstr != null) {
```
gilles-duboscq, line 228: These
```java
if (lastInstr != null) {
    lastInstr.setNext(X);
}
```
blocks are suspicious, i guess they are just never taken, otherwise they would create a problem in the graph with the begin nodes having 2 predecessors: the `If` and something else.

In the first instance, we just appended an `If`, so the logic in `append`->`updateLastInstruction` will ensure `lastInstr` is null.
In the second instance, we just appended a control flow sink, so the same logic also ensures `lastInstr` is null.

koutheir: Methods that update `lastInstr` are not documented, and usage patterns of `lastInstr` are not clear to me. Should I remove these calls to `setNext()`?

gilles-duboscq: I think in general it shouldn't be necessary to update `lastInstr` manually. Methods like `append` should take care of it.
However, as you already found out, when creating some control-flow you'll need to set `lastInstr` to allow other methods to link what they are building with your control-flow structure.

I think the way you set `lastInstr` to the being nodes makes sense.

> Should I remove these calls to `setNext()`?

Yes

koutheir: Done.

mukel:
I'm integrating this PR, should be merged soon.

PR #4719, opened 2022-07-11
Fix the implementation of CHECKCAST byte code instruction.

koutheir:
[`CHECKCAST`](https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-6.html#jvms-6.5.checkcast) must not throw an exception when the object is `null`, even if the type cannot be resolved.

### DIFF
--- a/compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java
+++ b/compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java
@@ -1227,7 +1227,15 @@ public class BytecodeParser extends CoreProvidersDelegate implements GraphBuilde
      */
     protected void handleUnresolvedCheckCast(JavaType type, ValueNode object) {
         assert !graphBuilderConfig.unresolvedIsError();
+
+        // The CHECKCAST byte code refers to a type that could not be resolved.
+        // CHECKCAST must throw an exception if, and only if, the object is not null.
+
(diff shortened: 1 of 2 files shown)


compiler/src/org.graalvm.compiler.java/src/org/graalvm/compiler/java/BytecodeParser.java
```diff
@@ -1227,7 +1227,30 @@ protected void handleUnresolvedLoadConstant(JavaType type) {
      */
     protected void handleUnresolvedCheckCast(JavaType type, ValueNode object) {
         assert !graphBuilderConfig.unresolvedIsError();
-        append(new FixedGuardNode(graph.addOrUniqueWithInputs(IsNullNode.create(object)), Unresolved, InvalidateRecompile));
```
gilles-duboscq, line 1230: Was this change necessary? Or was it just for symmetry with the SVM code?

This structure will just end up being turned back into the `FixedGuardNode` so i think we should just keep this as is. Unless it fixes some issue i'm missing.

koutheir: > Was this change necessary? Or was it just for symmetry with the SVM code?

Only for symmetry.

> This structure will just end up being turned back into the FixedGuardNode so i think we should just keep this as is. Unless it fixes some issue i'm missing.

I'll try that instead.

mukel:
I'm integrating this PR, should be merged soon.

Issue #4647, opened 2022-06-15
[GR-39182] Provide `@CEntryPoint`s for advanced debugging


fniephaus:
To further improve the debugging experience of native executables, we need three `@CEntryPoint`s:

- An entry point for allocating an object given a certain class/type, and
- two entry points to pin and unpin an object to ensure that a GC neither collects nor moves them at debugging time.

Potentially, we might want to allow allocating an object that is already pinned if the operation needs to be atomic.

fniephaus:
@vvlevchenko has started to work on this and might provide an initial implementation for this soon.

peter-hofer:
See #5184 for the current design of the advanced debugging API using JNI, addressing the issues of #4723. I have been busy with the 22.3 release up until now and there are quite a few issues that still need to be addressed. @vvlevchenko

fniephaus:
We don't have plans to move this any further. Instead, we are currently working on #5648 and hopefully have something to show.

PR #4723, opened 2022-07-14
[GR-39182] set of @CEntryPoint functions to enrich evaluation functionality in IDE.

vvlevchenko:
Set of functions to enrich evaluation in IDE debugger.

String functions:
- svm_dbg_string_to_utf8
- svm_dbg_utf8_to_string

Array operations:
- svm_dbg_create_array
- svm_dbg_array_set
- svm_dbg_array_get

Class resolving functions:
- svm_dbg_classloader
- svm_dbg_class_for_name
- svm_dbg_allocate_object
- svm_dbg_class_modifier
- svm_dbg_class_fields
- svm_dbg_class_methods
- svm_dbg_class_super
- svm_dbg_class_interfaces

Garbage collection functions:
- svm_dbg_pin_object
- svm_dbg_unpin_object

Misc:
- svm_dbg_null

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/DebugHelper.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/DebugHelper.java
@@ -34,14 +34,24 @@ import com.oracle.svm.core.graal.snippets.CEntryPointSnippets;
 import com.oracle.svm.core.heap.Heap;
 import com.oracle.svm.core.heap.ObjectHeader;
 import com.oracle.svm.core.heap.ReferenceAccess;
+import com.oracle.svm.core.heap.VMOperationInfo;
+import com.oracle.svm.core.heap.VMOperationInfos;
 import com.oracle.svm.core.hub.DynamicHub;
 import com.oracle.svm.core.hub.LayoutEncoding;
 import com.oracle.svm.core.log.Log;
 import com.oracle.svm.core.snippets.KnownIntrinsics;
 import com.oracle.svm.core.stack.JavaStackFrameVisitor;
 import com.oracle.svm.core.stack.JavaStackWalker;
+import com.oracle.svm.core.thread.JavaVMOperation;
+import com.oracle.svm.core.thread.NativeVMOperation;
+import com.oracle.svm.core.thread.NativeVMOperationData;
+import com.oracle.svm.core.thread.VMOperation;
 import com.oracle.svm.core.thread.VMThreads;
 import com.oracle.svm.core.util.Utf8;
+import com.sun.jdi.ClassType;
+import com.sun.jdi.ObjectReference;
+import com.sun.jdi.ReferenceType;
+import com.sun.jdi.Value;
 import jdk.internal.misc.Unsafe;
 import org.graalvm.compiler.word.Word;
 import org.graalvm.nativeimage.IsolateThread;
@@ -298,6 +308,13 @@ public class DebugHelper {
     public static class IdeDebugHelper {
         /**
          * Convert instance of java.lang.String to utf8 c-string.
+         * This function used from debugger:
+         *  - in value renderer to Object rendering via Object::toString()
+         *  - as auxiliary function to convert super {@link #classGetSuper(IsolateThread, Pointer)} and
+         *      interfaces {@link #classGetInterfaces(IsolateThread, Pointer)} name on fetching class information
+         *  - to decode field and method information from methods {@link #classGetFields(IsolateThread, Pointer)}
+         *      and {@link #classGetMethods(IsolateThread, Pointer)}
+         * This method may be called when debugger is in the breakpoint and process is paused
          * @param thread
          * @param ptr pointer to instance of java.lang.String
          * @return c-string
@@ -311,18 +328,26 @@ public class DebugHelper {
 
         /**
          * Creates instance of java.lang.String from c-string.
+         * This method called from debugger:
+         *   - on enum processing to pass value to {@link Enum#valueOf(Class, String)}
+         *      {@link com.sun.jdi.ClassType#getValue(com.sun.jdi.Field)}
+         *   - to pass string literals from evaluator
+         * This method may be called when debugger is in breakpoint and process is paused
          * @param thread
          * @param ptr c-string
          * @return instance java.lang.String
          */
         @CEntryPoint(name = "svm_dbg_utf8_to_string", include = IncludeDebugHelperMethods.class, publishAs = Publish.SymbolOnly)
         @CEntryPointOptions(prologue = SetThreadAndHeapBasePrologue.class)
-        public static Object utf8ToString(@SuppressWarnings("unused") IsolateThread thread, CCharPointer ptr) {
-            return Utf8.utf8ToString(ptr);
+        public static Pointer utf8ToString(@SuppressWarnings("unused") IsolateThread thread, CCharPointer ptr) {
+            return Word.objectToUntrackedPointer(Utf8.utf8ToString(ptr));
         }
 
         /**
          * Create instance of an array.
+         * This method called from debugger:
+         *   - on array creation {@link com.sun.jdi.ArrayType#newInstance(int)}
+         * This method may be called when debugger in on the breakpoint and process is paused
          * @param thread
          * @param signature signature of array type.
          * @param length size of array
@@ -376,6 +401,8 @@ public class DebugHelper {
 
         /**
          * Set array's element with index position.
+         * This method used to implement {@link com.sun.jdi.ArrayReference#setValue(int, Value)}
+         * This method may be called when debugger in on the breakpoint and process is paused.
          * @param thread
          * @param ptrArray pointer to array object.
          * @param csignature type of array.
@@ -443,6 +470,11 @@ public class DebugHelper {
 
         /**
          * Returns index element of array.
+         * This method used for:
+         *   - implementing {@link com.sun.jdi.ArrayReference#getValue(int)}
+         *   - as auxiliary function to process return value of {@link #classGetInterfaces(IsolateThread, Pointer)},
+         *      {@link #classGetFields(IsolateThread, Pointer)}, {@link #classGetMethods(IsolateThread, Pointer)}
+         * This method called when process is paused or on breakpoint
          * @param thread
          * @param ptrArray pointer to array.
          * @param csignature type of array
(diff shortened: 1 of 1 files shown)


peter-hofer (commented):
I don't think these entry points should return objects, in fact I consider it an oversight that this currently does not lead to failure. The return values would be absolute addresses, but that is undefined behavior and it would be better to explicitly return pointers, like which are being passed for arguments -- I suppose there, we enforce non-object types.

vvlevchenko:
> I don't think these entry points should return objects, in fact I consider it an oversight that this currently does not lead to failure. The return values would be absolute addresses, but that is undefined behavior and it would be better to explicitly return pointers, like which are being passed for arguments -- I suppose there, we enforce non-object types.

Do you mean wrap return "objects"  with `Word.objectToUntrackedPointer`?

peter-hofer:
Yes (more like unwrap :wink:)

vvlevchenko:
> Yes (more like unwrap 😉)

Thank you, I've changed return types/values.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/DebugHelper.java
```diff
@@ -281,6 +295,493 @@ public static void printLocationInfo(@SuppressWarnings("unused") IsolateThread t
...
+         * Pin object.
+         * @param thread
+         * @param ptrObj pointer to object to pin.
+         * @return "pin object" for given object, should be passed svm_dbg_unpin_object.
+         */
+        @CEntryPoint(name = "svm_dbg_pin_object", include = IncludeDebugHelperMethods.class, publishAs = Publish.SymbolOnly)
+        @CEntryPointOptions(prologue = SetThreadAndHeapBasePrologue.class)
+        public static Pointer pinObject(@SuppressWarnings("unused") IsolateThread thread, Pointer ptrObj) {
+            Object obj = ptrObj.toObject();
+            return Word.objectToUntrackedPointer(PinnedObject.create(obj));
```
christianhaeubl, line 588: The `PinnedObject` itself is allocated on the Java heap and may therefore be moved by the GC.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/DebugHelper.java
```diff
@@ -281,6 +295,493 @@ public static void printLocationInfo(@SuppressWarnings("unused") IsolateThread t
...
+         * @param ptrClass pointer of class.
+         * @return array of fields descriptors in format name:modifier:type or name:modifier, depends on how much information
+         * Graal VM can provide.
+         */
+        @CEntryPoint(name = "svm_dbg_class_fields", include = IncludeDebugHelperMethods.class, publishAs = Publish.SymbolOnly)
+        @CEntryPointOptions(prologue = SetThreadAndHeapBasePrologue.class)
+        public static Pointer classGetFields(@SuppressWarnings("unused") IsolateThread thread, Pointer ptrClass) {
+            Class<?> clazz = (Class<?>)ptrClass.toObject();
+            DynamicHub hub = DynamicHub.fromClass(clazz);
+            Field[] fields = hub.getReachableFields();
```
christianhaeubl, line 629: Most of the fields won't be available because this only includes the fields that are used via reflection. Same applies to other methods that access metadata.

christianhaeubl (changes requested):
As I already mentioned in https://github.com/oracle/graal/issues/4647#issuecomment-1156500279 and via Slack, this current approach won't work for multiple reasons and can introduce arbitrary deadlocks or crashes. I only added a few exemplary comments to some of the methods that were added in this PR. Most methods that were introduced in this PR have similar issues though.

Please document the following for each method that you added, so that we can try to come up with a design for a working solution:
- How is the method used?
- What state are other application or VM-internals threads in when this method is called (a. stopped in gdb at arbitrary code location, b. stopped in gdb at specific code location, c. running)?
- For each object/pointer argument, document if this is an arbitrary object or if this is always an object that is returned by one of your methods.
- Document for each object/pointer return value how this value will be used (a. used temporarily while debugging, b. may be written to the Java heap).

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/DebugHelper.java
vvlevchenko: > Most of the fields won't be available because this only includes the fields that are used via reflection. Same applies to other methods that access metadata.

This is actually why we're have this method for fields and methods, to detect which fields and methods could be accessed from evaluator.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/DebugHelper.java
```diff
@@ -281,6 +295,493 @@ public static void printLocationInfo(@SuppressWarnings("unused") IsolateThread t
...
+    public static class IdeDebugHelper {
+        /**
+         * Convert instance of java.lang.String to utf8 c-string.
+         * @param thread
+         * @param ptr pointer to instance of java.lang.String
+         * @return c-string
+         */
+        @CEntryPoint(name = "svm_dbg_string_to_utf8", include = IncludeDebugHelperMethods.class, publishAs = Publish.SymbolOnly)
+        @CEntryPointOptions(prologue = SetThreadAndHeapBasePrologue.class)
+        public static CCharPointer stringToUtf8(@SuppressWarnings("unused") IsolateThread thread, Pointer ptr) {
```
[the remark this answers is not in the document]
christianhaeubl: Let me be more specific how this (and many of your other methods) may break.

Scenario A
- Thread A tries to initiate a safepoint.
- Thread B hits a gdb breakpoint. All threads are paused in gdb.
- Thread B calls any of your methods that allocates Java memory.
- Thread B needs to initiate a garbage collection (GC) and deadlocks because thread A is already trying to initiate a safepoint but is blocked in gdb.

Scenario B
- Thread A executes some VM internal code.
- Thread B hits a gdb breakpoint. All threads are paused in gdb.
- Thread B allocates Java memory and needs to refill its thread-local allocation buffer (TLAB). The TLAB refill deadlocks because it tries to acquire a lock that is already locked by thread A in some VM internal code.

Scenario C
- The user sets a breakpoint in some JDK code that is used VM-internally as well.
- Thread A hits that breakpoint and gets paused in gdb while executing VM-internal code that must be allocation free.
- Thread A calls one of your methods that allocates Java heap memory and either crashes right away or destroys VM-internal state.

Scenario D
- Thread A hits a breakpoint and gets paused in gdb.
- Thread A calls one of your methods that allocates Java heap memory and the allocation throws an `OutOfMemoryError`.
- The process crashes because the `@CEntryPoint` uses an exception handler that kills the process by default.

Scenario E (with dedicated VM operation thread)
- Thread A hits a gdb breakpoint. All threads are paused in gdb.
- Thread A calls one of your methods that allocates Java memory.
- Thread A needs to initiate a GC and deadlocks because the dedicated VM operation thread is paused as well, so no one is going to execute the requested GC.

Scenario F
- Thread A hits a gdb breakpoint. Only thread A is paused in gdb all other threads are still running.
- Thread A calls `svm_dbg_allocate_object(...)`, which returns a raw pointer to a newly allocate object (`obj1`).
- Thread A calls `svm_dbg_allocate_object(...)` a second time. This time, the allocation needs a GC, which moves `obj1` that was returned by the first call. If this outdated pointer is access, it may result in crashes or may return invalid data.

Those are the scenarios that come to my mind without thinking about it too much in detail. I am definitely sure that there are more complex scenarios as well that I completely missed in this rather quick analysis and that will cause similar or worse issues.

Please provide the additional information that I requested yesterday so that I can try to come up with a design that doesn't have those issues.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/DebugHelper.java
```diff
@@ -281,6 +295,493 @@ public static void printLocationInfo(@SuppressWarnings("unused") IsolateThread t
...
+            Object obj = ptr.toObject();
+            return CTypeConversion.toCString((String) obj).get();
```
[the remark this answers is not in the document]
christianhaeubl: This leaks memory because `CTypeConversion.toCString(...)` allocates a `CCharPointerHolder` that contains the pinned object. You would need to call `CCharPointerHolder.close()` once you don't need the native memory anymore.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/DebugHelper.java
christianhaeubl: If the heap dumping support is enabled, we have information about all the fields that are present in the Native Image (in a compact, encoded form but it should still be good enough for the debugging support).

peter-hofer:
Like @chaeubl has already noted, we believe JNI is a better fit to support comprehensive debugging. The existing debug helper functions are intended as simple tools for direct command-line debugging.

Many of the functions in this PR already exist in JNI: string conversions, array access, class and member lookups, and object allocation. In JNI, objects are referred to by handles, which is more robust than addresses which can change during GC or pinning objects which needs bookkeeping. You can use JNI handle frames (`PushLocalFrame`) for successive JNI calls to clean up all their handles at once afterwards. Note that handles are thread-local by default and have to be recreated as global (`NewGlobalRef`) to be used in other threads (in which case they have to be explicitly destroyed later). JNI also offers functions for efficient access of primitive arrays.

Unlike for a Java application on HotSpot, our reflection dictionary (which you access here via `svm_dbg_class_for_name`, `svm_dbg_class_fields` and `svm_dbg_class_methods`) is not complete, but contains only elements which have been registered for reflective access beforehand (see our documentation on `reflect-config.json`). The same applies to the JNI dictionary (`jni-config.json`), and the reflection and JNI dictionaries can be different (and typically are). This means that any debug helper methods that you want to call via JNI must be registered during the image build, and that you can't call arbitrary methods or access arbitrary fields of arbitrary classes. This wouldn't be possible without JNI either though, because fields or methods or entire classes can be proven unreachable and excluded during the image build, and methods can be inlined at all call sites without ever being compiled individually.

Regarding GC and other VM operations, you will still need to consider which threads are suspended by the debugger because doing a JNI call or just single-stepping while other threads are suspended can block until another thread completes a pending VM operation. Another restriction is that the debugger should not be used on the JNI implementation itself (or any other VM internals, really) because this will most likely cause it to change data structures as they are used (particularly the handle tables).

In order to get started, we need a concept for a _debugger context_. I'm assuming that we always start out at a breakpoint in Java code, while JNI functions are intended to be called from native code. We could try putting the thread into native mode, or we could use the Java implementations of the JNI functions directly -- both approaches have their pitfalls and likely need some changes on our side. Then, we need to get the thread's JNI environment (this one is straightforward) and a way to create JNI handles for the object variables that we can access in the debugger. I'll have to experiment a bit to figure out how to best approach this and will also have a closer look at how/if NativeJDB does it.

Please let me know what you think and if there's something more we need to consider here.

PR #4725, opened 2022-07-14
Fixed #4724: -XX:+PrintGCTimes fails with assertions enabled

### LINKED ISSUE #4724: `-XX:+PrintGCTimes` fails with assertions enabled
The `com.oracle.svm.core.genscavenge.Timer` class seems to have as invariant that `wasOpened` is true iff `openNanos != 0`. The code at line 83 breaks that invariant by setting `openNanos` without setting `wasOpened`:

https://github.com/oracle/graal/blob/a7210c0b1f119a66d6c2504ae5e2e4dd31494778/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/Timers.java#L83

To reproduce, run `native-image -esa ...`, then run the resulting binary with `-XX:+VerboseGC -XX:+PrintGCTimes`. It will crash.

petermz:
The fix is to return the value without setting the field


christianhaeubl:
Thanks this change will be merged to master in the next few days.

PR #4799, opened 2022-08-08
Give host code a chance to create non-modifiable polyglot array

JaroslavTulach:
We are working on an implementation of a [functional programming language](http://enso.org) on top of Truffle. One of the important features of functional languages is [referential transparency](https://en.wikipedia.org/wiki/Referential_transparency). We can reach it in our language, but we have problems achieving it effectively when dealing with other languages across polyglot boundary.

For example, when we receive an object which _hasArrayElements_, we'd like to know whether it is _safe to use it_ as is, or whether we shall _defensively copy_ its content. Usually we have both sides (the Enso one and the Java one) under our control. E.g. the problem is just about communicating between these two whether the array can be considered immutable or not.

It has come to my attention that there is [isArrayElementModifiable](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/interop/InteropLibrary.html#isArrayElementModifiable-java.lang.Object-long-) which we could use to ask the array. Assuming that if all elements we read are not modifiable, then the array won't change. I consider it a good heuristic.

However then I tried to investigate what needs to be done on the (host) Java side to mark an array (or rather a `List`) as read only. I've noticed there is no such way (probably the `isArrayElementXyz` messages were added later than polyglot SDK and nobody needed to expose them yet). This PR is my attempt to allow (host) Java to express that a `List` is not mutable by using check for [friendly class implementations](https://github.com/oracle/graal/blob/01cdd0ac7b496711b8afd1dbc6649e335c4f6465/truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostObject.java#L104). I don't think the current code is mergable, but it should be good enough to start discussion.

What do you think, @chumer, @tzezula, @jdunkerley, @kustosz, @radeusgd, @hubertp?

### DIFF
--- a/truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/ProxyAPITest.java
+++ b/truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/ProxyAPITest.java
@@ -84,6 +84,8 @@ import org.junit.Before;
 import org.junit.Test;
 
 import com.oracle.truffle.tck.tests.ValueAssert.Trait;
+import java.util.Collections;
+import static org.junit.Assert.fail;
 
 /**
  * Testing the behavior of proxies API and proxy interfaces.
@@ -137,6 +139,21 @@ public class ProxyAPITest {
             return getSize.get();
         }
 
+    }
+
+    @Test
+    public void readOnlyUnmodifiableList() {
+        List<String> proxy = Collections.unmodifiableList(Collections.nCopies(10, "A"));
+        Value value = context.asValue(proxy);
+        assertTrue(value.hasArrayElements());
+        try {
+            value.setArrayElement(0, "B");
+        } catch (RuntimeException ex) {
+            assertEquals("Read-only array element", ex.getMessage());
+            assertEquals(UnsupportedOperationException.class, ex.getClass());
+            return;
+        }
+        fail("Runtime exception shall have been thrown");
     }
 
     @Test
--- a/truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostObject.java
+++ b/truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostObject.java
@@ -517,6 +517,9 @@ final class HostObject implements TruffleObject {
                         @Shared("isList") @Cached IsListNode isList,
                         @Shared("error") @Cached BranchProfile error) {
             try {
+                if (isReadOnlyList(receiver)) {
+                    return false;
+                }
                 long size = GuestToHostCalls.getListSize(receiver);
                 return index >= 0 && index < size;
             } catch (Throwable t) {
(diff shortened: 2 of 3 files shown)


JaroslavTulach:
> When singing the OCA

Cute. _Óóó sééé Ej_!

chumer:
isArrayElementWritable false only tells you that an element is not writable, not that it is immutable like you assume it. There is no immutable contract in interop you could reliably use for that. So I do not think you will get  around calling into a host-specific method to find out whether the List is immutable according to  host Java. The way you do this, by comparing the class, is also not really a specified way(or is it?), so I think you are on your own there.

I think we do have a bug here though. I see that WriteArrayElement#doList currently does not currently catch UnsupportedOperationException. It should probably throw an UnsupportedMessageException for UnsupportedOperationException.

JaroslavTulach:
> only tells you that an element is not writable, not that it is immutable

Yup, I know, it is just my interpretation.

> The way you do this, by comparing the class, is also not really a specified way(or is it?)

Comparing implementation class is already used in the _host interop_ package: https://github.com/oracle/graal/blob/01cdd0ac7b496711b8afd1dbc6649e335c4f6465/truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostObject.java#L104

JaroslavTulach:
However the real question for SDK & Truffle API:

> isArrayElementWritable false only tells you that an element is not writable

How does one create an object that `hasArrayElements` and (some of its) fields are `isArrayElementWritable == false` in host Java? Looks to me `Collections.unmodifyiableList` isn't enough right now, neither `ProxyArray` is enough...

Doing the implementation class check is one way. Having an `ProxyReadOnlyArray` interface another one that comes to my mind. Any contract will do, but right now there is just no way.

chumer:
> Comparing implementation class is already used in the host interop package:

That is done for performance reasons only. It's not observable by the user, as it would be if isArrayElementWritable were implemented that way.

> How does one create an object that hasArrayElements and (some of its) fields are isArrayElementWritable == false in host Java? 

Not everything is exposed to host java. So you cannot directly influence every interop message. But the semantics of interop are that you check whether isArrayElementWritable and then you call into writeArrayElement. WriteArrayElement is then still allowed to throw an UnsupportedMessageException. This can of course, have slightly different semantics in the languages, but in the proxy interfaces, there is only the latter way available, that unfortunately is currently bugged. As already mentioned.

> Doing the implementation class check is one way. Having an ProxyReadOnlyArray interface another one that comes to my mind. Any contract will do, but right now there is just no way.

One idea: Could you only implement ProxyIterable as a workaround? Or do you need array semantics?

I think I prefer ProxyReadOnlyArray. We could introduce ProxyReadOnlyArray as a base class to  ProxyArray. That should be compatible, right?

JaroslavTulach:
> > Doing the implementation class check is one way. Having an `ProxyReadOnlyArray` interface another one that comes to my mind. Any contract will do, but right now there is just no way.
> 
> One idea: Could you only implement ProxyIterable as a workaround? Or do you need array semantics?

Enso base type is `Vector` with `at` method. As such we need array semantics without needless copying.

> I think I prefer ProxyReadOnlyArray. We could introduce ProxyReadOnlyArray as a base class to ProxyArray. That should be compatible, right?

Adding a super interface `ProxyReadOnlyArray` with subset of `ProxyArray` methods would be compatible change. I'll see whether I can prepare it soon.

JaroslavTulach:
At the end we plan to [consolidate Vector & Array](https://www.pivotaltracker.com/n/projects/2539304/stories/183217534) rather than trying to differentiate between them.

Issue #4810, opened 2022-08-11
Linux perf does not consume the current debuginfo format


adinn:
**Describe the issue**
On Linux the perf tool provides details information about code execution and performance. It should be able to use image debuginfo to derive function names, files and line numbers for compiled method code, including inlined method code. There are several features of the current GraalVM native debug info layout that, while valid as DWARF and recognized by gdb, do not match perf's expectations. The debuginfo layout should be modified to accommodate perf's expectations as follows: 

1. Method declarations need to have an associated line number in order for perf to recognize them (we can use whatever line is associated with bytecode zero)
2. `.debug_arange` section entries need to be sorted low to high by range start address (this was always meant to be the case ...)
3. Concrete inline method definitions need to refer to an abstract inline method declaration replicated inliner's compile unit (CU) for the inliner's class using a CU local offset. At present they refer to a single definition in the CU of the class that declares the inlined method using a section global offset.
4. Inline method declarations also need to replicate their original method's parameters/local declarations in the importing class's CU so that the latter can be referred to from local/parameter location definitions using a CU local offset. At present local/parameter location definitions refer to a single, corresponding local/parameter declaration in the CU of the class that declares the inlined method.
5. CUs for instances classes whose methods only exist inline need to include a `stmt_list` attribute identifying an associated `.debug_line` subsection detailing line info for the class's methods. Clearly, this subsection will include no actual line number records. However, perf requires it so as to provide a dir and file table identifying the source path and file for the CU's class.
6. Perf expects method declarations to include a `linkage_name` attribute referencing a string section entry populated with the fully qualified method name (formatted using format specifier `"%H.%n"`). Unlike gdb, perf will not synthesize method names from other DWARF data. If a `linkage_name` attribute is omitted, perf picks up the symbol table entry with the nearest address below a profile target as a label for the corresponding function.
n.b. This field is also used by the debugger to present the method in debug backtraces etc, meaning that this patch results in some minor cosmetic changes to debug output. This also means that it would a very bad idea to use the GraalVM synthesized symbol name as the value of the linkage_name attribute (includes a message digest encoding of the arguments). 

**Steps to reproduce the issue**
Use `perf record -g` to record profile events for a generated native built with debug info enabled.
Use `perf report` to view the resulting profile info.
Java code events are reported either using a hex address or as if they belong to an unrelated C++ function such as _start.

**Describe GraalVM and your environment:**
Latest GraalVM head on Linux.

adinn:
[...]

PR #4811, opened 2022-08-11
[GR-40874] Perf support

adinn:
This PR implements changes to Linux debug info generation specified in issue #4810. These enable the Linux tool `perf` to display profile information for GraalVM native images.

This patch 'fixes' `perf` modulo from a few small remaining details:

`perf` expects many references from definitions to associated declarations (`specification` or `abstract_origin` attribute values) to be implemented as compile unit (CU) local offsets because it assumes that a single CU owns both the referring and referenced records. There are still a few cases in Java debug info where definitions and declarations need to be in different CUs and so must use an info section global offset. This is no problem for `gdb` but `perf` cannot cope with it.

`perf` does display generated machine code for sampled code addresses. However, it appears to be adding an extra 4K offset to the code address before decompiling, meaning that the wrong code is displayed. I am not clear why this is happening.

`perf` occasionally lists sampled method code addresses as hex numbers (e.g. something like `0x4061ba`) instead of using the fully qualified method name. I am not clear why this is happening. The odd thing is that `perf` delegates to binary `addr2line` to do name, file and line lookup and this program prints the correct name, file, line and even inline frame info for these addresses.

### DIFF
--- a/docs/reference-manual/native-image/DebugInfo.md
+++ b/docs/reference-manual/native-image/DebugInfo.md
@@ -18,6 +18,8 @@ The resulting image will contain debug records in a format the GNU Debugger (GDB
 Additionally, you can pass `-O0` to the builder which specifies that no compiler optimizations should be performed.
 Disabling all optimizations is not required, but in general it makes the debugging experience better.
 
+Debug information is not just useful to the debugger. It can also be used by the Linux performance profiling tools `perf` and `valgrind` to correlate execution statistics such as CPU utilization or cache misses with specific, named Java methods and even link them to individual lines of Java code in the original Java source file.
+
 By default, debug info will only include details of some of the values of parameters and local variables.
 This means that the debugger will report many parameters and local variables as being undefined. If you pass `-O0` to the builder then full debug information will be included.
 If you
@@ -35,7 +37,21 @@ However, it can significantly increase the size of the generated image on disk. 
 and local variable information by passing flag `-H:+SourceLevelDebug` can cause a program to be compiled
 slightly differently and for some applications this can slow down execution.
 
+The basic `perf report` command, which displays a histogram showing percentage execution time in each Java method , only requires passing flags `-g` and `-H:+SourceLevelDebug` to the `native-image` command.
+However, more sophisticated uses of `perf` (i.e. `perf annotate`) and use of
+`valgrind` requires debug info to be supplemented with linkage symbols identifying compiled Java methods.
+Java method symbols are omitted from the generated native image by default but they can be retained achieved by passing one extra flag to the `native-image` command
+
+```shell
+native-image -g -H:+SourceLevelDebug -H:-DeleteLocalSymbols Hello
+```
+
+Use of this flag will result in a small increase in the size of the
+resulting image file.
+
 > Note: Native Image debugging currently works on Linux with initial support for macOS. The feature is experimental.
+
+> Note: Debug info support for `perf` and `valgrind` on Linux is an experimental feature.
 
 ### Table of Contents
 
@@ -46,6 +62,7 @@ slightly differently and for some applications this can slow down execution.
 - [Checking Debug Info on Linux](#checking-debug-info-on-linux)
 - [Debugging with Isolates](#debugging-with-isolates)
 - [Debugging Helper Methods](#debugging-helper-methods)
+- [Special Considerations for using perf and valgrind](#using perf and valgrind)
 
 ## Source File Caching
 
@@ -773,3 +790,157 @@ For example, calling the method below prints high-level information about the Na
 ### Further Reading
 
 - [Debugging Native Image in VS Code](Debugging.md)
+
+## Special Considerations for using perf and valgrind
+
+Debug info includes details of address ranges for top level and
+inlined compiled method code as well as mappings from code addresses
+to the corresponding source files and lines.
+`perf` and `valgrind` are able to use this information for some of
+their recording and reporting operations.
+For example, `perf report` is able to associate code adresses sampled
+during a `perf record` session with Java methods and print the
+DWARF-derived method name for the method in its output histogram.
+
+```
+    . . .
+    68.18%     0.00%  dirtest          dirtest               [.] _start
+            |
+            ---_start
+               __libc_start_main_alias_2 (inlined)
+               |          
+               |--65.21%--__libc_start_call_main
+               |          com.oracle.svm.core.code.IsolateEnterStub::JavaMainWrapper_run_5087f5482cc9a6abc971913ece43acb471d2631b (inlined)
+               |          com.oracle.svm.core.JavaMainWrapper::run (inlined)
+               |          |          
+               |          |--55.84%--com.oracle.svm.core.JavaMainWrapper::runCore (inlined)
+               |          |          com.oracle.svm.core.JavaMainWrapper::runCore0 (inlined)
+               |          |          |          
+               |          |          |--55.25%--DirTest::main (inlined)
+               |          |          |          |          
+               |          |          |           --54.91%--DirTest::listAll (inlined)
+               . . .
+```              
+
+Unfortunately, other operations require Java methods to be identified
+by an ELF (local) function symbol table entry locating the start of
+the compiled method code.
+In particular, assembly code dumps provided by both tools identify
+branch and call targets using an offset from the nearest symbol.
+Omitting Java method symbols means that offsets are generally
+displayed relative to some unrelated global symbol, usually the entry
+point for a method exported for invocation by C code.
+
+As an illustration of the problem, the following excerpted output from
+`perf annotate` displays the first few annotated instructions of the
+compiled code for method `java.lang.String::String()`.
+
+```
+    . . .
+         : 501    java.lang.String::String():
+         : 521    public String(byte[] bytes, int offset, int length, Charset charset) {
+    0.00 :   519d50: sub    $0x68,%rsp
+    0.00 :   519d54: mov    %rdi,0x38(%rsp)
+    0.00 :   519d59: mov    %rsi,0x30(%rsp)
+    0.00 :   519d5e: mov    %edx,0x64(%rsp)
+    0.00 :   519d62: mov    %ecx,0x60(%rsp)
+    0.00 :   519d66: mov    %r8,0x28(%rsp)
+    0.00 :   519d6b: cmp    0x8(%r15),%rsp
+    0.00 :   519d6f: jbe    51ae1a <graal_vm_locator_symbol+0xe26ba>
+    0.00 :   519d75: nop
+    0.00 :   519d76: nop
+         : 522    Objects.requireNonNull(charset);
+    0.00 :   519d77: nop
+         : 524    java.util.Objects::requireNonNull():
+         : 207    if (obj == null)
+    0.00 :   519d78: nop
+    0.00 :   519d79: nop
+         : 209    return obj;
+    . . .
+```
+
+The leftmost column shows percentages for the amount of time recorded
+at each instruction in samples obtained during the `perf record` run.
+Each instruction is prefaced with it's address in the program's code
+section.
+The disassembly interleaves the source lines from which the code is
+derived, 521-524 for the top level code and 207-209 for the code
+inlined from from `Objects.requireNonNull()`.
+(note, however that the lines for the quoted comments are being computed
+and displayed incorrectly).
+Also, the start of the method is labelled with the name defined in the
+DWARF debug info, `java.lang.String::String()`.
+However, the branch instruction `jbe` at address `0x519d6f` uses a
+very large offset from `graal_vm_locator_symbol` to identify what is
+actually a target address within the compiled code for
+`String::String()`.
+
+Readability of the tool output is significantly improved if
+option `-H-DeleteLocalSymbols` is passed to the `native-image`
+command.
+The equivalent `perf annotate` output with this option enabled is as
+follows:
+
+```
+    . . .
+         : 5      000000000051aac0 <String_constructor_f60263d569497f1facccd5467ef60532e990f75d>:
+         : 6      java.lang.String::String():
+         : 521    *          {@code offset} is greater than {@code bytes.length - length}
+         : 522    *
+         : 523    * @since  1.6
+         : 524    */
+         : 525    @SuppressWarnings("removal")
+         : 526    public String(byte[] bytes, int offset, int length, Charset charset) {
+    0.00 :   51aac0: sub    $0x68,%rsp
+    0.00 :   51aac4: mov    %rdi,0x38(%rsp)
+    0.00 :   51aac9: mov    %rsi,0x30(%rsp)
+    0.00 :   51aace: mov    %edx,0x64(%rsp)
+    0.00 :   51aad2: mov    %ecx,0x60(%rsp)
+    0.00 :   51aad6: mov    %r8,0x28(%rsp)
+    0.00 :   51aadb: cmp    0x8(%r15),%rsp
+    0.00 :   51aadf: jbe    51bbc1 <String_constructor_f60263d569497f1facccd5467ef60532e990f75d+0x1101>
+    0.00 :   51aae5: nop
+    0.00 :   51aae6: nop
+         : 522    Objects.requireNonNull(charset);
+    0.00 :   51aae7: nop
+         : 524    java.util.Objects::requireNonNull():
+         : 207    * @param <T> the type of the reference
+         : 208    * @return {@code obj} if not {@code null}
+         : 209    * @throws NullPointerException if {@code obj} is {@code null}
+         : 210    */
+         : 211    public static <T> T requireNonNull(T obj) {
+         : 212    if (obj == null)
+    0.00 :   51aae8: nop
+    0.00 :   51aae9: nop
+         : 209    throw new NullPointerException();
+         : 210    return obj;
+    . . .
+```
+
+Note that the start address of the method is now labelled with the
+mangled symbol name `String_constructor_f60263d569497f1facccd5467ef60532e990f75d` as well as the DWARF name.
+
+Unfortunately, `perf` and `valgrind` do not correctly understand the
+mangling algorithm employed by GraalVM, nor are they currently able to
+replace the mangled name with the DWARF name in the disassembly even
+though both symbol and DWARF function data are known to idntify code
+starting at the same address.
+So, the branch instruction still prints its target using a symbol plus
+offset but it is at least using the method symbol this time.
+
+Executing command `perf annotate` will provide a disassembly listing
+for all methods and C functions in the image.
+It is possible to annotate a specific method by passing it's name as
+an argument to the perf annotate command.
+Note, however, that `perf` requries the mangled symbol name as
+argument rather than the DWARF name.
+So, in order to annotate method `java.lang.String::String()` it is
+necessary to run command `perf annotate
+String_constructor_f60263d569497f1facccd5467ef60532e990f75d`.
+
+The `valgrind` tool `callgrind` also requires local symbols to be
+retained in order to provide high quality output.
+When `callgrind` is used in combination with a viewer like
+`kcachegrind` it is possible to identify a great deal of valuable
+information about native image execution aand relate it back to
+specific source code lines.
--- a/docs/reference-manual/native-image/DebugInfo.md
+++ b/docs/reference-manual/native-image/DebugInfo.md
@@ -37,7 +37,7 @@ However, it can significantly increase the size of the generated image on disk. 
 and local variable information by passing flag `-H:+SourceLevelDebug` can cause a program to be compiled
 slightly differently and for some applications this can slow down execution.
 
-The basic `perf report` command, which displays a histogram showing percentage execution time in each Java method , only requires passing flags `-g` and `-H:+SourceLevelDebug` to the `native-image` command.
+The basic `perf report` command, which displays a histogram showing percentage execution time in each Java method, only requires passing flags `-g` and `-H:+SourceLevelDebug` to the `native-image` command.
 However, more sophisticated uses of `perf` (i.e. `perf annotate`) and use of
 `valgrind` requires debug info to be supplemented with linkage symbols identifying compiled Java methods.
 Java method symbols are omitted from the generated native image by default but they can be retained achieved by passing one extra flag to the `native-image` command
@@ -866,14 +866,14 @@ section.
 The disassembly interleaves the source lines from which the code is
 derived, 521-524 for the top level code and 207-209 for the code
 inlined from from `Objects.requireNonNull()`.
-(note, however that the lines for the quoted comments are being computed
-and displayed incorrectly).
 Also, the start of the method is labelled with the name defined in the
 DWARF debug info, `java.lang.String::String()`.
 However, the branch instruction `jbe` at address `0x519d6f` uses a
-very large offset from `graal_vm_locator_symbol` to identify what is
-actually a target address within the compiled code for
-`String::String()`.
+very large offset from `graal_vm_locator_symbol`.
+The printed offset does identify the correct address relative to the
+location of the symbol.
+However, this fails to make clear that the target address actually
+lies within the compiled code range for method `String::String()` i.e. that thsi is a method-local branch.
 
 Readability of the tool output is significantly improved if
 option `-H-DeleteLocalSymbols` is passed to the `native-image`
@@ -917,16 +917,25 @@ follows:
     . . .
 ```
 
-Note that the start address of the method is now labelled with the
-mangled symbol name `String_constructor_f60263d569497f1facccd5467ef60532e990f75d` as well as the DWARF name.
+In this version the start address of the method is now labelled with
+the mangled symbol name `String_constructor_f60263d569497f1facccd5467ef60532e990f75d`
+as well as the DWARF name.
+The branch target is now printed using an offset from that start
+symbol.
 
 Unfortunately, `perf` and `valgrind` do not correctly understand the
 mangling algorithm employed by GraalVM, nor are they currently able to
 replace the mangled name with the DWARF name in the disassembly even
-though both symbol and DWARF function data are known to idntify code
+though both symbol and DWARF function data are known to identify code
 starting at the same address.
 So, the branch instruction still prints its target using a symbol plus
 offset but it is at least using the method symbol this time.
+
+Also, because address `51aac0` is now recognized as a method start,
+`perf` has preceded the first line of the method with 5 context lines,
+which list the tail end of the method's javadoc comment.
+Unfortunately, perf has numbered these lines incorrectly, labelling
+the first comment with 521 rather than 516.
 
 Executing command `perf annotate` will provide a disassembly listing
 for all methods and C functions in the image.
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -31,6 +31,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import jdk.vm.ci.meta.ResolvedJavaMethod;
 import jdk.vm.ci.meta.ResolvedJavaType;
@@ -71,12 +72,20 @@ public class ClassEntry extends StructureTypeEntry {
      */
     private Map<ResolvedJavaMethod, MethodEntry> methodsIndex;
     /**
-     * A list recording details of all primary ranges included in this class sorted by ascending
-     * address range.
+     * A list recording details of all normal primary compiled methods included in this class sorted
+     * by ascending address range. Note that the associated address ranges are disjoint and
+     * contiguous.
      */
     private List<PrimaryEntry> primaryEntries;
     /**
-     * An index identifying primary ranges which have already been encountered.
+     * A list recording details of all deopt fallback primary compiled methods included in this
+     * class sorted by ascending address range. Note that the associated address ranges are
+     * disjoint, contiguous and above all ranges for normal primary compiled methods.
+     */
+    private List<PrimaryEntry> deoptPrimaryEntries;
+    /**
+     * An index identifying ranges for primary compiled method which have already been encountered,
+     * whether normal or deopt fallback methods.
      */
     private Map<Range, PrimaryEntry> primaryIndex;
     /**
@@ -95,10 +104,6 @@ public class ClassEntry extends StructureTypeEntry {
      * A list of the same dirs.
      */
     private List<DirEntry> localDirs;
-    /**
-     * This flag is true iff the entry includes methods that are deopt targets.
-     */
-    private boolean includesDeoptTarget;
 
     public ClassEntry(String className, FileEntry fileEntry, int size) {
         super(className, size);
@@ -107,6 +112,8 @@ public class ClassEntry extends StructureTypeEntry {
         this.methods = new ArrayList<>();
         this.methodsIndex = new HashMap<>();
         this.primaryEntries = new ArrayList<>();
+        // deopt methods list is created on demand
+        this.deoptPrimaryEntries = null;
         this.primaryIndex = new HashMap<>();
         this.localFiles = new ArrayList<>();
         this.localFilesIndex = new HashMap<>();
@@ -154,13 +161,16 @@ public class ClassEntry extends StructureTypeEntry {
     public void indexPrimary(Range primary, List<DebugFrameSizeChange> frameSizeInfos, int frameSize) {
         if (primaryIndex.get(primary) == null) {
             PrimaryEntry primaryEntry = new PrimaryEntry(primary, frameSizeInfos, frameSize, this);
-            primaryEntries.add(primaryEntry);
             primaryIndex.put(primary, primaryEntry);
             if (primary.isDeoptTarget()) {
-                includesDeoptTarget = true;
+                if (deoptPrimaryEntries == null) {
+                    deoptPrimaryEntries = new ArrayList<>();
+                }
+                deoptPrimaryEntries.add(primaryEntry);
             } else {
+                primaryEntries.add(primaryEntry);
                 /* deopt targets should all come after normal methods */
-                assert includesDeoptTarget == false;
+                assert deoptPrimaryEntries == null;
             }
             FileEntry primaryFileEntry = primary.getFileEntry();
             if (primaryFileEntry != null) {
@@ -247,13 +257,41 @@ public class ClassEntry extends StructureTypeEntry {
         return fileEntry;
     }
 
-    public List<PrimaryEntry> getPrimaryEntries() {
-        return primaryEntries;
+    /**
+     * Retrieve a stream of all primary compiled method entries for this class, including both
+     * normal and deopt fallback compiled methods.
+     * 
+     * @return a stream of all primary compiled method entries for this class.
+     */
+    public Stream<PrimaryEntry> primaryEntries() {
+        Stream<PrimaryEntry> stream = primaryEntries.stream();
+        if (deoptPrimaryEntries != null) {
+            stream = Stream.concat(stream, deoptPrimaryEntries.stream());
+        }
+        return stream;
     }
 
-    @SuppressWarnings("unused")
-    public Object primaryIndexFor(Range primaryRange) {
-        return primaryIndex.get(primaryRange);
+    /**
+     * Retrieve a stream of all normal primary compiled method entries for this class, excluding
+     * deopt fallback compiled methods.
+     * 
+     * @return a stream of all normal primary compiled method entries for this class.
+     */
+    public Stream<PrimaryEntry> normalPrimaryEntries() {
+        return primaryEntries.stream();
+    }
+
+    /**
+     * Retrieve a stream of all deopt fallback primary compiled method entries for this class.
+     * 
+     * @return a stream of all deopt fallback primary compiled method entries for this class.
+     */
+    public Stream<PrimaryEntry> deoptPrimaryEntries() {
+        if (includesDeoptTarget()) {
+            return primaryEntries.stream();
+        } else {
+            return Stream.empty();
+        }
     }
 
     public List<DirEntry> getLocalDirs() {
@@ -265,7 +303,7 @@ public class ClassEntry extends StructureTypeEntry {
     }
 
     public boolean includesDeoptTarget() {
-        return includesDeoptTarget;
+        return deoptPrimaryEntries != null;
     }
 
     public String getCachePath() {
@@ -371,47 +409,60 @@ public class ClassEntry extends StructureTypeEntry {
         return methods;
     }
 
+    /*
+     * Accessors for lo and hi bounds of this class's compiled method code ranges. See comments in
+     * class DebugInfoBase for an explanation of the layout of compiled method code.
+     */
+
+    /**
+     * Retrieve the lowest code section offset for compiled method code belonging to this class. It
+     * is an error to call this for a class entry which has no compiled methods.
+     * 
+     * @return the lowest code section offset for compiled method code belonging to this class
+     */
     public int lowpc() {
         assert isPrimary();
         return primaryEntries.get(0).getPrimary().getLo();
     }
 
+    /**
+     * Retrieve the lowest code section offset for compiled method code belonging to this class that
+     * belongs to a deoptimization fallback compiled method. It is an error to call this for a class
+     * entry which has no deoptimization fallback compiled methods.
+     * 
+     * @return the lowest code section offset for a deoptimization fallback compiled method
+     *         belonging to this class.
+     */
     public int lowpcDeopt() {
         assert isPrimary();
         assert includesDeoptTarget();
-        int lowpc = -1;
-        for (PrimaryEntry primaryEntry : primaryEntries) {
-            Range primary = primaryEntry.getPrimary();
-            if (primary.isDeoptTarget()) {
-                lowpc = primary.getLo();
-                break;
-            }
-        }
-        assert lowpc >= 0;
-        return lowpc;
+        return deoptPrimaryEntries.get(0).getPrimary().getLo();
     }
 
+    /**
+     * Retrieve the highest code section offset for compiled method code belonging to this class
+     * that does not belong to a deoptimization fallback compiled method. The returned value is the
+     * offset of the first byte that succeeds the code for that method. It is an error to call this
+     * for a class entry which has no compiled methods.
+     * 
+     * @return the highest code section offset for compiled method code belonging to this class
+     */
     public int hipc() {
         assert isPrimary();
-        if (!includesDeoptTarget()) {
-            return primaryEntries.get(primaryEntries.size() - 1).getPrimary().getHi();
-        } else {
-            Range lastPrimary = null;
-            for (PrimaryEntry primaryEntry : primaryEntries) {
-                Range primary = primaryEntry.getPrimary();
-                if (primary.isDeoptTarget()) {
-                    break;
-                }
-                lastPrimary = primary;
-            }
-            assert lastPrimary != null;
-            return lastPrimary.getHi();
-        }
+        return primaryEntries.get(primaryEntries.size() - 1).getPrimary().getHi();
     }
 
+    /**
+     * Retrieve the highest code section offset for compiled method code belonging to this class
+     * that belongs to a deoptimization fallback compiled method. It is an error to call this for a
+     * class entry which has no deoptimization fallback compiled methods.
+     * 
+     * @return the highest code section offset for a deoptimization fallback compiled method
+     *         belonging to this class.
+     */
     public int hipcDeopt() {
         assert isPrimary();
         assert includesDeoptTarget();
-        return primaryEntries.get(primaryEntries.size() - 1).getPrimary().getHi();
+        return deoptPrimaryEntries.get(deoptPrimaryEntries.size() - 1).getPrimary().getHi();
     }
 }
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
@@ -72,22 +72,21 @@ public class ClassEntry extends StructureTypeEntry {
      */
     private Map<ResolvedJavaMethod, MethodEntry> methodsIndex;
     /**
-     * A list recording details of all normal primary compiled methods included in this class sorted
-     * by ascending address range. Note that the associated address ranges are disjoint and
-     * contiguous.
+     * A list recording details of all normal compiled methods included in this class sorted by
+     * ascending address range. Note that the associated address ranges are disjoint and contiguous.
      */
-    private List<PrimaryEntry> primaryEntries;
+    private List<CompiledMethodEntry> normalCompiledEntries;
     /**
-     * A list recording details of all deopt fallback primary compiled methods included in this
-     * class sorted by ascending address range. Note that the associated address ranges are
-     * disjoint, contiguous and above all ranges for normal primary compiled methods.
+     * A list recording details of all deopt fallback compiled methods included in this class sorted
+     * by ascending address range. Note that the associated address ranges are disjoint, contiguous
+     * and above all ranges for normal compiled methods.
      */
-    private List<PrimaryEntry> deoptPrimaryEntries;
+    private List<CompiledMethodEntry> deoptCompiledEntries;
     /**
-     * An index identifying ranges for primary compiled method which have already been encountered,
-     * whether normal or deopt fallback methods.
+     * An index identifying ranges for compiled method which have already been encountered, whether
+     * normal or deopt fallback methods.
      */
-    private Map<Range, PrimaryEntry> primaryIndex;
+    private Map<Range, CompiledMethodEntry> compiledMethodIndex;
     /**
      * An index of all primary and secondary files referenced from this class's compilation unit.
      */
@@ -111,10 +110,10 @@ public class ClassEntry extends StructureTypeEntry {
         this.fileEntry = fileEntry;
         this.methods = new ArrayList<>();
         this.methodsIndex = new HashMap<>();
-        this.primaryEntries = new ArrayList<>();
+        this.normalCompiledEntries = new ArrayList<>();
         // deopt methods list is created on demand
-        this.deoptPrimaryEntries = null;
-        this.primaryIndex = new HashMap<>();
+        this.deoptCompiledEntries = null;
+        this.compiledMethodIndex = new HashMap<>();
         this.localFiles = new ArrayList<>();
         this.localFilesIndex = new HashMap<>();
         this.localDirs = new ArrayList<>();
@@ -159,18 +158,18 @@ public class ClassEntry extends StructureTypeEntry {
     }
 
     public void indexPrimary(Range primary, List<DebugFrameSizeChange> frameSizeInfos, int frameSize) {
-        if (primaryIndex.get(primary) == null) {
-            PrimaryEntry primaryEntry = new PrimaryEntry(primary, frameSizeInfos, frameSize, this);
-            primaryIndex.put(primary, primaryEntry);
+        if (compiledMethodIndex.get(primary) == null) {
+            CompiledMethodEntry compiledEntry = new CompiledMethodEntry(primary, frameSizeInfos, frameSize, this);
+            compiledMethodIndex.put(primary, compiledEntry);
             if (primary.isDeoptTarget()) {
-                if (deoptPrimaryEntries == null) {
-                    deoptPrimaryEntries = new ArrayList<>();
+                if (deoptCompiledEntries == null) {
+                    deoptCompiledEntries = new ArrayList<>();
                 }
-                deoptPrimaryEntries.add(primaryEntry);
+                deoptCompiledEntries.add(compiledEntry);
             } else {
-                primaryEntries.add(primaryEntry);
+                normalCompiledEntries.add(compiledEntry);
                 /* deopt targets should all come after normal methods */
-                assert deoptPrimaryEntries == null;
+                assert deoptCompiledEntries == null;
             }
             FileEntry primaryFileEntry = primary.getFileEntry();
             if (primaryFileEntry != null) {
@@ -183,10 +182,10 @@ public class ClassEntry extends StructureTypeEntry {
         Range primary = subrange.getPrimary();
         /* The subrange should belong to a primary range. */
         assert primary != null;
-        PrimaryEntry primaryEntry = primaryIndex.get(primary);
+        CompiledMethodEntry compiledEntry = compiledMethodIndex.get(primary);
         /* We should already have seen the primary range. */
-        assert primaryEntry != null;
-        assert primaryEntry.getClassEntry() == this;
+        assert compiledEntry != null;
+        assert compiledEntry.getClassEntry() == this;
         FileEntry subFileEntry = subrange.getFileEntry();
         if (subFileEntry != null) {
             indexLocalFileEntry(subFileEntry);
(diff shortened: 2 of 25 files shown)


adinn:
@stooke It would be good if you could review this patch even though it is mostly incidental to the Windows code. There is one semantic change which I believe does not affect the Windows code but I'd very much like you to validate that assumption.

I have repurposed fields `primaryClasses` and `primaryClassesIndex` of class `DebugInfoBase`, renaming them to `instanceClasses` and `instanceClassesIndex`. The associated iterator has also changed named to `getInstanceClasses()`.

The associated collection instances were originally populated with `ClassEntry` records for all classes notified via the `DebugCodeInfo` API i.e. classes which had at least one top level compiled method. When inline info arrived the list and associated index was supplemented with any extra classes for which methods were found in inline frames.

That meant that some of these so-called primary classes did not actually have any top level compiled methods. Luckily code that processes entries in these collections appears to always check whether the entry has any compiled methods, by calling `ClassEntry::isPrimary()`, or not care and simply iterate over an empty list of `primaryEntry` records.

In consequence I have repurposed this list so that it is populated with all instance classes that get presented via the `DebugTypeInfo` API (i.e. excluding array types,primitive types and the object header struct type). This then allows the Linux code to generate debug info for all instance classes by calling `getInstanceClasses()`. It would be very helpful if you could check whether this is ok as far as the Windows code is concerned.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -359,6 +362,10 @@ private int writeNonPrimaryClassUnit(DebugContext context, ClassEntry classEntry
         String compilationDirectory = classEntry.getCachePath();
         log(context, "  [0x%08x]     comp_dir  0x%x (%s)", pos, debugStringIndex(compilationDirectory), compilationDirectory);
         pos = writeAttrStrp(compilationDirectory, buffer, pos);
+        /* Non-primary classes have no compiled methods so they have no low or high pc. */
+        /* However, we still generate a statement list in cae they include inlined methods */
```
zakkak, line 366:
```suggestion
        /* However, we still generate a statement list in case they include inlined methods */
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -1317,8 +1363,8 @@ private int writeMethodLocation(DebugContext context, PrimaryEntry primaryEntry,
         log(context, "  [0x%08x]     specification  0x%x (%s)", pos, methodSpecOffset, methodKey);
         pos = writeAttrRefAddr(methodSpecOffset, buffer, pos);
         HashMap<DebugLocalInfo, List<Range>> varRangeMap = primary.getVarRangeMap();
-        pos = writeMethodParameterLocations(context, varRangeMap, primary, 2, buffer, pos);
-        pos = writeMethodLocalLocations(context, varRangeMap, primary, 2, buffer, pos);
+        pos = writeMethodParameterLocations(context, null, varRangeMap, primary, 2, buffer, pos);
+        pos = writeMethodLocalLocations(context, null, varRangeMap, primary, 2, buffer, pos);
```
zakkak, line 1367: The class entry is known in `writeMethodLocation` caller `writeMethodLocations`. Why don't we pass it through?

zakkak (changes requested):
Thanks @adinn. It looks good to me other than some typos and a few comments I added for consideration.

SergejIsbrecht:
@adinn, regarding perfs behavior you could probably ask the maintainer Arnaldo (Red Hat). This is his handle on Twitter https://twitter.com/acmel?t=Dt5BWUsXmY27r7REYbSCuw&s=09

adinn:
@SergejIsbrecht I resolved the problem with display of hex addresses for some methods in the most recent push. The problem is that perf only starts translating sampled code addresses from some given code section (`.text`, `.plt`, `.init`) to fn/method names if the address lies in the range starting at the first exported symbol. It seems to assume that code addresses preceding the first exported symbol occupy dead space.

So, if you don't preserve method symbols (i.e. pass `-H:-DeleteLocalSymbols`) then the first symbol will likely be some way down the .text segment. With the current GraalVM this is about 4M from the start, where class VMLocator exposes a method as a C entry point. Samples for methods which follow this symbol get looked up and reported using the DWARF name. Samples for methods that precede it are reported as hex addresses.

The fix is to add a dummy code symbol at the start of the .text section when `DeleteLocalSymbols` has value `true`.

adinn:
@SergejIsbrecht Also, assuming you do retain local symbols, the latest version of this patch will provide a pretty reliable perf annotate output for compiled Java methods. You can run a bare call

    $ perf annotate

to see stats for the whole program. Alternatively, you can annotate a single method by passing its symbol name

    perf annotate PrintStream_format_9338ddec90bb0bd89998e4b56896abba6d7c4ded

Unfortunately, at present you have to provide the symbolic name for the method, not the DWARF name `java.io.PrintStream::format`. That long string is an SHA1 digest of the actual method signature. If you want to identify the symbolic names you can run `nm` on your generated binary image. I'd be interested to hear your feedback on the current state of play with this feature.

SergejIsbrecht:
@adinn , I am currently compiling for x64. I will attach some FlameGraphs, when everything is running with your branch.

SergejIsbrecht:
@adinn ,

first of all: it look way batter in comparison to before, but I still see some "unknown" addresses

```
     0,47%     0,00%  ool-6-worker-23  bench               [.] com.oracle.svm.core.posix.thread.PosixPlatformThreads::setNativeName (inlined)                                                ▒
     0,46%     0,00%  ool-6-worker-13  bench               [.] org.jctools.queues.MpscUnboundedArrayQueue::poll (inlined)                                                                    ▒
     0,46%     0,00%  ool-6-worker-13  bench               [.] org.jctools.queues.BaseMpscLinkedArrayQueue::poll (inlined)                                                                   ▒
     0,43%     0,00%  te-jmh-worker-1  anon                [.] 0x00007fe4cf00116f                                                                                                            ▒
     0,42%     0,00%  te-jmh-worker-1  anon                [.] 0x00007fe51a7011f7                                                                                                            ▒
     0,42%     0,00%  te-jmh-worker-1  anon                [.] 0x00007fe4cea011f7                                                                                                            ▒
     0,40%     0,00%  te-jmh-worker-1  anon                [.] 0x00007fe518b02727                                                                                                            ▒
     0,36%     0,05%  te-jmh-worker-1  [vdso]              [.] __vdso_clock_gettime                                                                                                          ▒
     0,36%     0,00%  te-jmh-worker-1  libc-2.31.so        [.] __clock_gettime_2 (inlined)                                                                                                   ▒
     0,28%     0,00%  te-jmh-worker-1  anon                [.] 0x00007fe518b4e477                                                                                                            ▒
     0,24%     0,24%  te-jmh-worker-1  [vdso]              [.] 0x00000000000006c8                                                                                                            ▒
     0,24%     0,00%  te-jmh-worker-1  [vdso]              [.] 0x00007ffde6ffe6c8       
```

I compiled a JMH benchmark to a native-image and ran `perf record -F 99 --call-graph dwarf -- ./bench`
native_image_jmh1
native_image_jmh2
busercontent.com/4427685/185644021-efc9b85c-eabd-4bb3-a58d-fb63f7bb9669.png)

(Flamegraphs were created with https://github.com/flamegraph-rs/flamegraph)

```
./flamegraph --deterministic --palette java -c "record -F 99 --call-graph dwarf" -- ./bench
```

adinn:
@SergejIsbrecht Thanks very much for doing that! It certainly looks very interesting.

> I still see some "unknown" addresses ...
>      0,43%     0,00%  te-jmh-worker-1  anon                [.] 0x00007fe4cf00116f                                                                                                            
>      0,42%     0,00%  te-jmh-worker-1  anon                [.] 0x00007fe51a7011f7                   

I do no know where those sample addresses come from but most definitely look like they are not from method code compiled into the `.text` section (nor even code provided by SVM in the `.init` section). `.text` usually starts around 0x400500 for a small program, maybe a bit more for a large one. `objdump -h <imagefile>` will tell you the start VMA and extent of section `.text` for your image. Those addresses looks like they have come out of a dynamically allocated region. They might be in a dynamic shared library, I suppose, but I would have expected perf to be able to provide an address for whatever shared lib code your program relies on.

> If you would like I could create some FlameGraphs with async-profiler with an actual JVM for comparisons.

Yes, that might be interesting to see how easy it is to compare what is going on under the actual benchmark run methods. Of course, your graph will include a slab of profile data for both JVM startup and benchmark warmup (including a load of execution bundled as 'interpreter') all of which will be missing in the Graal case.

> Also there are still some SHA1 hashes around
> -    8,83%     8,83%  ool-6-worker-13  bench               [.] IsolateEnterStub__VmLocatorSymbol__vmLocatorSymbol__bec84cad1f8708102cd8814ef3e496531bf6ff5b__b6484ec9f5c47b82c12ad1112cecc290
>      8,63% java.util.concurrent.ForkJoinPool::runWorker (inlined)                                                                                                                            
>        start_thread                                                                                                                                                                         
>        com.oracle.svm.core.code.IsolateEnterStub::PosixPlatformThreads_pthreadStartRoutine_38d96cbc1a188a6051c29be1299afe681d67942e (inlined)                                               
>        com.oracle.svm.core.posix.thread.PosixPlatformThreads::pthreadStartRoutine (inlined)                                                                                                 

That's not necessarily an indication that the name is being derived from a symbol rather than from the debuginfo. In this case the outermost pthread start method whose name includes an SHA1 hash is a synthetic stub method. Methods like this one only ever get presented to the debug info generator with a method name that includes an SHA1 hash (there is no 'real' Java name). In this case I am certain that the printed name has been derived from debug info because the class name is package qualified. If a symbol had been used to name the sample then there would be no package qualifier (the symbol naming algorithm used by GraalVM constructs symbols as BareClass_Method_sha1hash).

adinn:
@SergejIsbrecht Thinking about it those weird addresses may be the result of improper stack unwinding. That could indicate an issue with the generation of the `.debug_frame` section. I'll see if I can reproduce similar weirdness and then try to debug it.

SergejIsbrecht:
@adinn,

> ll see if I can reproduce similar weirdness and then try to debug it.

Don't bother. I will give you one later that day.

SergejIsbrecht:
@adinn ,

I am currently working on an example. Here are some FlameGraphs with Java11 Hotspot C2. FlameGraphs are captured with async-profiler 2.8 on Linux (5.4)

JMH Gradle Plugin:
```
profilers.set(listOf("async:libPath=/home/user/Applications/async-profiler-2.8-linux-x64/build/libasyncProfiler.so;output=flamegraph;cstack=fp;event=cpu;interval=5000000;threads=false"))
```

https://gist.github.com/SergejIsbrecht/dda8abca7c3c1004f03f1507bc1e9240

(Just download the HTML files and open them)

https://gist.github.com/SergejIsbrecht/5c33bce34e0b1036bc827f5762c5c545

This FlameGraph was captured with your branch and flamegraph-rs.

```
flamegraph --deterministic --palette java -c "record -F 99 --call-graph dwarf" -- -/bench
```

The Hotspot C2 FlameGraph kind of looks different. There are a lot of `consumeCPU` stackframes, which are not available in the native-image ones. Maybe they were just inlined? It would be strange when the native-image would inline this calls, but the Hotspot C2 would not.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
```diff
@@ -369,4 +370,48 @@ public MethodEntry ensureMethodEntryForDebugRangeInfo(DebugRangeInfo debugRangeI
     public List<MethodEntry> getMethods() {
         return methods;
     }
+
+    public int lowpc() {
+        assert isPrimary();
+        return primaryEntries.get(0).getPrimary().getLo();
+    }
+
+    public int lowpcDeopt() {
```
olpaw, line 379: If you instead make `boolean` field `com.oracle.objectfile.debugentry.ClassEntry#includesDeoptTarget` into an `int` that is `-1` if there are no deoptTarget entries, of the index of the first deoptTarget entry if there are any then you can also avoid the loop here and just access the first deoptTarget entry to get its `lo` value. Would that work?

olpaw (changes requested):
Thanks a lot for a adding perf-support!

See individual review comments. In addition:

- [ ] Please also test if this works with images that are shared libraries (i.e. built with `--shared`). 

- [ ] Please also add documentation on how to use perf with native-images.
Either in `docs/reference-manual/native-image/DebugInfo.md` of you create an new `docs/reference-manual/native-image/Profiling.md`. It should contain info how the image needs to be built so that it can be profiled with perf, what is currently working, current limitations, plans/ideas for future improvements.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
```diff
@@ -369,4 +370,48 @@ public MethodEntry ensureMethodEntryForDebugRangeInfo(DebugRangeInfo debugRangeI
...
+            assert lastPrimary != null;
+            return lastPrimary.getHi();
+        }
+    }
+
+    public int hipcDeopt() {
+        assert isPrimary();
+        assert includesDeoptTarget();
+        return primaryEntries.get(primaryEntries.size() - 1).getPrimary().getHi();
+    }
```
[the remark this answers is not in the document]
adinn: Yes, I will add a comment to this effect, although this is already discussed in DebugInfoBase where the choice of a class per compilation unit (as opposed to a file per CU) is explained. I think it is probably best to refer to that comment rather than restate it.

I don't think two lists is a good idea. Classes rarely have large numbers of methods (the exception being a few cases like the reflection holder class where methods are mostly synthetic). Most classes have very few methods and no deopt methods. So adding an extra list per class to avoid a small amount of iteration sounds to me like a bad tradeoff of space against time. I may be wrong but I'd prefer to have some stats to prove me wrong before making any such change.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfARangesSectionImpl.java
```diff
@@ -213,62 +223,86 @@ public void writeContent(DebugContext context) {
...
-                pos = writeLong(0, buffer, pos);
-                pos = writeLong(0, buffer, pos);
             }
+            pos = writeLong(0, buffer, pos);
+            pos = writeLong(0, buffer, pos);
         }
         assert pos == size;
     }
 
+    public static int compare(int pc1, int pc2) {
```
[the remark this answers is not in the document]
adinn: Sure, but as you say below this is not needed.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfARangesSectionImpl.java
```diff
@@ -213,62 +223,86 @@ public void writeContent(DebugContext context) {
...
+            return -1;
+        } else if (pc1 > pc2) {
+            return 1;
+        } else {
+            return 0;
+        }
+    }
+
+    private int sortByLowPC(ClassEntry classEntry1, ClassEntry classEntry2) {
+        return compare(classEntry1.lowpc(), classEntry2.lowpc());
```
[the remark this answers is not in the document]
adinn: Yes, that's better.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfARangesSectionImpl.java
```diff
@@ -213,62 +223,86 @@ public void writeContent(DebugContext context) {
...
+    }
+
+    private int sortByLowPCDeopt(ClassEntry classEntry1, ClassEntry classEntry2) {
+        return compare(classEntry1.lowpcDeopt(), classEntry2.lowpcDeopt());
```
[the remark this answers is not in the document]
adinn: Yes.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfDebugInfo.java
```diff
@@ -73,7 +73,7 @@ public class DwarfDebugInfo extends DebugInfoBase {
     public static final int DW_ABBREV_CODE_builtin_unit = 1;
     public static final int DW_ABBREV_CODE_class_unit1 = 2;
     public static final int DW_ABBREV_CODE_class_unit2 = 3;
-    public static final int DW_ABBREV_CODE_class_unit3 = 4;
+    // public static final int DW_ABBREV_CODE_class_unit3 = 4;
```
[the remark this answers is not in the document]
adinn: I will actually remove this definition and renumber to avoid the gap in the CODE range.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -1713,6 +1732,14 @@ public int writeIndirectOopConversionExpression(boolean isHub, byte[] buffer, in
         return pos;
     }
 
+    /*
+     * Attributes of type ref4 which refer to an element in she same CU need to be written as an
+     * offset relative to the start of the current CU. This field is updated every time writing of a
+     * CU commences, providing as a section offset for the CU whcih can be subtracted from the
+     * section offset that are used to record element positions.
```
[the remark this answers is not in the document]
adinn: Done.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -902,7 +925,9 @@ protected abstract class NativeImageDebugBaseMethodInfo extends NativeImageDebug
             // method references in node source positions. So, we do the translation
             // here just to make sure we use a HostedMethod wherever possible.
             method = promoteAnalysisToHosted(m);
-            this.paramInfo = createParamInfo(method);
+            LineNumberTable lineNumberTable = method.getLineNumberTable();
+            line = (lineNumberTable != null ? lineNumberTable.getLineNumber(0) : 0);
```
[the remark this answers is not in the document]
adinn: Because when written out in the DWARF info 0 is an illegal line number and -1 is interpreted as the unsigned value 0xffffffff -- n.b. line numbers have to be strictly positive to be valid.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -351,7 +352,11 @@ private abstract class NativeImageDebugFileInfo implements DebugFileInfo {
             Class<?> clazz = hostedType.getJavaClass();
             SourceManager sourceManager = ImageSingletons.lookup(SourceManager.class);
             try (DebugContext.Scope s = debugContext.scope("DebugFileInfo", hostedType)) {
-                fullFilePath = sourceManager.findAndCacheSource(javaType, clazz, debugContext);
+                Path filePath = sourceManager.findAndCacheSource(javaType, clazz, debugContext);
+                if (filePath == null && hostedType instanceof HostedInstanceClass) {
+                    filePath = fullFilePathFromClassName((HostedInstanceClass) hostedType);
```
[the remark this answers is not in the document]
adinn: Yes, we would like all files to be in the cache. However, we only get here when `sourceManager.findAndCacheSource()` returns `null` which means we were not able to cache the file.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
adinn: If you look carefully you will notice that this `writeMethodParameterLocations` and `writeMethodLocalLocations` have two callers: `writeMethodDeclaration` and `writeAbstractInlineMethod`. The first caller passes `null` for the class entry because we are writing the params/locals for the original definition of the method. The second caller passes the class entry for the owner of the abstract inline method because we are writing a repeat copy of the locals/params.

In the first case the class of the (original) method is the class whose CU is currently being written. In the second case the class of the (abstract inline) method may not be the same as the class of the CU currently being written.

Passing null in the first case vs the abstract inline class entry is important. The classEntry argument is passed on to `writeMethodParameter/LocalDeclaration` which uses it to parameterize the call to `setMethodLocalIndex`. This setter method indexes values associated with a null class entry differently to those indexed with a non-null class entry because they represent different types of declaration. A value indexed with a null entry identifies a param/local declared under an original declaration. A value indexed with a non-null entry identifies a param/local declared under an abstract inline declaration

If you look at the points where calls to `getMethodLocalIndex` occur you will see that they use the same argument scheme. When defining the location of a param/local for a top level method `null` is passed. When defining the location of a param/local for a (concrete) inlined method the inlined method owner's class entry is passed.

The former step is unchanged from before this patch. The latter step is what ensures that an inline method's param/local location DIEs refer to a param/local specification that exists in the same CU.

adinn:
@SergejIsbrecht
> The Hotspot C2 FlameGraph kind of looks different. There are a lot of consumeCPU stackframes, which are not available in the native-image ones. Maybe they were just inlined? It would be strange when the native-image would inline this calls, but the Hotspot C2 would not.

Thanks for providing the flame graphs -- they are very interesting. I am not sure why you are seeing frames omitted in the native graph but I have a suspicion what is going on. Before mving on to that I'll note that both Hotspot and GraalVM should be able to allocate sample addresses to inlined methods rather than simply to the top level compiled method. However, this will rarely be 100% accurate. Because of the way code gets reorganized during optimization machine code that derives from one method often gets accounted for as belonging to another method. That said, I don't think this explains the disparity seen in your example.

So, here's my guess as to what is happenign. When you built your image did you pass option `-H:+SourceLevelDebug`? When you omit this option then very little frame information is passed through from the compiler to the debugger. That means that, apart from obvious places like call points and exception handler entries, where a top level and inline frame stack has to be well-defined,  and maybe one or two other places where frame info is available, the debug generator can only assume that code addresses belong to the top level method. That might account for why all of the samples that should belong to `consumeCPU` are being allocated to the `Lambda...run` method.

n.b. the reason this argument is not the default is that including extra frame info in the compiler graph may cause the register allocator to think that values in registers are alive for slightly longer than would otherwise be assumed. That's not necessarily going to be a problem for all compiled methods but it can easily cause a noticeable drop in performance for benchmarks, especially micro-benchmarks. So, be aware that if you do switch this flag on you might see your benchmark run a bit slower.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
```diff
@@ -369,4 +370,48 @@ public MethodEntry ensureMethodEntryForDebugRangeInfo(DebugRangeInfo debugRangeI
...
+        assert lowpc >= 0;
+        return lowpc;
+    }
+
+    public int hipc() {
+        assert isPrimary();
+        if (!includesDeoptTarget()) {
+            return primaryEntries.get(primaryEntries.size() - 1).getPrimary().getHi();
+        } else {
+            Range lastPrimary = null;
```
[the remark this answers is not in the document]
adinn: Afraid not. The name PrimaryEntry is rather misleading. What that type actually represents is a top level compiled method. The list we are iterating is a list of all top level compiled methods in a given class. Of course the list may be empty. If not then the list is sorted by method code range within the code section. Any two methods lie in disjoint ranges. The non-deopt methods (potentially plural) form an initial segment of the list and all their ranges are contiguous. Deopt methods (potentially plural) only exist when there are non-deopt equivalents. Deopt methods form a trailing segment of the list and all their ranges are contiguous. If both leading and trailing segment are non-empty then there will be a discontinuity between the ranges for these two segments. That's because the totality of non-deopt methods for all classes are generated before the totality of deopt methods for all classes.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
```diff
@@ -369,4 +370,48 @@ public MethodEntry ensureMethodEntryForDebugRangeInfo(DebugRangeInfo debugRangeI
...
+            }
+            assert lastPrimary != null;
+            return lastPrimary.getHi();
+        }
+    }
+
+    public int hipcDeopt() {
+        assert isPrimary();
+        assert includesDeoptTarget();
+        return primaryEntries.get(primaryEntries.size() - 1).getPrimary().getHi();
```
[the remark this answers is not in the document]
adinn: Yes, but I'm also leaving this assert in

         assert includesDeoptTarget(); 

We ought to have reliably detected that this class entry contains deopt targets.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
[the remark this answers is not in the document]
olpaw: It might be more convenient to map the two segments to two ArrayLists instead, then these methods would become trivial.
In places where you need them to be seen as one you just use `java.util.stream.Stream#concat`

SergejIsbrecht:
@adinn ,

it does not seem to matter whether I use `-H:+SourceLevelDebug` or `-H:-SourceLevelDebug` when generating the image. When I run `perf` with ` ~/D/g/g/e/p/b/n/nativeCompile  perf record -F 499 --call-graph dwarf -- ./bench` I get following `perf report`:

```
perf 279819 66305.036635:       1792 cycles: 
        ffffffffafc73616 native_write_msr+0x6 ([kernel.kallsyms])
        ffffffffafc0d335 intel_tfa_pmu_enable_all+0x35 ([kernel.kallsyms])
        ffffffffafc07addr2line: DWARF error: section .debug_info is larger than its filesize! (0xc14d2 vs 0x90ed8)
addr2line: DWARF error: section .debug_info is larger than its filesize! (0xc14d2 vs 0x90ed8)
addr2line: DWARF error: section .debug_info is larger than its filesize! (0xc14d2 vs 0x90ed8)
addr2line: DWARF error: section .debug_info is larger than its filesize! (0xc14d2 vs 0x90ed8)
addr2line: DWARF error: section .debug_info is larger than its filesize! (0xc14d2 vs 0x90ed8)
```

While importing following issue is displayed:
Screenshot from 2022-08-19 16-52-13

The hex address seems quite unusual. 

```
            55e712eb128a com.oracle.svm.core.posix.thread.PosixPlatformThreads::pthreadStartRoutine+0x6408a (inlined)
            55e712e4d05c com.oracle.svm.core.code.IsolateEnterStub::PosixPlatformThreads_pthreadStartRoutine_38d96cbc1a188a6051c29be1299afe681d67942e+0x17c (inlined)
            7efd88131608 start_thread+0xd8 (/usr/lib/x86_64-linux-gnu/libpthread-2.31.so)
                       7 [unknown] ([unknown])
```

Later today I will run some native rust / c++ application with perf in order to see whether I see such unusual addresses.

adinn:
> It does not seem to matter whether I use -H:+SourceLevelDebug or -H:-SourceLevelDebug when generating the image. At least the generated FlameGraphs looked the same. I did not see any consume stack-frames.

Ok, thanks very much for testing it.

> When I run perf with perf record -F 499 --call-graph dwarf -- ./bench I get following perf report:

>    perf 279819 66305.036635:       1792 cycles: 
>            ffffffffafc73616 native_write_msr+0x6 ([kernel.kallsyms])
>            ffffffffafc0d335 intel_tfa_pmu_enable_all+0x35 ([kernel.kallsyms])
>            ffffffffafc07addr2line: DWARF error: section .debug_info is larger than its filesize! (0xc14d2 vs 0x90ed8)
>    addr2line: DWARF error: section .debug_info is larger than its filesize! (0xc14d2 vs 0x90ed8)
>    addr2line: DWARF error: section .debug_info is larger than its filesize! (0xc14d2 vs 0x90ed8)
>    ...

Well it looks like the debuginfo is incorrect. Can you provide me with the benchmark source and detailed instructions on how to generate the binary? I'd really like to have a look at the debuginfo it contains. I'd also like to run addr2line under gdb to see what is causing it to print those errors.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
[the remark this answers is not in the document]
adinn: As I explained to Foivos I was wanting to avoid having an extra field and list in each class entry for the normal case where there are no deopt compiled methods. However, I guess I can create the deopt primaries list on demand and use a `null` list when there are no deopt methods. That also allows me to use the value of the list field as a way to get rid of boolean field `includesDeoptTarget`.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
adinn: That is a nice idea but it will be rendered unnecessary when I introduce a separate list for deopt methods.

SergejIsbrecht:
@adinn, sure. I will provide a working example in some hours. Anyways, this jmh bench is quite trivial, therefor any code should suffer from the same issue.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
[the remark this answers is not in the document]
zakkak: True, but couldn't we cache the file in the first place if we used `fullFilePathFromClassName` inside `findAndCacheSource` as a _new_ "last resort"?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -341,13 +343,14 @@ private int writeNonPrimaryClasses(DebugContext context, byte[] buffer, int pos)
 
     private int writeNonPrimaryClassUnit(DebugContext context, ClassEntry classEntry, byte[] buffer, int p) {
         int pos = p;
-        setCUIndex(classEntry, pos);
+        int lineIndex = getLineIndex(classEntry);
+        cuStart = pos;
+        setCUIndex(classEntry, cuStart);
```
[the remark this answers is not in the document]
zakkak: OK, not strong opinion here.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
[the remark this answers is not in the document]
adinn: `fullFilePathFromClassName` returns a *relative* path and file name based on the package qualified class name. This is done merely to conjure up a file name unique to the class that we can write into the debuginfo in order that perf and other tools don't give up early on trying to process samples.

We already know that this relative path cannot be extended to locate an actual source file. Any absolute path that works will be rooted in one of the possible locations where we might find an actual source: either the JDK src.zip root or a source jar root that corresponds to a class jar on the modulepath or classpath. The source cache manager will already have tried all available options to identify an absolute path to a source file using this relative path as a tail. So, we have nothing left to try here to find an actual source file.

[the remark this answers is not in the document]
adinn: I have added a comment to that effect at the point of call.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
```diff
@@ -71,12 +72,20 @@ public class ClassEntry extends StructureTypeEntry {
      */
     private Map<ResolvedJavaMethod, MethodEntry> methodsIndex;
     /**
-     * A list recording details of all primary ranges included in this class sorted by ascending
-     * address range.
+     * A list recording details of all normal primary compiled methods included in this class sorted
+     * by ascending address range. Note that the associated address ranges are disjoint and
+     * contiguous.
      */
     private List<PrimaryEntry> primaryEntries;
```
zakkak, line 79: Minor: I believe renaming this to `normalPrimaryEntries` will make more clear what this list is about (or at least hint the readed to check the javadoc :) ).

zakkak (approved):
Thanks for addressing all the comments Andrew. It looks good to me!

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfARangesSectionImpl.java
```diff
@@ -104,34 +105,31 @@ public void createContent() {
...
-                    for (PrimaryEntry classPrimaryEntry : classPrimaryEntries) {
-                        if (!classPrimaryEntry.getPrimary().isDeoptTarget()) {
-                            pos += 2 * 8;
-                        }
-                    }
-                } else {
-                    pos += classPrimaryEntries.size() * 2 * 8;
-                }
+                // count 16 bytes for each normal compiled method
+                pos += classEntry.normalPrimaryEntries().reduce(0,
```
olpaw, line 109: Use
```
pos += classEntry.normalPrimaryEntries().count() * 2 * 8;
```
instead of this convoluted reduce.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfARangesSectionImpl.java
```diff
@@ -104,34 +105,31 @@ public void createContent() {
...
                  */
                 pos += DW_AR_HEADER_PAD_SIZE;
-                List<PrimaryEntry> classPrimaryEntries = classEntry.getPrimaryEntries();
-                for (PrimaryEntry classPrimaryEntry : classPrimaryEntries) {
-                    if (classPrimaryEntry.getPrimary().isDeoptTarget()) {
-                        pos += 2 * 8;
-                    }
-                }
+                // count 16 bytes for each non-deopt compiled method
+                pos += classEntry.deoptPrimaryEntries().reduce(0,
```
olpaw, line 127: The comment and the code do not match

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfARangesSectionImpl.java
```diff
@@ -231,76 +193,60 @@ public void writeContent(DebugContext context) {
...
+         * Align to 2 * address size.
+         */
+        for (int i = 0; i < DW_AR_HEADER_PAD_SIZE; i++) {
+            pos = writeByte((byte) 0, buffer, pos);
         }
+        return pos;
+    }
+
+    int writeARanges(DebugContext context, Stream<PrimaryEntry> entries, byte[] buffer, int p) {
+        return entries.reduce(p,
```
olpaw, line 232: Using operations with side effects with stream methods like `reduce` is not recommend. It's better to use `forEach` (or even `forEachOrdered`) instead. Final `pos` is just `p + buffer_length_delta` right?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfFrameSectionImpl.java
```diff
@@ -137,36 +137,35 @@ private int writeCIE(byte[] buffer, int p) {
...
-                    long hi = range.getHi();
-                    int lengthPos = pos;
-                    pos = writeFDEHeader((int) lo, (int) hi, buffer, pos);
-                    pos = writeFDEs(primaryEntry.getFrameSize(), primaryEntry.getFrameSizeInfos(), buffer, pos);
-                    pos = writePaddingNops(buffer, pos);
-                    patchLength(lengthPos, buffer, pos);
-                }
+            if (classEntry.isPrimary()) {
+                pos = classEntry.normalPrimaryEntries().reduce(pos,
+                                (p1, entry) -> writeMethodFrame(entry, buffer, p1),
```
olpaw, line 159: See https://github.com/oracle/graal/pull/4811/files/e7072b49874ca1db66157b335d91caab31d28b9b..2e14ef6f1cf5017ae5a4470980e53193e37842a4#r953587450

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfFrameSectionImpl.java
```diff
@@ -137,36 +137,35 @@ private int writeCIE(byte[] buffer, int p) {
...
-                    long lo = range.getLo();
-                    long hi = range.getHi();
-                    int lengthPos = pos;
-                    pos = writeFDEHeader((int) lo, (int) hi, buffer, pos);
-                    pos = writeFDEs(primaryEntry.getFrameSize(), primaryEntry.getFrameSizeInfos(), buffer, pos);
-                    pos = writePaddingNops(buffer, pos);
-                    patchLength(lengthPos, buffer, pos);
-                }
+            if (classEntry.includesDeoptTarget()) {
+                pos = classEntry.deoptPrimaryEntries().reduce(pos,
```
olpaw, line 166: see https://github.com/oracle/graal/pull/4811/files/e7072b49874ca1db66157b335d91caab31d28b9b..2e14ef6f1cf5017ae5a4470980e53193e37842a4#r953587450

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -974,18 +977,13 @@ private int writeInterfaceType(DebugContext context, InterfaceClassEntry interfa
...
 
-        for (PrimaryEntry primaryEntry : classPrimaryEntries) {
-            Range primary = primaryEntry.getPrimary();
-            if (primary.isDeoptTarget() != deoptTargets) {
-                continue;
-            }
-            pos = writeMethodLocation(context, primaryEntry, buffer, pos);
-        }
+        Stream<PrimaryEntry> entries = (deoptTargets ? classEntry.deoptPrimaryEntries() : classEntry.normalPrimaryEntries());
+        pos = entries.reduce(pos,
```
olpaw, line 984: see https://github.com/oracle/graal/pull/4811/files/e7072b49874ca1db66157b335d91caab31d28b9b..2e14ef6f1cf5017ae5a4470980e53193e37842a4#r953587450

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfLocSectionImpl.java
```diff
@@ -155,13 +156,11 @@ private int writeDeoptClassLocations(DebugContext context, byte[] buffer, int po
...
-        List<PrimaryEntry> classPrimaryEntries = classEntry.getPrimaryEntries();
-
-        for (PrimaryEntry primaryEntry : classPrimaryEntries) {
-            if (primaryEntry.getPrimary().isDeoptTarget() != isDeopt) {
-                continue;
-            }
-            pos = writePrimaryRangeLocations(context, primaryEntry, buffer, pos);
+        if (!isDeopt || classEntry.includesDeoptTarget()) {
+            Stream<PrimaryEntry> entries = (isDeopt ? classEntry.deoptPrimaryEntries() : classEntry.normalPrimaryEntries());
+            pos = entries.reduce(pos,
```
olpaw, line 161: see https://github.com/oracle/graal/pull/4811/files/e7072b49874ca1db66157b335d91caab31d28b9b..2e14ef6f1cf5017ae5a4470980e53193e37842a4#r953587450

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
adinn: I have a follow up patch which includes that change. The follow-up also renames class `PrimaryEntry` to `CompiledMethodEntry`, adjusts parameters and locals to use the name `compiledEntry` or instead of `primaryEntry` and updates comments to use the phrase `compiled method` instead of `primary method`.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfARangesSectionImpl.java
adinn: Ah, there's a `Stream::count()` method?

One day I'll give up on maintaining Byteman (target 6 bytecode) and actually start learning & using jdk7+ features. :-]

adinn: Done.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfARangesSectionImpl.java
adinn: It's not just about knowing what the final position is. This reduction operation is essentially using a stream to implement an indexed for loop with index var `p1` and loop step computed by the write calls in the loop body. The index is used as a cursor into byte array `buffer` to identify where to start writing the range info for each successive method. The cursor is consumed and updated by the two write operations. The final lambda is used to pass the updated cursor value on to defined `p1` for the next execution of the loop body. The final cursor value pops out at the end /defining/ the final position.

Now, in this case each iteration appends the same fixed amount to the buffer. So, you are right to observe that we can compute in advance what the final cursor position would be. However, I don't think that means we can simply switch to a `forEach`. We still need to inject a cursor start position into each invocation of the lambda in order to write each successive method's range data at the appropriate offset.

A better alternative, one that would be usable with stream `forEach`, would be to wrap the byte array in an object which also embeds a cursor updated by each write and pass that to the lambda as a final object. We could define our own wrapper or rely on an existing type like ByteBuffer or ByteArrayOutputStream. That would avoid the need for a reduction.

However, it would be much better if we used a common type across the code base, one which parameterized all the `writeXXX` methods. This type should also really provide a base API that is DWARF aware, implementing the various 'putXXX' methods that the `writeXXX` methods depend on (since these are the ones that would actually need to update the cursor). That would work fine with a `forEach` model so long as all uses were sequential.

This really should have been done from the outset. Unfortunately when I started it was not clear what was needed so I simply started/stuck with peeking and poking bytes into a byte array using an externally maintained offset. I was relying on lists and iterators at that point, not streams, so using a byte buffer and explicit buffer index was not such an issue. It certainly would simplify and clarify the code if we switched to using some form of buffer/stream which maintained its own cursor position. However, I think it would be better to do this as a global clean-up rather than as a local or global change mixed in with this patch.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
```diff
@@ -419,19 +423,24 @@ private Range addSubrange(DebugLocationInfo locationInfo, Range primaryRange, Cl
     }
 
     private ClassEntry ensureClassEntry(ResolvedJavaType type) {
-        /* See if we already have an entry. */
-        ClassEntry classEntry = primaryClassesIndex.get(type);
+        /* We should already have an entry -- TODO prove that claim. */
+        ClassEntry classEntry = instanceClassesIndex.get(type);
         if (classEntry == null) {
+            /* We must have a type entry -- TODO prove we never reach here. */
```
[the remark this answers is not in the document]
adinn: I am confident we can now reduce this call to a lookup, with an assert that the result is found, and dispense with any need for proof. So, I'll push an extra change to that effect.

Why so? Well, the continued existence of this method is an artefact of creeping change in the meaning of the fields now renamed to `instancesClasses` and `instanceClassesIndex`. Originally these fields tracked instance classes with top level compiled methods. The methods and associated classes were presented via the DebugCodeInfo API and ultimately derived by iterating over the `NativeImageCodeCache`. The expectation was that all of these classes would already have been presented via the DebugTypeInfo API, ultimately derived by iterating over the Universe type. However, the list was originally only meant to include the proper (i.e. strict) subset of the notified instance classes which needed to be qualified with method location and range info.

When we started reporting frame data via the `DebugCodeInfo` API we found classes that were i) not already identified via the `DebugTypeInfo` API or ii) had inlined methods but no top level methods. These also required generation of inline method location and range info. So, we threw them into the list and index, creating a `ClassEntry` if none existed and then sorted them out case by case in code that used the list of classes.

The first problem was an input hygiene problem. We ran into 'missing' types because frames were actually referencing analysis types instead of hosted types. That has been fixed on the notification side (`NativeImageDebugInfoProvider`) by remapping classes from analysis to hosted domain before passing them through the `DebugCodeInfo` API. 

The second problem reflect the fact that we really need to iterate over all instance classes at certain points in the code and only over some instance classes at others. A `ClassEntry` now identifies whether or not it has top level compiled methods and/or inlined methods and all code that now consumes the `instanceClasses` list tests for the relevant cases.

So, both the problems we originally faced appear to be resolved. The current code populates this list and index with all expected instance classes during traversal of universe types set. So, the checks and updtes that used to be done in `ensureClassEntry` really ought to be redundant.

n.b. There is another follow-up task needed to consolidate the current position beyond redefining `ensureClassEntry`. From one point of view fields `instanceClasses` and `instanceClassesIndex` are redundant. We now only consume the instaceClasses list via a Stream API. The stream content manifested as a list in `instanceClasses` can be derived virtually by filtering field `types` with a type test for `ClassEntry` and cast before returning. Similarly, lookups in the index map `instanceClassesIndex` can be replaced with a lookup in the index map `typesIndex` that also employs an `instanceof` check. So, we have the option to virtualize the instance class list and index, trading off space against a small amount of stream processing time.

Whichever way we decide to go with this a code change is needed. There are a couple of cases left where iteration over all instance classes was and still is performed by streaming and filtering field `types` (this made sense when not all instance classes were listed in field `instanceClasses`).  I'm happy to fix this as part of this patch or in a follow-up. Do you have a preference for retaining two lists or virtualizing one of them?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfARangesSectionImpl.java
adinn: Ok, I had a rethink and came up with a relatively simple solution that 1) removes the reductions and 2) can either be extended later to deal with all cases or can be superseded by use of a ByteBuffer/ByteArrayOutputStream class.

I defined a class called `Cursor` nested in `DwarfSectionImpl` that wraps an `int` field called `pos`, with the obvious `get`, `set` and `add` methods. This allows it to be used in all the `DwarfXXXSection` classes when they are writing data to `byte[] content`.

A `Cursor` can be created and set to the current `pos` (or default to 0)  at any point where we want to iterate over a stream of class entries, compiled method entries etc. The cursor can be passed into the `forEach` lambda bodies and `get`, `set` or `add` can be used to read and/or update the current position.

This allows all existing reduce operations to be simplified to a `forEach` -- so long as it executes sequentially. We could extend this so that all `writeXXX` and `putXXX` operations take a `Cursor pos` argument in place of the current `int pos` argument. Alternatively, we can switch to using some other type to encapsulate both the `byte[] buffer` and `int pos`.

I'll push a patch soon.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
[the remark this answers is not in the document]
adinn: Fixed. I retained the separate list of instance classes but everything is now using streams derived either from the types list or from the classes list.

adinn:
@SergejIsbrecht Are you sure that your JMH benchmark is not the cause of the disparity you are seeing in the flamegraphs?

I ask that because when JMH runs on the JVM it relies on specific details of the compiler implementation to ensure that the body of a benchmark routine is not optimized away to nothing. Your flamegraph for GraalVM Native shows only the test thread run method. That looks very much like a JMH benchmark where the body of the benchmark routine has been optimized away to nothing.

Do you know for certain that the assumptions made by the version of JMH you are using about what does and does not get optimized will be met by the Graal compiler?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ClassEntry.java
adinn: This renaming patch has now been pushed

olpaw:
@adinn to get this into 22.3 this has to be merged by the end of this week. AFAICS documentation is still missing and a bit of commit message cleanups/fixups (e.g. consistent capitalization of first word of message, folding style only changes in previous commits where possible).

adinn:
@olpaw I just pushed a draft change to the docs that

- mentions the ability to use perf and valgrind in the overview
- describes how to retain local symbols to improve output quality
- provides full explanation of the various wrinkles ot using these tools in a dedicated subsection

All the other rework has been addressed and the cleanup pass to rename PrimaryEntry to CompiledMethodEntry plus various other related name changes has been performed.

So, I think all that is left is to squash the style fix commits and for you to give it a final review.

adinn:
@olpaw By the way, I am not seeing any style check errors when I run a check on my machine. However, the gate often picks up things that do not show locally. Can you point me at the things that need fixing?

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -35,8 +37,22 @@ However, it can significantly increase the size of the generated image on disk.
 and local variable information by passing flag `-H:+SourceLevelDebug` can cause a program to be compiled
 slightly differently and for some applications this can slow down execution.
 
+The basic `perf report` command, which displays a histogram showing percentage execution time in each Java method , only requires passing flags `-g` and `-H:+SourceLevelDebug` to the `native-image` command.
```
olpaw, line 40: s/method ,/method,

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -772,4 +789,158 @@ For example, calling the method below prints high-level information about the Na
...
+```
+
+The leftmost column shows percentages for the amount of time recorded
+at each instruction in samples obtained during the `perf record` run.
+Each instruction is prefaced with it's address in the program's code
+section.
+The disassembly interleaves the source lines from which the code is
+derived, 521-524 for the top level code and 207-209 for the code
+inlined from from `Objects.requireNonNull()`.
+(note, however that the lines for the quoted comments are being computed
```
olpaw, line 869: s/lines/assembly lines

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -772,4 +789,158 @@ For example, calling the method below prints high-level information about the Na
...
+         : 212    if (obj == null)
+    0.00 :   51aae8: nop
+    0.00 :   51aae9: nop
+         : 209    throw new NullPointerException();
+         : 210    return obj;
+    . . .
+```
+
+Note that the start address of the method is now labelled with the
+mangled symbol name `String_constructor_f60263d569497f1facccd5467ef60532e990f75d` as well as the DWARF name.
```
olpaw, line 921: Would be cool if `c++filt` would support our mangling.

docs/reference-manual/native-image/DebugInfo.md
adinn: Done.

docs/reference-manual/native-image/DebugInfo.md
adinn: Actually, the assembly lines are always going to be a bit out wrt to the source but but that is not what I meant. This comment is in the wrong place, which explains why you thought I was saying something different.

It is actually the line *numbers* in the trailing javadoc for the method in the *second* sample that are wrong. You can see that in this sample the first line of the method is correctly identified as being at 521. In the second sample that line is output preceded by the 5 trailing lines of the javadoc. For some reason perf has started numbering the trailing lines from 521 rather than from 516. This looks like a perf issue rather than a DWARF one.

I'll relocate the comment to after the second sample and adjust it to say that the line numbering has been applied incorrectly.

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -772,4 +789,158 @@ For example, calling the method below prints high-level information about the Na
...
+section.
+The disassembly interleaves the source lines from which the code is
+derived, 521-524 for the top level code and 207-209 for the code
+inlined from from `Objects.requireNonNull()`.
+(note, however that the lines for the quoted comments are being computed
+and displayed incorrectly).
+Also, the start of the method is labelled with the name defined in the
+DWARF debug info, `java.lang.String::String()`.
+However, the branch instruction `jbe` at address `0x519d6f` uses a
+very large offset from `graal_vm_locator_symbol` to identify what is
```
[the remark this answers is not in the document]
adinn: The offset from `graal_vm_locator_symbol` is correct. It's just that it's not very helpful. Do you want me to clarify that.

docs/reference-manual/native-image/DebugInfo.md
adinn: Yes it would. I have had a quick lopk into what perf does to unmangle names. By default it calls `bfd_demangle` which has its own slightly weird expectations about how to handle names e.g. arrays of `TYPE` are modelled as `JArray<TYPE>`. If (when) that fails it should punt to a routine in the perf code called `java_demangle_sym`. This knows how to demangle bytecode type and member signatures into Java syntax. For example it will turn `Ljava/lang/StringLatin1;equals([B[B)Z` into `boolean java.lang.StringLatin1.equals(byte[], byte[])"`. We might be able to use that for some cases but I guess it won't do for methods exposed as C entry points which actually need a linker-friendly name.

Note that anyway I am not currently using the symbol name as the `linkage_name` atttribute in the method.definition. At present I generate that attribute as  `<fully_qualified_class_name>::<bare_method_name>` in order to ensure that 1) `perf report` prints something sensible when there are no local symbols 2) `perf annotate` includes this name at the start of the listing for a given method. We can probably do better than this (but maybe in a follow-up patch).

olpaw: > but maybe in a follow-up patch

Definitely

docs/reference-manual/native-image/DebugInfo.md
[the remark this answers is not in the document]
olpaw: Yes that would help prevent confusion. So the address `graal_vm_locator_symbol+0xe26ba` points to is correct, BUT since its anchored to `graal_vm_locator_symbol` instead of the method itself it's quite cryptic fcrom the users pov.

olpaw:
> > Can you point me at the things that need fixing?
>
> I will run the PR in our CI to see what comes up.

`mx --kill-with-sigquit --strict-compliance debuginfotest --output-path /b/b/e/main/substratevm/svmbuild` failed:

```
(gdb) backtrace
Checker backtrace hello.Hello::main: match 3 failed at line 3 #3  com.oracle.svm.core.JavaMainWrapper::doRun (argc=<optimized out>, argv=<optimized out>) at com/oracle/svm/core/JavaMainWrapper.java:232

Checker backtrace hello.Hello::main {
  re.compile('#0[ \t]+hello\\.Hello::main .* at hello/Hello\\.java:76')
  re.compile('#1[ \t]+0x[0-9a-f]+ in com\\.oracle\\.svm\\.core\\.JavaMainWrapper::runCore0 .* at [a-z/]+JavaMainWrapper\\.java:[0-9]+')
  re.compile('#2[ \t]+0x[0-9a-f]+ in com\\.oracle\\.svm\\.core\\.JavaMainWrapper::runCore .* at [a-z/]+JavaMainWrapper\\.java:[0-9]+')
  re.compile('#3[ \t]+ com\\.oracle\\.svm\\.core\\.JavaMainWrapper::run .* at [a-z/]+JavaMainWrapper\\.java:[0-9]+')
  re.compile('#4[ \t]+0x[0-9a-f]+ in com\\.oracle\\.svm\\.core\\.code\\.IsolateEnterStub::JavaMainWrapper_run_[0-9a-f]+ .*')
}

#0  hello.Hello::main (args=0x7fa367e010f0) at hello/Hello.java:76
#1  0x000000000041447a in com.oracle.svm.core.JavaMainWrapper::runCore0 () at com/oracle/svm/core/JavaMainWrapper.java:175
#2  0x000000000041413d in com.oracle.svm.core.JavaMainWrapper::runCore () at com/oracle/svm/core/JavaMainWrapper.java:135
#3  com.oracle.svm.core.JavaMainWrapper::doRun (argc=<optimized out>, argv=<optimized out>) at com/oracle/svm/core/JavaMainWrapper.java:232
#4  0x0000000000442e1e in com.oracle.svm.core.JavaMainWrapper::run (argc=<optimized out>, argv=<optimized out>) at com/oracle/svm/core/JavaMainWrapper.java:218
#5  com.oracle.svm.core.code.IsolateEnterStub::JavaMainWrapper_run_5087f5482cc9a6abc971913ece43acb471d2631b (__0=<optimized out>, __1=<optimized out>) at com/oracle/svm/core/code/IsolateEnterStub.java:1

WARNING: psutil is not available, java process detection is less accurate
  File "/b/b/e/mxtool/mx_enter.py", line 41, in <module>
    mx._main_wrapper()
  File "/b/b/e/mxtool/mx.py", line 18093, in _main_wrapper
    main()
gate: 07 Sep 2022 13:32:35(+18:53) END:   image debuginfotest [0:01:46.338883] [disk (free/total): 362.5GB/511.3GB]
```

adinn:
> mx --kill-with-sigquit --strict-compliance debuginfotest --output-path /b/b/e/main/substratevm/svmbuild failed:

> ```
> (gdb) backtrace
> Checker backtrace hello.Hello::main: match 3 failed at line 3 #3  com.oracle.svm.core.JavaMainWrapper::doRun (argc=<optimized out>, argv=<optimized out>) at com/oracle/svm/core/JavaMainWrapper.java:232
> ```

Ok, I think that's from the latest rebase thanks to commit 1b6c1bf193d3c6d5835e067607657accc3d1a9a8 "Add support for running a native-image main entry point in a new thread"

It introduces a new call into the runCore hierarchy by delegating to `doRun` or `doRunInNewThread`

This should have been caught by an mx debuginfotest gate test when the patch was merged. Any idea why not?

A patch is on its way.

adinn:
Ah, that patch did update the test script. It looks like I broke it again when I did the merge.

olpaw:
> mx --kill-with-sigquit --strict-compliance debuginfotest --output-path /b/b/e/main/substratevm/svmbuild failed:

btw, apart from that everything looks good :crossed_fingers:

adinn:
> btw, apart from that everything looks good crossed_fingers

Ok, great. I just pushed a patch that fixes the debug info test. All that is left is to squash the commits.

adinn:
And there's the squashed version

olpaw:
> And there's the squashed version

Restarted internal CI

olpaw (approved):

olpaw:
@adinn the PR is in the merge queue.

PR #4838, opened 2022-08-23
Fix GetDirectBufferCapacity() implementation

koutheir:
[GetDirectBufferCapacity()](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#getdirectbuffercapacity) JNI must return -1 if the given object is not a direct java.nio.Buffer.

This is a continuation of #4715.

### DIFF
--- a/substratevm/mx.substratevm/suite.py
+++ b/substratevm/mx.substratevm/suite.py
@@ -226,6 +226,7 @@ suite = {
                 "java.base" : [
                     "sun.invoke.util",
                     "sun.net",
+                    "sun.nio.ch",
                     "sun.reflect.annotation",
                     "sun.reflect.generics.factory",
                     "sun.reflect.generics.reflectiveObjects",
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
@@ -32,7 +32,6 @@ import java.lang.reflect.Executable;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.nio.Buffer;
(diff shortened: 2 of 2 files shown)


koutheir:
@christianwimmer, the build complains that `The type sun.nio.ch.DirectBuffer is not accessible`. Is the class not actually `public`, like [you said](https://github.com/oracle/graal/pull/4715#discussion_r931523135)? Should I introduce a new `@TargetClass`?

peter-hofer:
The class is public, but restricted by the module system. @olpaw , can this be worked around easily?

christianwimmer:
> The class is public, but access is restricted by the module system

The `suite.py` file has the list of packages that are opened up in the module system for each SVM project. We add lines there whenever we need to open up something new.

koutheir:
> We add lines there whenever we need to open up something new.

Is there any kind of documentation on the structure of that file?

christianwimmer:
There is no documentation, we all do just copy&paste&edit when changing the suite.py...

In your case, it should be enough to add a line for the the package `sun.nio.ch` to the list of the com.oracle.svm.core project: https://github.com/oracle/graal/blob/master/substratevm/mx.substratevm/suite.py#L227

koutheir:
Can someone explain the reason behind the assertion failures during the build?

koutheir:
The build fails for an interesting reason:
```plain
FindBugs warnings were found
M C BC: instanceof will always return false in com.oracle.svm.core.jni.functions.JNIFunctions$Support.directBufferFromJNIHandle(JNIObjectHandle), since a com.oracle.svm.core.jni.functions.JNIFunctions$Target_java_nio_Buffer cannot be a sun.nio.ch.DirectBuffer  At JNIFunctions.java:[line 1318]
```

How can I deal with this issue?

koutheir:
@christianwimmer, any news regarding this issue?

peter-hofer:
@koutheir , findbugs/spotbugs just doesn't know how substitution classes works. Thank you, I'll get this merged.

PR #4868, opened 2022-09-02
Handle control flow split of WithExceptionNode nodes when replacing nodes

koutheir:


koutheir:
Without this fix, running `replaceAtUsages()` on a `WithExceptionNode` leaves pending nodes in the graph, and causes the assertion failure:
```plain
jdk.internal.vm.compiler/org.graalvm.compiler.virtual.phases.ea.VirtualUtil.assertNonReachable(VirtualUtil.java:79)
```

koutheir:
Thank you @gilles-duboscq for your assistance with this issue.

Issue #4431, opened 2022-03-25
[GR-40863] JFR on JDK17: Exception occurred when setting value "150/s" for class jdk.jfr.internal.Control


Karm:
**Describe the issue**
JDK 17 based native-image built application with JFR enabled logs the undermentioned warning:

```
Exception occurred when setting value "150/s" for class jdk.jfr.internal.Control
```

**Steps to reproduce the issue**
It seems rather arbitrary, e.g. with this tiny application [timezones](https://github.com/Karm/mandrel-integration-tests/blob/master/apps/timezones/src/main/java/timezone/Main.java):
```
mvn package

native-image -H:+AllowVMInspection -jar target/timezones.jar target/timezones

./target/timezones -XX:+FlightRecorder -XX:StartFlightRecording=maxsize=10m,filename=logs/flight-native.jfr -XX:FlightRecorderLogging=jfr
[info][jfr] Added periodic hook for EveryChunkPeriodEvents(5686)
[info][jfr] Added periodic hook for EndChunkPeriodEvents(5702)
[info][jfr] Flight Recorder initialized
[info][jfr] Repository base directory: /tmp
[warn][jfr,setting] Exception occurred when setting value "150/s" for class jdk.jfr.internal.Control
[info][jfr        ] Started recording "1" (1) {maxsize=10.0MB, dumponexit=true, filename=/home/karm/workspaceRH/mandrel-integration-tests/apps/timezones/logs/flight-native.jfr}
Fri Mar 25 23:40:50 CET 2022
Central European Standard Time
[info][jfr        ] Stopped recording "1" (1). Reason "Dump on exit".
[info][jfr        ] Wrote recording "1" (1) to /home/karm/workspaceRH/mandrel-integration-tests/apps/timezones/logs/flight-native.jfr
[info][jfr        ] Closed recording "1" (1)
[info][jfr        ] Removed repository /tmp/2022_03_25_23_40_50_379394


./target/timezones -XX:+FlightRecorder -XX:StartFlightRecording=maxsize=10000,delay=5h,filename=logs/flight-native.jfr -XX:FlightRecorderLogging=jfr
[info][jfr] Added periodic hook for EveryChunkPeriodEvents(5686)
[info][jfr] Added periodic hook for EndChunkPeriodEvents(5702)
[info][jfr] Flight Recorder initialized
[info][jfr] Repository base directory: /tmp
[info][jfr] Scheduled recording "1" (1) to start at 2022-03-26T04:40:49.792212
[...]

PR #4895, opened 2022-09-07
Fix exception due to unsupported event throttling

roberttoyonaga:
See related issue here: https://github.com/oracle/graal/issues/4431

### Problem 
`jdk.jfr.internal.JVM.setThrottle(long, long, long)` was being called. Since event throttling is not yet supported, an exception is thrown and and error is printed to the terminal. This is only an issue with JDK 17+

### Solution
Substitute this method so an exception is not thrown, and then log a descriptive warning explaining throttling feature is not supported yet: "`[warn][jfr,system,throttle] JFR event throttling is not yet supported in native images.`"

However, it might be better to silently do nothing instead of show a warning, similar to how the old object sample feature is handled (`Target_jdk_jfr_internal_JVM.emitOldObjectSamples(long, boolean)`).

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
@@ -48,6 +48,8 @@ import com.oracle.svm.core.util.VMError;
 import jdk.jfr.Configuration;
 import jdk.jfr.internal.JVM;
 import jdk.jfr.internal.LogTag;
+import jdk.jfr.internal.LogLevel;
+import jdk.jfr.internal.Logger;
 
 /**
  * Manager class that handles most JFR Java API, see {@link Target_jdk_jfr_internal_JVM}.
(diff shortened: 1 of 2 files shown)


fniephaus (approved):
LGTM. Thanks for the swift fix!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
```diff
@@ -530,6 +532,12 @@ public boolean setCutoff(long eventTypeId, long cutoffTicks) {
         return true;
     }
 
+    /** See {@link JVM#setThrottle}. */
+    public boolean setThrottle(long eventTypeId, long eventSampleSize, long periodMs) {
+        Logger.log(LogTag.JFR_SYSTEM_THROTTLE, LogLevel.WARN, "JFR event throttling is not yet supported in native images.");
```
fniephaus, line 537: This is not going to work on JDK11, but I don't think we need to introduce a JDK17+-specific package for this. I'm going to fix this this in https://github.com/oracle/graal/pull/4896.

fniephaus:
Merged via https://github.com/oracle/graal/commit/1de651475cbb21bc1ba217bd46fc21b7cbb7c3f9.

PR #4916, opened 2022-09-10
HostObject interop with wrapping TruffleObject

JaroslavTulach:
Enso language supports _values with warnings_. They are implemented as a `TruffleObject` that wraps original object, holds the warnings and delegates all messages via `ReflectionLibrary` to the original object. Find more details in the [implementation](https://github.com/enso-org/enso/commit/f6340361a0d82047c56b59f8f2d20a209ba6c0b7#diff-d784205295fc228b108d6ca365bf8371e6fa2337bd8f2c38e7a3bdac721f3a30R14).

This PR provides a [test showing the use-case](https://github.com/oracle/graal/pull/4916/commits/6b9dfae0ef8157a09510dcea9d1860e1411755b2) and demonstrating the failure as the first commit. Additional commits follow showing possible implementation making the test work: [629d3df](https://github.com/oracle/graal/pull/4916/commits/629d3df50769d3ca329d505dfdc6c0105ed46c8d)

### DIFF
--- a/truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/HostObjectWrappedTest.java
+++ b/truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/HostObjectWrappedTest.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or
+ * data (collectively the "Software"), free of charge and under any and all
+ * copyright rights in the Software, and any and all patent rights owned or
+ * freely licensable by each licensor hereunder covering either (i) the
+ * unmodified Software as contributed to or provided by such licensor, or (ii)
+ * the Larger Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ *
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ *
+ * The above copyright notice and either this complete permission notice or at a
+ * minimum a reference to the UPL must be included in all copies or substantial
+ * portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.oracle.truffle.api.test.polyglot;
+
+import org.graalvm.polyglot.Context;
+import org.graalvm.polyglot.HostAccess;
+import org.graalvm.polyglot.Value;
+import static org.junit.Assert.assertEquals;
+import org.junit.Test;
+
+import com.oracle.truffle.api.interop.TruffleObject;
+import com.oracle.truffle.api.library.CachedLibrary;
+import com.oracle.truffle.api.library.ExportLibrary;
+import com.oracle.truffle.api.library.ExportMessage;
+import com.oracle.truffle.api.library.Message;
+import com.oracle.truffle.api.library.ReflectionLibrary;
+import java.lang.reflect.Field;
+import java.math.BigInteger;
+
+public class HostObjectWrappedTest {
+    /**
+     * Enso language supports <em>values with warnings</em>. They are
+     * implemented as a {@link TruffleObject} that wraps original object,
+     * holds the warnings and delegates all messages via {@link ReflectionLibrary} to the original object. 
+     * Find more details in the
+     * <a href="https://github.com/enso-org/enso/commit/f6340361a0d82047c56b59f8f2d20a209ba6c0b7#diff-d784205295fc228b108d6ca365bf8371e6fa2337bd8f2c38e7a3bdac721f3a30R14">implementation</a>.
+     * <p>
+     * This {@code WrapInt} mimics that behavior and tries to perform
+     * <em>host interop</em>. That fails in 21.3 version and needs additional
+     * fixes to work.
+     */
+    @ExportLibrary(ReflectionLibrary.class)
+    public static final class WrapObject implements TruffleObject {
+        private final TruffleObject delegate;
+
+        private WrapObject(TruffleObject delegate) {
+            this.delegate = delegate;
+        }
+
+        @ExportMessage
+        Object send(Message message, Object[] args, @CachedLibrary(limit="3") ReflectionLibrary reflect) throws Exception {
+            return reflect.send(delegate, message, args);
+        }
+
+        public static Value wrap(Value i) throws Exception {
+            Field f = i.getClass().getSuperclass().getDeclaredField("receiver");
+            f.setAccessible(true);
+            TruffleObject o = (TruffleObject) f.get(i);
+            WrapObject wrap = new WrapObject(o);
+            return i.getContext().asValue(wrap);
+        }
+    }
+
+    @Test
+    public void testHostInteropWithWrapper() throws Exception {
+        HostAccess accessPolicy = HostAccess.newBuilder(HostAccess.ALL).allowPublicAccess(true).build();
+        try (Context context = Context.newBuilder().allowHostAccess(accessPolicy).build()) {
+            Value bigIntClass = context.asValue(BigInteger.class).getMember("static");
+            Value four = bigIntClass.invokeMember("valueOf", 444L);
+            Value five = bigIntClass.invokeMember("valueOf", 555L);
+
+            Value simplePlus = four.invokeMember("add", five);
+            assertEquals(444L + 555L, simplePlus.invokeMember("longValue").asLong());
+
+            Value wrapFive = WrapObject.wrap(five);
+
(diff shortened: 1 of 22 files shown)


truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/HostObjectWrappedTest.java
```diff
@@ -0,0 +1,112 @@
...
+            Value four = bigIntClass.invokeMember("valueOf", 444L);
+            Value five = bigIntClass.invokeMember("valueOf", 555L);
+
+            Value simplePlus = four.invokeMember("add", five);
+            assertEquals(444L + 555L, simplePlus.invokeMember("longValue").asLong());
+
+            Value wrapFive = WrapObject.wrap(five);
+
+            Value wrapPlus = four.invokeMember("add", wrapFive);
+            assertEquals(444L + 555L, wrapPlus.invokeMember("longValue").asLong());
```
JaroslavTulach, line 106: When the test is executed on 21.3 branch, it fails with:
```
1) com.oracle.truffle.api.test.polyglot.HostObjectWrappedTest#testHostInteropWithWrapper
java.lang.IllegalArgumentException: Invalid argument when invoking 'add' on '444'(language: Java, type: java.math.BigInteger). Cannot convert '555'(language: Java, type: java.math.BigInteger) to Java type 'java.math.BigInteger': Unsupported target type.Provided arguments: ['555'(language: Java, type: java.math.BigInteger)].
        at com.oracle.truffle.polyglot.PolyglotEngineException.illegalArgument(PolyglotEngineException.java:125)
        at com.oracle.truffle.polyglot.PolyglotValueDispatch.invalidInvokeArgumentType(PolyglotValueDispatch.java:1432)
```

chumer:
Are you using WrapObject.wrap in production code, if so why?

I agree the kind of refactoring sketched in https://github.com/oracle/graal/pull/4916/commits/6b9dfae0ef8157a09510dcea9d1860e1411755b2 would roughly do the trick. We cannot however just use uncached everywhere, that would lead to a performance regression. We would need to create cached libraries and use them where appropriate. We also have some dependent EE code we would need to fix on our side for this.

JaroslavTulach:
> Are you using WrapObject.wrap in production code, if so why?

No, that's just for the test. The production code has `EnsoLanguage extends TruffleLanguage` and obtains `TruffleObject` via [lookupHostSymbol](https://github.com/enso-org/enso/blob/22a371a9c6653ab1bbf9a978a0b086d65de37f28/engine/runtime/src/main/java/org/enso/interpreter/node/expression/builtin/interop/java/LookupClassNode.java#L18). I can rewrite the test to avoid reflection, if I get guidance what to copy. I have found tests for `HostObject` and tests for looking Java classes up, but none that mixes both together.

truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostExecuteNode.java
```diff
@@ -596,7 +595,7 @@ private SingleMethod findBestCandidate(List<SingleMethod> applicableByArity, Obj
                     for (int i = 0; i < paramCount; i++) {
                         if (!HostToTypeNode.canConvert(args[i], parameterTypes[i], genericParameterTypes[i], null,
                                         hostContext, priority, InteropLibrary.getFactory().getUncached(args[i]),
-                                        HostTargetMappingNode.getUncached())) {
+                                        HostTargetMappingNode.getUncached(), HostLibrary.getUncached())) {
```
JaroslavTulach, line 598: OK to use `HostLibrary.getUncached()` as there already was `HostTargetMappingNode.getUncached()`, I think.

truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostTargetMappingNode.java
```diff
@@ -149,12 +150,13 @@ abstract static class SingleMappingNode extends Node {
         protected Object doDefault(Object receiver, @SuppressWarnings("unused") HostTargetMapping cachedMapping,
                         HostContext context, InteropLibrary interop, boolean checkOnly,
                         @Cached ConditionProfile acceptsProfile,
+                        @CachedLibrary(limit="3") HostLibrary library,
```
JaroslavTulach, line 153: Using cached version of `HostLibrary` because there is a node to attach to and because other nodes are `@Cached` as well.

truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostToTypeNode.java
```diff
@@ -140,7 +141,7 @@ protected static Object doGeneric(
         return convertImpl(operand, targetType, genericType, allowsImplementation(context, targetType),
                         isPrimitiveTarget(targetType), context,
                         InteropLibrary.getUncached(operand),
-                        useTargetMapping,
+                        useTargetMapping, HostLibrary.getUncached(),
```
JaroslavTulach, line 144: Using uncached `HostLibrary.getUncached()` as the method is annotated by `@TruffleBoundary`.

JaroslavTulach (commented):
> We cannot however just use uncached everywhere, that would lead to a performance regression

Whenever there is a `Node` _"somewhere"_, the patch uses cached version. The goal is to use uncached only when behind `@TruffleBoundary` or in an `@Override` method (overridden methods are usually not subject to PE, right?). Also when there already is `InteropLibrary.getUncached()` or etc., it is OK to use `HostLibrary.getUncached()`.

truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostLanguageService.java
```diff
@@ -152,7 +152,7 @@ private boolean validHostValue(Object hostValue, HostContext context) {
 
     @Override
     public boolean isHostValue(Object value) {
-        Object obj = HostLanguage.unwrapIfScoped(language, value);
+        Object obj = HostLanguage.unwrapIfScoped(language, HostLibrary.getUncached(), value);
```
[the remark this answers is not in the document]
chumer: That is not true. Such a rule does not exist and does particularly not apply for most of the truffle.polyglot and truffle.host code. The only important thing is that the virtual call can be devirtualized reliably during PE otherwise it would cause a performance warning during compilation. This is generally true for all objects rooted in the PolyglotEngineImpl class like AbstractPolyglotHostService as the engine is always a constant in PE paths.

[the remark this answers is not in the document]
chumer: Note that AbstractPolyglotHostService exists for polyglot isolates, as all host operations need to be handled differently there.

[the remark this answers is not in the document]
JaroslavTulach: The `@Override` claim, was just a (wrong) guess. I know PE doesn't like overwritten methods, but if it resolves to a single method, then it of course works fine.

To compensate for my erroneous judgment, I suggest following check: Modify the `unwrapIfScoped` method to reference some blacklisted method:
```diff
graal/vm$ git diff
diff --git a/truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostLanguage.java b/truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostLanguage.java
index 9706e5f5e56..a045c0ce7b0 100644
--- a/truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostLanguage.java
+++ b/truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostLanguage.java
@@ -61,6 +61,7 @@ import com.oracle.truffle.api.interop.UnsupportedMessageException;
 import com.oracle.truffle.api.nodes.Node;
 import com.oracle.truffle.api.nodes.RootNode;
 import com.oracle.truffle.host.HostMethodScope.ScopedObject;
+import java.io.File;
 
 /*
  * Java host language implementation.
@@ -94,6 +95,7 @@ final class HostLanguage extends TruffleLanguage<HostContext> {
     }
 
     static Object unwrapIfScoped(HostLanguage language, HostLibrary library, Object o) {
+        if (!new File(".").exists()) throw new IllegalStateException();
         HostInstance host = library.asHostInstance(o);
         if (language == null || !language.methodScopingEnabled) {
             return host;
```
and then compile `js` with native image. It gives seven violations like:
```
graal/vm$ mx --native-image=js --java-home ~/bin/labs-11/ --dy=/graal-js,/espresso --env ni-ce build
=== Found 7 compilation blocklist violations ===

Blocklisted method
  java.io.UnixFileSystem.getBooleanAttributes0(File)
called from
  java.io.UnixFileSystem.getBooleanAttributes(UnixFileSystem.java:248)
  java.io.File.exists(File.java:831)
  com.oracle.truffle.host.HostLanguage.unwrapIfScoped(HostLanguage.java:98)
  com.oracle.truffle.host.HostObject.isInstance(HostObject.java:147) -> com.oracle.truffle.host.HostLanguageService.isHostObject(HostLanguageService.java:203)
  com.oracle.truffle.polyglot.EngineAccessor$EngineImpl.isHostObject(EngineAccessor.java:1080)
  com.oracle.truffle.api.TruffleLanguage$Env.isHostObject(TruffleLanguage.java:2116)
  com.oracle.truffle.js.builtins.ConstructorBuiltins$ConstructJavaImporterNode.constructJavaImporter(ConstructorBuiltins.java:2233)
  com.oracle.truffle.js.builtins.ConstructorBuiltinsFactory$ConstructJavaImporterNodeGen.execute(ConstructorBuiltinsFactory.java:4490)
  com.oracle.truffle.js.nodes.control.TopLevelAwaitModuleBodyNode$TopLevelAwaitModuleRootNode.execute(TopLevelAwaitModuleBodyNode.java:106)
  org.graalvm.compiler.truffle.runtime.OptimizedCallTarget.executeRootNode(OptimizedCallTarget.java:650)
  org.graalvm.compiler.truffle.runtime.OptimizedCallTarget.profiledPERoot(OptimizedCallTarget.java:622)
 
```
All of them go thru `com.oracle.truffle.api.TruffleLanguage$Env.isHostObject`. I'll fix it.

truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostInstance.java
```diff
@@ -0,0 +1,6 @@
+package com.oracle.truffle.host;
+
+/** Marker interface to recognize all host objects.
+ */
+interface HostInstance {
```
chumer, line 5: Why is the purpose of this?

truffle/src/com.oracle.truffle.host/src/com/oracle/truffle/host/HostLibrary.java
```diff
@@ -0,0 +1,25 @@
...
+        CompilerAsserts.neverPartOfCompilation("HostLibrary needs to be fast on fast-path");
+        return UNCACHED;
+    }
+    
+    static LibraryFactory<HostLibrary> getFactory() {
+        return FACTORY;
+    }
+    
+    public HostInstance asHostInstance(Object obj) {
+        return obj instanceof HostInstance ? (HostInstance) obj : null;
```
chumer, line 23: Interface type checks are very expensive. Any reason why there is not a isHostInstance and asHostInstance. 
The first returns true|false whether it is a host object. And the second returns not the HostObject wrapper but the underlying real host object. asHostInstance then could fail with an UnsupportedMessageException if it is used with the wrong implementation.

dougxc:
@JaroslavTulach has now signed an OCA.

JaroslavTulach:
I have to put this PR on hold as I don't have time to finish it right now.

JaroslavTulach:
- looks like #12670 provides the same functionality as this PR tried to offer
- thanks, @tzezula!

PR #4950, opened 2022-09-13
[GR-41047] use C++ symbol mangling on Linux to make tools happy

adinn:
This patch encodes local symbols on Linux using a C++-like bfd compatible encoding of the Java method name and signature. This allows both gdb and Linux tools to demangle method names in most circumstances, including in assembly code listings. If works best best used in combination with option `-H:-DeleteLocalSymbols` but still improved behaviour for tools without passing that option.

Note that for gdb to display demangled method names after this patch is enabled it is necessary to execute command `(gdb) set print asm-demangle`.


fniephaus:
Unfortunately, AOT-compiling Graal.js fails with this:

```
[libjsvm:51131] Failed generating 'libjsvm' after 5m 33s.
Fatal error: com.oracle.svm.core.util.VMError$HostedError: java.lang.RuntimeException: Illegal replacement of symbol table entry
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:697)
	at com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:536)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:403)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:580)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:128)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner$JDK9Plus.main(NativeImageGeneratorRunner.java:610)
Caused by: java.lang.RuntimeException: Illegal replacement of symbol table entry
	at com.oracle.objectfile.SymbolTable.tryReplace(SymbolTable.java:45)
	at com.oracle.objectfile.elf.ELFSymtab.lambda$addEntry$0(ELFSymtab.java:447)
	at java.base/java.util.HashMap.compute(HashMap.java:1229)
	at com.oracle.objectfile.elf.ELFSymtab.addEntry(ELFSymtab.java:447)
	at com.oracle.objectfile.elf.ELFSymtab.newDefinedEntry(ELFSymtab.java:435)
	at com.oracle.objectfile.elf.ELFObjectFile.createDefinedSymbol(ELFObjectFile.java:187)
	at com.oracle.svm.hosted.image.LIRNativeImageCodeCache$NativeTextSectionImpl.defineMethodSymbol(LIRNativeImageCodeCache.java:491)
	at com.oracle.svm.hosted.image.NativeImage$NativeTextSectionImpl.writeTextSection(NativeImage.java:951)
	at com.oracle.svm.hosted.image.NativeImage.build(NativeImage.java:457)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:685)
	... 5 more
```

Any idea what could be going on here, @adinn?

adinn:
@fniephaus Hi. I have an idea what might going on here but I may need to debug the problem to be sure.

The error means that the map from ResolvedJavaMethod to encoded name is not 1-1. Clearly this cannot the case for `uniqueShortName`. I can only think of two reasons why my code might be doign this:

1. I am relying on `ResolvedJavaMethod::getname` and `ResolvedJavaType::toJavaName` to provide the base strings that I embed in my encoding whereas ` uniqueShortName` relies on `ResolvedJavaType::toClassName`. I cannot see how that could make any difference but I may be missing something in how some of the overrides work.
2. I might be munging these text components together incorrectly in some corner case. The most likely cause I envisage for that would be that I have wrongly applied the C++ mangler substitutions that replace repeated prefixes with terms like `S_`, S1_` etc. However, there might well be some other coding mistake in there.

Can you provide me with instructions on how to run the gate test in a debugger? If that is not possible can you provide a reproducer I can run that recreates the problem?

adinn:
@fniephaus @olpaw 

Ok, scratch that last comment. I see now that the problem is upstream of method `SubstrateUtils.uniqueShortName`.

This relates to the patch Paul added to `HostedMethod::create` some while back to deal with multiple methods with the same print name. That current patched code for that method feeds `uniqueShortName` a classloader id string and, in the case of a print name clash in the same class loader, a numeric index on the method name. These both serve to disambiguate `HostedMethod` instances that would otherwise map to the same name.

I can easily inherit Paul's disambiguation fix by making `HostedMethod.create` invoke the BFD mangler when we are generating Linux debug info. I'll try that and get back to you with the results on the examples I have played with.

I fear this fix may not be the end of the story as the resulting changes to the demangled name may cause problems for the tools. For non-bootstrap classes or methods which require a disambiguation suffix the demangled name will fail to be an exact match for both the name encoded via DWARF and the Java source name. The disparity might not confuse users very much as far as the ability to identify which method is at play is concerned (although it would certainly not be completely transparent). However, I am not clear what problems these deviations might cause perf and/or valgrind to suffer: firstly, during reporting, when trying to connect profile data with the relevant sources and disassembliesl; and secondly, on the command line, when a user needs to specify the method(s) for which reports of profile data are desired.

I don't think the method name ambiguity will constitute a significant problem. This type of ambiguity only in rare cases. However, I think the loader issue is something we need to resolve.

One thing that strikes me is that we don't actually need classloader prefixes for the system or module loader. Names of classes in all 3 initial loaders should not duplicate each other (the delegation rules ought to ensure that is the case). If so then that would mean a lot less opportunity for tools to be stymied by the name change.

In the case where the class is in a user loader not on the system path a solution to the DWARF name mismatch might be to embed the class loader name into the DWARF info. We could just prepend it to the class info `name` attribute with an extra '.' or '::' separator? Alternatively, we might embed classes belonging to the loader into a namespace whose name is derived from the classloader name?

Any thoughts?

olpaw:
> I don't think the method name ambiguity will constitute a significant problem. This type of ambiguity only in rare cases. However, I think the loader issue is something we need to resolve.

Correct. `SubstrateUtil.toolFriendlyMangle` must have the same semantics as
```
com.oracle.svm.core.SubstrateUtil#uniqueShortName(java.lang.String, jdk.vm.ci.meta.ResolvedJavaType, java.lang.String, jdk.vm.ci.meta.Signature, boolean)
```
The classloader has to be taken into account in the returned String in a uniqueness-defining way.

> Names of classes in all 3 initial loaders should not duplicate each other (the delegation rules ought to ensure that is the case).

This can be implemented as an optimization in `SubstrateUtil.toolFriendlyMangle`.

olpaw (changes requested):
See comment

olpaw:
Also I see internal aarch64 gates failing with these changes with e.g.
```

[wasm:8777] [7/7] Creating image...                                                                      (0.0s @ 4.10GB)
[wasm:8777] ------------------------------------------------------------------------------------------------------------------------
[wasm:8777]                         7.7s (3.6% of total time) in 64 GCs | Peak RSS: 7.90GB | CPU load: 8.74
[wasm:8777] ========================================================================================================================
[wasm:8777] Failed generating 'wasm' after 3m 34s.
Fatal error: com.oracle.svm.core.util.VMError$HostedError: java.lang.RuntimeException: Illegal replacement of symbol table entry
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:697)
	at com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:536)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:403)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:580)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:128)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner$JDK9Plus.main(NativeImageGeneratorRunner.java:610)
Caused by: java.lang.RuntimeException: Illegal replacement of symbol table entry
	at com.oracle.objectfile.SymbolTable.tryReplace(SymbolTable.java:45)
	at com.oracle.objectfile.elf.ELFSymtab.lambda$addEntry$0(ELFSymtab.java:447)
	at java.base/java.util.HashMap.compute(HashMap.java:1229)
	at com.oracle.objectfile.elf.ELFSymtab.addEntry(ELFSymtab.java:447)
	at com.oracle.objectfile.elf.ELFSymtab.newDefinedEntry(ELFSymtab.java:435)
	at com.oracle.objectfile.elf.ELFObjectFile.createDefinedSymbol(ELFObjectFile.java:187)
	at com.oracle.svm.hosted.image.LIRNativeImageCodeCache$NativeTextSectionImpl.defineMethodSymbol(LIRNativeImageCodeCache.java:491)
	at com.oracle.svm.hosted.image.NativeImage$NativeTextSectionImpl.writeTextSection(NativeImage.java:951)
	at com.oracle.svm.hosted.image.NativeImage.build(NativeImage.java:457)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:685)
	... 5 more
```

adinn:
New version factors out unique short name generation into a provider and configures a BFD-compatible provider when generating debug info on Linux.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProvider.java
```diff
@@ -0,0 +1,163 @@
...
+     * runtime's builtin loaders and GraalVM Native's top level loaders an empty string is returned.
+     * There is no need to qualify the method name in this case since the method defined via these
+     * loaders is taken to be the primary instance. This is safe so long as the delegation model for
+     * this special set of loaders ensures that none of them will replicate a class loaded by another
+     * loader in the set.
+     * @param loader The loader whose identifier is to be returned.
+     * @return A unique identifier for the classloader or the empty string when the loader is one
+     * of the special set whose method names do not need qualification.
+     */
+    public abstract String classLoaderNameAndId(ClassLoader loader);
```
olpaw, line 104: If we really want this customizable then we should instead adjust the interface method parameter and use
```
String uniqueShortName(ClassLoader loader, ResolvedJavaType declaringClass, String methodName, Signature methodSignature, boolean isConstructor);
```
instead of
```
String uniqueShortName(String loaderName, ResolvedJavaType declaringClass, String methodName, Signature methodSignature, boolean isConstructor);
```
and then the `classLoaderNameAndId` implementation and its used fields are just private within `DefaultUniqueShortNameProvider`.

olpaw (changes requested):
See comments on recent changes

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProvider.java
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * An abstract provider class that defines a protocol for generating unique short names for
+ * Java methods and fields which are compatible for use as local linker symbols. An instance
+ * of this class may be registered as an image singleton in order to ensure that a specific
+ * naming convention is adopted.
+ */
+public abstract class UniqueShortNameProvider {
+    private static UniqueShortNameProvider provider = null;
+
+    public static synchronized UniqueShortNameProvider provider() {
```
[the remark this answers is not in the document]
adinn: I have changed this to use name `singleton` and simply do an unsynchronized lookup. The default provider is now factored out of the interface and installed as an `AutomaticallyRegisteredImageSingleton`. It uses an associated BooleanSupplier to enable registration when we have no debug info enabled or are not on Linux.
I have retained the `DebugInfoFeature` with an `afterRegistration` callback in order to ensure that the BFD provider gets registered when we do have debug info enabled and are on Linux.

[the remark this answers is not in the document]
olpaw: Sounds good.

Please make sure you use the same BooleanSupplier in both so that it's easy to find the belonging parts in an IDE.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProvider.java
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+        return provider;
+    }
+
+    /**
+     * Returns a short, reasonably descriptive, but still unique name for the provided method
+     * which can be used as a linker local symbol.
+     */
+    public abstract String uniqueShortName(ResolvedJavaMethod method);
```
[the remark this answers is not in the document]
adinn: I was fairly sure that this was not going to be work because of the recursive invocations of the `uniqueShortname` and `classLoaderId` methods in the BFD provider. However, you are right that the provider can be simplified if we move the wrapper method code out into `SubstrateUtil` and/or the BFD provider.

I have condensed the interface down so that we have the following 3 API methods:

```
static UniqueShortNameProvider singleton();
String uniqueShortName(ClassLoader, ResolvedJavaType, String, Signature, boolean);
String uniqueShortName(Member);
```

The API on `SubstrateUtil` provides the following methods which redirect where needed

```
String uniqueShortName(ResolvedJavaMethod);
String uniqueShortName(ClassLoader, ResolvedJavaType, String, Signature, boolean);
String uniqueShortName(Member m);
String uniqueStubName(ResolvedJavaMethod m);
```

I also restored this method to `SubstrateUtil`
```
static String classLoaderNameAndId(ClassLoader loader);
```

The BFD provider implements its own internal loader name generator, also called `classLoaderNameAndId`, which avoids generating names for builtin and Graal loaders. It piggybacks on the implementation in `SustrateUtil`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProvider.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Returns a short, reasonably descriptive, but still unique name derived from the provided method
+     * that can be used as the name of a stub method defined on some factory class that invokes it.
+     * Note that although the value returned by this method is unique to the called method it's purpose
+     * is to uniquely define a single component of the full name of the stub method i.e. the method
+     * selector.
+     */
+
+    public abstract String uniqueStubName(ResolvedJavaMethod method);
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProvider.java
adinn: I had to keep `classLoaderNameAndId` available as an API in `SubstrateUtil` because it is used by the method name comparison code. So, I moved the implementation back there. It does not need to appear as an overrideable API in `UniqueShortNameProvider`. A provider implementation can implement its own internal method for generating loader names, calling out to  `SubstrateUtil` when it wants the field value. The default provide can always just use the `SubstrateUtil` implementation.

adinn:
@olpaw I have just pushed commits for the rework and for the DWARF changes to embed child DIEs for class compile units with a named loader in a wrapper namespace DIE.

n.b. I had to update the API for `UniqueShortNameProvider` to expose the loader encoding method

```
String uniqueShortName(ClassLoader);
```

That's needed because `NativeImageDwarfInfoProvider` needs to pass on the loader encoding to the DWARF generator. The default provider just calls into `SubstrateUtil` to fetch field `classNameAndId`. However the BFD provider retruns "" for loaders that need to be ignored and for those it does not skip it edits the field value to make it more readable.

Assuming there are nor style or format errors this is ready for re-review.

adinn:
@olpaw The latest push is complete, has no style errors and passes the gate tests apart from two sulong errors unrelated to the patch. gdb, perf and valgrind all successfully recognize the DWARF info and demangle the ELF symbols correctly.

I also tested with your test program that has two versions of the same class. I had to tweak the pom to enable javac debug info and had build source jars for Main, Module1 and Module2. I also had to include the source jars for the Module classes in the source search path (the Main sources are found because they sit next to the Main jar).

gdb identifies both versions of the method located under a namespace derived from the relevant classloader id as shown in the session listing that follows:

```
(gdb) run
`/home/adinn/redhat/openjdk/graal/graal/substratevm/main' has changed; re-reading symbols.
Starting program: /home/adinn/redhat/openjdk/graal/graal/substratevm/main 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[New Thread 0x7ffff79ff640 (LWP 1237342)]

Thread 1 "main" hit Breakpoint 1, Main::main(java.lang.String[]*) (args=0x7ffff7a010f0) at Main.java:38
Missing separate debuginfos, use: dnf debuginfo-install zlib-1.2.11-31.fc35.x86_64
(gdb) info fun ::main
All functions matching regular expression "::main":

File Main.java:
38:	void Main::main(java.lang.String[]*);
(gdb) info func ::SameClass
All functions matching regular expression "::SameClass":
(gdb) info func SameClass
All functions matching regular expression "SameClass":

File foo/bar/SameClass.java:
9:	int URLClassLoader_12bcd0c0::foo.bar.SameClass::method();
5:	int URLClassLoader_1fdfafd2::foo.bar.SameClass::method();

File jdk/internal/reflect/Reflection.java:
252:	boolean jdk.internal.reflect.Reflection::isSameClassPackage(java.lang.Class*, java.lang.Class*);
(gdb) c
Continuing.
Hello

Thread 1 "main" hit Breakpoint 2, URLClassLoader_12bcd0c0::foo.bar.SameClass::method() () at foo/bar/SameClass.java:9
(gdb) bt
#0  URLClassLoader_12bcd0c0::foo.bar.SameClass::method() () at foo/bar/SameClass.java:9
#1  0x00000000004a69e6 in com.oracle.svm.core.reflect.ReflectionAccessorHolder::invoke_325e96a0b624d069773b50920fd700a64518797b(java.lang.Object*, java.lang.Object[]*, org.graalvm.nativeimage.c.function.CFunctionPointer*) (__0=<optimized out>, __1=<optimized out>, __2=<optimized out>) at com/oracle/svm/core/reflect/ReflectionAccessorHolder.java:1
#2  0x00000000004a8d0f in com.oracle.svm.core.reflect.SubstrateMethodAccessor::invoke(java.lang.Object*, java.lang.Object[]*) (this=<optimized out>, obj=<optimized out>, args=<optimized out>) at com/oracle/svm/core/reflect/SubstrateMethodAccessor.java:105
#3  0x00000000004060d3 in java.lang.reflect.Method::invoke(java.lang.Object*, java.lang.Object[]*) (this=<optimized out>, obj=<optimized out>, args=<optimized out>) at java/lang/reflect/Method.java:568
#4  Main::main(java.lang.String[]*) (args=<optimized out>) at Main.java:42
#5  0x0000000000413828 in com.oracle.svm.core.JavaMainWrapper::runCore0() () at com/oracle/svm/core/JavaMainWrapper.java:175
#6  0x000000000041353c in com.oracle.svm.core.JavaMainWrapper::runCore() () at com/oracle/svm/core/JavaMainWrapper.java:135
#7  com.oracle.svm.core.JavaMainWrapper::doRun(int, org.graalvm.nativeimage.c.type.CCharPointerPointer*) (argc=<optimized out>, argv=<optimized out>) at com/oracle/svm/core/JavaMainWrapper.java:232
#8  0x0000000000437feb in com.oracle.svm.core.JavaMainWrapper::run(int, org.graalvm.nativeimage.c.type.CCharPointerPointer*) (argc=<optimized out>, argv=<optimized out>) at com/oracle/svm/core/JavaMainWrapper.java:218
#9  com.oracle.svm.core.code.IsolateEnterStub::JavaMainWrapper_run_e6899342f5939c89e6e2f78e2c71f5f4926b786d(int, org.graalvm.nativeimage.c.type.CCharPointerPointer*) (__0=<optimized out>, __1=<optimized out>) at com/oracle/svm/core/code/IsolateEnterStub.java:1
(gdb) c
Continuing.
Calling from Module1
java.lang.Exception: Stack trace
	at java.base@17.0.1/java.lang.Thread.dumpStack(Thread.java:1380)
	at foo.bar.SameClass.method(SameClass.java:10)
	at java.base@17.0.1/java.lang.reflect.Method.invoke(Method.java:568)
	at Main.main(Main.java:42)
java.net.URLClassLoader@12bcd0c0, name: null

Thread 1 "main" hit Breakpoint 2, URLClassLoader_1fdfafd2::foo.bar.SameClass::method() () at foo/bar/SameClass.java:5
(gdb) bt
#0  URLClassLoader_1fdfafd2::foo.bar.SameClass::method() () at foo/bar/SameClass.java:5
#1  0x00000000004a69e6 in com.oracle.svm.core.reflect.ReflectionAccessorHolder::invoke_325e96a0b624d069773b50920fd700a64518797b(java.lang.Object*, java.lang.Object[]*, org.graalvm.nativeimage.c.function.CFunctionPointer*) (__0=<optimized out>, __1=<optimized out>, __2=<optimized out>) at com/oracle/svm/core/reflect/ReflectionAccessorHolder.java:1
#2  0x00000000004a8d0f in com.oracle.svm.core.reflect.SubstrateMethodAccessor::invoke(java.lang.Object*, java.lang.Object[]*) (this=<optimized out>, obj=<optimized out>, args=<optimized out>) at com/oracle/svm/core/reflect/SubstrateMethodAccessor.java:105
#3  0x0000000000406191 in java.lang.reflect.Method::invoke(java.lang.Object*, java.lang.Object[]*) (this=<optimized out>, obj=<optimized out>, args=<optimized out>) at java/lang/reflect/Method.java:568
#4  Main::main(java.lang.String[]*) (args=<optimized out>) at Main.java:43
#5  0x0000000000413828 in com.oracle.svm.core.JavaMainWrapper::runCore0() () at com/oracle/svm/core/JavaMainWrapper.java:175
#6  0x000000000041353c in com.oracle.svm.core.JavaMainWrapper::runCore() () at com/oracle/svm/core/JavaMainWrapper.java:135
#7  com.oracle.svm.core.JavaMainWrapper::doRun(int, org.graalvm.nativeimage.c.type.CCharPointerPointer*) (argc=<optimized out>, argv=<optimized out>) at com/oracle/svm/core/JavaMainWrapper.java:232
#8  0x0000000000437feb in com.oracle.svm.core.JavaMainWrapper::run(int, org.graalvm.nativeimage.c.type.CCharPointerPointer*) (argc=<optimized out>, argv=<optimized out>) at com/oracle/svm/core/JavaMainWrapper.java:218
#9  com.oracle.svm.core.code.IsolateEnterStub::JavaMainWrapper_run_e6899342f5939c89e6e2f78e2c71f5f4926b786d(int, org.graalvm.nativeimage.c.type.CCharPointerPointer*) (__0=<optimized out>, __1=<optimized out>) at com/oracle/svm/core/code/IsolateEnterStub.java:1
(gdb) c
Continuing.
Calling from Module2
java.lang.Exception: Stack trace
	at java.base@17.0.1/java.lang.Thread.dumpStack(Thread.java:1380)
	at foo.bar.SameClass.method(SameClass.java:6)
	at java.base@17.0.1/java.lang.reflect.Method.invoke(Method.java:568)
	at Main.main(Main.java:43)
java.net.URLClassLoader@1fdfafd2, name: null
true
java.net.URLClassLoader@12bcd0c0
java.net.URLClassLoader@1fdfafd2
true
true
jdk.internal.loader.ClassLoaders$AppClassLoader@7907ec20
true
TEST PASSED
[Thread 0x7ffff7f8a740 (LWP 1237341) exited]
[Thread 0x7ffff79ff640 (LWP 1237342) exited]
[New process 1237341]
[Inferior 1 (process 1237341) exited normally]
```

Notice that the two breakpoints are reported at different lines. Of course, the source cache only contains one version of the file -- the one found in the first source jar on the source path. So, the file view when at the second break displays the 'wrong' file, locating the current line in the comment. However, that is only to be expected.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * An automatic feature class which ensures that the Linux debug unique short name provider is
+ * registered when generating debug info for Linux.
+ */
+@AutomaticallyRegisteredFeature
+@SuppressWarnings("unused")
+class NativeImageDebugInfoFeature implements InternalFeature {
+    @Override
+    public void afterRegistration(AfterRegistrationAccess access) {
+        if (OS.LINUX.isCurrent() && SubstrateOptions.GenerateDebugInfo.getValue() > 0) {
```
olpaw, line 47: Use of `!new UniqueShortNameProviderDefaultImpl.UseDefault().getAsBoolean()` here.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProvider.java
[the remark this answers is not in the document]
olpaw: See https://github.com/oracle/graal/pull/4950/files#r985701099

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
olpaw: Or even better:
```patch
diff --git a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProviderDefaultImpl.java b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProviderDefaultImpl.java
index 85459841e5d..14454e39f49 100644
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProviderDefaultImpl.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/UniqueShortNameProviderDefaultImpl.java
@@ -90,11 +91,14 @@ public class UniqueShortNameProviderDefaultImpl implements UniqueShortNameProvid
         return SubstrateUtil.classLoaderNameAndId(classLoader);
     }
 
-    public static class UseDefault implements BooleanSupplier {
+    public static boolean useDefaultProvider() {
+        return !(OS.LINUX.isCurrent() && SubstrateOptions.GenerateDebugInfo.getValue() > 0);
+    }
 
+    public static class UseDefault implements BooleanSupplier {
         @Override
-            return !(OS.LINUX.isCurrent() && SubstrateOptions.GenerateDebugInfo.getValue() > 0);
+            return useDefaultProvider();
         }
     }
 }
diff --git a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
index 698c6523af2..8fcdf8d8ec0 100644
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
@@ -44,7 +44,7 @@ import java.util.List;
 class NativeImageDebugInfoFeature implements InternalFeature {
     @Override
     public void afterRegistration(AfterRegistrationAccess access) {
-        if (OS.LINUX.isCurrent() && SubstrateOptions.GenerateDebugInfo.getValue() > 0) {
+        if (!UniqueShortNameProviderDefaultImpl.useDefaultProvider()) {
             if (!ImageSingletons.contains(UniqueShortNameProvider.class)) {
                 // configure a BFD mangler to provide unique short names for method and field
                 // symbols
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
```diff
@@ -0,0 +1,71 @@
...
+                ClassLoader systemLoader = ClassLoader.getSystemClassLoader();
+                // the Graal app loader and image loader and their parent loader will not duplicate
+                // classes
+                ClassLoader appLoader = accessImpl.getApplicationClassLoader();
+                ClassLoader imageLoader = accessImpl.getImageClassLoader().getClassLoader();
+                ClassLoader imageLoaderParent = imageLoader.getParent();
+                // the app and image loader should both have the same parent
+                assert imageLoaderParent == appLoader.getParent();
+                // ensure the mangle ignores prefix generation for Graal loaders
+                List<ClassLoader> ignored = new ArrayList<>();
```
olpaw, line 62:
```patch
diff --git a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java	(revision Staged)
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java	(date 1664805087484)
@@ -59,11 +59,7 @@
                 // the app and image loader should both have the same parent
                 assert imageLoaderParent == appLoader.getParent();
                 // ensure the mangle ignores prefix generation for Graal loaders
-                List<ClassLoader> ignored = new ArrayList<>();
-                ignored.add(systemLoader);
-                ignored.add(imageLoaderParent);
-                ignored.add(appLoader);
-                ignored.add(imageLoader);
+                List<ClassLoader> ignored = List.of(systemLoader, imageLoaderParent, appLoader, imageLoader);
                 ImageSingletons.add(UniqueShortNameProvider.class, new NativeImageBFDNameProvider(ignored));
             }
         }
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
```diff
@@ -198,7 +207,18 @@ public void createContent() {
          *
          * Details of each specific DIE contents are as follows:
          *
-         * Primitive Types: For each non-void Java primitive type there is a level 0 DIE defining a
+         * A single instance of the level 0 <code>buitin_unit</code> compile unit provides details
```
olpaw, line 210: typo: `s/buitin_unit/builtin_unit`

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -673,6 +674,12 @@ public int headerSize() {
             return getObjectLayout().getFirstFieldOffset();
         }
 
+        @Override
+        public String loaderName() {
+
+            return UniqueShortNameProvider.singleton().uniqueShortName(hostedType.getJavaClass().getClassLoader());
```
olpaw, line 680: This gave me a heart attack for a moment. I though you are trying to use the FQ class-name of the classloader. But you have overloaded the `uniqueShortName` method name to have a variant that takes a loader :sweat_smile: 

It might be good not to use `uniqueShortName` for that but a more explicit `uniqueShortLoaderName`.

olpaw (changes requested):
Nice PR. A few changes are still needed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
adinn: Yes, nice thank you!

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
```diff
@@ -290,6 +310,20 @@ public void createContent() {
          *
          * </ul>
          *
+         * Namespace embedding:
+         *
+         * When the class loader associated with a class defined in a <code>class_unit</code>
+         * compile unit has a non-empty loader id string then a namespace DIE is used to wrap all
+         * its child DIEs. Otherwise the children are embedded directly. The namespace DIE has a
```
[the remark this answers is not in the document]
adinn: Yes, that's actually quiet a nice idea. I'll check it does not introduce any glitches and let you know.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Yes, indeed. Much healthier.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
adinn: Done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
adinn: Done.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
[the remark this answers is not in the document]
adinn: Hmm, I am afraid we suffered an unforeseen adverse reaction from the test subject ...

```
(gdb) ptype 'hello.Hello$DefaultGreeter'
No symbol "hello.Hello$DefaultGreeter" in current context.
info types hello
All types matching regular expression "hello":

File hello/Hello.java:
	(anonymous namespace)::_z_.hello.Hello;
	(anonymous namespace)::_z_.hello.Hello$DefaultGreeter;
	(anonymous namespace)::_z_.hello.Hello$Greeter;
	(anonymous namespace)::_z_.hello.Hello$NamedGreeter;
	(anonymous namespace)::hello.Hello;
	(anonymous namespace)::hello.Hello$DefaultGreeter;
	(anonymous namespace)::hello.Hello$Greeter;
	(anonymous namespace)::hello.Hello$NamedGreeter;

File hello/Target_hello_Hello_DefaultGreeter.java:
	(anonymous namespace)::_z_.hello.SubstituteHelperClass;
	(anonymous namespace)::hello.SubstituteHelperClass;
(gdb) info func hello
All functions matching regular expression "hello":

File hello/Hello.java:
37:	(anonymous namespace)::hello.Hello$Greeter *hello.Hello$Greeter::greeter(java.lang.String[]*);
71:	void hello.Hello$NamedGreeter::greet();
178:	void hello.Hello::inlineTailRecursion(int);
  . . .
(gdb) 
```
So, we will have to stick with selective use of namespaces.

adinn:
@olpaw I have addressed all the review comments. Assuming there are not style/format issues I believe the latest push is complete.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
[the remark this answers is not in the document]
olpaw: Yes, this is ugly. I agree, better not use namespaces when not needed then.

olpaw:
Our CI shows that this change causes aarch64 gates to fail with:
```

[libjsvm:51177] [7/7] Creating image...                                                                  (0.0s @ 7.05GB)
Fatal error: com.oracle.svm.core.util.VMError$HostedError: java.lang.RuntimeException: Illegal replacement of symbol table entry
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:696)
	at com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:535)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:403)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:580)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:128)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner$JDK9Plus.main(NativeImageGeneratorRunner.java:610)
Caused by: java.lang.RuntimeException: Illegal replacement of symbol table entry
	at com.oracle.objectfile.SymbolTable.tryReplace(SymbolTable.java:45)
	at com.oracle.objectfile.elf.ELFSymtab.lambda$addEntry$0(ELFSymtab.java:447)
	at java.base/java.util.HashMap.compute(HashMap.java:1229)
	at com.oracle.objectfile.elf.ELFSymtab.addEntry(ELFSymtab.java:447)
	at com.oracle.objectfile.elf.ELFSymtab.newDefinedEntry(ELFSymtab.java:435)
	at com.oracle.objectfile.elf.ELFObjectFile.createDefinedSymbol(ELFObjectFile.java:187)
	at com.oracle.svm.hosted.image.LIRNativeImageCodeCache$NativeTextSectionImpl.defineMethodSymbol(LIRNativeImageCodeCache.java:491)
	at com.oracle.svm.hosted.image.NativeImage$NativeTextSectionImpl.writeTextSection(NativeImage.java:938)
	at com.oracle.svm.hosted.image.NativeImage.build(NativeImage.java:458)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:684)
	... 5 more
```
I'm creating a reproducer right now that I can share with you.

olpaw:
I make the runtime exception more verbose and it now shows me:
```
Fatal error: com.oracle.svm.core.util.VMError$HostedError:
java.lang.RuntimeException: Illegal replacement of symbol table entry: 
oldEntry name: _ZN44com.oracle.js.parser.ir.Module$ModuleRequest21.Module$ModuleRequestEP44com.oracle.truffle.api.strings.TruffleStringP13java.util.Map, defined:true, 
newEntry name: _ZN44com.oracle.js.parser.ir.Module$ModuleRequest21.Module$ModuleRequestEP44com.oracle.truffle.api.strings.TruffleStringP13java.util.Map
```
So it seems that a symbol gets registered twice with the same name where that did not happen before.

adinn:
@olpaw It would really help if you could provide a reproducer for this failure.

olpaw:
> @olpaw It would really help if you could provide a reproducer for this failure.

On it. Unfortunately I could not yet get it into something nice. I will continue tomorrow.

adinn:
> On it. Unfortunately I could not yet get it into something nice. I will continue tomorrow.

Thanks, Paul.

olpaw:
> Is it possible that class Module$ModuleRequest could have been loaded and had its constructor compiled twice via two of the Graal loaders that I am discounting from including their name in the symbol definition?

I think that the most likely reason.

But still even if we would end up with two identical symbol names the name collision handling in `HostedMethod#create` (`uniqueShortName` setting) should prevent that we ever have two symbols with identical names. But looking into `com.oracle.svm.hosted.image.NativeImageBFDNameProvider.BFDMangler#mangle(java.lang.String, jdk.vm.ci.meta.ResolvedJavaType, java.lang.String, jdk.vm.ci.meta.Signature, boolean)` shows me that the collision handling would not work for Constructors because there you do not use the `memberName`. I'm now testing this theory by running the test with assertions enabled. I suspect I should get `assert selector.equals("<init>");` to fail in this case.

adinn:
> But looking into com.oracle.svm.hosted.image.NativeImageBFDNameProvider.BFDMangler#mangle(java.lang.String, jdk.vm.ci.meta.ResolvedJavaType, java.lang.String, jdk.vm.ci.meta.Signature, boolean) shows me that the collision handling would not work for Constructors because there you do not use the memberName.

Ah yes, the code assumes (asserts) the member name is `<init>` and substitutes the class name minus the package for the selector. So, if you have already detected a name collision then this will undo any correction previously applied. It may be that what we have here is a fallback compiled version of the constructor.

olpaw:
> I'm now testing this theory by running the test with assertions enabled. I suspect I should get assert selector.equals("<init>"); to fail in this case.

That's exactly what is happening!

> Ah yes, the code assumes (asserts) the member name is <init> and substitutes the class name minus the package for the selector. So, if you have already detected a name collision then this will undo any correction previously applied.

`NativeImageBFDNameProvider.BFDMangler#mangle` breaks my assumption that a name collision can always be solved by modifying the `name` of the method. To prevent such mistakes in the future, there should be the following addition to `com.oracle.svm.hosted.meta.HostedMethod#create`:
```patch
diff --git a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/meta/HostedMethod.java b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/meta/HostedMethod.java
index 559b731bc85..be7a80a4ed2 100644
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/meta/HostedMethod.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/meta/HostedMethod.java
@@ -109,6 +109,7 @@ public final class HostedMethod implements SharedMethod, WrappedJavaMethod, Grap
         if (collisionCount > 0) {
             name = name + METHOD_NAME_COLLISION_SUFFIX + collisionCount;
             uniqueShortName = SubstrateUtil.uniqueShortName(holder.getJavaClass().getClassLoader(), holder, name, signature, wrapped.isConstructor());
+            VMError.guarantee(!uniqueShortName.equals(uniqueShortName), "methodName is not part the uniqueShortName of the HostedMethod created for " + wrapped);
         }
         return new HostedMethod(wrapped, holder, signature, constantPool, handlers, deoptOrigin, name, uniqueShortName, localVariableTable);
     }
```

adinn:
> > I'm now testing this theory by running the test with assertions enabled. I suspect I should get assert selector.equals("<init>"); to fail in this case.
>
> That's exactly what is happening!

Ok, I will fix this by asserting `selector.startsWith("<init>") and making the encoding substitute `<init>` with the base class name. If the selector contains a suffix then it will be carried over into the new selector.

olpaw:
Just to be 100% sure, I did the following experimental change:
```patch
diff --git a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
index 7c911ce..bcacbce 100644
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
@@ -363,7 +363,7 @@ class NativeImageBFDNameProvider implements UniqueShortNameProvider {
         public String mangle(String loaderName, ResolvedJavaType declaringClass, String memberName, Signature methodSignature, boolean isConstructor) {
             String fqn = declaringClass.toJavaName();
             String selector = memberName;
-            if (isConstructor) {
+            if (isConstructor && false) {
                 assert methodSignature != null;
                 assert selector.equals("<init>");
                 int index = fqn.lastIndexOf('.');
```
which fixes the issue. In other words for constructors `mangle` breaks it's promise to produce truly unique names for a given set of input parameters because it currently does not take the value of `memberName` into account.

olpaw:
@matneu this PR needs a security review because it modified the symbol names we generate in native-images when the user builds with debug-info generation (`-g`) on Linux. In this case the symbol names contain the fully qualified type signatures of a method/constructor instead of generating hashes like we do per default. Thus more information is given to the user when this is used.

adinn:
@olpaw I just pushed a fix that includes the guarantee for uniqueness of the fixed name and modifies the generator to retain any suffix on the constructor name.

olpaw:
@adinn rerunning the CI gates right now ...

olpaw:
@adinn looks like all gates are passing now. There were two failing gates but I suspect them to be transients that I'm rerunning now.

olpaw (approved):

PR #4964, opened 2022-09-15
LLVM Source Level debug info support

rishikeshdevsot:
(1) the change or feature is needed

Graal substratevm has support for an LLVM backend but it is not possible to associate source level location information (e.g. line number, filename) with the output IR. That is what this commit aims to do, adding an option called "IncludeLLVMSourceDebugInfo" which includes source level debug information in the output IR built using the LLVM DI Builder interface . This information in the IR can be helpful for debugging purposes and performing static analysis on the IR. 

(2) how it is implemented

GraalVM SubstrateVM currently uses JavaCPP's JNI functions for llvm to build the LLVM IR. I am using the JNI methods of the LLVM DIBuilder class to create location information with corresponding line number, subprogram and filename. This location information is associated to the corresponding LLVM Instruction, function or basic block using the `LLVMSetCurrentDebugLocation2` and `LLVMSetInstDebugLocation` functions. The line number, method name and file name information is derived using a new class called LLVMDebugLineInfo which is synonymous to the NativeImageDebugLineInfo class from previous GraalVM releases (21.1.0).

The debug location is attached to an IR instruction by calling `buildDebugInfoForInstr()` from inside the setResult() method in NodeLLVMBuilder class. `buildDebugInfoForInstr()` is defined inside the LLVMIRBuilder class and uses the methods of the LLVM DIBuilder to generate the source level location information.

(3) Notes
1. I needed to use the `SourceManager` class to obtain the file and directory information associated with the node. Once `SourceManager` was defined, I added to the `ImageSingletons` map inside `emitLLVM()`. But the problem is that `emitLLVM()` seems to be run by multiple threads which would cause multiple additions of the `SourceManager` to `ImageSingletons`. I've currently sidestepped the problem by enclosing the addition to `ImageSingletons` within a lock but ideally I would like to perform this addition before the multiple threads are launched but I am not sure where exactly that is? 

2. When using the `-O0` flag to generate the native-image with an LLVM backend, I get an error saying "the LLVM backend doesn't support debug info generation" called from inside `visitFullInfoPointNode` in NodeLLVMBuilder and I am not sure how to avoid this problem. This problem was not present when I implemented this feature in Graal 21.1.0. 

3. I have tested the feature to compile Hadoop to output LLVM IR with location information using GraalVM 21.1.0. But for the latest build, I have only tested it with toy examples.
 
The flags required to obtain location information are `-g` and `-H=+IncludeLLVMSourceDebugInfo`.

This PR is hopefully just some startup code so that I can receive feedback and comments to understand how the code needs to structured and any other requirements that might me needed.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/NodeLLVMBuilder.java
+++ b/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/NodeLLVMBuilder.java
@@ -35,6 +35,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import com.oracle.svm.core.graal.llvm.util.LLVMOptions;
 import org.graalvm.compiler.code.CompilationResult;
 import org.graalvm.compiler.core.common.NumUtil;
 import org.graalvm.compiler.core.common.calc.Condition;
@@ -796,6 +797,12 @@ public class NodeLLVMBuilder implements NodeLIRBuilderTool, SubstrateNodeLIRBuil
                         "value type doesn't match node stamp (" + node.stamp(NodeView.DEFAULT).toString() + ")", llvmOperand.get());
 
         gen.getDebugInfoPrinter().setValueName(llvmOperand, node);
+        if (LLVMOptions.IncludeLLVMSourceDebugInfo.getValue()) {
+            if (llvmOperand.get() instanceof LLVMValueRef) {
+                LLVMValueRef instr = llvmOperand.get();
+                builder.buildDebugInfoForInstr(node, instr);
+            }
+        }
         valueMap.put(node, llvmOperand);
         return operand;
     }
--- a/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMBackend.java
+++ b/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMBackend.java
@@ -28,7 +28,9 @@ import static com.oracle.svm.core.util.VMError.unimplemented;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.locks.ReentrantLock;
 
+import com.oracle.svm.hosted.image.sources.SourceManager;
 import org.graalvm.compiler.code.CompilationResult;
 import org.graalvm.compiler.core.common.CompilationIdentifier;
 import org.graalvm.compiler.debug.CounterKey;
@@ -61,10 +63,12 @@ import jdk.vm.ci.meta.JavaConstant;
 import jdk.vm.ci.meta.JavaKind;
 import jdk.vm.ci.meta.ResolvedJavaMethod;
 import jdk.vm.ci.meta.VMConstant;
+import org.graalvm.nativeimage.ImageSingletons;
 
 public class SubstrateLLVMBackend extends SubstrateBackend {
     private static final TimerKey EmitLLVM = DebugContext.timer("EmitLLVM").doc("Time spent generating LLVM from HIR.");
     private static final TimerKey BackEnd = DebugContext.timer("BackEnd").doc("Time spent in EmitLLVM and Populate.");
+    private ReentrantLock imageSingletonesLock = new ReentrantLock();
 
     public SubstrateLLVMBackend(Providers providers) {
         super(providers);
@@ -121,6 +125,14 @@ public class SubstrateLLVMBackend extends SubstrateBackend {
             NodeLLVMBuilder nodeBuilder = newNodeLLVMBuilder(graph, generator);
 
             /* LLVM generation */
+            if (LLVMOptions.IncludeLLVMSourceDebugInfo.getValue()) {
+                //TODO: Avoid this lock if possible
+                imageSingletonesLock.lock();
+                if (ImageSingletons.contains(SourceManager.class) == false) {
+                    ImageSingletons.add(SourceManager.class, new SourceManager());
+                }
+                imageSingletonesLock.unlock();
+            }
             generate(nodeBuilder, graph);
             byte[] bitcode = generator.getBitcode();
             result.setTargetCode(bitcode, bitcode.length);
--- a/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
+++ b/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
@@ -1,0 +1,205 @@
+/*
+ * Copyright (c) 2018, 2019, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.core.graal.llvm.util;
+
+import com.oracle.graal.pointsto.infrastructure.OriginalClassProvider;
+import com.oracle.graal.pointsto.infrastructure.WrappedJavaMethod;
+import com.oracle.graal.pointsto.meta.AnalysisType;
+import com.oracle.svm.core.SubstrateOptions;
+import com.oracle.svm.hosted.annotation.CustomSubstitutionMethod;
+import com.oracle.svm.hosted.annotation.CustomSubstitutionType;
+import com.oracle.svm.hosted.image.NativeImage;
+import com.oracle.svm.hosted.image.sources.SourceManager;
+import com.oracle.svm.hosted.lambda.LambdaSubstitutionType;
+import com.oracle.svm.hosted.meta.HostedMethod;
+import com.oracle.svm.hosted.meta.HostedType;
+import com.oracle.svm.hosted.substitute.InjectedFieldsType;
+import com.oracle.svm.hosted.substitute.SubstitutionMethod;
+import com.oracle.svm.hosted.substitute.SubstitutionType;
+import jdk.vm.ci.meta.LineNumberTable;
+import jdk.vm.ci.meta.ResolvedJavaMethod;
+import jdk.vm.ci.meta.ResolvedJavaType;
+import org.graalvm.compiler.debug.DebugContext;
+import org.graalvm.compiler.graph.NodeSourcePosition;
+import org.graalvm.nativeimage.ImageSingletons;
+
+import java.nio.file.Path;
+
+import static org.graalvm.compiler.debug.GraalError.unimplemented;
+
+// Similar to NativeImageDebugLineInfo from previous releases
+public class LLVMDebugLineInfo {
+    private final int bci;
+    private final ResolvedJavaMethod method;
+    private final DebugContext debugContext;
+    private Path cachePath;
+    private Path fullFilePath;
+
+    LLVMDebugLineInfo(DebugContext debugContext, NodeSourcePosition position) {
+        int posbci = position.getBCI();
+        this.bci = (posbci >= 0 ? posbci : 0);
+        this.method = position.getMethod();
+        this.cachePath = SubstrateOptions.getDebugInfoSourceCacheRoot();
+        this.debugContext = debugContext;
+        computeFullFilePath();
+    }
+
+    private static ResolvedJavaType getOriginal(ResolvedJavaType javaType) {
+        if (javaType instanceof SubstitutionType) {
+            return ((SubstitutionType) javaType).getOriginal();
+        } else if (javaType instanceof CustomSubstitutionType<?, ?>) {
+            return ((CustomSubstitutionType<?, ?>) javaType).getOriginal();
+        } else if (javaType instanceof LambdaSubstitutionType) {
+            return ((LambdaSubstitutionType) javaType).getOriginal();
+        } else if (javaType instanceof InjectedFieldsType) {
+            return ((InjectedFieldsType) javaType).getOriginal();
+        }
+        return null;
+    }
+
+    protected static ResolvedJavaType getJavaType(HostedMethod hostedMethod, boolean wantOriginal) {
+        if (wantOriginal) {
+            /* Check for wholesale replacement of the original class. */
+            HostedType hostedType = hostedMethod.getDeclaringClass();
+            ResolvedJavaType javaType = hostedType.getWrapped().getWrappedWithoutResolve();
+            ResolvedJavaType originalType = getOriginal(javaType);
+            if (originalType != null) {
+                return originalType;
+            }
+            /* Check for replacement of the original method only. */
+            ResolvedJavaMethod javaMethod = hostedMethod.getWrapped().getWrapped();
+            if (javaMethod instanceof SubstitutionMethod) {
+                return ((SubstitutionMethod) javaMethod).getOriginal().getDeclaringClass();
+            } else if (javaMethod instanceof CustomSubstitutionMethod) {
+                return ((CustomSubstitutionMethod) javaMethod).getOriginal().getDeclaringClass();
+            }
+            return hostedType.getWrapped().getWrapped();
+        }
+        ResolvedJavaMethod javaMethod = hostedMethod.getWrapped().getWrapped();
+        return javaMethod.getDeclaringClass();
+    }
+
+    public String fileName() {
+        if (fullFilePath != null) {
+            Path fileName = fullFilePath.getFileName();
+            if (fileName != null) {
+                return fileName.toString();
+            }
+        }
+        return null;
+    }
+
+    public Path filePath() {
+        if (fullFilePath != null) {
+            return fullFilePath.getParent();
+        }
+        return null;
+    }
+
+    public Path cachePath() {
+        return cachePath;
+    }
+
+    public String className() {
+        return method.format("%H");
+    }
+
+    public String methodName() {
+        ResolvedJavaMethod targetMethod = method;
+        while (targetMethod instanceof WrappedJavaMethod) {
+            targetMethod = ((WrappedJavaMethod) targetMethod).getWrapped();
+        }
+        if (targetMethod instanceof SubstitutionMethod) {
+            targetMethod = ((SubstitutionMethod) targetMethod).getOriginal();
+        } else if (targetMethod instanceof CustomSubstitutionMethod) {
+            targetMethod = ((CustomSubstitutionMethod) targetMethod).getOriginal();
+        }
+        String name = targetMethod.getName();
+        if (name.equals("<init>")) {
+            if (method instanceof HostedMethod) {
+                name = getJavaType((HostedMethod) method, true).toJavaName();
+                if (name.indexOf('.') >= 0) {
+                    name = name.substring(name.lastIndexOf('.') + 1);
+                }
+                if (name.indexOf('$') >= 0) {
+                    name = name.substring(name.lastIndexOf('$') + 1);
+                }
+            } else {
+                name = targetMethod.format("%h");
+                if (name.indexOf('$') >= 0) {
+                    name = name.substring(name.lastIndexOf('$') + 1);
+                }
+            }
+        }
+        return name;
+    }
+
+    public String symbolNameForMethod() {
+        return NativeImage.localSymbolNameForMethod(method);
+    }
+
+    public int addressLo() {
+        throw unimplemented();
+    }
+
+    public int addressHi() {
+        throw unimplemented();
+    }
+
+    public int line() {
+        LineNumberTable lineNumberTable = method.getLineNumberTable();
+        if (lineNumberTable != null) {
+            return lineNumberTable.getLineNumber(bci);
+        }
+        return 0;
+    }
+
+    @SuppressWarnings("try")
+    public void computeFullFilePath() {
+        ResolvedJavaType declaringClass = method.getDeclaringClass();
+        Class<?> clazz = null;
+        if (declaringClass instanceof OriginalClassProvider) {
+            clazz = ((OriginalClassProvider) declaringClass).getJavaClass();
+        }
+        /*
+         * HostedType wraps an AnalysisType and both HostedType and AnalysisType punt calls to
+         * getSourceFilename to the wrapped class so for consistency we need to do the path
+         * lookup relative to the doubly unwrapped HostedType or singly unwrapped AnalysisType.
+         */
+        if (declaringClass instanceof HostedType) {
+            declaringClass = ((HostedType) declaringClass).getWrapped();
+        }
+        if (declaringClass instanceof AnalysisType) {
+            declaringClass = ((AnalysisType) declaringClass).getWrapped();
+        }
+        SourceManager sourceManager = ImageSingletons.lookup(SourceManager.class);
+        try (DebugContext.Scope s = debugContext.scope("DebugCodeInfo", declaringClass)) {
+            fullFilePath = sourceManager.findAndCacheSource(declaringClass, clazz, debugContext);
+        } catch (Throwable e) {
+            throw debugContext.handle(e);
+        }
+    }
+}
+
--- a/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMIRBuilder.java
+++ b/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMIRBuilder.java
@@ -32,9 +32,12 @@ import static com.oracle.svm.core.graal.llvm.util.LLVMUtils.dumpValues;
 import static org.graalvm.compiler.debug.GraalError.shouldNotReachHere;
 
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import com.oracle.svm.shadowed.org.bytedeco.llvm.LLVM.LLVMDIBuilderRef;
+import com.oracle.svm.shadowed.org.bytedeco.llvm.LLVM.LLVMMetadataRef;
 import org.graalvm.compiler.core.common.NumUtil;
 import org.graalvm.compiler.core.common.calc.Condition;
 import org.graalvm.compiler.core.common.memory.MemoryOrderMode;
@@ -54,6 +57,9 @@ import com.oracle.svm.shadowed.org.bytedeco.llvm.LLVM.LLVMModuleRef;
 import com.oracle.svm.shadowed.org.bytedeco.llvm.LLVM.LLVMTypeRef;
 import com.oracle.svm.shadowed.org.bytedeco.llvm.LLVM.LLVMValueRef;
 import com.oracle.svm.shadowed.org.bytedeco.llvm.global.LLVM;
+import org.graalvm.compiler.debug.DebugContext;
+import org.graalvm.compiler.graph.NodeSourcePosition;
+import org.graalvm.compiler.nodes.ValueNode;
 
 public class LLVMIRBuilder implements AutoCloseable {
     private static final String DEFAULT_INSTR_NAME = "";
@@ -62,7 +68,11 @@ public class LLVMIRBuilder implements AutoCloseable {
     private LLVMBuilderRef builder;
     private LLVMModuleRef module;
     private LLVMValueRef function;
-
+    public LLVMDIBuilderRef diBuilder;
+    // Maps filenames to a compile unit
+    public HashMap<String, LLVMMetadataRef> diFilenameToCU;
+    // Maps function names to a LLVM SubPrograms
+    public  HashMap<String, LLVMMetadataRef> diFunctionToSP;
     private final boolean primary;
     private final LLVMHelperFunctions helpers;
 
@@ -70,6 +80,15 @@ public class LLVMIRBuilder implements AutoCloseable {
         this.context = LLVM.LLVMContextCreate();
         this.builder = LLVM.LLVMCreateBuilderInContext(context);
         this.module = LLVM.LLVMModuleCreateWithNameInContext(name, context);
+        if (LLVMOptions.IncludeLLVMSourceDebugInfo.getValue()) {
+            String dbgInfoFlagName = "Debug Info Version";
+            LLVMValueRef dbgInfoVal = LLVM.LLVMConstInt(LLVM.LLVMInt32TypeInContext(context), LLVM.LLVMDebugMetadataVersion(), 0);
+            LLVM.LLVMAddModuleFlag(this.module, LLVM.LLVMModuleFlagBehaviorWarning, dbgInfoFlagName,
+                    dbgInfoFlagName.length(), LLVM.LLVMValueAsMetadata(dbgInfoVal));
+            this.diBuilder = LLVM.LLVMCreateDIBuilder(this.module);
+            diFilenameToCU = new HashMap<String, LLVMMetadataRef>();
+            diFunctionToSP = new HashMap<String, LLVMMetadataRef>();
+        }
         this.primary = true;
         this.helpers = new LLVMHelperFunctions(this);
     }
@@ -78,6 +97,11 @@ public class LLVMIRBuilder implements AutoCloseable {
         this.context = primary.context;
         this.builder = LLVM.LLVMCreateBuilderInContext(context);
         this.module = primary.module;
+        if (LLVMOptions.IncludeLLVMSourceDebugInfo.getValue()) {
+            this.diBuilder = LLVM.LLVMCreateDIBuilder(this.module);
+            diFilenameToCU = new HashMap<String, LLVMMetadataRef>();
+            diFunctionToSP = new HashMap<String, LLVMMetadataRef>();
+        }
         this.function = null;
         this.primary = false;
         this.helpers = null;
@@ -113,6 +137,12 @@ public class LLVMIRBuilder implements AutoCloseable {
     public void close() {
         LLVM.LLVMDisposeBuilder(builder);
         builder = null;
+        if (LLVMOptions.IncludeLLVMSourceDebugInfo.getValue()) {
+            LLVM.LLVMDIBuilderFinalize(diBuilder);
+            diFilenameToCU.clear();
+            diFunctionToSP.clear();
+            diBuilder = null;
+        }
         if (primary) {
             LLVM.LLVMDisposeModule(module);
             module = null;
@@ -797,6 +827,89 @@ public class LLVMIRBuilder implements AutoCloseable {
 
     public void buildDebugtrap() {
         buildIntrinsicCall("llvm.debugtrap", functionType(voidType()));
+    }
+
+    private LLVMMetadataRef getCompileUnit(String filename, LLVMMetadataRef diFile) {
+        final String producer = "Graal Compiler";
+        LLVMMetadataRef compileUnit;
+        if (diFilenameToCU.containsKey(filename)) {
+            compileUnit = diFilenameToCU.get(filename);
+        } else {
+            compileUnit = LLVM.LLVMDIBuilderCreateCompileUnit(diBuilder, LLVM.LLVMDWARFSourceLanguageJava, diFile, producer, producer.length(),
+                    0, "", 0, 0, "", 0, LLVM.LLVMDWARFEmissionFull,
+                    0, 1, 1, "", 0, "", 0);
+
+
+            diFilenameToCU.put(filename, compileUnit);
+
+        }
+        return compileUnit;
+    }
+
+    private LLVMMetadataRef getSubProgram(String filename, String funcName, LLVMMetadataRef diFile) {
+        // Key to map a function to its LLVM Sub Program
+        String key = filename + ":" + funcName;
+        LLVMMetadataRef subProgram;
+        if (diFunctionToSP.containsKey(key)) {
+            subProgram = diFunctionToSP.get(key);
+        } else {
+            //TODO: Create the correct function type, currently all functions are treated as "void" (DWARF type float: 0x4) return types with no paramenters
+            String typeName = "void";
+            LLVMMetadataRef[] paramType = {LLVM.LLVMDIBuilderCreateBasicType(diBuilder, typeName,
+                    typeName.length(), 32, 4, 0)};
+            PointerPointer<LLVMMetadataRef> paramTypeP = new PointerPointer<LLVMMetadataRef>(paramType);
+
+            LLVMMetadataRef spType = LLVM.LLVMDIBuilderCreateSubroutineType(diBuilder,
+                    diFile, paramTypeP, 1, 0);
+            //TODO: Get the function and scope line num if possible
+            subProgram = LLVM.LLVMDIBuilderCreateFunction(
+                    diBuilder, diFile, funcName, funcName.length(), funcName, funcName.length(),
+                    diFile, 0, // Func Line
+                    spType, 0, 1,
+                    0, // Scope Line
+                    0, 0);
+            diFunctionToSP.put(key, subProgram);
+        }
+        return subProgram;
+    }
+
+    public void buildDebugInfoForInstr(ValueNode node, LLVMValueRef instr) {
+        /* If the LLVMValueRef is of the type of instruction, function or basic block, then add source location information
+        to this reference if the `node` contains it
+         */
+        if ((LLVM.LLVMGetValueKind(instr) == LLVM.LLVMInstructionValueKind) ||
+                (LLVM.LLVMGetValueKind(instr) == LLVM.LLVMBasicBlockValueKind)
+                || (LLVM.LLVMGetValueKind(instr) == LLVM.LLVMFunctionValueKind)) {
+            if (node.getNodeSourcePosition() != null) {
+                NodeSourcePosition position = node.getNodeSourcePosition();
+                DebugContext debugContext = node.getDebug();
+                LLVMDebugLineInfo dbgLineInfo = new LLVMDebugLineInfo(debugContext, position);
+                int lineNum = dbgLineInfo.line();
+                dbgLineInfo.computeFullFilePath();
+                String filename = dbgLineInfo.fileName();
+
+                if (filename != null) {
+                    String directory = String.valueOf(dbgLineInfo.filePath());
+                    String funcName = dbgLineInfo.methodName();
+                    /* To add debug location information in the LLVM IR:
+                    1. A compile unit needs to be created which acts as a unit of translation, usually a compile unit is
+                    associated with a file
+                    2. Scope needs to be attached to each DILocation metadata. Here, a subProgram corresponding to the function
+                    is used to satisfy this scope requirement
+                    3. The DILocation is created and set to the instruction reference
+                    Source: https://llvm.org/docs/tutorial/MyFirstLanguageFrontend/LangImpl09.html and
+                    https://llvm.org/docs/SourceLevelDebugging.html
+                     */
+                    LLVMMetadataRef diFile = LLVM.LLVMDIBuilderCreateFile(diBuilder, filename, filename.length(), directory, directory.length());
+                    getCompileUnit(filename, diFile);
+                    LLVMMetadataRef subProgram = getSubProgram(filename, funcName, diFile);
+
+                    LLVMMetadataRef diLocation = LLVM.LLVMDIBuilderCreateDebugLocation(context, lineNum, 0, subProgram, null);
+                    LLVM.LLVMSetCurrentDebugLocation2(builder, diLocation);
+                    LLVM.LLVMSetInstDebugLocation(builder, instr);
+                }
+            }
+        }
     }
 
     public LLVMValueRef functionEntryCount(LLVMValueRef count) {
--- a/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMOptions.java
+++ b/substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMOptions.java
@@ -49,5 +49,8 @@ public class LLVMOptions {
     @Option(help = "Enable LLVM bitcode optimizations")//
     public static final HostedOptionKey<Boolean> BitcodeOptimizations = new HostedOptionKey<>(false);
 
+    @Option(help = "Include source code level debug info in the output LLVM IR", type = OptionType.Debug)//
+    public static final HostedOptionKey<Boolean> IncludeLLVMSourceDebugInfo = new HostedOptionKey<>(false);
+
     public static final List<HostedOptionKey<?>> allOptions = Arrays.asList(IncludeLLVMDebugInfo, DumpLLVMStackMap, LLVMMaxFunctionsPerBatch, CustomLD, BitcodeOptimizations);
 }
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImage.java
@@ -470,7 +470,9 @@ public abstract class NativeImage extends AbstractImage {
             if (SubstrateOptions.GenerateDebugInfo.getValue(HostedOptionValues.singleton()) > 0) {
                 Timer timer = TimerCollection.singleton().get(TimerCollection.Registry.DEBUG_INFO);
                 try (Timer.StopTimer t = timer.start()) {
-                    ImageSingletons.add(SourceManager.class, new SourceManager());
+                    if (ImageSingletons.contains(SourceManager.class) == false) {
+                        ImageSingletons.add(SourceManager.class, new SourceManager());
+                    }
                     DebugInfoProvider provider = new NativeImageDebugInfoProvider(debug, codeCache, heap, metaAccess);
                     objectFile.installDebugInfo(provider);
                 }


substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/NodeLLVMBuilder.java
```diff
@@ -796,6 +797,12 @@ public Value setResult(ValueNode node, Value operand) {
                         "value type doesn't match node stamp (" + node.stamp(NodeView.DEFAULT).toString() + ")", llvmOperand.get());
 
         gen.getDebugInfoPrinter().setValueName(llvmOperand, node);
+        if (LLVMOptions.IncludeLLVMSourceDebugInfo.getValue()) {
+            if (llvmOperand.get() instanceof LLVMValueRef) {
```
loicottet, line 801: As far as I can tell `llvmOperand.get()` is always an `LLVMValueRef`

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
```diff
@@ -0,0 +1,205 @@
...
+            return ((CustomSubstitutionType<?, ?>) javaType).getOriginal();
+        } else if (javaType instanceof LambdaSubstitutionType) {
+            return ((LambdaSubstitutionType) javaType).getOriginal();
+        } else if (javaType instanceof InjectedFieldsType) {
+            return ((InjectedFieldsType) javaType).getOriginal();
+        }
+        return null;
+    }
+
+    protected static ResolvedJavaType getJavaType(HostedMethod hostedMethod, boolean wantOriginal) {
```
loicottet, line 82: I think you can get the same result by looking up `hostedMethod.getDeclaringClass().getJavaClass()`. This will be cleaner than looking into the internals of `HostedType`

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
```diff
@@ -0,0 +1,205 @@
...
+            return lineNumberTable.getLineNumber(bci);
+        }
+        return 0;
+    }
+
+    @SuppressWarnings("try")
+    public void computeFullFilePath() {
+        ResolvedJavaType declaringClass = method.getDeclaringClass();
+        Class<?> clazz = null;
+        if (declaringClass instanceof OriginalClassProvider) {
```
loicottet, line 183: Since you're generating debug info during compilation, `method` should always be a `HostedMerthod`. This should allow you to use `getJavaClass()` directly here to get the class information.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMIRBuilder.java
```diff
@@ -78,6 +98,11 @@ public LLVMIRBuilder(LLVMIRBuilder primary) {
         this.context = primary.context;
         this.builder = LLVM.LLVMCreateBuilderInContext(context);
         this.module = primary.module;
+        if (LLVMOptions.IncludeLLVMSourceDebugInfo.getValue()) {
```
loicottet, line 101: This constructor is only used to create tiny helper methods that represent a single instruction in a main program method (e.g. a compare-and-swap between two objects). Does it make sense to emit debug info here as well? There wouldn't be anything more specific than the debug info attached to the call to this helper method in the main (primary) builder.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMIRBuilder.java
```diff
@@ -62,14 +68,28 @@ public class LLVMIRBuilder implements AutoCloseable {
     private LLVMBuilderRef builder;
     private LLVMModuleRef module;
     private LLVMValueRef function;
+    public LLVMDIBuilderRef diBuilder;
 
+    // Maps filenames to a compile unit
+    public HashMap<String, LLVMMetadataRef> diFilenameToCU;
```
loicottet, line 74: I think LLVM automatically merges identical metadata nodes when it encounters them, if that's the case we can probably get rid of those caches. Also, shouldn't all the debug information from a function have the same file name?

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
```diff
@@ -0,0 +1,205 @@
...
+import org.graalvm.compiler.debug.DebugContext;
+import org.graalvm.compiler.graph.NodeSourcePosition;
+import org.graalvm.nativeimage.ImageSingletons;
+
+import java.nio.file.Path;
+
+import static org.graalvm.compiler.debug.GraalError.unimplemented;
+
+// Similar to NativeImageDebugLineInfo from previous releases
+public class LLVMDebugLineInfo {
```
loicottet, line 53: `NativeImageDebugLineInfo` has been replaced with `NativeImageDebugLocationInfo`. You should base your changes on the newer version to be able to merge the changes to master.

loicottet: Also, is it necessary to reimplement this class? Can't you just reuse it as is?

loicottet (commented):
This looks like cool changes, thanks a lot! Let me know when you've had the time to look into my comments.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImage.java
```diff
@@ -470,7 +470,9 @@ public void build(String imageName, DebugContext debug) {
             if (SubstrateOptions.GenerateDebugInfo.getValue(HostedOptionValues.singleton()) > 0) {
                 Timer timer = TimerCollection.singleton().get(TimerCollection.Registry.DEBUG_INFO);
                 try (Timer.StopTimer t = timer.start()) {
-                    ImageSingletons.add(SourceManager.class, new SourceManager());
+                    if (ImageSingletons.contains(SourceManager.class) == false) {
```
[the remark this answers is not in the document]
adinn: The SourceManager singleton really ought to be registered in NativeImageDebugInfoFeature conditional on debuginfo generation being enabled.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMBackend.java
```diff
@@ -121,6 +125,14 @@ private void emitLLVM(StructuredGraph graph, CompilationResult result) {
             NodeLLVMBuilder nodeBuilder = newNodeLLVMBuilder(graph, generator);
 
             /* LLVM generation */
+            if (LLVMOptions.IncludeLLVMSourceDebugInfo.getValue()) {
+                //TODO: Avoid this lock if possible
+                imageSingletonesLock.lock();
+                if (ImageSingletons.contains(SourceManager.class) == false) {
```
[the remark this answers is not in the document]
adinn: No, DebugInfoFeature is the right place to do this as it is relevant to all debug info generation code not just the LLVM debug info generator.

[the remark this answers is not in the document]
adinn: Yes, `SourceManager` is definitely not thread safe at the moment.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
adinn: There are a great number of other changes that have happened along with that change.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
adinn: The lookup is performed this way in NativeImageDebugInfoProvider for a good reason. There are specific cases related to the use of substitutions where looking into the internals of HostedType is required in order to give a consistent view of the code base to a debugger.

The problematic situation occurs when we have a class C and substitution class S with method m1 defined by C and m2 substituted by S. We can only have one class in the debug info and we need to ensure m1 and m2 are reported as methods of that class. The current code in NativeImageDebugInfoProvider ensures that the class name is always reported as C and that the methods are reported as C.m1 and C.m2.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
adinn: That is true of classes that are presented as the declaring class of a compiled method. It did not turn out to be true of all classes that turned up as declaring classes of some inlined methods that are referenced in frame data derived from the CompilationResult. These inline classes also need to be associated with a source file. 

Note also that in this case the source file for a substituted method S.m1 needs to be the file for class S, not for the class C that it substitutes. That is why the lookup functions above take a flag describing whether or not the original is wanted.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMIRBuilder.java
adinn: I'm not sure what the answer is to that last question. However, I'd just note that a compiled method can include code from inlined methods which belong to a different source file.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMOptions.java
```diff
@@ -49,5 +49,8 @@ public class LLVMOptions {
     @Option(help = "Enable LLVM bitcode optimizations")//
     public static final HostedOptionKey<Boolean> BitcodeOptimizations = new HostedOptionKey<>(false);
 
+    @Option(help = "Include source code level debug info in the output LLVM IR", type = OptionType.Debug)//
+    public static final HostedOptionKey<Boolean> IncludeLLVMSourceDebugInfo = new HostedOptionKey<>(false);
```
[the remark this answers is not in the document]
adinn: Yes, I agree. Enabling debug info generation and selecting the generated target type are (must be) orthogonal choices.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
adinn: Note however, that the source file for m1 needs to be C.java while the source file for m2 needs to be S.java.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImage.java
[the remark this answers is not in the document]
loicottet: That seems to be a cleaner solution indeed.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMBackend.java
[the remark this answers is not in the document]
loicottet: I agree, having all management related to `SourceManager` in `NativeImageDebugInfoFeature` makes sense. The handling of debug information on the LLVM backend has to happen in the compilation phase, so making `SourceManager` can't be avoided, I think. It should be as simple as changing its data structures to thread-safe ones though, so not a big hurdle.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
loicottet: I see, that makes sense. However it would be cleaner to reuse those lookup methods from `NativeImageDebugInfoProvider` instead of copying them here.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
loicottet: That makes sense, I didn't consider inlined methods. As a general question related to that file, how is it different for LLVM? Couldn't we just reuse the one from the normal backend?

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/SubstrateLLVMBackend.java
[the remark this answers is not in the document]
adinn: Well, this is a cache manager so it is not quite so simple as just updating the data structures. The important thing to get right is to ensure that the check as to whether a source file is cached in the sources directory and any subsequent lookup and copy from some originating jar into the sources directory happens *atomically*. A secondary constraint is that a failed lookup must records the lack of any suitable source atomically so that we don't waste time repeating lookups.

Like all concurrent caches there are different choices available to enforce consistency, ranging from the sledgehammer solution of a synchronization on the whole cache at every lookup, through a synchronization on some proxy that exists per cached file, right down to a racy lookup and copy of sources that only synchronizes at the point of linking the copied file into the source tree.

I am happy to review any proposed fix. However, I am also happy to implement a solution, which may be quicker as I am already familiar with the code. @rishikeshdevsot @loicottet please confirm how you would like to proceed.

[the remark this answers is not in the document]
rishikeshdevsot: @adinn I am okay with you implementing a solution since you are familiar with the problems and the code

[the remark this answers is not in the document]
adinn: Ok, I'll prepare that in a separate PR and you can rebase this PR on it.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMOptions.java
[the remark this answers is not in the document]
rishikeshdevsot: Makes sense, I'll remove this option and just use `GenerateDebugInfo`

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/NodeLLVMBuilder.java
rishikeshdevsot: Agreed, I'll remove this check

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMIRBuilder.java
rishikeshdevsot: I see, I removed debug info emission for this constructor

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
rishikeshdevsot: Okay.

I pushed some changes where I attempted to use the standard `NativeImageDebugInfoProvider` backend for the location information. I had to make the class public and add a new subclass `NativeImageDebugLLVMLocationInfo`. 

I had one question regarding `NativeImageBaseMethodInfo` constructor. When I tried using the default constructor, I got a NullPointerException from `isPseudoObjectType()` called from `createParamInfo()` inside the constructor. I am not sure what these functions do, so I don't know how to fix the error for the LLVM backend. I was hoping if someone could let me know what they do. Currently, in the pushed commits, I have a workaround where I created another constructor inside `NativeImageBaseMethodInfo` that does not make these calls.

Let me know if this is similar to what you were expecting.

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMIRBuilder.java
rishikeshdevsot: From my understanding, LLVM automatically merges identical nodes during linking. ("...duplicate information is automatically merged by the linker" https://llvm.org/docs/SourceLevelDebugging.html)  But that means the individual bytecode files corresponding to the functions(e.g. `fXX.bc`) will have duplicate compile units and subprograms without the caches based on how the code is written. So the user will have to deal with duplicate debug info if they wish to perform static analysis on the function level bytecode files. Atleast, for my use case, I would like to perform static analysis on individual bytecode files which is why I added these caches, when I saw that there were multiple compile units present in the `fXX.bc` files. 

If you think that is not a general use case, then I can remove the caches.

adinn:
@rishikeshdevsot PR #5379 which implements a thread safe version of SourceManager is now under review. It would be helpful if you could merge the PR into a copy of your branch before I commit it in order to check that it correctly addresses the problem of multi-threaded access.

rishikeshdevsot:
@adinn I rebased your PR on the current master and then rebased my current changes on top of that. I compiled and ran `mx native-image` with an LLVM Backend for a simple Helloworld java program. I got the following error:
```
Fatal error: com.oracle.svm.core.util.VMError$HostedError: java.lang.NullPointerException
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.util.VMError.shouldNotReachHere(VMError.java:72)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.image.NativeImage.write(NativeImage.java:175)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.image.NativeImageViaCC.write(NativeImageViaCC.java:97)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:719)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:542)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:403)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:580)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:128)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.debugentry.ClassEntry.localFilesIdx(ClassEntry.java:235)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.debugentry.Range.getFileIndex(Range.java:284)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeInlineSubroutine(DwarfInfoSectionImpl.java:1565)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.generateConcreteInlinedMethods(DwarfInfoSectionImpl.java:1047)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeMethodLocation(DwarfInfoSectionImpl.java:1408)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.lambda$writeMethodLocations$11(DwarfInfoSectionImpl.java:1017)
	at java.base/java.util.stream.ReduceOps$1ReducingSink.accept(ReduceOps.java:80)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.reduce(ReferencePipeline.java:563)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeMethodLocations(DwarfInfoSectionImpl.java:1016)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeCompiledClassUnit(DwarfInfoSectionImpl.java:473)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.lambda$writeCompiledClasses$6(DwarfInfoSectionImpl.java:412)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeCompiledClasses(DwarfInfoSectionImpl.java:411)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.generateContent(DwarfInfoSectionImpl.java:157)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.createContent(DwarfInfoSectionImpl.java:90)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfSectionImpl.getOrDecideSize(DwarfSectionImpl.java:608)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.ELFUserDefinedSection.getOrDecideSize(ELFUserDefinedSection.java:105)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.ObjectFile.bake(ObjectFile.java:1625)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.ObjectFile.write(ObjectFile.java:1269)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.image.NativeImage.lambda$write$0(NativeImage.java:171)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.ObjectFile.withDebugContext(ObjectFile.java:1801)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.image.NativeImage.write(NativeImage.java:170
	```
	Rebasing my current changes (just pushed them) onto the master without your changes is able to generate the native image fine(without any errors) and have the relevant source mappings in the llvm bytecode files.

adinn:
@rishikeshdevsot

I'm not at all sure how this has happened. The line where the NPE occurs is the call to `get` in this method of `ClassEntry`:

    public int localFilesIdx(@SuppressWarnings("hiding") FileEntry fileEntry) {
        return localFilesIndex.get(fileEntry);
    }

That can only mean field `localFilesIndex` must be `null`. However, the field is always written in the constructor of ClassEntry:

    public ClassEntry(String className, FileEntry fileEntry, int size) {
        super(className, size);
        this.interfaces = new ArrayList<>();
        . . .
        this.localFilesIndex = new HashMap<>();
        this.localDirs = new ArrayList<>();
        . . .

Has your merge changed any of this code?

substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/util/LLVMDebugLineInfo.java
adinn: @rishikeshdevsot Method `isPseudoObjectType`  checks whether a type presented as a Java object type is derived from class WordBase. ThaIf so that means the Object type used as a proxy for a foreign numeric or pointer value. As far as the type info model is concerned the debugger still needs to present these types as object types. However, as far as the compiled code is concerned they get handled as 64 bit words. That only has influences debug info generation for local vars and values. A parameter or local var of this type will use the object type as its type but will use long as its JavaKind. A constant value for these types will use a long not an oop. This is important when processing frame data that identifies local values and when checking that repeat occurences of vars and values actually alias each other.

I'm not sure what you mean when you say "When I tried using the default constructor" for `NativeImageBaseMethodInfo`. Do you mean you added a constructor which does not receive a `ResolvedJavaMethod`? If so then the resulting `NullPointerException` is only going to be the start of your problems.

If instead you used the existing constructor and it gave you an NPE then it looks like something is very wrong with the `ResolvedJavaMethod` you passed it.

rishikeshdevsot:
@adinn I don't think my code touches any of these parts. I have now pushed all my latest changes merged with the current master to this PR, it is able to build an image and output LLVM bytecode without errors. When I just try your PR without any of my changes, the native image is still built fine and the LLVM bytecode is also output as expected. 
So the cause should be in someway your changes are interacting with my changes. I can try replaying my changes rebased on top of your PR one at a time to see which one leads to the error.

adinn:
@rishikeshdevsot Well looking into your PR it seems neither my patch to make Sourcemanager thread safe nor your PR modifies these lines. So, I have no idea how the NPE you report can possibly happen with this code as it stands.

The above problem represents an issue in the code that models the class base which is almost totally neutral as far as the back end code generator is concerned. However, I am very surprised to see that you are generating DWARF info using the DWARF-specific ELF sections generator code that I wrote for Linux when using the SubstrateVM configured variant of the GraalVM JIT back end when compiling an image using the LLVM back end.

Do you have a good reason to assume that the DWARF it generates will match the compiled program layout employed by LLVM? My assumption would be that this would be unlikely without at least some significant modifications. If your PR does make modifications to either the debug info provider code (DebugInfoProvider, NativeImageDebugInfoProvider and their many related sub-interfaces/classes), the modelling code (classes in package `com.oracle.objectfile.debugentry`) or the ELF section generator code (in package `com.oracle.objectfile.elf.dwarf`) could you perhaps summarize them so I can have some idea where to look for the source of this strange error?

adinn:
@rishikeshdevsot I have looked into your code and I think there is a lot more work to do here. The presence of a significant problem is clearly foreshadowed in your header comment note:

> When using the -O0 flag to generate the native-image with an LLVM backend, I get an error saying "the LLVM backend doesn't support debug info generation" called from inside visitFullInfoPointNode in NodeLLVMBuilder and I am not sure how to avoid this problem. This problem was not present when I implemented this feature in Graal 21.1.0.

There is a good reason why this error is being generated. It is deliberate and reflects the fact that the existing mechanism for generating debug enabled using `-g`  is not compatible with an LLVM generated image.

In order to fix this you need to do two things:
1. Make the LLVM compiler generate debug information that identifies: types and their methods and fields; file and line number info; information about parameter and local var names & types and their live value locations; etc which it will pass on to the ELF image writer for inclusion in the image.
2. Disable or modify the code that is currently used to generate debug information and pass it to the ELF image writer for inclusion in the image.

It looks like your code has done some of the work required for step 1 i.e. it labels individual instructions or instruction sequences with a file, line number and method name. The exception trace for the NPE that you printed above shows that you have not properly addressed step 2.

The above prescription assumes optimistically that LLVM provides a comprehensive API to drive debug info generation *in total*. The pessimistic alternative is that that it does so partially. In that case in order to get a good debugging experience you need to hope that the LLVM API operates in a way that allows the info it does support to be supplemented with whatever info the API does not cater for. You will also need to think about how you are going allow for that extra information to be added.

At this point, it is worth noting that interface `DebugInfoProvider` is the internal API that NativeImage provides for generating debug info *in total* and `NativeImageDebugInfoProvider` is the class that implements that API *in total*. So, if LLVM only provides partial support for generating debug info then you may be able to resue some of this existing implementation. However you may also need to do a lot of work to decouple functionality that fill in the gaps from `DebugInfoProvider` Depending on what LLVM does provide by way of API, this decoupling and reuse may or may not be possible. I am currently not sure how -- or even whether  -- you ought to be using interface `DebugInfoProvider` at all.

Clearly, you will have to do a lot more work to get full debug information into an LLVM image than simply add file, line and method name tags. If there are further hooks to allow this to be passed to LLVM during generation then you may be able to completely disable the existing debug info generation code. If not then you will have to look into the current implementation, understand in exact detail how it works and make changes to accommodate the steps that LLVM will take while somehow filling in the blanks for the steps it does not take. I may be able help you with that latter path or I may be just be able to identify that it is not possible. Before I can provide that help I'll need you to identify and explain a lot more about what LLVM will do to install debug info in the generated image, what it will not do and how easy it is to extend what it does to supplement the info it generates cater for any omissions.

There are several things that are clear from looking at the code and the above exception trace.

1. Your attempt to reuse class `NativeImageDebugLocationInfo` as a super of class `NativeImageDebugLLVMLocationInfo` does not work. It is not simply that this reuse is resulting in the NPE. It is also a bad design that is confusing and likely to be difficult to maintain. You have been forced to add a spurious constructor to `NativeImageDebugInfoProvider` solely in order to be able to create a throwaway instance which you can then use to instantiate a `NativeImageDebugLLVMLocationInfo` instance. The only point of all this is to reuse a small amount of functionality from `NativeImageDebugLocationInfo`, `NativeImageDebugBaseMethodInfo` and `NativeImageDebugFileInfo`.
This is a problem because it is incidental to your use of these types -- and hence to your type -- that they actually serve to implement the corresponding `DebugInfoProvider` interfaces and thereby drive the existing debug info generator implementations. That latter fact implies that changes to the `DebugInfoProvider` interface or implementation which arise during maintenance will likely cause unnecessary and avoidable problems for your implementation of `NativeImageDebugLLVMLocationInfo`.
I think it would be much better for you to factor out the necessary functionality into some auxiliary (static) methods of `NativeImageDebugInfoProvider` which your own class can depend. That avoids your code having to implement the `DebugInfoProvider` interfaces, subclass the associated `NativeImageDebugInfoProvider` inner classes and indirectly access the behaviour through `NativeImageDebugInfoProvider`, avoiding a lot of unnecessary complexity and confusion. Alternatively, you might want to introduce a separate utility class to house this functionality which your class and `NativeImageDebugInfoProvider` can both delegate to.
2. If you want debug info generation to work when using the LLVM back end then you need to ensure that the current generation steps that happen when `-g` is passed either do not happen or are modified to avoid generating info that is inappropriate.
Currently debug info generation occurs in two stages, one before image write and the next during image write. The point where you need to intercept the current implementation is the `beforeImageWrite` callback `NativeImageDebugInfoFeatur` where stage 1 is initiated.
At this point the Provider API is used by the object file code to build a generic model of the debug info content that needs to be included in the image. The object file also creates ELF sections into which the debug info can be written. This is done by creating a instance of `NativeImageDebugInfoProvider` and passing it to `ObjectFile::installDebugInfo`.
If you are relying on LLVM to generate all the debug info then you simply need to bypass these two steps conditional on LLVM being in use. If you can only get LLVM to generate some of the section content and want to try to reuse the existing code to do the rest then you will need to leave this code in place. You may well need to leave it up to LLVM to create some of the target sections and you will likely need to modify the code that builds the model to incorporate information about the debug inof elements that LLVM is managing (e.g., with your current code, files and line numbers).
The second stage of debug info generation happens when the ELF file is being written. The sections created at stage 1 are requested to provide a byte array that contains the section content. Each ELF debug section implementation class iterates over the generic model created at stage 1 and encodes the relevant info into a byte array, whether that is file and line info, string info, type info or whatever. If LLVM generates everything then you don't need to create any of these sections which means the request to encode and return the content will not occur. However, if you need to retain just some of these sections -- because LLVM does not generate this sort of content -- then you will need to modify the methods which encode these retained sections so that the info they include is compatible with/cross-references the info LLVM generates.

Rather than start continuing to implement something along these lines I think you need to investigate what LLVM provides by way of API to support debug info generation, see if it will provide all that is needed and, if not, see whether it can be supplemented by external code. If you report back to me with that I can help you to identify whether any of the existing code can be reused and how best to approach doing that, including what possible refactoring might be appropriate.

rishikeshdevsot:
@adinn Thanks a lot for your insightful comments on my code and sorry for the delay in my response. I can start off by looking at what LLVM provides in terms of debug info generation and let you know.

rishikeshdevsot:
Hi @adinn, 

From my understanding of the source code, the output native image generated with or without the llvm backend being enabled is the same. I think the native image generation code is decoupled from the LLVM IR generation code even when the LLVM backend is enabled. The only thing enabling the LLVM backend does is generate LLVM IR using the GraalVM AST during the compilation phase, the generated IR is then compiled and linked into an object file using LLVM's LLD; but this object file is different from the output native image. I don’t think LLVM takes part in native image generation. The debug info being generation code I wrote is also only being included in the LLVM IR, the debug info being written to the native image should be the same as without the llvm backend flag being enabled. So when you say the following
> There is a good reason why this error is being generated. It is deliberate and reflects the fact that the existing mechanism for generating debug enabled using `-g` is not compatible with an LLVM-generated image.

There is no LLVM-generated image per se. The native image generated when the LLVM backend is enabled is the same as when it is disabled.

Correct me if I am wrong about the above @loicottet 

Nonetheless, I have looked into LLVM API for including debug info, and based on my reading of https://llvm.org/docs/SourceLevelDebugging.html  I think LLVM provides an API for debug info generation in _total_. LLVM provides:

1. A set of intrinsic functions to declare and track source local variables.
2. API calls to set line number, file name, and scope information
3. API calls to declare types, methods, and method parameters
4. API calls to declare global variables
etc

I can get started on including this debug information in the LLVM IR output by Graal but I don't think this would affect the output native image. 

Let me know if I misunderstood something or need to provide more details.

adinn:
Hi @rishikeshdevsot

> The only thing enabling the LLVM backend does is generate LLVM IR using the GraalVM AST during the compilation phase, the generated IR is then compiled and linked into an object file using LLVM's LLD; but this object file is different from the output native image.

> The native image generated when the LLVM backend is enabled is the same as when it is disabled.

I am not clear what these two statements are meant to be suggesting. It would help if you could provide a more detailed explanation of what the LLVM back end is doing.

Maybe you could start by clarifying what input LLVM consumes. What does 'the GraalVM AST' mean? Which input (program code in some high level language? LLVM bitcode?) is this AST derived from?

You go on to say "this object file is different from the output native image". A native image is a self-contained executable (I'm assuming you are not generating a shared library). If the image is really 'no different' when the LLVM back end is used then how does the object code generated by the LLVM back end relate to the executable? Is it a completely separate artifact?

Alternatively, do you just mean that the object code that is derived from Java bytecode and generated by the standard back end is generated in the same way and that enabling the LLVM back end serves to handle some other (non-Java) input, generate extra object code which needs to be linked in to produce the final executable? If so then

- what is this extra input?
- what are the symbols this extra object exports?
- how do those symbols relate to some input to the native image front end?
- how are those symbols connected to the compiled Java code output by the standard back end?

loicottet:
@rishikeshdevsot At the moment the LLVM backend is only replacing the regular backend to emit the code section of the final Native Image executable from Graal's intermediate representation (IR). This is done by transforming this IR into LLVM bitcode and feeding it to the LLVM compiler to generate an object file containing the code of the program, with symbols pointing to the Graal-generated data sections.

This LLVM-generated object file is then linked with the object file generated by Graal, which is identical to what it would be when not using the LLVM backend except it has an empty code section and exports symbols to individual objects in its data sections. These symbols are only used during linking, and do not end up in the final image.

As a consequence, the final executable _looks_ mostly the same as the executable produced by the standard backend, but the code section contents are different, since they were produced by different compilers.

adinn:
@loicottet Thanks for that explanation.

@rishikeshdevsot I believe this explains why the debug info currently generated for Linux does not allow work with an image generated using the LLVM back end.

There are various places in the debug info generation scheme where the generator relies on information generated by the GraalVM compiler (more precisely, the compiler as configured for use with SubstrateVM) during compilation or, in some cases, makes assumptions about the layout and operation of the code that compiler generates. This info and the associated assumptions are closely tied to specific, key decisions taken by the compiler during processing of the IR and generation of the machine code. Those decisions don't just determine where the code starts and ends or which internal code address maps to which source line. They also determine things like the stack frame layouts, inline hierarchies, and local var, inlined parameter or constant locations. n.b. given what Loic states, I don't believe LLVM will take any different decisions regarding data layouts for Java types (otherwise the layout of non-code sections, such as the initial heap, would be affected).

Now, if generation of the code (.text) section is handed over to LLVM then this will have two consequences. Firstly, information needed to generate the debug info may no longer be created and passed to the generator. Secondly, LLVM may choose to lay out the compiled code, stack frames and var locations in ways that do not conform to the assumptions made by the current debug info generator.

There are two ways to deal with this. One way would be to 1) ensure LLVM provides comparable info the the current generator and b) modify the generator to revise its fixed assumptions so that it can accommodate the different decisions made when using LLVM.

The alternative is, as I suggested, to plug in a different debug info generator that understand how LLVM operates and produces the required debug sections using whatever debug info generation API LLVM provides.

Clearly, a major attraction of the first option is that the generator already knows how to produce complete debug information about the relevant Java types. Another less obvious (but, arguably, just as important) attraction is that the generator handles creation and installation of all the different ELF object debug sections needed to store the various different suites of debug information. It provides utility code for writing basic elements of any given section in the expected format (i.e. at the individual record level below that where compiler decisions inform the structure).

That said, it may not be easy to get LLVM to provide all the desired information needed by the current generator to produce the non-type related debug info -- especially as we would prefer it to arrive in much the same format as the GraalVM compiler does. Likewise it may not be easy a) to identify and precisely define the assumptions LLVM makes about code layouts, stack frames, line numbering, inlining, local/param/constant placements etc or b) to adapt the generator to accommodate them.

The attractiveness of the second option really depends on how much support LLVM provides for debug info generation. Clearly, one downside of using an LLVM API would involve redoing the work that the current generator does to write details of which types exist in the system. How much work that is really depends on how much help the LLVM API provides. Likewise, the nature of the API may determine how much work needs to be done in order to support creation and installation of the ELF debug sections and writing of individual records.

Finally, as a middle path, it might, perhaps, be possible to factor out some of the current generator code as one or more libraries and reuse it in a dedicated LLVM generator.

I cannot really help you make a judgement as to which path to follow as I don't know any of the details of the LLVM API. If you want to investigate further and try to decide how to proceed I'll be happy to read any reports you can provide of what you think is involved in pursuing either path (or the middle route) and provide what advice I can.

rishikeshdevsot:
Hi @addin,

I did some investigation and the way LLVM debug info generation works is as follows:

1. The LLVM [[DIBuilder API](https://llvm.org/doxygen/classllvm_1_1DIBuilder.html)] is used to include debug information in the LLVM IR in the form of LLVM IR metadata and [[intrinsic functions](https://llvm.org/docs/SourceLevelDebugging.html#format-common-intrinsics)]. DIBuilder can be used to include type definitions, function type declarations, source mappings as metadata in the LLVM IR. Local variable declaration, live value locations are included as intrinsic functions in the LLVM IR.
2. The same llvm compiler (`llc`) which is currently used to generate the code section of the ELF executable by taking LLVM IR as input, can also be used to perform DWARF emission and generate the debug section when the input LLVM IR has the debug info metadata from step 1.

For example: creating the double type using the DIBuilder API:

`DIType DblTy = DBuilder->createBasicType("double", 64, dwarf::DW_ATE_FLOAT)`

where the arguments are name, size in bits, and encoding (DWARF encoding code). This shows up as metadata inside the LLVM IR as:

`!3 = {.} ; [DW_TAG_ base_type ] [double] [line 0, size 64, align 64, offset
O, enc DW_ATE_float]`

which further gets compiled to dwarf as 

`0X00000097: TAG_base_type [5]
AT name( "double" )
AT_encoding( DW_ATE_float
AT byte size 0x08 )`

Together steps 1 and 2 will drive debug info generation in *total.*

I believe the approach we should take is to include the debug info metadata in the LLVM IR when the `-g` flag is enabled and let the LLVM compiler handle the debug info generation. This has the following advantages:

1. The debug information metadata included in the LLVM IR is designed to be target agnostic and hence supports multiple debug information formats(DWARF/Microsoft Code view)
2. “LLVM debug information always provides information to accurately read the source-level state of the program, regardless of which LLVM optimizations have been run”…” LLVM debug information is automatically optimized along with the rest of the program, using existing facilities. For example, duplicate information is automatically merged by the linker, and unused information is automatically removed.” ([https://llvm.org/docs/SourceLevelDebugging.html#ccxx-frontend](https://llvm.org/docs/SourceLevelDebugging.html#ccxx-frontend)). So if we were to take a different approach and use a separate debug info generator instead of using `llc`  then we would not be able to perform any LLVM IR optimizations(for example inlining, basic block reordering/merging/cleanup, etc) when `-g` is enabled. Or we would have to make the debug info generator aware of any LLVM optimizations being performed.
3. It allows performing static analysis on the LLVM IR which requires mappings to the source code. This is my current use case, I require a way to be able to perform static analysis in the LLVM IR and then be able to map certain LLVM IR instructions back to the Java source which I can do using the debug info metadata in the LLVM IR.

This approach does have the disadvantage that it would involve re-implementing a lot of the existing debug info generator but I think it is still worth the effort. 

 Let me know if there are any questions about anything I’ve mentioned above.

Also, can you point me to where all the debug info type definitions are currently being done for all the Java types?

adinn:
> I believe the approach we should take is to include the debug info metadata in the LLVM IR when the -g flag is enabled and let the LLVM compiler handle the debug info generation

I think that sound like the correct approach.

> This approach does have the disadvantage that it would involve re-implementing a lot of the existing debug info generator but I think it is still worth the effort.

I agree that it is not going to be possible to reuse much of the existing debug info generator code. The bulk of this code is designed to generate complete debug info after compilation has completed based on the specific data currently collected by the compiler as a side-effect of compilation. The LLVM approach appears to require injecting metadata into the LLVM IR as an auxiliary part of the compilation process. However, the current debug info generator code is very tightly designed around consuming the auxiliary (non-IR/non-code) data structures the compiler outputs and transforming them into linker sections organized as DWARF or CodeView debug records. It is hard to see how much of that existing code will be of use when it comes to  modifying the compilation process to inject LLVM metadata whose format is different to and independent from the target debug section formats.

What that implies is that when using the LLVM back end the current debug info generation step needs to be disabled. This is provided via an internal feature (class `NativeImageDebugInfoFeature`) so switching it off merley requires a tweak to method `isInConfiguration`.

After that implementing LLVM debug info generation is going to require by making modifications to `SubstrateLLVMBackend` so it injects debug metadata into the generated LLVM IR. If the information stored in a CompilationResult is not enough to produce all the debug metadata you need then you may well also have to modify some of the classes in package `com.oracle.svm.core.graal.llvm` to ensure it is collected during compilation. That could well involve overriding methods in class `Backend`. I don't know if it will also necessitate making changes in classes that belong to package `org.graalvm.compiler` but I hope that will not be not necessary.

Obviously, you will also need to inform LLVM of details of the Java type model. That will very likely need to be done before you can inject debug metadata into the LLVM IR for some compiled method.

> Also, can you point me to where all the debug info type definitions are currently being done for all the Java types?

I'm not sure exactly what you mean by 'all the Java types' but let me clarify what the current generator caters for. The range of types for which type info is generated is as follows:

1. Java primitive types
2. Special (non-Java )object header type (named _objhdr)
3. Java instance types
4. Java interface types
5. Java array types

The data that describes all these different types is constructed in class `NativeImageDebugInfoProvider` which is an implementation of interface `DebugInfoProvider`. The main API method that retrieves this data is `DebugInfoProvider#typeInfoProvider()`. Most of the type records retrieved via this API method are derived from data attached to the heap (`NativeImageHeap`) and the code cache (`NativeImageCodeCache`). Some of it (primitive type/header) is synthesized from scratch.

Interface `DebugInfoProvider` allows the generator to retrieve this type information plus also info about compiled methods (DebugInfoProvider#codeInfoProvider() and heap data (DebugInfoProvider#dataInfoProvider() which are all used together to encode complete DWARF or CodeView records. This main interface defines a host of nested interfaces, such as DebugTypeInfo, DebugPrimitiveTypeInfo, DebugHeaderTypeInfo etc, which are implemented by `NativeImageDebugInfoProvider` using correspondingly named implementation types, NativeImageDebugPrimitiveTypeInfo, NativeImageDebugHeaderTypeInfo, etc. These nested interfaces declare further methods which expose related content like super info, fields, methods, local var info, etc.

I think this API (suite of interfaces) is probably unlikely to suit your needs and even more so the implementation. The API includes methods that you will not need/want to implement and the current implementation can only be created by passing in data that you may well not have available at the point where you need to communicate the type into to LLVM.

I don't really know how/where you would be able to notify to LLVM of this type information in order to produce complete LLVM debug metadata, although I assume that you will be able to use the builder API to do so. You may be in a position to to achieve this by iterating over the types in the `NativeImageHeap` as is done in `NativeImageDebugInfoProvider`. Alternatively, you may need to reach further back into the structures generated by the points to analysis. I say that because I suspect you will probably need to identify what types will be used in the image before you can inject info into the LLVM IR. That's for the obvious reason that the metadata injected into the LLVM IR will need to refer to types (e.g. to define method owner, param, return and local var types). That might require you to notify type info before a 
`NativeImageHeap` has been created or is available to your code.

rishikeshdevsot:
Hi @adinn @loicottet 

I wanted to update you on my progress and I have pushed the code I have worked on so far. I’ve been able to set subprograms to functions and add line information to instructions inside functions which then gets compiled to DWARF sections by LLVM’s compiler. The debug_line section shows the location information and the debug_info section shows the function subprogram information. I’ve also added support to generate primitive type information by using the NativeImageHeap which is available before LLVM compilation occurs. 
I have refactored the code to avoid the problem adinn highlighted 

> Your attempt to reuse class NativeImageDebugLocationInfo as a super of class NativeImageDebugLLVMLocationInfo does not work.

 by using a helper class.

I’ve used the LLVM DIBuilder API to set subprograms, create types and set debug locations. The API includes debug info metadata in the output LLVM IR. This LLVM IR is then compiled using LLVM’s llc compiler which generates the code section and the debug section.

There are a couple of things I wanted to bring to your attention/want advice on:

1. The llvm-link executable used in the `LLVMNativeImageCodeCache.java` to link all the individual llvm bytecode files into a single bytecode file seems to be of a Release+Asserts build. Currently, there is a problem I am facing with that, the llvm-link has an assertion that ensures that call instructions have debug location when there are present in functions with subprograms assigned to them. But there are several call instructions (mostly from buildStatepointCall) whose nodes don’t have an associated NodeSourcePosition so I cannot get their surrounding functions or their source location. Which means I will not be able to assign these call instructions debug locations. But if the function enclosing this call instruction has a subprogram set to it then llvm-link assertion `inlinable function call in a function with debug info must have a !dbg location` gets triggered. When I disable this assertion and use llvm-link and llc (LLVM’s compiler that generates the code section and the debug section), the compilation is successful and DWARF information is present in the output object file.
2. I am thinking we should switch from using the llvm-link’s Release+Asserts build executable to just the Release executable. This problem was also brought up in the official llvm-project repository and there was a comment about turning this assertion into a warning [https://github.com/llvm/llvm-project/issues/57727#issuecomment-1251596368](https://github.com/llvm/llvm-project/issues/57727#issuecomment-1251596368)
3. I have also tried setting the no-inline attribute to the callsites but that did not help with the problem.
4. Another thing I tried was just to provide a placeholder function name and file name to these call instruction but that triggers another assertion which checks whether information of the surrounding function’s subprogram matches with the debug information in the call instruction.
5. Let me know how I can approach this problem.

I think the remaining parts are to include all type information, live variable locations and verifying all the debug information is accurate.

adinn:
Hi @rishikeshdevsot Thanks for pursuing this. I am still busy with other changes to the debug info generator but I will look into your refactoring of the debug info provider helper code as soon as I have time.

rishikeshdevsot:
Hi @loicottet @adinn 
I believe I have completed almost everything required for this PR. Debug information metadata for types, locations, variables, subprograms, and compile units is included in the LLVM IR, and the LLVM compiler compiles the metadata to DWARF in the final executable. Running dwarfdump on the final executable shows the debug information generated by LLVM. Since my last update, I have made the following changes:

1. Added a workaround that allows the LLVM linker to pass without disabling asserts (continuing from my previous post): In `LLVMGenerator::buildStatepointCall` and `LLVMGenerator::buildStatepointInvoke`, I included generating placeholder debug information (method is the main method and line number is 0) using the `LLVMIRBuilder::setPlaceholderInformation` function. I have also added generating placeholder information when the node source position is null for a ValueNode for which debug information is being added.
2. Generated debug information for function parameters and local variables: For function parameters, I created `alloca` instructions for each parameter as specified by the LLVM documentation and added declarations for the same. The variable declarations are at the end of each block. So when generating instructions, the variable information is obtained using the `getLocalsBySlot` method and stored in a linked list called localVarListPerBlock. When NodeLLVMBuilder::doBlock reaches the end of the block, the declarations for all the variables are added.
3. Generated type information: Types are generated recursively as new type names are encountered when adding LLVM IR instructions or when creating function parameters. Whenever a type name is encountered, the `LLVMIRBuilder::getDiType` function is called on that type name. If debug information for that type has already been generated, then that is returned. Otherwise, the type is looked up in the `NativeImageHeap`, and the hostedType is used to generate debug type information using the `createDebugTypeInfo` function. I have added the class `LLVMImageDebugTypeInfo` inside the `LLVMDebugInfoProvider`, which implements `DebugTypeInfo` similar to the `NativeImageDebugTypeInfo` class and its child classes. In LLVM, the debug type information cannot be shared across modules because each debug type info is associated with an LLVM DIBuilder, which is module-specific. That is why instead of generating all the debug type information at once, I have opted for the recursive approach where type information is only generated as new types are encountered. There can be cases where a class can have members which are of the same type as the enclosing class (e.g., class A { A memberX = new A(); }). To deal with such cases so that we don’t have a stack overflow, I created an array `typesVisitedRecursively` which keeps track of already visited classes.
4. Included location information: The location information and the subprograms for inlined methods are included if the `NodeSourcePosition` contains this information.
5. Disabled the NativeImageDebugInfo feature when the LLVM backend is enabled with debug symbols as adinn mentioned 

> What that implies is that when using the LLVM back end the current debug info generation step needs to be disabled. This is provided via an internal feature (class NativeImageDebugInfoFeature) so switching it off merely requires a tweak to method isInConfiguration.

6. Refactored the code so that the `LLVMDebugInfoProvider` implements the `DebugInfoProvider` interfaces similar to `NativeImageDebugInfoProvider` to make the code more readable. I have also moved functions common to both the DebugInfo providers to `DebugInfoProviderHelper` class.

There were a few things that I am not sure how to implement or deal with:

1. When a type name is encountered which is not present in the NativeImageHeap (E.g. DynamicHub). I am not sure what to do. For now I’ve just generated a basic placeholder type with the same name
2. Obtain enum values: I could not figure out how to obtain the value of an enum using its hostedField. I think I have to use readStorageValue() function but I don’t understand how it is supposed to be used.
3. Obtain array length: I could not figure out how to obtain the number of elements present in array given a hostedType of the array. So for now I have given them a constant value.
4. For cases when the member of the class is of the same type as the enclosing class, I didn’t know what to do. I thought about creating a pointer to the class but that requires the class type to be generated and that in turn requires all the member types to be generated. So for now I have ignored generating a type for such members.

P.S. Do you have any tests for NativeImageDebugInfo that I can use to generate and verify the debug info being generated for llvm backend?

rishikeshdevsot:
Following up on this in case anything else is needed from me to move the PR forward.

Issue #4985, opened 2022-09-16
[GR-41098] HeapDumpWriterImpl retains memory after dump


SergejIsbrecht:
**Describe the issue**
After `kill -s USR1 <pid>` a object `HeapDumpWriterImpl` with a field `fieldsMap` is left behind. This instance is visible in the .hprof file. In my case it consumes up to 32MiB of JavaHeap space. 

**Steps to reproduce the issue**
* build native-image with `-H:+AllowVMInspection` and SerialGC
* run native-image -- I my case with -Xmx256 -Xms256
* `kill -s USR1 <pid>`
* open hprof file with VisualVM and search for HeapDumpWriterImpl object. 

![native_image_heapdump](https://user-images.githubusercontent.com/4427685/190729852-31373c7d-f914-4dd9-9e5f-8b7e3581b450.png)
(it is not possible to provide the actual hprof file, but it should be easy to create a re-producible)

**Describe GraalVM and your environment:**
 - GraalVM version (latest snapshot builds can be found [here](https://github.com/graalvm/graalvm-ce-dev-builds/releases)), or commit id if built from source: **CE 22.2.0**
 - JDK major version: **11**
 - OS: **Linux 4.x**
 - Architecture: **aarch64**

**More details**
As far as I see it the object is referenced via JNI global, which means it will not be GC'ed.

The code is as follows:
```
        byte[] fieldsMapData = heapDumpUtils.getFieldsMap();
        if (fieldsMapData.length == 0) {
            throw new IOException("Empty fieldsMap");
        }
        fieldsMap = createFieldsMap(fieldsMapData);
```

`fieldsMap` is never zeroed out and therefore can not be collected (as far as I see it in the source code). 

** Idea ** 
I would clear the reference after the hprof file was written. After the HeapDump is done, the HashMap could be collected by the GC.

** Note **
Is it really a good idea to collect the data into an object on the Java-Heap? When the native-image is already running out of Java-Heap it would certainly fail, when a heap dump is triggered. Furthermore the instance of `HeapDumpWriterImpl` pollutes every created .hprof file. The measurement (hprof file) in this sense is not valid, because it is hard to reason about it. 
Therefore, I would suggest to somehow allocate memory outside of the Java-Heap and throw away the memory after the heap dump is done.

fniephaus:
Thanks for filling this. See https://github.com/oracle/graal/issues/4637 for the issue on heap dump pollution. We'll look into a fix (unless you'd like to open a PR).

[...]

PR #4994, opened 2022-09-18
[GR-41098] clear allocated memory after heapdump

SergejIsbrecht:
HeapDumpWriterImpl is shown as global JNI reference, which will not be garbage collected (GC root). When a heapdump is written, some fields are set, which contain a lot of memory. Those fields are not null'ed after the .hprof file has been written. This will result in lost memory, which is quite expensive in some cases. The fields can not be GC'ed, because they are referenced from `HeapDumpWriterImpl` and `HeapDumpWriterImpl` is a GC root, which means it will not be GC'ed. This fix nulls set fields after the .hprof file has been written in order to be GC'ed.

Issue-Id: GR-41098
Issue-Id: #4985

## Testing

I did a quick test with master (HEAD) and this branch and it seems the fields set to null will be GC'ed after the .hprof file is created.

### Bad case
```
// stable at ~50MiB heap before heap dump
[Full GC (CollectOnAllocation) 120960K->54400K, 0.0825642 secs]

// stable at ~80MiB after heap dump
[Full GC (CollectOnAllocation) 158522K->84794K, 0.0976779 secs]
```

## Good case (with fix)
```
// stable at ~50MiB before heap dump is triggered
[Full GC (CollectOnAllocation) 133248K->51328K, 0.0907885 secs]

// trigger heap dump
[Full GC (java.lang.System.gc()) 124032K->51328K, 0.0989828 secs]

// after heap dump is triggered
[Full GC (CollectOnAllocation) 120773K->52165K, 0.0863788 secs]
```

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heapdump/HeapDumpWriterImpl.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heapdump/HeapDumpWriterImpl.java
@@ -530,7 +530,10 @@ public class HeapDumpWriterImpl extends HeapDumpWriter {
 
         /* Flush buffer stream and throw it away. */
         out.flush();
+        /* zero out allocated memory in order to be garbage collected */
         out = null;
+        fieldsMap = null;
+        classDataCache = null;
     }
 
     private void writeHeapRecordPrologue() throws IOException {


SergejIsbrecht:
I signed the OCA, which is currently under review.

fniephaus:
Thanks for the PR. Could you please sign the OCA, so that I can integrate this?

SergejIsbrecht:
@fniephaus, I already did, but it seems it is currently under review. I would think some human interaction is needed on the side of oracle to approve it.

fniephaus:
Thanks for raising this issue and contributing. I already had a similar fix in the pipeline (https://github.com/oracle/graal/commit/4e8f80e2b8779d1020eacb27ae79d66de8b7eb92), hope you don't mind that I merged that one instead of yours.

PR #5002, opened 2022-09-19
Remove most truffle dependencies from `com.oracle.svm.graal`

zakkak:
This PR tries to completely remove any dependencies on the truffle-api.jar for running `native-image` without runtime compilation support and truffle features.

It:

* Makes flags that depend on LibGraal optional
* Makes the native-image driver dependency on truffle-api.jar optional
* Removes most truffle dependencies from `com.oracle.svm.graal`
  * The last dependency I have not yet managed to properly remove is https://github.com/oracle/graal/blob/2f9a32bbc9fbe53952fa6e85b55979e3754cc70c/substratevm/src/com.oracle.svm.graal/src/com/oracle/svm/graal/meta/RuntimeCodeInstaller.java#L111 (suggestions are welcome)

### DIFF
--- a/sdk/mx.sdk/mx_sdk_vm_impl.py
+++ b/sdk/mx.sdk/mx_sdk_vm_impl.py
@@ -1308,9 +1308,10 @@ class NativePropertiesBuildTask(mx.ProjectBuildTask):
                 '--no-fallback',
                 '-march=compatibility',  # Target maximum portability of all GraalVM images.
                 '-H:+AssertInitializationSpecifiedForAllClasses',
-                '-H:+EnforceMaxRuntimeCompileMethods',
                 '-Dorg.graalvm.version={}'.format(_suite.release_version()),
             ]
+            if has_component('LibGraal'):
+                build_args += ['-H:+EnforceMaxRuntimeCompileMethods']
             if _debug_images():
                 build_args += ['-ea', '-O0', '-H:+PreserveFramePointer', '-H:-DeleteLocalSymbols']
             if _get_svm_support().generate_debug_info(image_config):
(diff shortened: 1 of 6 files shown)


zakkak:
Hi @fniephaus, 

I am not sure who is the most suitable to review this kind of change. 
Could you please have a look and assign it accordingly?

Thank you

zakkak:
Hi @pejovica, this is kind reminder.

zakkak:
@fniephaus could we please revisit this? 

There have been some efforts to remove/separate Truffle dependencies, but this PR seems still relevant.

christianwimmer:
@zakkak We are currently working on a large refactoring to decouple Truffle from the compiler. The end result of this should be that only the `svm.truffle` project depends on Truffle, and that project is also in a separate module. On the way to this, @olpaw and @chumer might be able to re-use some of the changes from this PR.

zakkak:
Thanks for the update @christianwimmer. 

Do I understand correctly that this will also remove all the truffle related packages from the `jdk.internal.vm.compiler` module (as introduced in https://github.com/oracle/graal/pull/6503)?

Furthermore, is this planned for the 23.1 release?

I have also noticed that https://github.com/oracle/graal/pull/6503 moved

chumer:
The goal is described here:
https://github.com/oracle/graal/issues/6852

So in the future truffle and all languages will be consumed from the application module-path. Except the SVM specific Truffle code which will be for the time being picked up by the native-image driver if finds a truffle runtime on the module path.

The compiler will still depend on truffle-compiler.jar, which is a small set of interfaces shared between Truffle and the compiler. It is not quite minimal yet (it still contains some libgraal specific classes), but we will work on further reducing it.

zakkak:
No longer relevant.

PR #5014, opened 2022-09-20
[GR-40529] [GR-40812] Move tracing data into static state and remove the native-image hosted agent.

graalvmbot:
This really should be two PRs, but these changes are very related to each other so I created one for both to help alleviate CI pressure. If needed, I can split them into two PRs.

This PR is broken into 4 commits:
 - the first 3 move the class initialization and object instantiation data into static state. Previously, this data was saved in an object stored in the `ImageSingletons` - however, we could have cases where these events happen and need to be traced before `ImageSingletons` are available.
Saving this data into static state should be safe - this flag only used for diagnostics.
We now also rightly accept `--trace-class-initialization` specified multiple times on the command line.

- the last commit handles a bug where we could generate a `clinit` for some classes.
This commit completely removes the native-image hosted agent, and instead does all instrumentation inside our JVMTI diagnostics agent. We now pass in a different ASM target version depending on which Java version native-image gets built for


zakkak:
Thank you @gradinac this PR indeed improves tracing of object instantiation (see https://github.com/quarkusio/quarkus/discussions/27739#discussioncomment-3690554 for an example)

PR #4375, opened 2022-03-07
Extract standalone pointsto analysis tool out of native image building process

ziyilin:
### Overview
Pointsto analysis is one of the corner stones in native image generation, and also the foundation of many static analysises. The `com.oracle.graal.pointsto` project is the state-of-the-art of pointsto analysis implementation, but it is embeded in the native image generation process and very difficult to be used by other static analysises. This PR extracts the pointsto project from embeded to a standalone tool, so it can be easily used from command line or API without dependency on svm.jar. 

### Prerequisites
Standalone pointsto is a pure Java project. It requires:

- JDK11
- JVMCI support
- Graal Compiler

### Usage
Standalone pointsto's basic functions dependend on pointsto.jar, native-image-base.jar and some graal compiler jmods. Run it with GraalVM JDK would be the easiest way.
#### From Command Line
Assume the GraalVM JDK home is `$GRAALVM_HOME`,  and there is a `./bin/HelloWorld.class` file under current directory. Using the GraalVM JDK will start the poitnsto analysis and print the analysis results to the `reports` directory:

```shell
$GRAALVM_HOME/bin/java --add-exports=java.base/jdk.internal.module=ALL-UNNAMED  --add-exports=jdk.internal.vm.compiler/org.graalvm.compiler.phases.util=ALL-UNNAMED -cp $GRAALVM_HOME/lib/svm/builder/pointsto.jar:$GRAALVM_HOME/lib/svm/builder/native-image-base.jar com.oracle.graal.pointsto.PointsToAnalyzer -H:AnalysisTargetAppCP=bin -H:+PrintAnalysisCallTree -H:AnalysisEntryClass=HelloWorld
```
Descriptions:

- The JVM option `-cp` sepcifies the analysis framework classpath, **NOT** the analysis tagets.
- `com.oracle.graal.pointsto.PointsToAnalyzer` is the analysis laucher class.
- The `-H:AnalysisTargetAppCP` option specifies where the analysis targets are.
- The `-H:AnalysisEntryClass=HelloWorld` option specify the analysis entrypoint wihch would be HelloWorld class' main method.

#### From API
A more common usage scenario of standalone pointsto would be invoked via API inside user's own program to get pointsto results for further analysis. The following code shows an example of invoking standalone pointsto API and getting the results.
```java
 String[] arguments = new String[]{"-H:AnalysisTargetAppCP=bin", "-H:AnalysisEntryClass=HelloWorld"};
[...]

PR #5024, opened 2022-09-21
Clear options before parsing

ziyilin:
The standalone pointsto analysis can be programmatically invoked multiple times. Each invocation should have its own options which are parsed independently but all invocations can share the same allAnalysisOptions.


cstancu:
Also please add a comment mentioning why you reset the map.

zakkak:

PR #5053, opened 2022-09-23
Add unit tests for standalone pointsto analysis

ziyilin:
Adding unit tests and a tester framework to run standalone pointsto analysis and verifies the results.

### DIFF
--- a/substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/StandaloneAnalyzerReachabilityTest.java
+++ b/substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/StandaloneAnalyzerReachabilityTest.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2022, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2022, Alibaba Group Holding Limited. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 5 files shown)


substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/StandaloneAnalyzerReachabilityTest.java
```diff
@@ -0,0 +1,105 @@
...
+        PointstoAnalyzerTester tester = new PointstoAnalyzerTester(this.getClass());
+        tester.setAnalysisArguments(tester.getTestClassName(),
+                        "-H:AnalysisTargetAppCP=" + tester.getTestClassJar());
+        String classC = C.class.getName();
+        String classC1 = C1.class.getName();
+        String classC2 = C2.class.getName();
+        String classD = D.class.getName();
+        tester.setExpectedReachableTypes(classC, classC1, classC2);
+        tester.setExpectedUnreachableTypes(classD);
+        tester.setExpectedReachableFields(classC + ".val");
```
cstancu, line 92: Instead of using the `String` representation of reachable elements you could pass in the actual `java.lang.reflect.Field` object and query the `com.oracle.graal.pointsto.meta.AnalysisUniverse.fields` map directly. Of course you'd need to first transform the `java.lang.reflect.Field` to a `jdk.vm.ci.meta.ResolvedJavaField` using the `jdk.vm.ci.hotspot.HotSpotMetaAccessProvider.lookupJavaField`. We currently don't have an accessor method `ResolvedJavaField get(JavaField field)` in `AnalysisUniverse` but we could add one. Note that the `AnalysisMetaAccess.lookupJavaField(JavaField)` cannot be used in this case as it has side effects. 

By querying the elements directly you won't have to walk down the element list, which is not a big problem now, but the more test you add the more overhead there will be, especially if we add these tests to the GitHub CI.

Having a handle of the actual `Analysis*` elements will make it easier later to also add tests that check more specific information, e.g., like field type information. Also, using the `java.lang.reflect.*` objects in general will make it easier to evolve these tests, moving and renaming code, while using IDE support. 

This is not a big problem now, but for the long term I think the testing infrastructure needs to be refactored.

ziyilin: Thanks for the suggestion. I'll add a refactoring commit in the PR queue.

zakkak:

PR #5070, opened 2022-09-26
Add StandaloneClassInitializationFeature

ziyilin:
All clinit methods should be taken as root methods for standalone pointsto analysis.
Adding a new StandaloneClassInitializationFeature to register all discovered clinit method during analysis.

### DIFF
--- a/substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/ConstantFieldTest.java
+++ b/substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/ConstantFieldTest.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2022, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2022, Alibaba Group Holding Limited. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.graal.pointsto.standalone.test;
+
+import org.junit.Test;
+
+/**
+ * This test can test 2 facts for standalone pointsto analysis:
+ * <ol>
+ * <li>The clinit method is taken as analysis target instead of being executed to initialize the
+ * class.</li>
+ * <li>The static final field is not taken as final in the analysis, i.e. it is not considered as
+ * constant at analysis time.</li>
+ * </ol>
+ */
+public class ConstantFieldTest {
+
+    @Test
+    public void testConstantField() {
+        PointstoAnalyzerTester tester = new PointstoAnalyzerTester(ConstantFieldCase.class);
+        tester.setAnalysisArguments(tester.getTestClassName(),
+                        "-H:AnalysisTargetAppCP=" + tester.getTestClassJar());
+        tester.setExpectedReachableMethods(ConstantFieldCase.ConstantType.class.getName() + ".foo()",
+                        ConstantFieldCase.class.getName() + ".<clinit>()");
+        tester.setExpectedReachableFields(ConstantFieldCase.class.getName() + ".constantField");
+        tester.runAnalysisAndAssert();
+    }
+}
--- a/substratevm/src/com.oracle.graal.pointsto.standalone/src/com/oracle/graal/pointsto/standalone/heap/StandaloneImageHeapScanner.java
+++ b/substratevm/src/com.oracle.graal.pointsto.standalone/src/com/oracle/graal/pointsto/standalone/heap/StandaloneImageHeapScanner.java
@@ -30,9 +30,13 @@ import com.oracle.graal.pointsto.BigBang;
 import com.oracle.graal.pointsto.ObjectScanningObserver;
 import com.oracle.graal.pointsto.heap.ImageHeap;
 import com.oracle.graal.pointsto.heap.ImageHeapScanner;
+import com.oracle.graal.pointsto.heap.value.ValueSupplier;
+import com.oracle.graal.pointsto.meta.AnalysisField;
 import com.oracle.graal.pointsto.meta.AnalysisMetaAccess;
+import com.oracle.graal.pointsto.meta.UninitializedStaticFieldValueReader;
 import com.oracle.graal.pointsto.util.AnalysisError;
 import jdk.vm.ci.meta.ConstantReflectionProvider;
+import jdk.vm.ci.meta.JavaConstant;
 import org.graalvm.compiler.api.replacements.SnippetReflectionProvider;
 
 public class StandaloneImageHeapScanner extends ImageHeapScanner {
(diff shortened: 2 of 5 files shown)


oubidar-Abderrahim:
Thank you for you contribution, we'll check you PR shortly

oubidar-Abderrahim:
@cstancu could you please review this PR? please let me know if I should mirror it to bitbucket

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/heap/ImageHeapScanner.java
```diff
@@ -464,6 +464,9 @@ protected String formatReason(String message, ScanReason reason) {
     protected ValueSupplier<JavaConstant> readHostedFieldValue(AnalysisField field, JavaConstant receiver) {
         // Wrap the hosted constant into a substrate constant
         JavaConstant value = universe.lookup(hostedConstantReflection.readFieldValue(field.wrapped, receiver));
+        if (value == null) {
+            value = JavaConstant.NULL_POINTER;
```
cstancu, line 468: `null` and `JavaConstant.NULL_POINTER` are not semantically equivalent in this case. If `hostedConstantReflection.readFieldValue()` returns `null` then something is potentially wrong and the caller of this needs to be defensive. The hosted provider in this case is [`HotSpotConstantReflectionProvider`](https://github.com/openjdk/jdk/blob/master/src/jdk.internal.vm.ci/share/classes/jdk.vm.ci.hotspot/src/jdk/vm/ci/hotspot/HotSpotConstantReflectionProvider.java#L164).

substratevm/src/com.oracle.graal.pointsto.standalone/src/com/oracle/graal/pointsto/standalone/features/StandaloneClassInitializationFeature.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class StandaloneClassInitializationFeature implements Feature {
+
+    Set<AnalysisMethod> addedClinits = ConcurrentHashMap.newKeySet();
+
+    @Override
+    public void duringAnalysis(DuringAnalysisAccess a) {
+        StandaloneAnalysisFeatureImpl.DuringAnalysisAccessImpl access = (StandaloneAnalysisFeatureImpl.DuringAnalysisAccessImpl) a;
+        for (AnalysisType type : access.getUniverse().getTypes()) {
+            if (type.isReachable()) {
```
cstancu, line 44: You could avoid using a feature here by registering the initializer as a root method in `com.oracle.graal.pointsto.standalone.StandaloneHost.initializeType()`. This is called from the `com.oracle.graal.pointsto.meta.AnalysisType.initializationTask` when the type is marked as reachable.

substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/ConstantFieldTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public static void main(String[] args) {
+        constantField.foo();
+    }
+
+    static class ConstantType {
+        public void foo() {
+        }
+    }
+
```
cstancu, line 49: I think that it would be better style to separate the tested code from the test code, i.e., everything above this line should be in a different class than the `@Test` method.

substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/ConstantFieldTest.java
```diff
@@ -0,0 +1,58 @@
...
+package com.oracle.graal.pointsto.standalone.test;
+
+import org.junit.Test;
+
+/**
+ * This test can test 2 facts for standalone pointsto analysis:
+ * <li>The clinit method is taken as analysis target instead of being executed to initialize the
+ * class.</li>
+ * <li>The static final field is not taken as final in the analysis, i.e. it is not considered as
+ * constant at analysis time.</li>
```
cstancu, line 36: The comment implies that the test asserts facts about the `<clinit>` but you only test that `foo` is reachable. If that's something that you plan to extend the test with later please keep it separated.

substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/ConstantFieldTest.java
cstancu: This actually goes beyond style: you risk making JUnit reachable in your analyzed code and I guess you don't want that. Please also fix the already added tests (`StandaloneAnalyzerReachabilityTest` and `ClassEqualityTest`).

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/heap/ImageHeapScanner.java
ziyilin: As there is no class initialization at standalone pointsto analysis time, there is no value for static fields. This assignment was to avoid the NPE. You're right, it confuses the `null` and `NULL_POINTER` sematic.
I added `StandaloneImageHeapScanner#isValueAvailable` to override the `ImageHeapScanner#isValueAvailable` to mark the static field's value is not available, so that the value shall never be read.

substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/ConstantFieldTest.java
ziyilin: You are right. The tested code is removed to a new class. `StandaloneAnalyzerReachabilityTest` and `ClassEqualityTest` will be modified in the following test refactoring patch.

substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/ConstantFieldTest.java
ziyilin: Added the assertion for the reachability of `<clinit>`.

cstancu: What about `constantField`? Is it reachable or is it constant folded?

ziyilin: I thought the reachability of `ConstantType#foo` is the proof of the reachability of `constantField`. But it could also be constant folded. 
The assertion of `constantField` is also added.

substratevm/src/com.oracle.graal.pointsto.standalone/src/com/oracle/graal/pointsto/standalone/heap/StandaloneImageHeapScanner.java
```diff
@@ -54,4 +55,11 @@ protected Class<?> getClass(String className) {
             return null;
         }
     }
+
+    @Override
+    public boolean isValueAvailable(AnalysisField field) {
+        // The static field is not available in standalone analysis because the class is not
+        // initialized at analysis time.
+        return !field.isStatic();
```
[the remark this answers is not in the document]
ziyilin: Add the `StandaloneImageHeapScanner#readHostedFieldValue` method to read the uninitialized static value.

ziyilin:
@cstancu Any review updates?

substratevm/src/com.oracle.graal.pointsto.standalone/src/com/oracle/graal/pointsto/standalone/heap/StandaloneImageHeapScanner.java
```diff
@@ -54,4 +58,15 @@ protected Class<?> getClass(String className) {
             return null;
         }
     }
+
+    @Override
+    protected ValueSupplier<JavaConstant> readHostedFieldValue(AnalysisField field, JavaConstant receiver) {
+        ValueSupplier<JavaConstant> ret = super.readHostedFieldValue(field, receiver);
+        if (ret.get() == null) {
+            JavaConstant constant = UninitializedStaticFieldValueReader.readUninitializedStaticValue(field, value -> universe.getSnippetReflection().forObject(value));
+            return ValueSupplier.eagerValue(constant);
```
cstancu, line 67: I think this is fine in the end since the doc on [`ConstantReflectionProvider.readFieldValue`](https://github.com/openjdk/jdk/blob/f43bb9feaa03008bad9708a4d7ed850d2532e102/src/jdk.internal.vm.ci/share/classes/jdk.vm.ci.meta/src/jdk/vm/ci/meta/ConstantReflectionProvider.java#L35) states that `@return the value of this field or {@code null} if the value is not available (e.g., because the field holder is not yet initialized).`, but you should probably me more defensive since `UninitializedStaticFieldValueReader.readUninitializedStaticValue` assumes this is a static field.

cstancu (approved):

substratevm/src/com.oracle.graal.pointsto.standalone/src/com/oracle/graal/pointsto/standalone/heap/StandaloneImageHeapScanner.java
ziyilin: I added a check for static field to avoid the potential assertion failure.

cstancu: This got merged by https://github.com/oracle/graal/commit/5e100c36e3da1797c7df22b6bc40c564864a3ed7 without this check. Please include it in the next PR.

ziyilin: I see. Should I close this PR?

cstancu: Yes, the important parts are already merged.

zakkak:

PR #5093, opened 2022-09-28
Record reflection exceptions in native image agent

lusou-zhangquan:
# Feature description
It helps to record reflection exceptions happened during execution based on native image agent, such as `java.lang.ClassNotFoundExcetion`, `java.lang.NoClassDefFoundError`, etc. And those records will be written to config file in json format together with other config files.

# Background
While conducting pointsto analysis before building native-image , reflection configuration registry is needed for close world. But the key is that we have no idea about whether those exceptions happened during reflection registry are normal.   With the help of these exception configurations recorded during execution on JVM, the uncertainty could be eliminated.

# Implementation
After executing reflection methods such as `Class.forName` in breakpoint, which is designed to record reflection configuration, unhandled exception check will be conducted. If there exists, record them in configuration set and write them to configuration file.

### DIFF
--- a/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
+++ b/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
@@ -206,6 +206,14 @@ final class BreakpointInterceptor {
         }
     }
 
+    private static void traceException(JNIEnvironment env, String className, JNIObjectHandle exception) {
+        JNIObjectHandle exceptionClass = jniFunctions().getGetObjectClass().invoke(env, exception);
+        String exceptionClassName = Support.fromJniString(env, Support.callObjectMethod(env, exceptionClass, agent.handles().javaLangClassGetName));
+        if (tracer != null && exceptionClassName != null) {
+            tracer.traceException(className, exceptionClassName);
+        }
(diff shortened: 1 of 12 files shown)


oubidar-Abderrahim:
Hi, Thank you for contributing to GraalVM, could you please resolve the conflicts on your branch before proceeding to review your PR?
Are there members from the GraalVM team that you discussed those changes with and prefer if they review your PR? if yes, please add them as reviewers

vjovanov:
We had an implementation like this but we decided to take a different route. The reason is that adding exceptions is not really related to the metadata but to the shape of the classpath.

Our new approach is described here: 

https://github.com/orgs/oracle/projects/6

Every exception thrown at runtime can be reconstructed at build time based on the classpath. Therefore, at build time, we simply record the exception that would be thrown if you will reach this element at runtime. The upside is that there are no changes to the metadata required.

vjovanov:
Thank you for the contribution, and apologies for the long delay. We have since taken the implementation in a different direction and do not plan to encode failed reflection lookups in reachability metadata. Metadata should describe the dynamic accesses that must succeed, while lookup failures are handled by Native Image build-time analysis and missing-registration diagnostics.

As a result, this PR no longer aligns with the current design, so we are closing it rather than asking you to resolve the conflicts. We appreciate the work and the discussion it prompted.

Issue #5119, opened 2022-09-30
[GR-41675] libmanagement_ext.a unconditionally being added when libawt_headless.a is linked in


jerboaa:
**Describe the issue**
For an application, which doesn't use class `com.sun.management.internal.OperatingSystemImpl`, but does use `ImageIO` classes for example via libawt_headless, then `libmanagement_ext.a` static library ends up getting linked into the image no matter what.

**Steps to reproduce the issue**
1. `git clone --branch imageio-management-ext https://github.com/jerboaa/mandrel-integration-tests.git`
2. `cd mandrel-integration-tests`
3. `export JAVA_HOME=/path/to/graalvm; export GRAALVM_HOME=$JAVA_HOME; export PATH=$JAVA_HOME/bin:$PATH`
4. `mvn clean verify -Dquarkus.version=2.12.3.Final -Ptestsuite -Dtest=AppReproducersTest#imageioAWTTest`

**Describe GraalVM and your environment:**
 - GraalVM version: built from source, revision 59c0c26add86221fa9e6af577891eb9d9d118c4f
 - JDK major version: **17**
 - OS: **Linux**
 - Architecture: **AMD64**

**More details**
This code seems to add `libmanagement_ext.a` irrespective of it actually needed:

```
    @Override
    public void afterAnalysis(AfterAnalysisAccess access) {
        if (NativeLibrarySupport.singleton().isPreregisteredBuiltinLibrary("awt_headless")) {
            /*
             * Ensure that `management_ext` comes before `awt_headless` on the linker command line.
             * This is necessary to prevent linker errors such as JDK-8264047.
             */
            nativeLibraries.addStaticNonJniLibrary("management_ext", "awt_headless");
        }
    }
```

Originally found here: https://github.com/Karm/mandrel-integration-tests/issues/117

PR #5122, opened 2022-09-30
[GR-41675] Only require libmanagement_ext if it's actually needed.

jerboaa:
On some systems/apps, libawt_headless gets pulled in. This currently brings in libmanagement_ext unconditionally. However, libmanagement_ext should only be present for linking iff
com.sun.management.internal.OperatingSystemImpl class becomes reachable. The fix is to only add `management_ext` as a dependency for `awt_headless` when class `OperatingSystemImpl` is actually reachable.

Closes: #5119

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationManagementExt.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationManagementExt.java
@@ -43,15 +43,17 @@ import com.oracle.svm.util.ReflectionUtil;
 
 @AutomaticallyRegisteredFeature
 public class JNIRegistrationManagementExt extends JNIRegistrationUtil implements InternalFeature {
+
+    private static final String OPERATING_SYSTEM_IMPL = "com.sun.management.internal.OperatingSystemImpl";
     private NativeLibraries nativeLibraries;
 
     @Override
     public void beforeAnalysis(BeforeAnalysisAccess access) {
         nativeLibraries = ((BeforeAnalysisAccessImpl) access).getNativeLibraries();
 
-        rerunClassInit(access, "com.sun.management.internal.OperatingSystemImpl");
+        rerunClassInit(access, OPERATING_SYSTEM_IMPL);
 
-        access.registerReachabilityHandler(this::linkManagementExt, clazz(access, "com.sun.management.internal.OperatingSystemImpl"));
+        access.registerReachabilityHandler(this::linkManagementExt, clazz(access, OPERATING_SYSTEM_IMPL));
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationManagementExt.java
```diff
@@ -77,7 +79,7 @@ private void linkManagementExt(@SuppressWarnings("unused") DuringAnalysisAccess
 
     @Override
     public void afterAnalysis(AfterAnalysisAccess access) {
-        if (NativeLibrarySupport.singleton().isPreregisteredBuiltinLibrary("awt_headless")) {
+        if (managementExtNeeded && NativeLibrarySupport.singleton().isPreregisteredBuiltinLibrary("awt_headless")) {
```
fniephaus, line 82: Instead of using a field, we can't check somehow that "management_ext" is already added?

jerboaa: OK. Done now. It requires iterating over the internal list of static libraries, though (using `NativeLibraries`).

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationManagementExt.java
```diff
@@ -77,7 +77,8 @@ private void linkManagementExt(@SuppressWarnings("unused") DuringAnalysisAccess
 
     @Override
     public void afterAnalysis(AfterAnalysisAccess access) {
-        if (NativeLibrarySupport.singleton().isPreregisteredBuiltinLibrary("awt_headless")) {
+        boolean managementExtNeeded = nativeLibraries.getStaticLibraries().stream().anyMatch(l -> l.equals(nativeLibraries.getStaticLibraryPath("management_ext")));
```
pejovica, line 80: I think it would be simpler to just check if the `com.sun.management.internal.OperatingSystemImpl` class is reachable.
```suggestion
        boolean managementExtNeeded = access.isReachable(clazz(access, "com.sun.management.internal.OperatingSystemImpl"));
```

jerboaa: Thanks, done in the latest version.

jerboaa:
Thanks for the reviews!

jerboaa:
If somebody could help with getting it integrated, I'd appreciate it.

jerboaa:
@fniephaus Could you please help getting this fix integrated? Thank you!

fniephaus:
Sure, but I'm afraid this won't make it into 22.3.

fniephaus:
You got it!

fniephaus:
Internal PR is at https://github.com/oracle/graal/pull/5195.

PR #5184, opened 2022-10-07
WiP: advanced debug interface based on JNI.

peter-hofer:


peter-hofer:
See `JniDebugHelper` code for current `TODO`s -- there are quite a few.

Example application and usage from GDB as follows.
1. Build using `native-image -g -O0 -H:+IncludeDebugHelperMethods -cp . HelloDebug`
2. Run with `gdb -x gdb-commands ./hellodebug`, or just `gdb ./hellodebug` and paste individual commands.

```java
class HelloDebug {

    String[] elements = { "water", "air", "earth", "fire"};

    for (String elem : elements) {
      System.out.println("Hello " + elem);
    }
  }
}
```

GDB command file with explanations:
```
# Set a breakpoint in PrintStream to intercept System.out.println(), before this line: writeln(String.valueOf(x));
break PrintStream.java:1028

# Run to the breakpoint
run

# Initialize breakpoint environment
call (void)svm_jnidbg_begin_break()

# Create JNI object handle for 'x', the argument passed to println()
set $strh=svm_jnidbg_ref_to_handle(x)

# Create JNI handle for 'x.value', the byte array storing the characters of 'x'
set $arrh=svm_jnidbg_ref_to_handle(x.value)

# Enter the JNI context, which returns the JNIEnv* for calling JNI functions
set $env=svm_jnidbg_enter_jni_context($rsp, $rip)

# Call JNI GetObjectClass on 'x', giving us a JNI handle for 'x.getClass()', which is 'java.lang.String.class'
set $strclassh=((JNIEnv *)$env)->GetObjectClass($env, $strh)

# Call JNI AllocObject on the handle for 'java.lang.String.class', giving us a new String object on which no constructor has been called
set $newstrh=((JNIEnv *)$env)->AllocObject($env, $strclassh)

# Call JNI NewStringUTF to create a String object with contents "UTF-8", and return a handle
set $utf8h=((JNIEnv *)$env)->NewStringUTF($env, "UTF-8")

# Leave the JNI environment, going back to "Java" mode
call (void)svm_jnidbg_leave_jni_context($env)

# Call a constructor on the String object that we newly created via JNI above, using the handles we created earlier.
# We need to use a helper method because the native calling convention differs from the internal Java method calling convention.
# It is crucial to always use handles and resolve them to references (addresses) only just before they are used, because otherwise, a garbage collection can run in between that leads to an object getting moved to a different address. Strictly, even the following leads to caching references across multiple calls of svm_jnidbg_handle_to_ref(), but that is supported.
call (void)svm_jnidbg_invokevoid_word_word_int_int_word('java.lang.String::String(byte[]*, int, int, java.lang.String*)', \
    svm_jnidbg_handle_to_ref($newstrh), \
    svm_jnidbg_handle_to_ref($arrh), \
    6, \
    3, \
    svm_jnidbg_handle_to_ref($utf8h))

# Set the local variable 'x' so that it refers to the newly created object instead of its original value
set x=svm_jnidbg_handle_to_ref($newstrh)

# Discard the breakpoint environment. This invalidates all created handles so that objects which are not otherwise reachable can be garbage-collected.
# For handles which should remain alive and/or be valid in threads other than the current one, use the JNI NewGlobalRef function to create handles and call DeleteGlobalRef to release the handle when it is no longer needed.
call (void)svm_jnidbg_end_break()

# Disable the breakpoint, continue printing and executing the rest of the program
disable 1
continue
```

Issue #5251, opened 2022-10-19
Make all Class objects available to debuggers.


vvlevchenko:
**make reachable classes accessible via JNI FindClass for `-g` mode**

Please include the following information:

as discussed on meeting with @fniephaus and @olpaw for expression evaluation in debugger we need get JNI reference for class 

**Describe who do you think will benefit the most.**
Debuggers might need this feature 

**Describe alternatives you've considered.**
[alternatively currently used function for prototyping](https://github.com/vvlevchenko/graal/blob/minamoto/ide-support/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/DebugHelper.java#L580)

peter-hofer:
Note that `jclass` in JNI is merely an object handle for a `Class<?>` object, which already exist for each class.

I think adding classes to the dictionary used by `FindClass` is not the way to go because it can change the observable behavior of the application and can mask problems with the JNI configuration. Using `Heap.getLoadedClasses()` is also not great because it must do a complete heap walk to find all `Class<?>` objects in some safe context first.

I think the best solution is to enrich the type information in the debug data with the offset of the corresponding `Class<?>` object in the image heap (if it's not already there). This does not affect the rest of the application, and, when the debug information is kept in a separate file, does not affect image size either.

adinn:
@peter-hofer @olpaw @fniephaus

> I think the best solution is to enrich the type information in the debug data with the offset of the corresponding Class<?> object in the image heap (if it's not already there)

It is relatively easy to identify the location of the class instance and pas it through to the debug info generator. However, this is not already in the Linux debug info for a good reason. There is no standard way to encode the location of a class's 'Class' instance in DWARF info because the idea of having such a metadata object modeled in the language is a Java-specific notion.

We could, of course, easily model a class pointer as a static member of the associated class e.g. make it look like as if the class included this definition:

```
class org.my.Foo : public Object {
  final public static java.lang.Class *'Foo.class' = org.my.Foo.class;
  . . .
}
```

Note that this would mean that in gdb you would reference the class as follows:

```
(gdb) p/x org.my.Foo::Foo.class
```

We might also want to expose the relevant name as a local linker symbol (indeed I think we could probably usefully make all static field names visible as local linker symbols).

Whether this is the right or best way to do this is a different question. I will consult with my colleagues to see if I can find out how this was handled with gcj.

adinn:
@stooke Do you know whether/how this is handled in the Microsoft debugger?

[...]

PR #5312, opened 2022-10-27
[GR-42601] Model class constants in debug info

adinn:
This PR exposes the location of java.lang.Class (aka DynamicHub) instances located in the image heap to debug info generators, allowing them to be encoded into debug info using the standard class name as an identifier for the corresponding object.

It includes a modification to the Linux debug info generator that encodes details of the class constant object in the info section as a DWARF constant DIE. The implementation enables reference to the class constants using symbolic names like

```
  hello.Hello.class
  java.util.List.class
  java.long.String[].class
  long.class
  byte[].class
```

The type of all the above constants is `class java.lang.Class` (i.e. they identify the underlying object layout rather than an oop).

These symbols are not currently entered into the symbol table as local symbols. That might be desirable for a follow up PR.

### DIFF
--- a/docs/reference-manual/native-image/DebugInfo.md
+++ b/docs/reference-manual/native-image/DebugInfo.md
@@ -140,6 +140,7 @@ The currently supported features include:
   - access through object networks via path expressions
   - reference by name to methods and static field data
   - reference by name to values bound to parameter and local vars
+  - reference by name to class constants
 
 Note that single stepping within a compiled method includes file and line number info for inlined code, including inlined GraalVM methods.
 So, GDB may switch files even though you are still in the same compiled method.
@@ -392,6 +393,46 @@ end
 
 (gdb) hubname $1
 0x8779c8:	"[Ljava.lang.String;"
+```
+
+The native image heap contains a unique hub object (i.e. instance of
+`java.lang.Class`) for every Java type that is included in the
+image. It is possible to refer to these class constants using the
+standard Java class literal syntax:
+
+```
+(gdb) print 'Hello.class'
+$6 = {
+  <java.lang.Object> = {
+    <_objhdr> = {
+      hub = 0xaabd00,
+      idHash = 1589947226
+    }, <No data fields>}, 
+  members of java.lang.Class:
+  typeCheckStart = 13,
+  name = 0xbd57f0,
+  ...
+```
+
+Unfortunately it is necessary to quote the class constant literal to
+avoid gdb interpreting the embedded `.` character as a field access.
+
+Note that the type of a class constant literal is `java.lang.Class`
+rather than `java.lang.Class *`.
+
+Class constants exist for Java instance classes, interfaces, array
+classes and arrays, including primitive arrays:
+
+```
+(gdb)  print 'java.util.List.class'.name
+$7 = (java.lang.String *) 0xb1f698
+(gdb) print 'java.lang.String[].class'.name->value->data
+$8 = 0x8e6d78 "[Ljava.lang.String;"
+(gdb) print 'long.class'.name->value->data
+$9 = 0xc87b78 "long"
+(gdb) x/s  'byte[].class'.name->value->data
+0x925a00:	"[B"
+(gdb) 
 ```
 
 Interface layouts are modelled as C++ union types.
--- a/substratevm/mx.substratevm/testhello.py
+++ b/substratevm/mx.substratevm/testhello.py
@@ -179,6 +179,8 @@ def test():
     execute("set print pretty on")
     # enable demangling of symbols in assembly code listings
     execute("set print asm-demangle on")
+    # disable printing of address symbols
+    execute("set print symbol off")
 
     # Print DefaultGreeter and check the modifiers of its methods and fields
     exec_string = execute("ptype 'hello.Hello$DefaultGreeter'")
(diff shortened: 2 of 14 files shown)


adinn:
This is being proposed as a solution for issue #5251.

fniephaus:
Thanks @adinn! This LGTM but I'll let the others comment on the implementation.

BTW: Running `mx checkstyle --primary` in `substratevm/` before pushing has saved me a lot of time fighting with the CI 😅.

zakkak (approved):
LGTM, thanks Andrew!

zakkak:
> BTW: Running `mx checkstyle --primary` in `substratevm/` before pushing has saved me a lot of time fighting with the CI sweat_smile.

Sometimes even that is not enough and I need to run the following as well:

```
mx --primary-suite-path substratevm --J @-Xmx2g eclipseformat -e ~/.mx/eclipse/eclipse --primary
```

adinn:
@fniephaus @zakkak I ran both checkstyle and eclipseformat but the local runs don't always show all the errors that turn up in the gate checks.

I have pushed further style fixes. Let's hope this one catches everything.

fniephaus:
> I have pushed further style fixes. Let's hope this one catches everything.

Thanks! Looks like this has worked :)

docs/reference-manual/native-image/DebugInfo.md
```diff
@@ -394,6 +395,46 @@ end
...
+avoid gdb interpreting the embedded `.` character as a field access.
+
+Note that the type of a class constant literal is `java.lang.Class`
+rather than `java.lang.Class *`.
+
+Class constants exist for Java instance classes, interfaces, array
+classes and arrays, including primitive arrays:
+
+```
+(gdb)  print 'java.util.List.class'.name
```
olpaw, line 427: @adinn, an alternative approach would be to add an artificial static `class` field to each type so that the hub of a given type can be accessed via
```
p *'java.util.List'::class
```
wouldn't that fit better to the existing code C++ based code model that we have?

peter-hofer: I believe the field would better be called `.class` so that it's not a valid Java bytecode field name, if DWARF permits that and it doesn't make problems with GDB, that is:
`'java.util.List'::.class`

adinn: > @adinn, an alternative approach would be to add an artificial static class-field to each type so that the hub of a given type can be accessed via
> ```
> p *'java.util.List'::class
> ```
> wouldn't that fit better to the existing C++ based code model that we have?

@olpaw That is indeed a nicer fit. The only problem with it is that it doesn't work ;-).

The only real option we have to make this data value visible via debug info is to encode it as a variable/constant, whether at top level or as a (static) member of some class. Any corresponding info definition must identify where the variable is located in memory. So, a static field info definition for a pseudo-field as described above would need specify as location the address of a memory slot whose content is an `oop`, a reference to the `Class` instance. We can make that reference a direct or indirect( (encoded) address by using a suitable location expression. However, we cannot get rid of the need for there to be an indirection.

Unfortunately, the only thing we have available to encode as the location is the address of the `Class` instance. *That* is the reason why I have had to declare the type of constant `java.util.List.class` as `java.lang.Class` not `java.lang.Class *`. It's also why when using isolates the type remains `java.lang.Class` not `_z_.java.lang.Class` or `_z_.java.lang.Class *`

That said, it is still possible to embed this constant as a field of the class. The required change is simply to generate the declaration and definition with the scope of the class and adjust the name and linkage name. The inescapable constraint we are under is that the field has to be an object value not an object reference.

So, we can support either:
```
(gdb) p 'java.util.List'::class
```
or
```
(gdb) p 'java.util.List'::.class
```

n.b. the reason I chose to make the constant a top level value rather than a value embedded in the class was:

1. the syntax and use of name `java.util.List.class` is very much like that of its Java original i.e. it functions as a global name for the class (modulo the associated network of class loader spaces within which it is visible)
2. in Java the class object is not actually a field of the class so presenting it as such appeared to me to risk engendering unnecessary confusion

I'm very happy to change this if you still want me to after taking the above explanation into account.

@peter-hofer yYour worry about `Foo::class` clashing with a legitimate field name is unnecessary. `class` is a reserved keyword so it cannot be used as a field name.

peter-hofer: @adinn `class` is a keyword in the Java _language_, but I would be surprised if it's explicitly forbidden in Java bytecode and can therefore not be used by other languages and especially bytecode generating tools.

adinn: @peter-hofer Oh yes, of course `class` could well be inserted into synthetic or transformed bytecode as a class pool constant. I'd discounted that as something to worry about but now that you mention it I suppose there is quite probably someone somewhere crazy enough to have decided to use it as as field name.

Mind you, the same concern would apply for `.class`, perhaps yet more so in that it is not a keyword and therefore more attractive to less crazy people ;-).

I actually think this is all the more reason to keep encoding the class constants as a top level name. All other Java names are going to exist within the scope of a class (which effectively means a namespace). I think it is relatively unlikely that something like `org.my.Foo.class` is going to turn up as a name for a variable in non-Java code. Of course, that is asking to be proven wrong ...

peter-hofer: @adinn `.class` is not a valid identifier in bytecode because `.` is not allowed in identifiers (JVMS 4.2.2), hopefully that rule is enforced too. `org.my.Foo.class` could again refer to a qualified class or interface named `class` in package `org.my.Foo` -- hopefully it never does, but it's not prohibited.

olpaw: @peter-hofer is right, it's trivial to end up with names that are reserved keywords in Java with bytecode producers other than `javac`. E.g.
```scala
object Hello {
    val `class`: String = "WHY NOT?"
    def main(args: Array[String]) = {
        println(`class`)
    }
}
```
Unfortunately I'm not sure about `.class` either, because the following also works just fine:
```scala
object Hello {
    val `.class`: String = "WHY NOT?"
    def main(args: Array[String]) = {
        println(`.class`)
    }
}
```
But it could be that here Scala applies some magic to make this possible.

peter-hofer: @olpaw, Scala seems to do just that because the field is actually named `$u002Eclass` when using reflection.

peter-hofer:
```
object Hello {
  val `class`: String = "OH NO";
  val `.class`: String = "WHY NOT?";

  def main(args: Array[String]) = {
    for (f <- getClass.getDeclaredFields) {
      println(f.getName())
    }
  }
}
```
prints
```
class
$u002Eclass
MODULE$
```

adinn:
@olpaw I have fixed up the commit messages and squashed the style fixes.
@peter-hofer  @vvlevchenko Just waiting for your decision on whether to continue with the current use of a top level var `org.my.Foo.class` var or switch to a static field `org.my.Foo::class`?

peter-hofer:
@adinn, could the top-level variable alias with anything else (a type)? If not, the current solution seems fine to me.

adinn:
> @adinn, could the top-level variable alias with anything else (a type)? If not, the current solution seems fine to me.

@peter-hofer Yes, there is potential for such a name clash in the scenario you described i.e. we have a Java class `org.my.Foo` and also encounter a bytecode file crafted to define a class named `class` in package `org.my.Foo`. Then with my proposal `org.my.Foo.class` would be the name for both the class constant and the hand-crafted object type. If we adopt your suggestion then there is no possibility of a clash.

adinn:
@peter-hofer Hmm, I was looking into what code changes would be needed to switch to implementing class constants using static field syntax (e.g. `'org.my.Foo':.class`) and ran into a problem. There is no direct equivalent for primitive type class constants because we don't (can't) generate a corresponding info class DIE. The names for the Java primitive types must used to declare primitive types so that they can then be referenced from class definitions. So, we cannot also have a class with that same name.

We might be able to get round this by embedding the class constant for primitive type in a namespace e.g. for `int.class` we create a namespace called `int` and declare a variable called `.class` in that namespace. I'm not sure whether that is going to create a problem with `int` doing double duty as the identifier for a primitive type and a namespace.

adinn:
@olpaw ^^^ see above comment

olpaw (approved):

olpaw:
Approved. Sorry for the late response!

peter-hofer:
Perhaps there is GDB syntax to distinguish a global variable `org.my.Foo.class` from a type of the same name, too. I do hope that this never occurs in practice anyway though.

olpaw:
Running in our CI now

olpaw:
Merged as https://github.com/oracle/graal/pull/5478

PR #5313, opened 2022-10-27
[GR-41921] Centralize and reduce usage of HotSpot JVMCI internals.

graalvmbot:


substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/infrastructure/OriginalFieldProvider.java
```diff
@@ -35,16 +35,8 @@ public interface OriginalFieldProvider {
...
-        Class<?> declaringClass = OriginalClassProvider.getJavaClass(reflectionProvider, field.getDeclaringClass());
-        try {
-            return declaringClass.getDeclaredField(field.getName());
-        } catch (Throwable e) {
-            /*
-             * Return null if there is some incomplete classpath issue or the field is either
-             * missing or hidden from reflection.
-             */
-            return null;
+        } else {
```
DanelRod, line 38: Do You need here now else statement, just after if statement 
```
    if(){
    ...
    return ... 
    }
 return reflectionProvider.originalField(field);
```

christianwimmer: In this case I find the code easier to read with the `else` statement.

Issue #5326, opened 2022-10-31
Upgrade native image debug info class SourceManager implementation to be thread-safe


adinn:
## Issue
The LLVM debug info implementation will benefit from re-use of the API on class SourceManager to locate and cache source files for a generated native image. However, as explained in PR #4964 this will require the API to be usable from concurrently executing threads, which is not currently the case. The implementation should be upgraded to make it thread safe.

# Preferred Solution
Ideally it should be possible to execute independent source file lookups concurrently. However, it is also important to ensure that concurrent threads do not perform unnecessary work trying to lookup and cache the same source file. A sensible compromise is to provide exclusion at the level of each individual target cached file, allowing requests for different target sources to proceed concurrently.

Class SourceManager retains the results of previous lookups in a lookup table. So, normally, repeat lookups find a cached file and complete very quickly. However, there can still be a significant exclusion window for concurrent lookups when a thread first has to copy a file into the cache. If lookups for the same file tend to arrive as a group then this might have an observable impact on progress of the compilation phase (for example, independent, compilations might be initiated concurrently for methods of the same class). This could be resolved by posting the target cached file details and closing the exclusion window before actually copying the file. However, that optimization should only be attempted if copying is observed to slow down compilation.

PR #5379, opened 2022-11-07
Make SourceManager thread safe

adinn:
Proposed implementation for issue #5326

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
@@ -350,10 +350,10 @@ class NativeImageDebugInfoProvider implements DebugInfoProvider {
             SourceManager sourceManager = ImageSingletons.lookup(SourceManager.class);
             try (DebugContext.Scope s = debugContext.scope("DebugFileInfo", hostedType)) {
                 Path filePath = sourceManager.findAndCacheSource(javaType, clazz, debugContext);
-                if (filePath == null && hostedType instanceof HostedInstanceClass) {
+                if (filePath == null && (hostedType.isInstanceClass() || hostedType.isInterface())) {
                     // conjure up an appropriate, unique file name to keep tools happy
                     // even though we cannot find a corresponding source
-                    filePath = fullFilePathFromClassName((HostedInstanceClass) hostedType);
+                    filePath = fullFilePathFromClassName(hostedType);
                 }
                 fullFilePath = filePath;
             } catch (Throwable e) {
@@ -425,8 +425,8 @@ class NativeImageDebugInfoProvider implements DebugInfoProvider {
         }
     }
 
-    private static Path fullFilePathFromClassName(HostedInstanceClass hostedInstanceClass) {
-        String[] elements = hostedInstanceClass.toJavaName().split("\\.");
+    private static Path fullFilePathFromClassName(HostedType hostedClass) {
+        String[] elements = hostedClass.toJavaName().split("\\.");
         int count = elements.length;
         String name = elements[count - 1];
         while (name.startsWith("$")) {
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/Source.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/Source.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) 2022, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2022, Red Hat Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.hosted.image.sources;
+
+import java.nio.file.Path;
+
+/**
+ * A proxy for a potentially cached file, recording progress of any attempted caching operation and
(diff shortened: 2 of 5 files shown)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceCache.java
```diff
@@ -331,112 +310,64 @@ protected Path tryCacheFile(Path filePath, Class<?> clazz) {
...
             try {
-                if (tryCheckCacheFile(sourcePath, targetPath)) {
-                    return filePath;
-                }
+                // only copy if the size is different or the target is older than the source
+                FileTime sourceTime = Files.getLastModifiedTime(sourcePath);
+                FileTime destTime = Files.getLastModifiedTime(targetPath);
+                long sourceSize = Files.size(sourcePath);
+                long destSize = Files.size(targetPath);
+                doCopy = sourceSize != destSize || destTime.compareTo(sourceTime) < 0;
```
zakkak, line 347: What if the size is different but the target file is newer than the source? Shouldn't we through an error in that case or at least keep the cached one?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceCache.java
```diff
@@ -331,112 +310,64 @@ protected Path tryCacheFile(Path filePath, Class<?> clazz) {
...
             } catch (IOException e) {
-                /* delete the target file as it is invalid */
-                targetPath.toFile().delete();
-                /* have another go at caching it */
-                return tryCacheFile(filePath, clazz);
+                // copy anyway just in case
+                System.out.println("IOException " + e);
+                e.printStackTrace();
```
zakkak, line 351: I think this should be a warning.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceCache.java
```diff
@@ -331,112 +310,64 @@ protected Path tryCacheFile(Path filePath, Class<?> clazz) {
...
-            FileTime sourceTime = Files.getLastModifiedTime(sourcePath);
-            FileTime destTime = Files.getLastModifiedTime(targetPath);
-            if (destTime.compareTo(sourceTime) < 0) {
+        if (doCopy) {
+            try {
                 Files.copy(sourcePath, targetPath, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.COPY_ATTRIBUTES);
+            } catch (IOException e) {
+                // cached file may be invalid so remove it
+                cachedFile.delete();
+                source.updateStatus(CacheStatus.MISSING);
```
zakkak, line 366: Consider also printing a warning in this case.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/Source.java
```diff
@@ -0,0 +1,143 @@
...
+     * into the cache. It can also upgrade the status from CACHING to MISSING if the file cannot be
+     * found or downgrade it from LOCATED to MISSING if an attempt to copy a located source into the
+     * cache fails.
+     *
+     * @return True if the calling thread is expected to try to cache the source otherwise false.
+     */
+    synchronized boolean shouldCache() {
+        if (isUnknown()) {
+            // first caller to reach here can promote status to CACHING
+            // and return to try to popuate the cache
```
zakkak, line 128:
```suggestion
            // and return to try to populate the cache
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -425,8 +425,8 @@ public Path cachePath() {
         }
     }
 
-    private static Path fullFilePathFromClassName(HostedInstanceClass hostedInstanceClass) {
-        String[] elements = hostedInstanceClass.toJavaName().split("\\.");
+    private static Path fullFilePathFromClassName(HostedType hostedClass) {
+        String[] elements = hostedClass.toJavaName().split("\\.");
```
zakkak, line 429: It's not clear to me how this change relates to making the SourceManager thread safe, could you please explain?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/Source.java
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * The caller that attempts the cache operation performs all other transitions. It upgrades the
+     * status to LOCATED when it finds a matching source and then to CACHED when this file is copied
+     * into the cache. It can also upgrade the status from CACHING to MISSING if the file cannot be
+     * found or downgrade it from LOCATED to MISSING if an attempt to copy a located source into the
+     * cache fails.
+     *
+     * @return True if the calling thread is expected to try to cache the source otherwise false.
+     */
+    synchronized boolean shouldCache() {
```
zakkak, line 125: The name `shouldCache` implies that this method does far less than what it actually does. Unfortunately, I can't think of any better one. Maybe `shouldCacheOrBlock`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceManager.java
```diff
@@ -53,39 +51,35 @@ public class SourceManager {
...
+            cache.resolve(source, moduleName);
+
+            assert !source.isCaching();
         }
-        /* memoize the lookup */
-        verifiedPaths.put(resolvedType, (path != null ? path : INVALID_PATH));
 
-        return path;
+        /* cache attempt has been tried - return any associated path */
+        return source.getPath();
```
zakkak, line 82: What's the point of blocking before returning this value? Even if the source is still in the "CACHING" state does it really matter?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceCache.java
adinn: Not necessarily. Firstly, this is all a bit of a heuristic. The way to properly ensure you get the right files is to nuke the cache.

A change of size is the most reliable indicator that we have a different jar but it is not fool proof. The time stamp check fixes some cases but it is not really that reliable. Imagine two different versions of a file Foo.java are in foo-1.0.jar and foo-1.1.jar but happen to have the same size. Consider what might happen if the user upgrades version from 1.0 to 1.1.

- If both versions are downloaded from Maven and were minted before the first cache operation took place then the cache copy of Foo.java will be newer than both of the jar versions (the copy does not preserve the original time stamps) so we will not recache -- bzzzt!
- If foo-1.1.jar has been released since we last cached then its Foo.java will be newer so we recache - ding!
- If foo-1.0 was built locally, Foo.java was cached and then a local foo-1.0 was built then we recache -- ding!
- If foo-1.1 was built locally and just happened to be built after the local foo-1.0 then we don't recache -- bzzzt!

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceCache.java
adinn: This was supposed to be removed. It was only there for debugging. We could possibly generate a (verbose level) log message here but I think it might be better to do that later where we catch errors when performing the copy.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/Source.java
adinn: Fixed

adinn:
I'd prefer to put this on hold until it is decided whether this is really needed to support the model of debug info generation proposed for LLVM. So, I am reverting it to draft status.

Issue #5396, opened 2022-11-09
[GR-42465] JFR jdk.ContainerConfiguration event reports wrong values in a container


jerboaa:
**Describe the issue**
The `jdk.ContainerConfiguration` JFR event reports incorrect memory limits when run in a container. It reports unlimited values instead of the set container memory limit.

**Steps to reproduce the issue**
1. `git clone https://github.com/Karm/mandrel-integration-tests.git`
2. `cd mandrel-integration-tests/apps/debug-symbols-smoke`
3. `mvn clean package`
4. `native-image --enable-monitoring -jar target/debug-symbols-smoke.jar target/debug-symbols-smoke`
5. Run the result in a container for example with `sudo podman run --memory=1G --memory-swap=1G --rm -ti -v $(pwd)/target:/opt/target:z -v /path/to/graalvm:/opt/jdk:z fedora:37`
6. `[root@62988404d0a9 opt]# /opt/target/debug-symbols-smoke  -XX:+FlightRecorder -XX:StartFlightRecording=filename=flight-native.jfr   -XX:FlightRecorderLogging=jfr`
8. Inspect the JFR events for example with the `jfr` tool: `/opt/jdk/bin/jfr print --events jdk.ContainerConfiguration flight-native.jfr`

Observe the `memoryLimit` and `swapMemoryLimit` being reported incorrectly. Shows `-1`, while it should be showing `1 GB`.

**Describe GraalVM and your environment:**
 - GraalVM version: Graal master, revision `7b1952da4031ae9f8446b2e04e6d9b8575607004`
 - JDK major version: **17**
 - OS: **Linux**
 - Architecture: **AMD64**

**More details**
It's not reproducible with JVM mode:
```
[root@62988404d0a9 opt]# ./jdk/bin/java -XX:StartFlightRecording=filename=flight-java.jfr -Xlog:jfr -jar ./target/debug-symbols-smoke.jar
[0.301s][info][jfr] Flight Recorder initialized
[0.301s][info][jfr] Repository base directory: /tmp
[0.386s][info][jfr] Started recording "1" (1) {dumponexit=true, filename=/opt/flight-java.jfr}
[0.386s][info][jfr,startup] Started recording 1. No limit specified, using maxsize=250MB as default.
[0.386s][info][jfr,startup] 
[0.386s][info][jfr,startup] Use jcmd 13 JFR.dump name=1 to copy recording data to file.
Q to quit
Q
[...]

PR #5390, opened 2022-11-08
[GR-42460] Add support for CgroupMetrics.getTotalMemory0

jerboaa:
Latest JDK code received an update to bound the
reported container memory above by the total physical memory
of the host running the container. See [JDK-8292541](https://bugs.openjdk.org/browse/JDK-8292541)

Implement the same in GraalVM, only substituting the class/methods
if they exist in the JDK.

Closes #5387

PR #5400, opened 2022-11-09
[GR-42465] Reset JDKs cgroup metrics instances

jerboaa:
This is needed in order to get correct memory limits detected for some code paths (one of which is JFR).

Closes: #5396

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKContainerSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKContainerSubstitutions.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2022, 2022, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.svm.core.jdk;
+
+import org.graalvm.nativeimage.Platform.LINUX;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import org.graalvm.nativeimage.Platforms;
+
+import com.oracle.svm.core.annotate.Alias;
+import com.oracle.svm.core.annotate.RecomputeFieldValue;
+import com.oracle.svm.core.annotate.RecomputeFieldValue.Kind;
+
+import com.oracle.svm.core.annotate.TargetClass;
+
+@TargetClass(className = "jdk.internal.platform.cgroupv1.CgroupV1Subsystem")
+@Platforms(LINUX.class)
+final class Target_jdk_internal_platform_cgroupv1_CgroupV1Subsystem {
+    @Alias //
+    @RecomputeFieldValue(kind = Kind.Reset) //
+    private static volatile Target_jdk_internal_platform_cgroupv1_CgroupV1Subsystem INSTANCE;
+}
+
+@TargetClass(className = "jdk.internal.platform.cgroupv2.CgroupV2Subsystem")
+@Platforms(LINUX.class)
+final class Target_jdk_internal_platform_cgroupv2_CgroupV2Subsystem {
+    @Alias //
+    @RecomputeFieldValue(kind = Kind.Reset) //
+    private static volatile Target_jdk_internal_platform_cgroupv2_CgroupV2Subsystem INSTANCE;
+}
+
+@TargetClass(className = "jdk.jfr.internal.instrument.JDKEvents")
+@Platforms(LINUX.class)
+final class Target_jdk_jfr_internal_instrument_JDKEvents {
+    @Alias //
+    @RecomputeFieldValue(kind = Kind.Reset) //
+    private static Target_jdk_internal_platform_Metrics containerMetrics;
+
+    @Alias //
+    @RecomputeFieldValue(kind = Kind.Reset) //
+    private static boolean initializationTriggered;
+}
+
+@TargetClass(className = "jdk.jfr.internal.RequestEngine")
+@Platforms(LINUX.class)
+final class Target_jdk_jfr_internal_RequestEngine {
+    @Alias //
+    @RecomputeFieldValue(kind = Kind.NewInstance, declClass = CopyOnWriteArrayList.class) //
+    private static List<?> entries;
+}
+
+// Only present in JDKs without JDK-8268398
+@TargetClass(className = "jdk.jfr.internal.Utils", onlyWith = {JDK17OrEarlier.class, JDK17OrLater.class})
+@Platforms(LINUX.class)
+final class Target_jdk_jfr_internal_Utils {
+    @Alias //
+    @RecomputeFieldValue(kind = Kind.Reset) //
+    private static Target_jdk_internal_platform_Metrics[] metrics;
+}
+
+@TargetClass(className = "jdk.internal.platform.Metrics")
+@Platforms(LINUX.class)
+final class Target_jdk_internal_platform_Metrics {
+}
+
+/** Dummy class to have a class with the file's name. */
+public final class JDKContainerSubstitutions {
+}


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/Containers.java
```diff
@@ -220,5 +220,10 @@ public void duringSetup(DuringSetupAccess access) {
         RuntimeClassInitializationSupport classInitSupport = ImageSingletons.lookup(RuntimeClassInitializationSupport.class);
         classInitSupport.initializeAtRunTime("com.oracle.svm.core.containers.cgroupv1.CgroupV1Subsystem", "for cgroup support");
         classInitSupport.initializeAtRunTime("com.oracle.svm.core.containers.cgroupv2.CgroupV2Subsystem", "for cgroup support");
+        // JDK platform mirror classes need to get re-initialized at runtime so as to update to the
+        // correct path of the cgroup interface files stored in controllers. This code-path is being used for example via
+        // JFR tracing.
+        classInitSupport.rerunInitialization("jdk.internal.platform.cgroupv1.CgroupV1Subsystem", "for cgroup support");
```
zakkak, line 226: It looks like `jdk.internal.platform.CgroupV1MetricsImpl` will need some special handling as well. I wonder if we indeed need to re-run initialization on these classes or if it would suffice to reset the `INSTANCE` field value.

The error from the CI with the current approach is:
```
 com.oracle.svm.core.util.UserError$UserException: No instances of jdk.internal.platform.cgroupv1.CgroupV1Subsystem are allowed in the image heap as this class should be initialized at image runtime. To see how this object got instantiated use --trace-object-instantiation=jdk.internal.platform.cgroupv1.CgroupV1Subsystem.
Detailed message:
Trace: Object was reached by
  reading field jdk.internal.platform.CgroupMetrics.subsystem of constant 
    jdk.internal.platform.CgroupV1MetricsImpl@38b7a16c: jdk.internal.platform.CgroupV1MetricsImpl@38b7a16c
  reading static field jdk.jfr.internal.instrument.JDKEvents.containerMetrics
    at jdk.jfr.internal.instrument.JDKEvents.emitContainerIOUsage(JDKEvents.java:242)
  parsing method jdk.jfr.internal.instrument.JDKEvents.emitContainerIOUsage(JDKEvents.java:242) reachable via the parsing context
    at jdk.jfr.internal.instrument.JDKEvents$$Lambda$69/0x00000007c00e7aa8.run(Unknown Source)
    at java.lang.Shutdown.runHooks(Shutdown.java:130)
    at java.lang.Shutdown.exit(Shutdown.java:173)
    at com.oracle.svm.core.graal.snippets.CEntryPointSnippets.initializeIsolate(CEntryPointSnippets.java:314)
```

zakkak (changes requested):
LGTM,

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/Containers.java
jerboaa: Thanks, good point. That might work.

jerboaa: Updated the patch with resetting various statics instead. Lets see what CI says to this.

zakkak (approved):
LGTM, thanks @jerboaa !

fniephaus (commented):
LGTM, thanks!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKContainerSubstitutions.java
```diff
@@ -0,0 +1,76 @@
...
+@TargetClass(className = "jdk.jfr.internal.instrument.JDKEvents")
+@Platforms(LINUX.class)
+final class Target_jdk_jfr_internal_instrument_JDKEvents {
+    @Alias //
+    @RecomputeFieldValue(kind = Kind.Reset) //
+    private static Target_jdk_internal_platform_Metrics containerMetrics;
+}
+
+// Only present in JDKs without JDK-8268398
+@TargetClass(className = "jdk.jfr.internal.Utils", onlyWith = JDK17OrEarlier.class)
```
fniephaus, line 61: Going to adjust this to:
```suggestion
@TargetClass(className = "jdk.jfr.internal.Utils", onlyWith = {JDK17OrLater.class, JDK17OrEarlier.class})
```
as the `metrics` field only exists in JDK17.

zakkak: Maybe check for the presence of `metrics` through reflection instead (in a similar manner to https://github.com/oracle/graal/pull/5390/files#diff-107062ee1800e4a17640d4d352ebe6b1fb7fa260ab7055d4bef4565b833e8e61)?

jerboaa: Yes, I can update. @fniephaus sound good to you or are you already on it?

fniephaus: I'm already on it. Don't think we need a check based on reflection at this point.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKContainerSubstitutions.java
```diff
@@ -0,0 +1,76 @@
...
+    @RecomputeFieldValue(kind = Kind.Reset) //
+    private static volatile Target_jdk_internal_platform_cgroupv2_CgroupV2Subsystem INSTANCE;
+}
+
+@TargetClass(className = "jdk.jfr.internal.instrument.JDKEvents")
+@Platforms(LINUX.class)
+final class Target_jdk_jfr_internal_instrument_JDKEvents {
+    @Alias //
+    @RecomputeFieldValue(kind = Kind.Reset) //
+    private static Target_jdk_internal_platform_Metrics containerMetrics;
```
fniephaus, line 57: How is this re-initialized at run time? Unless I'm missing something, this will just disable container events.

jerboaa: It's initialized with `PlatformRecorder()` => `JDKEvents.initialize()` => `initializeContainerEvents()`. What this change does is preventing build-time state to populate to runtime unless I'm missing something. In JVM mode it's set to `null` here:
https://github.com/openjdk/jdk17u-dev/blob/master/src/jdk.jfr/share/classes/jdk/jfr/internal/instrument/JDKEvents.java#L116

https://github.com/oracle/graal/issues/5396 has a reproducer and according to my testing it's fixed with this. Have you determined otherwise?

fniephaus: And `PlatformRecorder` is initialized at run-time?

jerboaa: That's my understanding, yes. When I run the reproducer:

```
[root@7de05057c67d /]# /opt/target/debug-symbols-smoke -XX:+FlightRecorder -XX:StartFlightRecording=filename=flight-native.jfr -XX:FlightRecorderLogging=jfr
[info][jfr] Added periodic hook for EveryChunkPeriodEvents(12336)
[info][jfr] Added periodic hook for EndChunkPeriodEvents(12314)
[info][jfr] Flight Recorder initialized
[info][jfr] Repository base directory: /tmp
[info][jfr] Started recording "1" (1) {dumponexit=true, filename=/flight-native.jfr}
Q to quit
```

Note the line `Flight Recorder initialized` which comes from https://github.com/openjdk/jdk17u-dev/blob/master/src/jdk.jfr/share/classes/jdk/jfr/FlightRecorder.java#L188 at which point a new instance of `PlatformRecorder()` would already been created, triggering the initialization chain.

fniephaus: Thanks for checking!

jerboaa: Note that this initially came from https://github.com/oracle/graal/pull/5400#discussion_r1018769862 where I've opted for class re-init of the cgroup instances (then changed it to instance reset). However, I'll run some more tests with JFR enabled at image build time. It might be we need to also reset https://github.com/openjdk/jdk17u-dev/blob/master/src/jdk.jfr/share/classes/jdk/jfr/internal/instrument/JDKEvents.java#L117

jerboaa: > I'll run some more tests with JFR enabled at image build time.

This also works with an image produced with JFR enabled at image build time according to my testing. So I think this is resolved?

fniephaus: Again, thanks for checking. Yes, I think this is resolved. I've put the PR in the merge queue. Enjoy your weekend. :)

fniephaus: I had to take it out of the merge queue and will be on PTO until the end of the month. Internal reviewers noted:

>I hope I'm not missing something obvious, but based on the `JDKEvents.initialize()` source code, this substitution just doesn't make sense - it's either redundant or incomplete. ...

and

>... I checked both the JDK 17 and JDK 19 sources and as you stated, this does not make any sense. There are two cases:
a. Native image build with JFR at runtime only: `JDKEvents.initialize()` is not executed at image build time, so the `containerMetrics` are also not initialized. Resetting containerMetrics is useless but not harmful.
b. Native image build with JFR at both build and runtime: `JDKEvents.initialize()` is executed at image build time and `containerMetrics` is initialized. Resetting `containerMetrics` alone is not sufficient, and reexecuting `JDKEvents.initialize()` can be harmful.

jerboaa: Thanks. Apparently in my testing I've used an old image :-( Sorry about that. Yes, those JDKEvents resets are necessary to cater the case where JFR is enabled at build and runtime. As expected, https://github.com/openjdk/jdk17u-dev/blob/master/src/jdk.jfr/share/classes/jdk/jfr/internal/instrument/JDKEvents.java#L117 needed resets too as well as the list recording registered hooks. The updated patch should have this now.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKContainerSubstitutions.java
jerboaa: FWIW, I've added this to the latest patch. As it needed changes for the JFR resets anyway.

jerboaa:
@fniephaus It's my understanding that `jdk.Container*` events can be ticked off on the list in https://github.com/oracle/graal/issues/5410 too with this fix. /cc @roberttoyonaga

roberttoyonaga:
>It's my understanding that `jdk.Container*` events can be ticked off on the list in #5410 too with this fix

Yup they seem to be working now. I'll update the task list. Thank you!

jerboaa:
@fniephaus Are there any other issues with this patch that aren't addressed, yet? Thanks!

fniephaus:
Thanks for resolving the open issue and for the delay. I've asked for another internal review and will merge this when all is well.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKContainerSubstitutions.java
christianhaeubl: If `JDKEvents.initialize()` was executed at build time and is then reexecuted at run-time, I think that this can cause issues. Doesn't for example the following call chain throw an `IllegalArgumentException("Hook has already been added")` on JDK 19?
`JDKEvents.initialize()` -> `JDKEvents.initializeContainerEvents()` -> `RequestEngine.addTrustedJDKHook(...)` -> `RequestEngine.addHookInternal(...)`

fniephaus:
The PR is now in our merge queue.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKContainerSubstitutions.java
jerboaa: By `disregard the example below` you meant this, right?

```
I think that reexecuting `JDKEvents.initialize()` at runtime can cause issues if the method was already executed at image build time. For example, doesn't the following call chain throw an `IllegalArgumentException("Hook has already been added")` on JDK 19?

`JDKEvents.initialize()` -> `RequestEngine.addTrustedJDKHook(...)` -> `RequestEngine.addHookInternal(...)`
```

fniephaus:
This was merged via https://github.com/oracle/graal/commit/a01e8b772a5802e6cad159bc6fcf1c1486d004a1.

jerboaa:
Thanks!

PR #5441, opened 2022-11-16
[GR-42375] Introduce `-H:±GenerateBuildArtifactsFile` option.

graalvmbot:
and deprecate `.build_artifacts.txt` files.


fniephaus:
@jerboaa please feel free to take a look at this. We've decided to not include this in the build output JSON because the two files target different consumers: build output for monitoring, build artifacts for packaging tools.
Also note the new (internal) `-H:BuildArtifactsJSON` option only accepts a path and is inconsistent with `-H:BuildOutputJSONFile`, which also accepts a filename. I will share more details on how we plan to unified all of this within a public API in the future, would be great to get your feedback on that as well.

fniephaus:
Thanks for the quick review.

>As long as we have a way of knowing what the produced output filename is (for -HBuildOutputJSONFile) this should be fine.

Yes, although we probably won't allow users to customize the filename (easier to unify output and to match against corresponding schema), only the target directory can be changed.

fniephaus:
>Please consider that, thanks!

Sure, will do. Thank you

fniephaus:
Please feel free to take another look, @jerboaa:
- The option is now a boolean flag. Not sure whether it should be called `-H:±BuildArtifacts`, `-H:±ReportBuildArtifacts`, or something else?
- Paths are now relative again, using `uri-reference`, which seems to allow relative paths (see [here](https://json-schema.org/understanding-json-schema/reference/string.html#resource-identifiers))
- Since we couldn't identify any external dependents, I disabled the creation of the old file by default and added support for `NATIVE_IMAGE_DEPRECATED_BUILD_ARTIFACTS_TXT=true` as a temporary fallback

jerboaa:
I'll take a look next week, thanks.

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -415,26 +415,29 @@ def help_stdout_check(output):
...
             except ImportError:
                 mx.abort('Unable to import jsonschema')
+            with open(join(suite.dir, '..', 'docs', 'reference-manual', 'native-image', 'assets', 'build-artifacts-schema-v0.9.0.json')) as f:
+                build_artifacts_schema = json.load(f)
             with open(join(suite.dir, '..', 'docs', 'reference-manual', 'native-image', 'assets', 'build-output-schema-v0.9.1.json')) as f:
-                json_schema = json.load(f)
+                build_output_schema = json.load(f)
             with tempfile.NamedTemporaryFile(prefix='build_json') as json_file:
-                helloworld(['--output-path', svmbuild_dir(), f'-H:BuildOutputJSONFile={json_file.name}'])
+                helloworld(['--output-path', svmbuild_dir(), f'-H:BuildOutputJSONFile={json_file.name}', f'-H:BuildArtifactsJSON={svmbuild_dir()}'])
```
jerboaa, line 431: Isn't this option no longer valid? `-H:+BuildArtifacts`?

This is also the reason for the gate failure it seems:
```
 Running: mx --primary-suite-path substratevm --java-home=/home/runner/work/graal/graal/jdk helloworld --output-path /home/runner/work/graal/graal/substratevm/svmbuild -H:BuildOutputJSONFile=/tmp/build_jsontyc_wdu_ -H:BuildArtifactsJSON=/home/runner/work/graal/graal/substratevm/svmbuild
Error: Could not find option 'BuildArtifactsJSON'. Did you mean one of these: BuildArtifacts. Use -H:PrintFlags= to list all available options.
```

fniephaus: Ype, forgot to update that test when I pushed this. Should be fixed now.

jerboaa (approved):

PR #5447, opened 2022-11-16
Thin locking for object monitors

petermz:
Hi --
This patch implements Thin Lock monitor support.

### Description
The idea is to make uncontended monitor enter/exit as fast as possible, and fall back to current approach in case of contention. The algorithm uses mark word in object header to hold information about owner thread and lock counter. In case of contention the algorithm replaces the mark word with newly created JavaMonitor and switches to old slow path.

The patch introduces new option, `-H:±ThinLock`, `true` by default.

### Performance
A typical Spring application starts about 5% faster (as measured on Ubuntu x86_64):
```
thin lock off : 0.110 ms
thin lock on : 0.104 ms
```

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/GreyToBlackObjRefVisitor.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/GreyToBlackObjRefVisitor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2017, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -24,18 +24,19 @@
  */
 package com.oracle.svm.core.genscavenge;
 
-import org.graalvm.compiler.word.Word;
-import org.graalvm.nativeimage.Platform;
-import org.graalvm.nativeimage.Platforms;
-import org.graalvm.word.Pointer;
-import org.graalvm.word.UnsignedWord;
-
 import com.oracle.svm.core.AlwaysInline;
 import com.oracle.svm.core.genscavenge.remset.RememberedSet;
 import com.oracle.svm.core.heap.ObjectReferenceVisitor;
 import com.oracle.svm.core.heap.ReferenceAccess;
 import com.oracle.svm.core.hub.LayoutEncoding;
 import com.oracle.svm.core.log.Log;
+import org.graalvm.compiler.word.Word;
+import org.graalvm.nativeimage.Platform;
+import org.graalvm.nativeimage.Platforms;
+import org.graalvm.word.Pointer;
+import org.graalvm.word.UnsignedWord;
+
+import static com.oracle.svm.core.monitor.ThinLockMonitorSupport.isThinLockMonitor;
 
 /**
  * This visitor is handed <em>Pointers to Object references</em> and if necessary it promotes the
@@ -81,6 +82,10 @@ final class GreyToBlackObjRefVisitor implements ObjectReferenceVisitor {
 
         if (HeapImpl.getHeapImpl().isInImageHeap(p)) {
             counters.noteNonHeapReferent();
+            return true;
+        }
+
+        if (isThinLockMonitor(p)) {   // Ignore if it is thin lock mark
             return true;
         }
 
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
@@ -342,6 +342,9 @@ public class SubstrateOptions {
     @Option(help = "Enable support for threads and and thread-local variables (disable for single-threaded implementation)")//
     public static final HostedOptionKey<Boolean> MultiThreaded = new HostedOptionKey<>(true);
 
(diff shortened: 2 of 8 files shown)


SergejIsbrecht:
I have been running a JMH benchmark on native image CE 17 with your branch (22.3-dev). The benchmark contains a _lot_ of synchronization.

(NOTE: I do not know how reliable the results are. Furthermore I am not not able to share this benchmark. I am currently trying to extract the IP code in order for you to run it.)

-H:-ThinLock
```plain
 ~/D/native_image_bench  ./dispatcher_threadpool.ce.java17.amd64.minus_thinLock -Xmx10g -Xms10g -XX:+PrintGC -XX:+VerboseGC                          2.7m  So 20 Nov 2022 22:41:59 CET
# JMH version: 1.35
# VM version: JDK 17.0.5, Substrate VM, GraalVM dev Java 17 CE
# *** WARNING: This VM is not supported by JMH. The produced benchmark data can be completely wrong.
# VM invoker: null/bin/java
# VM options: -Xmx10g -Xms10g -XX:+PrintGC -XX:+VerboseGC

Benchmark                       (blackHoleCnt)  (dispatcherCnt)  (jobCnt)  Mode  Cnt     Score     Error  Units
DispatcherBenchmark.threadpool            1000               20     10000  avgt   10  1883.084 ± 137.380  ms/op
```

-H:+ThinLock
```plain
 ~/D/native_image_bench  ./dispatcher_threadpool.ce.java17.amd64.minus_thinLock -Xmx10g -Xms10g -XX:+PrintGC -XX:+VerboseGC                          2.7m  So 20 Nov 2022 22:41:59 CET
# JMH version: 1.35
# VM version: JDK 17.0.5, Substrate VM, GraalVM dev Java 17 CE
# *** WARNING: This VM is not supported by JMH. The produced benchmark data can be completely wrong.
# VM invoker: null/bin/java
# VM options: -Xmx10g -Xms10g -XX:+PrintGC -XX:+VerboseGC

Benchmark                         (blackHoleCnt)  (dispatcherCnt)  (jobCnt)  Mode  Cnt     Score    Error  Units
DispatcherBenchmark.threadpool            1000               20     10000  avgt   10  1108.916 ± 23.762  ms/op
```

native-image GraalVM CE 22.3.0
```
 ~/D/native_image_bench  ./dispatcher_threadpool.ce.java17.amd64 -Xmx10g -Xms10g -XX:+PrintGC -XX:+VerboseGC                                         2.7m  So 20 Nov 2022 22:44:51 CET
# JMH version: 1.35
# VM version: JDK 17.0.5, Substrate VM, GraalVM 22.3.0 Java 17 CE
# *** WARNING: This VM is not supported by JMH. The produced benchmark data can be completely wrong.
# VM invoker: null/bin/java
# VM options: -Xmx10g -Xms10g -XX:+PrintGC -XX:+VerboseGC

Benchmark                         (blackHoleCnt)  (dispatcherCnt)  (jobCnt)  Mode  Cnt     Score    Error  Units
DispatcherBenchmark.threadpool            1000               20     10000  avgt   10  1021.065 ± 41.117  ms/op
```

As far as I see it the results show a decrease in performance.

petermz:
```
-H:-ThinLock: 1883.084 ± 137.380  ms/op
-H:+ThinLock: 1108.916 ± 23.762  ms/op
```
This looks like an (unbelievably) huge speedup rather than a performance decrease. Am I missing something? Anyway, I don't think this change could result in such a large speedup, **especially** in contended case.

Do you really have to pass `-XX:+PrintGC -XX:+VerboseGC` to the benchmark ? These could easily skew the results I guess...

zakkak:
@SergejIsbrecht as mentioned in the warnings, _This VM is not supported by JMH. The produced benchmark data can be completely wrong_. Please don't use JMH for your experiments as it doesn't support `native-image` (which among others might be optimizing out large parts of the benchmark because of that).

SergejIsbrecht:
@petermz , I mixed up the results. When the native-image is build with `+ThinLock` the results get worse. I will run it again and update my values.

> Do you really have to pass -XX:+PrintGC -XX:+VerboseGC to the benchmark ?

This is just for sanity, because I am benchmarking a legacy threadpool implementation. The SerialGC of CE is quite bad, so I have to see whether the GC somehow changes my results.

As @zakkak already said, JMH is not supported, therefore I will just use the main later on this week. I will also look into the perf output.

peter-hofer:
Hi @petermz , thank you for your contribution. Thin locks can be helpful to improve uncontended performance, but in the presence of contention and without deflation, the additional code paths can also hurt performance. Therefore, careful benchmarking is necessary to determine whether an implementation provides a speedup in typical workloads.

I see conceptual problems for Native Image given that we're using Java objects for inflated locks. The monitor field is an object field, which means that accesses to it use GC barriers and do heap address arithmetic, both of which need to be carefully avoided with thin lock words. GraalVM Enterprise Edition (EE) uses compressed 32-bit references which are insufficient for unique thread ids (especially considering virtual threads) and have no spare bit to mark a thin lock either. We could make the monitor field a `long`, but that would increase the size of objects with EE. For both object or `long` monitor fields, we would need to add explicit handling for the field in many more places (including the G1 GC in EE) and ensure correct barriered accesses, which can introduce subtle but horrible bugs, and would also increase the size of objects with EE. An implementation would need to provide _substantial_ speedups to justify these tradeoffs. @christianhaeubl @christianwimmer 

I also see the following issues with the code:
* `ThinLockMonitorSupport.inflateMonitor()` is not uninterruptible, so after invoking `getCompressedRepresentation` a GC can happen and move the lock object and turn the obtained lock reference representation invalid.
* The GC is not the only place that can encounter a thin lock (see other implementations of `ObjectReferenceVisitor`, for example)
* Checking for thin locks should be restricted to the monitor field
* `Thread.getId()` can be overridden, we have `JavaThreads.getThreadId()` to work around that (note that Java 19 introduced `Thread.threadId()` for the same reason)

peter-hofer:
Closing due to inactivity.

PR #5460, opened 2022-11-17
[GR-4247] Initial Native image Replay bundle Support.

graalvmbot:


docs/reference-manual/native-image/ReplayBundles.md
```diff
@@ -0,0 +1,179 @@
...
+### `build`:
+
+This folder is used to document the build process that lead to the image that was created alongside the bundle.
+The `report` sub-folder holds `build.log`. It is equivalent to what would have been created if the user had appended
+`|& tee build.log` to the original native-image command line. Additionally, we have several json-files:
+* `analysis_results.json`: Contains the results of the static analysis. A rerun should compare the new
+`analysis_results.json` file with this one and report deviations in a user-friendly way.
+* `build_artifacts.json`: Contains a list of the artifacts that got created during the initial build. As before,
+changes should be reported to the user. 
+* `build_output.json`: Similar information as `build.log`.
```
jerboaa, line 137: I think `build_output.json` (is that `-H:BuildOutputJSONFile`?), doesn't include linked native libs from the JDK. Maybe that's also something worth collecting as it might be a source of differences.

jerboaa (commented):
A couple of thoughts from my end. Seems a nice proposal!

docs/reference-manual/native-image/ReplayBundles.md
```diff
@@ -0,0 +1,179 @@
...
+
+File `all.env` contains the environment variables that we allowed the builder to see during the initial build and
+`all.properties` the respective system-properties.
+
+### `build`:
+
+This folder is used to document the build process that lead to the image that was created alongside the bundle.
+The `report` sub-folder holds `build.log`. It is equivalent to what would have been created if the user had appended
+`|& tee build.log` to the original native-image command line. Additionally, we have several json-files:
+* `analysis_results.json`: Contains the results of the static analysis. A rerun should compare the new
```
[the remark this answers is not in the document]
olpaw: We need to record/document if the initial build was performed in a container.
In which case we also need to record:
* The version of docker that was used. E.g. `podman version 4.3.1`
* Which kernel it was running on. E.g. `6.0.9-300.fc37.x86_64`
* Which Dockerfile was used. E.g.
```Dockerfile
FROM registry.fedoraproject.org/fedora:37
RUN dnf -y install gcc zlib-static
WORKDIR /ni-build
ENTRYPOINT ["/graalvm/bin/native-image"]
CMD ["--help"]
```
* The docker command line (without the native-image part)
```
podman run --network=none -it --mount type=bind,input="<path-to-graalvm>",target=/graalvm,readonly --mount type=bind,source=<path-to-input-folder>,target=/input container-graalvm
```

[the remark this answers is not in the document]
olpaw: > Is the premise that build a) and b) must both run in the restricted container env?

Yes. Using a containerized build binds you to further use the same container setup for future rebuilds.

docs/reference-manual/native-image/ReplayBundles.md
```diff
@@ -0,0 +1,179 @@
...
+## Enforced sanitized image building
+
+### Containerized image building on supported platforms
+
+If available, docker/podman should be used to run the image builder inside a well-defined container image. **This allows
+us to prevent the builder from using the network during image build**, thus guaranteeing that the image build result did
+not depend on some unknown (and therefore unreproducible) network state. Another advantage is that we can mount
+`input/classes` and `$GRAALVM_HOME` read-only into the container and only allow read-write access to the mounted `out`
+and `build` directories. This will prevent the application code that runs at image build time to mess with anything
+other than those directories.
```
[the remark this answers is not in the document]
olpaw: Thanks for the heads up. We should record if cgroup v2 or cgroup v1 was used in a containerized build.

docs/reference-manual/native-image/ReplayBundles.md
olpaw: Are you suggesting to record which static native libs have been linked into the image by the native-linker?

I would consider that redundant as it is fully determined by the version of GraalVM + the native-image command line options.

olpaw:
I moved the contents of `ReplayBundles.md` into issue https://github.com/oracle/graal/issues/5473 and made it part of the roadmap.

docs/reference-manual/native-image/ReplayBundles.md
```diff
@@ -0,0 +1,207 @@
...
+with:
+
+```shell
+native-image --replay .../path/to/alaunch.nirb.jar
+```
+
+this will rebuild the `alaunch` image with the same image arguments, environment variables, system properties
+settings, classpath and module-path options as in the initial build.
+
+To support the use-case of image-building-as-a-service, there should also be a way to create a bundle without
```
olpaw, line 52: Added based on suggestion from @stooke

docs/reference-manual/native-image/ReplayBundles.md
```diff
@@ -0,0 +1,207 @@
...
+**even the initial build that created the bundle already uses a class- and/or module-path that refers to the contents
+of the `input/classes` folder**. This way we can guarantee that a replay build sees exactly the same relocated
+class/module-path entries as the initial build. The use of `run.cmd` is explained later.
+
+File `all.env` contains the environment variables that we allowed the builder to see during the initial build and
+`all.properties` the respective system-properties.
+
+#### `input/stage/container`:
+
+If the image builder runs in a container environment, this subdirectory holds all information necessary to redo the
```
olpaw, line 148: Added based on suggestions of @jerboaa

docs/reference-manual/native-image/ReplayBundles.md
```diff
@@ -0,0 +1,207 @@
...
+├── build
+│   └── report <- Contains information about the build proccess.
+│       │         In case of replay these will be compared against. 
+│       ├── analysis_results.json
+│       ├── build_artifacts.json
+│       ├── build.log
+│       ├── build_output.json
+│       ├── jni_access_details.json
+│       └── reflection_details.json
+├── input
```
petermz, line 76: - I think JSON config files should be included here (JNI/reflection/resources) etc. Those can be passed via command line options
- Does it make sense to capture compiler/linker versions? There are also dynamic libs, e.g. an AWT program can link `libX11.so` etc, but this is probably too much. But might make sense to capture the linker command that gets executed at the final step

olpaw: > I think JSON config files should be included here (JNI/reflection/resources) etc. Those can be passed via command line options

@petermz good point! All the config files that are passed to native image that were not embedded into on of the input jars need to be put in here as well. We should have an extra subdirectory for such files/directories e.g. `input/auxiliary` would make sense.

> Does it make sense to capture compiler/linker versions?

For non-containerized builds it does. We mostly already have that info in the builder (see class `com.oracle.svm.hosted.c.codegen.CCompilerInvoker`)  we just need to make sure its a mandatory part of `build_output.json`.

petermz:
One more note: the `nirb` in `<imagename>.nirb.jar` feels pretty cryptic to me. Is `<imagename>.replay.jar` any better? At least it's consistent with how command line options are named

docs/reference-manual/native-image/ReplayBundles.md
jerboaa: > Are you suggesting to record which static native libs have been linked into the image by the native-linker?

Yes, the ones coming from the JDK. For example:

```
$ strings ./debug-symbols-smoke-mandrel | grep com.oracle.svm.core.VM.Target.StaticLibraries
com.oracle.svm.core.VM.Target.StaticLibraries=liblibchelper.a|libnet.a|libnio.a|libjava.a|libfdlibm.a|libzip.a|libjvm.a
```

> I would consider that redundant as it is fully determined by the version of GraalVM + the native-image command line options.

I'm not sure, we've seen issues with - a supposedly headless AWT - image build, which did include a different set depending on the build host (`GraphicsEnvironment.isHeadless()`). See https://github.com/oracle/graal/issues/5119

olpaw:
> One more note: the `nirb` in `<imagename>.nirb.jar` feels pretty cryptic to me. Is `<imagename>.replay.jar` any better? At least it's consistent with how command line options are named

Good idea. See https://github.com/oracle/graal/issues/5473#issuecomment-1326388734

Issue #4921, opened 2022-09-11
Improve AWT Support


christianwimmer:
### TL;DR

Improve AWT support to use dynamic linking of the AWT native code libraries. Generalize the dynamic linking approach, which is currently used only on Windows, to all platforms. Drop the current static linking approach used for Linux

### Goals

The current support for AWT on Windows uses dynamic linking of the native code libraries: the unmodified shared libraries of the JDK are loaded dynamically at image run time using the regular JNI mechanism. The only "downside" of this approach is that the shared libraries need to be shipped together with the application executable.

On Linux, currently a static linking approach is used: the AWT native code is linked statically into the application. This approach has significant compatibility issues and would require changes to the native code for future JDK versions and to support additional platforms. It is therefore a goal to remove the static linking approach on Linux and switch to the dynamic linking. Once dynamic linking is working on Windows and Linux, adding support for MacOS will be not much additional work.

### Non-goals

* We will remove the current static linking approach used for Linux. This approach can be considered a failed experiment because it is not maintainable and has no significant benefits over the dynamic linking approach.
* It is not a goal to provide a complete reflection and resource registration of all parts of AWT. Each application still needs to provide its own registrations, which can be collected using the tracing agent.

anasmughal:
Thank you for slating this to a future release. Do we have any idea when it will be available? 

Out product release depends on this. 

Thanks.

bflorat:
"The only "downside" of this approach is that the shared libraries need to be shipped together with the application executable." 
Could you please detail what you think of ?

As a packager, what will I be required to ship within my application ? provide libawt.so ?  a full JRE (which would obviously defeat the goal of graalvm native image feature) ?

kristofdho:
> As a packager, what will I be required to ship within my application ? provide libawt.so ? a full JRE (which would obviously defeat the goal of graalvm native image feature) ?

I imagine this will work the same for unix as it currently works for Windows.
An example of how my (cleaned up) target folder looks with a completed build:
<img width="168" alt="image" src="https://user-images.githubusercontent.com/43339887/205896844-d22e3ea1-5e1f-462a-b76f-8dd772a55220.png">

The content of `native)image.build_artifacts.txt`:
```ini
[EXECUTABLE]
native-image.exe

[JDK_LIB]
awt.dll
fontmanager.dll
freetype.dll

[JDK_LIB_SHIM]
# Note that shim JDK libraries depend on this
# particular native image (including its name)
# and therefore cannot be used with others.
[...]

PR #5466, opened 2022-11-18
Register JNI and Reflection classes for AWT/Swing on Linux

AlexanderScherbatiy:
GraalVM provides tracing agent which allows to find JNI and Reflection classes used by an application and store them in configuration files.  Using the tracing agent with UI application requires checking non obvious use cases to cover different application code paths: Drag & Drop, file choosers, input dialogs, key strokes, Look & Feels.

 Key strokes provide the most vivid use case. [AWTKeyStroke.getVKValue(String)](https://github.com/openjdk/jdk19u/blob/045c1f64897dbd8c12795ba84396ef12598f70a0/src/java.desktop/share/classes/java/awt/AWTKeyStroke.java#L532
) method uses reflection to get a key code for for the requested `KeyEvent.VK_*` string value. It requires pressing all keys used as key strokes in the application during the tracing agent usage.

The proposed fix registers common classes used by AWT/Swing application on Linux System.

L&F UI classes are registered only for Basic and Metal L&Fs as [UIManager.setLookAndFeel(String)](https://github.com/openjdk/jdk19u/blob/045c1f64897dbd8c12795ba84396ef12598f70a0/src/java.desktop/share/classes/javax/swing/UIManager.java#L626) method directly sets MetalLookAndFeel class. It is necessary to use the tracing agent when an application uses GTK, Nimbus, and other Look & Feels.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationAwt.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JNIRegistrationAwt.java
@@ -1,6 +1,5 @@
 /*
  * Copyright (c) 2020, 2022, Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2022, 2022, BELLSOFT. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


zakkak:
Related to https://github.com/oracle/graal/issues/4921

alina-yur:
Hi @AlexanderScherbatiy, sorry for the delay on this PR. Your OCA submission is now approved. Can you please resubmit this PR using your company email so the OCA checks pass? Thank you!

AlexanderScherbatiy:
The PR is resubmitted as https://github.com/oracle/graal/pull/5870

PR #5475, opened 2022-11-21
[GR-43600] Additional debuginfo memory footprint improvements

adinn:
This is a second round of improvements to debug info footprint. It involves one change to the COMPILER phase and four to the DWARF phase:
### COMPILER phase
1. Each entry in a compiled method's `Infopoint` list includes a leaf `BytecodeFrame` which links an associated linear chain of caller `BytecodeFrame` instances. In many cases the chains in adjacent `Infopoint` instances are either identical or share most ancestors (based on `BytecodeFrame.equals`). Storage can be compressed by merging this list of lists to an equivalent tree where common ancestors are reduced to unique instances. This can be done as a final step during compilation, ensuring that the reduction in memory is delivered incrementally over the compilation phase rather than as one final step.
n.b. It would be preferable to generate the `BytecodeFrame` instances in a tree structure during LIR generation. However, that is not possible with the current compile strategy which is why conversion to a tree is being done when the compilation has completed. There are several reasons for this. A frame chain created during LIR generation of a given `Infopoint`includes an unprocessed `values` array attached to each `BytecodeFrame`. These `values` are initialized in a second pass down the chain. Reuse of existing `BytecodeFrame` ancestors to build a shared tree structure would, at least, require this initialization pass to be merged with the construction pass. However, these `values` may also be updated by subsequent compiler phases, which also assume independent linear `BytecodeFrame` chains. So, the later passes would also need to be modified to process a shared tree and, possibly, restructure it if modifications led to further merge opportunities (this also assumes that earlier merges did not elide nodes that might later need to be distinguished) 

For the HelloWorld app used in to test previous debug info footprint improvements these changes save roughly 8Mb storage for basic debug info generation. However, for full debug info generation (where full local var info implies many more Infpoint instances) it results in a very significant 100Mb saving. The saving is roughly 75% of the memory previously used to record frame information, comprising not just repeated instances of `BytecodeFrame` but also (necessarily for equality) repeats of their associated arrays of type `JavaValue[]` and `JavaKind[]`.

### DWARF phase
1. Full method names were being stored in `Range` objects and redundantly entered into the string table for inclusion in the `.debug_string` section. The first change creates these names on demand, which only happens when performing logging.
2. A single class, `Range`, was being used to model: primary ranges that detail a top level method; leaf ranges that model an instruction range comprising instructions derived from a single caller method (whether primary or inlined) and call ranges that model an instruction range derived from an inlined call, possibly including recursively inlined ranges. The change uses three subclasses to model each type of range more succinctly, allowing each class to omit some field data:
- `PrimaryRange` and `CallRange` employ fields `firstCallee` and `lastCallee` to daisy chain child ranges but these can be omitted from `LeafRange`.
- `CallRange` and `LeafRange` employ fields `caller` and `primary` as a back link to, respectively, their direct caller range and (indirect) primary range but these fields can be omitted from `PrimaryRange`.
- `CallRange` and `LeafRange` employ field `siblingCallee` to daisy chain adjacent `Range` instances sharing the same caller but this can be omitted from `PrimaryRange`.
3.Details of a local or parameter variable value locations were being modelled with distinct instances of classes `NativeImageDebugRegisterValue`,  `NativeImageDebugStackValue` and `NativeImageDebugConstantValue` even though the same register, stack slot and, occasionally but less frequently, constant will be referenced many times. This was changed to reuse instances (across all compilations).
4. A single implementation class, `NativeImageDebugLocalValueInfo` was being used to implement interface `DebugLocalInfo` and extending interface `DebugLocalValueInfo` which model, respectively, details of a local or parameter variable’s name, type etc and details of where the variable’s value is located. Splitting the implementation into a superclass, `NativeImageDebugLocalInfo`, and subclass,  `NativeImageDebugLocalValueInfo`, allows superclass inastances to omit some fields.

For the HelloWorld app used in to test previous debug info footprint improvements these changes save roughly 6/15Mb extra storage for basic/full debug info generation. 

### Updated heap dump stats
For the specific classes affected by these updates
[Footprint3.pdf](https://github.com/oracle/graal/files/10117859/Footprint3.pdf)

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
@@ -34,6 +34,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import com.oracle.objectfile.debugentry.range.PrimaryRange;
+import com.oracle.objectfile.debugentry.range.Range;
+import com.oracle.objectfile.debugentry.range.SubRange;
 import com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugFileInfo;
 import jdk.vm.ci.meta.ResolvedJavaType;
 import org.graalvm.collections.EconomicMap;
@@ -299,7 +302,7 @@ public abstract class DebugInfoBase {
             /* Search for a method defining this primary range. */
             ClassEntry classEntry = lookupClassEntry(ownerType);
             MethodEntry methodEntry = classEntry.ensureMethodEntryForDebugRangeInfo(debugCodeInfo, this, debugContext);
-            Range primaryRange = new Range(stringTable, methodEntry, lo, hi, primaryLine);
+            PrimaryRange primaryRange = Range.createPrimary(methodEntry, lo, hi, primaryLine);
             debugContext.log(DebugContext.INFO_LEVEL, "PrimaryRange %s.%s %s %s:%d [0x%x, 0x%x]", ownerType.toJavaName(), methodName, filePath, fileName, primaryLine, lo, hi);
             classEntry.indexPrimary(primaryRange, debugCodeInfo.getFrameSizeChanges(), debugCodeInfo.getFrameSize());
             /*
@@ -427,12 +430,13 @@ public abstract class DebugInfoBase {
      * @param locationInfo
      * @param primaryRange
      * @param classEntry
+     * @param subRangeIndex
      * @param debugContext
      * @return the subrange for {@code locationInfo} linked with all its caller subranges up to the
      *         primaryRange
      */
     @SuppressWarnings("try")
-    private Range addSubrange(DebugLocationInfo locationInfo, Range primaryRange, ClassEntry classEntry, EconomicMap<DebugLocationInfo, Range> subRangeIndex, DebugContext debugContext) {
+    private Range addSubrange(DebugLocationInfo locationInfo, PrimaryRange primaryRange, ClassEntry classEntry, EconomicMap<DebugLocationInfo, Range> subRangeIndex, DebugContext debugContext) {
         /*
          * We still insert subranges for the primary method but they don't actually count as inline.
          * we only need a range so that subranges for inline code can refer to the top level line
@@ -458,7 +462,7 @@ public abstract class DebugInfoBase {
         final int line = locationInfo.line();
         ClassEntry subRangeClassEntry = lookupClassEntry(ownerType);
         MethodEntry subRangeMethodEntry = subRangeClassEntry.ensureMethodEntryForDebugRangeInfo(locationInfo, this, debugContext);
-        Range subRange = new Range(stringTable, subRangeMethodEntry, lo, hi, line, primaryRange, isTopLevel, caller);
+        SubRange subRange = Range.createSubrange(subRangeMethodEntry, lo, hi, line, primaryRange, caller, locationInfo.isLeaf());
         classEntry.indexSubRange(subRange);
         subRangeIndex.put(locationInfo, subRange);
         debugContext.log(DebugContext.DETAILED_LEVEL, "SubRange %s.%s %d %s:%d [0x%x, 0x%x] (%d, %d)",
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
@@ -309,7 +309,7 @@ public abstract class DebugInfoBase {
              * Record all subranges even if they have no line or file so we at least get a symbol
              * for them and don't see a break in the address range.
              */
-            EconomicMap<DebugLocationInfo, Range> subRangeIndex = EconomicMap.create();
+            EconomicMap<DebugLocationInfo, SubRange> subRangeIndex = EconomicMap.create();
             debugCodeInfo.locationInfoProvider().forEach(debugLocationInfo -> addSubrange(debugLocationInfo, primaryRange, classEntry, subRangeIndex, debugContext));
         }));
 
@@ -436,7 +436,7 @@ public abstract class DebugInfoBase {
      *         primaryRange
      */
     @SuppressWarnings("try")
-    private Range addSubrange(DebugLocationInfo locationInfo, PrimaryRange primaryRange, ClassEntry classEntry, EconomicMap<DebugLocationInfo, Range> subRangeIndex, DebugContext debugContext) {
+    private Range addSubrange(DebugLocationInfo locationInfo, PrimaryRange primaryRange, ClassEntry classEntry, EconomicMap<DebugLocationInfo, SubRange> subRangeIndex, DebugContext debugContext) {
         /*
          * We still insert subranges for the primary method but they don't actually count as inline.
          * we only need a range so that subranges for inline code can refer to the top level line
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/PrimaryRange.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/PrimaryRange.java
@@ -28,10 +28,8 @@ package com.oracle.objectfile.debugentry.range;
 
 import com.oracle.objectfile.debugentry.ClassEntry;
 import com.oracle.objectfile.debugentry.MethodEntry;
-import com.oracle.objectfile.debuginfo.DebugInfoProvider;
 
 public class PrimaryRange extends Range {
-    private static final DebugInfoProvider.DebugLocalInfo[] EMPTY_LOCAL_INFOS = new DebugInfoProvider.DebugLocalInfo[0];
     /**
      * The first subrange in the range covered by this primary or null if this primary as no
      * subranges.
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/PrimaryRange.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/PrimaryRange.java
@@ -41,21 +41,6 @@ public class PrimaryRange extends Range {
      * The last subrange in the range covered by this primary.
      */
     protected SubRange lastCallee;
-    /**
-     * Values for the associated method's local and parameter variables that are available or,
-     * alternatively, marked as invalid in this range.
-     */
-    private DebugInfoProvider.DebugLocalValueInfo[] localValueInfos;
-    /**
-     * The set of local or parameter variables with which each corresponding local value in field
-     * localValueInfos is associated. Local values which are associated with the same local or
-     * parameter variable will share the same reference in the corresponding array entries. Local
-     * values with which no local variable can be associated will have a null reference in the
-     * corresponding array. The latter case can happen when a local value has an invalid slot or
-     * when a local value that maps to a parameter slot has a different name or type to the
-     * parameter.
-     */
-    private DebugInfoProvider.DebugLocalInfo[] localInfos;
 
     protected PrimaryRange(MethodEntry methodEntry, int lo, int hi, int line) {
         super(methodEntry, lo, hi, line, -1);
@@ -97,51 +82,5 @@ public class PrimaryRange extends Range {
     @Override
     public boolean isLeaf() {
         return firstCallee == null;
-    }
-
-    @Override
-    public boolean includesInlineRanges() {
-        SubRange child = firstCallee;
-        while (child != null && child.isLeaf()) {
-            child = child.siblingCallee;
-        }
-        return child != null;
-    }
-
-    public int getLocalValueCount() {
-        assert !this.isPrimary() : "primary range does not have local values";
-        return localValueInfos.length;
-    }
-
-    public DebugInfoProvider.DebugLocalValueInfo getLocalValue(int i) {
-        assert !this.isPrimary() : "primary range does not have local values";
-        assert i >= 0 && i < localValueInfos.length : "bad index";
-        return localValueInfos[i];
-    }
-
-    public DebugInfoProvider.DebugLocalInfo getLocal(int i) {
-        assert !this.isPrimary() : "primary range does not have local vars";
-        assert i >= 0 && i < localInfos.length : "bad index";
-        return localInfos[i];
-    }
-
-    public void setLocalValueInfo(DebugInfoProvider.DebugLocalValueInfo[] localValueInfos) {
-        int len = localValueInfos.length;
-        this.localValueInfos = localValueInfos;
-        this.localInfos = (len > 0 ? new DebugInfoProvider.DebugLocalInfo[len] : EMPTY_LOCAL_INFOS);
-        // set up mapping from local values to local variables
-        for (int i = 0; i < len; i++) {
-            localInfos[i] = methodEntry.recordLocal(localValueInfos[i]);
-        }
-    }
-
-    public DebugInfoProvider.DebugLocalValueInfo lookupValue(DebugInfoProvider.DebugLocalInfo local) {
-        int localValueCount = getLocalValueCount();
-        for (int i = 0; i < localValueCount; i++) {
-            if (getLocal(i) == local) {
-                return getLocalValue(i);
-            }
-        }
-        return null;
     }
 }
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/PrimaryRange.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/PrimaryRange.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2022, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2022, Red Hat Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.objectfile.debugentry.range;
+
+import com.oracle.objectfile.debugentry.ClassEntry;
+import com.oracle.objectfile.debugentry.MethodEntry;
+import com.oracle.objectfile.debuginfo.DebugInfoProvider;
+
+public class PrimaryRange extends Range {
+    private static final DebugInfoProvider.DebugLocalInfo[] EMPTY_LOCAL_INFOS = new DebugInfoProvider.DebugLocalInfo[0];
+    /**
+     * The first subrange in the range covered by this primary or null if this primary as no
+     * subranges.
+     */
+    protected SubRange firstCallee;
+    /**
+     * The last subrange in the range covered by this primary.
+     */
+    protected SubRange lastCallee;
+    /**
+     * Values for the associated method's local and parameter variables that are available or,
+     * alternatively, marked as invalid in this range.
+     */
+    private DebugInfoProvider.DebugLocalValueInfo[] localValueInfos;
+    /**
+     * The set of local or parameter variables with which each corresponding local value in field
+     * localValueInfos is associated. Local values which are associated with the same local or
+     * parameter variable will share the same reference in the corresponding array entries. Local
+     * values with which no local variable can be associated will have a null reference in the
+     * corresponding array. The latter case can happen when a local value has an invalid slot or
+     * when a local value that maps to a parameter slot has a different name or type to the
+     * parameter.
+     */
+    private DebugInfoProvider.DebugLocalInfo[] localInfos;
+
+    protected PrimaryRange(MethodEntry methodEntry, int lo, int hi, int line) {
+        super(methodEntry, lo, hi, line, -1);
+        this.firstCallee = null;
+        this.lastCallee = null;
+    }
+
+    @Override
+    public int getFileIndex() {
+        ClassEntry owner = methodEntry.ownerType();
+        return owner.localFilesIdx(getFileEntry());
+    }
+
+    @Override
+    public boolean isPrimary() {
+        return true;
+    }
+
+    @Override
+    protected void addCallee(SubRange callee) {
+        assert this.lo <= callee.lo;
+        assert this.hi >= callee.hi;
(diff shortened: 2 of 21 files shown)


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/PrimaryRange.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public DebugInfoProvider.DebugLocalValueInfo lookupValue(DebugInfoProvider.DebugLocalInfo local) {
+        int localValueCount = getLocalValueCount();
+        for (int i = 0; i < localValueCount; i++) {
+            if (getLocal(i) == local) {
+                return getLocalValue(i);
+            }
+        }
+        return null;
+    }
```
zakkak, line 146: Since primary ranges do not have local values and vars these methods seem wrong

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/SubRange.java
```diff
@@ -0,0 +1,140 @@
...
+    @Override
+    public abstract boolean isLeaf();
+
+    public int getLocalValueCount() {
+        assert !this.isPrimary() : "primary range does not have local values";
+        return localValueInfos.length;
+    }
+
+    public DebugInfoProvider.DebugLocalValueInfo getLocalValue(int i) {
+        assert !this.isPrimary() : "primary range does not have local values";
```
zakkak, line 106:
```suggestion
```

redundant?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/SubRange.java
```diff
@@ -0,0 +1,140 @@
...
+        assert i >= 0 && i < localValueInfos.length : "bad index";
+        return localValueInfos[i];
+    }
+
+    public DebugInfoProvider.DebugLocalInfo getLocal(int i) {
+        assert !this.isPrimary() : "primary range does not have local vars";
```
zakkak, line 112:
```suggestion
```

redundant?

zakkak (changes requested):
Great to see a follow up on this topic! 
Thank you @adinn 

I have added a few comments for your consideration.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/code/SubstrateBackend.java
```diff
@@ -152,14 +164,83 @@ public CompilationResult newCompilationResult(CompilationIdentifier compilationI
         return new CompilationResult(compilationIdentifier, name) {
             @Override
             public void close(OptionValues options) {
-                /*
-                 * Do nothing, we do not want our CompilationResult to be closed because we
-                 * aggregate all data items and machine code in the native image heap.
-                 */
+                FrameOptimizer.compressInfoPoints(this);
```
zakkak, line 167: I wonder if it's worth skipping this altogether when debug info generation is disabled.

adinn:
Hmm, it seems that there is an error in `BytecodeFrame.equals` which is defined as follows:

```
    public boolean equals(Object obj) {
        if (this == obj) {
            return true;
        }
        if (obj instanceof BytecodeFrame && super.equals(obj)) {
            BytecodeFrame that = (BytecodeFrame) obj;
            // @formatter:off
            if (this.duringCall == that.duringCall &&
                this.rethrowException == that.rethrowException &&
                this.numLocals == that.numLocals &&
                this.numLocks == that.numLocks &&
                this.numStack == that.numStack &&
                Arrays.equals(this.values, that.values)) { <---- does not compare slotKinds!
                return true;
            }
            // @formatter:off
            return true;    <---- hmm, this is not going to work!
        }
        return false;
    }
```

I believe this is what is causing the gate failure when building sulong -- two frames with different settings for duringCall are being falsely equated and hence wrongly merged. I'll modify the hash map to use a key which wraps the BytecodeFrame and does a correct equality test. This may well lead to a smaller footprint saving so I will recheck that too.

adinn:
Ok, so my first version cut the wrapper type correctly tested the first frame but still relied on the broken equals method to check caller frames. This latest version performs the missing checks all the way up the caller chain, bypassing the equals method altogether. In consequence, it's a little fragile but it also performs better as it delays recursive checking of callers as long as possible. This version includes asserts to verify the copied chains are valid. As well as the frame and JavaKind/Value array gains it also provides a few more savings on DWARF phase data.

adinn:
Oops, I reverted parts of the fix to the comparator class before the last push. New version coming.

adinn:
@olpaw This now builds the llvmvm and pythonvm (.so) libraries without problem. It also passed the gate. So, I believe it is ready for review.

adinn:
n.b. I'll address Foivos review comments once I have yours.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVLineRecordBuilder.java
```diff
@@ -66,9 +67,9 @@ CVLineRecord build(CompiledMethodEntry entry) {
         debug("CVLineRecord.computeContents: processing primary range %s", primaryRange);
 
         processRange(primaryRange);
-        Iterator<Range> iterator = compiledEntry.leafRangeIterator();
+        Iterator<SubRange> iterator = compiledEntry.leafRangeIterator();
```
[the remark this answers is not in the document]
adinn: Well, perhaps but I don't see any reason to just now. These subclasses are not currently exposed outside of the type hierarchy (they are declared as protected) because their existence is merely an implementation detail. They both implement the same API, defined at the level of SubRange. So, even if a client uses what can only be a LeafRange via this API it cannot invoke a behaviour that is only available for a LeafRange.
What I can do though is simplify some of the behaviours that are still distributed across all Range subclasses e.g. A Primary does not need to store local values Method includesInlineRanges can be defined once in Range, rather than in three different versions in Primary, Call and Leaf.
I'll do that in the next reworking.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/code/SubstrateBackend.java
adinn: Perhaps. However, since there will be very few infopoints it will not cost much. What I must do is update the comment.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debuginfo/DebugInfoProvider.java
```diff
@@ -274,7 +275,7 @@ public interface DebugMethodInfo extends DebugMemberInfo {
 
     /**
      * Access details of a compiled top level or inline method producing the code in a specific
-     * {@link com.oracle.objectfile.debugentry.Range}.
+     * {@link Range}.
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -35,6 +35,7 @@
 import java.util.stream.Stream;
 import java.util.function.Predicate;
 
+import com.oracle.objectfile.debugentry.range.SubRange;
```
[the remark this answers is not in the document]
adinn: Fixed

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
```diff
@@ -419,12 +422,13 @@ public ClassEntry lookupObjectClass() {
...
      * @param primaryRange
      * @param classEntry
+     * @param subRangeIndex
      * @param debugContext
      * @return the subrange for {@code locationInfo} linked with all its caller subranges up to the
      *         primaryRange
      */
     @SuppressWarnings("try")
-    private Range addSubrange(DebugLocationInfo locationInfo, Range primaryRange, ClassEntry classEntry, EconomicMap<DebugLocationInfo, Range> subRangeIndex, DebugContext debugContext) {
+    private Range addSubrange(DebugLocationInfo locationInfo, PrimaryRange primaryRange, ClassEntry classEntry, EconomicMap<DebugLocationInfo, Range> subRangeIndex, DebugContext debugContext) {
```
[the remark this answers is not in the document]
adinn: Fixed

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/SubRange.java
```diff
@@ -0,0 +1,140 @@
...
+
+import com.oracle.objectfile.debugentry.ClassEntry;
+import com.oracle.objectfile.debugentry.MethodEntry;
+import com.oracle.objectfile.debuginfo.DebugInfoProvider;
+
+public abstract class SubRange extends Range {
+    private static final DebugInfoProvider.DebugLocalInfo[] EMPTY_LOCAL_INFOS = new DebugInfoProvider.DebugLocalInfo[0];
+    /**
+     * This is null for a primary range. For sub ranges it holds the root of the call tree they
+     * belong to.
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/Range.java
```diff
@@ -66,143 +69,75 @@
...
+     * @param caller the range for which this is a subrange, either an inlined call range or the
+     *            primary range.
+     * @param isLeaf true if this is a leaf range with no subranges
+     * @return a new subrange to be linked into the range tree below the primary range.
      */
-    public Range(StringTable stringTable, MethodEntry methodEntry, int lo, int hi, int line) {
-        this(stringTable, methodEntry, lo, hi, line, null, false, null);
+    public static SubRange createSubrange(MethodEntry methodEntry, int lo, int hi, int line, PrimaryRange primary, Range caller, boolean isLeaf) {
+        assert primary != null;
+        assert primary.isPrimary();
```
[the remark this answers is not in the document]
adinn: Removed

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/PrimaryRange.java
adinn: Yes, they are gone.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/SubRange.java
adinn: Removed

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/SubRange.java
adinn: Yes

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/range/SubRange.java
```diff
@@ -0,0 +1,140 @@
...
+        // the primary range's class entry indexes all files defined by the compilation unit
+        Range primaryRange = (isPrimary() ? this : getPrimary());
+        ClassEntry owner = primaryRange.methodEntry.ownerType();
+        return owner.localFilesIdx(getFileEntry());
+    }
+
+    @Override
+    public boolean isPrimary() {
+        return getPrimary() == null;
+    }
```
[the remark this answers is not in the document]
adinn: Fixed

zakkak (approved):
LGTM, thanks @adinn !

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/code/SubstrateBackend.java
```diff
@@ -152,14 +166,220 @@ public CompilationResult newCompilationResult(CompilationIdentifier compilationI
...
             }
         };
     }
 
+    private static class FrameOptimizer {
+        /**
+         * Wrapper class that makes up for deficiencies in the implementation of
+         * BytecodeFrame.equals.
+         */
+        private static class BytecodeFrameComparator {
```
olpaw, line 179: This is a very unusual implementation of a `Comparator` in Java. Usually you implement a `java.util.Comparator` class specialized to the type you want to compare. E.g. see `com.oracle.svm.hosted.meta.HostedUniverse.TypeComparator`. Then you have one comparator instance (and not one for each `BytecodeFrame`) that you use with the two `BytecodeFrames` that you want to compare.

olpaw: Since the only state you have in `BytecodeFrameComparator` is the `BytecodeFrame` you want to compare against I suggest you choose an implementation that does not require creating a `BytecodeFrameComparator` for each `BytecodeFrame` you want to compare against.

Also the term Comparator here is misleading because you only care about equality and hash-code so this isn't a Comparator in the sense of `java.util.Comparator`. I suggest replacing this class with static methods that take two `BytecodeFrame` instances. And in the `frameMap` you have two `BytecodeFrames` instead.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/code/SubstrateBackend.java
```diff
@@ -152,14 +166,220 @@ public CompilationResult newCompilationResult(CompilationIdentifier compilationI
         return new CompilationResult(compilationIdentifier, name) {
             @Override
             public void close(OptionValues options) {
-                /*
```
olpaw, line 155: Please keep the comment but amend it with: "Do nothing, except .....".

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/code/SubstrateBackend.java
```diff
@@ -152,14 +166,220 @@ public CompilationResult newCompilationResult(CompilationIdentifier compilationI
...
+                    }
+                    thisFrame = thisFrame.caller();
+                    thatFrame = thatFrame.caller();
+                }
+                return thisFrame == thatFrame;
+            }
+        }
+
+        private static void compressInfoPoints(CompilationResult result) {
+            List<Infopoint> infopointsCopy = new ArrayList<>();
```
olpaw, line 240: ArrayList has a java.util.ArrayList#ArrayList(java.util.Collection<? extends E>) constructor that copies the elements.

olpaw (changes requested):
This PR breaks truffle image building.

This can be reproduced easily with:
```
cd graal/vm
mx --env ce-js build
```
This will end up in
```
[libjsvm:31055] [7/8] Layouting methods...    [*]                                                        (0.2s @ 3.74GB)

[libjsvm:31055] ------------------------------------------------------------------------------------------------------------------------
[libjsvm:31055]                         7.3s (6.2% of total time) in 66 GCs | Peak RSS: 7.35GB | CPU load: 7.73
[libjsvm:31055] ------------------------------------------------------------------------------------------------------------------------
[libjsvm:31055] Produced artifacts:
[libjsvm:31055]  /home/pwoegere/OLabs/issues/GR-43600/graal/sdk/mxbuild/linux-amd64/libjsvm.so.image/svm_err_b_20230117T111612.116_pid31055.md (build_info)
[libjsvm:31055] ========================================================================================================================
[libjsvm:31055] Failed generating 'libjsvm' after 1m 57s.
[libjsvm:31055] 
[libjsvm:31055] The build process encountered an unexpected error:
[libjsvm:31055] 
com.oracle.svm.core.util.VMError$HostedError: Encoding two infopoints at same offset. Conflicting infopoint: 41[HostedMethod<Throwable.toString -> AnalysisMethod<java.lang.Throwable.toString -> HotSpotMethod<Throwable.toString()>>>][]] com.ibm.icu.impl.Assert.fail(Assert.java:14) [bci: 1] #locals=1 #expr=0
	at com.oracle.svm.core.code.CodeInfoEncoder.addMethod(CodeInfoEncoder.java:213)
	at com.oracle.svm.hosted.image.NativeImageCodeCache.encodeMethod(NativeImageCodeCache.java:392)
	at com.oracle.svm.hosted.image.NativeImageCodeCache.buildRuntimeMetadata(NativeImageCodeCache.java:252)
	at com.oracle.svm.hosted.image.LIRNativeImageCodeCache.layoutMethods(LIRNativeImageCodeCache.java:221)
	at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:663)
	at com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:538)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:401)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:609)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.start(NativeImageGeneratorRunner.java:133)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:93)
	at com.oracle.svm.hosted.NativeImageGeneratorRunner$JDK9Plus.main(NativeImageGeneratorRunner.java:623)
````

I suspect `FrameOptimizer.compressInfoPoints` breaks some assumptions about infopoints that are important for runtime compilation.

File: substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/code/SubstrateBackend.java (hunk 12 above)
Comment (adinn) [reply substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/code/SubstrateBackend.java:179]: Yes, this is not actually a Comparator in the sense of `java.util.Comparator`. It was just a wrapper type so I could redefine the equals and hash methods.

I see that `EconomicMap.create` accepts an argument of type `Equivalence` to define the `equals` and `hash` operations used to manage map inserts and lookups (the map is actually being used as a hashset here). As you say, I believe all I need to do here is implement `Equivalence` appropriately with, say, class `BytecodeFrameEquivalence` and then create an EconomicMap<BytecodeFrame,BytecodeFrame> parameterised passing an instance of `BytecodeFrameEquivalence` to the create call. I will only ever need one instance across all uses.

**Comment by adinn** [comment]
> This PR breaks truffle image building.
. . .
> I suspect FrameOptimizer.compressInfoPoints breaks some assumptions about infopoints that are important for runtime compilation.

So it seems.

This looks exactly like the error that occurred with the python library. In that case the problem happened when I replaced a `Call` infopoint with a compressed replacement.

The original `Call` infopoint had been saved in the list of valid call deopt states (by a call to `CompilationResult.recordCallInvalidForDeoptimization`). When a `Call` infopoint is in the saved set the test for it being a `DeoptEntry` succeeds which means it is not considered for processing by `FrameInfoEncoder.addDebugInfo`.

If a replacement for the saved infopoint is not also added to the saved set then the replacement will get processed by `addDebugInfo` and we can run into a subsequent infopoint that refers to the same location. This was fixed by the code at SubstrateBackend:249.

                if (newInfopoint != infopoint && infopoint instanceof Call && !result.isValidCallDeoptimizationState((Call) infopoint)) {
                    // result must discount the replacement infopoint as a deopt location
                    result.recordCallInvalidForDeoptimization((Call) newInfopoint);
                }

That avoided the problem when building libpython. I'll have to debug to see why that does not work for the js case.

File: substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/code/SubstrateBackend.java [hunk 15]
```diff
@@ -152,14 +166,220 @@ public CompilationResult newCompilationResult(CompilationIdentifier compilationI
...
+                    return false;
+                }
+                BytecodeFrameComparator that = (BytecodeFrameComparator) obj;
+                BytecodeFrame thisFrame = frame;
+                BytecodeFrame thatFrame = that.frame;
+                while (thisFrame != thatFrame && thisFrame != null && thatFrame != null) {
+                    if (thisFrame.getBCI() == thatFrame.getBCI() &&
+                                    Objects.equals(thisFrame.getMethod(), thatFrame.getMethod()) &&
+                                    Objects.equals(thisFrame.caller(), thatFrame.caller()) &&
+                                    // BytecodeFrame.equals returns true even when these tests fail!
```
christianwimmer, line 207: @dougxc is fixing this in https://bugs.openjdk.org/browse/JDK-8300590

If you find JVMCI bugs like this, please let us know instead of working around the problem. We could have had the proper fix in all the just released updates of the JDK already.

adinn: Hi @christianwimmer. Apologies for not raising this as a bug. I had assumed that there was a good reason for disabling this check in the subtype implementation. In future I'll attempt to curb my natural assumption that the GraalVM team must know what they are doing! ;-)

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/graal/code/SubstrateBackend.java
```diff
@@ -152,14 +166,220 @@ public CompilationResult newCompilationResult(CompilationIdentifier compilationI
         return new CompilationResult(compilationIdentifier, name) {
             @Override
             public void close(OptionValues options) {
-                /*
-                 * Do nothing, we do not want our CompilationResult to be closed because we
-                 * aggregate all data items and machine code in the native image heap.
-                 */
+                FrameOptimizer.compressInfoPoints(this);
```
olpaw, line 169: A better place to to do the deduplication of BytecodeFrames (incrementally) would be to add it to `org.graalvm.compiler.code.CompilationResult#addInfopoint` and the again later when there is no more mutation in the `jdk.vm.ci.code.BytecodeFrame#values` array.

Its very unfortunate that we have a mutable state in `BytecodeFrame` while at the same time the `jdk.vm.ci.code.DebugInfo#bytecodePosition` field is final. This forces us to instantiate new Infopoints to be able to have that `bytecodePosition` field updated. This creates new problems since the infopoint instances are at that point already referenced at other places as well requiring shenanigans like in line 250. @adinn @christianwimmer

adinn: @olpaw @christianwimmer Yes, my first attempt was to deduplicate `BytecodeFrame` instances when they are created so that we never incur the penalty of modelling what we know will actually be a tree as a set of paths. I quickly found that they are not stable at create time so had to pick a later point to convert from one representation to the other.

I chose to do it in (the Substrate override of) `CompilationResult.close` because that means the resulting footprint compression is pipelined with compilation rather than delayed until compilation completes. However, that choice requires wholesale replacement of the set of Infopoint instances in `CompilationResult.infopoints` after the collection has been assembled, at which point references to the old values may have been stashed away -- e.g in list `CompilationResult.invalidCallDeoptimizationStates`. Doing it earlier, say at the call to `addInfopoint`,  could mean we can swap in a replacememt before it is used elsewhere.

@olpaw I am not sure what you mean by this:

> A better place to do the deduplication of BytecodeFrames (incrementally) would be to add it to org.graalvm.compiler.code.CompilationResult#addInfopoint and then again later when there is no more mutation in the jdk.vm.ci.code.BytecodeFrame#values array.

Does the `values` array get modified after the call to `addInfopoint`? If so then how can we deduplicate frames? at the call? Will we not risk eliding two frames that later turn out to be different because they end up with different values?

n.b. I have not been able to find any use of an `Infopoint` added via a call to `addInfopoint` other than via `CompilationResult.recordCall`. However, in all the latter cases the `Call` returned by this seems only to be used to update `CompilationResult.invalidCallDeoptimizationStates`. Is there some other use of a newly added `Infopoint` that I am missing?

Also, I have run command ` mx --env ce-js build` about a dozen times and not been able to see an error. Is the problem very infrequent or could the failure to manifest be in some way related to my setup?

adinn:
I have amended this PR to no longer performs deduplication of BytecodeFrame instances. That will be presented in a separate PR which can be tested and reviewed independently of the changes remaining in this PR.

adinn:
@olpaw This should be ok to retest and merge now that the frame dedup has been removed.

I will address the outstanding review comments which relate to that removed code when I open a new PR.

olpaw (approved):
Running in internal CI now

olpaw:
PR is in the merge queue

PR #5511, opened 2022-11-28
Support delegating SVM's feature

ziyilin:
Using DelegateFeature to delegate a compatible SVM feature in standalone pointsto, so that the existing SVM feature can be reused.

DashboardDumpFeature is also modified for invoking from standalone pointsto.

This PR is part of https://github.com/oracle/graal/pull/4375

### DIFF
--- a/substratevm/mx.substratevm/suite.py
+++ b/substratevm/mx.substratevm/suite.py
@@ -490,6 +490,7 @@ suite = {
             "sourceDirs": ["src"],
             "dependencies": [
                 "com.oracle.graal.pointsto",
+                "sdk:GRAAL_SDK",
             ],
             "requires" : [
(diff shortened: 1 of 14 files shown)


cstancu:
I don't understand the usefulness of this feature. The points-to analysis should not depend on SVM features. Also, you should not depend on the `dashboard` code, that is currently Native Image specific.

ziyilin:
This feature dumps standalone analysis result so that it can be visually inspected in the dashboard. 
It is an optional feature. All dependencies to SVM are made by reflection. If user would like to dump the results to dashboard,  he/she must explicitly provide svm.jar.
Furthermore, it provides a mechanism to invoke SVM features from standalone analysis. Though I don't see other usefulness besides DashboardDump for now.

zakkak:

Issue #5635, opened 2022-12-16
Windows Debug Information for local variables.


stooke:
Windows debugger support now includes line numbers and type information.  Static variables are accessible, but local variables and function parameters are not, unless the user casts a register to a type in the debugger.

I would like to add local variable information to Graal, based on the Dwarf support implemented by @adinn and the type info support #3732  that was just integrated.  I propose a two-stage process:  in the first pull request, I would add the new CodeView records required to implement this support, and simple code to define function parameters upon function entry.
A second pull request would add full local variable support.

I think this support would greatly benefit Graal users and developers who are working on Windows.

I have a preliminary implementation of the first stage almost ready for review, if it's decided to move forward.

PR #5638, opened 2022-12-16
Windows Debug Information for local variables - prerequisite record definitions

stooke:
This PR implements issue #5635 by adding Windows CodeView records support to define local variables.  

Method parameter variables (including 'this') are defined as a demonstration.  No other local variable information is implemented by this PR as this PR is not aware of the lifetime of variables or stack frame information.  Method parameter variables are available to the debugger only upon entry to a method.

There is a body of unused code (symbol record definitions) in this PR - it will be required for local variable definition in the followup PR.

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
@@ -27,9 +27,12 @@
 package com.oracle.objectfile.pecoff.cv;
 
 import com.oracle.objectfile.debugentry.ClassEntry;
+import org.graalvm.compiler.debug.GraalError;
 
 import java.util.HashMap;
 import java.util.Map;
+
+import static com.oracle.objectfile.pecoff.cv.CVDebugConstants.S_BLOCK32;
 
 /*
  * A CVSymbolSubrecord is a record in a DEBUG_S_SYMBOL record within a .debug$S section within a PECOFF file.
@@ -301,6 +304,181 @@ abstract class CVSymbolSubrecord {
         }
     }
 
+
+    public static class CVSymbolRegisterRecord extends CVSymbolSubrecord {
+
+        private final String name;
+        private final int typeIndex;
+        private final short register;
+
+        CVSymbolRegisterRecord(CVDebugInfo debugInfo, String name, int typeIndex, short register) {
+            super(debugInfo, CVDebugConstants.S_REGISTER);
+            this.name = name;
+            this.typeIndex = typeIndex;
+            this.register = register;
+        }
+
+        @Override
+        protected int computeContents(byte[] buffer, int initialPos) {
+            int pos = CVUtil.putInt(typeIndex, buffer, initialPos);
+            pos = CVUtil.putShort(register, buffer, pos);
+            pos = CVUtil.putUTF8StringBytes(name, buffer, pos);
+            return pos;
+        }
+
+        @Override
+        public String toString() {
+            return String.format("S_REGISTER   name=%s  r%d type=0x%x)", name, register, typeIndex);
+        }
+    }
+
+    public static class CVSymbolLocalRecord extends CVSymbolSubrecord {
+
+        private final String name;
+        private final int typeIndex;
+        private final int flags;
+
+        CVSymbolLocalRecord(CVDebugInfo debugInfo, String name, int typeIndex, int flags) {
+            super(debugInfo, CVDebugConstants.S_LOCAL);
+            this.name = name;
+            this.typeIndex = typeIndex;
+            this.flags = flags;
+        }
+
+        @Override
+        protected int computeContents(byte[] buffer, int initialPos) {
+            int pos = CVUtil.putInt(typeIndex, buffer, initialPos);
+            pos = CVUtil.putShort((short) flags, buffer, pos);
+            pos = CVUtil.putUTF8StringBytes(name, buffer, pos);
+            return pos;
+        }
+
+        @Override
+        public String toString() {
+            return String.format("S_LOCAL   name=%s type=0x%x flags=0x%x)", name, typeIndex, flags);
+        }
+    }
+
+    private abstract static class CVSymbolDefRangeBase extends CVSymbolSubrecord {
+
+        protected final int gapCount;
+        protected final String procName;
+        protected final int procOffset;
+        protected final short range;
+
+        protected CVSymbolDefRangeBase(CVDebugInfo debugInfo, short recordType, String procName, int procOffset, short range) {
+            super(debugInfo, recordType);
+            this.procName = procName;
+            this.procOffset = procOffset;
+            this.range = range;
+            this.gapCount = 0;
+            if (procOffset != 0) {
+                GraalError.shouldNotReachHere("procOffset not implemented");
+            }
+        }
+
+        void addGap() {
+            /* TODO */
+        }
+
+        int computeRange(byte[] buffer, int initialPos) {
+            /* Emit CV_LVAR_ADDR_RANGE. */
+            // int pos = cvDebugInfo.getCVSymbolSection().markRelocationSite(buffer, initialPos,
+            // procName, (long) procOffset);
+            int pos = cvDebugInfo.getCVSymbolSection().markRelocationSite(buffer, initialPos, procName);
+            pos = CVUtil.putShort(range, buffer, pos);
+            return pos;
+        }
+
+        int computeGaps(@SuppressWarnings("unused") byte[] buffer, int initialPos) {
+            /* Emit gaps. Not Yet implemented. */
+            assert gapCount == 0;
+            return initialPos;
+        }
+
+        int computeRangeAndGaps(byte[] buffer, int initialPos) {
+            int pos = computeRange(buffer, initialPos);
+            return computeGaps(buffer, pos);
+        }
+    }
+
+    public static class CVSymbolDefRangeFramepointerRelFullScope extends CVSymbolSubrecord {
+
+        private final int frameOffset;
+
+        CVSymbolDefRangeFramepointerRelFullScope(CVDebugInfo debugInfo, int frameOffset) {
+            super(debugInfo, CVDebugConstants.S_DEFRANGE_FRAMEPOINTER_REL_FULL_SCOPE);
+            this.frameOffset = frameOffset;
+        }
+
+        @Override
+        protected int computeContents(byte[] buffer, int initialPos) {
+            return CVUtil.putInt(frameOffset, buffer, initialPos);
+        }
+
+        @Override
+        public String toString() {
+            return String.format("S_DEFRANGE_FRAMEPOINTER_REL_FULL_SCOPE frameOffset=0x%x", frameOffset);
+        }
+    }
+
+    public static class CVSymbolDefRangeRegisterRecord extends CVSymbolDefRangeBase {
+
+        private final short register;
+        private final short attr;
+
+        CVSymbolDefRangeRegisterRecord(CVDebugInfo debugInfo, short register, String procName, int procOffset, int length) {
+            super(debugInfo, CVDebugConstants.S_DEFRANGE_REGISTER, procName, procOffset, (short) length);
+            this.register = register;
+            this.attr = 0;
+        }
+
+        @Override
+        protected int computeContents(byte[] buffer, int initialPos) {
+            int pos = CVUtil.putShort(register, buffer, initialPos);
+            pos = CVUtil.putShort(attr, buffer, pos);
+            pos = computeRangeAndGaps(buffer, pos);
+            return pos;
+        }
+
+        @Override
+        public String toString() {
+            return String.format("S_DEFRANGE_REGISTER r%d attr=0x%x %s+0x%x range=0x%x %d gaps)", register, attr, procName, procOffset, range, gapCount);
+        }
+    }
+
+    public static class CVSymbolDefRangeRegisterRel extends CVSymbolDefRangeBase {
+
+        private final short baseRegister;
+        private final short spilledUdtMember;
+        private final short parentOffset;
+        private final int bpOffset;
+
+        CVSymbolDefRangeRegisterRel(CVDebugInfo debugInfo, String procName, int procOffset, short range, short baseRegister) {
+            super(debugInfo, CVDebugConstants.S_DEFRANGE_REGISTER_REL, procName, procOffset, range);
+            this.baseRegister = baseRegister;
+            this.spilledUdtMember = 0;
+            this.parentOffset = 0;
+            this.bpOffset = 0;
+        }
+
+        @Override
+        protected int computeContents(byte[] buffer, int initialPos) {
+            int pos = CVUtil.putShort(baseRegister, buffer, initialPos);
+            /* bitfield spilledUDTMember : 1, pad : 3, parentOffset : 12 */
+            pos = CVUtil.putShort((short) (spilledUdtMember + (parentOffset << 4)), buffer, pos);
+            pos = CVUtil.putInt(bpOffset, buffer, pos);
+            pos = computeRangeAndGaps(buffer, pos);
+            return pos;
+        }
+
+        @Override
+        public String toString() {
+            return String.format("S_DEFRANGE_REGISTER_REL r%d spilled=%d parentOffset=0x%x %s+0x%x range=0x%x %d gaps)", baseRegister, spilledUdtMember, parentOffset, procName, procOffset, range,
+                            gapCount);
+        }
+    }
+
     public static class CVSymbolRegRel32Record extends CVSymbolSubrecord {
 
         private final String name;
@@ -328,6 +506,61 @@ abstract class CVSymbolSubrecord {
         @Override
         public String toString() {
             return String.format("S_REGREL32   name=%s  offset=(r%d + 0x%x) type=0x%x)", name, register, offset, typeIndex);
+        }
+    }
+
+    public static class CVSymbolDefRangeFramepointerRel extends CVSymbolDefRangeBase {
+
+        private final int fpOffset;
+
+        CVSymbolDefRangeFramepointerRel(CVDebugInfo debugInfo, String name, int procOffset, short range, int fpOffset) {
+            super(debugInfo, CVDebugConstants.S_DEFRANGE_FRAMEPOINTER_REL, name, procOffset, range);
+            this.fpOffset = fpOffset;
+        }
+
+        @Override
+        protected int computeContents(byte[] buffer, int initialPos) {
+            int pos = CVUtil.putInt(fpOffset, buffer, initialPos);
+            pos = computeRangeAndGaps(buffer, pos);
+            return pos;
+        }
+
+        @Override
+        public String toString() {
+            return String.format("S_DEFRANGE_FRAMEPOINTER_REL  name=%s+0x%x fpOfffset=0x%x)", procName, procOffset, fpOffset);
+        }
+    }
+
+    public static class CVSymbolBlock32Record extends CVSymbolSubrecord {
+        // K32 name= parent=0x0 end=0x0 len=0x8 codeoffset=0x0:70
+        // reloc addr=0x0008e5 type=11 sym=main IMAGE_REL_AMD64_SECREL(0x0b)
+        final String procName;
+        String name = "";
+        int parent = 0;
+        int end = 0;
+        int len = 8;
+        int offset = 0;
+        short segment = 0;
+
+        CVSymbolBlock32Record(CVDebugInfo cvDebugInfo, String procName) {
+            super(cvDebugInfo, S_BLOCK32);
+            this.procName = procName;
+            /* TODO probably need to implement procOffset here, too. */
+        }
+
+        @Override
+        protected int computeContents(byte[] buffer, int initialPos) {
+            int pos = CVUtil.putInt(parent, buffer, initialPos);
+            pos = CVUtil.putInt(end, buffer, pos);
+            pos = CVUtil.putInt(len, buffer, pos);
+            pos = cvDebugInfo.getCVSymbolSection().markRelocationSite(buffer, pos, procName);
+            pos = CVUtil.putUTF8StringBytes(name, buffer, pos);
+            return pos;
+        }
+
+        @Override
+        public String toString() {
+            return String.format("S_BLOCK32   name=%s parent=0x%x end=0x%x len=0x%x offset=0x%x:0x%x", name, parent, end, len, segment, offset);
         }
     }
 
@@ -388,6 +621,17 @@ abstract class CVSymbolSubrecord {
     }
 
     public static final class CVSymbolFrameProcRecord extends CVSymbolSubrecord {
+
+        /* TODO: This may change in the presence of isolates. */
+
+        /* Async exception handling (vc++ uses 1, clang uses 0). */
+        public static final int FRAME_ASYNC_EH = 1 << 9;
+
+        /* Local base pointer = SP (0=none, 1=sp, 2=bp 3=r13). */
+        public static final int FRAME_LOCAL_BP = 1 << 14;
+
+        /* Param base pointer = SP. */
+        public static final int FRAME_PARAM_BP = 1 << 16;
 
         private final int framelen;
         private final int padLen;
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
@@ -304,7 +304,7 @@ abstract class CVSymbolSubrecord {
         }
     }
 
-
+    @SuppressWarnings("unused")
     public static class CVSymbolRegisterRecord extends CVSymbolSubrecord {
 
         private final String name;
@@ -379,6 +379,7 @@ abstract class CVSymbolSubrecord {
 
         void addGap() {
             /* TODO */
+            GraalError.shouldNotReachHere("addGap() not implemented");
         }
 
         int computeRange(byte[] buffer, int initialPos) {
@@ -391,7 +392,7 @@ abstract class CVSymbolSubrecord {
         }
 
         int computeGaps(@SuppressWarnings("unused") byte[] buffer, int initialPos) {
-            /* Emit gaps. Not Yet implemented. */
+            /* Emit gaps. Not yet implemented. */
             assert gapCount == 0;
             return initialPos;
         }
@@ -447,6 +448,7 @@ abstract class CVSymbolSubrecord {
         }
     }
 
+    @SuppressWarnings("unused")
     public static class CVSymbolDefRangeRegisterRel extends CVSymbolDefRangeBase {
 
         private final short baseRegister;
@@ -509,6 +511,7 @@ abstract class CVSymbolSubrecord {
         }
     }
 
+    @SuppressWarnings("unused")
     public static class CVSymbolDefRangeFramepointerRel extends CVSymbolDefRangeBase {
 
         private final int fpOffset;
@@ -531,6 +534,7 @@ abstract class CVSymbolSubrecord {
         }
     }
 
+    @SuppressWarnings("unused")
     public static class CVSymbolBlock32Record extends CVSymbolSubrecord {
         // K32 name= parent=0x0 end=0x0 len=0x8 codeoffset=0x0:70
         // reloc addr=0x0008e5 type=11 sym=main IMAGE_REL_AMD64_SECREL(0x0b)
@@ -545,7 +549,7 @@ abstract class CVSymbolSubrecord {
         CVSymbolBlock32Record(CVDebugInfo cvDebugInfo, String procName) {
             super(cvDebugInfo, S_BLOCK32);
             this.procName = procName;
-            /* TODO probably need to implement procOffset here, too. */
+            /* TODO - may need to implement procOffset here. */
         }
 
         @Override
@@ -622,7 +626,7 @@ abstract class CVSymbolSubrecord {
 
     public static final class CVSymbolFrameProcRecord extends CVSymbolSubrecord {
 
-        /* TODO: This may change in the presence of isolates. */
+        /* This may change in the presence of isolates. */
 
         /* Async exception handling (vc++ uses 1, clang uses 0). */
         public static final int FRAME_ASYNC_EH = 1 << 9;
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
@@ -30,15 +30,63 @@ import com.oracle.objectfile.SectionName;
 import com.oracle.objectfile.debugentry.ClassEntry;
 import com.oracle.objectfile.debugentry.CompiledMethodEntry;
 import com.oracle.objectfile.debugentry.FieldEntry;
+import com.oracle.objectfile.debugentry.MethodEntry;
 import com.oracle.objectfile.debugentry.Range;
 import com.oracle.objectfile.debugentry.TypeEntry;
+import org.graalvm.compiler.debug.GraalError;
 
 import java.lang.reflect.Modifier;
+import java.util.ArrayList;
 
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_CL;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_CX;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_DI;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_DIL;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_DL;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_DX;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_ECX;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_EDI;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_EDX;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_ESI;
 import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_R8;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_R8B;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_R8D;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_R8W;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_R9;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_R9B;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_R9D;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_R9W;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_RCX;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_RDI;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_RDX;
+import static com.oracle.objectfile.pecoff.cv.CVConstants.CV_AMD64_RSI;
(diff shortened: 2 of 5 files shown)


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -152,6 +190,86 @@ private void buildFunction(CompiledMethodEntry compiledEntry) {
...
+            gpRegisterIndex++;
+        }
+
+        /* define function parameters (p1, p2...) according to the calling convention */
+        for (int i = 0; i < method.getParamCount(); i++) {
+            final TypeEntry paramType = method.getParamType(i);
+            final int typeIndex = cvDebugInfo.getCVTypeSection().addTypeRecords(paramType).getSequenceNumber();
+            final String paramName = "p" + (i + 1);
+            if (typeIndex == T_REAL64 || typeIndex == T_REAL32) {
+                /* floating point primitive */
```
adinn, line 220: These two cases can also be distinguished by first testing for a primitive type (i.e. folding into the branch below) and then checking whether the primitive type entry flags include NUMERIC but omit INTEGRAL.

adinn (approved):
Sorry, I cannot comment on the correctness of the PECOOF record layouts or what CodeView expects. However, as far as use of the generic param info is concerned this looks ok.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
stooke: Done.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -152,14 +191,96 @@ private void buildFunction(CompiledMethodEntry compiledEntry) {
         addLineNumberRecords(compiledEntry);
     }
 
+    void addLocals(CompiledMethodEntry primaryEntry) {
```
pejovica, line 194: @stooke I was kind of expecting this code to at least make some use of the debug info location support added in #4505, but that doesn't seem to be the case. Is the functionality exposed via `DebugLocalInfo` (and friends) insufficient for this use case?

stooke: @pejovica this PR is a prerequisite to adding that support.  This PR mostly implements the CodeView records required, and provides a quick but working proof of concept to examine register-based function parameters.  A follow up PR will extend the functionality to use DebugLocalInfo, which was not available when this code was written.

stooke: @pejovica, I have modified this PR to use the actual parameter names instead of synthetic names (no more 'p1, 'p2', etc), which means this PR as is is fairly useful to Windows developers, which can now set a breakpoint at function start and examine parameters in the Visual Studio IDE.  I have also merged the latest Graal head. Please have a look when you get the chance.

I continue to work on the follow-up PR, which will take full advantage of @adinn 's PR #4505.

adinn:
@pejovica Is there any reason why this PR has still not been integrated?

Simon (@stooke) has code already lined up that will use location info. However, that code is not included here because he was explicitly asked to implement the Windows behaviour in stages.

Can we please make progress on this PR which has been sitting waiting for a final review for months?

adinn:
> In particular, I don't think we should hardcode the same calling convention for all methods as is done in this PR because that is clearly incorrect.

Clearly? ... Err .. I mean ... Clearly, yes! of course!

> Instead, I think the logic in CVSymbolSubsectionBuilder.addLocals really needs to be based on the actual calling convention that was used when compiling the method.

I was not aware that that different compiled methods had different calling conventions (for a given architecture, that is).

> I actually expected that the API introduced in https://github.com/oracle/graal/pull/4505 already took care of this, but that doesn't seem to be the case. As far as I can tell, when creating a synthetic location record, ParamLocationProducer also uses the same calling convention for all methods.

Yes, for Linux `ParamLocationProducer` currently only takes CPU (AMD64/AArch64) into account when generating synthetic locations for parameters and this was certainly not questioned when #4505 was reviewed. The current implementation assumes that all compiled methods retrieved from the native image code heap will pass arguments using a single calling convention specific to the CPU type. Is that incorrect on Linux? Or is it just an issue on Windows (or MacOS)?

If there is a need for different calling conventions according to the type of the compiled method then can you clarify 1) what OS/CPU combinations it applies for; also 2) what property of the method leads to it being compiled with a given calling convention?

Is this a requirement for native image code or is it, say, only for code generated by Truffle? or some other type of generated code?

> So I think the location info API should be fixed to use the actual calling convention (see SubstrateCallingConvention), and this PR should be reworked to use the location info API. That should simplify the logic in CVSymbolSubsectionBuilder.addLocals quite a bit, but more importantly it should ensure that it is correct.

So, it seems you are suggesting that `ParamLocationProducer` needs to generalized to operate with reference to a `SubstrateCallingConvention` for all cases, even Linux, and that a PR to that effect is needed as a prerequisite change before the current PR can proceed. Is that correct? If so then I'll really need answers to the above questions.

stooke:
Looking at the code in my repo (which needs a rebase so is slightly our of date), in SubstrateCallingConventionKind.java, there are only three calling conventions - Java (needed), Native (generated by the underlying C/C++ compilers and ForwardReturnValue, which is _never_ used, only checked for.  I think it's perfectly valid, if inelegant, for the debuginfo generator to simply check that the calling convention is Java and complain otherwise.

pejovica:
@adinn @stooke There seems to be some misunderstanding about how calling from native works, but literally anything that can be called from native must be compiled with the `Native` calling convention for the call to succeed. This includes all methods annotated with the `@CEntryPoint` annotation, but as you've already noticed we also have a bunch of generated wrappers for Java methods that can be called via JNI (that's what `JNICallTrampolineMethod`s are for). So for example, if I were to set a breakpoint on the entry to `main`, the parameters `argc` and `argv` would be located according to the `Native` calling convention, meaning that PR as is would display the wrong thing. This sounds like a problem to me.

adinn:
@pejovica I take your point and agree that `ParamLocationProducer` needs to be updated to respect use of the `Native` calling convention by the relevant subset of the generated method base. That is required to ensure *both* Windows *and* Linux record the correct local var locations for the those methods in the associated debug info. I will create an issue and raise a PR to fix that.

However, I don't understand why you appear to be treating that as a blocker for this patch. I don't think any of the code here will need to be changed after the above issue is fixed i.e. pushing it now does not imply rework later. I don't believe pushing the patch will risk any serious consequences, such as crashing the runtime or Windows debugger (otherwise I would expect to see the same outcome on Linux). It will mean that Windows debugger users may be misled about input arguments for a small number of methods (as currently happens on Linux). However, in the case of almost all methods the patch will provide valuable information to Windows debugger users (just as it does on Linux) and would be useful to have right now. Not only that but pushing it will also allow Simon to present the follow up patches which shodl significantly improve windows debugging. Is there a reason not to proceed with this PR that I am missing?

pejovica:
> @pejovica I take your point and agree that `ParamLocationProducer` needs to be updated to respect use of the `Native` calling convention by the relevant subset of the generated method base. That is required to ensure _both_ Windows _and_ Linux record the correct local var locations for the those methods in the associated debug info. I will create an issue and raise a PR to fix that.

@adinn Thanks, that would be great.

> However, I don't understand why you appear to be treating that as a blocker for this patch. I don't think any of the code here will need to be changed after the above issue is fixed i.e. pushing it now does not imply rework later...

No, I don't think it is a blocker, but I also don't think we need another copy of code dealing with calling conventions. Note that `CVSymbolSubsectionBuilder.addLocal` effectively duplicates the `ParamLocationProducer` code, and since it doesn't use `DebugLocalValueInfo` at all, I don't see how we can avoid reworking this code later. That's why I suggested that this PR should be reworked to use the appropriate API.

stooke:
@pejovica , as discussed before, this code lays the groundwork for the full implementation, which will use ParamLocationProducer.  ParamLocationProducer did not exist with this PR ware written.

stooke:
@pejovica, I have rebased the repo on Graal head, and modified my code to take registers and stack offsets from @adinn's ParamLocationProducer code, thus eliminating duplication.  When his PR #7256 is integrated, this code will reap the benefits.

stooke:
@pejovica, hello! I have rebased the repo on Graal head including the integration of @adinn 's [PR #7256](https://github.com/oracle/graal/pull/7256).  When you have some time, could you please take another look at this PR?

pejovica:
@maxandersen No, I've been busy with support for Windows x64 unwind info, so I haven't had much chance to work on this. I plan to come back to this once I'm done with that feature, which is almost finished, so it should be soon.

stooke:
@pejovica , I have rebased the repo and tested it with the latest Graal native.  I have also update the copyright dates on the changed files.

pejovica:
@stooke Thanks for the rebase and the heads-up! I'll take another look now.

wirthi:
Sorry for the delay on this PR. We'll see to prioritize this in the next month so it can be merged. It's quite a substantial change with lots of implicit (Windows) knowledge that we want to understand before merging. 

Having some tests for that would be awesome, but probably hard to pull off and quite brittle?

stooke:
> Sorry for the delay on this PR. We'll see to prioritize this in the next month so it can be merged. It's quite a substantial change with lots of implicit (Windows) knowledge that we want to understand before merging.
> 
> Having some tests for that would be awesome, but probably hard to pull off and quite brittle?

It's great to hear this.  I have worked on making some reliable, repeatable tests, but that work would probably need to be a PR in itself, due to the changing nature of debug output.

wirthi:
@stooke thanks for your updates. The responsible engineer on our side is currently on leave, but I expect we can review and merge this mid August.

stooke:
> @stooke thanks for your updates. The responsible engineer on our side is currently on leave, but I expect we can review and merge this mid August.

I look forward to it!  I notice that there have been some changes in the mainline, should I update the PR - I'm hesitant to reset the clock...

stooke:
@wirthi @pejovica there have been changed in the mainline - should I update this branch?  I really don't want to reset the clock again on this.

wirthi:
@stooke there is no downside in updating the branch, if anything it would make it easier for us to adopt it.

pejovica:
@stooke Hey, I’ve finally picked this back up—thanks for your patience. I did some initial testing with the debugger against the current tip, and things look good. I’m digging back into the details and will follow up with a deeper pass soon.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVRegisterUtil.java
```diff
@@ -0,0 +1,277 @@
...
+    private static final short CV_AMD64_XMM1_0 = 166;
+    private static final short CV_AMD64_XMM2_0 = 170;
+    private static final short CV_AMD64_XMM3_0 = 174;
+    private static final short CV_AMD64_XMM4_0 = 178;
+    private static final short CV_AMD64_XMM5_0 = 182;
+    private static final short CV_AMD64_XMM6_0 = 186;
+    private static final short CV_AMD64_XMM7_0 = 190;
+
+    private static final short CV_AMD64_XMM8_0 = 260;
+    private static final short CV_AMD64_XMM9_0 = 261;
```
pejovica: `CV_AMD64_XMM9_0` and later values should stride by +4.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVRegisterUtil.java
```diff
@@ -0,0 +1,277 @@
...
+            this.register = r;
+            this.cv1 = -1;
+            this.cv2 = -1;
+            this.cv4 = cv4;
+            this.cv8 = cv8;
+        }
+    }
+
+    /* List of Graal assignable registers and CodeView register IDs. */
+    private static final CvRegDef[] compactRegDefs = {
```
pejovica: `compactRegDefs` can be made local to the static block; it’s only used to initialize `javaToCvRegisters` there.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVRegisterUtil.java
```diff
@@ -0,0 +1,277 @@
...
+    static {
+        for (CvRegDef def : compactRegDefs) {
+            assert 0 <= def.register.number && def.register.number <= MAX_JAVA_REGISTER_NUMBER;
+            javaToCvRegisters[def.register.number] = def;
+        }
+    }
+
+    /* Convert a Java register number to a CodeView register code */
+    /* the CodeView code depends upon the register type and size */
+    static short getCVRegister(int javaReg, TypeEntry typeEntry) {
```
pejovica: The contract of this method seems unclear. It should either (a) be defensive—use runtime checks and return -1 on any error; or (b) assume validated inputs—keep the asserts and drop runtime checks. The current mix is inconsistent; moreover, some conditions directly conflict (assert `0 <= javaReg <= MAX`, then check `javaReg > MAX`; assert `cvReg != null`, then check `cvReg == null`).

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVRegisterUtil.java
stooke: Done.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVRegisterUtil.java
stooke: Done - and thanks.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVRegisterUtil.java
stooke: Understood.  I have clarified the intent via reordering, comments, and a warning message.
The intent is that production code would continue with a warning (and no symbol emitted) while debug code would trigger an assert so the code could be fixed and not ignored.

stooke:
@pejovica thank you for your review!  I have tried to address your points with the latest push.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVTypeSectionBuilder.java
```diff
@@ -489,6 +500,14 @@ CVTypeRecord.CVTypeMFunctionRecord buildMemberFunction(ClassEntry classEntry, Me
         return addTypeRecord(mFunctionRecord);
     }
 
+    CVTypeRecord buildFuncIdRecord(CVTypeRecord.CVTypeMFunctionRecord mFunctionRecord, String functionName) {
+        if (mFunctionRecord.getClassType() != 0) {
```
pejovica: As far as I can tell, this can’t be 0, so the else branch creating `LF_FUNC_ID` looks dead. But even if it were, I don’t think `LF_FUNC_ID` should reference an `LF_MFUNCTION`. My understanding is that for methods we should use `LF_MFUNC_ID` only.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVTypeRecord.java
```diff
@@ -313,6 +317,94 @@ public boolean equals(Object obj) {
         }
     }
 
+    static final class CVTypeFuncIdRecord extends CVTypeRecord {
```
pejovica: I don’t think we’ll need this (see https://github.com/oracle/graal/pull/5638#discussion_r2382022923), so I’d suggest removing it.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVTypeRecord.java
```diff
@@ -357,6 +449,113 @@ public boolean equals(Object obj) {
...
+        public boolean equals(Object obj) {
+            if (!super.equals(obj)) {
+                return false;
+            }
+            CVTypeModifierRecord other = (CVTypeModifierRecord) obj;
+            return this.typeIndex == other.typeIndex && this.attrs == other.attrs;
+        }
+    }
+
+    static final class CVTypeProcedureRecord extends CVTypeRecord {
```
pejovica: This is already unused. Given the above comments (https://github.com/oracle/graal/pull/5638#discussion_r2382022923, https://github.com/oracle/graal/pull/5638#discussion_r2382135062), I’d suggest removing it as well.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVTypeSectionBuilder.java
stooke: Done, and renamed function to buildMFuncIdRecord()

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVTypeRecord.java
stooke: Gone.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVTypeRecord.java
stooke: Gone.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
```diff
@@ -301,6 +305,211 @@ public String toString() {
...
+            super(debugInfo, CVDebugConstants.S_LOCAL);
+            this.name = name;
+            this.typeIndex = typeIndex;
+            this.flags = flags;
+        }
+
+        @Override
+        protected int computeContents(byte[] buffer, int initialPos) {
+            int pos = CVUtil.putInt(typeIndex, buffer, initialPos);
+            pos = CVUtil.putShort((short) flags, buffer, pos);
```
pejovica: Seems cleaner to declare `flags` as `short`.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
```diff
@@ -301,6 +305,211 @@ public String toString() {
...
+            return String.format("S_DEFRANGE_FRAMEPOINTER_REL_FULL_SCOPE frameOffset=0x%x", frameOffset);
+        }
+    }
+
+    public static class CVSymbolDefRangeRegisterRecord extends CVSymbolDefRangeBase {
+
+        private final short register;
+        private final short attr;
+
+        CVSymbolDefRangeRegisterRecord(CVDebugInfo debugInfo, String procName, int procOffset, int length, short register) {
```
pejovica: I’d declare `length` as `short` here to match the base class.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
```diff
@@ -331,33 +540,109 @@ public String toString() {
         }
     }
 
+    @SuppressWarnings("unused")
```
pejovica: Seems not needed; class is used.
```suggestion
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
```diff
@@ -331,33 +540,109 @@ public String toString() {
...
+        protected final int debugEnd;
+        protected final int typeIndex;
+        protected final short segment;
+        protected final byte flags;
+        protected final String symbolName;
+        protected final String displayName;
 
-        CVSymbolGProc32Record(CVDebugInfo cvDebugInfo, String symbolName, String displayName, int pparent, int pend, int pnext, int proclen, int debugStart, int debugEnd, int typeIndex,
+        @SuppressWarnings("unused")
+        CVSymbolGProc32Record(CVDebugInfo cvDebugInfo, String symbolName, String displayName, int pparent, int pend, int pnext, long proclen, int debugStart, int debugEnd, int typeIndex,
```
pejovica: Any reason to keep `S_GPROC32` since it’s no longer used? If not, I’d suggest renaming `CVSymbolGProc32Record` to `CVSymbolGProc32IdRecord` and switching the record type to `S_GPROC32_ID`.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -26,35 +26,55 @@
...
+import java.util.Map;
+
+import static com.oracle.objectfile.pecoff.cv.CVRegisterUtil.CV_AMD64_R8;
+import static com.oracle.objectfile.pecoff.cv.CVSymbolSubrecord.CVSymbolFrameProcRecord.FRAME_LOCAL_BP;
+import static com.oracle.objectfile.pecoff.cv.CVSymbolSubrecord.CVSymbolFrameProcRecord.FRAME_PARAM_BP;
+import static com.oracle.objectfile.pecoff.cv.CVTypeConstants.MAX_PRIMITIVE;
 
 final class CVSymbolSubsectionBuilder {
 
+    private static final int S_LOCAL_FLAGS_IS_PARAM = 1;
```
pejovica: Might be better placed in `CVSymbolLocalRecord`.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -26,35 +26,55 @@
...
         this.cvDebugInfo = cvDebugInfo;
         this.cvSymbolSubsection = new CVSymbolSubsection(cvDebugInfo);
         this.lineRecordBuilder = new CVLineRecordBuilder(cvDebugInfo);
         this.heapName = SectionName.SVM_HEAP.getFormatDependentName(cvDebugInfo.getCVSymbolSection().getOwner().getFormat());
-        /* For isolates, Graal currently uses r14; this code will handle r8-r15. */
-        assert 8 <= cvDebugInfo.getHeapbaseRegister() && cvDebugInfo.getHeapbaseRegister() <= 15;
-        this.heapRegister = (short) (CV_AMD64_R8 + cvDebugInfo.getHeapbaseRegister() - 8);
+        /* For isolates, Graal currently uses r14 as the heap base; this code will handle r8-r15. */
+        assert AMD64.r8.number <= cvDebugInfo.getHeapbaseRegister() && cvDebugInfo.getHeapbaseRegister() <= AMD64.r15.number;
+        this.heapRegister = (short) (CV_AMD64_R8 + cvDebugInfo.getHeapbaseRegister() - AMD64.r8.number);
```
pejovica: Feels like using `CVRegisterUtil.getCVRegister` would be cleaner here. Maybe add an overload that takes size in bytes?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -126,40 +146,118 @@ private void buildFunction(CompiledMethodEntry compiledEntry) {
...
 
-        /* S_PROC32 add function definition. */
-        int functionTypeIndex = addTypeRecords(compiledEntry);
-        byte funcFlags = 0;
-        CVSymbolSubrecord.CVSymbolGProc32Record proc32 = new CVSymbolSubrecord.CVSymbolGProc32Record(cvDebugInfo, externalName, debuggerName, 0, 0, 0,
-                        primaryRange.getHiOffset() - primaryRange.getLoOffset(), 0, 0, functionTypeIndex, (short) 0, funcFlags);
+        /* add function definition. */
+        final int functionTypeIndex = addTypeRecords(compiledEntry);
+        final byte funcFlags = 0;
+        CVSymbolSubrecord.CVSymbolGProc32IdRecord proc32 = new CVSymbolSubrecord.CVSymbolGProc32IdRecord(cvDebugInfo, externalName, debuggerName, 0, 0, 0, primaryRange.getHi() - primaryRange.getLo(),
```
pejovica: Why switch to `getHi() - getLo()`? Doesn’t seem functionally different and unnecessarily widens `proclen` to `long`.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -126,40 +146,118 @@ private void buildFunction(CompiledMethodEntry compiledEntry) {
...
-        int paramBP = 1 << 16; /* Param base pointer = SP. */
-        int frameFlags = asynceh + localBP + paramBP; /* NB: LLVM uses 0x14000. */
+        final int frameFlags = FRAME_LOCAL_BP + FRAME_PARAM_BP;
         addSymbolRecord(new CVSymbolSubrecord.CVSymbolFrameProcRecord(cvDebugInfo, compiledEntry.frameSize(), frameFlags));
 
-        /* TODO: add parameter definitions (types have been added already). */
-        /* TODO: add local variables, and their types. */
-        /* TODO: add block definitions. */
+        /* it's costly to compute this, so only compute it once */
+        Map<LocalEntry, List<Range>> varRangeMap = primaryRange.getVarRangeMap();
```
pejovica: Looks like `getVarRangeMap` already caches its result, so is this local cache still needed?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -126,40 +146,118 @@ private void buildFunction(CompiledMethodEntry compiledEntry) {
...
 
         /* Add line number records. */
         addLineNumberRecords(compiledEntry);
     }
 
+    void addParameters(CompiledMethodEntry primaryEntry, Map<LocalEntry, List<Range>> varRangeMap) {
+        final Range primaryRange = primaryEntry.primary();
+        /* The name as exposed to the linker. */
+        final String externalName = primaryRange.getSymbolName();
+        final MethodEntry method = primaryRange.isPrimary() ? primaryRange.getMethodEntry() : primaryRange.getCallees().getFirst().getMethodEntry();
```
pejovica: This can be simplified now that `primaryEntry.primary()` returns `PrimaryRange`.
```suggestion
        final MethodEntry method = primaryRange.getMethodEntry();
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -126,40 +146,118 @@ private void buildFunction(CompiledMethodEntry compiledEntry) {
...
+
+        /* define function parameters */
+        assert Modifier.isStatic(method.getModifiers()) || !method.getParams().isEmpty();
+        for (LocalEntry paramInfo : method.getParams()) {
+            final TypeEntry paramType = paramInfo.type();
+            final int typeIndex = method.isThisParam(paramInfo)
+                            ? cvDebugInfo.getCVTypeSection().getIndexForPointer(paramType)
+                            : cvDebugInfo.getCVTypeSection().addTypeRecords(paramType).getSequenceNumber();
```
pejovica: I think we should use `getIndexForPointer` for all params because `addTypeRecords` would return a non-pointer type index for object params.
```suggestion
            final int typeIndex = cvDebugInfo.getCVTypeSection().getIndexForPointer(paramType);
```

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -126,40 +146,118 @@ private void buildFunction(CompiledMethodEntry compiledEntry) {
...
+        int flags = isParam ? S_LOCAL_FLAGS_IS_PARAM : 0;
+        List<Range> ranges = varRangeMap.get(info);
+        addSymbolRecord(new CVSymbolSubrecord.CVSymbolLocalRecord(cvDebugInfo, name, typeIndex, flags));
+        long currentHigh = Integer.MIN_VALUE;
+        int registerOrSlot = 0; /* -slot or +register or 0=unknown */
+        CVSymbolSubrecord.CVSymbolDefRangeBase currentRecord = null;
+        for (Range subrange : ranges) {
+            LocalValueEntry value = subrange.lookupValue(info);
+            if (value != null) {
+                if (subrange.getLo() == currentHigh && registerOrSlot == infoTypeToInt(value)) {
```
pejovica, line 201: Have you considered merging discontinuous ranges via gaps? `cvdump` suggests there are plenty of opportunities.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
```diff
@@ -126,40 +146,118 @@ private void buildFunction(CompiledMethodEntry compiledEntry) {
...
+                    /* if we can, merge records */
+                    currentHigh = subrange.getHi();
+                    if (currentRecord != null) {
+                        currentRecord.length = (short) (currentHigh - currentRecord.procOffset - range.getLo());
```
pejovica: I don’t think we should assume the combined length fits in a `short`. Could we check before merging and skip otherwise?

wirthi:
Hey @stooke 

is everything clear regarding the review comments, is there anything else we should provide? Can you fix/comment the commented parts?

I hope we can merge this soonish, this has been open embarrassingly long.

Christian

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
stooke: done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
stooke: done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
stooke: done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
stooke: Done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
stooke: done.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
stooke: done.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubrecord.java
stooke: done.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
stooke: Because this PR has taken so long to review getHi() and getLo()  changed their return type to long 4 months ago.  getHiOffset() didn't exist until then. Changed to the new method.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
stooke: This can be a future optimization, for a future PR.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
stooke: A test has been added; if it won't fit, we emit a record up to the 16-bit boundary (length = 0xffff).

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/pecoff/cv/CVSymbolSubsectionBuilder.java
stooke: done.

stooke:
@wirthi @pejovica I have addressed the latest round of comments, and rebased to head.  Please have a good winter break!

pejovica:
@stooke Thanks for the update. Enjoy your break too!

maxandersen:
@pejovica hi - any updates on this?

pejovica:
@maxandersen Hey! I basically started back this week after the break, so I haven’t had a chance to get to this yet. I should be able to pick it up next week.

maxandersen:
Thanks for the update @pejovica ! We are looking forward to it!

maxandersen:
Friendly ping here @pejovica

pejovica:
@maxandersen I'm on it now. Will update later today.

substratevm/src/native-image-module-tests/hello.app/target/classes/META-INF/native-image/resource-config.json
pejovica, line 1: This looks like Maven `target/` build output. Can we remove it?

substratevm/src/native-image-module-tests/hello.app/target/classes/resource-file.txt
pejovica, line 1: Also build output. Please remove.

substratevm/src/native-image-module-tests/hello.lib/target/classes/resource-file.txt
pejovica, line 1: Also build output. Please remove.

pejovica (commented):
@stooke Looks like some build output got included in this PR. Could you remove these files?

substratevm/src/native-image-module-tests/hello.app/target/classes/META-INF/native-image/resource-config.json
stooke: Apparently my setup has a background process that was "helpfully" running the pom files and recreating these targets; they are deleted now and I'll be more careful about the files I check in.  Thanks.

substratevm/src/native-image-module-tests/hello.lib/target/classes/resource-file.txt
stooke: done

substratevm/src/native-image-module-tests/hello.app/target/classes/resource-file.txt
stooke: done

stooke:
> @stooke Looks like some build output got included in this PR. Could you remove these files?

Done, a background process created these and I was unfortunately careless in my checkin.  Thanks fo r the catch.

stooke:
@pejovica I have squashed the PR (this might also be an option on the integration side if enabled for the Graal repository).  Unfortunately, the changes (prseumably the force push) have triggered a 'workflow awaiting approval' message that requires approval from a maintainer; do you have maintainer status?  Until the workflow is approved, the GraalVM gate won't run.

pejovica:
> do you have maintainer status?

@stooke Yes. I approved the workflow, so the gate should be running now.

Issue #5636, opened 2022-12-16
Add support for `jdk.ObjectAllocationOutsideTLAB` and `jdk.ObjectAllocationInNewTLAB` JFR events


roberttoyonaga:
**TL;DR**
Add support for the`jdk.ObjectAllocationOutsideTLAB` and `jdk.ObjectAllocationInNewTLAB` JFR events.

**Goals**
These events should detail the object class and allocation size when an object is allocated outside or in a new TLAB.

When this work is done, the[ JFR Support tracking Issue](https://github.com/oracle/graal/issues/5410) should be updated to reflect the changes.

**Describe who do you think will benefit the most.**
GraalVM users will benefit from the additional information JFR will report. GraalVM developers will also benefit because object allocations outside TLAB events can help debug performance issues. 

**Express whether you'd like to help by contributing to this feature**
I volunteer to work on this.

fniephaus:
#5654 is merged, does this mean this issue is done and we need to update #5410?

PR #5654, opened 2022-12-20
Object allocation in new TLAB JFR event

roberttoyonaga:
### Description of Changes
This adds support for the JFR event `jdk.ObjectAllocationInNewTLAB`.  This PR resolves the issue https://github.com/oracle/graal/issues/5636

The JFR event`jdk.ObjectAllocationOutsideTLAB` also exists, but does not seem applicable to native images because allocations are always done in new TLABs.  For example, in Hotspot, if an allocation is larger than the largest possible TLAB size, or if TLAB allocation fails, then allocation is done outside of the TLAB. This is also true if the `UseTLAB` option is false. However, in SVM, all the aforementioned scenarios still result in using a new TLAB.


fniephaus:
Is there any way we can test each JFR event and make sure we don't regress on anything?

roberttoyonaga:
> Is there any way we can test each JFR event and make sure we don't regress on anything?

I've added some tests in [this internal PR](https://github.com/oracle/graal/pull/5528/files#diff-c5f3ec6b0d3596f19c9e9c5ec809a42ad63faae2a562a5b5c3ab4ac68c799130). I have some additional tests for the other events I've worked on, including this one, but I've been waiting for that internal PR to be merged before adding them to a new PR to avoid conflicts.  Those tests only check for correctness though, they don't analyze performance.

christianhaeubl:
Thanks for implementing this event. The changes look good, I will do a few minor cleanups and merge your PR into master.

PR #5691, opened 2023-01-03
[GR-38651] Add `-H:StripDebugInfo` and other tweaks for Debug Info.

graalvmbot:


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoStripFeature.java
```diff
@@ -0,0 +1,120 @@
...
+        try {
+            objcopyAvailable = FileUtils.executeCommand(objcopyExe, "--version") == 0;
+        } catch (IOException e) {
+            /* Fall through to `if (!objcopyAvailable)` */
+        } catch (InterruptedException e) {
+            throw new InterruptImageBuilding("Interrupted during checking for " + objcopyExe + " availability");
+        }
+
+        if (!objcopyAvailable) {
+            System.out.printf("Warning: %s not available. Skip generation of debuginfo file %s%n", objcopyExe, debugInfoName);
```
zakkak, line 97: I find this message a bit ambiguous. It's not clear whether debuginfo won't be generated at all or not.

I suggest:
```suggestion
            System.out.printf("Warning: %s not available. Skipping generation of separate debuginfo file %s, debuginfo will remain embedded in the executable%n", objcopyExe, debugInfoName);
```

zakkak (changes requested):

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -510,6 +521,16 @@ private void mangleType(ResolvedJavaType type) {
             }
         }
 
+        private void mangleType(UnresolvedJavaType type) {
+            if (type.isArray()) {
```
adinn, line 525: We don't have a class loader for this unresolved type so I am not sure that we can guarantee that its name will not clash with a name for some valid resolved type. It might be safer to put all unresolved types into their own namespace to ensure this does not happen. That would also involve modifying the debug type info generation code to adopt the same namespace convention.

I'll leave that change for a follow-up PR once we can identify what other changes are required to ensure that debug info generator does not break if an unresolved type pops up as an element of the declaration closure for the runtime type set.

adinn (approved):
Looks ok to push. I think we will need to do more work to ensure that unresolved types don't cause any grief during debug info generation.

olpaw (approved):

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
```diff
@@ -227,8 +227,8 @@ public void installDebugInfo(DebugInfoProvider debugInfoProvider) {
          */
         int oopTagsMask = debugInfoProvider.oopTagsMask();
 
-        /* Tag bits must be between 1 and 32 for us to emit as DW_OP_lit<n>. */
-        assert oopTagsMask > 0 && oopTagsMask < 32;
+        /* Tag bits must be between 0 and 32 for us to emit as DW_OP_lit<n>. */
```
adinn, line 230: Doh!

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -1760,10 +1760,13 @@ public int writeIndirectOopConversionExpression(boolean isHub, byte[] buffer, in
...
-                    /* We need two shifts to remove the bits. */
-                    rightShift = oopTagsShift;
+                    /* We need one or two shifts to remove the bits. */
+                    if (oopTagsShift != 0) { // -H:+UseCompressedReferences
+                        rightShift = oopTagsShift;
+                        exprSize += 2;
+                    }
                     leftShift = oopCompressShift;
-                    exprSize += 4;
+                    exprSize += 2;
```
adinn, line 1769: Double doh! At least the spec is right there in the code :-)

substratevm/mx.substratevm/testhello.py
```diff
@@ -219,7 +219,7 @@ def test():
                        r"#1%s%s in com\.oracle\.svm\.core\.JavaMainWrapper::runCore0%s %s at %sJavaMainWrapper\.java:[0-9]+"%(spaces_pattern, address_pattern, no_param_types_pattern, no_arg_values_pattern, package_pattern),
                        r"#2%s%s in com\.oracle\.svm\.core\.JavaMainWrapper::runCore%s %s at %sJavaMainWrapper\.java:[0-9]+"%(spaces_pattern, address_pattern, no_param_types_pattern, no_arg_values_pattern, package_pattern),
                        r"#3%scom\.oracle\.svm\.core\.JavaMainWrapper::doRun%s %s at %sJavaMainWrapper\.java:[0-9]+"%(spaces_pattern, param_types_pattern, arg_values_pattern, package_pattern),
-                       r"#4%s%s in com\.oracle\.svm\.core\.JavaMainWrapper::run%s %s at %sJavaMainWrapper\.java:[0-9]+"%(spaces_pattern, address_pattern, param_types_pattern, arg_values_pattern, package_pattern),
+                       r"#4%s(%s in )?com\.oracle\.svm\.core\.JavaMainWrapper::run%s %s at %sJavaMainWrapper\.java:[0-9]+"%(spaces_pattern, address_pattern, param_types_pattern, arg_values_pattern, package_pattern),
```
adinn, line 222: I'm not clear why this and the following change were needed. Is this just because of a different inlining decision?

fniephaus: Yes, `JavaMainWrapper::run` is inlined by the EE compiler.

olpaw:
@christianhaeubl to switch our debuginfo implementation to the new improved one by RedHat @fniephaus found an issue when it is used in combination with G1 GC. Commit https://github.com/oracle/graal/pull/5691/commits/058563c782934cd3a4a36669ba5c6b9705892064 contains our attempt to fix the problem. Please let us know if this looks legit to you. The code is question is creating dwarf expressions (small dwarf sub-programms that get interpreted by gdb to resolve oop references) based on the object representation for the  image that gets built.

christianhaeubl:
@olpaw thanks, looks good to me.

substratevm/mx.substratevm/testhello.py
adinn: Ok, thought it was just that. All looks good.

adinn (approved):

substratevm/mx.substratevm/testhello.py
```diff
@@ -784,13 +783,14 @@ def test():
...
-    checker = Checker('x/i $pc', rexp)
-    checker.check(exec_string)
+    if asm_insns:
+        exec_string = execute("x/i $pc")
+        if arch == 'aarch64':
+            rexp = r"%ssub%ssp, sp, #0x%s"%(wildcard_pattern, spaces_pattern, hex_digits_pattern)
+        else:
+            rexp = r"%ssub %s\$0x%s,%%rsp"%(wildcard_pattern, spaces_pattern, hex_digits_pattern)
+        checker = Checker('x/i $pc', rexp)
+        checker.check(exec_string)
```
adinn, line 793: @fniephaus What was the reason for making this optional?

fniephaus: This part of the test is a bit overfitting as it specifies what instructions the compiler is allowed to emit for method prologues. For example, GraalVM EE uses `-:H+StackOverflowCheckInPrologue` by default and therefore emits different instructions. Instead of making this optional, we could also remove the check altogether. After all, I believe the tests should simulate representative debugging scenarios to ensure we provide a decent user experience, without relying on specific decisions made by the compiler.

adinn: @fniephaus I understand that we want to avoid the test depending on what the compiler does but there was a good reason why this rule was broken here. Indeed, the fact that this check caught a difference in EE means it has served its purpose ;-)

The instruction check is actually just part of a sequence of checks designed to ensure that parameter variable encodings are correct. That sequence includes a prior and subsequent validation of the output from the `info args` command. The method being debugged has been deliberately chosen so that it will have a mix of int greg and freg arguments *plus* also integer and floating arguments passed via the stack.

The break point is carefully placed at the first instruction of the method, an address guaranteed to provide var info for all params. That is where we first run `info args`. The point of the stepi instruction on x86/two stepi instructions on aarch64 is to transition to the next infopoint at the end of the method prologue, also guaranteed to provide an updated set of var info for all params. The point of testing before and after the stack push is to ensue that the stack offsets for the stack-passed parameters have been encoded correctly both before and after push

So, this means that on EE we won't have completed that stack push transition and so the second validation of the `info args` command is not going to be checking the offsets post stack-push. That does not mean that the test will fail. What it does mean is that the test has been weakened and may catch an error in the calculation of the stack offsets post-push. It would be better to come up with something that allows the pre- and post-push offset calculations to be validated on EE as well as CE.

What we need is to be able to guarantee we stop at the point where i) the stack push is complete and ii) param info is available. I can be sure there is param info available at the stack push instruction because I ensure that the generator explicitly writes info at that point. The alternative would be to stop using a break point, say at a specific line. The problem with that is that param var location info cannot be guaranteed to be available at any chosen line break -- it comes and goes according to the whim of the compiler's infopoint encoding scheme, which is far more complex than any decisions it makes regarding the stack push.

What we could perhaps do instead of stepping a fixed number of instructions is to step until we actually see a stack extend instruction, execute that instruction and only then test the params values. This would build in a slightly weaker assumption i.e. that last instruction emitted by the compiler in a method prologue is a stack adjustment. That's not too presumptious (after all we already assume a fixed size frame). Making this work means we would need to detect any instruction on the current CPU that adjusts the stack register. Does that sound like a good idea? @olpaw any thoughts on this?

fniephaus: Thanks for the shedding some light on this, @adinn. Your suggestion seems very reasonable to me, so I have removed the option for making the check optional and replaced the logic that skips over the method prologue with a new one that steps instructions until the first that adjusts the stack pointer register (see https://github.com/oracle/graal/commit/66d1486241a1aa84aab79624a4937f79f42e4cb4). I hope I've documented this well enough in the test code.

adinn: That looks good but I think it still needs a small adjustment. If you look at the previous test code you will see that the aarch64 branch executes two `stepi` commands after he check for a stack push. That's because on aarch64 the stack build sequence requires both a stack extend and then a save of sp to fp. The compiler only posts a PROLOGUE_END notification after the second instruction. This also means that the param var locations in the debug info are only redefined after the move instruction. So, if you don't do the 2nd stepi then the pre-push var locations will be used. So, I would expect this version of the test to fail on aarch64.

fniephaus: Doh! I think this should fixes it: https://github.com/oracle/graal/pull/5691/commits/fc40db371521d883068d4bf819f9765a88c1dbfe#diff-911ff5ce4770f7bdd9410ad3374ae2d3f4fc7fbc69b661cac4b2d7077062ee1cR805-R813.

adinn: Yes, looks good.

PR #5694, opened 2023-01-04
[GR-42662] Espresso method overloading interop and varargs support.

graalvmbot:


JaroslavTulach:
Hello Allan,
this seems like a great step forward! I am able to execute following Enso program:
```haskell
import Standard.Base.IO

main = IO.println <| "Hello World!"
```
that program loads a lot of classes from Enso standard library and then uses `IO.println` - everything seems to be working. We are able to [move forward on the Enso side](https://www.pivotaltracker.com/n/projects/2539304/stories/183260380) now. Thanks a lot!

Issue #5410, opened 2022-11-10
[WIP] Outline of JFR Support In Native Image vs JIT 


roberttoyonaga:
### Purpose
This issue outlines the the current and in progress scope of JFR support in native images compared to JIT. This outline is a work in progress and should be kept up-to-date as tasks are identified/completed.

### Events
The following built-in JFR events are supported in native images.
- jdk.ThreadStart
- jdk.ThreadEnd
- jdk.DataLoss
- jdk.ClassLoadingStatistics
- jdk.InitialEnvironmentVariable
- jdk.InitialSystemProperty
- jdk.JavaThreadStatistics
- jdk.JVMInformation
- jdk.OSInformation
- jdk.PhysicalMemory
- jdk.ExecutionSample
- jdk.GarbageCollection
- jdk.GCPhasePause
- jdk.SafepointBegin
- jdk.SafepointEnd
- jdk.ExecuteVMOperation
- jdk.JavaMonitorEnter
- jdk.ThreadSleep
- jdk.JavaMonitorWait
- jdk.ThreadPark

### Features
|Feature| Native Image | JIT |Description|
| ---        |    ----   |    ----  |----|
|Event Streaming| In progress | Supported  | |
|Leak profiling| In progress | Supported  | Via old object sampling |
|Remote JMX | In progress | Supported  | Manage beans over JMX connection|
|Stacktrace data for events | Supported | Supported  | |
|In-memory chunk rotation| Not Supported | Supported||

### Optimizations
|Optimization| Native Image | JIT |Description|
| ---        |    ----   |    ----  |----|
|String pool|Not implemented| Implemented|Constant pool to deduplicate strings|

PR #5701, opened 2023-01-04
Add Java monitor inflation JFR event

roberttoyonaga:
This pull request adds the `jdk.JavaMonitorInflate` JFR event.

 In Hotspot, object monitors can be inflated and deflated. Inflation is the slow path and it happens when there is contention for the monitor, among other reasons. Once there are no longer threads waiting for the monitor, the monitor is deflated and the fast path can be used again. In SVM, inflation essentially happens whenever a `JavaMonitor` is lazily created for an object. Deflation doesn't ever happen. So in SVM this event can only help describe how often `JavaMonitors` are being created and why (the "cause"), it doesn't provide insight into how often the slow path is taken (because it's always taken).  In the future, if SVM monitors become more complex like in hotspot, then this event would become more useful.

When this is done, the[ JFR Support tracking Issue](https://github.com/oracle/graal/issues/5410) should be updated to reflect the changes.


oubidar-Abderrahim:
Thank you for contributing to GraalVM. 
@fniephaus @christianhaeubl could you please review this PR? or assign someone better suited to do so? Thank you

christianhaeubl (commented):
Thanks for implementing this event. The changes look good, I will do a few cleanups and merge your PR into master.

PR #5726, opened 2023-01-09
[GR-43141] Removed an invalid code path in System.nanoTime().

graalvmbot:


christianhaeubl:
This PR was replaced with another one.

PR #5751, opened 2023-01-12
Make GraalVM compatible with JDK 20+27 plus

### LINKED ISSUE #5750: Mandrel fails to build with JDK 20+30
**Describe the issue**
Mandrel builds fail when using JDK 20 starting with JDK 20+27.

```
wget https://api.adoptium.net/v3/binary/latest/20/ea/linux/x64/jdk/hotspot/normal/eclipse -o jdk.tar.gz
wget https://api.adoptium.net/v3/binary/latest/20/ea/linux/x64/staticlibs/hotspot/normal/eclipse -o jdk-static-libs.tar.gz
tar xf jdk.tar.gz
tar xf /tmp/jdk-static-libs.tar.gz -C jdk-20+30/ --strip-components=1
export JAVA_HOME=`pwd`/jdk-20+30
git clone https://github.com/graalvm/mandrel-packaging
git clone https://github.com/graalvm/mandrel
git clone https://github.com/graalvm/mx
cd mandrel-packaging
$JAVA_HOME/bin/java -ea build.java --mx-home ../mx --mandrel-repo ../mandrel
```

 - GraalVM version: **92cfa766d1f059561e6c6a5631d6b082fc9f7765**
 - JDK major version: **20**
 - OS: **Fedora 37**
 - Architecture: **AMD64**

**More details**
GraalVM tries to substitute methods `extentLocalCache` and `setExtentLocalCache` if the JDK is > 19:

https://github.com/oracle/graal/blob/92cfa766d1f059561e6c6a5631d6b082fc9f7765/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Target_java_lang_Thread.java#L649-L659

However, both methods are removed in https://github.com/openjdk/jdk20u/commit/221e1a426070088b819ddc37b7ca77d9d8626eb4#diff-2a5b4b651fa41e70f7953f7ee06e964d8396fcd1fe5755a1609e81b3f84d9224L398-L403 resulting in Mandrel builds failing with:

```
Error: Could not find target method: static java.lang.Object[] com.oracle.svm.core.thread.Target_java_lang_Thread.extentLocalCache()
```

This is expected to impact GraalVM as well once https://github.com/graalvm/labs-openjdk-20 syncs with upstream.

zakkak:
https://bugs.openjdk.org/browse/JDK-8298177 and https://bugs.openjdk.org/browse/JDK-8286666 introduced some breaking changes.

Closes #5750


Karm:
@zakkak contemporary JDK 20 (Januray 13 e0081462f4561), build passes with this branch, although building Hello world fails at linker stage:

```
hello.o:(.data+0x2a8): undefined reference to `Java_java_lang_Thread_ensureMaterializedForStackWalk'
```

It's the added [static native void ensureMaterializedForStackWalk\(Object o\);](https://github.com/openjdk/jdk20/blob/master/src/java.base/share/classes/java/lang/Thread.java#L418)

Not sure what to do with it, mimicking what you did is not what helps...
```
+    @Inject @TargetElement(onlyWith = { HasSetScopedValueCache.class, HasScopedValueCache.class, LoomJDK.class }) //
+    @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.Reset) //
+    Object bindings;
+
...
+    @Substitute
+    @TargetElement(onlyWith = HasSetScopedValueCache.class)
+    static void ensureMaterializedForStackWalk(Object o) {
+        JavaThreads.toTarget(currentCarrierThread()).bindings = o;
+    }
+
```

/me goona look at the JDK JIRA to figure out what it does...

zakkak:
> @zakkak contemporary JDK 20 (Januray 13 e0081462f4561), build passes with this branch, although building Hello world fails at linker stage:
> 
> ```
> hello.o:(.data+0x2a8): undefined reference to `Java_java_lang_Thread_ensureMaterializedForStackWalk'
> ```
> 
> It's the added [static native void ensureMaterializedForStackWalk(Object o);](https://github.com/openjdk/jdk20/blob/master/src/java.base/share/classes/java/lang/Thread.java#L418)
> 
> Not sure what to do with it, mimicking what you did is not what helps...

Yes I noticed that too. My understanding is that we should not substitute this method, but instead find out how to allow GraalVM to invoke it. If you come to any better conclusion please share.

Karm:
@zakkak 
I tried fiddling with `JNIRegistrationJava.java`, although it still escapes...

I can see `libjava.a`being passed to the linker, yet it does not contain the expected symbol. It contains:
```
Java_java_security_AccessController_ensureMaterializedForStackWalk
Java_java_security_AccessController_ensureMaterializedForStackWalk
.text.Java_java_security_AccessController_ensureMaterializedForStackWalk
ensureMaterializedForStackWalk
```
While the missing one is  `Java_java_lang_Thread_ensureMaterializedForStackWalk`:
```
hello.o:(.data+0x230): undefined reference to `Java_java_lang_Thread_ensureMaterializedForStackWalk'
```
...and that is nowhere to be seen.

Looking at some other symbols in `libjava.a`, e.g.
```
Java_java_lang_Thread_registerNatives
Java_java_lang_Thread_clearInterruptEvent
Java_java_lang_Thread_registerNatives
Java_java_lang_Thread_clearInterruptEvent
```
I am wondering why `ensureMaterializedForStackWalk` symbol is not JNI style prefixed...

Karm:
My autogenerated `/home/karm/workspaceRH/jdk20/build/linux-x86_64-server-release/support/headers/java.base/java_lang_Thread.h`

contains the symbol called correctly though:
```
/*
 * Class:     java_lang_Thread
 * Method:    ensureMaterializedForStackWalk
 * Signature: (Ljava/lang/Object;)V
 */
JNIEXPORT void JNICALL Java_java_lang_Thread_ensureMaterializedForStackWalk
  (JNIEnv *, jclass, jobject);
```

zakkak:
If I get it right, `Java_java_lang_Thread_ensureMaterializedForStackWalk` is not actually defined, due to being an intrinsic. When hotspot spots an invocation to `Java_java_lang_Thread_ensureMaterializedForStackWalk` it just replaces it
with a `NOP`, at least in AMD64 and AArch64 [1, 2]. As per the comment in [1, 2], the NOP doesn't serve any other purpose other than preventing a change in the code size because of the removal of the initial call/branch instruction.

`JVM_EnsureMaterializedForStackWalk_func` appears to serve as a wrapper to the macro `JVM_EnsureMaterializedForStackWalk` [3, 4] so that it can be referenced and registered as the implementation of
`Java_java_lang_Thread_ensureMaterializedForStackWalk` [5] for other architectures (? as well as for the interpreter?) that don't replace invocations to `Java_java_lang_Thread_ensureMaterializedForStackWalk` with NOPS. Not surprisingly, the definition of the macro `JVM_EnsureMaterializedForStackWalk` is a NOP itself [6]. The comment in the macro `JVM_EnsureMaterializedForStackWalk` definition indicates that it is essentially used to force the value passed as an argument to it to escape to a native method. This way the JIT compiler can no longer consider it safe for optimizations that would "unmaterialize" it.

I suspect that the registration in [5] is the one that leads in the JNI export in the generated header file, despite it not being necessary (as there is no code referencing it, and there is no implementation of the method either), one could say it's actually missleading as one might thing they can invoke the method from their JNI code.

Now on the GraalVM/Mandrel land. My guestimate is that GraalVM not being aware that `Java_java_lang_Thread_ensureMaterializedForStackWalk` is an intrinsic doesn't optimize the call out. So when it tries to link the object files it complains that the method doesn't exist. So the solution is probably to make `Java_java_lang_Thread_ensureMaterializedForStackWalk` an intrinsic in GraalVM as well and replace invocations to it with NOPs (can we just delete it instead in GraalVM?) just like in HotSpot.

As noted by @adinn, when doing ^^ we need to ensure that the optimization/substitution happens late enough in order to prevent Graal from figuring out that the invocation is actually a NOP. 

[1] https://github.com/openjdk/jdk/commit/221e1a426070088b819ddc37b7ca77d9d8626eb4#diff-018aa61d1a7aafcf70a535fcd40a318a4bd6511fd40ac39ce4be90cc52216749
[2] https://github.com/openjdk/jdk/commit/221e1a426070088b819ddc37b7ca77d9d8626eb4#diff-e5266a3774f26ac663dcc67e0be18608b1735f38c0576673ce36e0cd689bab4a
[3] https://github.com/openjdk/jdk/commit/221e1a426070088b819ddc37b7ca77d9d8626eb4#diff-a911b98b446b5d5ea166ce7b8babf62a92c11c586a16da09d74076c903d3e208L742-R749
[4] https://github.com/openjdk/jdk/commit/221e1a426070088b819ddc37b7ca77d9d8626eb4#diff-50437abcc14d6d69f89f6a6566f1ed016f60a8c2173d43cd681727b46a0dd314R4063-R4071
[5] https://github.com/openjdk/jdk/commit/221e1a426070088b819ddc37b7ca77d9d8626eb4#diff-8859eadf574f551df2ed16f890de5ab77147d94be0e00dcde0c9f1cc9a1b716dR57-R58
[6] https://github.com/openjdk/jdk/commit/221e1a426070088b819ddc37b7ca77d9d8626eb4#diff-a911b98b446b5d5ea166ce7b8babf62a92c11c586a16da09d74076c903d3e208L737-L744

zakkak:
I prepared https://github.com/oracle/graal/pull/5751/commits/9668ae46b175c75a5fb0798247c6c4db35487fc4 which treats `java.lang.Thread.ensureMaterializedForStackWalk` as an intrinsic and elides it, but I still have not figured out how to force its argument to be materialized. Could someone explain how I can achieve this? Looking for some special graph Node I stumbled across `MaterializedObjectState` but I don't see how to use it.

zakkak:
Thanks to @gergo- (for pointing me to `BlackholeNode`) this is now ready for review!

Karm:
There might be something wrong with my local setup, although the very latest JDK 20:
```
Failed generating 'libnative-image-agent' after 3.1s.

The build process encountered an unexpected error:

> com.oracle.svm.core.util.VMError$HostedError: Header file include/riscv64cpufeatures.h not found at main search location(s): 
/home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/svm/clibraries/linux-amd64/include/riscv64cpufeatures.h
 Use option -H:CLibraryPath to specify header file search locations.
```

Whereas:
```
$ find /home/karm/workspaceRH/ -name riscv64cpufeatures.h
/home/karm/workspaceRH/graal/graal/substratevm/src/com.oracle.svm.native.libchelper/include/riscv64cpufeatures.h
```

TBH, I was unaware that there is an ongoing GraalVM RISC-V work...

zakkak:
> Gonna check Mandrel packaging later...

Hi @Karm, that's fixed by https://github.com/graalvm/mandrel-packaging/pull/300, please pull `master` from mandrel-packaging.

jerboaa:
@zakkak There are some style issues the CI build complains about. Want to fix those?

adinn:
@zakkak I believe that translating the call to `ensureMaterializedForStackWalk` to a `BlackholeNode` that consumes the object argument should be enough to ensure it is allocated on the heap rather than on the stack. The `BlackholeNode` should contribute no instructions to the final code (not even a NOP), which is different to OpenJDK. However, I don't believethe lack of any NOP will cause any problems. That said, I'll leave it up to a pukka compiler dev/reviewer to provide an actual review.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Target_java_lang_Thread.java
```diff
@@ -182,10 +186,14 @@ public final class Target_java_lang_Thread {
     @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.Reset) //
     Object lockHelper;
 
-    @Inject @TargetElement(onlyWith = LoomJDK.class) //
+    @Inject @TargetElement(onlyWith = {HasSetExtentLocalCache.class, HasExtentLocalCache.class, LoomJDK.class}) //
     @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.Reset) //
     Object[] extentLocalCache;
```
peter-hofer, line 191: There is no need to _inject_ differently-named fields based on the name in the JDK that we're building on. You can rename the field to `scopedValueCache` in either case and use it from `extentLocalCache` and `setExtentLocalCache` too.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
```diff
@@ -2282,4 +2283,25 @@ public SnippetTemplate.SnippetInfo getSnippet(BigIntegerSnippets.Templates templ
...
+        InvocationPlugin threadPlugin = new InvocationPlugin("ensureMaterializedForStackWalk", Object.class) {
+            @Override
+            public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver, ValueNode object) {
+                b.add(new BlackholeNode(object));
+                return true;
+            }
+        };
+        Registration r = new Registration(plugins, "java.lang.Thread", replacements);
+        r.registerConditional(hasEnsureMaterializedForStackWalk(), threadPlugin);
+    }
```
peter-hofer, line 2306: @mur47x111 @dougxc this intrinsic would also be used on HotSpot, please review.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/HasSetExtentLocalCache.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.oracle.svm.core.jdk;
+
+import java.util.function.BooleanSupplier;
+
+/**
+ * A predicate that returns {@code true} iff {@code boolean java.lang.Thread.setExtentLocalCache()}
+ * exists.
+ */
+public class HasSetExtentLocalCache implements BooleanSupplier {
```
peter-hofer, line 35: I don't see a need to have this class in addition to `HasExtentLocalCache` when both of them were changed in the same commit. The same goes for `HasSetScopedValueCache`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/HasExtentLocalCache.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A predicate that returns {@code true} iff {@code boolean java.lang.Thread.extentLocalCache()}
+ * exists.
+ */
+public class HasExtentLocalCache implements BooleanSupplier {
+
+    @Override
+    public boolean getAsBoolean() {
+        try {
+            Class.forName("java.lang.Thread").getDeclaredMethod("extentLocalCache");
```
peter-hofer, line 40: You can use `ReflectionUtil.lookupMethod(false, Thread.class, "extentLocalCache") != null`. (Not sure why you look up `Thread` by name here and elsewhere)

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/HasExtentLocalCache.java
```diff
@@ -0,0 +1,47 @@
...
+
+package com.oracle.svm.core.jdk;
+
+import java.util.function.BooleanSupplier;
+
+/**
+ * A predicate that returns {@code true} iff {@code boolean java.lang.Thread.extentLocalCache()}
+ * exists.
+ */
+public class HasExtentLocalCache implements BooleanSupplier {
```
peter-hofer, line 35: I think those classes should just go into `Target_java_lang_Thread.java` since it doesn't seem like they'll be needed anywhere else.
(You can also remove the Javadoc comment because it's quite self-explanatory)

peter-hofer (changes requested):
Thanks for looking into this!

Karm:
@zakkak My local setup:

* Graal 775187be8b (foivos/2023-01-12-jdk20-27-plus-compatibility)
* JDK 20 1c84050610e (January 19th)
* mandrel-packaging fd0d69647bb3 (fixed riscv header)
* mx master

Still ends up in this when trying to link a hello world app:

```
GraalVM Native Image: Generating 'hello' (executable)...
[1/8] Initializing...                                                                                    (3.1s @ 0.10GB)
 Version info: 'GraalVM 23.0.0-dev775187be8be8 Java 20-internal-adhoc.karm.jdk20 Mandrel Distribution'
 Java version info: '20-internal-adhoc.karm.jdk20'
 C compiler: gcc (redhat, x86_64, 8.5.0)
 Garbage collector: Serial GC (max heap size: unlimited)
[2/8] Performing analysis...  [*****]                                                                    (9.2s @ 1.39GB)
   3,045 (73.30%) of  4,154 types reachable
   3,726 (50.42%) of  7,390 fields reachable
  14,745 (46.76%) of 31,530 methods reachable
     908 types,    83 fields, and   479 methods registered for reflection
      57 types,    55 fields, and    52 methods registered for JNI access
[3/8] Building universe...                                                                               (1.5s @ 0.46GB)
[4/8] Parsing methods...      [*]                                                                        (1.3s @ 0.50GB)
[5/8] Inlining methods...     [***]                                                                      (0.5s @ 1.02GB)
[6/8] Compiling methods...    [***]                                                                      (7.5s @ 2.37GB)
[7/8] Layouting methods...    [*]                                                                        (1.2s @ 2.60GB)

[8/8] Creating image...                                                                                  (0.0s @ 2.91GB)
                        0.8s (2.9% of total time) in 19 GCs | Peak RSS: 3.31GB | CPU load: 8.61
Produced artifacts:
 /home/karm/tmp/svm_err_b_20230120T120729.220_pid335608.md (build_info)
Failed generating 'hello' after 27.0s.

The build process encountered an unexpected error:

> java.lang.RuntimeException: There was an error linking the native image: Linker command exited with 1

Linker command executed:
/bin/gcc -z noexecstack -Wl,--gc-sections -Wl,--version-script,/tmp/SVM-7042596703295858125/exported_symbols.list -Wl,-x -o /home/karm/tmp/hello hello.o /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/svm/clibraries/linux-amd64/liblibchelper.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libnet.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libnio.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libjava.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libfdlibm.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libzip.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/svm/clibraries/linux-amd64/libjvm.a -Wl,--export-dynamic -v -L/tmp/SVM-7042596703295858125 -L/home/karm/tmp/mandrel-23.0-SNAPSHOT/lib -L/home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/svm/clibraries/linux-amd64 -lz -lpthread -ldl -lrt

Linker command output:
Using built-in specs.
COLLECT_GCC=/bin/gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/8/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-redhat-linux
Configured with: ../configure --enable-bootstrap --enable-languages=c,c++,fortran,lto --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-shared --enable-threads=posix --enable-checking=release --enable-multilib --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-gcc-major-version-only --with-linker-hash-style=gnu --enable-plugin --enable-initfini-array --with-isl --disable-libmpx --enable-offload-targets=nvptx-none --without-cuda-driver --enable-gnu-indirect-function --enable-cet --with-tune=generic --with-arch_32=x86-64 --build=x86_64-redhat-linux
Thread model: posix
gcc version 8.5.0 20210514 (Red Hat 8.5.0-17) (GCC) 
COMPILER_PATH=/usr/libexec/gcc/x86_64-redhat-linux/8/:/usr/libexec/gcc/x86_64-redhat-linux/8/:/usr/libexec/gcc/x86_64-redhat-linux/:/usr/lib/gcc/x86_64-redhat-linux/8/:/usr/lib/gcc/x86_64-redhat-linux/
LIBRARY_PATH=/usr/lib/gcc/x86_64-redhat-linux/8/:/usr/lib/gcc/x86_64-redhat-linux/8/../../../../lib64/:/lib/../lib64/:/usr/lib/../lib64/:/usr/lib/gcc/x86_64-redhat-linux/8/../../../:/lib/:/usr/lib/
COLLECT_GCC_OPTIONS='-z' 'noexecstack' '-o' '/home/karm/tmp/hello' '-v' '-L/tmp/SVM-7042596703295858125' '-L/home/karm/tmp/mandrel-23.0-SNAPSHOT/lib' '-L/home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/svm/clibraries/linux-amd64' '-mtune=generic' '-march=x86-64'
 /usr/libexec/gcc/x86_64-redhat-linux/8/collect2 -plugin /usr/libexec/gcc/x86_64-redhat-linux/8/liblto_plugin.so -plugin-opt=/usr/libexec/gcc/x86_64-redhat-linux/8/lto-wrapper -plugin-opt=-fresolution=/tmp/cc1AQGz8.res -plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lgcc_s -plugin-opt=-pass-through=-lc -plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lgcc_s --build-id --no-add-needed --eh-frame-hdr --hash-style=gnu -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o /home/karm/tmp/hello -z noexecstack /usr/lib/gcc/x86_64-redhat-linux/8/../../../../lib64/crt1.o /usr/lib/gcc/x86_64-redhat-linux/8/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/8/crtbegin.o -L/tmp/SVM-7042596703295858125 -L/home/karm/tmp/mandrel-23.0-SNAPSHOT/lib -L/home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/svm/clibraries/linux-amd64 -L/usr/lib/gcc/x86_64-redhat-linux/8 -L/usr/lib/gcc/x86_64-redhat-linux/8/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib/gcc/x86_64-redhat-linux/8/../../.. --gc-sections --version-script /tmp/SVM-7042596703295858125/exported_symbols.list -x hello.o /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/svm/clibraries/linux-amd64/liblibchelper.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libnet.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libnio.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libjava.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libfdlibm.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/libzip.a /home/karm/tmp/mandrel-23.0-SNAPSHOT/lib/svm/clibraries/linux-amd64/libjvm.a --export-dynamic -lz -lpthread -ldl -lrt -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-redhat-linux/8/crtend.o /usr/lib/gcc/x86_64-redhat-linux/8/../../../../lib64/crtn.o
hello.o:(.data+0x358): undefined reference to `Java_java_lang_Thread_ensureMaterializedForStackWalk'
collect2: error: ld returned 1 exit status

```

Any idea what should I do differently?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/methodhandles/MethodHandleFeature.java
```diff
@@ -121,7 +122,10 @@ public void duringSetup(DuringSetupAccess access) {
         memberNameIsMethod = ReflectionUtil.lookupMethod(memberNameClass, "isMethod");
         memberNameIsConstructor = ReflectionUtil.lookupMethod(memberNameClass, "isConstructor");
         memberNameIsField = ReflectionUtil.lookupMethod(memberNameClass, "isField");
-        memberNameGetParameterTypes = ReflectionUtil.lookupMethod(memberNameClass, "getParameterTypes");
+        memberNameGetMethodType = ReflectionUtil.lookupMethod(memberNameClass, "getMethodType");
+
+        Class<?> methodTypeClass = access.findClassByName("java.lang.invoke.MethodType");
```
gilles-duboscq, line 127: why is this happening reflectively? `MethodType` and `parameterArray` are public

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/HasExtentLocalCache.java
zakkak: My bad, copy pasted code without giving it much thought.

zakkak: Shouldn't I set `optional` to true in `lookupMethod` though?

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
```diff
@@ -2282,4 +2283,25 @@ public SnippetTemplate.SnippetInfo getSnippet(BigIntegerSnippets.Templates templ
...
+            return false;
+        }
+        return true;
+    }
+
+    private static void registerThreadPlugin(InvocationPlugins plugins, Replacements replacements) {
+        InvocationPlugin threadPlugin = new InvocationPlugin("ensureMaterializedForStackWalk", Object.class) {
+            @Override
+            public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver, ValueNode object) {
+                b.add(new BlackholeNode(object));
```
gergo-, line 2300: I think it would be helpful to add an extra `String` property to `BlackholeNode` so you can pass `"Thread.ensureMaterializedForStackWalk"` here and make it show up in IGV as a node labeled `Blackhole Thread.ensureMaterializedForStackWalk`. Otherwise people seeing a random blackhole in a graph might be very confused where it comes from, since this is a new use case for this node.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/methodhandles/MethodHandleFeature.java
zakkak: My bad, copy pasted code without giving it much thought.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
zakkak: Do I understand correctly that the following would suffice?

```diff
modified   compiler/src/org.graalvm.compiler.nodes/src/org/graalvm/compiler/nodes/debug/BlackholeNode.java
@@ -51,12 +51,18 @@ public final class BlackholeNode extends FixedWithNextNode implements LIRLowerab
 
     public static final NodeClass<BlackholeNode> TYPE = NodeClass.create(BlackholeNode.class);
     @Input ValueNode value;
+    @OptionalInput String comment;
 
     public BlackholeNode(ValueNode value) {
         super(TYPE, StampFactory.forVoid());
         this.value = value;
     }
 
+    public BlackholeNode(ValueNode value, String comment) {
+        this(value);
+        this.comment = comment;
+    }
+
     public ValueNode getValue() {
         return value;
     }
modified   compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
@@ -2297,7 +2297,7 @@ public class StandardGraphBuilderPlugins {
         InvocationPlugin threadPlugin = new InvocationPlugin("ensureMaterializedForStackWalk", Object.class) {
             @Override
             public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver, ValueNode object) {
-                b.add(new BlackholeNode(object));
+                b.add(new BlackholeNode(object, "ensureMaterializedForStackWalk"));
                 return true;
             }
         };
```

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Target_java_lang_Thread.java
zakkak: Done

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/HasSetExtentLocalCache.java
zakkak: Done

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/HasExtentLocalCache.java
zakkak: Done

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
gergo-: Yes, that's about it. You have to remove the `@OptionalInput` on the `String`, only nodes can (and must) be inputs. Other properties are not annotated.
You should also add something like `nameTemplate = "Blackhole {p#reason}"` to the class's `@NodeInfo` so that the reason is shown as part of the name of the node itself. This might require `reason` to be non-`null`, so probably it should have a default value of `""`.

zakkak:
> Any idea what should I do differently?

@Karm the key difference I see is that you are building your own JDK 20 while I am using `Temurin-20+30-202301110331 (build 20-beta+30-202301110331)`

A mistake I made while working on this, was using an old build instead of the latest one generated by mandrel-packaging when testing... Although trivial, please make sure you are using the latest build. 

You can a successfull build here https://github.com/graalvm/mandrel/actions/runs/3958942405
And I started another one with the latest updates https://github.com/graalvm/mandrel/actions/runs/3970219099

Karm:
> > Any idea what should I do differently?
> 
> @Karm the key difference I see is that you are building your own JDK 20 while I am using `Temurin-20+30-202301110331 (build 20-beta+30-202301110331)`
> 
> A mistake I made while working on this, was using an old build instead of the latest one generated by mandrel-packaging when testing... Although trivial, please make sure you are using the latest build.
> 

I am sure that is not the case. I did code changes before (printf debug...), I know the build script correctly cleans up and uses the correct dir.

> You can a successfull build here https://github.com/graalvm/mandrel/actions/runs/3958942405 And I started another one with the latest updates https://github.com/graalvm/mandrel/actions/runs/3970219099

I can se that.  Yet no luck.

I got paranoid and did this:

Graal:
```
$ git clone --branch 2023-01-12-jdk20-27-plus-compatibility https://github.com/zakkak/mandrel.git graal-foivos
```

And I used downloaded Temurin instead of my own JDK build. It didn't help:

## OK
Aforementioned Ubuntu Github action:
```
 Version info: 'GraalVM 23.0.0-dev44c1201d Java 20-beta+31-202301130331 Mandrel Distribution'
 Java version info: '20-beta+31-202301130331'
 C compiler: gcc (linux, x86_64, 11.3.0)
```

## NOK
My CentOS 8 Stream workstation:
```
 Version info: 'GraalVM 23.0.0-dev Java 20-beta+31-202301130331 Mandrel Distribution'
 Java version info: '20-beta+31-202301130331'
 C compiler: gcc (redhat, x86_64, 8.5.0)
```
Dunno. Could it be the older toolchain I am using?

zakkak:
> Dunno. Could it be the older toolchain I am using?

No, apparently some change after the feedback breaks the PR.
https://github.com/graalvm/mandrel/actions/runs/3970219099/jobs/6806021385 is not working !
I'll have a closer look.

Karm:
> > Dunno. Could it be the older toolchain I am using?
> 
> No, apparently some change after the feedback breaks the PR. https://github.com/graalvm/mandrel/actions/runs/3970219099/jobs/6806021385 is not working ! I'll have a closer look.

I am glad I am not out of my mind :-D

I did try on a clean room Ubuntu 20 VM to check how a different toolchain behaves and it's the same...

https://karms.biz/pastebin/ubuntu-20-cleanroom.txt

zakkak:
Sorry for that @Karm 

The issue was in `hasEnsureMaterializedForStackWalk()`

```diff
modified   compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
@@ -2286,7 +2286,7 @@ public class StandardGraphBuilderPlugins {
 
     private static boolean hasEnsureMaterializedForStackWalk() {
         try {
-            Thread.class.getDeclaredMethod("ensureMaterializedForStackWalk");
+            Thread.class.getDeclaredMethod("ensureMaterializedForStackWalk", Object.class);
         } catch (Exception e) {
             return false;
```

Should be fixed now if you want to give it another go.

Karm:
@zakkak It works for me now, JDK 20 (1c840506 from Jan 19th).

gilles-duboscq:
For issues or PRs like this one that relate to compatibility with upstream OpenJDK changes, i'd like to introduce a new label so that we can easily find them, filter notifications etc.
This is relevant for a few different teams (Mandrel, Native Image, Compiler, labsjdk, ...) and i think a lael would help collaboration.
Discussing with @zakkak, he proposed `openjdk-updates` which i think is a good name for such a label.
Any comments/remarks?
/cc @pejovica, @dougxc, @marwan-hallaoui

dougxc:
Yes, I think `openjdk-updates` is a suitable label for such changes.

compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
dougxc: Looks good. This is the same change @gilles-duboscq made on a similar PR.

marwan-hallaoui:
Here is the integration PR(https://github.com/oracle/graal/pull/5831) for the current changes related to JDK-20.
cc @zakkak @gilles-duboscq @dougxc

zakkak:
Superseded by https://github.com/oracle/graal/pull/5831

PR #5775, opened 2023-01-17
[GR-43616] Avoid NPEs when `ExitStatus.of()` is used.

graalvmbot:


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/util/ExitStatus.java
```diff
@@ -33,15 +33,16 @@ public enum ExitStatus {
...
+    UNKNOWN(255);
 
     public static ExitStatus of(int status) {
         for (ExitStatus s : values()) {
             if (s.getValue() == status) {
                 return s;
             }
         }
-        return null;
+        return UNKNOWN;
```
JaroslavTulach, line 45: Exit code 137 is common and important - maybe you should have a constant for it. Returning `UNKNOWN` may hide the actual reason why it is happening (out of memo ry of Linux system) and may make it hard to understand how that can be fixed (close Chrome & Firefox & co. or set a swap space).

JaroslavTulach: At least I was [quite desperate](https://graalvm.slack.com/archives/CN9KSFB40/p1660539227687399) when I got 137 for the first time.

fniephaus: This PR will fix the NPE. Also, this method only has one caller and there, the exit code will still be printed.
Anyway, I'm also looking into a better error message in case of OOMEs, although as part of another PR.

fniephaus: See https://github.com/oracle/graal/pull/5782

PR #5816, opened 2023-01-23
Remove BytecodeFrame duplicates

adinn:
This is a compiler change that was factored out of PR #5475 (Debug info footprint improvements part 2) for separate review and testing.

The PR significantly reduces `BytecodeFrame` storage by merging the linear chains of BytecodeFrame instances associated 1-1 with each `Infopoint` of a `CompilationResult` into an equivalent tree that shared common frames.


adinn:
@olpaw I think this is ready for a re-test and re-review. It is still using a workaround for the broken equals method on BytecodeFrame but that workaround should still be correct. Before removing it I'd prefer to retry the internal gate to see whether the patch is still breaking some of the .so builds that previously stopped it being pushed. If you would prefer I can remove the special case comparison and revert to reliance on `BytecodeFrame::equals`.

PR #5855, opened 2023-01-30
[GR-38651] Tweak debuginfotest for aarch64.

graalvmbot:


adinn:
@fniephaus

That change looks fine and I verified it runs correctly on Linux/AArch64.

n.b. Full disclosure involves a small additional comment:

The correctness of the current stepping logic depends on two details that determine the instruction sequence generated in the frame header:

1. The frame size is valid for use as an immediate operand to a sub instruction
2. The test is run with the default option for `preserveFramePointer` (`false`)

Neither of these is a concern as far as robustness of the test against changes to the compiler code is concerned.

The current frame size (0x#a0) is unlikely ever to change enough to exceed the limit for immediates (12 bits) and, if it did, the test would fail since we would observe the instructions that adjust sp in a different order and not get a match.

If the default for `preserveFramePointer` were to be changed then the prologue would include an extra instruction, `add fp, sp, #frameSize`, which would not have been stepped over. This should make the test fail because the second check of the stack-passed parameters would be done using the pre-push stack offsets rather than the post-push ones. Indeed, I verified this is the case by running the test with `-H:+PreserveFramePointer`.

adinn (approved):
Looks good. Verified on Linux/AArch64.

PR #5860, opened 2023-01-31
Word-by-word card table scanning

petermz:
This is a port of a [Hotspot fix](https://bugs.openjdk.org/browse/JDK-7068625). The idea is, currently card table is scanned byte by byte. But the majority of cards are clean, so we could move by 8-byte steps as long as none of those 8 cards are marked, and fall back to single-byte steps otherwise.

Testing shows this speeds up things a bit, especially when heap is large. The charts below show length of the `blackenDirtyCardRoots` stage for HyperAlloc running with `OnlyIncrementally` collection policy and different heap sizes.

HyperAlloc-Xmx1g
HyperAlloc-Xmx32g

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/remset/AlignedChunkRememberedSet.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/remset/AlignedChunkRememberedSet.java
@@ -55,6 +55,11 @@ final class AlignedChunkRememberedSet {
     }
 
     @Fold
+    public static int wordSize() {
+        return ConfigurationValues.getTarget().wordSize;
+    }
+
+    @Fold
     public static UnsignedWord getHeaderSize() {
         UnsignedWord headerSize = getFirstObjectTableLimitOffset();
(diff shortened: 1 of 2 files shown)


oubidar-Abderrahim:
Hi, Thank you for contributing to GraalVM, Is there a specific member of the GraalVM team you prefer to review your PR?

petermz:
Hi Abderrahim --
No, I don't think I have any preference

oubidar-Abderrahim:
Hi @wirthi, could you please assign someone from the team to review this PR? Once the PR gets approved I can integrate it internally, Thank you.

christianhaeubl (approved):
Thanks for working on the performance of the card table scanning. The change looks good, I will integrate this change after I did some additional testing and benchmarking.

christianhaeubl:
I did some benchmarking and your PR had a negative impact on a small number of benchmarks. I opened a PR that potentially replaces your PR: https://github.com/oracle/graal/pull/5963. The implementation there, is closer to the one in HotSpot and doesn't seem to have any negative performance impacts.

If possible, please re-run your benchmarks with the PR from above and verify that the speedups are similar to the ones that you saw with your implementation.

petermz:
I've measured speedups with your PR, they are about the same. Thanks!

christianhaeubl:
Thanks for checking, then I am closing this PR in favor of https://github.com/oracle/graal/pull/5963.

PR #5943, opened 2023-02-09
[GR-44105] Enable building graalvm with debug info in CE.

graalvmbot:


sdk/mx.sdk/mx_sdk_vm_impl.py
```diff
@@ -1231,8 +1250,10 @@ def contents(self):
             ]
             if _debug_images():
                 build_args += ['-ea', '-O0', '-H:+PreserveFramePointer', '-H:-DeleteLocalSymbols']
-            if _get_svm_support().is_debug_supported(image_config):
+            if _get_svm_support().generate_debug_info(image_config):
```
gilles-duboscq, line 1253: * `-g` is now passed by default also while building GraalVM CE's native images on linux and windows (like what's already happening for EE)
* It's also passed on darwin if the user explicitly requests `--debug-images`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -659,8 +659,17 @@ public static Path getDebugInfoSourceCacheRoot() {
         }
     }
 
-    @Option(help = "Strip debug info from the binary.")//
-    public static final HostedOptionKey<Boolean> StripDebugInfo = new HostedOptionKey<>(true);
+    @Option(help = "Use a separate file for debug info.")//
+    public static final HostedOptionKey<Boolean> StripDebugInfo = new HostedOptionKey<>(OS.getCurrent() != OS.DARWIN, SubstrateOptions::validateStripDebugInfo);
+
+    private static void validateStripDebugInfo(HostedOptionKey<Boolean> optionKey) {
```
gilles-duboscq, line 665: Ideally this flag should be called `SeparateDebugInfo` rather than `StripDebugInfo` since "strip" means different things on different platform:
* stripping can have an effect on linux even if no debug info was generated (e.g., stripping symbols)
* on windows debug info is always separated to a pdb file. That pdb file can also be ["stripped"](https://learn.microsoft.com/en-us/cpp/build/reference/pdbstripped-strip-private-symbols) to only contain public symbols and stack walking data. Currently that's controlled by `DeleteLocalSymbols` on windows.

I think the PR is already a step in the right direction but maybe later we should look into how to improve/rename those options to have a more uniform meaning on different platforms.

gilles-duboscq:
@adinn, FYI, with this change debug information is will start to be generated by default during GraalVM CE builds.

adinn:
@gilles-duboscq Ok by me. Does that also mean that compilation defaults to -O0?

zakkak:
> @gilles-duboscq Ok by me. Does that also mean that compilation defaults to -O0?

No, that would be a disaster :)

gilles-duboscq:
> @gilles-duboscq Ok by me. Does that also mean that compilation defaults to -O0?

No, `-O0` is not passed by default. Using `--debug-images` does pass `-ea -O0 -H:+PreserveFramePointer -H:-DeleteLocalSymbols` but that's only by explicit request.

PR #5961, opened 2023-02-13
[GR-43952] Emit `DW_AT_(comp_dir|stmt_list)` attr for compile units.

graalvmbot:
Emit `DW_AT_comp_dir` and `DW_AT_stmt_list` attributes for all DIEs with the `DW_TAG_compile_unit` tag. `gprofng` and other tools expect these attributes to always be present.


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
```diff
@@ -927,6 +927,10 @@ private int writeBuiltInUnitAbbrev(@SuppressWarnings("unused") DebugContext cont
         pos = writeFlag(DwarfDebugInfo.DW_CHILDREN_yes, buffer, pos);
         pos = writeAttrType(DwarfDebugInfo.DW_AT_language, buffer, pos);
         pos = writeAttrForm(DwarfDebugInfo.DW_FORM_data1, buffer, pos);
+        pos = writeAttrType(DwarfDebugInfo.DW_AT_comp_dir, buffer, pos);
```
gilles-duboscq, line 930: I think also need to emit the corresponding data in `com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl#writeBuiltInUnit`

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
```diff
@@ -983,6 +987,10 @@ private int writeArrayUnitAbbrev(@SuppressWarnings("unused") DebugContext contex
         pos = writeAttrForm(DwarfDebugInfo.DW_FORM_data1, buffer, pos);
         pos = writeAttrType(DwarfDebugInfo.DW_AT_name, buffer, pos);
         pos = writeAttrForm(DwarfDebugInfo.DW_FORM_strp, buffer, pos);
+        pos = writeAttrType(DwarfDebugInfo.DW_AT_comp_dir, buffer, pos);
```
gilles-duboscq, line 990: Same thing, i think you need to write the corresponding values in `com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl#writeArrayTypeUnit`

adinn:
Hi @fniephaus 

Yes, your patch needs to be supplemented with some extra code that writes the content for these extra attributes into the info records.

I think it's fairly obvious what to do for the `comp_dir` attribute. For class units that model an instance class we call `classEntry.getCachePath()` which uses the root dir of the sources cache when it has an associated file or `""` if it has no file.

For an array type or a type in the built-in unit there is no associated source file. I guess `grpofng` will be happy just to use `""`. So, this attribute needs to be emitted in method 
`DwarfInfoSectionImpl#writeBuiltInUnit`:

        log(context, "  [0x%08x]     language  %s", pos, "DW_LANG_Java");
        pos = writeAttrData1(DwarfDebugInfo.LANG_ENCODING, buffer, pos);
    +   String compilationDirectory = "";
    +   log(context, "  [0x%08x]     comp_dir  0x%x (%s)", pos, debugStringIndex(compilationDirectory), compilationDirectory);
    +   pos = writeStrSectionOffset(compilationDirectory, buffer, pos);
        . . .

and in method `DwarfInfoSectionImpl#writeArrayTypeUnit`:

        log(context, "  [0x%08x]     name 0x%x (%s)", pos, debugStringIndex(name), name);
        pos = writeStrSectionOffset(name, buffer, pos);
    +   String compilationDirectory = "";
    +   log(context, "  [0x%08x]     comp_dir  0x%x (%s)", pos, debugStringIndex(compilationDirectory), compilationDirectory);
    +   pos = writeStrSectionOffset(compilationDirectory, buffer, pos);
        . . .

Let us hope that once `gprofng` does not trip up over the lack of a `file` attribute once it gets past this hurdle.

It is less clear what to do for the `stmt_list` attribute. I ran into this problem with `perf` when generating a class unit for a class which had no associated code range (i.e. all its methods only ever occurred inline in methods of other classes). I had to generate a dummy record in the `debug_line` section for each such class, providing with a header, an an empty file/dir table and an empty line number table.

We might actually be able to get away with setting `stmt_list` to zero because built-in and array units have no methods. So, `gprofng` will probably never look at the content at offset zero (which will probably be legitimate content for some real class unit with compiled code of its own). However, it might be safer to make sure that the line generator adds an empty record for the built-in and array units. We can probably use the same record and ensure we generate it first. Indeed we may also find that a single empty record also works for classes which have no compiled code.

So, the basic change to try is as per above e..g for `DwarfInfoSectionImpl#writeBuiltInUnit`:

        log(context, "  [0x%08x]     language  %s", pos, "DW_LANG_Java");
        pos = writeAttrData1(DwarfDebugInfo.LANG_ENCODING, buffer, pos);
    +   final String compilationDirectory = "";
    +   log(context, "  [0x%08x]     comp_dir  0x%x (%s)", pos, debugStringIndex(compilationDirectory), compilationDirectory);
    +   pos = writeStrSectionOffset(compilationDirectory, buffer, pos);
    +   final int lineIndex = 0;
    +   log(context, "  [0x%08x]     stmt_list  0x%08x", pos, lineIndex);
    +   pos = writeAttrData4(lineIndex, buffer, pos);
        . . .

Likewise for method `DwarfInfoSectionImpl#writeArrayTypeUnit`.

You can try that for now to see if it is enough to keep `gprofng` happy. Meanwhile, I will look into what is needed to ensure there is an empty record at the start of the debug_line section and also see if we can reuse this for classes which have no compiled code.

BTW, I was surprised to see that your branch passed the debuginfotest gate. It turns out to be a fluke. gdb fails to read your debug info because the abbrev record layouts do not match the info record content. The script fails in the code recently introduced by Peter Hofer when trying to read the object header size. I think the problem is that the script quits the gdb session rather than exiting with an error code. The test thinks all is well and continues. The test should probably use class Checker to check the against a regexp and return a regexp match value for the hub size. It is guaranteed to exit if no match is found.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfAbbrevSectionImpl.java
adinn: Yes, @gilles-duboscq  is right. See my comment below.

fniephaus:
> Yes, your patch needs to be supplemented with some extra code that writes the content for these extra attributes into the info records.

Apologies, I couldn't find the location where those attributes are written. So I ran all the tests, tried to reproduce the error with `gprofng` and when everything passed, I thought I open a PR and hope that a reviewer points out what I'm doing wrong. I hope this is now addressed correctly...

> I think it's fairly obvious what to do for the comp_dir attribute. For class units that model an instance class we call classEntry.getCachePath() which uses the root dir of the sources cache when it has an associated file or "" if it has no file.

It turns out that `gprofng` doesn't like `comp_dir` to be set to `""`. EE debuginfo seems to emit `System.getProperty("user.dir")`, but I felt that the cache path seems better. Then I decided to move cache path out of `FileEntry` and into `DebugInfoBase`, which allowed me to simplify the code base a bit. I also pushed [another minor refactoring](https://github.com/oracle/graal/pull/5961/commits/9f38d668c92da2e096288e6b2874a2ea1f6dd4d2).

> We might actually be able to get away with setting stmt_list to zero because built-in and array units have no methods.

I agree, this should do the trick.

> BTW, I was surprised to see that your branch passed the debuginfotest gate.

So was I! [I've adjusted `testhello.py`](https://github.com/oracle/graal/pull/5961/commits/82172ad90134177c1533d9b59acdc8e07fe977f1) so that it fails the build if a `gdb.error` is thrown. That way, we don't even need a checker for Peter's new code.

adinn (approved):
Looks good.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
```diff
@@ -258,8 +264,11 @@ public void installDebugInfo(DebugInfoProvider debugInfoProvider) {
         /* Reference alignment must be 8 bytes. */
         assert oopAlignment == 8;
 
-        /* Ensure we have a null string in the string section. */
+        cachePath = debugInfoProvider.getCachePath() != null ? debugInfoProvider.getCachePath().toString() : "";
+
+        /* Ensure we have a null string and the cache path in the string section. */
         stringTable.uniqueDebugString("");
+        stringTable.uniqueDebugString(cachePath);
```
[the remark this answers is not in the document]
fniephaus: Done

PR #5962, opened 2023-02-13
List<Value> parameter may get List<anything>

JaroslavTulach:
While working on https://github.com/enso-org/enso/issues/5590 I realized there is a limitation in _host interop conversions_. If you have a `String[]` host interop value and you try to pass it into `List<org.graalvm.polyglot.Value>` parameter, everything works properly. However, if you have `List<String>` host interop value, one gets `ClassCastException` as the conversion to `Value` doesn't happen.

b2afd2f2aa7298a3c237580aeeda95d51ed64627 contains a test showing the good (with `[]`) and bad (with `List`) behavior.

d60e7822cf9767c7dda6b8cf71d139844397814b proposes a fix. I am not saying it is the correct fix, but it seems to work. The idea is to treat `java.util.List` target type in a special way and never pass it directly as _host value_. Rather always wrap it with a delegate to `InteropLibrary.readArrayElement`.

CCing @woess, @chumer

### DIFF
--- a/truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/HostAccessTest.java
+++ b/truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/HostAccessTest.java
@@ -2396,6 +2396,45 @@ public class HostAccessTest extends AbstractHostAccessTest {
         }
     }
 
+    public final class WorkWithArray {
+        WorkWithArray() {
+        }
+
+        public List<String> stringsAsList() {
+            return Arrays.asList("Hello", "World");
+        }
+
+        public String[] stringsAsArray() {
+            return Arrays.asList("Hello", "World").toArray(String[]::new);
+        }
+
+        public String concatenate(List<Value> values) {
+            StringBuilder sb = new StringBuilder();
+            for (Value val : values) {
+                sb.append(val.toString());
+            }
+            return sb.toString();
+        }
+    }
+
+    private void checkListOfValue(String factoryName) throws Exception {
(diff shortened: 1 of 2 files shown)


JaroslavTulach:
We have found only [one workaround](https://github.com/enso-org/enso/pull/5592#discussion_r1104029551):
```java
  public static Column fromItems(String name, List<Value> items) {
    InferredBuilder builder = new InferredBuilder(items.size());
    for (Value item : items) {
      Object converted = Polyglot_Utils.convertPolyglotValue(item);
    for (Object item : items) {
      Object converted = item instanceof Value v ? Polyglot_Utils.convertPolyglotValue(v) : item;
    }
}
```
e.g. be ready that even we request `List<Value>` we may get items that aren't a `Value`.

chumer:
Thanks for the report. I assume we have the same problem with `Map` `Iterable` and other types as well.
Can you file this as an issue? 

I think the fix is not quite correct, i think it can cause issues with List<HostType> mappings.

JaroslavTulach:
> Thanks for the report. Can you file this as an issue?

Reported as a really nice round issue  #6000. I guess we can close this pull request then. Feel free to copy/reuse the unit test [b2afd2f](https://github.com/oracle/graal/pull/5962/commits/b2afd2f2aa7298a3c237580aeeda95d51ed64627).

> I assume we have the same problem with `Map` `Iterable` and other types as well. 

That's very likely.

> I think the fix is not quite correct, i think it can cause issues with List mappings.

[d60e782](https://github.com/oracle/graal/pull/5962/commits/d60e7822cf9767c7dda6b8cf71d139844397814b) isn't production ready fix. It's too trivial to not cause regressions. Surprisingly all `truffle$ mx unittest` passed OK.

PR #5964, opened 2023-02-13
Represent boolean flags as `unsigned char`, instead of `char`

koutheir:
The `char` type in C is sometimes signed, and sometimes unsigned.

Boolean flags should be represented as unsigned, so this patch makes sure of that.


vjovanov (approved):

vjovanov:
Was this causing a problem on some platforms?

koutheir:
Yes.

PR #5965, opened 2023-02-13
Throw no exceptions when actual extraction should be skipped

koutheir:
When the `skip` flag is true, any `BufferUnderflowException`s should be reported
to the caller by returning `null`.


koutheir:
The failure in checks is unrelated to this change.

christianwimmer:
What problem is this PR fixing? When I last looked at the annotation code not so long ago, I added the necessary catch of `BufferUnderflowException` more towards the entry points, in #5625 because I concluded that catching the exception early actually screws up the annotation parsing code more than it helps.

koutheir:
I did not have the changes made by [#5625](https://github.com/oracle/graal/pull/5625) delivered in our Graal EE 22.3.1. If you think this change is unnecessary, then we can close it.

loicottet:
I agree that #5625 is the best way to fix this. It matches the JDK behavior in `AnnotationParser`

PR #5969, opened 2023-02-13
Detect usage of resources loaded through `jdk.internal.jimage.BasicImageReader`

koutheir:
Add support, in `native-image-agent`, for detection of resources loaded through `jdk.internal.jimage.BasicImageReader` and classes derived from it.

### DIFF
--- a/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
+++ b/substratevm/src/com.oracle.svm.agent/src/com/oracle/svm/agent/BreakpointInterceptor.java
@@ -616,6 +616,60 @@ final class BreakpointInterceptor {
         return true;
     }
 
+    private static boolean jimageReaderFindLocation1(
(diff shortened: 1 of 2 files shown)


koutheir:
The failure in checks do not seem related to this change.

peter-hofer:
How does application code use these methods? The problem with instrumenting methods within the JDK is that we can't distinguish JDK-internal usage from usage in user code and cannot accurately filter, so it is preferable to instrument those methods that are called from application code directly.

koutheir:
I think it was `Module.getResourceAsStream()` and methods like it.

peter-hofer (changes requested):
Thank you, please change the PR accordingly.

koutheir:
Too many methods seem to be calling `BasicImageReader.findLocation(String)` and `BasicImageReader.findLocation(String,String)`. I don't think I can find all of them without extensive testing and tracing.

PR #5971, opened 2023-02-13
`System.arraycopy()` should not copy `Object[]` into `Object`

koutheir:
`System.arraycopy​(Object src, int, Object dest, int, int)` should throw an `ArrayStoreException` when:
- `src` is an instance of `Object[]`, and
- `dest` is an instance of `Object`, and
- `src.getClass()` != `dest.getClass()`.

In the previous implementation, the following condition is satisfied:
```java
DynamicHub.toClass(toHub).isAssignableFrom(DynamicHub.toClass(fromHub))
```
where:
- `DynamicHub.toClass(toHub)` is `Object`, and
- `DynamicHub.toClass(fromHub)` is `Object[]`

because an `Object` variable can be assigned an `Object[]` value.

In this situation, `JavaMemoryUtil.copyObjectArrayForward()` is called, and the behavior is undefined.

The correct behavior should be throwing an `ArrayStoreException`.


christianhaeubl (approved):
Thanks for finding and fixing this issue. I will merge a slightly modified version to master.

PR #5972, opened 2023-02-13
Refactoring of calls to `FrameAccess.getWordKind()`.

koutheir:
Refactoring of calls to `FrameAccess.getWordKind()`.


christianwimmer:
Why this change? I don't see how it fixes a problem or makes anything better.

koutheir:
It's a refactoring change, so it has no semantic effect.

peter-hofer:
Thank you @koutheir , but there is no benefit to this change.

PR #5974, opened 2023-02-13
Support headless AWT without linking against unneeded native code

koutheir:
When building a native image in headless AWT mode (*i.e.,* `-Djava.awt.headless=true`), allow building without linking against unneeded native code.


christianwimmer:
The new AWT mode where we always use dynamic linking for AWT makes this PR unnecessary, so there is no real point in merging it.

koutheir:
@christianwimmer, is the new AWT mode going to be merged into the `22.3.*` branch? If that is not the case, then please merge these changes into the `22.3.*` branch.

christianwimmer:
> If that is not the case, then please merge these changes into the 22.3.* branch.

No, we do not want to give the false impression that AWT support is working in the 22.3 branch.

PR #5975, opened 2023-02-13
More checks in `java.lang.reflect.Array.get*()`.

koutheir:
More `null` checks in `java.lang.reflect.Array.get*()`.


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangReflectSubstitutions.java
```diff
@@ -44,6 +44,9 @@ final class Target_java_lang_reflect_Array {
 
     @Substitute
     private static boolean getBoolean(Object array, int index) {
+        if (array == null) {
+            throw new NullPointerException();
+        }
```
christianwimmer, line 49: Please use `} else if(...` to avoid an unnecessary line (applies to every method)

koutheir:
The failure in checks is unrelated to this change.

christianwimmer (approved):
@oubidar-Abderrahim please integrate this PR

PR #5976, opened 2023-02-13
`jdk.internal.loader.BootLoader.getDefinedPackage()` should return `null` for undefined packages.

koutheir:
`jdk.internal.loader.BootLoader.getDefinedPackage()` should return `null` for undefined packages.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
@@ -758,15 +758,20 @@ final class Target_jdk_internal_loader_ClassLoaders {
 @TargetClass(value = jdk.internal.loader.BootLoader.class)
 final class Target_jdk_internal_loader_BootLoader {
 
+    /**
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JavaLangSubstitutions.java
```diff
@@ -758,15 +758,20 @@ final class Target_jdk_internal_loader_ClassLoaders {
...
     @Substitute
-    static Package getDefinedPackage(String name) {
+    public static Package getDefinedPackage(String name) {
         if (name != null) {
-            Target_java_lang_Package pkg = new Target_java_lang_Package(name, null, null, null,
-                            null, null, null, null, null);
-            return SubstrateUtil.cast(pkg, Package.class);
-        } else {
-            return null;
+            for (Package pkg : (Iterable<Package>) packages()::iterator) {
```
christianwimmer, line 768: Why convert the `Stream` to an `Iteraable` instead of using proper stream expressions (`filter`, `empty`)?

koutheir: Because it seemed simpler.

PR #5979, opened 2023-02-13
Persist the `java.awt.headless` Java system property.

koutheir:
If the `java.awt.headless` Java system property is defined during the native image build, then it is defined to the same value at runtime too. A lot of behavior in AWT is based on this value, so it should be the same at build and run time, so that classes initialized at build time have the expected state at run time.


fniephaus:
I think this may be something that users should be able to control. What if I do want to run my app headfull and headless? We are going to allow users to "burn in" properties at build-time and obviously, `java.awt.headless` would be a very interesting property to burn in when using AWT/Swing code.

koutheir:
> What if I do want to run my app headfull and headless?

Then you should build two different native images, for the moment. At least until more AWT classes are initialized at run time instead of build time.

fniephaus:
>Then you should build two different native images, for the moment. 

But do you really need to? We are dynamically linking AWT so you will have headless and headful native libraries anyway. How much would you save if you burn in the property?

koutheir:
This is not about "saving" anything. It is about making sure that classes initialized at build time while assuming headful versus headless, continue to work properly at runtime because their assumption remains valid at runtime. Building a class with a headful configuration and then using a headless configuration at runtime causes unexpected behavior.

PR #5980, opened 2023-02-13
Interrupting a thread that is not alive need not have any effect

koutheir:
Interrupting a thread that is not alive need not have any effect. See [Thread.interrupt()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Thread.html#interrupt()).

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Target_java_lang_Thread.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Target_java_lang_Thread.java
@@ -458,6 +458,13 @@ public final class Target_java_lang_Thread {
      */
     @Substitute
     void interrupt0() {
+        Thread thread = JavaThreads.fromTarget(this);
+
+        // Interrupting a thread that is not alive need not have any effect.
+        if (!PlatformThreads.isAlive(thread)) {
+            return;
+        }
+
         if (JavaVersionUtil.JAVA_SPEC <= 11) {
             interruptedJDK11OrEarlier = true;
         } else {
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Target_java_lang_Thread.java
```diff
@@ -458,6 +458,13 @@ public static boolean interrupted() {
      */
     @Substitute
     void interrupt0() {
+        Thread thread = JavaThreads.fromTarget(this);
+
+        // Interrupting a thread that is not alive need not have any effect.
+        if (!PlatformThreads.isAlive(thread)) {
```
christianhaeubl, line 464: Did you see any particular issue without this check?

The check that you added is racy and it is also something that HotSpot does not do (i.e., HotSpot doesn't look at the Java thread state).

christianhaeubl:
The Native Image behavior matches the HotSpot behavior from OpenJDK 17 and 20, so I am closing this issue.

koutheir:
@christianhaeubl, can you please merge these changes into the `22.3.*` branch?

christianhaeubl:
I will merge a modified version of your PR to GraalVM 22.3.

koutheir:
@christianhaeubl, did you actually merge this PR to GraalVM 22.3?

christianhaeubl:
Yes.

koutheir:
@christianhaeubl, in [this commit](03488e5967cf88ff52d1c945bba8e05e22a9ef08), you replaced the call to `PlatformThreads.isAlive()` by a call to `PlatformThreads.canBeInterrupted()`, whose implementation doesn't seem to check for terminated threads, as `acquireThreadData()` only returns `null` for **detached** threads, and not for terminated threads.

Am I missing something? Or did that commit introduce a regression on JDK11?

koutheir:
Could you please elaborate on which use cases break when using `isAlive()`?

PR #5991, opened 2023-02-14
Check UTF-8 encoding in JNI functions

koutheir:
The following JNI functions accept some UTF-8 strings as arguments:
- `FindClass`
- `GetFieldID`
- `GetMethodID`
- `GetStaticFieldID`
- `GetStaticMethodID`
- `RegisterNatives`

This patch checks that the UTF-8 strings provided to these methods are correctly encoded. If they are not, then the proper [JNI error reporting behavior](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html) is performed, mainly:
- `FindClass` throws `NoClassDefFoundError`.
- `GetFieldID` and `GetStaticFieldID` throw `NoSuchFieldError`.
- `GetMethodID`, `GetStaticMethodID` and `RegisterNatives` throw `NoSuchMethodError`.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
@@ -341,6 +341,10 @@ public final class JNIFunctions {
     @CEntryPointOptions(prologue = JNIEnvEnterPrologue.class, prologueBailout = ReturnNullHandle.class)
     static JNIObjectHandle FindClass(JNIEnvironment env, CCharPointer cname) {
         CharSequence name = Utf8.wrapUtf8CString(cname);
+        if (name == null) {
+            throw new NoClassDefFoundError("Class name is either null or invalid UTF-8 string");
+        }
+
         Class<?> clazz = JNIReflectionDictionary.singleton().getClassObjectByName(name);
         if (clazz == null) {
             throw new NoClassDefFoundError(name.toString());
@@ -363,7 +367,15 @@ public final class JNIFunctions {
         for (int i = 0; i < nmethods; i++) {
             JNINativeMethod entry = (JNINativeMethod) p;
             CharSequence name = Utf8.wrapUtf8CString(entry.name());
+            if (name == null) {
+                throw new NoSuchMethodError("Method name at index " + i + " is either null or invalid UTF-8 string");
+            }
+
             CharSequence signature = Utf8.wrapUtf8CString(entry.signature());
+            if (signature == null) {
+                throw new NoSuchMethodError("Method signature at index " + i + " is either null or invalid UTF-8 string");
+            }
+
             CFunctionPointer fnPtr = entry.fnPtr();
 
             String declaringClass = MetaUtil.toInternalName(clazz.getName());
@@ -1243,7 +1255,15 @@ public final class JNIFunctions {
             DynamicHub.fromClass(clazz).ensureInitialized();
 
             CharSequence name = Utf8.wrapUtf8CString(cname);
+            if (name == null) {
+                throw new NoSuchMethodError("Method name is either null or invalid UTF-8 string");
+            }
+
             CharSequence signature = Utf8.wrapUtf8CString(csig);
+            if (signature == null) {
+                throw new NoSuchMethodError("Method signature is either null or invalid UTF-8 string");
+            }
+
             return getMethodID(clazz, name, signature, isStatic);
         }
 
@@ -1269,6 +1289,10 @@ public final class JNIFunctions {
             DynamicHub.fromClass(clazz).ensureInitialized();
 
             CharSequence name = Utf8.wrapUtf8CString(cname);
+            if (name == null) {
+                throw new NoSuchFieldError("Field name is either null or invalid UTF-8 string");
+            }
+
             JNIFieldId fieldID = JNIReflectionDictionary.singleton().getFieldID(clazz, name, isStatic);
             if (fieldID.isNull()) {
                 throw new NoSuchFieldError(clazz.getName() + '.' + name);


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
```diff
@@ -425,13 +445,21 @@ static JNIMethodId GetStaticMethodID(JNIEnvironment env, JNIObjectHandle hclazz,
     @CEntryPoint(exceptionHandler = JNIExceptionHandlerReturnNullWord.class, include = CEntryPoint.NotIncludedAutomatically.class, publishAs = Publish.NotPublished)
     @CEntryPointOptions(prologue = JNIEnvEnterPrologue.class, prologueBailout = ReturnNullPointer.class)
     static JNIFieldId GetFieldID(JNIEnvironment env, JNIObjectHandle hclazz, CCharPointer cname, CCharPointer csig) {
-        return Support.getFieldID(hclazz, cname, csig, false);
+        final JNIFieldId result = Support.getFieldID(hclazz, cname, csig, false);
+        if (result.isNull()) {
+            throw new NoSuchFieldError("<INVALID-NAME>");
+        }
```
peter-hofer, line 451: `Support.getFieldID` already throws when it would return a null handle.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
```diff
@@ -425,13 +445,21 @@ static JNIMethodId GetStaticMethodID(JNIEnvironment env, JNIObjectHandle hclazz,
...
     }
 
     @CEntryPoint(exceptionHandler = JNIExceptionHandlerReturnNullWord.class, include = CEntryPoint.NotIncludedAutomatically.class, publishAs = Publish.NotPublished)
     @CEntryPointOptions(prologue = JNIEnvEnterPrologue.class, prologueBailout = ReturnNullPointer.class)
     static JNIFieldId GetStaticFieldID(JNIEnvironment env, JNIObjectHandle hclazz, CCharPointer cname, CCharPointer csig) {
-        return Support.getFieldID(hclazz, cname, csig, true);
+        final JNIFieldId result = Support.getFieldID(hclazz, cname, csig, true);
+        if (result.isNull()) {
+            throw new NoSuchFieldError("<INVALID-NAME>");
+        }
```
peter-hofer, line 461: `Support.getFieldID` already throws when it would return a null handle.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
```diff
@@ -872,6 +900,9 @@ interface NewObjectWithObjectArrayArgFunctionPointer extends CFunctionPointer {
     static int ThrowNew(JNIEnvironment env, JNIObjectHandle clazzHandle, CCharPointer message) throws Throwable {
         Class<?> clazz = JNIObjectHandles.getObject(clazzHandle);
         JNIMethodId ctor = Support.getMethodID(clazz, "<init>", "(Ljava/lang/String;)V", false);
+        if (ctor.isNull()) {
+            return JNIErrors.JNI_EINVAL();
+        }
```
peter-hofer, line 905: `Support.getMethodID` already throws when it cannot find the constructor, and so does the HotSpot implementation of `ThrowNew`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/util/Utf8.java
```diff
@@ -226,12 +225,20 @@ public static final class WrappedAsciiCString implements CharSequence {
...
             this.chars = chars;
             this.length = length;
             this.hashCode = hashCode;
+        }
+
+        private static WrappedAsciiCString create(CCharPointer chars, int length, int hashCode) {
+            final String s = utf8ToString(chars);
+            if (s == null || s.length() != length || s.hashCode() != hashCode) {
+                return null;
+            }
```
peter-hofer, line 238: This always transforms the characters to a `java.lang.String` instead of doing so only as an assertion, losing the advantage of `WrappedAsciiCString` entirely. What does this accomplish anyway? If the string contains non-ASCII characters, the caller will instead invoke `utf8ToString` and add any additional checks there.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
```diff
@@ -341,6 +341,10 @@ static JNIObjectRefType GetObjectRefType(JNIEnvironment env, JNIObjectHandle han
     @CEntryPointOptions(prologue = JNIEnvEnterPrologue.class, prologueBailout = ReturnNullHandle.class)
     static JNIObjectHandle FindClass(JNIEnvironment env, CCharPointer cname) {
         CharSequence name = Utf8.wrapUtf8CString(cname);
+        if (name == null) {
+            throw new NoClassDefFoundError("<INVALID-NAME>");
```
peter-hofer, line 345: Please change to `Class name is not a valid UTF-8 string or NULL`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
```diff
@@ -363,7 +367,15 @@ static int RegisterNatives(JNIEnvironment env, JNIObjectHandle hclazz, JNINative
         for (int i = 0; i < nmethods; i++) {
             JNINativeMethod entry = (JNINativeMethod) p;
             CharSequence name = Utf8.wrapUtf8CString(entry.name());
+            if (name == null) {
+                throw new NoSuchMethodError("<INVALID-NAME>");
```
peter-hofer, line 371: Please change to: `"Method name at index " + i + " is not a valid UTF-8 string or NULL"`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
```diff
@@ -363,7 +367,15 @@ static int RegisterNatives(JNIEnvironment env, JNIObjectHandle hclazz, JNINative
...
+            }
+
             CharSequence signature = Utf8.wrapUtf8CString(entry.signature());
+            if (signature == null) {
+                throw new NoSuchMethodError("<INVALID-SIGNATURE>");
```
peter-hofer, line 376: Please change to: `"Signature at index " + i + " is not a valid UTF-8 string or NULL"`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
```diff
@@ -1244,7 +1275,9 @@ static JNIMethodId getMethodID(JNIObjectHandle hclazz, CCharPointer cname, CChar
 
             CharSequence name = Utf8.wrapUtf8CString(cname);
             CharSequence signature = Utf8.wrapUtf8CString(csig);
-            return getMethodID(clazz, name, signature, isStatic);
+            return (name == null || signature == null)
+                            ? WordFactory.nullPointer()
+                            : getMethodID(clazz, name, signature, isStatic);
```
peter-hofer, line 1280: Please simply throw `NoSuchMethodError` here.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
```diff
@@ -1269,6 +1302,10 @@ static JNIFieldId getFieldID(JNIObjectHandle hclazz, CCharPointer cname, CCharPo
             DynamicHub.fromClass(clazz).ensureInitialized();
 
             CharSequence name = Utf8.wrapUtf8CString(cname);
+            if (name == null) {
+                return WordFactory.nullPointer();
+            }
+
```
peter-hofer, line 1308: Please simply throw `NoSuchFieldError` here.

peter-hofer (changes requested):

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
koutheir: That's true, but that's not what I understand from [the JNI specifications](https://docs.oracle.com/en/java/javase/11/docs/specs/jni/functions.html#thrownew). `ThrowNew()` is supposed to throw the "newly constructed `java.lang.Throwable` object", and return a negative value of failure to do that (in this case, `JNI_EINVAL`).

koutheir:
The CI failure is unrelated to these changes.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/functions/JNIFunctions.java
peter-hofer: This is a generic phrase in the JNI specification. In HotSpot's implementation which we consider the reference for implementation details, I can't see a code path in which it wouldn't return `JNI_OK`.

koutheir: I don't like that, but okay.

koutheir:
The CI failure is unrelated to these changes.

PR #5992, opened 2023-02-14
Proper input validation in replaced implementation of `Array.newInstance()`

koutheir:
The JDK implementation of `Array.newInstance()` simply calls the private `Array.newArray()`.

The previous implementation replacement only checked for `null` `componentType` and strictly negative `length`, but did not check that the dimension of `componentType` (when it is an array class) is less than 255.

The previous incomplete checks were removed, and moved into the replacement implementation of `Array.newInstance()` in `Target_java_lang_reflect_Array`.
That implementation performs the complete set of checks before calling `Array.newArray()`.

### DIFF
--- a/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
+++ b/compiler/src/org.graalvm.compiler.replacements/src/org/graalvm/compiler/replacements/StandardGraphBuilderPlugins.java
@@ -468,12 +468,10 @@ public class StandardGraphBuilderPlugins {
 
     private static void registerArrayPlugins(InvocationPlugins plugins, Replacements replacements) {
         Registration r = new Registration(plugins, Array.class, replacements);
-        r.register(new InvocationPlugin("newInstance", Class.class, int.class) {
+        r.register(new InvocationPlugin("newArray", Class.class, int.class) {
             @Override
             public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver unused, ValueNode componentType, ValueNode length) {
-                ValueNode componentTypeNonNull = b.nullCheckedValue(componentType);
-                ValueNode lengthPositive = b.maybeEmitExplicitNegativeArraySizeCheck(length);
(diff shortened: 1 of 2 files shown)


koutheir:
I don't think the CI failure is related to this patch.

peter-hofer:
Thank you @koutheir , but this also affects Graal JIT compilation, so it's not valid to move checks to a substitution that is used only for native images. @mur47x111, @christianwimmer, could you please confirm the issues with the current `Array.newInstance` plugin?

mur47x111:
> this also affects Graal JIT compilation, so it's not valid to move checks to a substitution that is used only for native image

Correct. The JIT compilation demands these checks.

koutheir:
So are all checks supposed to happen in `StandardGraphBuilderPlugins.java`?

mur47x111:
yes unless you upstream these checks in JDK. (E.g., `CounterMode.crypt` does all the bound check and then calls `CounterMode.implCrypt` which is a compiler intrinsic.) Then we don't need these checks in `StandardGraphBuilderPlugins` nor `Target_java_lang_reflect_Array`. We only need to change the plugin to be applicable on `Array.newArray`

mur47x111:
I attempted to create a test case and cannot reproduce on JIT (both C2 and Graal). I think the reason is that HotSpot is consistent on the array instantiation code and will not cache the array class if its dimension exceeds 255. We will then hit https://github.com/oracle/graal/blob/04b38a35884ed20b75777ae795bdf723820fe5da/compiler/src/org.graalvm.compiler.hotspot/src/org/graalvm/compiler/hotspot/replacements/HotSpotAllocationSnippets.java#L245-L248 and deoptimize in the JITed code. 

I think we can make this plugin HotSpot only, and in `Target_java_lang_reflect_Array`, instead of creating an `@Alias`, we do a `DynamicNewArrayNode.newArray`.

koutheir:
> I think we can make this plugin HotSpot only, and in `Target_java_lang_reflect_Array`, instead of creating an `@Alias`, we do a `DynamicNewArrayNode.newArray`.

I'm not sure how to do this.

PR #6029, opened 2023-02-20
Fixes and features for LLVM backend

Zeavee:
This PR regroups fixes and new features for the LLVM backend.

- When emitting a shift, according to the documentation, we have to allow a number with a bit width of 8 or 16 and treat it as a 32 bit number.
- On a `Word.objectToTrackedPointer` call, we need to emit the `ptrToInt` LLVM call at the latest moment, to ensure the value used is correctly put in the statepoint.
- Handle `ComputeIndirectCallTargetNode` in the LLVM backend.
- Handle `ImageHeapConstant` in the LLVM backend.
- Make the `personality` function uninterruptible to avoid recurring callbacks when handling LLVM exceptions.
- Clobber the heap base register in functions that are entry points when isolates are enabled to ensure we return the correct isolate even when we create a new one in a JNI call for example.


christianhaeubl:
@loicottet : please have a look at this PR.

Zeavee:
Pull request is merged.

PR #6099, opened 2023-03-01
[DRAFT] Use a dense BitSet when TypeState cardinality is small

tekknolagi:
cc @cstancu


substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/typestate/MultiTypeState.java
```diff
@@ -185,11 +266,78 @@ public boolean equals(Object o) {
...
+            for (int i = other.nextSetBit(0); i >= 0; i = other.nextSetBit(i + 1)) {
+                set[cardinality++] = i;
+            }
+            assert cardinality == other.cardinality();
+        }
+
+        public boolean get(int bit) {
+            assert bit >= 0 : "get() bit must be non-negative";
+            for (int i = 0; i < cardinality; i++) {
+                if (set[i] == bit) {
```
cstancu, line 289: The type IDs should already be sorted. You could use `java.util.Arrays#binarySearch(int[], int)`, although for such a small size it probably won't make a lot of difference.

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/typestate/DefaultAnalysisPolicy.java
```diff
@@ -337,7 +337,7 @@ public TypeState doUnion(PointsToAnalysis bb, MultiTypeState s1, MultiTypeState
          * No need for a deep equality check (which would need to iterate the arrays), since the
          * speculation logic below is doing that anyway.
          */
-        if (s1.typesBitSet() == s2.typesBitSet()) {
+        if (s1.bitSetEquals(s2)) {
```
cstancu, line 340: Switching from `==` to `equals()`  invalidates the speculation.

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/typestate/MultiTypeState.java
tekknolagi: I figured linear search would be faster and potentially even loop unrolled if the compiler can prove cardinality <= MAX_CARDINALITY. I would have to actually investigate this

substratevm/src/com.oracle.graal.pointsto/src/com/oracle/graal/pointsto/typestate/DefaultAnalysisPolicy.java
tekknolagi: Ah, good point. I will make something that does the shallow check without materializing a `BitSet`.

PR #6139, opened 2023-03-07
[GR-42112] Missing registration errors for reflection calls

graalvmbot:


loicottet:
Introduces `MissingReflectionRegistrationError` when it is impossible to tell whether a reflective query should fail because the queried element has not been registered at build-time, or because it doesn't exist. The PR also adds support for `Class.getNestMembers` and `Class.getSigners`.
The changes are hidden behind the `-H:+ThrowMissingRegistrationErrors` flag for now, with the objective of enabling this flag by default in the future and eventually getting rid of it.

PR #6202, opened 2023-03-14
[GR-40463] RJMX Configuration Optimization.

roberttoyonaga:
This PR tries to optimize some of the configuration needed for remote JMX support. Another round of optimizations will be done at some point in the future.

Unfortunately, even after these changes, the native image doesn't get much smaller. But this might help increase maintainability at least. 

### Summary of Changes
Added some more tests in substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jmx/JmxTest.java
Removed BufferPoolMXBean test because it's not actually implemented (so the test always passes anyway).

Slimmed down reflection configuration for platform MXBeans in substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JmxServerFeature.java 
(this doesn't actually make the image much smaller unfortunately). Only the dynamic proxy configuration and reflection configuration for the classes is kept.

Removed some reflection registrations from JmxCommon. The "stub" and "skel" class reflection registrations could not be removed because they are needed for creating the remote stubs. Additionally all the methods for "Info" type classes that have a `from(CompositeData)` method require that all their methods be registered for reflections. The same is true for RMIConnection and RMIServer who's every method is accessed via RMIServerImpl  and RMIConnectionImpl_Stub.

None of the dynamic proxy registrations could be removed. The PlatformMXBeans can be accessed remotely via dynamic proxies that abstract away the connection handling with the remote MBean server (The other way is direct interaction through the MBeanServer). So any platform MBean that we want to be available needs to have a dynamic proxy registrations. VisualVM/JMC also need this in order to display PlatformMXBean info.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JmxCommonFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JmxCommonFeature.java
@@ -126,24 +126,22 @@ public class JmxCommonFeature implements InternalFeature {
 
     private static void configureProxy(BeforeAnalysisAccess access) {
         DynamicProxyRegistry dynamicProxySupport = ImageSingletons.lookup(DynamicProxyRegistry.class);
-
-        dynamicProxySupport.addProxyClass(access.findClassByName("jdk.management.jfr.FlightRecorderMXBean"),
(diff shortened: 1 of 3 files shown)


fniephaus:
Running the internal gates on this...

fniephaus (approved):
LGTM, thanks!

PR #6249, opened 2023-03-21
[doc](readme)Enhancement in url

nikam14:
Modified the url into a hyperlink.


fniephaus:
Thanks for the suggestion, but the link actually works because `[website]` references this: https://github.com/oracle/graal/pull/6249/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L102.

PR #6264, opened 2023-03-22
Ensure raw struct field getter identifies same field setter.

adinn:
This fixes a problem in class `SubstrateDiagnostics$ErrorContext` where it declares a raw structure field getter with name `frameHasCalleeSavedRegisters` and a corresponding setter with name `setFrameHasCalleeSavedRegisters`. The result is to treat them as independent accessors for distinct, foreign, single bit fields, one with name `frameHasCalleeSavedRegisters` and the other with name `FrameHasCalleeSavedRegisters`, with the result that accesses do not reflect prior updates.

The fix provided by this PR simply employs the expected name `getFrameHasCalleeSavedRegisters`. However, the PR also inserts a sanity check assertion to ensure that raw structure fields are always provided with both a getter and a setter.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateDiagnostics.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateDiagnostics.java
@@ -534,7 +534,7 @@ public class SubstrateDiagnostics {
                 log.string("General purpose register values:").indent(true);
                 RegisterDumper.singleton().dumpRegisters(log, context.getRegisterContext(), printLocationInfo, allowJavaHeapAccess, allowUnsafeOperations);
                 log.indent(false);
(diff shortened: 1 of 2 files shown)


christianhaeubl:
Thanks for the fix. The verification that you added is a bit too strict because we have `@RawStruct` fields that don't have a setter (e.g., `JfrBufferNode.Lock`). I will try to come up with an improved verification (this will need some internal discussion because `@RawStruct` is public API).

adinn:
Would it be best to push this change now minus the verification?

christianhaeubl:
I improved the verification and it now works on pretty much the whole code base (for the very small number of exceptions, I added unused getters to make the verification happy). This will be merged to master next week.

For GraalVM 23.0 and 22.3, I plan to merge the fix without the verification.

PR #6272, opened 2023-03-24
AMD64 intrinsics for `UnmanagedMemoryUtil.copyLongsForward/Backward`

petermz:
This PR adds AMD64 intrinsics for `UnmanagedMemoryUtil.copyLongs` methods. They are designed along the lines of `copy_bytes` Hotspot stubs. AVX registers are used when available; otherwise bytes are transferred via 64 bit registers in 32-byte chunks (this corresponds to what the current implementation is doing).

Because these methods are extensively used by GC, I've measured length of GC pause as a performance metric. The charts show that the intrinsics (orange line) consistently win about 3% over compiled code (blue line), even in the absence of AVX instructions.

avx128
avx256
avx512
noavx


oubidar-Abderrahim:
Hi, thank you for contributing to GraalVM, could you please provide the following info:

- Is the PR ready for review?
- Is there someone from the Graal team that you want to review this PR? please mention them

PR #6300, opened 2023-03-28
Ensure compiled methods are presented in ascending address order.

adinn:


adinn:
This PR has been closed in favour of #6414

PR #6302, opened 2023-03-28
[GR-44722] Adjust version strings and properties.

graalvmbot:


fniephaus:
I've added a description to the PR and am working with @jerboaa on fixing the Quarkus version detection. We also need to update the version detection in the Native Build Tools.

PR #6320, opened 2023-03-30
Thread user cpu time implementation for LinuxThreadCpuTimeSupport.getThreadCpuTime() method

AlexanderScherbatiy:
This is a proposal for a thread user cpu time implementaion in `LinuxThreadCpuTimeSupport.getThreadCpuTime()` method:  https://github.com/oracle/graal/blob/be7610ba0c935325098746b7431f1577661c89b3/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxThreadCpuTimeSupport.java#L63

The logic which retrieves thread user time is based on jdk [static jlong slow_thread_cpu_time(Thread *thread, bool user_sys_cpu_time)](https://github.com/openjdk/jdk/blob/9df20600592427550998c6685f103737e3115a51/src/hotspot/os/linux/os_linux.cpp#L5035) method.

- `PosixVMThreads.getCurrentOSThreadId()` is updated to return `gettid()` on Linux platform.
-  `ThreadCpuTimeSupport.getThreadCpuTime(OSThreadHandle, boolean)` method is changed to  `getThreadCpuTime(IsolateThread, boolean)`. This allows to use `IsolateThread` to get `OSThreadId` for thread user time and `OSThreadHandle` for thread system time.
- `/proc/self/task/tid/stat` file is used to get a user cpu time by kernel thread id using regular expressions.
- `ThreadCpuTimeTest` junit test is added,

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxThreadCpuTimeSupport.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxThreadCpuTimeSupport.java
@@ -75,7 +75,7 @@ final class LinuxThreadCpuTimeSupport implements ThreadCpuTimeSupport {
             return getThreadUserTime(kernelThreadId.get(isolateThread));
         }
 
-        return getThreadCpuTime(VMThreads.findOSThreadHandleForIsolateThread(isolateThread), includeSystemTime);
+        return getThreadCpuTime(VMThreads.findOSThreadHandleForIsolateThread(isolateThread));
     }
 
     /**
@@ -86,12 +86,8 @@ final class LinuxThreadCpuTimeSupport implements ThreadCpuTimeSupport {
      * @param includeSystemTime if {@code true} includes both system and user time, if {@code false}
      *            returns user time.
      */
-    @Override
     @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
-    public long getThreadCpuTime(OSThreadHandle osThreadHandle, boolean includeSystemTime) {
-        if (!includeSystemTime) {
-            return -1;
-        }
+    private long getThreadCpuTime(OSThreadHandle osThreadHandle) {
         CIntPointer threadsClockId = StackValue.get(Integer.BYTES);
         if (LinuxPthread.pthread_getcpuclockid((pthread_t) osThreadHandle, threadsClockId) != 0) {
             return -1;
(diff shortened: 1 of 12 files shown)


AlexanderScherbatiy:
`@SuppressWarnings("unused")` annotation is added for `isolateThread` parameter in `ThreadCpuTimeSupport.init(...)` method to fix `The value of the parameter isolateThread is not used` problem.

oubidar-Abderrahim:
Hi, thank you for contributing to GraalVM, could you please provide the following info:

- Is the PR ready for review?
- Is there someone from the Graal team that you want to review this PR? Please mention them

AlexanderScherbatiy:
- The PR is ready for review. The idea of the fix was to implement the `LinuxThreadCpuTimeSupport.getThreadCpuTime()` method in the similar way as it is done in jdk.
- No.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxThreadCpuTimeSupport.java
```diff
@@ -34,22 +35,49 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.BufferedReader;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 @AutomaticallyRegisteredImageSingleton(ThreadCpuTimeSupport.class)
 final class LinuxThreadCpuTimeSupport implements ThreadCpuTimeSupport {
 
+    private static final FastThreadLocalInt kernelThreadId = FastThreadLocalFactory.createInt("LinuxThreadCpuTimeSupport.kernelThreadId");
```
christianhaeubl, line 55: I don't think that we need a new thread local. It should be possible to change the existing code so that `VMThreads.OSThreadIdTL` contains the thread id on Linux and macOS (instead of `pthread_t`, which isn't really correct and redundant, see `OSThreadHandleTL`).

christianhaeubl (commented):
Thanks for the PR, I added a few comments.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxThreadCpuTimeSupport.java
AlexanderScherbatiy: The code is updated to return `gettid()` from `PosixVMThreads.getCurrentOSThreadId()` method on Linux platform. The thread local is removed from `LinuxThreadCpuTimeSupport` and its `getThreadCpuTime(isolateThread, includeSystemTime)` method now uses OSThreadId to get the thread user time.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxThreadCpuTimeSupport.java
```diff
@@ -79,4 +107,45 @@ private static long getThreadCpuTimeImpl(int clockId) {
         }
         return time.tv_sec() * TimeUtils.nanosPerSecond + time.tv_nsec();
     }
+
+    @Uninterruptible(reason = "Used as a transition between uninterruptible and interruptible code", calleeMustBe = false)
```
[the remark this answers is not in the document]
AlexanderScherbatiy: `getThreadUserTime(tid)` method is moved into C code.

christianhaeubl:
Thanks. I fixed a few issues and did further cleanups, see https://github.com/oracle/graal/pull/6591/files. This PR should get merged in the next few days.

Issue #6319, opened 2023-03-30
[native-image --version] No way to tell version in use when JDK isn't jlinked nor compiled to binary


jerboaa:
**Describe the issue**
After https://github.com/oracle/graal/pull/6302 there is no way for us (Mandrel) to tell which codeline is in use based on the `native-image --version` output. We use the bash launcher (deliberately) and don't jlink the base JDK that is in use (plain OpenJDK).

`java.vendor.version` is a protected property, which cannot be set on the command line:

```
$ java -Djava.vendor.version=foo -XshowSettings:properties --version 2>&1 | grep vendor
    java.specification.vendor = Oracle Corporation
    java.vendor = Red Hat, Inc.
    java.vendor.url = https://www.redhat.com/
    java.vendor.url.bug = https://bugzilla.redhat.com/enter_bug.cgi?product=Fedora&component=java-17-openjdk&version=37
    java.vendor.version = (Red_Hat-17.0.6.0.10-1.fc37)
    java.vm.specification.vendor = Oracle Corporation
    java.vm.vendor = Red Hat, Inc.
```

That means, the new `native-image --version` output solely contains JDK version properties, which is unfortunate since we cannot tell which code line is in use (23.x or 24.x or...).

Example (of a recent mandrel build from graal master):

```
$ native-image --version
native-image 20.0.1-internal 2023-04-18
OpenJDK Runtime Environment null (build 20.0.1-internal-adhoc.sgehwolf.jdk20u)
OpenJDK 64-Bit Server VM null (build 20.0.1-internal-adhoc.sgehwolf.jdk20u, mixed mode)
```

jerboaa:
WIP patch to solve this issue for us would be something like this:

https://github.com/oracle/graal/compare/master...jerboaa:graal:version-mandrel?expand=1

Would that be acceptable?

jerboaa:
@fniephaus Would you have any eta for something like this getting integrated? The version change produces a lot of noise in our CI.

fniephaus:
I will reach out tomorrow and hope we can merge a fix asap and within the next 48 hours.

fniephaus:
I've just opened #6356 with a more comprehensive fix for this (incl updating the build output infra). PTAL and let me know what you think. Thanks!

jerboaa:
Thank you!

PR #6321, opened 2023-03-30
Use java.vendor.version only if IS_AOT

jerboaa:
This allows for the bash launcher to print a sensible `native-image --version` output. Otherwise it would only contain OpenJDK properties, which gives no info on the actual code level in use.

Also add the vendor version to the build output as otherwise, the same issue applies. Only Java version information is present.

Closes: #6319


jerboaa:
@fniephaus PTAL. Thanks!

jerboaa:
See also https://github.com/graalvm/mandrel/issues/491#issuecomment-1490523562 for example version outputs.

zakkak (approved):
LGTM

jerboaa:
Replaced by #6356

PR #6327, opened 2023-03-31
Avoid NPE when build JDK has no source zip.

adinn:
This patch fixes a potential NPE looking up JDK sources when the build JDK does not include a src.zip file.


zakkak:
FTR this PR is superseded by https://github.com/oracle/graal/pull/6328

PR #6328, opened 2023-03-31
Avoid NPE when build JDK has no source zip.

adinn:
This patch avoids a potential NPE when the buildJDK does not include sources file `lib/src.zip`.

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceCache.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/sources/SourceCache.java
@@ -91,7 +91,7 @@ public class SourceCache {
             Path javaHomePath = Paths.get("", javaHome);
             Path srcZipPath = javaHomePath.resolve("lib").resolve("src.zip");
             if (!srcZipPath.toFile().exists()) {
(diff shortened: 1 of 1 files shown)


fniephaus:
Should we print some sort of warning that the JDK sources are missing? Otherwise, users may think debugging does not work for the JDK parts.

adinn:
> Should we print some sort of warning that the JDK sources are missing? Otherwise, users may think debugging does not work for the JDK parts.

Yes, that's probably a good idea. Can you point me at an example of how such a warning should be generated -- (unless, of course, it is merely a matter of writing to System.err :-)

fniephaus:
>Can you point me at an example of how such a warning should be generated -- (unless, of course, it is merely a matter of writing to System.err

We unfortunately still don't have an infrastructure for that, but in the meantime, please use `System.out` and start the warning with `Warning: <user-friendly description>`, similar to this:
https://github.com/oracle/graal/blob/84df4df8d6514a029354c5427d305ce87f9271a9/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/util/CPUTypeAMD64.java#L164-L165

adinn:
@fniephaus @olpaw I added a warning message and fixed the style issues. Should be good to go now.

adinn:
@fniephaus Sorry for the delay -- was busy on #6414. I have fixed the remaining style issue. This should be good to go, assuming the gate does not come up with any further style issues.

fniephaus (approved):
LGTM, thanks!

PR #6381, opened 2023-04-06
Fix reading cgroup controller error

ziyilin:
Multiple controllers can be set on the same hierarchy in /proc/self/cgroup. Assume a few fixed patterns is not sufficient to handle all the cases.


oubidar-Abderrahim:
Thank you @ziyilin for opening this PR,
@christianwimmer could you please assign someone to review this PR?

christianwimmer:
How is this handled / fixed in OpenJDK (where our code is copied from)?

ziyilin:
Although not exactly the same, OpenJDK17 uses the similar approach as this PR to handle the problem.
https://github.com/openjdk/jdk17u-dev/blob/master/src/java.base/linux/classes/jdk/internal/platform/CgroupSubsystemFactory.java#L259-L281

ziyilin:
Hi @christianwimmer, is there any review updates?

christianwimmer:
We are working on a solution that brings our code closer to the JDK again (ideally just using the JDK code).

VictoryWangCN:
Is there any update?

PR #6392, opened 2023-04-10
Print reports after standalone pointsto analysis

ziyilin:
Refactor the SubstrateOptions.Path option to a common place for both pointsto and SVM, so that the analysis report can be printed to the location specified by Path in standalone pointsto.

This PR is part of https://github.com/oracle/graal/pull/4375

### DIFF
--- a/substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/PointstoAnalyzerTester.java
+++ b/substratevm/src/com.oracle.graal.pointsto.standalone.test/src/com/oracle/graal/pointsto/standalone/test/PointstoAnalyzerTester.java
@@ -157,7 +157,11 @@ public class PointstoAnalyzerTester {
         try {
             try {
                 int ret = pointstoAnalyzer.run();
-                assertEquals("Analysis return code is expected to 0", 0, ret);
+                if (expectPass) {
+                    assertEquals("Analysis return code is expected to 0", 0, ret);
+                } else {
+                    assertNotEquals("The analysis is expected to fail, but succeeded", 0, ret);
+                }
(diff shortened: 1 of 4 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -274,9 +275,6 @@ public static void setOptimizeValueUpdateHandler(ValueUpdateHandler<Optimization
     @Option(help = "Path passed to the linker as the -rpath (list of comma-separated directories)")//
     public static final HostedOptionKey<LocatableMultiOptionValue.Strings> LinkerRPath = new HostedOptionKey<>(LocatableMultiOptionValue.Strings.buildWithCommaDelimiter());
 
-    @Option(help = "Directory of the image file to be generated", type = OptionType.User)//
-    public static final HostedOptionKey<String> Path = new HostedOptionKey<>(null);
```
cstancu, line 278: `Path` has  a specific meaning in the context of NI and I don't think that moving it to `CommonOptions` is a good idea. I think it's better if you add a specific options (e.g., `ReportsPath`) somewhere in the standalone project and use that instead.

ziyilin: Fixed. Move the `ReportsPath` option into `StandaloneOptions` as it is only used by standalone pointsto, it's not necessary to put it in the common place.

cstancu (approved):

ziyilin:
Hi @oubidar-Abderrahim , can you help to merge this PR? It has been a while. Thank you very much.

zakkak:

PR #6422, opened 2023-04-14
[GR-46055] Add Javadocs and reflection registrations for `ServerProvider` and `ClientProvider` in JMX Features

roberttoyonaga:
This adds reflection configuration for `com.sun.jmx.remote.protocol.rmi.ClientProvider` and `com.sun.jmx.remote.protocol.rmi.ServerProvider`.

When the JMX mbean server is set up to  accept remote connections, [`JMXConnectorServerFactory.newJMXConnectorServer`](https://github.com/openjdk/jdk/blob/master/src/java.management/share/classes/javax/management/remote/JMXConnectorServerFactory.java#L275) is called, which eventually can result in a call to `Class.forName()` with `com.sun.jmx.remote.protocol.rmi.ServerProvider`.

A similar path starting from [`JMXConnectorFactory.newJMXConnector`](https://github.com/openjdk/jdk/blob/master/src/java.management/share/classes/javax/management/remote/JMXConnectorFactory.java#L327) can lead to requiring reflection metadata for `com.sun.jmx.remote.protocol.rmi.ClientProvider`.

Edit: Added reflection registration for methods of `java.lang.management.MemoryNotificationInfo` and `java.lang.management.LockInfo`. These are necessary for the same reason `java.lang.management.ThreadInfo,com.sun.management.GcInfo`,` java.lang.management.MonitorInfo`, `java.lang.management.MemoryUsage` are necessary (these already exist in the reflection configuration).  “Info” classes have their methods reflectively accessed from their static “from” methods.  Remote JMX infrastructure uses the following pattern: the "info" classes have a corresponding "CompositeData" class which is used to construct them [using the static "from" method](https://github.com/openjdk/jdk/blob/master/src/java.management/share/classes/sun/management/MappedMXBeanType.java#L622). (ie. `SomeInfo` → `SomeInfoCompositeData` extends `LazyCompositeData`).  I have not yet seen runtime errors related to these specific classes requiring  reflection registration (and the tracing agent has not suggested them), but based on the pattern JMX uses, it would be conservative to include them. Adding/omitting these registrations does not seem to make a difference in image size anyway.  

Edit: Added some java docs to try to document why certain classes need to be registered.

These additional reflection registrations should be backported to the GraalVM 23 release branch.

### DIFF
--- a/substratevm/mx.substratevm/suite.py
+++ b/substratevm/mx.substratevm/suite.py
@@ -620,6 +620,13 @@ suite = {
                     "sun.util.locale",
                     "sun.invoke.util",
                 ],
+                "java.management": [
+                    "com.sun.jmx.mbeanserver",
+                    "sun.management",
+                ],
+                "java.rmi": [
+                    "sun.rmi.server",
+                ],
                 "jdk.internal.vm.ci" : [
                     "jdk.vm.ci.meta",
                     "jdk.vm.ci.code",
@@ -633,6 +640,11 @@ suite = {
                 "jdk.jfr": [
                     "jdk.jfr.internal",
                     "jdk.jfr.internal.jfc",
+                ],
+                "java.management.rmi": [
+                    "com.sun.jmx.remote.internal.rmi",
+                    "com.sun.jmx.remote.protocol.rmi",
+                    "javax.management.remote.rmi",
                 ],
             },
             "javaCompliance" : "17+",
--- a/substratevm/mx.substratevm/suite.py
+++ b/substratevm/mx.substratevm/suite.py
@@ -621,11 +621,11 @@ suite = {
                     "sun.invoke.util",
                 ],
                 "java.management": [
-                    "com.sun.jmx.mbeanserver",
-                    "sun.management",
+                    "com.sun.jmx.mbeanserver", # Needed for javadoc links (MXBeanIntrospector,DefaultMXBeanMappingFactory, MXBeanProxy)
+                    "sun.management", # Needed for javadoc links (MappedMXBeanType)
                 ],
                 "java.rmi": [
-                    "sun.rmi.server",
+                    "sun.rmi.server",  # Needed for javadoc links (UnicastRef, UnicastRef2)
                 ],
                 "jdk.internal.vm.ci" : [
                     "jdk.vm.ci.meta",
@@ -642,9 +642,9 @@ suite = {
                     "jdk.jfr.internal.jfc",
                 ],
                 "java.management.rmi": [
-                    "com.sun.jmx.remote.internal.rmi",
-                    "com.sun.jmx.remote.protocol.rmi",
-                    "javax.management.remote.rmi",
+                    "com.sun.jmx.remote.internal.rmi", # Needed for javadoc links (ProxyRef)
+                    "com.sun.jmx.remote.protocol.rmi", # Needed for javadoc links (ClientProvider, ServerProvider)
+                    "javax.management.remote.rmi", # Needed for javadoc links (RMIServer, RMIServerImpl_Stub, RMIConnection, RMIConnectionImpl_Stub)
                 ],
             },
             "javaCompliance" : "17+",
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JmxCommonFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JmxCommonFeature.java
@@ -26,6 +26,8 @@
 
 package com.oracle.svm.hosted.jdk;
 
+import java.lang.reflect.Method;
+import java.rmi.Remote;
 import java.util.Arrays;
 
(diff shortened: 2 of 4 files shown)


roberttoyonaga:
Hi @fniephaus can you have a look at this when you have time?

fniephaus:
Thanks for the PR, @roberttoyonaga. Are you looking in move issues that require similar config changes? Is there any way we can add a test for this to make sure we don't regress?

roberttoyonaga:
Hi @fniephaus I'm not sure about the best way to create a test for this. I did some debugging and found that the existing JMX native unit tests do the same reflective lookup on those classes, but I was only able to see this issue when using a Quarkus native app. Is it possible the [automatic reflection detection ](https://www.graalvm.org/22.0/reference-manual/native-image/Reflection/#automatic-detection) hid this bug up until now, but misses those reflection uses when building the Quarkus app? 

Yes I am trying to catalog reflection uses in JDK remote JMX code to see if there's anything that could yet break.

fniephaus:
>I'm not sure about the best way to create a test for this

That's alright, I was hoping we can have a simple end-to-end test. Maybe in the future, we could have a scriptable client, similar to how we test our DWARF debugging infra.

>Is it possible the [automatic reflection detection](https://www.graalvm.org/22.0/reference-manual/native-image/Reflection/#automatic-detection) hid this bug up until now, but misses those reflection uses when building the Quarkus app?

Why would the automatic detection of reflection work any different when building a Quarkus app? Maybe due to some side-effects, but I can't really think of any at this point.

roberttoyonaga:
@fniephaus Can you please have another look at this? I've also added/expanded some javadocs to try and make maintenance of these features easier.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JmxServerFeature.java
```diff
@@ -108,5 +119,8 @@ private static void configureReflection() {
             Class<?> clazz = p.getClass();
             RuntimeReflection.register(clazz);
         }
+
+        RuntimeReflection.register(access.findClassByName("com.sun.jmx.remote.protocol.rmi.ServerProvider"));
```
fniephaus, line 123: I'd be nice if we could add a reason for each registration, similar to how `RuntimeClassInitializationSupport#register()` accepts a reason (see for example `JDKInitializationFeature`). But I don't think we have anything for this yet, right @cstancu?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JmxCommonFeature.java
```diff
@@ -121,6 +123,32 @@ public void beforeAnalysis(BeforeAnalysisAccess access) {
         configureProxy(access);
     }
 
+    /**
+     * This method handles proxy registrations for PlatformMXBeans. We are able to do the
+     * registrations for PlatformMXBeans because they are known and there are a finite number of
+     * them. If a user wishes to register a custom standard MBean with the MBeanServer, they will
+     * have to provide their own proxy configuration in a JSON file. This is documented in the
```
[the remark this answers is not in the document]
roberttoyonaga: That's a good idea. I'll add a link here. I think when I was making these changes the web page was not yet live.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JmxCommonFeature.java
```diff
@@ -121,6 +123,32 @@ public void beforeAnalysis(BeforeAnalysisAccess access) {
...
+     * remote JMX guide.
+     *
+     * <P>
```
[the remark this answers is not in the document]
roberttoyonaga: No reason. I'll change it to be lowercase.

fniephaus:
Our internal gates are passing, so I guess this is good to go as soon as you've addressed the comments. :)

fniephaus (approved):
LGTM, thanks!

substratevm/mx.substratevm/suite.py
```diff
@@ -642,6 +642,13 @@
                     "sun.util.locale",
                     "sun.invoke.util",
                 ],
+                "java.management": [
+                    "com.sun.jmx.mbeanserver",
```
fniephaus, line 646: Do you only add those for the references in javadocs?

fniephaus: Well the references help navigation, so I'd be for keeping them. It's just a bit unfortunate that we loose the reason in the `suite.py`. In there, it looks like a strong code dependency.

roberttoyonaga: I could add comments in `suite.py` explaining those imports are only needed for javadocs. Would this be better?

fniephaus: Yes, that would be good. Thanks!

Issue #6142, opened 2023-03-07
Debugging Information for @RawStructure and @CStruct.


christianhaeubl:
Some parts of Native Image (JFR support, GC, ...) define C-like structures via the annotations `@RawStructure` and `@CStruct`. At the moment, it seems that no debugging information is getting emitted for structs like that. Just in case that you want to add this support, `NativeLibraries.findElementInfo(...)` would be the starting point to query information about those structs. /cc @olpaw

adinn:
Hi @christianhaeubl. Yes, it seems like it would be of use to be able to recognise these types.

I'll look into the information that `NativeLibraries.findElementInfo(...)` returns and come back with an indication of what info we might be able to encode for these types, how we might need modify the debug info generator to do so and how much work it would entail.

adinn:
@christianhaeubl @olpaw 

I looked into this and I think it may well be possible to use the element info data to provide richer debug info for types derived from `WordBase` and `PointerBase`.  I have some initial ideas for how to do this. However, before doing that I need to clarify some details.

1. The element info appears only to be present for types that subclass `WordBase` (which, therefore, includes those that subclass `PointerBase`). Is that always going to be the case? Or are there classes with element info that do not subclass `WordBase`?

2. I looked through the list of classes returned by `heap.getUniverse().getTypes()` to see which ones were subclasses of WordBase. I also checked which of them have element info. The summary counts are shown below.
  Clearly, a lot of them do not model foreign struct types. Is there any default interpretation that can be used to describe the classes which have no element info? For example, are they all really just the equivalent to a typedef for a machine word?

|#Subclasses |#With Element Info |#No Element Info|
|        ---:|               ---:|            ---:|
|         352|               174 |             178|

3. WordBase and its subclasses are currently presented as interfaces which means they are type compatible with (assignable to) type Object. That's obviously inappropriate for struct types since it implies they have a Java object header embedded at offset 0. Is there a good reason why they are typed using interfaces?

4. I am currently thinking that WordBase types with element info need to be described as pointers to foreign C struct/class types whereas those that have no element info need to be described as typedefs for a machine word or, when they inherit from PointerBase, a machine pointer. Is that the right model? In the latter case is there anything that would indicate signedness for the non-pointer word types?

christianhaeubl:
Thanks for looking into that!

ad 1. `WordBase` is always the top-level type.

ad 2. The default interpretation is really just the equivalent to a typedef for a machine word. The typical uses are:
[...]

PR #6625, opened 2023-05-17
[GR-46599] Foreign debuginfo types

adinn:
This PR provides a fix for issue #6142.

It involves 4 sets of changes:

1) Modification of the DebugInfoProvider API to separate out and present, via new API `DebugForeignTypeInfo`, information about Java interfaces (and a small number of classes) derived from `WordBase` and `PointerBase` whose purpose is to model foreign data (or, in the case of classes, foreign, raw pointers to Java data).

2. Modification of `NativeImageDebugInfoProvider` to identify the relevant types and advertise them via the new API.

3. Modification of the generic debug info model in package `com.oracle.objectfile.debugentry` to accomodate foreign type info.

4. Modification of the Linux (and Windows) info generator to output a DWARF (or PECOFF) encoding of the foreign types that is intelligible to the relevant debugger.


adinn:
### Current Progress

The current PR implements all of steps 1 2 and 3.

Step 4 is complete for Linux.

Step 4 for Windows still requires the PECOFF generator to be updated to generate an appropriate encoding for foreign types.

n.b. as a minimum the PECOFF generator could simply continue to treat these types as though they were Java classes or interfaces. However, it would be better if it were able to use an equivalent encoding to the DWARF one.

### Examples

```
(gdb) info types org.graalvm.nativeimage.c.type
All types matching regular expression "org.graalvm.nativeimage.c.type":

File JAVA:
	org.graalvm.nativeimage.c.type.CTypeConversion$CCharPointerHolder;
	typedef int8_t * org.graalvm.nativeimage.c.type.CCharPointer;
	typedef org.graalvm.nativeimage.c.type.CCharPointer * org.graalvm.nativeimage.c.type.CCharPointerPointer;
	typedef double * org.graalvm.nativeimage.c.type.CDoublePointer;
	typedef float * org.graalvm.nativeimage.c.type.CFloatPointer;
	typedef int32_t * org.graalvm.nativeimage.c.type.CIntPointer;
	typedef org.graalvm.nativeimage.c.type.CIntPointer * org.graalvm.nativeimage.c.type.CIntPointerPointer;
	typedef int64_t * org.graalvm.nativeimage.c.type.CLongPointer;
	typedef int16_t * org.graalvm.nativeimage.c.type.CShortPointer;
	typedef void * org.graalvm.nativeimage.c.type.VoidPointer;
	typedef org.graalvm.word.WordBase * org.graalvm.nativeimage.c.type.WordPointer;

File org/graalvm/nativeimage/c/type/CTypeConversion.java:
	org.graalvm.nativeimage.c.type.CTypeConversion;
	org.graalvm.nativeimage.c.type.CTypeConversion$CCharPointerPointerHolder;
(gdb) ptype 'org.graalvm.nativeimage.c.type.CIntPointer'
type = int32_t *
(gdb) info types IsolateThread
All types matching regular expression "IsolateThread":

File JAVA:
	typedef org.graalvm.nativeimage.IsolateThread * com.oracle.svm.core.c.function.CEntryPointNativeFunctions$IsolateThreadPointer;
	typedef void * org.graalvm.nativeimage.IsolateThread;
(gdb) ptype 'com.oracle.svm.core.jni.headers.JNIJavaVM'
type = struct JavaVM_ {
    com.oracle.svm.core.jni.headers.JNIInvokeInterface functions;
} *
(gdb) ptype 'com.oracle.svm.core.jni.headers.JNIInvokeInterface'
type = struct JNIInvokeInterface_ {
    org.graalvm.nativeimage.Isolate reserved0;
    org.graalvm.nativeimage.c.function.CFunctionPointer DestroyJavaVM;
    org.graalvm.nativeimage.c.function.CFunctionPointer AttachCurrentThread;
    org.graalvm.nativeimage.c.function.CFunctionPointer DetachCurrentThread;
    com.oracle.svm.core.jni.headers.JNIFunctionPointerTypes$GetEnvFunctionPointer GetEnv;
    org.graalvm.nativeimage.c.function.CFunctionPointer AttachCurrentThreadAsDaemon;
} *
```



### Notable Changes

Foreign types appear as interfaces in the Java type system i.e. as object references. However, they manifest (for the most part) as raw pointers at runtime, whether that be to pointers to foreign primitive values, structs, other pointers or simply undifferentiated (except, perhaps, as signed or unsigned) foreign words.

The DWARF encoding uses the Java type name as a typdef to label the relevant pointer types rather than the data referenced by the pointer. So, for example, the foreign types used to model foreign pointers of type `char *` and a `char **` appear in gdb as follows

```
(gdb) info types CCharPointer
All types matching regular expression "CCharPointer":

File JAVA:
	org.graalvm.nativeimage.c.type.CTypeConversion$CCharPointerHolder;
	typedef int8_t * org.graalvm.nativeimage.c.type.CCharPointer;
	typedef org.graalvm.nativeimage.c.type.CCharPointer * org.graalvm.nativeimage.c.type.CCharPointerPointer;
    . . .
(gdb) ptype 'org.graalvm.nativeimage.c.type.CCharPointer'
type = int8_t *
(gdb) ptype 'org.graalvm.nativeimage.c.type.CCharPointerPointer'
type = int8_t **
```
n.b. As an aside, appropriate at this point, I'll mention that I have deliberately chosen to use 'int8_t' as the name for the target type rather than `char` because of the obvious name clash with the Java type. Also, it is int8_t rather than uint8_t because the element info that defines the target type as a 1 byte integer does not mark it as unsigned.

This naming convention differs from the approach used with proper Java types where the Java name is used to label the underlying struct. It makes sense because a `CCharPointer` is clearly meant to be a pointer to a `char`. However, this also implies that in method or function signatures a foreign type parameter  or return valueappears without the dereference operator `*` e.g.

```
(gdb) info func dumpToLog
All functions matching regular expression "dumpToLog":

File com/oracle/svm/core/threadlocal/VMThreadLocalInfos.java:
63:	void com.oracle.svm.core.threadlocal.VMThreadLocalInfos::dumpToLog(com.oracle.svm.core.log.Log*, org.graalvm.nativeimage.IsolateThread, bool);
```

Note that this different naming convention has required a corresponding modification to the mangled name encoding employed when generating local or global symbols for the foreign types.

adinn:
@olpaw @christianhaeubl You might want to have a preliminary look at this and comment on the choices I have made regarding the DWARF encoding for foreign types and how that appears when viewed via gdb.

@christianhaeubl It would be very helpful if you could direct me at a simple example program that employs foreign data (whether explicitly or implicitly via, say, the GC) that I can use to verify the corrcetness of the encoding and its clarity/utility.

@stooke I would welcome your input on what might be done to ensure the Windows debug info is complete now that some of the Java types appear in the generic debug model via `ForeignTypeEntry` rather than `InterfaceEntry` or `ClassEntry`.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/TypeEntry.java
```diff
@@ -98,8 +99,25 @@ public boolean isEnum() {
...
+     * properties of the java type. This also allows them to be decorated with properties that
+     * record details of the generated debug info. When it comes to encoding the model type as DWARF
+     * or PECOFF method {@link #isForeign()} may need to be called in order to allow foreign types
+     * ot be special cased.
+     *
+     * @return true if this entry is a class type otherwise false.
+     */
     public boolean isClass() {
-        return isInstance() | isInterface() || isEnum();
+        return isInstance() | isInterface() || isEnum() || isForeign();
```
stooke, line 120: inadvertent bitwise OR found!

return isInstance() || isInterface() || isEnum() || isForeign();

stooke:
@adinn I've had  a chance to (quickly) review your changes as they impact the Windows PE/COFF code.  Your fix to CVTypeSectionBuilder.java should prevent the code from dying a death the first time a foreign type is encountered.  
It's difficult for me to see in the abstract what the issues will be, but optimistically it should only take a few lines in getIndexForPointerOrPrimitive().  If I can, I'll try to build from your PR before I head off on PTO and see how far I get.

As the Windows code maps java primitive types to C types, unfortunately I don't think I can have the debugger differentiate between java char and c++ uint8_t.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/TypeEntry.java
adinn: Oops ... good catch!
Fixed.

adinn:
> @adinn I've had a chance to (quickly) review your changes as they impact the Windows PE/COFF code. Your fix to CVTypeSectionBuilder.java should prevent the code from dying a death the first time a foreign type is encountered.
> It's difficult for me to see in the abstract what the issues will be, but optimistically it should only take a few lines in getIndexForPointerOrPrimitive(). If I can, I'll try to build from your PR before I head off on PTO and see how far I get.

Ok, fingers crossed. Let me know if there are problems that need I need to address in the stages that feed the PECOFF generator.

> As the Windows code maps java primitive types to C types, unfortunately I don't think I can have the debugger differentiate between java char and c++ uint8_t.

Contrariwise, if that is already a lost cause then introducing foreign type info the PECOFF encoding cannot compound the damage :-)

zakkak (commented):
The changes look good to me.

> @christianhaeubl It would be very helpful if you could direct me at a simple example program that employs foreign data (whether explicitly or implicitly via, say, the GC) that I can use to verify the corrcetness of the encoding and its clarity/utility.

+1

> I have deliberately chosen to use 'int8_t' as the name for the target type rather than char because of the obvious name clash with the Java type. 

Is `char8_t` (available since C++20) too new to be used instead of `int8_t`?

adinn:
> Is char8_t (available since C++20) too new to be used instead of int8_t?

No, I don't think it is too new but it may be unadvisable.

At the moment all the generator knows is that some foreign POINTER types point to integral values with known byte size and signedness. So, I generate names accordingly like uint8_t, int32_t, etc. Much as I could swap char for int when the bit count is 8, I think I prefer the above uniformity to consistency with the slightly weird norms of C nomenclature.

The original distinction between char for single byte integral values vs various, possibly short or long, flavours of int for multi byte integral values is a somewhat unfortunate artefact of C pre-history. It has eventually landed us with the nonsense of char8_t and int8_t both having the same sense in terms of the underlying integral representation yet being distinguished as a sop to different intended usage (we won't even consider the merits of long64_t and int64_t). That distinction also foreshadows the need for char16_t (wide chars) vs, say, short in or int16_t. I think I'd rather avoid the whole can of worms.

christianhaeubl:
@adinn: here are a few examples for `@RawStructure`s and how you could inspect them with gdb:

- `com.oracle.svm.core.genscavenge.AlignedHeapChunk.AlignedHeader` (signed/unsigned data, inheritence, pinned object data): run any non-trivial application (it only needs to perform a GC) and set a breakpoint in `YoungGeneration.promoteAlignedObject(...)`

- `com.oracle.svm.core.jfr.JfrBuffer` (primitive and native pointers): build and run a JFR test case and set a breakpoint at the end of `JfrThreadLocal.getNativeBuffer(...)`. Command line for building and executing: `graal/substratevm$ mx native-unittest -p TestJavaMonitorWaitEvent --build-args -g -O0`

- `com.oracle.svm.core.heap.dump.HeapDumpMetadata.ClassInfo` (pointers to other `@RawStructure`s, pinned object data): build any native image with `--enable-monitoring=heapdump`. Run the image with `-XX:+DumpHeapAndExit` and set a breakpoint in `HeapDumpMetadata.initialize()`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -911,19 +1100,287 @@ public int flags() {
...
+            debugContext.log("typedef %s = *%s", typedefName, hostedType.toJavaName());
+        }
+        dumpElementInfo(pointerToInfo);
+    }
+
+    private void logStructInfo(HostedType hostedType, StructInfo structInfo) {
+        debugContext.log(DebugContext.VERBOSE_LEVEL, "Foreign struct type %s %s", hostedType.toJavaName(), elementKind(structInfo).toString());
+        assert hostedType.isInterface();
+        boolean isIncomplete = structInfo.isIncomplete();
+        Integer size;
```
christianhaeubl, line 1177: I don't see any reason why you need a boxed integer here (`elementSize(...)` returns a primitive int value). Same applies to `Integer size` in `logRawStructureInfo(...)`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1146,24 +1603,46 @@ private List<DebugLocalInfo> createParamInfo(ResolvedJavaMethod method, int line
...
-     * 
+     * Identify a Java type which is being used to model a foreign memory word or pointer type.
+     *
      * @param type the type to be tested
      * @param accessingType another type relative to which the first type may need to be resolved
-     * @return true if the type is a pseudo object type
+     * @return true if the type models a foreign memory word or pointer type
+     */
+    private boolean isForeignWordType(JavaType type, ResolvedJavaType accessingType) {
+        assert accessingType instanceof HostedType : "must be!";
```
christianhaeubl, line 1621: I tend to avoid assertions like that because they don't provide much benefit over the `ClassCastException` that would be thrown anyways.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -128,15 +142,16 @@ class NativeImageDebugInfoProvider implements DebugInfoProvider {
     int primitiveStartOffset;
     int referenceStartOffset;
     private final Set<HostedMethod> allOverrides;
-    HostedType wordBaseType;
     HostedType hubType;
+    private HostedType wordBaseType;
```
christianhaeubl, line 146: Can be final.

substratevm/src/com.oracle.svm.test/src/hello/#Hello.java#
```diff
@@ -0,0 +1,247 @@
...
+import com.oracle.svm.core.NeverInline;
+
+import org.graalvm.word.PointerBase;
+import org.graalvm.word.WordFactory;
+import org.graalvm.word.UnsignedWord;
+import org.graalvm.word.SignedWord;
+
+import org.graalvm.nativeimage.UnmanagedMemory;
+
+public class Hello {
```
christianhaeubl, line 41: Where is this class/file used?

christianhaeubl (commented):
Thanks for implementing this. This will definitely improve the debugging experience by a lot in some areas (e.g., JFR support).

I tested your implementation by debugging a few examples (mainly related to `@RawStructure`) and I did not encounter any issues when I used `gdb` directly. When I briefly tested with Eclipse CDT, I encountered a few problems (e.g., if a struct is in a local variable, it sometimes fails to show fields of the struct) but I would assume that those are more likely to be related to Eclipse CDT.

I did a rough code review and added a few comments. @olpaw will need to do the detailed review because he knows this code much better than I do.

> The current code assumes at most one such 'parent' will be present. Should I add an assert that there is not more than one such parent?

Yes, please add an assertion.

substratevm/src/com.oracle.svm.test/src/hello/#Hello.java#
adinn: This is left over from an experiment. UnmanagedMemory is not used by Hello. Removed.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/DebugInfoBase.java
```diff
@@ -143,6 +143,10 @@ public abstract class DebugInfoBase {
      * Handle on type entry for header structure.
      */
     private HeaderTypeEntry headerType;
+    /**
+     * Handle on type entry for void typ-e.
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ForeignTypeEntry.java
```diff
@@ -0,0 +1,125 @@
...
+import jdk.vm.ci.meta.ResolvedJavaType;
+import org.graalvm.compiler.debug.DebugContext;
+
+public class ForeignTypeEntry extends ClassEntry {
+    private static final int FLAG_WORD = 1 << 0;
+    private static final int FLAG_POINTER = 1 << 1;
+    private static final int FLAG_INTEGRAL = 1 << 2;
+    private static final int FLAG_SIGNED = 1 << 3;
+    private static final int FLAG_FLOAT = 1 << 4;
+    String typedefName;
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
```diff
@@ -85,6 +85,22 @@ public void afterRegistration(AfterRegistrationAccess access) {
...
+    public void beforeAnalysis(BeforeAnalysisAccess access) {
+        /*
+         * Make the name provider aware of the native libs
+         */
+        if (!UniqueShortNameProviderDefaultImpl.UseDefault.useDefaultProvider()) {
+            if (ImageSingletons.contains(UniqueShortNameProvider.class)) {
+                UniqueShortNameProvider provider = ImageSingletons.lookup(UniqueShortNameProvider.class);
+                if (provider instanceof NativeImageBFDNameProvider) {
+                    var accessImpl = (FeatureImpl.BeforeAnalysisAccessImpl) access;
+                    ((NativeImageBFDNameProvider) provider).setNativeLibs(((FeatureImpl.BeforeAnalysisAccessImpl) access).getNativeLibraries());
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -848,6 +870,169 @@ private int writeInterfaceImplementor(DebugContext context, ClassEntry classEntr
...
+        int layoutOffset = pos;
+        if (foreignTypeEntry.isWord()) {
+            // define the type as a typedef for a signed or unsigned word i.e. we don't have a
+            // layout type
+            pos = writeForeignWordLayout(context, foreignTypeEntry, size, foreignTypeEntry.isSigned(), buffer, pos);
+        } else if (foreignTypeEntry.isIntegral()) {
+            // use a suitably sized signed or unsigned integral type as the layout type
+            pos = writeForeignIntegerLayout(context, foreignTypeEntry, size, foreignTypeEntry.isSigned(), buffer, pos);
+        } else if (foreignTypeEntry.isFloat()) {
+            // use a suitably sized signed or unsigned float type as the layout type
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -848,6 +870,169 @@ private int writeInterfaceImplementor(DebugContext context, ClassEntry classEntr
...
+            pos = writeForeignFloatLayout(context, foreignTypeEntry, size, buffer, pos);
+        } else {
+            // pointer or unknown - layout id as a foreign stucture if we have fields otherwise use
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -911,19 +1100,287 @@ public int flags() {
...
+            totalSize += nextSize;
+        }
+        if (totalSize != size) {
+            debugContext.log("    // int gap : %d", size - totalSize);
+        }
+        debugContext.log("  } %s;", rawStructureInfo.getName());
+    }
+
+    private HostedType getFieldType(StructFieldInfo field) {
+        // we should always some sort of accessor, preferably a GETTER or a SETTER
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -911,19 +1100,287 @@ public int flags() {
...
+        if (elementInfo instanceof RawStructureInfo || elementInfo instanceof StructInfo) {
+            return elementInfo.getChildren().stream().filter(elt -> isFieldWithGetterOrSetter(elt))
+                            .map(elt -> ((StructFieldInfo) elt))
+                            .sorted(this::structFieldComparator);
+        } else {
+            return Stream.empty();
+        }
+    }
+
+    boolean isFieldWithGetterOrSetter(ElementInfo elementInfo) {
```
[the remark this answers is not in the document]
adinn: The log code also needed serious pruning. It was mostly a heuristic so I could view a print rendering of the element info and work out how to model the types.
Done.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Sure, removed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -911,19 +1100,287 @@ public int flags() {
...
+            debugContext.log("    typedefName = %s", typedefName);
+        }
+        dumpElementInfo(structInfo);
+        // The target may be a struct or a typedef for a struct. In the former case
+        // the keyword struct will be part of the name (the typedefName for the
+        // elementInfo appears always to be null).
+        debugContext.log("typedef %s *%s;", structInfo.getName(), hostedType.toJavaName());
+        List<StructFieldInfo> orderedFields = orderedFieldsStream(structInfo).toList();
+        if (orderedFields.isEmpty()) {
+            debugContext.log("raw struct with no fields!");
```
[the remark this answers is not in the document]
adinn: No longer applicable.

adinn:
@christianhaeubl @olpaw

I spotted two more issues that I have resolved with the most recent push.

1. Foreign fields  in a struct which are accessed using an `Accessor` of type `ADDRESS` can correspond to embedded composite data, either another struct or, as another possibility, an array of some foreign type. Such fields were being omitted, resulting in gaps in the struct definition. I have fixed that by pushing field info through to the generator for `ADDRESS` struct fields.
  There is a small wrinkle at play here though which I encountered when trying to get embedded struct fields to work. The field descriptor expects to be provided with a ResolvedJavaType that 1) defines the type of the embedded struct and 2) also serves as a key to identify the corresponding (foreign) ClassEntry in the debug info model. That's problematic because when writing the field's type the generator resolves this ClassEntry to the pointer type for the foreign struct layout rather than layout itself.
  I finessed this problem by modifying 1) the field descriptor type passed through the `DebugInfo` interface (`DebugFieldInfo`) and 2) the generic debug info model type for a field (FieldEntry) to track whether the value type is being used to model an embedded data value or a reference. The DWARF generator respects that setting and types the field accordingly by setting the value type to identify the layout info rather than the info for the pointer to the layout.
  So, for example, The timeval struct now looks like this
```
(gdb) ptype 'com.oracle.svm.core.posix.headers.Time$itimerval'
type = struct itimerval {
    timeval it_interval;
    timeval it_value;
} *
(gdb) ptype 'com.oracle.svm.core.posix.headers.Time$timeval'
type = struct timeval {
    long tv_sec;
    long tv_usec;
} *
```

2. In the process of implementing the above solution for embedded structs I noticed that a foreign field whose type is an array of some other type (e.g. a `char[N]`) field also has an `ADDRESS` type `Accessor`. In such a case the `Accessor` return type that provides the type for the field is ambiguous: if the return type is `char *` then it could be returning the address of a single char field or of the first char in an array. 
  Luckily, the size field can be used to clarify this question. If it is a multiple of the value size then this is an array (and one which can be sized accordingly). So, for example, the debug info for `Utsname$utsname` now lists the type as follows:

```
(gdb) ptype 'com.oracle.svm.core.posix.headers.Utsname$utsname'
type = struct utsname {
    int8_t sysname[65];
    int8_t nodename[65];
    int8_t release[65];
    int8_t version[65];
    int8_t machine[65];
} *
```

I am now working on extending the debuginfotest test code to ensure that foreign structures are correctly handled. I'll ping you when that is complete.

adinn:
@olpaw @christianhaeubl 
I have extended `mx debuginfotest` to test operation of `CStructTests`. A few small points:

1. `mx debuginfotest` now builds and runs two separate native images `hello` and `cstructtests` (each of them built and tested with isolates on and off).
2. Each generated image is run inside gdb using its own python script, respectively `testhello.py` and `testcstruct.py`
3. Common utility code has been factored out into script `gdb_utils.py`.
4. I modified the `mx debuginfotest` command to accept option `--omit_build` which avoids regenerating the native images -- this is not for normal use but it makes life much, much easier when making changes to the gdb python scripts.
5. I had to modify class `CStructTests` to avoid calling `SubstrateUtil::strlen(CCharPointer)` (I just call `String::length()` instead) because I was seeing occasional (intermittent) exceptions during execution:
```
Exception in thread "main" java.lang.NoClassDefFoundError: com.oracle.svm.core.SubstrateUtil
	at com.oracle.svm.test.debug.CStructTests.weird(CStructTests.java:137)
	at com.oracle.svm.test.debug.CStructTests.main(CStructTests.java:223)
```

I'd be interested to hear if you have any idea as to why the exceptions happen (especially why they do not occur consistently).

zakkak:
> 4\. I modified the `mx debuginfotest` command to accept option `--omit_build` which avoids regenerating the native images -- this is not for normal use but it makes life much, much easier when making changes to the gdb python scripts.

FWIW I use:

```
env "debuginfotest_isolates=yes" gdb -ex 'python "ISOLATES=False"' -x substratevm/mx.substratevm/testhello.py ./substratevm/svmbuild/hello.hello
```

adinn:
> FWIW I use:

>     env "debuginfotest_isolates=yes" gdb -ex 'python "ISOLATES=False"' -x substratevm/mx.substratevm/testhello.py ./substratevm/svmbuild/hello.hello

Yes, I realize that it is possible to drive a test run with a suitably configured call to gdb. Unfortunately, the above command is out of date wrt the state of the script -- even before my changes (the script now builds the two versions of the hello image in  `./substratevm/svmbuild/isolates_off`and `substratevm/svmbuild/isolates_on`). I think it is better that this is provided as a command option which can be maintained as the script changes.

olpaw:
> 4\. I modified the `mx debuginfotest` command to accept option `--omit_build` which avoids regenerating the native images -- this is not for normal use but it makes life much, much easier when making changes to the gdb python scripts.

@adinn this is not needed. Since my modified that code last time it's now possible to just use
```
mx debuginfotest --dry-run
```
to achieve that. The `--dry-run` get passed to be image build command which then does nothing.

olpaw:
> I'd be interested to hear if you have any idea as to why the exceptions happen (especially why they do not occur consistently).

Sounds like a compiler bug, do you agree @christianhaeubl ?

christianhaeubl:
> I'd be interested to hear if you have any idea as to why the exceptions happen (especially why they do not occur consistently).

@adinn : if you build a single image and run it multiple times, does the exception just happen randomly or reliably?

adinn:
> @adinn this is not needed. Since I modified that code last time, it's now possible to just use

>     mx debuginfotest --dry-run

Ah, ok. I was not aware that this option bypassed the native image build step. I'll remove the --omit-build option in the next version.

substratevm/src/com.oracle.svm.test/src/hello/#Hello.java#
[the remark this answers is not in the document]
adinn: Yes. I have removed it.

adinn:
@olpaw @fniephaus

Just for the record:

I have fixed the problem that was causing a link failure in the vm and sulong gates and they appear to be working correctly now. The link error happened because the generator was writing offset -1 as the value of the type attribute of a struct field definition. It is now producing the correct offset to the correct type.

The problem only manifested in a very special case: where a POINTER type is used to type  an embedded struct field (i.e. one with an ADDRESS accessor) and the types involved appear in a specific order. During the first sizing pass the index of type referenced by  the pointer may be undefined (-1) at the point where the POINTER type is being generated. That will happen if the target type comes later in the type list.

That's not a problem for writing the POINTER type because the offset of the referent /is/ defined by the end of the first pass. However, it means that the layout index recorded for the POINTER type is set to -1 (undefined) in the first pass and only reset to the correct offset when the POINTER type is written during the second pass.

The error happened because a struct which preceded the POINTER type was relying on using the layout type. This was needed to type an embedded field whose ADDRESS type is the POINTER type. When the struct definition is written during the second pass the layout index is still -1.

I fixed this by ensuring that embedded fields of POINTER type use the type index of the POINTER type's referent rather than the layout index of the POINTER type. These (eventually) have the same values but the type index will always be set by the end of the first pass.

I believe this is the only case where this can happen. Manifestation of the problem relies on the presence of an indirection from one definition to another to derive an offset (POINTER types only), the need to consume a layout index for an unrelated type rather than the normal pointer index  (embedded struct fields only) and an unfortunate ordering of the types that have these unusual cross-dependencies (STRUCT precedes POINTER precedes referent). The fix works by systematically bypassing that second condition.

Note that a similar issue arises with void * POINTER types (i.e. types with a POINTER tag but with no referent). The original patch already finessed that case by generating the void type very early on, recording its position explicitly (in a DwarfInfoSectionImpl field) and using its offset as the layout index for all void * POINTER types.

adinn:
@olpaw @fniephaus

The PR is now passing the gate tasks that were previously failing. There is a problem with the helloworld and helloworld_debug gates but it looks like an unrelated error introduced into the mx scripts.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
```diff
@@ -85,6 +85,22 @@ public void afterRegistration(AfterRegistrationAccess access) {
...
+    @Override
+    public void beforeAnalysis(BeforeAnalysisAccess access) {
+        /*
+         * Make the name provider aware of the native libs
+         */
+        if (!UniqueShortNameProviderDefaultImpl.UseDefault.useDefaultProvider()) {
+            if (ImageSingletons.contains(UniqueShortNameProvider.class)) {
+                UniqueShortNameProvider provider = ImageSingletons.lookup(UniqueShortNameProvider.class);
+                if (provider instanceof NativeImageBFDNameProvider) {
+                    var accessImpl = (FeatureImpl.BeforeAnalysisAccessImpl) access;
```
olpaw, line 97: Instead of this complex chain of checks to determine if you have a `NativeImageBFDNameProvider` instance just add a private field to class `NativeImageDebugInfoFeature` that holds a ref to the `NativeImageBFDNameProvider` you instantiated in `afterRegistration`. I.e. in `afterRegistration` you change:
```java
ImageSingletons.add(UniqueShortNameProvider.class, new NativeImageBFDNameProvider(ignored));
``` to
```java
bfdNameProvider = new NativeImageBFDNameProvider(ignored);
ImageSingletons.add(UniqueShortNameProvider.class, bfdNameProvider);
```
Then here you can just use
```java
if (bfdNameProvider != null) { ... }
```

substratevm/src/com.oracle.svm.test/src/hello/#Hello.java#
```diff
@@ -0,0 +1,247 @@
+/*
```
olpaw: I guess this file should not be committed, right?

olpaw (changes requested):
We are very close now. Only one small fix needed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
adinn: I just pushed a fix for this

olpaw (approved):

olpaw:
@adinn unfortunately I do get NPEs in EE gates. They fail with:
```
com.oracle.svm.core.util.VMError$HostedError: java.lang.UnsupportedOperationException: null not supported as key!
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.util.VMError.shouldNotReachHere(VMError.java:82)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.image.NativeImage.write(NativeImage.java:167)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.image.NativeImageViaCC.write(NativeImageViaCC.java:98)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:735)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:544)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:408)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:594)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.start(NativeImageGeneratorRunner.java:134)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:94)
Caused by: java.lang.UnsupportedOperationException: null not supported as key!
	at org.graalvm.sdk/org.graalvm.collections.EconomicMapImpl.checkKeyNonNull(EconomicMapImpl.java:640)
	at org.graalvm.sdk/org.graalvm.collections.EconomicMapImpl.get(EconomicMapImpl.java:242)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfDebugInfo.lookupTypeProperties(DwarfDebugInfo.java:558)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfDebugInfo.getTypeIndex(DwarfDebugInfo.java:592)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfSectionImpl.getTypeIndex(DwarfSectionImpl.java:797)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeEmbeddedArrayDataType(DwarfInfoSectionImpl.java:1366)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeStructField(DwarfInfoSectionImpl.java:344)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.lambda$writeStructFields$4(DwarfInfoSectionImpl.java:323)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1707)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeStructFields(DwarfInfoSectionImpl.java:322)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeForeignStructLayout(DwarfInfoSectionImpl.java:982)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeForeignLayout(DwarfInfoSectionImpl.java:923)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeInstanceClassInfo(DwarfInfoSectionImpl.java:409)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.lambda$writeInstanceClasses$5(DwarfInfoSectionImpl.java:388)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1707)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeInstanceClasses(DwarfInfoSectionImpl.java:387)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.generateContent(DwarfInfoSectionImpl.java:190)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfInfoSectionImpl.writeContent(DwarfInfoSectionImpl.java:117)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.ObjectFile.debugContext(ObjectFile.java:1833)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.dwarf.DwarfSectionImpl.getOrDecideContent(DwarfSectionImpl.java:643)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.elf.ELFUserDefinedSection.getOrDecideContent(ELFUserDefinedSection.java:110)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.ObjectFile.bake(ObjectFile.java:1625)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.ObjectFile.write(ObjectFile.java:1282)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.ObjectFile.lambda$write$5(ObjectFile.java:1274)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.ObjectFile.withDebugContext(ObjectFile.java:1814)
	at org.graalvm.nativeimage.objectfile/com.oracle.objectfile.ObjectFile.write(ObjectFile.java:1273)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.image.NativeImage.write(NativeImage.java:165)
	... 7 more
```
I think it has to do with compressed references. I will keep you updated.

adinn:
@olpaw Hmm, it would be interesting to know what the struct and struct field are when this blows up.

The assumption at the point where the NPE happens is that an ADDRESS struct field will be typed by dereferencing a foreign pointer type to a known foreign target type. The pointer type is the derived from return type of the ADDRESS getter. Which means the filed itself must either be a value of the target type or an array of such values. That explains why the call to getPointerTo() is unchecked (it should, of course, have been asserted non-null).

It might well be that in this specific case we have a struct field whose ADDRESS getter return a pointer to a foreign primitive type (or maybe a Word type?). I may not have correctly catered for that case.

If you have a problem debugging this and can provide a reproducer I will be happy to help investigate.

adinn:
@olpaw n.b. when I say a pointer type I mean a ForeignTypeEntry derived from a ResolvedJavaType which 1) is a subclass of PointerBase and 2) has attached element info with kind `POINTER`. Of course, that does not on its own guarantee that `getPointerTo()` will return non-null. The latter method uses a value established in the call to `ForeignTypeEntry::addDebugInfo`. The lookup checks the target of the `CPointerTo` annotation on the ResolvedJavaType. If it is non-null then it ought to map to another ForeignTypeEntry that was created when traversing the image's type list. So, a null return from `getPointerTo()` could indicate 1) a missing `CPointerTo` annotation (maybe because it is inherited?) or 2) a lookup failure when translating from the ResolvedJavaClass referenced in the annotation to the expected ForeignTypeEntry.

adinn:
@olpaw I think I may understand where the problem is coming from here. It is quite possible for ForeignTypeEntry::isPointer() to return true and ForeignTypeEntry::getPointerTo() to return false. Indeed that case is catered for when writing the info for the type by generating a `DW_TAG_pointer_type` record that references the corresponding  `DW_TAG_base_type` record for type `void`.

So, when defining the type for an embedded (ADDRESS getter) struct element/element array field this case needs to be detected and handled correctly. The current code assumes that the 'isPointer()' `ForeignTypeEntry` identified by the return type of the ADDRESS getter will return a non-null value from `getPointerTo()`.

Unfortunately, when this return value is null because I have encoded the pointer type as `void *`. So, using the same case handling would mean declaring the struct field to have type `void` or `void[]` which is obviously not going to work. So, I think I will need to use a suitable pointer to integral (WORD) type for these cases e.g. uint64_t *.

olpaw:
> Hmm, it would be interesting to know what the struct and struct field are when this blows up.

@adinn the struct in question is `com.oracle.svm.core.code.CodeInfoPointer` which is a `@RawPointerTo` the `com.oracle.svm.core.code.CodeInfo` `@RawStructure`.

olpaw:
Adding
```patch
diff --git a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ForeignTypeEntry.java b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ForeignTypeEntry.java
index 825984255b8..05c543da943 100644
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ForeignTypeEntry.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ForeignTypeEntry.java
@@ -78,6 +78,8 @@ public class ForeignTypeEntry extends ClassEntry {
             ResolvedJavaType referent = debugForeignTypeInfo.pointerTo();
             if (referent != null) {
                 pointerTo = debugInfoBase.lookupTypeEntry(referent);
+            } else {
+                System.out.println("FLAG_POINTER without pointerTo: " + debugForeignTypeInfo.typeName() + '(' + debugForeignTypeInfo.typedefName() + ')');
             }
         } else if (debugForeignTypeInfo.isIntegral()) {
             flags = FLAG_INTEGRAL;
```
shows me the following problematic cases:
```
FLAG_POINTER without pointerTo: com.oracle.svm.core.code.CodeInfoPointer(CodeInfo*)
FLAG_POINTER without pointerTo: com.oracle.svm.core.heap.dump.HeapDumpMetadata$FieldInfoPointer(FieldInfo*)
FLAG_POINTER without pointerTo: com.oracle.svm.core.heap.dump.HeapDumpMetadata$FieldNamePointer(FieldName*)
FLAG_POINTER without pointerTo: com.oracle.svm.core.thread.PlatformThreads$OSThreadHandlePointer(OSThreadHandle*)
```

adinn:
@olpaw I believe the assumption that an ADDRESS-only struct field has a known foreign type is actually valid. What is missing -- apart from the asserts that `getPointerTo()` returns non-null -- is a lookup of the `pointerTo` value via a `RawPointerTo` annotation when the debug info provider is creating a `DebugForeignTypeInfo` instance. The current code only looks up target types declared via a `CPointerTo` annotation.

I have just pushed a patch which adds the secondary lookup. It also asserts that `getPointerTo()` returns non-null when we are attempting to type an embedded struct field. It passes `debuginfotest` without G1. Please let me know if this resolves the issue when running with G1.

olpaw:
> Please let me know if this resolves the issue when running with G1.

Yes. https://github.com/oracle/graal/pull/6625/commits/75a2a27effdfab0ecc081373a5838511d57778e9 fixes the issue.
I'm cherry-picking it over to https://github.com/oracle/graal/pull/6774

olpaw:
There is one more aarch64 specific issue. On EE I see
```
(gdb) stepi
method prologue is unexpectedly long, did not reach end after 6 stepis
``` on some gates. I assume I just need to increase `max_num_stepis` a bit for aarch64.

adinn:
> There is one more aarch64 specific issue. On EE I see
> 
>       (gdb) stepi
>       method prologue is unexpectedly long, did not reach end after 6 stepis
> 
> on some gates. I assume I just need to increase max_num_stepis a bit for aarch64.

Yes, I guess that is all that will be needed. It would probably be best to eyeball the instruction sequence just to be sure that the EE build is still terminating the prologue with an instruction that matches `instruction_adjusts_sp_register_pattern`.

olpaw:
> Yes, I guess that is all that will be needed. It would probably be best to eyeball the instruction sequence just to be sure that the EE build is still terminating the prologue with an instruction that matches `instruction_adjusts_sp_register_pattern`

The issue was something else entirely.
For some reason (I have yet to figure out) in our aarch gates we see 'amd64' for `arch` in `testhello.py` and thus the `if arch == "aarch64"` branch is never taken.

olpaw:
PR is now in the merge queue and should land on master today.

olpaw:
Merged as https://github.com/oracle/graal/pull/6774

adinn:
> Merged as https://github.com/oracle/graal/pull/6774

@olpaw Thanks for all your help getting this merged.

PR #6627, opened 2023-05-17
Fix for JFR virtual thread tests

roberttoyonaga:
It's possible for threads to reuse IDs of inactive threads. This means that we have to replace the `expectedThreads` set with a map that records a count of seen threads for each ID. Additionally, the previous implementation of `expectedThreads` was not thread safe. Writes from some threads may not be immediately visible to other threads.  It's been changed to a `ConcurrentHashMap`. This should resolve some of the JFR virtual thread test failures.

This PR does not fix all the problems with `TestVirtualThreadsChunkRotation`. There is still a problem related to dumping snapshots I must investigate further, so it is still skipped for now.

This is related to : https://github.com/oracle/graal/pull/6464#issuecomment-1551367336

### DIFF
--- a/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestVirtualThreadsBasic.java
+++ b/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestVirtualThreadsBasic.java
@@ -31,6 +31,7 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assume.assumeTrue;
 
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


roberttoyonaga:
> I don't see any reason why the thread IDs of inactive threads would be reused

Yes, you are correct. I was looking at an old version of the specification docs. Since JDK20 (and in the SVM implementation) IDs are never re-used. However, I do still think there is a need for the `ConcurrentHashMap`. I found that, with the original Hashset implementation of `expectedThreads`, the main thread occasionally does not see the all the data worker threads have produced until after all the recorded events have been iterated. This results in`expectedThreads` not being empty when the assertion checks it.

roberttoyonaga:
oh I see that [you already made some of the sets used in the JFR tests synchronized](https://github.com/oracle/graal/commit/435d31855e74794a1e062f9e04a8da255512e6ce).  I've changed this PR to make `TestVirtualThreadsChunkRotation` and `TestVirtualThreadsBasic` also use synchronized sets. The problem I've described in the above comment should be fixed now so I've removed the `@Ignore` on `TestVirtualThreadsJfrStreaming`.

christianhaeubl:
Thanks, this PR will be merged in the next few days.

PR #6646, opened 2023-05-20
Injecting JVM arguments via `-Dnative-image.benchmark.extra-jvm-arg`

ceresek:
Benchmarks with JNI classes, such as Renaissance with the PAPI agent plugin, require adding `-Djava.library.path` to the JVM in the agent and run phases. This patch permits injecting the option via `-Dnative-image.benchmark.extra-jvm-arg`.

@farquet Could you please take a look if this or something with similar functionality could be added ?


farquet:
That's a useful addition. Thanks @ceresek for the contribution! I'll do a round of internal testing of the new feature.

PR #6681, opened 2023-05-25
[GR-44217] Truffle opens itself in runtime to languages and instruments on the module-path.

graalvmbot:


substratevm/mx.substratevm/suite.py
```diff
@@ -1331,6 +1331,7 @@
                 "POINTSTO",
                 "compiler:GRAAL",
                 "NATIVE_IMAGE_BASE",
+                "truffle:TRUFFLE_API",
```
zakkak, line 1334: @tzezula is this really necessary?

PR #6746, opened 2023-06-06
Remove CI jobs using non existing tags

zakkak:
Fixes recent (since the merge of https://github.com/oracle/graal/pull/6738)  CI failures, e.g.:

* https://github.com/oracle/graal/actions/runs/5176166290/jobs/9324638530
* https://github.com/oracle/graal/actions/runs/5177381458/jobs/9327406668

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -84,10 +84,6 @@ jobs:
               PIP_PACKAGES: "jsonschema==4.6.1"
           - env:
               JDK_VERSION: "21"
-              GATE_TAGS: "build,helloworld_debug"
-              PRIMARY: "substratevm"
-          - env:
-              JDK_VERSION: "21"
               GATE_TAGS: "build,debuginfotest"
               PRIMARY: "substratevm"
           - env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,7 +79,7 @@ jobs:
               PRIMARY: "substratevm"
           - env:
               JDK_VERSION: "21"
-              GATE_TAGS: "build,test,helloworld"
+              GATE_TAGS: "build,helloworld,native_unittests"
               PRIMARY: "substratevm"
               PIP_PACKAGES: "jsonschema==4.6.1"
           - env:


zakkak:
For the record, the strict tag checking was added to `mx` with https://github.com/graalvm/mx/commit/b209e64b352788f4ae462a5263fff6aaac7d68bd

.github/workflows/main.yml
```diff
@@ -79,13 +79,9 @@ jobs:
               PRIMARY: "substratevm"
           - env:
               JDK_VERSION: "21"
-              GATE_TAGS: "build,test,helloworld"
```
zapster, line 82: I guess it makes sense to replace that with something else. This is what the internal CI is running:
https://github.com/oracle/graal/blob/master/substratevm/ci/ci.jsonnet#L115
So maybe run `native_unittests` here...

.github/workflows/main.yml
```diff
@@ -79,13 +79,9 @@ jobs:
...
+              GATE_TAGS: "build,helloworld"
               PRIMARY: "substratevm"
               PIP_PACKAGES: "jsonschema==4.6.1"
-          - env:
-              JDK_VERSION: "21"
-              GATE_TAGS: "build,helloworld_debug"
```
zapster, line 87: ... and `debuginfotest` here?

zapster:
Thanks @zakkak for the fix. I guess the tags were originally added to test something but we did not notice they no longer exist (that is what motivates the strict tag checking in the first place). So we should probably replace them with something else. Some suggestions in-line.

.github/workflows/main.yml
zakkak: Interesting, I didn't notice we were not running `native_unittests`... Adding it, thanks

.github/workflows/main.yml
zakkak: We already run `debuginfotest` in https://github.com/oracle/graal/blob/f658f09475f468ed9e50934267c44aa6e4c4987a/.github/workflows/main.yml#L85-L88

zapster (approved):
Thanks again. I'll integrate it ASAP.

PR #6799, opened 2023-06-14
Fix constant pool JFR ID bug

roberttoyonaga:
**Description of Bug**

The `JfrSymbolRepository` was allowing for JFR IDs to be duplicated across epochs. This results in incorrect symbol constant pool data. Consumer code that reads the JFR data on disk will confuse the values that share the same IDs. VisualVm also interprets the wrong symbol values due to the ID duplication.
 
Although `JfrSymbolHashtable.nextId` monotonically increases and is never reset, there were separate `JfrSymbolHashtable`s for each epoch, resulting in two `nextId` counts. This effectively results in allowing the same ID key to be used twice. 

The[ hotspot code that handles symbol IDs forces them to be unique.](https://github.com/openjdk/jdk/blob/master/src/hotspot/share/jfr/support/jfrSymbolTable.cpp#L37-L39) 

The stacktrace repo suffers from the same problem. The other repositories do not have the same problem because they don't create the IDs themselves (thread repo uses already available TIDs and method repo uses method IDs etc).

**Fix**

Make `JfrSymbolHashtable.nextId` and `JfrStackTraceTable.nextId` static. This way the count is shared across the two different `epochData`s.
This fix also allows `TestVirtualThreadsChunkRotation` to reliably pass. I have removed the `@Ignore` annotation.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrStackTraceRepository.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrStackTraceRepository.java
@@ -301,7 +301,7 @@ public class JfrStackTraceRepository implements JfrRepository {
     }
 
     public static final class JfrStackTraceTable extends AbstractUninterruptibleHashtable {
-        private long nextId;
+        private static long nextId;
 
         @Override
         @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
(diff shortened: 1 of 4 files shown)


christianhaeubl:
Nice find, thanks for fixing that. This PR will be merged in the next few days.

roberttoyonaga:
@christianhaeubl since this bug affects previous versions of JFR in SVM, should it be backported to GraalVM 23.0? Should I just open a PR to the 23.0 release branch?

PR #6823, opened 2023-06-16
load .jnilib on MacOS from the native-image executable directory

### LINKED ISSUE #6822: .jnilib not loaded from the executable directory
**Describe the issue**
As of #5932 native libs are loaded from the directory containing the native image. However on MacOS this only works for `.dylib` libraries and not for `.jnilib` because `ClassLoaderHelper.mapAlternativeName` is not considered foro `sysPath` in this coode:
https://github.com/oracle/graal/blob/f9ef31f0c872872f7008ae771540f3020223414c/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java#L122-L135

I don't have a minimal example, but the CI on this PR fails because of this: https://github.com/xerial/sqlite-jdbc/pull/922

 - GraalVM version: **GraalVM CE 17.0.7 and 20.0.1**
 - OS: **macos-latest** on GitHub Actions

**More details**
N/A

kkriske:
fixes #6822

I'm not sure how I can add a test for this.
I have no way to actually test this locally either, because it only fails in GitHub Actions and I do not own a MacOS device.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java
@@ -120,20 +120,27 @@ public final class NativeLibrarySupport {
             usrPaths = tokens;
         }
         String libname = System.mapLibraryName(name);
-        if (sysPath != null && loadLibrary0(new File(sysPath, libname), false)) {
+        if (sysPath != null && loadLibraryFromPath(sysPath, libname)) {
             return;
         }
         for (String path : usrPaths) {
-            File libpath = new File(path, libname);
-            if (loadLibrary0(libpath, false)) {
-                return;
-            }
-            File altpath = Target_jdk_internal_loader_ClassLoaderHelper.mapAlternativeName(libpath);
-            if (altpath != null && loadLibrary0(altpath, false)) {
+            if (loadLibraryFromPath(path, libname)) {
                 return;
             }
         }
         throw new UnsatisfiedLinkError("no " + name + " in java.library.path");
+    }
+
+    private boolean loadLibraryFromPath(String path, String libname) {
+        File libpath = new File(path, libname);
+        if (loadLibrary0(libpath, false)) {
+            return true;
+        }
+        File altpath = Target_jdk_internal_loader_ClassLoaderHelper.mapAlternativeName(libpath);
+        if (altpath != null && loadLibrary0(altpath, false)) {
+            return true;
+        }
+        return false;
     }
 
     /** Returns the directory containing the native image, or {@code null}. */


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java
```diff
@@ -120,22 +120,29 @@ public void loadLibraryRelative(String name) {
...
+            if (loadLibraryFromPath(path, libname)) {
                 return;
             }
         }
         throw new UnsatisfiedLinkError("no " + name + " in java.library.path");
     }
 
+    private boolean loadLibraryFromPath(String path, String libname) {
+        File libpath = new File(path, libname);
+        if (loadLibrary0(libpath, false)) {
```
abimarank, line 136: No need for this `if`, could return `loadLibrary0(libpath, false)`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java
```diff
@@ -120,22 +120,29 @@ public void loadLibraryRelative(String name) {
...
+            return true;
+        }
+        File altpath = Target_jdk_internal_loader_ClassLoaderHelper.mapAlternativeName(libpath);
+        if (altpath != null && loadLibrary0(altpath, false)) {
```
abimarank, line 140: Here also, `if` condition can be removed.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibrarySupport.java
kkriske: This is incorrect, directly returning `loadLibrary0` makes the altpath flow unreachable making the issue these changes are supposed to fix even worse.
If I may be so bold, why are you reviewing this as a non oracle employee without being specifically asked to review this? This now gives the impression someone is looking into this which will cause this to go unnoticed by people that have an actual say in what happens with this PR.
Please refrain from adding random reviews on PRs you have no affiliation with whatsoever, especially if the input you provide is wrong.

oubidar-Abderrahim:
Thank you for opening this PR.

oubidar-Abderrahim:
@christianwimmer would you mind assigning someone to review and approve this PR? Thank you

kkriske:
@oubidar-Abderrahim The location of the relevant code ahs moved (to NativeLibraries.java) but the issue is still there I think.
We simply just switched the extension we were using, but for people not in charge of that, this is quite an obscure suble bug. A bit unfortunate is was completely ignored especially since it's so easy to fix.

Issue #6814, opened 2023-06-16
Exception on reading MemoryUsage (possible wrong max memory value)


joggeli34:
**Describe the issue**
With the newest graalvm release, it is now possible to read out the memory usage. We use quarkus with micrometer, where we get following exception:

```
2023-06-15 14:18:58,409 WARN  [io.mic.cor.ins.int.DefaultGauge] (executor-thread-3) Failed to apply the value function for the gauge 'jvm.memory.used'. Note that subsequent logs will be logged at debug level.: java.lang.IllegalArgumentException: committed = 8573528 should be < max = 8388608
	at java.management@17.0.7/java.lang.management.MemoryUsage.<init>(MemoryUsage.java:166)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.genscavenge.AbstractMemoryPoolMXBean.memoryUsage(AbstractMemoryPoolMXBean.java:75)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.genscavenge.GenScavengeMemoryPoolMXBeans$EdenMemoryPoolMXBean.getUsage(GenScavengeMemoryPoolMXBeans.java:94)
	at io.micrometer.core.instrument.binder.jvm.JvmMemory.getUsage(JvmMemory.java:76)
	at io.micrometer.core.instrument.binder.jvm.JvmMemory.getUsageValue(JvmMemory.java:66)
	at io.micrometer.core.instrument.binder.jvm.JvmMemoryMetrics.lambda$bindTo$0(JvmMemoryMetrics.java:82)
	at io.micrometer.core.instrument.internal.DefaultGauge.value(DefaultGauge.java:53)
	at io.micrometer.prometheus.PrometheusMeterRegistry.lambda$newGauge$5(PrometheusMeterRegistry.java:331)
	at io.micrometer.prometheus.MicrometerCollector.collect(MicrometerCollector.java:75)
	at io.prometheus.client.Collector.collect(Collector.java:45)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.findNextElement(CollectorRegistry.java:204)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.nextElement(CollectorRegistry.java:219)
	at io.prometheus.client.CollectorRegistry$MetricFamilySamplesEnumeration.nextElement(CollectorRegistry.java:152)
	at io.prometheus.client.exporter.common.TextFormat.write004(TextFormat.java:71)
[...]

PR #6930, opened 2023-07-04
Use -1 as max value for eden and survivor spaces

petermz:
This PR addresses #6814.

It turns out max eden size is enforced post factum, i.e. if 'memory used' >= 'max size' then GC is started (see [`AbstractCollectionPolicy::shouldCollectOnAllocation`](https://github.com/oracle/graal/blob/8b81a03fee458013b8b703e234e77edc4836be43/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractCollectionPolicy.java#L90)). This applies also to young generation as a whole -- see [`BasicPolicy::shouldCollectOnAllocation`](https://github.com/oracle/graal/blob/8b81a03fee458013b8b703e234e77edc4836be43/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/BasicCollectionPolicies.java#L60). So we cannot assume that maximum size, as reported by collection policy, will never be exceeded. It seems there is no meaningful estimation for the maximum size of young generation spaces, other than `MaxHeapSize`.

The solution proposed here is to return -1 for young gen (eden and survivor) spaces. This is also how Hotspot behaves in JDKs 17 and 19.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractMemoryPoolMXBean.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractMemoryPoolMXBean.java
@@ -45,12 +45,14 @@ import sun.management.Util;
 
 public abstract class AbstractMemoryPoolMXBean extends AbstractMXBean implements MemoryPoolMXBean {
 
+    protected static final UnsignedWord UNDEFINED = WordFactory.signed(UNDEFINED_MEMORY_USAGE);
+    private static final UnsignedWord UNINITIALIZED = WordFactory.zero();
+
     private final String name;
     private final String[] managerNames;
     protected final UninterruptibleUtils.AtomicUnsigned peakUsage = new UninterruptibleUtils.AtomicUnsigned();
 
-    private static final UnsignedWord UNDEFINED = WordFactory.zero();
-    protected UnsignedWord initialValue = UNDEFINED;
+    protected UnsignedWord initialValue = UNINITIALIZED;
 
     @Platforms(Platform.HOSTED_ONLY.class)
     protected AbstractMemoryPoolMXBean(String name, String... managerNames) {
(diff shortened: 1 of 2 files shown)


oubidar-Abderrahim:
Hi, Thank you for contributing to GraalVM, could you please specify the following

- Is your PR ready for review?
- Is there a specific engineer from the GraalVM team that you want to review this PR? (because he has more context or you discussed these changes with them previously)

petermz:
It's ready for review; as for the reviewer, I don't have any preference.

christianhaeubl:
Thanks for the PR. For eden, this change makes sense.
Why is it necessary to use `UNDEFINED` for the survivor space?

petermz:
Currently `CollectionPolicy.getMaximumSurvivorSize()` is used. Apparently this method isn't called anywhere else, so I felt it was unsafe to rely on.

Furthermore, I now can reproduce this issue with old gen bean as well. This makes sense because `OldGeneration` does no size checks upon adding new chunks (see e.g. `requestAlignedChunk`, or `promoteUnalignedObject`), so it may grow beyond what the policy thinks its max size is.

And it seems `CollectionPolicy.getMaximumHeapSize()` is unsafe too, as it may as well be temporarily exceeded -- see e.g. check in `GCImpl.doCollectOnce`. So I guess all the beans should better return -1 for max size.

joggeli34:
This would really help us, as this blocks us from updating to the newest graalvm, as creates a lot of exceptions.

manofthepeace:
@joggeli34 if you use something like quarkus you can alleviate the issue this way. Not perfect but...

```java
    @Produces
    @Singleton
    // TODO remove once https://github.com/oracle/graal/issues/6814 is fixed and released
    public MeterFilter removeJvmUsed() {
        return MeterFilter.denyNameStartsWith("jvm.memory");
    }
```

christianhaeubl:
Thanks, I will try to get this PR merged.

Issue #6944, opened 2023-07-06
Thresholded JFR events in SubstrateVM are not obeying the threshold


jbachorik:
**Describe the issue**
When running a native image application with JFR the recording may get swamped with threads like `Monitor Wait` or `Thread Park` which are very short and should have been ignored according to the threshold limit (the default is 10ms).

**Steps to reproduce the issue**
Run any native image application which is using thread pools with JFR and capture the recording (one such recording is attached).

Here are the `Java Monitor Wait Events` with duration of several nanoseconds
<img width="709" alt="Java Monitor Wait Events" src="https://github.com/oracle/graal/assets/738413/59fcd544-2381-47be-8cb9-737bb97fe982">
while the config is defining the threshold of 20ms
<img width="727" alt="Java Monitor Wait Config" src="https://github.com/oracle/graal/assets/738413/ee0c2d65-3c69-450e-8186-090aa659ad25">

**Describe GraalVM and your environment:**
 - Oracle GraalVM 20.0.1+9
 - OS: **Linux/Ubuntu**
 - Architecture: **AMD64**

[JFR Recording](https://drive.google.com/file/d/1DaFr1sGMBe0CMIWoe9dkLs0h4dD_yvn4/view?usp=sharing) (I had to use GDrive because of the 25MiB limit on GitHub)

roberttoyonaga:
Hi @jbachorik. Thanks for bringing this up. I'll work on fixing this next week.

PR #6964, opened 2023-07-10
Check threshold for native JFR events

roberttoyonaga:
fixes: https://github.com/oracle/graal/issues/6944 

**Description**
Adds checking that duration threshold is exceeded before emitting native events. Java level events are not affected because their threshold values are maintained by `jdk.jfr.internal.PlatformEventType`

This issue is present in previous versions of Native Image JFR and might be worth backporting to 23.0 because the changes are minimal.


christianhaeubl:
Thanks for the PR. I did a few cleanups and minor fixes on top. This should get merged in the next few days.

PR #6966, opened 2023-07-10
Optimize JFR Type Repository Serialization

roberttoyonaga:
### Description of Problem
I noticed that significant time was being spent to serialize the JFR type repository. This was discovered using the SIGPROF based JFR method sampler. The JFR type repository must be serialized every chunk rotation as well as periodically (every second, or more frequently if specified by the user). Additionally, when the `jdk.ObjectAllocationInNewTLAB` event is enabled , it can be observed that the allocation slow path is often taken here as well. The underlying reason the type repository serialization is slow is because [all classes in the heap are iterated and the "loaded" ones get put in a newly allocated list each time](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrTypeRepository.java#L90C31-L90C64). The large new list often requires a new TLAB and the iteration itself takes time. 

### Solution
This PR implements caching of the list of "loaded" classes and checks if the cache is stale before regenerating it.

### Profiling Experiment and Results
A simple test was done with 10 trials comparing with/without the caching mechanism. The test involved running a stock quarkus hibernate quickstart app and uses the SIGPROF sampler to determine the time spent in the JFR type repository code retrieving the "loaded" classes data. The `jdk.ObjectAllocationInNewTLAB` event was enabled to see if there would be a reduction in the number of times the allocation slow path needs to be taken. Each JFR recording ran for 20s, during which time [Hyperfoil](https://hyperfoil.io/) was used to make 10 requests/second  to the app's endpoints.  Results are below:

**% of total time spent  in `Heap.getLoadedClasses()`**
  | original | optimized
-- | -- | --
Mean | 0.06590249953 | 0.03987949529
Median | 0.05914918415 | 0.03721942281
STDV | 0.03863853669 | 0.01660919135
MAX | 0.1346153846 | 0.06818181818

p-value over 10 trials unequal variance, two tails: 0.0736362926

**Total TLABs created by Heap.getLoadedClasses()**
  | original | optimized
-- | -- | --
Mean | 12 | 4
Median | 12 | 4
STDV | 0 | 0
MAX | 12 | 4

Note: These figures don't represent every scenario because the percentage of time spent doing JFR periodic operations will depend on how busy the program is, but this should at least indicate some improvement. 

 
The following repo has more details and code used the experiment. You can also find the 20 generated ".jfr" snapshot files there (10 with the optimization, 10 control).
https://github.com/roberttoyonaga/quarkus-quickstarts/tree/optimization-study/hibernate-orm-quickstart 

###  RSS Experiment and Results
A simple plaintext quarkus app was used to measure the RSS after first response running JFR with and without the new development commits.

**RSS measured after first response upon start-up (KB)**
  | original | optimized
-- | -- | --
Mean | 64341.9 | 62115
Median | 64803 | 62194
STDV | 1701.106793 | 923.1938523
MAX | 66507 | 63045

p-value over 10 trials unequal variance, two tails: 0.00272094022

See here for the test code and scripts: https://github.com/rh-jmc-team/jfr-native-image-performance


substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapImpl.java
```diff
@@ -118,6 +119,10 @@ public final class HeapImpl extends Heap {
     /** A cached list of all the classes, if someone asks for it. */
     private List<Class<?>> classList;
 
+    /** A cached list of all the loaded classes, if someone asks for it. */
+    private List<Class<?>> loadedClasses;
```
galderz, line 123: I don't know if this will work at all, but did you consider using `PredefinedClassesSupport.loadedClassesByName` instead of creating a new cache?

galderz, line 123: If the cache is necessary, you should also consider the impact of this cache from a memory consumption perspective. To assess that you would need to run your experiments before/after and compare max RSS consumption.

roberttoyonaga: That's a great idea! I did just that and found some very interesting results.

Experiment routine: Use the same hibernate-orm Quarkus quickstart that I used in the previous two experiments. Use use the `ps` command to obtain the RSS as described in the [Quarkus docs](https://quarkus.io/guides/native-reference#rss). Repeat 5 times with/without the optimization changes.

Results:
For both with/without the optimization changes, the first RSS measurement taken after starting the app is not statistically different (between 101000 - 110000 with much variance in between). For comparison, running the app in java mode results in RSS 200000+.   However, the interesting result is that, in the unoptimized version, RSS steadily creeps up a total of about 7000 over a 10s interval (with no activity hitting the app endpoints). I believe this is due to the JFR periodic task allocating new TLABs every second. The optimized version does not have this problem.

roberttoyonaga: Hmm that's a good idea.  I gave it a try, but I think that `PredefinedClassesSupport.loadedClassesByName` doesn't actually cache **all** loaded classes, just the ones that are "dynamically" loaded at runtime (although they aren't truly "loaded" in the usual sense).
If I try to reuse `PredefinedClassesSupport.loadedClassesByName`, most of the classes show up as `null`. See below: 

image

galderz (approved):
Looks good, thanks @roberttoyonaga

christianhaeubl:
Thanks for doing the measurements and for the PR. The list of loaded classes can be pretty big, so I would like to avoid having one extra copy in the heap. I replaced the method with a visitor, see https://github.com/oracle/graal/pull/7317

PR #6985, opened 2023-07-12
[GR-49816] Throw exception for null pointers passed to RuntimeJNIAccess / RuntimeReflection register methods

zakkak:
Don't allow null values to be passed to the `register` method of `RuntimeJNIAccess` and `RuntimeReflection`. Since these are public APIs GraalVM should either handle null values (by ignoring them in this case) or throw a `NullPointerException` before creating an asynchronous task to perform the registration in the analysis, which then results in `NullPointerException`s being thrown later when it's no longer possible to understand where the null value originate from.

Closes: https://github.com/quarkusio/quarkus/issues/34698 

If accepted, it would also be good to get this backported to 23.1 and 23.0 (and 22.3?).

### DIFF
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
@@ -46,13 +46,7 @@ import java.util.Arrays;
 
 public interface ReflectionRegistry {
     default void register(ConfigurationCondition condition, Class<?>... classes) {
-        Arrays.stream(classes).forEach(clazz -> {
-            if (clazz == null) {
-                throw new NullPointerException("Cannot register null value as class for reflection. " +
-                                "Please ensure that all values you register are not null.");
-            }
-            register(condition, false, clazz);
-        });
+        Arrays.stream(classes).forEach(clazz -> register(condition, false, clazz));
     }
 
     void register(ConfigurationCondition condition, boolean unsafeAllocated, Class<?> clazz);
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
@@ -46,7 +46,13 @@ import java.util.Arrays;
 
 public interface ReflectionRegistry {
     default void register(ConfigurationCondition condition, Class<?>... classes) {
-        Arrays.stream(classes).forEach(clazz -> register(condition, false, clazz));
+        Arrays.stream(classes).forEach(clazz -> {
+            if (clazz == null) {
+                throw new NullPointerException("Cannot register null value as class for reflection. " +
+                        "Please ensure that all values you register are not null.");
+            }
+            register(condition, false, clazz);
+        });
     }
 
     void register(ConfigurationCondition condition, boolean unsafeAllocated, Class<?> clazz);
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
@@ -49,7 +49,7 @@ public interface ReflectionRegistry {
         Arrays.stream(classes).forEach(clazz -> {
             if (clazz == null) {
                 throw new NullPointerException("Cannot register null value as class for reflection. " +
-                        "Please ensure that all values you register are not null.");
+                                "Please ensure that all values you register are not null.");
             }
             register(condition, false, clazz);
         });
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ConditionalConfigurationRegistry.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ConditionalConfigurationRegistry.java
@@ -38,6 +38,14 @@ public abstract class ConditionalConfigurationRegistry {
     private final Map<String, Collection<Runnable>> pendingReachabilityHandlers = new ConcurrentHashMap<>();
 
     protected void registerConditionalConfiguration(ConfigurationCondition condition, Runnable runnable) {
+        if (condition == null) {
+            throw new NullPointerException("Cannot use null value as condition for conditional configuration. " +
+                            "Please ensure that you register a non-null condition.");
+        }
+        if (runnable == null) {
+            throw new NullPointerException("Cannot use null value as runnable for conditional configuration. " +
+                            "Please ensure that you register a non-null runnable.");
+        }
         if (ConfigurationCondition.alwaysTrue().equals(condition)) {
             /* analysis optimization to include new types as early as possible */
             runnable.run();
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ConditionalConfigurationRegistry.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ConditionalConfigurationRegistry.java
@@ -26,6 +26,7 @@ package com.oracle.svm.hosted;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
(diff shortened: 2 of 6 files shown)


fniephaus:
> Since these are public APIs GraalVM should either handle null values (by ignoring them in this case) or throw a NullPointerException before creating an asynchronous task to perform the registration in the analysis, which then results in NullPointerExceptions being thrown later when it's no longer possible to understand where the null value originated from.

+1 for throwing NPEs, so that users are forced to look into registrations and other API calls that do nothing (e.g., when a class/field/method lookup returned `null`).

zakkak:
@fniephaus this one slept through the cracks. I have updated it to thrown `NullPointerException`s as discussed. Please have a look.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
```diff
@@ -46,7 +46,12 @@
 
 public interface ReflectionRegistry {
     default void register(ConfigurationCondition condition, Class<?>... classes) {
-        Arrays.stream(classes).forEach(clazz -> register(condition, false, clazz));
+        Arrays.stream(classes).forEach(clazz -> {
+            if (clazz == null) {
+                throw new NullPointerException("Trying to register null for reflection");
```
fniephaus, line 51: I'm no native speaker but "trying" may not be clear enough for users here. How about:

```suggestion
                throw new NullPointerException("Cannot register null value as class for reflection. Please ensure that all values you register are not null.");
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ConditionalConfigurationRegistry.java
```diff
@@ -38,6 +38,9 @@ public abstract class ConditionalConfigurationRegistry {
     private final Map<String, Collection<Runnable>> pendingReachabilityHandlers = new ConcurrentHashMap<>();
 
     protected void registerConditionalConfiguration(ConfigurationCondition condition, Runnable runnable) {
+        if (runnable == null) {
+            throw new NullPointerException("Trying to register null runnable");
```
fniephaus, line 42:
```suggestion
            throw new NullPointerException("Cannot register null value as runnable for conditional configuration. Please ensure that all values you register are not null.");
```

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
```diff
@@ -46,7 +46,12 @@
 
 public interface ReflectionRegistry {
     default void register(ConfigurationCondition condition, Class<?>... classes) {
```
fniephaus, line 48: Should we also throw a NPE if `condition` is null?

zakkak: That's already handled in https://github.com/oracle/graal/pull/6985/files#diff-cfb00f555d030e4938d0ee92feb053eb8ce70f3c182b36b80c0241bdd412e8a1R41-R43

fniephaus: I'm not sure I see it. The first check seems `ConfigurationCondition.alwaysTrue().equals(condition)`, which is save but doesn't guarantee `condition` is not-null, no?

zakkak: Look at the highlighted part above the line you mention (the comment block makes things a bit harder).

image

zakkak: Oh.... Sorry that's not testing the condition :facepalm:

fniephaus: yes, that is just testing the runnable.

zakkak: Added https://github.com/oracle/graal/pull/6985/files#diff-cfb00f555d030e4938d0ee92feb053eb8ce70f3c182b36b80c0241bdd412e8a1R41-R44

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/impl/ReflectionRegistry.java
zakkak: I updated all messages following this pattern.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ConditionalConfigurationRegistry.java
zakkak: I updated all messages following this pattern.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/ReflectionRegistrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+        @SuppressWarnings("unused")
+        int unusedVariableTwo = 2;
+
+        @Override
+        public void beforeAnalysis(final BeforeAnalysisAccess access) {
+            try {
+                RuntimeReflection.register((Class<?>) null);
+                assert false;
+            } catch (NullPointerException e) {
+                // expected
```
fniephaus, line 57: Should we do this instead of the comments?
```suggestion
                assert e.getMessage().startsWith("Cannot use null value");
```

zakkak: +1 that would be even better.

zakkak: Done

fniephaus (approved):
LGTM

fniephaus:
Could you please fix the style gate?

zakkak:
> Could you please fix the style gate?

Done

vjovanov:
Thank you greatly for this fix. LGTM.

fniephaus:
I discussed this internally and changed the PR as follows:
- Lower the null-check into `register(ConfigurationCondition condition, boolean unsafeAllocated, Class<?> clazz)`. It was missing there.
- Move null-checks out of runnables and right to the beginning of the register methods.
- Apply the same strategy to `JNIAccessFeature`. It seems you, @zakkak, forgot about that.

zakkak:
> I discussed this internally and changed the PR as follows:

Sounds good, are you going to create a different PR or force push on this?

>     * Lower the null-check into `register(ConfigurationCondition condition, boolean unsafeAllocated, Class<?> clazz)`. It was missing there.

That makes sense, this way it won't cover only the public facing APIs.

>     * Move null-checks out of runnables and right to the beginning of the register methods.

Doesn't this result in iterating over the arrays twice? Once for the check and another for the runnables?

>     * Apply the same strategy to `JNIAccessFeature`. It seems you, @zakkak, forgot about that.

Indeed. Thanks for noticing and fixing this.

fniephaus:
> Doesn't this result in iterating over the arrays twice? Once for the check and another for the runnables?

Yes, but there are no guarantees when the runnables run. What if they run later on in some other thread?

zakkak:
> > Doesn't this result in iterating over the arrays twice? Once for the check and another for the runnables?
> 
> Yes, but there are no guarantees when the runnables run. What if they run later on in some other thread?

This is only a concern because we are using Streams instead of iterators (for loop), e.g. [here](https://github.com/oracle/graal/pull/6985/files#diff-b8e2db8c4bcc18db72364049b7ece5b148822706f95ee985470ab9570d58723dR49-R55), right?

Given that the arrays being iterated are not expected to be that big, what's the benefit of using streams?

fniephaus:
> Sounds good, are you going to create a different PR or force push on this?

I just push my changes manually.


> This is only a concern because we are using Streams instead of iterators (for loop), e.g. [here](https://github.com/oracle/graal/pull/6985/files#diff-b8e2db8c4bcc18db72364049b7ece5b148822706f95ee985470ab9570d58723dR49-R55), right?
> Given that the arrays being iterated are not expected to be that big, what's the benefit of using streams?

I'm not using streams for the null-checks. I didn't touch the other code you references.

zakkak:
> I just pushed my changes manually.

Looks good to me. Thank you Fabio.

Well runnables are inevitable in `registerConditionalConfiguration` either way. Sorry for the noise.

fniephaus:
> Well runnables are inevitable in registerConditionalConfiguration either way. Sorry for the noise.

What do you mean? For reachability handlers, some runnables may never run (if something did not get reachable). We should still fail if `null` values were registered, so that users know. In case the runnables to run later on, there is no way for users to understand which part of their code base registered a `null` class/method/fields, that's why we should fail as early as possible instead.

zakkak:
> > Well runnables are inevitable in registerConditionalConfiguration either way. Sorry for the noise.
> 
> What do you mean? For reachability handlers, some runnables may never run (if something did not get reachable). We should still fail if `null` values were registered, so that users know. In case the runnables to run later on, there is no way for users to understand which part of their code base registered a `null` class/method/fields, that's why we should fail as early as possible instead.

Yes, I agree. Sorry if it was not clear.

PR #7012, opened 2023-07-18
Add support for -XX:+HeapDumpOnOutOfMemoryError

### LINKED ISSUE #4641: [GR-38994] Add support for `-XX:+HeapDumpOnOutOfMemoryError`
**Feature request**

Following on the addition of heap dumps, support for `-XX:+HeapDumpOnOutOfMemoryError` should be added. This flag enables the heap dump to be dumped in the event of an out of memory error.

`-XX:+HeapDumpOnOutOfMemoryError` is a flag that is commonly used in Java programs. Now that heap dumps can be retrieved on demand, Java developers would expect this flag to be available. Heap dumps are needed the most when the JVM is running out of memory.

The alternative would be for a user to somehow be alerted when an OOME is reported and for the user to request a heap dump manually or via some automation.

I'd like to contribute this feature.

galderz:
Closes #4641.

Implements support for `-XX:+HeapDumpOnOutOfMemoryError` flag.

I've made some small changes to the `HeapDumpOperation` to be able to use it for this use case. The key thing is that all code in the operation has to be allocation free, which meant tweaking the error reporting method to not allocate. After discussing with @christianhaeubl, a new set of `open` and `create` methods have been added that take a `CCharPointer`. On startup, if `-XX:+HeapDumpOnOutOfMemoryError` is passed, the filename is pre-calculated and stored temporarily in case it's needed. Then, when OOME hits, the `CCharPointer` is used to create the file descriptor and pass it to the native VM operation.

To use `-XX:+HeapDumpOnOutOfMemoryError`, native image has to be built with `--enable-monitoring=heapdump`.

I will push the documentation updates shortly.

### DIFF
--- a/docs/reference-manual/native-image/guides/create-heap-dump-from-native-executable.md
+++ b/docs/reference-manual/native-image/guides/create-heap-dump-from-native-executable.md
@@ -9,12 +9,13 @@ permalink: /reference-manual/native-image/guides/create-heap-dump/
 
 You can create a heap dump of a running executable to monitor its execution. Just like any other Java heap dump, it can be opened with the [VisualVM](../../../tools/visualvm.md) tool.
 
-To enable heap dump support, native executables must be built with the `--enable-monitoring=heapdump` option. Heap dumps can then be created in three different ways:
+To enable heap dump support, native executables must be built with the `--enable-monitoring=heapdump` option. Heap dumps can then be created in different ways:
 
 1. Create heap dumps with VisualVM.
 2. Dump the initial heap of a native executable using the `-XX:+DumpHeapAndExit` command-line option.
 3. Create heap dumps sending a `SIGUSR1` signal at run time.
 4. Create heap dumps programmatically using the [`org.graalvm.nativeimage.VMRuntime#dumpHeap`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/VMInspectionOptions.java) API.
+5. If `-XX:+HeapDumpOnOutOfMemoryError` is passed in at runtime and an `OutOfMemoryError` is encountered, a heap dump will be produced.
 
 All approaches are described below.
 
@@ -274,6 +275,18 @@ The condition to create a heap dump is provided as an option on the command line
     The resulting heap dump can be then opened with the [VisualVM](../../../tools/visualvm.md) tool like any other Java heap dump, as illustrated below.
 
     ![Native Image Heap Dump View in VisualVM](img/heap-dump-api.png)
+
+## Create a Heap Dump on `OutOfMemoryError`
+
+Start the application with `-XX:+HeapDumpOnOutOfMemoryError` option to get a heap dump when an `OutOfMemoryError` occurs.
+The generated heap dump file name will have the `svm-heapdump-<PID>.hprof` naming format.
+For example:
+
+```shell
+./example -XX:+HeapDumpOnOutOfMemoryError
+Dumping heap to svm-heapdump-67799.hprof ...
+Exception in thread "main" java.lang.OutOfMemoryError: Garbage-collected heap size exceeded.
+```
 
 ### Related Documentation
 
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/VMRuntime.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/VMRuntime.java
@@ -92,6 +92,20 @@ public final class VMRuntime {
         ImageSingletons.lookup(HeapDumpSupport.class).dumpHeap(outputFile, live);
     }
 
+    /**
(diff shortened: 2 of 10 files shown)


galderz:
Pushed a couple of commits:
* Added log message similar to the HotSpot on when heap is dumped on OOME.
* Added documentation for feature.

This PR is ready for review now. @christianhaeubl can you have a look? @zakkak can you make this PR not a draft? I'm not able to do it, nor can I add reviewers...etc.

galderz:
By the way, the style errors reported are unrelated to the code changes in this PR.

zakkak:
> @zakkak can you make this PR not a draft? I'm not able to do it, nor can I add reviewers...etc.

I can't do it either, It says:

```
Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can mark a draft pull request as ready for review.
```

fniephaus:
unblocked the PR :)

galderz:
Thanks @fniephaus.

Having had a day to think about this, there's one doubt I have lingering, which is what to do if the heap dump file already exists. As it stands, any existing heap dumps would be overwritten, but I'm unsure if more than one OOME can be fired. In HotSpot you can have code that catches and ignores and so you'll get more OOMEs fired, but in SVM, an OOME does not bubble up all the way to the application. The main thread simply halts and the program finishes.

In any case, assuming multiple OOMEs can be fired, what HotSpot does is append a number to the file, but in SVM, the code that fires the OOME can't create new file names. So, you would have to pre-allocate a number of files in advance in case they're needed, which is a bit clunky. E.g. how many would you pre-allocate?

Side note: I've not been able to trigger HotSpot to generate multiple heap dump files on OOME, even after catching the OOME and continue trying to allocate objects. Multiple OOMEs are fired but only one dump file.

galderz:
Can someone add the redhat-interest label? @zakkak @roberttoyonaga can you do that?

fniephaus:
> In any case, assuming multiple OOMEs can be fired, what HotSpot does is append a number to the file, but in SVM, the code that fires the OOME can't create new file names. So, you would have to pre-allocate a number of files in advance in case they're needed, which is a bit clunky. E.g. how many would you pre-allocate?

Why can't the code that fires the OOME create new file names? If it's because the code is not allowed to allocate managed memory, I'm sure @christianhaeubl can show you how to do this with native/unmanaged memory. Or am I missing something?

> Side note: I've not been able to trigger HotSpot to generate multiple heap dump files on OOME, even after catching the OOME and continue trying to allocate objects. Multiple OOMEs are fired but only one dump file.

Me neither, but maybe possible if you're running multiple threads and they all see OOMEs? Regardless, I'd argue we should stay as close to the HotSpot impl as possible.

galderz:
@fniephaus Actually, I misread the HotSpot code. Appending a number to the heap dump file is only done when multiple heap dumps are produced in the case of VM errors other than OOME, e.g.

```
// Called by error reporting by a single Java thread outside of a JVM safepoint,
// or by heap dumping by the VM thread during a (GC) safepoint. Thus, these various
// callers are strictly serialized and guaranteed not to interfere below. For more
// general use, however, this method will need modification to prevent
// inteference when updating the static variables base_path and dump_file_seq below.
void HeapDumper::dump_heap() {
  HeapDumper::dump_heap(false);
}
```

The `false` there means the heap dump is not an OOME. This is something independent of this PR.

What is relevant to this PR, which I have only realised when I tried OOMEs with multiple threads, is that if multiple threads reach the same point, only one heap dump is produced:

```
void report_java_out_of_memory(const char* message) {
  static int out_of_memory_reported = 0;

  // A number of threads may attempt to report OutOfMemoryError at around the
  // same time. To avoid dumping the heap or executing the data collection
  // commands multiple times we just do it once when the first threads reports
  // the error.
  if (Atomic::cmpxchg(&out_of_memory_reported, 0, 1) == 0) {
    // create heap dump before OnOutOfMemoryError commands are executed
    if (HeapDumpOnOutOfMemoryError) {
      tty->print_cr("java.lang.OutOfMemoryError: %s", message);
      HeapDumper::dump_heap_from_oome();
    }

    if (OnOutOfMemoryError && OnOutOfMemoryError[0]) {
      VMError::report_java_out_of_memory(message);
    }

    if (CrashOnOutOfMemoryError) {
      tty->print_cr("Aborting due to java.lang.OutOfMemoryError: %s", message);
      report_fatal(OOM_JAVA_HEAP_FATAL, __FILE__, __LINE__, "OutOfMemory encountered: %s", message);
    }

    if (ExitOnOutOfMemoryError) {
      tty->print_cr("Terminating due to java.lang.OutOfMemoryError: %s", message);
      os::_exit(3); // quick exit with no cleanup hooks run
    }
  }
}
```

I need to add the necessary protection so that only one heap dump is produced. I'll add that next.

galderz:
Added support for only generating one heap dump on OOME. Also fixed the conflict.

christianhaeubl:
Thanks for the PR, I did a bunch of changes and fixes on top (see https://github.com/oracle/graal/pull/7088). Feel free to review and comment the changes that I pushed. It might take a few more days until this PR gets merged.

PR #7104, opened 2023-08-02
Do not expose canonicalized names for mapped relative directories in `fd_prestat_dir_name` so WASI SDK can resolve relative file names correctly.

nirvdrum:
Exposing the canonicalized names creates problems with `__wasilibc_find_abspath` in WASI SDK 20. That implementation has differing behavior based on whether a fully-qualified or a relative path is supplied. Canonicalization bypasses the relative path handling causing other relative paths to fail to resolve correctly.

Additionally, GraalWasm does various path prefix checks when opening files. The prefix check fails when opening a relative file path because the canonicalized directory name does not appear in the relative path name. Retaining the directory's originally mapped relative path name allows the prefix check to pass.


nirvdrum:
Ping @woess, in case Florian is no longer active on the project.

PR #7109, opened 2023-08-02
allow java.home config files to be included in the build output

kkriske:
set up basic support for copying files from the java.home directory to the build output


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/BuildArtifacts.java
```diff
@@ -47,6 +47,8 @@ enum ArtifactType {
         JDK_LIBRARY("jdk_libraries"),
         /* For all library shims for the JDK needed at run-time. */
         JDK_LIBRARY_SHIM(JDK_LIBRARY.getJsonKey()), // distinction should not be important to users.
+        /* For all configuration files from java.home needed at run-time */
+        JDK_CONFIG_FILE("jdk_config"),
```
fniephaus, line 51: If you extend this enum, you'll   also have to extend [this JSON schema](https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/assets/build-artifacts-schema-v0.9.0.json). Since `v0.9.0` is already release, you'll have to copy it and call it `v0.9.1`. Also, search for references to the file name and update the help text of the corresponding option. We should probably add a note about this to the enum's javadoc.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
```diff
@@ -0,0 +1,86 @@
...
+
+    JDKConfigFiles() {
+    }
+
+    public void register(String configFile) {
+        if (configFile.startsWith("/")) {
+            configFiles.add(configFile.substring(1));
+        } else {
+            configFiles.add(configFile);
+        }
```
fniephaus, line 59: How about this?
```suggestion
            configFiles.add(configFile.startsWith("/") ? configFile.substring(1) : configFile);
```
Also, shouldn't we use `File.separator` instead of `"/"` here?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
```diff
@@ -0,0 +1,86 @@
...
+@AutomaticallyRegisteredFeature
+final class JDKConfigFilesFeature implements InternalFeature {
+    @Override
+    public void afterRegistration(AfterRegistrationAccess access) {
+        ImageSingletons.add(JDKConfigFiles.class, new JDKConfigFiles());
+    }
+
+    @Override
+    public void afterImageWrite(AfterImageWriteAccess access) {
+        Path jdkHomeDir = Path.of(System.getProperty("java.home")).normalize();
```
fniephaus, line 72: `java.home` cannot be null, right?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
```diff
@@ -0,0 +1,86 @@
...
+        for (String configFile : new TreeSet<>(JDKConfigFiles.singleton().configFiles)) {
+            Path jdkConfigPath = jdkHomeDir.resolve(configFile).normalize();
+            VMError.guarantee(jdkConfigPath.startsWith(jdkHomeDir));
```
fniephaus, line 75: Doesn't `resolve().normalize()` take care of this? If you want to keep the guarantee, please add a descriptive message.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
```diff
@@ -0,0 +1,86 @@
...
+            Path imageConfigPath = access.getImagePath().resolveSibling(configFile);
+            try {
+                Files.createDirectories(imageConfigPath.getParent());
+                Files.copy(jdkConfigPath, imageConfigPath, REPLACE_EXISTING);
+                BuildArtifacts.singleton().add(BuildArtifacts.ArtifactType.JDK_CONFIG_FILE, imageConfigPath);
+            } catch (IOException e) {
+                VMError.shouldNotReachHere(e);
```
fniephaus, line 82: Please add a descriptive reason such as 
```suggestion
                VMError.shouldNotReachHere("Failed to copy JDK configuration files", e);
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
```diff
@@ -0,0 +1,86 @@
...
+            configFiles.add(configFile);
+        }
+    }
+}
+
+@AutomaticallyRegisteredFeature
+final class JDKConfigFilesFeature implements InternalFeature {
+    @Override
+    public void afterRegistration(AfterRegistrationAccess access) {
+        ImageSingletons.add(JDKConfigFiles.class, new JDKConfigFiles());
```
fniephaus, line 67: You could annotate `JDKConfigFiles` with `AutomaticallyRegisteredImageSingleton` and drop this manual registration if you like, but it can also stay as is.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
```diff
@@ -0,0 +1,86 @@
...
+        return ImageSingletons.lookup(JDKConfigFiles.class);
+    }
+
+    final Set<String> configFiles = Collections.newSetFromMap(new ConcurrentHashMap<>());
+
+    JDKConfigFiles() {
+    }
+
+    public void register(String configFile) {
+        if (configFile.startsWith("/")) {
```
fniephaus, line 55: I wonder whether it would make sense to check the existence of the configFile here, so that we can fail when an incorrect file is registered, rather than later before the image is being created.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
kkriske: The thing that's validated here is user input, so it all depends on how `JDKConfigFiles.singleton().register("")` is called.
In theory, removing the leading slash isn't required either if nobody passes one in.. So depends on how much input sanitisation you want to do.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
kkriske: I still need to validate this, but the idea was to protect against passing in `../../../someSystemFile`, which after `.normalize()` would no longer contain the `..` directories, but would point to a fille outside of the `java.home` dir. That is unless I'm missing something.
Since it's not actually exposes to end user (will it ever be?) this validation is maybe a bit overly cautious.

fniephaus: I agree that we probably don't need to be too cautious here as it's internal API. However, the validation again happens late in the process, I'd prefer if we validate such things as early as possible (in the `register()` hook in this case).

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
fniephaus: Right, how about changing this to something like `register(String... configPathParts)`, which can then be used like `register("lib", "bar", "my.file");`. Parts are not allowed to be `..` or to start with `/` or `\`. `register()` then joins the parts using the correct `File.separator` and validates that the joined path exists in `java.home`. If any of this is not the case, fail with an error.

fniephaus: You could even just use `Paths.get()` in `register()`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
kkriske: Makes sense, I moved the validation to the register call.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
kkriske: A vararg does make more sense indeed. I've modified it and changed the validation accordingly.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JDKConfigFiles.java
kkriske: I don't think it can be. What are you hinting at?

fniephaus: Just checking. As you know, java.home isn't always set in SVM 😉

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/BuildArtifacts.java
kkriske: example build artifacts file:
```json
{
  "executables": [
    "native-tests.exe.exe"
  ],
  "jdk_config": [
    "lib\\fontconfig.bfc"
  ],
  "jdk_libraries": [
    "awt.dll",
    "fontmanager.dll",
    "freetype.dll",
    "javaaccessbridge.dll",
    "javajpeg.dll",
    "jawt.dll",
    "lcms.dll",
    "java.dll",
    "jvm.dll"
  ]
}
```
I'm assuming the double backslash may have to be normalized?

Issue #7146, opened 2023-08-07
Building SubstrateVM fails after https://github.com/oracle/graal/pull/7102


zakkak:
**Describe the issue**
Since the merge of https://github.com/oracle/graal/pull/7102 building substrate VM with:

```
mx --native=native-image,lib:jvmcicompiler --components="Native Image,LibGraal" build
```

fails with:

```
...
Archiving NATIVE_IMAGE_BASE... [dependency com.oracle.svm.common updated]
Building Java module org.graalvm.nativeimage.base (native-image-base.jar) from NATIVE_IMAGE_BASE
Compiling com.oracle.truffle.host with javac-daemon(JDK 21)... [dependency TRUFFLE_DSL_PROCESSOR updated]
Error: Module org.graalvm.sdk not found, required by jdk.internal.vm.compiler.management
java.lang.module.FindException: Module org.graalvm.sdk not found, required by jdk.internal.vm.compiler.management
	at java.base/java.lang.module.Resolver.findFail(Resolver.java:892)
	at java.base/java.lang.module.Resolver.resolve(Resolver.java:192)
	at java.base/java.lang.module.Resolver.resolve(Resolver.java:141)
	at java.base/java.lang.module.Configuration.resolve(Configuration.java:420)
	at java.base/java.lang.module.Configuration.resolve(Configuration.java:254)
	at jdk.jlink/jdk.tools.jlink.internal.Jlink$JlinkConfiguration.resolve(Jlink.java:217)
	at jdk.jlink/jdk.tools.jlink.internal.JlinkTask.createImageProvider(JlinkTask.java:523)
	at jdk.jlink/jdk.tools.jlink.internal.JlinkTask.createImage(JlinkTask.java:411)
	at jdk.jlink/jdk.tools.jlink.internal.JlinkTask.run(JlinkTask.java:286)
	at jdk.jlink/jdk.tools.jlink.internal.Main.run(Main.java:56)
	at jdk.jlink/jdk.tools.jlink.internal.Main.main(Main.java:34)

Building graalvm-jimage: Failed due to error: 1
Note: Some input files use or override a deprecated API that is marked for removal.
Note: Recompile with -Xlint:removal for details.
```

This causes the Quarkus CI nightly runs to fail, see https://github.com/oracle/graal/actions/runs/5746129140/job/15575150905

**Steps to reproduce the issue**
1. git clone https://github.com/oracle/graal.git
2. cd graal/substratevm
[...]

PR #7147, opened 2023-08-07
The Graal compiler component depends on Graal SDK as well

zakkak:
See the `distDependencies` of the GRAAL distribution in mx.compiler/suite.py

Closes https://github.com/oracle/graal/issues/7146


chumer:
Thanks for the report. I will merge a fix this week that will take care of this, but that would conflict with your change.

PR #7153, opened 2023-08-08
[GR-43849] Using devtoolset version 10 for linux_aarch64 jobs.

graalvmbot:


petermz:
May I ask what the reason for this change is? Isn't it possible to use the same devtoolset for both?

iamstolis:
> May I ask what the reason for this change is? Isn't it possible to use the same devtoolset for both?

We do not have devtoolset version 11 for `linux_aarch64`. In other words, this PR is just a fix of the previous PR that requested devtoolset version 11 unconditionally. This request could not be satisfied on `linux_aarch64` in our CI. Some default/old devtoolset was used as a consequence. This old devtoolset was not able to build some stuff (like the latest version of `graal-nodejs`).

PR #7158, opened 2023-08-08
[GR-47755] Refactor dwarf constants.

graalvmbot:


koutheir:
@adinn, the constant `DW_CFA_CIE_version` is defined as `1`. According to DWARF standards version 2 through 5, the value that should be specified is never `1`. Is this a bug? If so, which CIE version are you generating in GraalVM?

| DWARF version | CIE version number |
| --- | --- |
| 2.0.0 | 2 |
| 3 | 3 |
| 4 | 4 |
| 5 | 4 |

koutheir:
@adinn, given that GraalVM emits the fields `address_size` and `segment_size` of the CIE, this version must be 4. DWARF versions 3 and below did not define those fields.

koutheir:
@adinn, given that GraalVM does not emit the fields `address_size` and `segment_size` of the CIE, this version must be 2 or 3.

Am I missing something?

adinn:
@koutheir

I believe that the frame section generates DWARF that is compatible with DWARF version 2 (which, I think, is what you were suggesting).

Section 7.23 of the [DWARF 2 standard](https://dwarfstd.org/doc/dwarf-2.0.0.pdf) states that the initial version is 1, that being the version appropriate to the current standard (i.e. DWARF 2).

Likewise, Appendix F of the [DWARF 4 standard](https://dwarfstd.org/doc/DWARF4.pdf) states that the correct version for for DWARF 2 content is 1.

So, I don't think we are generating invalid frame content.

However it is arguable that the enum should not actually specify `DW_CFA_CIE_version` as 1 since version values 3 and 4 are also legitimate.

In other cases where the generator has made a choice for a version or some other range value it has used a local definition rather than an enum (see for example `DwarfDebugInfo.LANG_ENCODING`).

It might be better to define enum tags providing the encoding appropriate to each DWARF level e.g.
```
DW_CFA_CIE_version_DWARF_2(byte)1)
DW_CFA_CIE_version_DWARF_3(byte)3)
DW_CFA_CIE_version_DWARF_4(byte)4)
```

If you raise an issue for that I will be happy to fix it.

koutheir:
You're right, I confused the CIE version with another version in the standard text of version 2.0.0. I'm sorry.

PR #7204, opened 2023-08-14
Correct error introduced into debug info log trace

adinn:
This PR fixes a breakage that was introduced into the debug info generator's message logging routines. It's a trivial fix to method formatting code.

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
@@ -181,7 +181,7 @@ public class DwarfInfoSectionImpl extends DwarfSectionImpl {
         pos = writeCUHeader(buffer, pos);
         assert pos == lengthPos + DIE_HEADER_SIZE;
         AbbrevCode abbrevCode = AbbrevCode.BUILTIN_UNIT;
-        log(context, "  [0x%08x] <0> Abbrev Number %d", pos, abbrevCode);
+        log(context, "  [0x%08x] <0> Abbrev Number %d", pos, abbrevCode.ordinal());
         pos = writeAbbrevCode(abbrevCode, buffer, pos);
         log(context, "  [0x%08x]     language  %s", pos, "DW_LANG_Java");
         pos = writeAttrLanguage(DwarfDebugInfo.LANG_ENCODING, buffer, pos);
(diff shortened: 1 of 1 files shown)


adinn:
@fniephaus Could you push this through the Oracle gate ASAP. Debugging the generator code without being able to log messages is a pain.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/DwarfInfoSectionImpl.java
```diff
@@ -181,7 +181,7 @@ private int writeBuiltInTypes(DebugContext context, byte[] buffer, int p) {
         pos = writeCUHeader(buffer, pos);
         assert pos == lengthPos + DIE_HEADER_SIZE;
         AbbrevCode abbrevCode = AbbrevCode.BUILTIN_UNIT;
-        log(context, "  [0x%08x] <0> Abbrev Number %d", pos, abbrevCode);
+        log(context, "  [0x%08x] <0> Abbrev Number %d", pos, abbrevCode.ordinal());
         pos = writeAbbrevCode(abbrevCode, buffer, pos);
```
fniephaus, line 185: Looking at the diff, shouldn't the log statement be part of `writeAbbrevCode()`?

adinn: Perhaps. However, this is part of a larger pattern where the writeXXX and putXXX operations do the serialization into the buffer and the caller determines whether/how it is logged. That's because logging may involve reporting more data than just the value being written (e.g. sometimes you want both a numeric value and a string representation) and may need to use a format specific to the calling context.

In this case the log message is meant look like the output of an `objdump` of the `.debug_info` section -- that helps to spot differences between what the generator thinks it is writing and what objdump reports (n.b. that's why it writes the actual code, not the enum tag name). Keeping the log messages in each of the functions that writes a specific segment of the info data helps ensure that the log statements actually match up with the format used by objdump.

Some or all of the logging logic could probably be folded into the `writeXXX` methods. However that would make them more complicated than they really need to be for their primary purpose, which is to serialize data into the buffer (or to size it when the buffer is null). This would be especially so for writeXXX methods are called from a variety of different callers.

I can rework this across the code base if you want but I'd rather do that as a separate PR as not having this fix is making other development tricky.

fniephaus (approved):

olpaw (approved):

fniephaus:
Merged as part of https://github.com/oracle/graal/pull/7218.

PR #7205, opened 2023-08-14
[GR-48144] Fix native-image build with --no-jlinking

### LINKED ISSUE #7085: GR-47186 "Fix native-image --add-modules" breaks `--no-jlinking` builds
**Describe the issue**
https://github.com/oracle/graal/pull/7076 breaks builds with `--no-jlinking` (which includes Mandrel builds https://github.com/graalvm/mandrel/issues/554).

Building with `--no-jlinking` results in the following error:

```
Exception in thread "main" java.lang.IllegalArgumentException: Missing scheme
	at java.base/java.nio.file.Path.of(Path.java:200)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.callListModules(NativeImage.java:1616)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.listModulesFromPath(NativeImage.java:1585)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.buildImage(NativeImage.java:1449)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.completeImageBuild(NativeImage.java:1196)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.build(NativeImage.java:1772)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.performBuild(NativeImage.java:1737)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.main(NativeImage.java:1711)
Exception in thread "main" java.lang.IllegalArgumentException: Missing scheme
	at java.base/java.nio.file.Path.of(Path.java:200)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.callListModules(NativeImage.java:1616)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.listModulesFromPath(NativeImage.java:1585)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.buildImage(NativeImage.java:1449)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.completeImageBuild(NativeImage.java:1196)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.build(NativeImage.java:1772)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.performBuild(NativeImage.java:1737)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.main(NativeImage.java:1711)
```

Further inspection reveals that this happens because when using the `--no-jlinking` option the `--list-modules` command fails with:

```
$ /home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/bin/java -p /home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/lib/jvmci/graal-sdk.jar:/home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/lib/truffle/truffle-api.jar:/home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/lib/truffle/truffle-compiler.jar:/home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/lib/truffle/truffle-runtime.jar:/home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/lib/svm/builder/svm.jar:/home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/lib/svm/builder/objectfile.jar:/home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/lib/svm/builder/pointsto.jar:/home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/lib/svm/builder/native-image-base.jar:/home/zakkak/code/graal/sdk/mxbuild/linux-amd64/GRAALVM_73636D9934_JAVA21_STAGE1/graalvm-73636d9934-java21-23.1.0-dev/lib/svm/library-support.jar --list-modules

Error occurred during initialization of boot layer
java.lang.module.FindException: Module org.graalvm.jniutils not found, required by org.graalvm.nativeimage.builder
```

```
 mx --primary-suite-path substratevm --no-jlinking --components="Native Image,LibGraal" --native-images=native-image,lib:jvmcicompiler build
```

 - GraalVM version: **70b50a854f615a9e9388de46895bcda92b4dfc0a**
 - JDK major version: **21-jvmci-23.1-b10**
 - OS: **Fedora 38**
 - Architecture: **AMD64**

cc @olpaw

jerboaa:
In #7171 `graal-sdk` got split into several parts. This patch accounts for that and adds
the new dependencies to the module path when the JDKs `jimage` is being queried for built-in modules.

`--add-modules org.graalvm.polyglot` has been dropped from the native-image launcher. It
gets added to the module path on demand if needed.

Closes #7085

### DIFF
--- a/sdk/mx.sdk/mx_sdk_vm_impl.py
+++ b/sdk/mx.sdk/mx_sdk_vm_impl.py
@@ -2232,13 +2232,9 @@ class GraalVmBashLauncherBuildTask(GraalVmNativeImageBuildTask):
             extra_jvm_args = mx.list_to_cmd_line(image_config.extra_jvm_args)
             if not _jlink_libraries():
                 if mx.is_windows():
-                    extra_jvm_args = ' '.join([extra_jvm_args, r'--upgrade-module-path "%location%\..\..\jvmci\graal.jar"',
-                                               r'--add-modules org.graalvm.polyglot',
-                                               r'--module-path "%location%\..\..\truffle\truffle-api.jar:%location%\..\..\jvmci\polyglot.jar"'])
+                    extra_jvm_args = ' '.join([extra_jvm_args, r'--upgrade-module-path "%location%\..\..\jvmci\graal.jar"'])
                 else:
-                    extra_jvm_args = ' '.join([extra_jvm_args, '--upgrade-module-path "${location}/../../jvmci/graal.jar"',
-                                               '--add-modules org.graalvm.polyglot',
-                                               '--module-path "${location}/../../truffle/truffle-api.jar:${location}/../../jvmci/polyglot.jar"'])
+                    extra_jvm_args = ' '.join([extra_jvm_args, '--upgrade-module-path "${location}/../../jvmci/graal.jar"'])
             return extra_jvm_args
 
         def _get_option_vars():
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -576,17 +576,27 @@ public class NativeImage {
         }
 
         private List<Path> createTruffleBuilderModulePath() {
-            List<Path> jars = getJars(rootDir.resolve(Paths.get("lib", "truffle")), "truffle-api", "truffle-runtime", "truffle-enterprise");
+            Path libTruffleDir = rootDir.resolve(Paths.get("lib", "truffle"));
+            List<Path> jars = getJars(libTruffleDir, "truffle-api", "truffle-runtime", "truffle-enterprise");
             if (!jars.isEmpty()) {
                 /*
                  * If Truffle is installed as part of the JDK we always add the builder modules of
                  * Truffle to the builder module path. This is legacy support and should in the
                  * future no longer be needed.
                  */
-                jars.addAll(getJars(rootDir.resolve(Paths.get("lib", "truffle")), "truffle-compiler"));
+                jars.addAll(getJars(libTruffleDir, "truffle-compiler"));
                 Path builderPath = rootDir.resolve(Paths.get("lib", "truffle", "builder"));
                 if (Files.exists(builderPath)) {
                     jars.addAll(getJars(builderPath, "truffle-runtime-svm", "truffle-enterprise-svm"));
+                    if (libJvmciDir != null) {
+                        // truffle-runtime-svm depends on polyglot, which is not part of non-jlinked
+                        // JDKs
+                        jars.addAll(getJars(libJvmciDir, "polyglot"));
+                    }
+                }
+                if (libJvmciDir != null) {
+                    // truffle-runtime depends on polyglot, which is not part of non-jlinked JDKs
+                    jars.addAll(getJars(libTruffleDir, "jniutils"));
                 }
             }
             /*
@@ -594,7 +604,7 @@ public class NativeImage {
              * image builder still needs it
              */
             if (libJvmciDir != null) {
-                jars.addAll(getJars(rootDir.resolve(Paths.get("lib", "truffle")), "truffle-compiler"));
+                jars.addAll(getJars(libTruffleDir, "truffle-compiler"));
             }
 
             return jars;
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -587,8 +587,9 @@ public class NativeImage {
                 jars.addAll(getJars(libTruffleDir, "truffle-compiler"));
                 Path builderPath = rootDir.resolve(Paths.get("lib", "truffle", "builder"));
                 if (Files.exists(builderPath)) {
-                    jars.addAll(getJars(builderPath, "truffle-runtime-svm", "truffle-enterprise-svm"));
-                    if (libJvmciDir != null) {
+                    List<Path> truffleRuntimeSVMJars = getJars(builderPath, "truffle-runtime-svm", "truffle-enterprise-svm");
+                    jars.addAll(truffleRuntimeSVMJars);
+                    if (libJvmciDir != null && !truffleRuntimeSVMJars.isEmpty()) {
                         // truffle-runtime-svm depends on polyglot, which is not part of non-jlinked
                         // JDKs
                         jars.addAll(getJars(libJvmciDir, "polyglot"));
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -561,10 +561,12 @@ public class NativeImage {
          */
         public List<Path> getBuilderModulePath() {
             List<Path> result = new ArrayList<>();
-            // Non-jlinked JDKs need truffle and graal-sdk on the module path since they
-            // don't have those modules as part of the JDK.
+            // Non-jlinked JDKs need truffle and word, collections, nativeimage on the
+            // module path since they don't have those modules as part of the JDK. Note
+            // that graal-sdk is now obsolete after the split in GR-43819 (#7171)
             if (libJvmciDir != null) {
-                result.addAll(getJars(libJvmciDir, "graal-sdk", "enterprise-graal"));
+                result.addAll(getJars(libJvmciDir, "enterprise-graal"));
+                result.addAll(getJars(libJvmciDir, "word", "collections", "nativeimage"));
             }
             if (modulePathBuild) {
                 result.addAll(createTruffleBuilderModulePath());
@@ -586,6 +588,13 @@ public class NativeImage {
                 if (Files.exists(builderPath)) {
                     jars.addAll(getJars(builderPath, "truffle-runtime-svm", "truffle-enterprise-svm"));
                 }
+            }
+            /*
+             * Non-Jlinked JDKs don't have truffle-compiler as part of the JDK, however the native
+             * image builder still needs it
+             */
+            if (libJvmciDir != null) {
+                jars.addAll(getJars(rootDir.resolve(Paths.get("lib", "truffle")), "truffle-compiler"));
             }
 
             return jars;


substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -556,9 +556,12 @@ public List<String> getBuilderJavaArgs() {
         public List<Path> getBuilderModulePath() {
             List<Path> result = new ArrayList<>();
             // Non-jlinked JDKs need truffle and graal-sdk on the module path since they
-            // don't have those modules as part of the JDK.
+            // don't have those modules as part of the JDK. They also need
+            // word/collection/nativeimage modules on the module path after the split in
+            // GR-43819 (#7171)
             if (libJvmciDir != null) {
                 result.addAll(getJars(libJvmciDir, "graal-sdk", "enterprise-graal"));
```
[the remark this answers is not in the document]
jerboaa: OK, sure.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -556,9 +556,12 @@ public List<String> getBuilderJavaArgs() {
...
+                result.addAll(getJars(libJvmciDir, "word", "collections", "nativeimage"));
```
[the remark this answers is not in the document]
jerboaa: I'm unsure about this one, we need some magic to get `polyglot.jar` copied to the dists dir:

```
Building native-image.image... [dependency SVM_DRIVER updated]
Error occurred during initialization of boot layer
java.lang.module.FindException: Module org.graalvm.polyglot not found

Building native-image.image: Failed due to error: 1
```

```
$ find sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/graal.src.zip
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/collections.jar
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/collections.src.zip
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/nativeimage.src.zip
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/word.jar
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/graal.jar
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/graal-management.src.zip
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/graal-management.jar
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/word.src.zip
sdk/mxbuild/linux-amd64/GRAALVM_44571D84E8_JAVA21_STAGE1/graalvm-44571d84e8-java21-23.1.0-dev/lib/jvmci/nativeimage.jar
```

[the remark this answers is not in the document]
zakkak: What's the command you are using to build GraalVM in this case? If `polyglot` is not part of the dependencies then not copying `polyglot.jar` over is correct.

E.g.

```
mx --primary-suite-path substratevm --no-jlinking --components="Native Image,LibGraal" --native-images=native-image,lib:jvmcicompiler build
```

Shouldn't copy `polyglot.jar`.

On the other hand:

```
mx --primary-suite-path substratevm --no-jlinking --components="Native Image,LibGraal,poly" --native-images=native-image,lib:jvmcicompiler build
```

Should probably copy it over.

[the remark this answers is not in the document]
zakkak: > Any idea on how to achieve that?

I am not sure what you refer to with _that_.

I think the reason you are getting:

```
Building native-image.image... [dependency SVM_DRIVER updated]
Error occurred during initialization of boot layer
java.lang.module.FindException: Module org.graalvm.polyglot not found
```

Is because we explicitly pass `--add-modules org.graalvm.polyglot` to the command line despite `polyglot.jar` not being present (or needed).

Are you getting this when you run `getJars(libJvmciDir, "polyglot"` as well?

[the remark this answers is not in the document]
jerboaa: "That" being making `mx --primary-suite-path substratevm --no-jlinking --components="Native Image,LibGraal" --native-images=native-image,lib:jvmcicompiler build` pass. It looks like the build invokes the native-image launcher macro, which in-turns fails to boot because of the missing `polyglot.jar` getting copied to the sdk's image directory and the launcher script referencing it using `--add-modules`.

> Are you getting this when you run `getJars(libJvmciDir, "polyglot"` as well?

Yes.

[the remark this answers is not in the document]
zakkak: > and the launcher script referencing it using `--add-modules`.

Exactly, that's the issue. That's why I believe we should remove it from the bash/native launcher and have `NativeImage` add it only when necessary as I think the launcher doesn't really need it...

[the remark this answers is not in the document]
jerboaa: OK. I don't know why `--add-modules ...polyglot` was added and have, therefore, no way of testing it. Do you have ideas, suggestions what executes that codepath?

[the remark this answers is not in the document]
zakkak: Unfortunately not, @chumer could you please shed some light on this?

zakkak (commented):
@jerboaa I was able to get rid of the `polyglot.jar` and `truffle-api.jar` in the launcher with https://github.com/zakkak/mandrel/commit/85088b19f8062246d5f0a95c83ace120ee034c25 (on top of your branch)

To test it I used:

```
~/code/mx/mx --primary-suite vm --no-jlinking --env ce build
```

which is supposed to build GraalVM CE (including polyglot and truffle):

```
❯ ~/code/mx/mx --primary-suite vm --no-jlinking --env ce graalvm-show             
GraalVM distribution: GRAALVM_COMMUNITY_JAVA21
Version: 23.1.0-dev
Config name: community
Components:
 - GraalVM compiler ('cmp', /graal, experimental)
 - GraalVM Coverage ('cov', /coverage, experimental)
 - GraalVM Debug Protocol Server ('dap', /dap, experimental)
 - Component installer ('gu', /installer, experimental)
 - GraalVM license files ('gvm', /., experimental)
 - GraalVM Chrome Inspector ('ins', /chromeinspector, experimental)
 - Insight ('insight', /insight, experimental)
 - Insight Heap ('insightheap', /insightheap, experimental)
 - LibGraal ('lg', /False, experimental)
 - GraalVM Language Server ('lsp', /lsp, experimental)
 - Truffle NFI LIBFFI ('nfi-libffi', /nfi, experimental)
 - Native Image ('ni', /svm, experimental)
 - Native Image Configure Tool ('nic', /svm, experimental)
 - Native Image licence files ('nil', /svm, experimental)
 - Polyglot Launcher ('poly', /polyglot, experimental)
 - Polyglot Native API ('polynative', /polyglot, experimental)
 - GraalVM Profiler ('pro', /profiler, experimental)
 - Graal SDK ('sdk', /graalvm, experimental)
 - GraalVM Launcher Common ('sdkl', /graalvm, experimental)
 - SubstrateVM ('svm', /svm, experimental)
 - Truffle Runtime SVM ('svmt', /truffle, experimental)
 - SVM Truffle NFI Support ('svmnfi', /nfi, experimental)
 - SubstrateVM Static Libraries ('svmsl', /False, experimental)
 - Truffle ('tfl', /truffle, experimental)
 - Truffle API ('tfla', /truffle, experimental)
 - Truffle Compiler ('tflc', /truffle, experimental)
 - Truffle Macro ('tflm', /truffle, experimental)
 - Truffle JSON Library ('truffle-json', /truffle-json, experimental)
 - Graal SDK Compiler ('sdkc', /graalvm, experimental)
 - Truffle NFI ('nfi', /nfi, experimental)
 - SubstrateVM Foreign API Preview Feature ('svmforeign', /svm-preview, experimental)
 - Graal SDK Native Image ('sdkni', /graalvm, experimental)
 - Polyglot Library ('libpoly', /polyglot, experimental)
Launchers:
 - gu (bash, rebuildable)
 - native-image (native, rebuildable)
 - native-image-configure (bash, rebuildable)
 - polyglot (bash, rebuildable)
Libraries:
 - libjvmcicompiler.so (native, non-rebuildable)
 - libnative-image-agent.so (native, rebuildable)
 - libnative-image-diagnostics-agent.so (native, rebuildable)
 - libpolyglot.so (skipped, rebuildable)
Installables:
 - NATIVE_IMAGE_INSTALLABLE_SVM_JAVA21
No standalone
```

Please have a look.

jerboaa:
Fantastic! Let me merge that in and I'll test myself as well.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1494,7 +1513,7 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> cp, LinkedHa
         Function<Path, Path> substituteModulePath = useBundle() ? bundleSupport::substituteModulePath : Function.identity();
         List<Path> substitutedImageModulePath = imagemp.stream().map(substituteModulePath).toList();
 
-        Map<String, Path> modules = listModulesFromPath(javaExecutable, Stream.concat(mp.stream(), imagemp.stream()).distinct().toList());
+        Map<String, Path> modules = listModulesFromPath(javaExecutable, Stream.concat(mp.stream(), imagemp.stream()).distinct().toList(), upgradeModulePath);
```
olpaw, line 1516: How about we just take the passed in `javaArgs` that we are also using for the builder. There we already have the `upgradeModulePath` you are seeking for.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1621,16 +1640,16 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> cp, LinkedHa
     /**
      * Resolves and lists all modules given a module path.
      *
-     * @see #callListModules(String, List)
+     * @see #callListModules(String, List, String)
      */
-    private Map<String, Path> listModulesFromPath(String javaExecutable, Collection<Path> modulePath) {
+    private Map<String, Path> listModulesFromPath(String javaExecutable, Collection<Path> modulePath, String upgradeModulePath) {
```
olpaw, line 1645: Then you would modify this to pass in `List<String> javaArgs` as the second parameter.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1641,13 +1660,19 @@ private Map<String, Path> listModulesFromPath(String javaExecutable, Collection<
      * <p>
      * This is a much more robust solution then trying to parse the JDK file structure manually.
      */
-    private static Map<String, Path> callListModules(String javaExecutable, List<String> arguments) {
+    private static Map<String, Path> callListModules(String javaExecutable, List<String> arguments, String upgradeModulePath) {
```
olpaw, line 1663: Also here

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1641,13 +1660,19 @@ private Map<String, Path> listModulesFromPath(String javaExecutable, Collection<
...
         Process listModulesProcess = null;
         Map<String, Path> result = new LinkedHashMap<>();
         try {
             var pb = new ProcessBuilder(javaExecutable);
             pb.command().addAll(arguments);
```
olpaw, line 1668: add add them here before the actual `arguments`.

olpaw (changes requested):

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
jerboaa: OK, sure. That passes in a whole lot of other things, but it seems to work. Thanks!

jerboaa:
Thanks for the review, @olpaw! Please take another look.

olpaw (commented):
The changes look good now. But there is still a failing gate.

jerboaa:
> The changes look good now. But there is still a failing gate.

@olpaw The gate failures seem to be caused by passing in `-XX:StartFlightRecording=dumponexit=true` to the `native-image` command, which results in the `java --list-modules` call to have that option too. The failing command is:

```
/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/bin/native-image --dry-run --verbose --no-fallback -H:+UnlockExperimentalVMOptions -H:+EnforceMaxRuntimeCompileMethods -H:+ReportExceptionStackTraces -H:Path=/disk/graal/upstream-sources/graal/substratevm/svmbuild -H:-UnlockExperimentalVMOptions --verbose -J-ea -J-esa --native-image-info -o /disk/graal/upstream-sources/graal/substratevm/svmbuild/helloworld -H:+UnlockExperimentalVMOptions -H:+VerifyNamingConventions -H:-UnlockExperimentalVMOptions -cp /disk/graal/upstream-sources/graal/substratevm/svmbuild HelloWorld -J-XX:StartFlightRecording=dumponexit=true -Dllvm.bin.dir=/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/LLVM_TOOLCHAIN/bin/
Apply jar:file:///disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/library-support.jar!/META-INF/native-image/com.oracle.svm/polyglot/native-image.properties
Apply jar:file:///disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/library-support.jar!/META-INF/native-image/com.oracle.svm/thirdparty/native-image.properties
Exception in thread "main" java.lang.IllegalArgumentException: Missing scheme
	at java.base/java.nio.file.Path.of(Path.java:200)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.callListModules(NativeImage.java:1688)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.listModulesFromPath(NativeImage.java:1652)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.buildImage(NativeImage.java:1516)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.completeImageBuild(NativeImage.java:1263)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.build(NativeImage.java:1839)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.performBuild(NativeImage.java:1804)
	at org.graalvm.nativeimage.driver/com.oracle.svm.driver.NativeImage.main(NativeImage.java:1778)
```

When I add `-XX:StartFlightRecording=dumponexit=true` option to `--list-modules`, there is unified logging output from `jfr+startup` which is what `--list-modules` parsing throws off. Example:

```
$ java -XX:StartFlightRecording=dumponexit=true --list-modules
[0.365s][info][jfr,startup] Started recording 1. No limit specified, using maxsize=250MB as default.
[0.365s][info][jfr,startup] 
[0.365s][info][jfr,startup] Use jcmd 81612 JFR.dump name=1 filename=FILEPATH to copy recording data to file.
com.oracle.graal.graal_enterprise
com.oracle.svm.enterprise.truffle
com.oracle.svm.extraimage_enterprise
com.oracle.svm.svm_enterprise
com.oracle.svm_enterprise.ml_dataset
com.oracle.truffle.enterprise
com.oracle.truffle.enterprise.svm
java.base@21
java.compiler@21
java.datatransfer@21
java.desktop@21
java.instrument@21
java.logging@21
java.management@21
java.management.rmi@21
java.naming@21
java.net.http@21
java.prefs@21
java.rmi@21
java.scripting@21
java.se@21
java.security.jgss@21
java.security.sasl@21
java.smartcardio@21
java.sql@21
java.sql.rowset@21
java.transaction.xa@21
java.xml@21
java.xml.crypto@21
jdk.accessibility@21
jdk.attach@21
jdk.charsets@21
jdk.compiler@21
jdk.crypto.cryptoki@21
jdk.crypto.ec@21
jdk.dynalink@21
jdk.editpad@21
jdk.hotspot.agent@21
jdk.httpserver@21
jdk.incubator.vector@21
jdk.internal.ed@21
jdk.internal.jvmstat@21
jdk.internal.le@21
jdk.internal.opt@21
jdk.internal.vm.ci@21
jdk.internal.vm.compiler
jdk.internal.vm.compiler.management
jdk.jartool@21
jdk.javadoc@21
jdk.jcmd@21
jdk.jconsole@21
jdk.jdeps@21
jdk.jdi@21
jdk.jdwp.agent@21
jdk.jfr@21
jdk.jlink@21
jdk.jpackage@21
jdk.jshell@21
jdk.jsobject@21
jdk.jstatd@21
jdk.localedata@21
jdk.management@21
jdk.management.agent@21
jdk.management.jfr@21
jdk.naming.dns@21
jdk.naming.rmi@21
jdk.net@21
jdk.nio.mapmode@21
jdk.random@21
jdk.sctp@21
jdk.security.auth@21
jdk.security.jgss@21
jdk.unsupported@21
jdk.unsupported.desktop@21
jdk.xml.dom@21
jdk.zipfs@21
org.graalvm.collections
org.graalvm.extraimage.builder
org.graalvm.jniutils
org.graalvm.locator
org.graalvm.nativeimage
org.graalvm.polyglot
org.graalvm.sdk
org.graalvm.truffle
org.graalvm.truffle.compiler
org.graalvm.truffle.runtime
org.graalvm.word
```

This didn't fail with the earlier patch as not all options got passed whole-sale to the `list-modules` call. So there are a couple of options how to fix it:

1. Silence unified logging for the `-XX:StartFlightRecording=dumponexit=true` test with `-Xlog:jfr\*=off`
2. Go back to the previous version of the patch (only passing `--upgrade-module`  option when needed)
3. Fix `callListModules` routine to ignore such unexpected output.

Which one should it be?

olpaw:
>  1.   Silence unified logging for the -XX:StartFlightRecording=dumponexit=true test with -Xlog:jfr\*=off
>  2.  Go back to the previous version of the patch (only passing --upgrade-module option when needed)
>  3.  Fix callListModules routine to ignore such unexpected output.
> Which one should it be?

Pleas go for `3`.

IMO, `2.` is not an option because there might be `javaArgs` used by the builder that influence the output of `--list-modules`. E.g. `--patch-module`, `--enable-preview` ... we want our `--list-modules` query to run in the same JVM context that be builder is also running afterwards.

olpaw (approved):

jerboaa:
Thanks for the review! Let me know if there is anything more that you need from my end.

olpaw:
> Running in internal CI ... ⏳

This found issues with the current approach.

We are regressing on
```
mx helloworld -J--limit-modules=org.graalvm.nativeimage.builder
```
On master this works fine but on this PR we get:
```
GraalVM Native Image: Generating 'helloworld' (executable)...
[1/8] Initializing...
                                                                                    (0.0s @ 0.08GB)
Error: Feature com.oracle.svm.thirdparty.gson.GsonFeature class not found on the classpath. Ensure that the name is correct and that the class is on the classpath.
com.oracle.svm.core.util.UserError$UserException: Feature com.oracle.svm.thirdparty.gson.GsonFeature class not found on the classpath. Ensure that the name is correct and that the class is on the classpath.
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.util.UserError.abort(UserError.java:73)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.FeatureHandler.registerFeatures(FeatureHandler.java:183)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.setupNativeImage(NativeImageGenerator.java:897)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:591)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:551)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:528)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:710)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.start(NativeImageGeneratorRunner.java:142)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:97)
                         0.1s (5.3% of total time) in 6 GCs | Peak RSS: 0.47GB | CPU load: 5.73
Finished generating 'helloworld' in 0.9s.
```

The reason is that the `-J--limit-modules=org.graalvm.nativeimage.builder` constrains the builder down to the transitive closure of the module org.graalvm.nativeimage.builder (itself + all modules it depends on).

But luckily we do have our `--list-modules` option implemented in the builder. So instead of using the  `--list-modules` option implemented in `java` use our version, then `-J--limit-modules=org.graalvm.nativeimage.builder` cannot hurt the `org.graalvm.nativeimage.librarysupport` module the the builder requires (it contains that `com.oracle.svm.thirdparty.gson.GsonFeature`). Run
```
mx native-image -J--limit-modules=org.graalvm.nativeimage.builder --list-modules
```
to see what I mean (and compare it to the output of `callListModules`.

jerboaa:
Thanks for the update. I'll look into it.

jerboaa:
Hmm, `mx --primary-suite-path substratevm helloworld -J--limit-modules=org.graalvm.nativeimage.builder` produces for me:

```
$ mx --primary-suite-path substratevm helloworld -J--limit-modules=org.graalvm.nativeimage.builder
GraalVM Native Image: Generating 'helloworld' (executable)...
[1/8] Initializing...
                                                                                    (0.0s @ 0.07GB)
Error: Feature com.oracle.svm.polyglot.groovy.GroovyIndyInterfaceFeature class not found on the classpath. Ensure that the name is correct and that the class is on the classpath.
com.oracle.svm.core.util.UserError$UserException: Feature com.oracle.svm.polyglot.groovy.GroovyIndyInterfaceFeature class not found on the classpath. Ensure that the name is correct and that the class is on the classpath.
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.util.UserError.abort(UserError.java:73)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.FeatureHandler.registerFeatures(FeatureHandler.java:183)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.setupNativeImage(NativeImageGenerator.java:897)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:591)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:551)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:528)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:710)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.start(NativeImageGeneratorRunner.java:142)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:97)
                         0.1s (5.9% of total time) in 6 GCs | Peak RSS: 0.44GB | CPU load: 4.48
Finished generating 'helloworld' in 1.6s.

```

Is that still the right reproducer?

jerboaa:
Adding the `-J--limit-modules=org.graalvm.nativeimage.builder` parameter, seems to cause the driver to omit `-imagemp <path/to/library-support.jar>` option to the `NativeImageGeneratorRunner`. Investigating why...

olpaw:
> As already hinted, using the --list-modules of the native-image builder itself does not suffer from that problem.

You will have to use the raw-variant `-H:+ListModules` instead of the `--list-modules` when you uase it as `imageArgs`.

jerboaa:
Thanks, Paul!

jerboaa:
> As already hinted, using the `--list-modules` of the native-image builder itself does not suffer from that problem.

What's really confusing is that if that is the case shouldn't this diff produce more output (specifically showing librarysupport - `org.graalvm.nativeimage.librarysupport` in the diff)?

```
$ diff -u <(/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/bin/java -p /disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/truffle/truffle-runtime.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/truffle/truffle-api.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/truffle/truffle-compiler.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/truffle/builder/truffle-runtime-svm.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/native-image-base.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/javacpp-platform-specific-shadowed.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/llvm-wrapper-shadowed.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/pointsto.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/javacpp-shadowed.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/objectfile.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/llvm-platform-specific-shadowed.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/svm-llvm.jar:/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/svm.jar --limit-modules org.graalvm.nativeimage.builder --list-modules | sort) <(/disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/bin/native-image -J--limit-modules=org.graalvm.nativeimage.builder --list-modules | sort)
--- /dev/fd/63	2023-08-23 10:49:15.599907604 +0200
+++ /dev/fd/62	2023-08-23 10:49:15.599907604 +0200
@@ -1,3 +1,4 @@
+
 java.base@21
 java.compiler@21
 java.logging@21
```

The only difference I see is that `native-image --list-modules` shows an extra empty line...

olpaw:
> The only difference I see is that native-image --list-modules shows an extra empty line...

That is strange. I would expect something like
```
graal/substratevm (master)$ mx native-image --list-modules
org.graalvm.nativeimage.librarysupport file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/library-support.jar

com.oracle.svm.shadowed.org.bytedeco.javacpp file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/javacpp-shadowed.jar
com.oracle.svm.shadowed.org.bytedeco.javacpp.linux.x86_64 file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/javacpp-platform-specific-shadowed.jar open
com.oracle.svm.shadowed.org.bytedeco.llvm file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/llvm-wrapper-shadowed.jar
com.oracle.svm.shadowed.org.bytedeco.llvm.linux.x86_64 file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/llvm-platform-specific-shadowed.jar open
java.base@21
java.compiler@21
java.datatransfer@21
java.desktop@21
java.instrument@21
java.logging@21
java.management@21
java.management.rmi@21
java.naming@21
java.net.http@21
java.prefs@21
java.rmi@21
java.scripting@21
java.security.jgss@21
java.security.sasl@21
java.smartcardio@21
java.sql@21
java.sql.rowset@21
java.transaction.xa@21
java.xml@21
java.xml.crypto@21
jdk.accessibility@21
jdk.attach@21
jdk.charsets@21
jdk.compiler@21
jdk.crypto.cryptoki@21
jdk.crypto.ec@21
jdk.dynalink@21
jdk.editpad@21
jdk.httpserver@21
jdk.internal.ed@21
jdk.internal.jvmstat@21
jdk.internal.le@21
jdk.internal.opt@21
jdk.internal.vm.ci@21
jdk.internal.vm.compiler
jdk.internal.vm.compiler.management
jdk.jartool@21
jdk.javadoc@21
jdk.jconsole@21
jdk.jdeps@21
jdk.jdi@21
jdk.jdwp.agent@21
jdk.jfr@21
jdk.jlink@21
jdk.jpackage@21
jdk.jshell@21
jdk.jsobject@21
jdk.jstatd@21
jdk.localedata@21
jdk.management@21
jdk.management.agent@21
jdk.management.jfr@21
jdk.naming.dns@21
jdk.naming.rmi@21
jdk.net@21
jdk.nio.mapmode@21
jdk.random@21
jdk.sctp@21
jdk.security.auth@21
jdk.security.jgss@21
jdk.unsupported@21
jdk.unsupported.desktop@21
jdk.xml.dom@21
jdk.zipfs@21
org.graalvm.collections
org.graalvm.jniutils
org.graalvm.locator
org.graalvm.nativeimage
org.graalvm.nativeimage.base file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/native-image-base.jar
org.graalvm.nativeimage.builder file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/svm.jar
org.graalvm.nativeimage.llvm file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/svm-llvm.jar
org.graalvm.nativeimage.objectfile file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/objectfile.jar
org.graalvm.nativeimage.pointsto file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/builder/pointsto.jar
org.graalvm.polyglot
org.graalvm.sdk
org.graalvm.truffle
org.graalvm.truffle.compiler
org.graalvm.truffle.runtime
org.graalvm.truffle.runtime.svm file:///home/pwoegere/OLabs/bitbucket/main/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/truffle/builder/truffle-runtime-svm.jar
org.graalvm.word
```

olpaw:
`java --list-modules` only lists the boot ModuleLayer whereas ` native-image --list-modules` lists the ModuleLayer of the image module path and afterwards the boot ModuleLayer.

Since `org.graalvm.nativeimage.librarysupport` is put on the image module path (i.e. `library-support.jar` is among the entries of the  `-imagemp` option you should see it.

jerboaa:
It's there if `-J--limit-modules=org.graalvm.nativeimage.builder` **isn't** being passed:

```
$ native-image --list-modules -J--limit-modules=org.graalvm.nativeimage.builder | grep librarysupport
$ native-image --list-modules | grep librarysupport
org.graalvm.nativeimage.librarysupport file:///disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/library-support.jar
```

jerboaa:
I revise. It's there for master, **but not** for this branch.

master (as compared to https://github.com/oracle/graal/pull/7205#issuecomment-1689872035):
```
$ native-image --list-modules -J--limit-modules=org.graalvm.nativeimage.builder | grep librarysupport
org.graalvm.nativeimage.librarysupport file:///disk/graal/upstream-sources/graalvm/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/library-support.jar
$ native-image --list-modules | grep librarysupport
org.graalvm.nativeimage.librarysupport file:///disk/graal/upstream-sources/graalvm/sdk/mxbuild/linux-amd64/GRAALVM_5852D06D07_JAVA21/graalvm-5852d06d07-java21-23.1.0-dev/lib/svm/library-support.jar
```

olpaw:
> Well, and it all starts to make sense now. Prior this branch java --list-modules isn't passing other arguments. Thus it lists all observable modules on the module path. Specifically --limit-modules isn't being passed. Therefore, the set of modules [in this code](https://github.com/oracle/graal/blob/7ea6d09948340317c6b6b37007481319a06d05de/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java#L1546) returns the empty set (with this patch), and a single-entry set on master)

Correct. And all of that because in the current impl. where we call `java --list-modules` we actually make the mistake that we put the entries from `imagemp` also on the `-p` of Java. But this is not what the situation actually looks for the builder (for the builder the entries from `imagemp` are **not** on the boot module layer and thus not subject to the effects of `-J--limit-modules ...`).

jerboaa:
I've pushed an update to this PR which always adds `org.graalvm.nativeimage.librarysupport` to the `list-modules` call. Not sure if it's the right fix, but it should solve the internal gate problem.

olpaw:
> Not sure if it's the right fix, but it should solve the internal gate problem.

Yes, but this is not what I would like to merge on our master.

Let's use the approach I already suggested in https://github.com/oracle/graal/pull/7205#issuecomment-1688559996
_But luckily we do have our --list-modules option implemented in the builder. So instead of using the --list-modules option implemented in java use our version, then -J--limit-modules=org.graalvm.nativeimage.builder cannot hurt the org.graalvm.nativeimage.librarysupport module the the builder requires ...._

jerboaa:
> > Not sure if it's the right fix, but it should solve the internal gate problem.
> 
> Yes, but this is not what I would like to merge on our master.
> 
> Let's use the approach I already suggested in [#7205 (comment)](https://github.com/oracle/graal/pull/7205#issuecomment-1688559996) _But luckily we do have our --list-modules option implemented in the builder. So instead of using the --list-modules option implemented in java use our version, then -J--limit-modules=org.graalvm.nativeimage.builder cannot hurt the org.graalvm.nativeimage.librarysupport module the the builder requires ...._

Sure, but what does it mean exactly? Instead of `java --list-modules` use `native-image --list-modules` in `callListModules()` is that what you're suggesting? I.e. call the image generator twice for each native-image run?

jerboaa:
@olpaw Gentle ping. Could you please clarify? Thanks!

jerboaa:
@olpaw @fniephaus I've pushed an update which now calls the `-H:+ListModules` implemented in the native image generator. Hopefully that gets us closer to getting it integrated. Please take another look, thanks!

sdk/mx.sdk/mx_sdk_vm_impl.py
```diff
@@ -125,7 +125,7 @@ def svm_experimental_options(experimental_options):
     dir_name='polyglot',
     launcher_configs=[mx_sdk_vm.LauncherConfig(
         destination='bin/<exe:polyglot>',
-        jar_distributions=['sdk:LAUNCHER_COMMON'],
+        jar_distributions=['sdk:LAUNCHER_COMMON', 'truffle:TRUFFLE_API'],
```
fniephaus, line 128: What is this change for? A launcher should never directly depend on truffle. Could you please revert this?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -558,10 +558,12 @@ public List<String> getBuilderJavaArgs() {
          */
         public List<Path> getBuilderModulePath() {
             List<Path> result = new ArrayList<>();
-            // Non-jlinked JDKs need truffle and graal-sdk on the module path since they
-            // don't have those modules as part of the JDK.
+            // Non-jlinked JDKs need truffle and word, collections, nativeimage on the
+            // module path since they don't have those modules as part of the JDK. Note
+            // that graal-sdk is now obsolete after the split in GR-43819 (#7171)
             if (libJvmciDir != null) {
-                result.addAll(getJars(libJvmciDir, "graal-sdk", "enterprise-graal"));
```
fniephaus, line 564: Thanks for catching!

jerboaa:
> Let's use the approach I already suggested in [#7205 (comment)](https://github.com/oracle/graal/pull/7205#issuecomment-1688559996)
> _But luckily we do have our --list-modules option implemented in the builder. So instead of using the --list-modules option implemented in java use our version, then -J--limit-modules=org.graalvm.nativeimage.builder cannot hurt the org.graalvm.nativeimage.librarysupport module the the builder requires ...._

This seems to cause the following gate failure on the `sulong` suite:

```
Fatal error: java.lang.module.ResolutionException: Module org.graalvm.truffle.runtime.svm reads more than one module named org.graalvm.truffle
	at java.base/java.lang.module.Resolver.resolveFail(Resolver.java:900)
	at java.base/java.lang.module.Resolver.checkExportSuppliers(Resolver.java:721)
	at java.base/java.lang.module.Resolver.finish(Resolver.java:380)
	at java.base/java.lang.module.Configuration.<init>(Configuration.java:139)
	at java.base/java.lang.module.Configuration.resolve(Configuration.java:422)
	at java.base/java.lang.module.Configuration.resolve(Configuration.java:254)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageClassLoaderSupport.createModuleLayer(NativeImageClassLoaderSupport.java:280)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageClassLoaderSupport.<init>(NativeImageClassLoaderSupport.java:165)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.installNativeImageClassLoader(NativeImageGeneratorRunner.java:274)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.start(NativeImageGeneratorRunner.java:132)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:97)
```

See: https://github.com/oracle/graal/actions/runs/6014210829/job/16313405565?pr=7205#step:14:1813

Any suggestions on how to fix? I cannot reproduce this failure locally :-(

jerboaa:
> I cannot reproduce this failure locally :-(

Finally reproduced it locally with:

```
mx --dynamicimports '/substratevm,/sulong' --primary-suite-path vm --native-images="graalvm-native-binutil,graalvm-native-clang,graalvm-native-clang-cl,graalvm-native-clang++,graalvm-native-ld,lib:llvmvm" --disable-polyglot --disable-libpolyglot gate --strict-mode --no-warning-as-error --tags build,sulong
```

Lets see if I can fix it...

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1505,7 +1524,8 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> cp, LinkedHa
         Function<Path, Path> substituteModulePath = useBundle() ? bundleSupport::substituteModulePath : Function.identity();
         List<Path> substitutedImageModulePath = imagemp.stream().map(substituteModulePath).toList();
 
-        Map<String, Path> modules = listModulesFromPath(javaExecutable, Stream.concat(mp.stream(), imagemp.stream()).distinct().toList());
+        List<String> mainClassArg = config.getGeneratorMainClass();
```
olpaw, line 1527: No point in introducing `mainClassArg` when you are not also using it below when the builder gets called for real.

olpaw:
> Lets see if I can fix it...

Now that https://github.com/oracle/graal/pull/7302 got merged I have a bit more time now. Let me know if you get stuck.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
jerboaa: Why not? `-H:+ListModules` is implemented in the generator, not the driver. In order to call the generator I need it's main class, no?

jerboaa: OK.

jerboaa:
@olpaw Fix for the sulong gate added. Please let me know what you think!

jerboaa:
Public gates pass. Ready for a final review. Hopefully we can get this version integrated soon-ish?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1510,7 +1529,20 @@ protected int buildImage(List<String> javaArgs, LinkedHashSet<Path> cp, LinkedHa
...
+        /*
+         * We are using the image generator for --list-modules. That cannot handle the same modules
+         * on the module path (--module-path) and image module path (-imagemp). Only pass modules
+         * via the image module path if it isn't already on the module path. Note that the path to
+         * the individual jars might be different between module path and image module path, so we
+         * use the file name for filtering.
+         */
+        Set<Path> imagempForListMods = imagemp.stream().filter(Predicate.not(a -> {
```
olpaw, line 1540: Instead of this hack we should find out why in the case of running those sulong tests we do get a module on the `-imagemp` that we already have on the module-path of the builder. I would expect this to be solved on mx level.

jerboaa: Fair enough, who can I get involved to look at this from the sulong side? My main aim is to get the Mandrel build unbroken, which is the case for about a month now and this fix is dragging along longer than I'd have thought. :-(

Here is what I know...

For the sulong tests you get this modulepath:
```
--module-path \
/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/truffle/truffle-api.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/truffle/truffle-runtime.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/truffle/truffle-compiler.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/truffle/builder/truffle-runtime-svm.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/builder/native-image-base.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/builder/svm.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/builder/svm-llvm.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/builder/javacpp-shadowed.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/builder/llvm-wrapper-shadowed.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/builder/objectfile.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/builder/llvm-platform-specific-shadowed.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/builder/javacpp-platform-specific-shadowed.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/builder/pointsto.jar
```
and this image module path:

```
-imagemp \
/home/runner/work/graal/graal/truffle/mxbuild/jdk21/dists/jdk21/truffle-api.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/truffle/builder/truffle-runtime-svm.jar:/home/runner/work/graal/graal/sdk/mxbuild/jdk21/dists/jdk17/polyglot.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/macros/junit/junit-support.jar:/home/runner/work/graal/graal/sdk/mxbuild/linux-amd64/GRAALVM_F06648F35C_JAVA21/graalvm-f06648f35c-java21-24.0.0-dev/lib/svm/library-support.jar
```

So you get `truffle-api.jar` and `truffle-runtime-svm.jar` as duplicates causing the issue. I don't know why :-/

I'm sort of out of my depth when it comes to `mx`, let alone `sulong`. So again, who would be able to help? Or better, yet, could we get this fix in, *as is*, first and file a follow-up issue somewhere to do this cleanup? We really want to get the Mandrel build working again and this patch is the missing piece. Thoughts?

christianwimmer:
@jerboaa after following this thread and discussing with Paul, it seems that there is no simple way forward where we can get a solution that is clean enough for us to merge in time for the next release. It seems like a larger cleanup is necessary, and we can do that in the next weeks.
Which means that for now you should keep a stripped-down and minimal version of this PR (without all the recent workarounds to get our gates pass) in the Mandrel repository, so that you can release with it. By "minimal" I mean that you can ignore all the complications with Truffle, which you do not need anyways.
What do you think about that?

jerboaa:
> @jerboaa after following this thread and discussing with Paul, it seems that there is no simple way forward where we can get a solution that is clean enough for us to merge in time for the next release.

Thanks for the heads-up!

> It seems like a larger cleanup is necessary, and we can do that in the next weeks. Which means that for now you should keep a stripped-down and minimal version of this PR (without all the recent workarounds to get our gates pass) in the Mandrel repository, so that you can release with it. By "minimal" I mean that you can ignore all the complications with Truffle, which you do not need anyways. What do you think about that?

@christianwimmer Sure that works for us. Please let us know when there is a suitable fix available that we can test and subsequently drop our downstream patch. We strive to keep divergence from upstream minimal.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -571,19 +573,36 @@ public List<Path> getBuilderModulePath() {
...
                  */
-                jars.addAll(getJars(rootDir.resolve(Paths.get("lib", "truffle")), "truffle-compiler"));
+                jars.addAll(getJars(libTruffleDir, "truffle-compiler"));
                 Path builderPath = rootDir.resolve(Paths.get("lib", "truffle", "builder"));
                 if (Files.exists(builderPath)) {
                     jars.addAll(getJars(builderPath, "truffle-runtime-svm", "truffle-enterprise-svm"));
+                    if (libJvmciDir != null) {
+                        // truffle-runtime-svm depends on polyglot, which is not part of non-jlinked
+                        // JDKs
+                        jars.addAll(getJars(libJvmciDir, "polyglot"));
```
chumer, line 591: This is most certainly incorrect. Starting with 23.1 polyglot is no longer part of the JDK build and passed by the user on the module-path. We still have some code that picks up jars from the JDK but it will no longer be reachable starting with 23.1.

jerboaa:
@olpaw Gentle ping on as to how to move forward with this for 24.0.

olpaw:
> @olpaw Gentle ping on as to how to move forward with this for 24.0.

I'm playing with using `jdk.internal.jimage.ImageReader` and `java.lang.module.ModuleFinder`. If all goes well there will be no extra JVM execution to query what we need to know about modules before calling the builder to build the image.

jerboaa:
Status update: This is waiting on https://github.com/oracle/graal/pull/7454 to merge and the fixes for the `--no-jlinking` case should then be fairly straight forward. No need for the list modules gymnastics any more since it'll be gone with #7454

olpaw:
Thanks! I will take a look in the next days.

fniephaus (approved):
LGTM, thanks!

jerboaa:
> All our internal CI gates passed with this change. I'm now just waiting for @chumer to respond to [#7205 (comment)](https://github.com/oracle/graal/pull/7205#discussion_r1315830587)

Thanks for the update!

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
jerboaa: Sorry, but I'm not sure I fully understand what is needed here. I took a guess and now only add `polyglot.jar` if `truffle-runtime-svm` (and the enterprise equivalent) jars are there as well. See https://github.com/oracle/graal/pull/7205/commits/0c518c76252cc597c265bf11916a84f7de415492

@chumer Is that the correct thing to do? If not, what would it be?

> truffle-runtime.svm.jar now gets added using the truffle-svm macro.

Could you clarify what the `truffle-svm` macro is?

zakkak: > Could you clarify what the truffle-svm macro is?

@jerboaa if you build with:
```
mx --primary-suite-path substratevm --components="Native Image,LibGraal,Truffle SVM Macro" --native-images=native-image,lib:jvmcicompiler build
```

notice the `Truffle SVM Macro` part the generated distribution will include the `truffle-svm` macro:

```
❯ cat sdk/latest_graalvm_home/lib/svm/macros/truffle-svm/native-image.properties 
ImageModulePath = ${.}/../../../truffle/builder/truffle-runtime-svm.jar
Args = -H:CLibraryPath=${.}/builder
IgnoreIfBuilderOnClasspath = true
```

You can then use the macro with:
```
native-image --macro:truffle-svm --other-flag my-app
```

jerboaa: OK, thanks. It's still not clear to me what **exactly** is being requested by @chumer to get this patch approved :-)

jerboaa:
@chumer Could you please take another look? Does the latest update address your concerns? Thanks in advance!

jerboaa:
Is there something missing for this PR to get merged? If so, please let me know. Thanks!

olpaw:
> Is there something missing for this PR to get merged? If so, please let me know. Thanks!

Still waiting for @chumer to respond to https://github.com/oracle/graal/pull/7205#issuecomment-1770493190

If there is no further feedback by the end of the week I will merge the PR as is.

jerboaa:
> > Is there something missing for this PR to get merged? If so, please let me know. Thanks!
> 
> Still waiting for @chumer to respond to [#7205 (comment)](https://github.com/oracle/graal/pull/7205#issuecomment-1770493190)
> 
> If there is no further feedback by the end of the week I will merge the PR as is.

Thank you!

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
chumer: I need clarification on why these changes are necessary. You should not need changes in the driver at all. You hardcode many library names here, which should not be done.

The --macro:truffle-svm should *never* be used manually. It's activated implicitly as part of adding truffle-runtime.jar on the module/class-path.

chumer (approved):

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -574,19 +576,37 @@ public List<Path> getBuilderModulePath() {
         }
 
         private List<Path> createTruffleBuilderModulePath() {
-            List<Path> jars = getJars(rootDir.resolve(Paths.get("lib", "truffle")), "truffle-api", "truffle-runtime", "truffle-enterprise");
+            Path libTruffleDir = rootDir.resolve(Paths.get("lib", "truffle"));
+            List<Path> jars = getJars(libTruffleDir, "truffle-api", "truffle-runtime", "truffle-enterprise");
             if (!jars.isEmpty()) {
```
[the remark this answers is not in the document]
zakkak: @chumer these changes were introduced by me in https://github.com/zakkak/mandrel/commit/85088b19f8062246d5f0a95c83ace120ee034c25

As I explain in https://github.com/oracle/graal/pull/7205#pullrequestreview-1582356653 (back then) these changes allowed me to build GraalVM CE with `--no-jlinking` using:

```
~/code/mx/mx --primary-suite vm --no-jlinking --env ce build
```

Looking at the CE configuration that was used while testing the above I see more components being included which probably explains the need for these changes back then: https://github.com/zakkak/mandrel/blob/85088b19f8062246d5f0a95c83ace120ee034c25/vm/mx.vm/ce#L2

As @jerboaa said we are more than happy to drop these if they are no longer relevant.
@jerboaa let me know if you need any help with this.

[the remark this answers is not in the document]
jerboaa: It doesn't (usually) trigger for us. I think this code is pre-existing and we wanted to make sure we don't break anything when this path is being used. This code seems to account for the fact that certain (modular) jar files have dependencies. If this path is taken and a `truffle-runtime-svm` is in `lib/truffle/builder`, `polyglot.jar` and `jniutils.jar` needs to be added when the JDK isn't jlinked (and, therefore, doesn't include those modules):

```
$ jar -d -f sdk/mxbuild/linux-amd64/GRAALVM_COMMUNITY_JAVA22/graalvm-community-openjdk-22+17.1/lib/truffle/builder/truffle-runtime-svm.jar | grep -E 'polyglot|truffle\.runtime'
org.graalvm.truffle.runtime.svm jar:file:///disk/graal/upstream-sources/graal/sdk/mxbuild/linux-amd64/GRAALVM_COMMUNITY_JAVA22/graalvm-community-openjdk-22+17.1/lib/truffle/builder/truffle-runtime-svm.jar!/module-info.class
requires org.graalvm.polyglot transitive
requires org.graalvm.truffle.runtime static
qualified exports com.oracle.svm.truffle to com.oracle.truffle.enterprise.svm org.graalvm.truffle.runtime.svm
qualified exports com.oracle.svm.truffle.api to com.oracle.truffle.enterprise.svm org.graalvm.truffle org.graalvm.truffle.runtime.svm
qualified exports com.oracle.svm.truffle.isolated to com.oracle.truffle.enterprise.svm org.graalvm.truffle.runtime.svm
qualified exports com.oracle.svm.truffle.nfi to com.oracle.truffle.enterprise.svm org.graalvm.truffle.runtime.svm
qualified exports com.oracle.svm.truffle.nfi.libffi to com.oracle.truffle.enterprise.svm org.graalvm.truffle.runtime.svm
qualified exports com.oracle.svm.truffle.nfi.posix to com.oracle.truffle.enterprise.svm org.graalvm.truffle.runtime.svm
qualified exports com.oracle.svm.truffle.nfi.windows to com.oracle.truffle.enterprise.svm org.graalvm.truffle.runtime.svm
$ jar -d -f ./truffle/mxbuild/jdk22/dists/jdk17/truffle-runtime.jar | grep jniutils
requires org.graalvm.jniutils transitive
$ jar -d -f sdk/mxbuild/jdk22/dists/jdk17/jniutils.jar | head -n1
org.graalvm.jniutils jar:file:///disk/graal/upstream-sources/graal/sdk/mxbuild/jdk22/dists/jdk17/jniutils.jar!/module-info.class
```

Does that sound OK?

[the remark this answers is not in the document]
jerboaa: @chumer As @zakkak said, we can drop those changes if you are happy for us to drop them. Or we can clean it up in a subsequent patch? Either way, those changes aren't important for us. The key piece is to have the `native-image` component build working in the `--no-jlinking` config.

jerboaa:
Thank you, @chumer for the review and approval!

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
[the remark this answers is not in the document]
chumer: @jerboaa @zakkak Your CE env config explains why this still triggers for you. As you know, we have gone unchained and many of your CE env configuration components are about to be removed on master. I'd recommend switching to a config similar to this: https://github.com/oracle/graal/blob/master/vm/mx.vm/ce#L2
We do change these component lists whenever we need to, so I'd recommend undoing the custom changes to your env config.

> Does that sound OK?

You should never need to put `truffle-runtime-svm` on the module-path manually in the driver after >= 23.1, as you mentioned it has further module dependencies so can only activate automatically if certain jars are put on the image module/class-path by the user. The driver no longer decides this >= 23.1, which is why I mentioned that the condition should not trigger above. If the truffle-runtime.jar gets put on the image module-path by the user the truffle-svm macro gets activated which then activates truffle-runtime-svm. truffle-runtime.jar has all the necessary dependencies and already expects them on the image module-path, so when truffle-runtime-svm gets activated through the macro all dependencies are satisfied, including polyglot.

Ultimately @olpaw is in charge of this code and needs to decide what goes in and what not. I do not know enough about --no-jlinking to judge whether these changes are justified. I know that many of your changes probably stem from an outdated/wrong environment config, and I think these changes will mostly no longer be necessary after you update the env. @olpaw should decide whether it is fine to merge this as an intermediate step.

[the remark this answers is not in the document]
jerboaa: Thanks for your input. So the the comment inside the `if` block is accurate: `This is legacy support and should in the future no longer be needed.` Question is when we remove this. @olpaw Opinions?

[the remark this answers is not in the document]
olpaw: @jerboaa your PR is already in our merge queue. Feel free to provide a cleanup PR after this one got merged that reduces the changes needed for your usecase to a minimum.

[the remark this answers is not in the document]
jerboaa: Great, thank you. Will do that, then.

jerboaa:
@olpaw Hi! Just checking, is this still in the merge queue?

fniephaus:
> Hi! Just checking, is this still in the merge queue?

Yes, it is... it unfortunately failed due to some infra issue, but it's currently on the top again. If all goes well, it should be merged within the next couple of hours.

jerboaa:
> If all goes well, it should be merged within the next couple of hours.

OK, thanks Fabio.

fniephaus:
Sorry, but it failed again...I hope we can resolve the infra issues very soon.

PR #7070, opened 2023-07-26
Update committing of java level events

roberttoyonaga:
Resolves related issues: https://github.com/oracle/graal/issues/7052 and https://github.com/graalvm/mandrel/issues/546

Draft because I'm waiting for LabsJDK to be updated with the changes in OpenJDK21+32 that caused the breakage so I can test. For now, I have tested with mandrel and Temurin build of openjdk21+32.

**Summary**
This change https://github.com/openjdk/jdk21/pull/129 caused some substitutions to become out of date. Additionally, the way Java-level events are committed by the JavaEventWriter has also changed. In openjdk, committing is now done in  native Hotspot code in order to avoid safepointing potentially interrupting the commit operation. This PR achieves the same behaviour by subsituting the new `JVM#commit` method and making it `uninterruptible`.

Issue #6854, opened 2023-06-21
Adopt support for JDK 21


fniephaus:
### TL;DR

The next feature release of GraalVM will exclusively support JDK 21.

### Goals

As GraalVM is adopting the [JDK release process](https://www.graalvm.org/2022/openjdk-announcement/), GraalVM will exclusively support JDK 21 in the [September release](https://www.graalvm.org/release-calendar/). This means users can take advantage of [all new features introduced with JDK 21](https://openjdk.org/projects/jdk/21/) with the Graal JIT compiler. All preview features and [generational ZGC](https://openjdk.org/jeps/439) will not be supported in Native Image. [Virtual threads](https://openjdk.org/jeps/444) preview support was available in Native Image since [JDK 19](https://medium.com/graalvm/graalvm-22-3-is-here-jdk-19-builds-jlink-support-new-monitoring-features-and-more-f6e2b2eeff95) and will be available as a finalized feature in JDK 21.

The community and end-users are encouraged to upgrade to JDK 21, but have several options if they need to stay on older JDK versions: The community edition of GraalVM for JDK 17 will see [another CPU update in October](https://www.graalvm.org/release-calendar/), while Oracle GraalVM for JDK 17 is [supported until 2025](https://docs.oracle.com/en/graalvm/jdk/17/docs/release-calendar/) with security updates and important bugfixes.

### Non-Goals

- Add Native Image support for all preview features

### Related PRs

- https://github.com/oracle/graal/pull/4802
- https://github.com/oracle/graal/pull/6835

thomaswue:
We would prefer if other distributions release 17 from the same branch as the Community Edition. If there are important changes on master, they should be backported. We are taking this approach also for the Oracle GraalVM for Java 17 distribution.

PR #7211, opened 2023-08-15
Fix recent JFR changes to be compatible with JDK version < 21

roberttoyonaga:
This is a follow up to https://github.com/oracle/graal/pull/7070 which was closed because these commits (https://github.com/oracle/graal/commit/ad853ee6d0a540ffafba6b01ee8c31209558da43 https://github.com/oracle/graal/commit/8e778030ad096b26ac4a82c8540d6040c7d23156 https://github.com/oracle/graal/commit/1a87440137ae595d65179bdbf5e9b2f1dd08cd4d) essentially did the same thing. However, the merged commits may cause issues for the JFR java level infrastructure for JDK<21. This PR aims to maintain the support for JDK < 21. 

It's worth mentioning that native image builds  with JDK17 are currently failing for unrelated reasons with or without this change. See issue: https://github.com/oracle/graal/issues/7212


zapster:
Hi and thanks for your contribution! The next release 23.1 will be JDK 21 only (#6854), so we are gradually dropping support for other version, essentially by not fixing them. Also, we only support the "our latest" JDK 21, which is the version that is specified here: https://github.com/oracle/graal/blob/master/common.json#L38

PR #7252, opened 2023-08-22
Fix bug in JFR periodic chunk events and fix thread data in ThreadCPULoad events

roberttoyonaga:
### Summary
This PR fixes two native image JFR problems. 

1. Problem: Periodic end/begin chunk events do not get emitted when the JFR API is used. This is because the appropriate periodic events were previously only registered when JFR is started from the command line. However, this is a problem if recordings are instead started using the JFR API (which is what the unit tests use).
 
This part of the PR should be backported to previous GraalVM versions because it affects every version the JFR custom event API is supported. 

2. Problem: ThreadCPULoad events attach thread data of the thread that is emitting the event instead of the thread that the data belongs to. This problem only manifests when the event is emitted at the end of a chunk (one thread will emit events on behalf of all running threads). The existing test for ThreadCPULoad event only accounts for the case where the event is emitted as a result of a thread exiting, so the bug does not manifest. 

This part of the PR does not need to be backported because ThreadCPULoad was only recently added after the 23.0 release.


christianhaeubl:
Thanks - this will be merged to master in a few days. I will also take care of the backports.

PR #7256, opened 2023-08-23
Use calling convention to derive incoming param register and stack slot locations

adinn:
This PR modifies the debug info provider to use the compiler calling convention when generating locations for incoming parameters. This ensures that the correct locations are used appropriate to the current operating system, target CPU and incoming call API type (Java or Native).

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
@@ -29,7 +29,7 @@ import java.util.function.Function;
 
 import jdk.graal.compiler.debug.DebugContext;
 import jdk.graal.compiler.printer.GraalDebugHandlersFactory;
-import jdk.vm.ci.code.RegisterConfig;
+import com.oracle.svm.core.graal.meta.RuntimeConfiguration;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.Platform;
 
@@ -44,7 +44,6 @@ import com.oracle.svm.core.UniqueShortNameProvider;
 import com.oracle.svm.core.UniqueShortNameProviderDefaultImpl;
 import com.oracle.svm.core.feature.AutomaticallyRegisteredFeature;
 import com.oracle.svm.core.feature.InternalFeature;
-import com.oracle.svm.core.graal.meta.SubstrateRegisterConfig;
 import com.oracle.svm.core.option.HostedOptionValues;
 import com.oracle.svm.hosted.FeatureImpl;
 import com.oracle.svm.hosted.ProgressReporter;
@@ -109,10 +108,9 @@ class NativeImageDebugInfoFeature implements InternalFeature {
             var accessImpl = (FeatureImpl.BeforeImageWriteAccessImpl) access;
             var image = accessImpl.getImage();
             var debugContext = new DebugContext.Builder(HostedOptionValues.singleton(), new GraalDebugHandlersFactory(GraalAccess.getOriginalSnippetReflection())).build();
-            RegisterConfig registerConfig = ((FeatureImpl.BeforeImageWriteAccessImpl) access).getRuntimeConfiguration().getBackendForNormalMethod().getCodeCache().getRegisterConfig();
-            assert registerConfig instanceof SubstrateRegisterConfig : "unexpected register configuration!";
+            RuntimeConfiguration runtimeConfiguration = ((FeatureImpl.BeforeImageWriteAccessImpl) access).getRuntimeConfiguration();
             DebugInfoProvider provider = new NativeImageDebugInfoProvider(debugContext, image.getCodeCache(), image.getHeap(), image.getNativeLibs(), accessImpl.getHostedMetaAccess(),
-                    (SubstrateRegisterConfig) registerConfig);
+                            runtimeConfiguration);
             var objectFile = image.getObjectFile();
             objectFile.installDebugInfo(provider);
 
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
@@ -29,6 +29,7 @@ import java.util.function.Function;
 
 import jdk.graal.compiler.debug.DebugContext;
 import jdk.graal.compiler.printer.GraalDebugHandlersFactory;
+import jdk.vm.ci.code.RegisterConfig;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.Platform;
 
@@ -43,6 +44,7 @@ import com.oracle.svm.core.UniqueShortNameProvider;
 import com.oracle.svm.core.UniqueShortNameProviderDefaultImpl;
 import com.oracle.svm.core.feature.AutomaticallyRegisteredFeature;
 import com.oracle.svm.core.feature.InternalFeature;
+import com.oracle.svm.core.graal.meta.SubstrateRegisterConfig;
 import com.oracle.svm.core.option.HostedOptionValues;
 import com.oracle.svm.hosted.FeatureImpl;
 import com.oracle.svm.hosted.ProgressReporter;
@@ -107,7 +109,10 @@ class NativeImageDebugInfoFeature implements InternalFeature {
             var accessImpl = (FeatureImpl.BeforeImageWriteAccessImpl) access;
             var image = accessImpl.getImage();
             var debugContext = new DebugContext.Builder(HostedOptionValues.singleton(), new GraalDebugHandlersFactory(GraalAccess.getOriginalSnippetReflection())).build();
-            DebugInfoProvider provider = new NativeImageDebugInfoProvider(debugContext, image.getCodeCache(), image.getHeap(), image.getNativeLibs(), accessImpl.getHostedMetaAccess());
+            RegisterConfig registerConfig = ((FeatureImpl.BeforeImageWriteAccessImpl) access).getRuntimeConfiguration().getBackendForNormalMethod().getCodeCache().getRegisterConfig();
+            assert registerConfig instanceof SubstrateRegisterConfig : "unexpected register configuration!";
+            DebugInfoProvider provider = new NativeImageDebugInfoProvider(debugContext, image.getCodeCache(), image.getHeap(), image.getNativeLibs(), accessImpl.getHostedMetaAccess(),
+                    (SubstrateRegisterConfig) registerConfig);
             var objectFile = image.getObjectFile();
             objectFile.installDebugInfo(provider);
 
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
@@ -2457,7 +2457,12 @@ class NativeImageDebugInfoProvider extends NativeImageDebugInfoProviderBase impl
         }
 
         static NativeImageDebugStackValue create(StackSlot value, int framesize) {
-            return memoizedCreate(value.getOffset(framesize));
+            // Work around a problem on AArch64 where StackSlot asserts if it is
+            // passed a zero frame size, even though this is what is expected
+            // for stack slot offsets provided at the point of entry (because,
+            // unlike x86, lr has not been pushed).
+            int offset = (framesize == 0 ? value.getRawOffset() : value.getOffset(framesize));
+            return memoizedCreate(offset);
         }
 
         static NativeImageDebugStackValue create(DebugLocalValueInfo previous, int adjustment) {
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
@@ -25,9 +25,6 @@
  */
 package com.oracle.svm.hosted.image;
 
-import static com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugFrameSizeChange.Type.CONTRACT;
-import static com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugFrameSizeChange.Type.EXTEND;
-
 import java.lang.reflect.Modifier;
 import java.nio.file.Path;
 import java.util.ArrayList;
@@ -58,7 +55,6 @@ import com.oracle.graal.pointsto.infrastructure.WrappedJavaMethod;
 import com.oracle.graal.pointsto.infrastructure.WrappedJavaType;
 import com.oracle.graal.pointsto.meta.AnalysisMethod;
 import com.oracle.objectfile.debuginfo.DebugInfoProvider;
-import com.oracle.svm.core.OS;
 import com.oracle.svm.core.SubstrateOptions;
 import com.oracle.svm.core.UniqueShortNameProvider;
 import com.oracle.svm.core.code.CompilationResultFrameTree.Builder;
@@ -67,7 +63,7 @@ import com.oracle.svm.core.code.CompilationResultFrameTree.FrameNode;
 import com.oracle.svm.core.code.CompilationResultFrameTree.Visitor;
 import com.oracle.svm.core.config.ConfigurationValues;
 import com.oracle.svm.core.graal.code.SubstrateBackend.SubstrateMarkId;
-import com.oracle.svm.core.graal.meta.SubstrateRegisterConfig;
+import com.oracle.svm.core.graal.meta.RuntimeConfiguration;
 import com.oracle.svm.core.image.ImageHeapPartition;
 import com.oracle.svm.hosted.DeadlockWatchdog;
 import com.oracle.svm.hosted.c.NativeLibraries;
@@ -97,7 +93,6 @@ import com.oracle.svm.util.ClassUtil;
 
 import jdk.vm.ci.aarch64.AArch64;
 import jdk.vm.ci.amd64.AMD64;
-import jdk.vm.ci.code.Architecture;
 import jdk.vm.ci.code.BytecodeFrame;
 import jdk.vm.ci.code.BytecodePosition;
 import jdk.vm.ci.code.Register;
@@ -116,6 +111,9 @@ import jdk.vm.ci.meta.ResolvedJavaType;
 import jdk.vm.ci.meta.Signature;
 import jdk.vm.ci.meta.Value;
 
+import static com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugFrameSizeChange.Type.CONTRACT;
+import static com.oracle.objectfile.debuginfo.DebugInfoProvider.DebugFrameSizeChange.Type.EXTEND;
+
 /**
  * Implementation of the DebugInfoProvider API interface that allows type, code and heap data info
  * to be passed to an ObjectFile when generation of debug info is enabled.
@@ -125,8 +123,8 @@ class NativeImageDebugInfoProvider extends NativeImageDebugInfoProviderBase impl
     private final Set<HostedMethod> allOverrides;
 
     NativeImageDebugInfoProvider(DebugContext debugContext, NativeImageCodeCache codeCache, NativeImageHeap heap, NativeLibraries nativeLibs, HostedMetaAccess metaAccess,
-                            SubstrateRegisterConfig registerConfig) {
-        super(codeCache, heap, nativeLibs, metaAccess, registerConfig);
+                    RuntimeConfiguration runtimeConfiguration) {
+        super(codeCache, heap, nativeLibs, metaAccess, runtimeConfiguration);
         this.debugContext = debugContext;
         /* Calculate the set of all HostedMethods that are overrides. */
         allOverrides = heap.hUniverse.getMethods().stream()
@@ -1891,8 +1889,9 @@ class NativeImageDebugInfoProvider extends NativeImageDebugInfoProviderBase impl
                     // need to redefine the value for this param using a stack slot value
                     // that allows for the stack being extended by framesize. however we
                     // also need to remove any adjustment that was made to allow for the
-                    // difference between the caller SP and the pre-extend callee SP.
-                    int adjustment = frameSize - PRE_EXTEND_FRAME_ADJUSTMENT;
+                    // difference between the caller SP and the pre-extend callee SP
+                    // because of a stacked return address.
+                    int adjustment = frameSize - PRE_EXTEND_FRAME_SIZE;
                     NativeImageDebugLocalValue value = NativeImageDebugStackValue.create(localInfo, adjustment);
                     NativeImageDebugLocalValueInfo nativeLocalInfo = (NativeImageDebugLocalValueInfo) localInfo;
                     NativeImageDebugLocalValueInfo newLocalinfo = new NativeImageDebugLocalValueInfo(nativeLocalInfo.name,
@@ -1970,12 +1969,11 @@ class NativeImageDebugInfoProvider extends NativeImageDebugInfoProviderBase impl
                 JavaKind kind = ownerType.getJavaKind();
                 JavaKind storageKind = isForeignWordType(ownerType, ownerType) ? JavaKind.Long : kind;
                 assert kind == JavaKind.Object : "must be an object";
-                // NativeImageDebugLocalValue value = locProducer.nextLocation(kind);
                 NativeImageDebugLocalValue value = locProducer.thisLocation();
                 debugContext.log(DebugContext.DETAILED_LEVEL, "locals[%d] %s type %s slot %d", localIdx, name, ownerType.getName(), slot);
                 debugContext.log(DebugContext.DETAILED_LEVEL, "  =>  %s kind %s", value, storageKind);
                 localInfos.add(new NativeImageDebugLocalValueInfo(name, value, storageKind, ownerType, slot, firstLine));
-                slot += storageKind.getSlotCount();
+                slot += kind.getSlotCount();
                 localIdx++;
             }
             for (int i = 0; i < parameterCount; i++) {
@@ -1984,12 +1982,11 @@ class NativeImageDebugInfoProvider extends NativeImageDebugInfoProviderBase impl
                 ResolvedJavaType paramType = (ResolvedJavaType) signature.getParameterType(i, ownerType);
                 JavaKind kind = paramType.getJavaKind();
                 JavaKind storageKind = isForeignWordType(paramType, ownerType) ? JavaKind.Long : kind;
-                // NativeImageDebugLocalValue value = locProducer.nextLocation(kind);
                 NativeImageDebugLocalValue value = locProducer.paramLocation(i);
                 debugContext.log(DebugContext.DETAILED_LEVEL, "locals[%d] %s type %s slot %d", localIdx, name, ownerType.getName(), slot);
                 debugContext.log(DebugContext.DETAILED_LEVEL, "  =>  %s kind %s", value, storageKind);
                 localInfos.add(new NativeImageDebugLocalValueInfo(name, value, storageKind, paramType, slot, firstLine));
-                slot += storageKind.getSlotCount();
+                slot += kind.getSlotCount();
                 localIdx++;
             }
             return localInfos;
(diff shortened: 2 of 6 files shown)


adinn:
This PR is now ready for review.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProviderBase.java
```diff
@@ -426,6 +444,22 @@ private static HashMap<JavaKind, HostedType> initJavaKindToHostedTypes(HostedMet
...
+     * @param method The method whose calling convention is required.
+     * @return The calling convention for the method.
+     */
+    protected SubstrateCallingConvention getCallingConvention(HostedMethod method) {
+        SubstrateCallingConventionKind callingConventionKind = method.getCallingConventionKind();
+        HostedType declaringClass = method.getDeclaringClass();
+        HostedType receiverType = method.isStatic() ? null : declaringClass;
+        Signature signature = method.getSignature();
+        SubstrateCallingConventionType type = callingConventionKind.toType(false);
+        return (SubstrateCallingConvention) registerConfig.getCallingConvention(type, signature.getReturnType(declaringClass), signature.toParameterTypes(receiverType), valueKindFactory);
```
pejovica, line 460: You can use `ResolvedJavaMethod.toParameterTypes()` to avoid dealing with `receiverType`.
```suggestion
        return (SubstrateCallingConvention) registerConfig.getCallingConvention(type, signature.getReturnType(declaringClass), method.toParameterTypes(), valueKindFactory);
```

adinn: Calling `method.toParameterTypes()` fails when the method is a constructor. In that case the calling convention needs to take account of the 'this' parameter as well as the constructor arguments. However, if I were to call `method.toParameterTypes()` when `method` was a constructor it would call `method.getSignature().toParameterTypes(null)` and the resulting calling convention would omit the 'this' argument. That is why my code handles this case specially and only passes null when the method is static.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoFeature.java
```diff
@@ -107,7 +109,10 @@ public void beforeImageWrite(BeforeImageWriteAccess access) {
             var accessImpl = (FeatureImpl.BeforeImageWriteAccessImpl) access;
             var image = accessImpl.getImage();
             var debugContext = new DebugContext.Builder(HostedOptionValues.singleton(), new GraalDebugHandlersFactory(GraalAccess.getOriginalSnippetReflection())).build();
-            DebugInfoProvider provider = new NativeImageDebugInfoProvider(debugContext, image.getCodeCache(), image.getHeap(), image.getNativeLibs(), accessImpl.getHostedMetaAccess());
+            RegisterConfig registerConfig = ((FeatureImpl.BeforeImageWriteAccessImpl) access).getRuntimeConfiguration().getBackendForNormalMethod().getCodeCache().getRegisterConfig();
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProviderBase.java
```diff
@@ -84,11 +92,21 @@ public abstract class NativeImageDebugInfoProviderBase {
     protected final HostedType wordBaseType;
     final HashMap<JavaKind, HostedType> javaKindToHostedType;
     private final Path cachePath = SubstrateOptions.getDebugInfoSourceCacheRoot();
+    /*
+     * A factory used when deriving the calling convention for a specific method
+     */
+    private static final ValueKindFactory<LIRKind> valueKindFactory = new ValueKindFactory<>() {
```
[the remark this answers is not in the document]
adinn: Done

adinn:
@pejovica Thank you for the review. I have responded to your comments and modified the code where appropriate. I also rebased on master to incorporate a recent change that added a heartbeat callback to the debug info generator. Could you please review again?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -42,6 +42,20 @@
...
+import com.oracle.svm.hosted.c.info.PointerToInfo;
+import com.oracle.svm.hosted.c.info.PropertyInfo;
+import com.oracle.svm.hosted.c.info.RawStructureInfo;
+import com.oracle.svm.hosted.c.info.SizableInfo;
+import com.oracle.svm.hosted.c.info.SizableInfo.ElementKind;
+import com.oracle.svm.hosted.c.info.StructFieldInfo;
+import com.oracle.svm.hosted.c.info.StructInfo;
+import com.oracle.svm.util.ClassUtil;
+import jdk.vm.ci.code.CallingConvention;
+import jdk.vm.ci.meta.AllocatableValue;
```
zakkak, line 58: Nitpick: The ordering of this imports seems unaligned with the rest. How come checkstyle is not complaining?

adinn: Hmm, I'm not sure why they are out of order nor why checkstyle is not complaining. I had to rebase the patch to deal with the addition of a heartbeat callback to the debug info generator and this seemed to be the order in head at the time of merge.

adinn: I have re-ordered the imports so they are grouped alphabetically.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProviderBase.java
```diff
@@ -85,10 +96,14 @@ public abstract class NativeImageDebugInfoProviderBase {
     final HashMap<JavaKind, HostedType> javaKindToHostedType;
     private final Path cachePath = SubstrateOptions.getDebugInfoSourceCacheRoot();
 
-    public NativeImageDebugInfoProviderBase(NativeImageCodeCache codeCache, NativeImageHeap heap, NativeLibraries nativeLibs, HostedMetaAccess metaAccess) {
+    /*
+     * A factory used when deriving the calling convention for a specific method
```
pejovica, line 100: A leftover?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProviderBase.java
```diff
@@ -426,6 +441,25 @@ private static HashMap<JavaKind, HostedType> initJavaKindToHostedTypes(HostedMet
...
+    protected SubstrateCallingConvention getCallingConvention(HostedMethod method) {
+        SubstrateCallingConventionKind callingConventionKind = method.getCallingConventionKind();
+        HostedType declaringClass = method.getDeclaringClass();
+        HostedType receiverType = method.isStatic() ? null : declaringClass;
+        Signature signature = method.getSignature();
+        SubstrateCallingConventionType type = callingConventionKind.toType(false);
+        Backend backend = runtimeConfiguration.lookupBackend(method);
+        RegisterConfig registerConfig = backend.getCodeCache().getRegisterConfig();
+        assert registerConfig instanceof SubstrateRegisterConfig;
+        return (SubstrateCallingConvention) registerConfig.getCallingConvention(type, signature.getReturnType(declaringClass), signature.toParameterTypes(receiverType), backend);
```
pejovica, line 460: I'd pass `null` here since `WrappedSignature.getReturnType` doesn't use the `accessingClass` argument anyway.
```suggestion
        return (SubstrateCallingConvention) registerConfig.getCallingConvention(type, signature.getReturnType(null), signature.toParameterTypes(receiverType), backend);
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProviderBase.java
adinn: Hmm, not sure how this got in there? I think it may have happened when I rebased. I'll delete it.

adinn: Done.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProviderBase.java
adinn: Done.

adinn:
@pejovica Thanks for the re-review. I have corrected the code as per your comments.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -2160,164 +2169,102 @@ public NativeImageDebugLocationInfo split(int stackDecrement, int adjustment) {
...
-     * pushed lr and fp registers these have not been rolled into the offsets derived for the
-     * unadjusted sp.
+     * The value is zero because nothing is pushed until the stack extend occurs.
      */
-    static final int AARCH64_FRAMESIZE_ADJUSTMENT = 0;
+    static final int AARCH64_PRE_EXTEND_FRAME_ADJUSTMENT = 0;
 
-    static int adjustFrameSize(int frameSize) {
-        // make sure this is the right arch and os
+    static {
```
pejovica, line 2192: It seems that this entire initialization block can be reduced to just this:
```java
    static {
        PRE_EXTEND_FRAME_ADJUSTMENT = ConfigurationValues.getTarget().arch.getReturnAddressSize();
    }
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -2160,164 +2169,102 @@ public NativeImageDebugLocationInfo split(int stackDecrement, int adjustment) {
...
-
-        ParamLocationProducer(ResolvedJavaMethod method) {
-            Architecture arch = ConfigurationValues.getTarget().arch;
+        private final HostedMethod hostedMethod;
+        private final CallingConvention callingConvention;
+        private boolean usesStack;
+
+        private static final int adjustment;
+
+        static {
```
pejovica, line 2212: Same here. The entire initialization block can be reduced to:
```java
    static {
        adjustment = ConfigurationValues.getTarget().arch.getReturnAddressSize();
    }
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1964,7 +1971,8 @@ private List<DebugLocalValueInfo> initSyntheticInfoList(ParamLocationProducer lo
                 JavaKind kind = ownerType.getJavaKind();
                 JavaKind storageKind = isForeignWordType(ownerType, ownerType) ? JavaKind.Long : kind;
                 assert kind == JavaKind.Object : "must be an object";
-                NativeImageDebugLocalValue value = locProducer.nextLocation(kind);
+                // NativeImageDebugLocalValue value = locProducer.nextLocation(kind);
```
pejovica, line 1974: Do we still need this comment here and a few lines below?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -2160,164 +2169,102 @@ public NativeImageDebugLocationInfo split(int stackDecrement, int adjustment) {
...
+                RegisterValue registerValue = (RegisterValue) value;
+                return NativeImageDebugRegisterValue.create(registerValue);
+            } else {
+                // call argument must be a stack slot if it is not a register
+                StackSlot stackSlot = (StackSlot) value;
+                this.usesStack = true;
+                // the calling convention provides raw offsets from the SP at point
+                // of call but these need adjusting to allow for a stacked return
+                // address and/or frame pointer according to the architecture
+                return new NativeImageDebugStackValue(stackSlot.getRawOffset() + adjustment);
```
pejovica, line 2262: Shouldn't we be using `create` here?
```suggestion
                return NativeImageDebugStackValue.create(stackSlot, adjustment);
```

Also, there's no need to use `getRawOffset` because `StackSlot.getOffset` calculates the correct location of the incoming stack arguments as long as you use the correct frame size at the point of interest. And since at the method entry point we only have the part of the frame that contains the return address, the size of the frame is equal to the size of the return address, which is what `adjustment` actually is.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: No. Removed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: This static field can be dropped in favour of the previous one.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Yes, we can use `memoizedCreate(int)` to avoid duplication of the same value.
I renamed `PRE_EXTEND_FRAME_ADJUSTMENT` to `PRE_EXTEND_FRAME_SIZE` and passed `getOffset(PRE_EXTEND_FRAME_SIZE)` to `memoizedCreate`.

adinn:
@pejovica Further rework completed and pushed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1964,7 +1970,7 @@ private List<DebugLocalValueInfo> initSyntheticInfoList(ParamLocationProducer lo
                 JavaKind kind = ownerType.getJavaKind();
                 JavaKind storageKind = isForeignWordType(ownerType, ownerType) ? JavaKind.Long : kind;
                 assert kind == JavaKind.Object : "must be an object";
-                NativeImageDebugLocalValue value = locProducer.nextLocation(kind);
+                NativeImageDebugLocalValue value = locProducer.thisLocation();
                 debugContext.log(DebugContext.DETAILED_LEVEL, "locals[%d] %s type %s slot %d", localIdx, name, ownerType.getName(), slot);
                 debugContext.log(DebugContext.DETAILED_LEVEL, "  =>  %s kind %s", value, storageKind);
                 localInfos.add(new NativeImageDebugLocalValueInfo(name, value, storageKind, ownerType, slot, firstLine));
```
pejovica, line 1976: There is a small bug in the next line that I found while testing the PR. I've noticed that some arguments in synthetic frames are reported as optimized out, and the reason is that we're miscounting slots for `Word` type arguments: in the next line we use `storageKind` to count the slots, but it should be `kind`.
```diff
-                slot += storageKind.getSlotCount();
+                slot += kind.getSlotCount();
```
This also reminded me that we should test this. So could you please add tests that check that we can list the arguments of (1) an entry point (they use a different calling convention than Java now), e.g., I think checking `main` (the entry point for `JavaMainWrapper.run`) should be sufficient, and (2) a method with a mixture of `Word` and non-`Word` type arguments (as a regression test for this bug).

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -1977,7 +1983,7 @@ private List<DebugLocalValueInfo> initSyntheticInfoList(ParamLocationProducer lo
                 ResolvedJavaType paramType = (ResolvedJavaType) signature.getParameterType(i, ownerType);
                 JavaKind kind = paramType.getJavaKind();
                 JavaKind storageKind = isForeignWordType(paramType, ownerType) ? JavaKind.Long : kind;
-                NativeImageDebugLocalValue value = locProducer.nextLocation(kind);
+                NativeImageDebugLocalValue value = locProducer.paramLocation(i);
                 debugContext.log(DebugContext.DETAILED_LEVEL, "locals[%d] %s type %s slot %d", localIdx, name, ownerType.getName(), slot);
                 debugContext.log(DebugContext.DETAILED_LEVEL, "  =>  %s kind %s", value, storageKind);
                 localInfos.add(new NativeImageDebugLocalValueInfo(name, value, storageKind, paramType, slot, firstLine));
```
pejovica, line 1989: Here as well. In the next line we use `storageKind` to count the slots, but it should be `kind`.
```diff
-                slot += storageKind.getSlotCount();
+                slot += kind.getSlotCount();
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
```diff
@@ -2160,164 +2166,54 @@ public NativeImageDebugLocationInfo split(int stackDecrement, int adjustment) {
...
+                return NativeImageDebugRegisterValue.create(registerValue);
             } else {
-                return nextStackLocation();
+                // call argument must be a stack slot if it is not a register
+                StackSlot stackSlot = (StackSlot) value;
+                this.usesStack = true;
+                // the calling convention provides offsets from the SP relative to the current
+                // frame size. At the point of call the frame may or may not include a return
+                // address depending on the architecture.
+                return NativeImageDebugStackValue.memoizedCreate(stackSlot.getOffset(PRE_EXTEND_FRAME_SIZE));
```
pejovica, line 2211: nit: I'd use `create` here for consistency.
```suggestion
                return NativeImageDebugStackValue.create(stackSlot, PRE_EXTEND_FRAME_SIZE);
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: I still need to update the tests.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: @pejovica The change suggested above results in a failure on AArch64. The call to 
```
NativeImageDebugStackValue.create(stackSlot, PRE_EXTEND_FRAME_SIZE)
```
results in a call to
```
NativeImageDebugStackValue.memoizedCreate(value.getOffset(PRE_EXTEND_FRAME_SIZE))
```
On AARCH64 we have `PRE_EXTEND_FRAME_SIZE == 0` which means that we run up against the following assert in `StackSlot.getOffset()`
```
public int getOffset(int totalFrameSize) {
    assert totalFrameSize > 0 || !addFrameSize;
    . . .
```
The problem is that the pre-frame push stack slots coming out of the AArch64 calling convention satisfy `addFrameSize == true` even though at this point on AArch64 the total frame size is zero. The problem does not arise on x86_64 because the call will have pushed a return address meaning that the total frame size will not be zero.

I have finessed this in NativeImageDebugStackValue.create by special casing `framesize == 0` to call `StackSlot.getRawOffset()`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
pejovica: Yes, something like that, but for a method that has more arguments after the first `WordBase` type argument (e.g., like `CEntryPointSnippets.createIsolate`), because if you were to set a breakpoint on such a method, you wouldn't be able to see those arguments without this fix.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
pejovica: Looks good, thanks!

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageDebugInfoProvider.java
adinn: I just pushed a change to test the mixed argument case. It recognizes all arguments with the patch to `initSyntheticInfoList` in place (i.e. incrementing slot count using `kind.getSlotCount()`). It fails I revert that specific fix (i.e. increment by `storageKind.getSlotCount()`). So, this looks like a wrap.

pejovica: Yes, looks good, thanks!

pejovica (approved):
@adinn I'm going to start the integration.

adinn:
@pejovica I am afraid the newly added mixed argument test is failing on AArch64. That's not because the debug info for the incoming arguments at method entry is incorrect. The problem is that on AArch64 gdb does not always place a method breakpoint at the first instruction of the method. I'm not clear exactly why it does this but it is a known issue. What it means is that the test fails to see the initial bindings for the method arguments because the breakpoint ends up being placed on an instruction that is well beyond the range between the 1st instruction and the stack push.

Specifically, when command `break com.oracle.svm.test.debug.CStructTests::testMixedArguments` is passed to gdb on AArch64 the break is placed at offset 28 in the method, where the first nop is encountered.

```
(gdb) break com.oracle.svm.test.debug.CStructTests::testMixedArguments
Breakpoint 17 at 0x6360ac: file com/oracle/svm/test/debug/CStructTests.java, line 236.

(gdb) continue

Thread 1 "hello_image" hit Breakpoint 17, com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*) (this=<optimized out>, m1=0x21f69b8, s=<optimized out>, ss1=<optimized out>, l=<optimized out>, m2=0x21f69d8, ss2=<optimized out>, m3=0x21f69f8) at com/oracle/svm/test/debug/CStructTests.java:236
236	        System.out.println("You find " + m1);
(gdb) info args
this = <optimized out>
m1 = 0x21f69b8
s = <optimized out>
ss1 = <optimized out>
l = <optimized out>
m2 = 0x21f69d8
ss2 = <optimized out>
m3 = 0x21f69f8
(gdb) x/i $pc
=> 0x6360ac <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)+28>:	nop
(gdb) x/i 'com.oracle.svm.test.debug.CStructTests::testMixedArguments'
   0x636090 <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)>:	sub	sp, sp, #0x60
(gdb) 
   0x636094 <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)+4>:	stp	x29, x30, [sp, #80]
(gdb) 
   0x636098 <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)+8>:	ldr	x0, [x28, #8]
(gdb) 
   0x63609c <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)+12>:	cmp	sp, x0
(gdb) 
   0x6360a0 <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)+16>:	
    b.ls	0x636704 <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)+1652>  // b.plast
(gdb) 
   0x6360a4 <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)+20>:	adrp	x2, 0x1865000
(gdb) 
   0x6360a8 <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)+24>:	ldr	x2, [x2, #1752]
(gdb) 
=> 0x6360ac <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)+28>:	nop
```
This is well into the method body code and certainly beyond the stack push which is complete when we reach the instruction at offset 8. The initial argument debug info is posted for the code between from offset 0 to offset 8.

I modified the test to identify the address of the first instruction in `CStructTests'::testMixedArguments` and place a breakpoint at that specific address. In that case the test passes on AArch64:

```
(gdb) x/i 'com.oracle.svm.test.debug.CStructTests'::testMixedArguments
   0x636090 <com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*)>:	sub	sp, sp, #0x60

bp = 636090 636090
(gdb) break *0x636090
Breakpoint 17 at 0x636090: file com/oracle/svm/test/debug/CStructTests.java, line 236.

(gdb) continue

Thread 1 "hello_image" hit Breakpoint 17, com.oracle.svm.test.debug.CStructTests::testMixedArguments(java.lang.String*, short, com.oracle.svm.test.debug.CStructTests$SimpleStruct, long, java.lang.String*, com.oracle.svm.test.debug.CStructTests$SimpleStruct2, java.lang.String*) (this=0xfffff7ca8040, m1=0x21f69b8, s=1, ss1=0x2db5050, l=123456789, m2=0x21f69d8, ss2=0x2db5070, m3=0x21f69f8) at com/oracle/svm/test/debug/CStructTests.java:236
236	        System.out.println("You find " + m1);
(gdb) info args
this = 0xfffff7ca8040
m1 = 0x21f69b8
s = 1
ss1 = 0x2db5050
l = 123456789
m2 = 0x21f69d8
ss2 = 0x2db5070
m3 = 0x21f69f8

(gdb) p m1->value->data
$15 = 0x1a7ac58 "a message in a bottle"

(gdb) p m2->value->data
$16 = 0x1a7ac30 "a ship in a bottle"
  . . .
```

This also shows that the argument info at the initial instruction is complete and correct.

Although my patch fixes the test and proves that the initial arg info is correct it does not fix the breakpoint placement problem. In particular, users can only benefit from the availability of param info if they know that they need to place a break on the initial instruction by address rather than by simply naming the target method. Ideally we should try to generate aarch64 debug info that encourages gdb to place a method breakpoint at the first instruction rather than at some point where it thinks the prologue is complete and the method is ready to be debugged. There was at one point an attempt to work around this break placement problem by posting a repeated initial line info records for the first instruction. That got removed some while ago when the code was refactored to include initial variable bindings.

I would prefer to push the extra patch as part of this PR. However, I could postpone it to a follow-up PR or, possibly, try to find a way of rendering it unnecessary by getting gdb to place the method breakpoint on the initial instruction. What would you like me to do?

pejovica:
@adinn I think we should address the placement of breakpoints on AArch64 in a separate issue, and for this PR we should merge your patch that fixes the test as a workaround until we find a proper solution.

adinn:
@pejovica This version of the debuginfo test script should fix the aarch64 failure.

adinn:
@olpaw I have squashed the commits to an appropriate subset and reworded them to include a leading upper case word.

PR #7265, opened 2023-08-23
[GR-46420] Use cgroup support from JDK.

graalvmbot:


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/Containers.java
```diff
@@ -110,64 +104,28 @@ public static int activeProcessorCount() {
...
-            if (info.isContainerized()) {
-                long memoryLimit = info.getMemoryLimit();
-                if (memoryLimit > 0) {
-                    return memoryLimit;
-                }
+            Metrics metrics;
+            try {
+                metrics = Container.metrics();
+            } catch (StackOverflowError e) {
+                throw VMError.shouldNotReachHere("Could not get container metrics, likely due to using NIO in the container code of the JDK (JDK-8309191).", e);
```
jerboaa, line 123: https://github.com/openjdk/jdk/pull/15416 should fix this on the JDK side.

jerboaa (approved):
This seems like a good move. I'll give it a spin for further testing...

PR #7308, opened 2023-08-29
Fix typo in Shape.newBuilder().layout()

duncanmak:
The other reference to the method is correct, but the call to `layout` was missing here.


olyagpl:
@duncanmak, your commit is merged, see [DynamicObjectModel.md](https://github.com/oracle/graal/blob/master/truffle/docs/DynamicObjectModel.md). Thank you for the correction!

PR #7406, opened 2023-09-13
[GR-48529] Use the API flag in the init hint.

graalvmbot:


zakkak:
@vjovanov @gilles-duboscq is this going to get backported to 23.1 as well?

Issue #7421, opened 2023-09-14
Problem with  mangling of primitive types in BFD name mangler


adinn:
A problem was introduced when the debug info generator introduced support for Java method symbol name mangling on Linux that is compatible with the BFD library used by the Linux debugger and tools, gdb, perf and valgrind. The problem occurs in the mangling of methods with boolean, byte or char arguments e.g.

```
void java.util.Arrays::fill(byte[]*, int, int, byte)
java.lang.String[]* java.lang.String::split(java.lang.String*, int, boolean)
```

The mangling algorithm encodes primitive argument types using the single letter encoding employed by gcc for the equivalent integral or floating types e.g. int -> i, float -> f etc. This is fine for most primitive types but does not work for `boolean`, `byte` or `char`. The above methods are mangled, respectively, as

```
_ZN16java.util.Arrays4fillEJvP6byte[]iib
_ZN16java.lang.String5splitEJP18java.lang.String[]PS_ic
```

which then demangle as follows

```
$ c++filt _ZN16java.util.Arrays4fillEJvP6byte[]iic
void java.util.Arrays::fill(byte[]*, int, int, char)
$ c++filt _ZN16java.lang.String5splitEJP18java.lang.String[]PS_ib
java.lang.String[]* java.lang.String::split(java.lang.String*, int, bool)
```

The token `c` at the end of the first mangled name identifies a signed 8 bit parameter but the print name for that C++ type in the demangled output is `char` which, in the context of a Java name, is very confusing.

Likewise the token `b` at the end of the second identifies a boolean parameter but the C++ print name for that type in the demangled output is `bool` rather than the expected Java name `boolean`.

The token `t` used to denote an unsigned 16 bit parameter decodes to the token `unsigned short` rather than the desired `char`.

The mangler should be updated so that the encoded names for primitive types decode to the correct Java names.

zakkak:
> when the debug info generator introduced support for Java method symbol name mangling on Linux that is compatible with the BFD library used by the Linux debugger and tools, gdb, perf and valgrind

FTR the support was added with https://github.com/oracle/graal/pull/4950

oubidar-Abderrahim:
Thank you for reporting this, we'll take a look into it shortly
Tracked internally on GR 48875

PR #7423, opened 2023-09-14
[GR-49402] Correct mangling of primitive types in BFD name mangler

adinn:
This PR fixes issue #7421 (q.v.)

The issue is susceptible to a simple fix and a more sophisticated fix as outlined below. The current patch implements the more sophisticated patch in order to minimize symbol footprint

## Basic fix

The simplest solution to the problem is to encode primitive type name using embedded symbol names. So, given the two examples cited in the issue

```
void java.util.Arrays::fill(byte[]*, int, int, byte)
java.lang.String[]* java.lang.String::split(java.lang.String*, int, boolean)
```

the encodings would be

```
_ZN16java.util.Arrays4fillEJvP6byte[]ii4byte
_ZN16java.lang.String5splitEJP18java.lang.String[]PS_i7boolean
```

which successfully decode as desired

```
$ c++filt _ZN16java.util.Arrays4fillEJvP6byte[]ii4byte
void java.util.Arrays::fill(byte[]*, int, int, byte)
$ c++filt _ZN16java.lang.String5splitEJP18java.lang.String[]PS_i7boolean
java.lang.String[]* java.lang.String::split(java.lang.String*, int, boolean)
```

Note that the names `4byte`, `7boolean` and `4char` refer to typedefs for primitive types defined in the debug info section so analysis tools which process these names will still be able to resolve them to appropriate machine layouts.

## Better fix

The problem with the basic fix is that it significantly enlarges symbol names when signatures include repeat occurrences of the same primitive type:

```
int jdk.internal.misc.Unsafe::makeInt(byte, byte, byte, byte)
  --> _ZN24jdk.internal.misc.Unsafe7makeIntEJi4byte4byte4byte4byte

$ c++filt _ZN24jdk.internal.misc.Unsafe7makeIntEJi4byte4byte4byte4byte
int jdk.internal.misc.Unsafe::makeInt(byte, byte, byte, byte)
````

A better solution profits from the use of short names to refer to elements that have already been encoded.

In the above case the encoding of the class name(space) as part of the method name introduces the short name `S_` for `jdk.internal.misc.Unsafe`. Likewise, encoding of the first parameter type introduces short name `S0_` for type name `byte`. The latter binding meant that repeated occurrences of `4byte` can be replaced with `S0_`.

```
int jdk.internal.misc.Unsafe::makeInt(byte, byte, byte, byte)
  --> _ZN24jdk.internal.misc.Unsafe7makeIntEJi4byteS0_S0_S0_

$ c++filt _ZN24jdk.internal.misc.Unsafe7makeIntEJi4byteS0_S0_S0_
int jdk.internal.misc.Unsafe::makeInt(byte, byte, byte, byte)
````

Implementing short name substitution only provides a small gain in footprint when encoding primitive symbol names (2 bytes per `char` or `byte` parameter, 5 per `boolean` parameter). However, it can shorten symbols much more significantly when object type names are repeated. For example without short name substitution the following encoding arises for method `Arrays::deepEquals`:

```
boolean java.util.Arrays::deepEquals(java.lang.Object[]*, java.lang.Object[]*)
  --> _ZN16java.util.Arrays10deepEqualsEJ7booleanP18java.lang.Object[]P18java.lang.Object[]

$ c++filt _ZN16java.util.Arrays10deepEqualsEJ7booleanP18java.lang.Object[]P18java.lang.Object[]
boolean java.util.Arrays::deepEquals(java.lang.Object[]*, java.lang.Object[]*)
```

The repeated `Object[]*` argument can be encoded using a short name as follows:

```
boolean java.util.Arrays::deepEquals(java.lang.Object[]*, java.lang.Object[]*)
  --> _ZN16java.util.Arrays10deepEqualsEJ7booleanP18java.lang.Object[]S2_

$ c++filt _ZN16java.util.Arrays10deepEqualsEJ7booleanP18java.lang.Object[]S2_
boolean java.util.Arrays::deepEquals(java.lang.Object[]*, java.lang.Object[]*)
```

providing a much greater saving in footprint (18 bytes).

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
@@ -279,16 +279,35 @@ class NativeImageBFDNameProvider implements UniqueShortNameProvider {
          *
          * Primitive Parameter Types
          *
-         * Primitive parameter types (or return types) are encoded using a single letter as follows:
+         * Primitive parameter types (or return types) may be encoded using a single letter cases
+         * where the symbol encodes a C++ primitive type with the same name and bit layout and
+         * interpretation as the Java type:
          *
-         *   bool -> b
-         *   byte -> a
          *   short -> s
-         *   char -> t
          *   int -> i
          *   long -> l
          *   float -> f
          *   double -> d
+         *   void -> v
+         *
+         * Other primitive types need to be encoded as symbols
+         *
+         *   boolean -> 7boolean
+         *   byte -> 4byte
+         *   char -> 4char
+         *
+         * In these latter cases the single letter encodings that identifies a C++ type with the
+         * same bit layout and interpretation (respectively, b, c and t) encode for differently
+         * named C++ type (respectively "bool", "char", and unsigned short). Their use would cause
+         * method signatures to be printed with a wrong and potentially misleading name e.g. if
+         * method void Foo::foo(boolean, byte, char) were encoded as _ZN3Foo3fooEJvbct it would
+         * then demangle as Foo::foo(bool, char, unsigned short).
+         *
+         * It would be possible to encode the name "char" using symbol c. However, that would
+         * suggest that the type is a signed 8 bit value rather than a 16 bit unsigned value.
+         * Note that the info section includes appropriate definitions of these three primitive
+         * types with the appropriate names so tools which attempt to resolve the symbol names to
+         * type info should work correctly.
          *
          * Object parameter types (which includes interfaces and enums) are encoded using the class
          * name encoding described above preceded by  prefix 'P'.
@@ -328,28 +347,46 @@ class NativeImageBFDNameProvider implements UniqueShortNameProvider {
          *
          * Substitutions:
          *
-         * Namespace prefix elements (i.e. the leading elements of an N...E encoding) are indexed and
-         * can be referenced using a shorthand index, saving space in the symbol encoding. In the
-         * standard Java encoding this means that common package and class name prefixes can be
-         * independently substituted (also template prefixes).
+         * Symbol elements are indexed and can be referenced using a shorthand index, saving
+         * space in the symbol encoding.
          *
-         * For example, when encoding the first parameter type, the standard Java encoding of method
-         * startsWith of class java.lang.String can refer to the common prefixes java, lang and String
-         * established by the namespace encoding of the method name:
+         * For example, consider C++ equivalent of String.compareTo and its corresponding encoding
          *
-         *   boolean startsWith(java.lang.String, int) -> _ZN4java4lang6String10startsWithEJbPN4java4lang6StringEi
-         *      -> _ZN4java4lang6String10startsWithEJbPN$_$0_$1_Ei
+         *   int java::lang::String::compareTo(java::lang::String)
+         *      -> _ZN4java4lang6String9compareToEJiPS1_
          *
-         * The namespace prefixes 4java, 4lang and 6String establish successive bindings for the
-         * indexed substitution variables $_, $0_ and $1_.
+         * The symbol encodings 4java, 4lang and 6String and 9compareTo establish
+         * successive bindings for the indexed substitution variables S_, S0_, S1_ and S2_ to
+         * the respective names java, java::lang, java::lang::String and java::lang::String::compareTo
+         * (note that bindings accumulate preceding elements when symbols occur inside a namespace).
+         * The encoding of the argument list uses S1_ references the 3rd binding i.e. the class
+         * name. The problem with this, as seen before when using namespaces is that the demangler
+         * accumulates names using a '::' separator between the namespace components rather than the
+         * desired  '.' separator.
          *
-         * The Graal encoding makes much more limited use of namespace prefixes but it can still profit
-         * from them to produce more concise encodings:
+         * The Graal encoding can work around the namespace separator as shown earlier and still employ
+         * symbols as prefixes in order to produce more concise encodings. The full encoding for the
          *
-         *   boolean startsWith(java.lang.String, int) -> _ZN16java.lang.String10startsWithEJbP16java.lang.Stringi
-         *      -> _ZN16java.lang.String10startsWithEJbP$_i
+         *   int String.compareTo(java.lang.String)
+         *      -> _ZN16java.lang.String9compareToEJiP16java.lang.String
          *
-         * Indexed prefix references are encoded as "S_", "S0_", ... "S9_", "SA_", ... "SZ_", "S10_", ...
+         * However, it is also possible to encode this more concisely as
+         *
+         *   int String.compareTo(java.lang.String)
+         *      -> _ZN16java.lang.String9compareToEJiPS_
+         *
+         * S_ translates to the first symbol introduced in the namespace i.e. java.lang.String.
+         *
+         * Substitutions can also occur when parameter types are repeated e.g.
+         *
+         * int Arrays.NaturalOrder.compare(Object first, Object second)
+         *      -> _ZN19Arrays$NaturalOrder7compareEJiPP16java.lang.ObjectPS_2
+         * 
+         * In this case the class name symbol 19Arrays$NaturalOrde binds $_ to Arrays$NaturalOrder,
+         * the method name symbol 7compare binds $1_ to Arrays$NaturalOrder::compareTo and the
+         * first parameter type name symbol 16java.lang.Object binds $2_ to java.lang.Object.
+         *
+         * Indexed symbol references are encoded as "S_", "S0_", ... "S9_", "SA_", ... "SZ_", "S10_", ...
          * i.e. after "$_" for index 0, successive encodings for index i embded the base 36 digits for
          * (i - 1) between "S" and "_".
          */
@@ -383,14 +420,166 @@ class NativeImageBFDNameProvider implements UniqueShortNameProvider {
     }
 
     private static class BFDMangler {
+
+        /*
+         * The mangler tracks certain elements as they are inserted into the mangled name and bind
+         * short symbol name to use as substitutions. If the same element needs to be re-inserted
+         * the mangler can embed the short symbol instead of generating the previously mangled full
+         * text. Short symbol names are bound in a well-defined sequence at well-defined points
+         * during mangling. This allows the demangler to identify exactly which text in the previous
+         * input must be used to replace a short symbol name.
+         *
+         * A binding occurs whenever a simple name is mangled outside of a namespace. For example, a
+         * top level class name mangled as 3Hello will bind the next available short name with the
+         * result that it will demangle to Hello and can be used to replace later occurrences of the
+         * string Hello.
+         *
+         * When a sequence of simple names is mangled inside a namespace substitution bindings are
+         * recorded for each successive composite namespace prefix but not for the final symbol
+         * itself. For example, when method name Hello::main is mangled to N5Hello4main4 a single
+         * binding is recorded which demangles to Hello. If a class has been loaded by an
+         * application loader and, hence, has a name which includes a loader namespace prefix (to
+         * avoid the possibility of the same named class being loaded by two different loaders) then
+         * the method name, say AppCL504::Hello::main, would mangle to a namespace encoding with 3
+         * elements N8AppCL5045Hello4mainE which would introduce two bindings, the first demangling
+         * to AppCL504 and the second to AppCL504::Hello.
+         *
+         * A binding is also recorded whenever a pointer type is mangled. The bound symbol demangles
+         * to whatever the text decoded from the scope of the P followed by a '*' to translate it to
+         * a pointer type. For example, when the type in parameter list (Foo*) is encoded as P3Foo a
+         * binding is recorded for the pointer as well as for the type Foo. The first binding
+         * demangles to Foo. The second binding demangles to Foo*.
+         *
+         * n.b. repeated use of the pointer operator results in repeated bindings. So, if a
+         * parameter with type Foo** were to be encoded as PP3Foo three bindings would recorded
+         * which demangle to Foo, Foo* and Foo**. However, multiple indirections do not occur in
+         * Java signatures.
+         *
+         * n.b.b. repeated mentions of the same symbol would redundantly record a binding. For
+         * example if int Hello::compareTo(Hello*) were mangled to _ZN5Hello9compareToEJiP5Hello the
+         * resulting bindings would be S_ ==> Hello, S0_ ==> Hello::compareTo, S1_ ==> Hello and S2_
+         * ==> Hello* i.e. both S_ and S1_ would demangle to Hello. This situation should never
+         * arise. A name can always be correctly encoded without repeats. In the above example that
+         * woudl be _ZN5Hello9compareToEJiPS_.
+         */
         final NativeImageBFDNameProvider nameProvider;
         final StringBuilder sb;
-        final List<String> prefixes;
+
+        // A list of lookup names identifying substituted elements. A prospective name for an
+        // element that is about to be encoded can be looked up in this list. If a match is found
+        // the list index can be used to identify the relevant short symbol. If it is not found
+        // then inserting the name serves to allocate the short name associated with the inserted
+        // element's result index.
+        List<LookupName> bindings;
+
+        /**
+         * A lookup name is used as a key to record and subsequently lookup a short symbol that can
+         * replace one or more elements of a mangled name. A lookup name is usually just a simple
+         * string, i.e. some text that is to be mangled and to which the corresponding short symbol
+         * should decode. However, substitutions for namespace prefixes (e.g. AppLoader506::Foo)
+         * require a composite lookup name that composes a namespace prefix (AppLoader506) with a
+         * trailing simple name (Foo). The corresponding short symbol will demangle to the string
+         * produced by composing the prefix and simple name with a :: separator (i.e. restoring
+         * AppLoader506::Foo). Short symbols for pointer types (e.g. Foo* or AppLoader506::Foo*)
+         * require a pointer lookup name that identifies the substituted text as a pointer to some
+         * underlying type (e.g. Foo or AppLoader506::Foo). The corresponding short symbol will
+         * demangle to the string produced by demangling the underlying type with a * suffix. In
+         * theory the prefix for a pointer type lookup name might be defined by another pointer type
+         * lookup name (if, say, we needed to encode type Foo**). In practice, that case should not
+         * arise with Java method signatures.
+         */
+        private abstract class LookupName {
+            String value;
+
+            protected LookupName(String value) {
+                assert value != null;
+                this.value = value;
+            }
+
+            @Override
+            public abstract String toString();
+        }
+
+        private class SimpleLookupName extends LookupName {
+            SimpleLookupName(String value) {
+                super(value);
+            }
+
+            @Override
+            public boolean equals(Object other) {
+                if (other != null && other instanceof SimpleLookupName) {
+                    return this.value.equals(((SimpleLookupName) other).value);
+                }
+                return false;
+            }
+
+            @Override
+            public int hashCode() {
+                return value.hashCode() + 29;
+            }
+
+            @Override
+            public String toString() {
+                return value;
+            }
+        }
+
+        private abstract class CompositeLookupName extends LookupName {
+            LookupName tail;
+
+            CompositeLookupName(String value, LookupName tail) {
+                super(value);
+                assert tail != null;
+                // it makes sense to concatenate Foo with Bar or Foo::Bar with Baz and
+                // Foo* with * but we not Foo* with Bar or Foo::Bar* with Baz
+                // assert !(tail instanceof PointerLookupName) || this instanceof PointerLookupName;
+                this.tail = tail;
+            }
+
+            public boolean equals(Object other) {
+                if (other != null && other instanceof NamespaceLookupName) {
+                    CompositeLookupName otherCompositeLookupName = (CompositeLookupName) other;
+                    assert tail != null && otherCompositeLookupName.tail != null;
+                    return value.equals(otherCompositeLookupName.value) && tail.equals(otherCompositeLookupName.tail);
+                }
+                return false;
+            }
+
+            @Override
+            public int hashCode() {
+                return value.hashCode() ^ tail.hashCode() + 37;
+            }
+        }
+
+        private class NamespaceLookupName extends CompositeLookupName {
+            NamespaceLookupName(String value, LookupName tail) {
+                super(value, tail);
+            }
+
+            @Override
+            public String toString() {
+                return value + "::" + tail.toString();
+            }
+        }
+
+        private class PointerLookupName extends NamespaceLookupName {
+            PointerLookupName(LookupName prefix) {
+                // use * as the value to compose with the prefix because it cannot
+                // arise as a simple name
+                super("*", prefix);
+            }
+
+            @Override
+            public String toString() {
+                // don't insert a '::' separator when composing
+                return tail.toString() + value;
+            }
+        }
 
         BFDMangler(NativeImageBFDNameProvider provider) {
             nameProvider = provider;
             sb = new StringBuilder("_Z");
-            prefixes = new ArrayList<>();
+            bindings = new ArrayList<>();
         }
 
         public String mangle(String loaderName, ResolvedJavaType declaringClass, String memberName, Signature methodSignature, boolean isConstructor) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by zakkak** [comment]
> A better solution profits from the use of short names to refer to elements that have already been encoded.

FWIW, this is documented in https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-compression

File: substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java [hunk 1]
```diff
@@ -383,14 +420,166 @@ public void setNativeLibs(NativeLibraries nativeLibs) {
...
+         * which demangle to Foo, Foo* and Foo**. However, multiple indirections do not occur in
+         * Java signatures.
+         *
+         * n.b.b. repeated mentions of the same symbol would redundantly record a binding. For
+         * example if int Hello::compareTo(Hello*) were mangled to _ZN5Hello9compareToEJiP5Hello the
+         * resulting bindings would be S_ ==> Hello, S0_ ==> Hello::compareTo, S1_ ==> Hello and S2_
+         * ==> Hello* i.e. both S_ and S1_ would demangle to Hello. This situation should never
+         * arise. A name can always be correctly encoded without repeats. In the above example that
+         * woudl be _ZN5Hello9compareToEJiPS_.
+         */
```
zakkak, line 464: Perhaps it's worth mentioning that the above are based on https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-compression and make clear that GraalVM doesn't really create any bindings/lookup-tables in the debug info sections.

zakkak (approved):
LGTM, I added a couple of comments/questions that might help improve the PR.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -383,14 +420,166 @@ public void setNativeLibs(NativeLibraries nativeLibs) {
...
+
+        private abstract class CompositeLookupName extends LookupName {
+            LookupName tail;
+
+            CompositeLookupName(String value, LookupName tail) {
+                super(value);
+                assert tail != null;
+                // it makes sense to concatenate Foo with Bar or Foo::Bar with Baz and
+                // Foo* with * but we not Foo* with Bar or Foo::Bar* with Baz
+                // assert !(tail instanceof PointerLookupName) || this instanceof PointerLookupName;
```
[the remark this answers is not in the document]
adinn: Not obsolete, just unnecessary. I'll remove them.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -441,30 +630,70 @@ public String mangle(Member member) {
...
+            // this should only be called when inserting a sequence into a namespace
+            assert sb.charAt(sb.length() - 1) == 'N';
+            // we can substitute both symbols
+            mangleWriteSubstitutableNameRecord(prefix);
+            // in theory the trailing simple name may be substitutable but cannot
+            // have an associated binding. in practice gdb will not translate
+            // a substitution in this position even though the demangler will
+
+            // mangleWriteSubstitutableNameNoRecord(name);
+            mangleWriteSimpleName(name);
```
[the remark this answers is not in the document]
adinn: No, that comment is correct. The demangler will actually translate a substitution if it is inserted at the end of a namespace e.g.
```
$ c++filt _ZN3foo3barS_E
foo::bar::foo
```

However, gdb will barf on the $_ and refuse to translate the whole symbol. So, in order to keep gdb happy we have to generate _ZN3foo3bar3fooE.
The commented out call reflect the fact that it ought to be ok here to call `mangleWriteSubstitutableNameNoRecord(name)` but in fact we have to call `mangleWriteSimpleName(name)`. I will explicitly spell that out in the comment.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -441,30 +630,70 @@ public String mangle(Member member) {
...
+                mangleWriteSubstitutableNameRecord(prefix1);
+                // we cannot establish a binding for the trailing prefix
+                mangleWriteSubstitutableNameNoRecord(prefix2);
+                recordName(lookup);
+            }
+            // in theory the trailing simple name may be substitutable but cannot
+            // have an associated binding. in practice gdb will not translate
+            // a substitution in this position even though the demangler will
+
+            // mangleWriteSubstitutableNameNoRecord(name);
```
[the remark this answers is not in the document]
adinn: Same as above

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: Unfortunately, it is not based on that document. That document merely reflects some aspects of the implementation (and falls a little short as a specification). The correct reference would be to a GNU g++ spec (which I have not yet found) or, in its absence, to the BFD source code (which can be found in both the GNU binutils and gdb trees).

I don't think there is a need for any specific reference. The gdb source tree is definitive for this and many other aspects of the way the Linux debug info generator works and really needs to be read to understand and verify many details of the implementation. I summarized the relevant points here rather than referenced the gdb code because we only need to consider a subset of the encoding scheme when it comes to Java i.e. this is just a TL;DR for the code.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -383,14 +420,166 @@ public void setNativeLibs(NativeLibraries nativeLibs) {
...
+                    CompositeLookupName otherCompositeLookupName = (CompositeLookupName) other;
+                    assert tail != null && otherCompositeLookupName.tail != null;
+                    return value.equals(otherCompositeLookupName.value) && tail.equals(otherCompositeLookupName.tail);
+                }
+                return false;
+            }
+
+            @Override
+            public int hashCode() {
+                return value.hashCode() ^ tail.hashCode() + 37;
```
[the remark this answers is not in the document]
zakkak: Oh, that makes sense.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -383,14 +420,163 @@ public void setNativeLibs(NativeLibraries nativeLibs) {
...
+        private class SimpleLookupName extends LookupName {
+            SimpleLookupName(String value) {
+                super(value);
+            }
+
+            @Override
+            public boolean equals(Object other) {
+                if (other instanceof SimpleLookupName) {
+                    return this.value.equals(((SimpleLookupName) other).value);
+                }
```
zakkak, line 512: TIL (so I thought you might want to learn some ["new", since Java 14, tricks](https://openjdk.org/jeps/305) as well) you can do 
```suggestion
                if (other instanceof SimpleLookupName simpleLookupName) {
                    return this.value.equals(simpleLookupName.value);
                }
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -383,14 +420,163 @@ public void setNativeLibs(NativeLibraries nativeLibs) {
...
+            LookupName tail;
+
+            CompositeLookupName(String value, LookupName tail) {
+                super(value);
+                assert tail != null;
+                this.tail = tail;
+            }
+
+            public boolean equals(Object other) {
+                if (other instanceof NamespaceLookupName) {
```
zakkak, line 537: Shouldn't this be `CompositeLookupName`?

```suggestion
                if (other instanceof CompositeLookupName) {
```

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: Oh my, I am an old dog and find these new tricks so very challenging . . .

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: Well caught!

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
zakkak: If it makes you feel any better I find it hard to keep track of these as well  (especially remembering in which version each trick works) despite being a younger dog :)

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -279,16 +279,35 @@ public String bfdMangle(String loaderName, ResolvedJavaType declaringClass, Stri
          *
          * Primitive Parameter Types
          *
-         * Primitive parameter types (or return types) are encoded using a single letter as follows:
+         * Primitive parameter types (or return types) may be encoded using a single letter cases
+         * where the symbol encodes a C++ primitive type with the same name and bit layout and
```
olpaw, line 283: using a single letter ~cases~ where

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -383,14 +420,163 @@ public void setNativeLibs(NativeLibraries nativeLibs) {
...
+            protected LookupName(String value) {
+                assert value != null;
+                this.value = value;
+            }
+
+            @Override
+            public abstract String toString();
+        }
+
+        private class SimpleLookupName extends LookupName {
```
olpaw, line 503: If you make SimpleLookupName, CompositeLookupName and NamespaceLookupName Java `record`s you get the hashCode, equals and toString implementations for free.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -441,30 +627,75 @@ public String mangle(Member member) {
...
+                mangleWriteSubstitutableNameNoRecord(prefix2);
+                recordName(lookup);
+            }
+            // see comment in previous method as to why we call mangleWriteSimpleName
+            // instead of mangleWriteSubstitutableNameNoRecord(name)
+            mangleWriteSimpleName(name);
+        }
+
+        private boolean substituteName(LookupName name) {
+            int index = bindings.indexOf(name);
```
olpaw, line 696: Linear search ... are we sure that this does not explode (computationally)

adinn: The lookup cost here will be o(N^2) where N is the number of substitutable elements in the name and signature. N accumulates at most

  1 for the package (namespace) prefix
  2 for an object return type
  2 for an object or (any) array type argument
  1 if we have a boolean argument
  1 if we have a char argument
  1 if we have a byte argument

The count for an object argument/return type can be lowered if

  the object type is a foreign type (not passed as a pointer) - subtract 1
  the object type repeats the namespace prefix (e.g. java.lang.String) - subtract 1
  the object type repeats a previous occurrence - subtract 2

I added a temporary tweak to the patch to count the bindings length for each symbol encoded and tested it by building the image used for the debuginfo test. The worst case I found was this one:
 
```
java.util.Collection*
sun.security.provider.certpath.DistributionPointFetcher::getCRLs(java.security.cert.X509CRLSelector*,
sun.security.x509.X509CertImpl*, sun.security.x509.DistributionPoint*, boolean[]*,
boolean, java.security.PublicKey*, java.security.cert.X509Certificate*, java.lang.String*,
java.util.List*, java.util.Set*, java.util.Date*, java.lang.String*, java.security.cert.TrustAnchor*)
```
Encoding it generates 26 bindings. It has a lot of args and the only repeat is the second occurrence of `java.lang.String*` which appears in the encoding as `SG_` i.e. binding 18.

That case is very much an outlier. Normally the number of bindings is quite small. Here is a table summarizing the distribution for the 65667 different encodings generated during generation of one of the test images:

|Bindings Count | Frequency  | %ge |
|-:|-:|-:|
|1 | 12721| 19.4|
|2 | 7482| 11.4|
|3 | 24745| 37.7|
|4 | 6049| 9.2|
|5 | 8164| 12.4|
|6 | 2509| 3.8|
|7 | 2578| 3.9|
|8 | 358| 5.4|
|9 | 665| 1.0|
|10 | 108| 0.1|
|11-15 | 260|0.39|
|16-20 | 24| ~0.0|
|20-25 | 3| ~0.0|
|26 | 1| ~0.0|

I don't think this means we must switch to a constant time lookup model such as an EconomicMap but the trade off is not clear. What do you think?

olpaw: Thanks for the detailed breakdown.

Of course one can think of pathological methods (potentially from generated source code or test suites) that could be much worse than what you found. The question is if it's practically relevant. I would suggest you build a large test image (huge quarkus or springboot native app ... @zakkak might be able to help with that) and check if you can see any real world build time regressions that are distinguishable from noise when building with `-g`.

zakkak: I could even use the "temporary tweak to the patch to count the bindings length" to report some more stats, other than the build times. @adinn would you mind sharing that patch, to make sure we are reporting the same thing?

galderz:
> FWIW, this is documented in https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-compression

I looked at @adinn's examples and the documentation and couldn't make sense why `S2_` would convert to `java.lang.Object[]*`. I initially thought `S0_` would be enough but not true. I found a better explaination in https://github.com/gchatelet/gcc_cpp_mangling_documentation. That one explains why `S0_` is `boolean`, but I still had doubts why `S1_` would not do it. So, here's some further clarification:

```
$ c++filt "_ZN16java.util.Arrays10deepEqualsEJ7booleanP18java.lang.Object[]S1_"
boolean java.util.Arrays::deepEquals(java.lang.Object[]*, java.lang.Object[])
```

`java.lang.Object[]` is a symbol per se:

```
$ c++filt "_Z18java.lang.Object[]"
java.lang.Object[]
```

So the fact that is a parameter makes an additional symbol FWIU.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
oraluben: It seems the only GNU's extension is the `J` for return type, other part follows the itanium's spec (not sure, at least respected by llvm and gcc)

https://github.com/gcc-mirror/gcc/blob/12039c9f40a81605e59951f80007827bc0413573/libiberty/cp-demangle.c#L3059

I also didn't find llvm's c++filt supporting mangling return type, does that mean debug information on MacOS is infeasible?

adinn: The BFD name mangling support is implemented for Linux in order support debugging using gdb and use of Linux analysis tool such as perf and valgrind. These programs expect to see return types in mangled names. I don't know what LLVM expects on Linux. If it cannot recognize return types then it is goign to have a problem debugging librraies generated by gcc/g++.

Debug info generation is not currently available for MacOS. So, whether the current code works correctly on MacOS is not an issue. If we want to reuse some of the current debug generator code on MacOS and that requires name mangling, for llvm or for other tools, then it may be necessary to implement a new mangler. If MacOS tools will accept GNUv3 mangled names without return types then it might be possible to customize the current mangler to cater for that. However, I believe there are cases (synthetic bridge methods) where two distinct Java methods are only differentiated by return type. In that case omitting the return type would map both methods to the same mangled name. So, that problme amy need ot be finessed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: @zakkak Not much to share really. The temporary tweak was simply a call to `System.out.format()` at the end of `BFDMangle::mangle(String, ResolvedJavaType, String, Signature, boolean)`, printing the resulting mangled name and the length of `bindings`. The rest of the work was done using shell commands on the resulting output.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -328,28 +347,46 @@ public String bfdMangle(String loaderName, ResolvedJavaType declaringClass, Stri
...
          *
-         *   boolean startsWith(java.lang.String, int) -> _ZN16java.lang.String10startsWithEJbP16java.lang.Stringi
-         *      -> _ZN16java.lang.String10startsWithEJbP$_i
+         * Substitutions can also occur when parameter types are repeated e.g.
          *
-         * Indexed prefix references are encoded as "S_", "S0_", ... "S9_", "SA_", ... "SZ_", "S10_", ...
+         * int Arrays.NaturalOrder.compare(Object first, Object second)
+         *      -> _ZN19Arrays$NaturalOrder7compareEJiPP16java.lang.ObjectPS_2
+         * 
+         * In this case the class name symbol 19Arrays$NaturalOrde binds $_ to Arrays$NaturalOrder,
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -383,14 +420,166 @@ public void setNativeLibs(NativeLibraries nativeLibs) {
...
+                super(value);
+                assert tail != null;
+                // it makes sense to concatenate Foo with Bar or Foo::Bar with Baz and
+                // Foo* with * but we not Foo* with Bar or Foo::Bar* with Baz
+                // assert !(tail instanceof PointerLookupName) || this instanceof PointerLookupName;
+                this.tail = tail;
+            }
+
+            public boolean equals(Object other) {
+                if (other != null && other instanceof NamespaceLookupName) {
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -383,14 +420,166 @@ public void setNativeLibs(NativeLibraries nativeLibs) {
...
+        }
+
+        private class SimpleLookupName extends LookupName {
+            SimpleLookupName(String value) {
+                super(value);
+            }
+
+            @Override
+            public boolean equals(Object other) {
+                if (other != null && other instanceof SimpleLookupName) {
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -383,14 +420,166 @@ public void setNativeLibs(NativeLibraries nativeLibs) {
...
+                    return this.value.equals(((SimpleLookupName) other).value);
+                }
+                return false;
+            }
+
+            @Override
+            public int hashCode() {
+                return value.hashCode() + 29;
```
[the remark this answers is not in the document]
adinn: Done

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
zakkak: @adinn this is a kind reminder

zakkak: Well it looks like I was looking at a cached version of this when I pinged you :/

Thanks for the reply @adinn , I will try to get some number (hopefully this week)

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -375,22 +412,113 @@ public String bfdMangle(Member m) {
...
+         * parameter with type Foo** were to be encoded as PP3Foo three bindings would recorded
+         * which demangle to Foo, Foo* and Foo**. However, multiple indirections do not occur in
+         * Java signatures.
+         *
+         * n.b.b. repeated mentions of the same symbol would redundantly record a binding. For
+         * example if int Hello::compareTo(Hello*) were mangled to _ZN5Hello9compareToEJiP5Hello the
+         * resulting bindings would be S_ ==> Hello, S0_ ==> Hello::compareTo, S1_ ==> Hello and S2_
+         * ==> Hello* i.e. both S_ and S1_ would demangle to Hello. This situation should never
+         * arise. A name can always be correctly encoded without repeats. In the above example that
+         * woudl be _ZN5Hello9compareToEJiPS_.
```
olpaw, line 463: typo

adinn: Thanks for re-reviewing, Paul. I fixed this and also squashed the style fixes into the main patch and review rework patches.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -375,22 +412,113 @@ public String bfdMangle(Member m) {
...
      * @param nativeLibs the current native libraries singleton.
      */
     public void setNativeLibs(NativeLibraries nativeLibs) {
         this.nativeLibs = nativeLibs;
     }
 
     private static class BFDMangler {
+
+        /*
+         * The mangler tracks certain elements as they are inserted into the mangled name and bind
```
olpaw, line 425: Excellent explanation. Thanks for adding that!

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -375,22 +412,113 @@ public String bfdMangle(Member m) {
...
         final NativeImageBFDNameProvider nameProvider;
         final StringBuilder sb;
-        final List<String> prefixes;
+
+        // A list of lookup names identifying substituted elements. A prospective name for an
+        // element that is about to be encoded can be looked up in this list. If a match is found
+        // the list index can be used to identify the relevant short symbol. If it is not found
+        // then inserting the name serves to allocate the short name associated with the inserted
+        // element's result index.
+        List<LookupName> bindings;
```
olpaw, line 473: As already mentioned. I would prefer a map based approach for recording and lookup of current state of bindings but if @zakkak measurements show that it does not matter in practice I'm fine wit that.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -375,22 +412,113 @@ public String bfdMangle(Member m) {
...
+         * produced by composing the prefix and simple name with a :: separator (i.e. restoring
+         * AppLoader506::Foo). Short symbols for pointer types (e.g. Foo* or AppLoader506::Foo*)
+         * require a pointer lookup name that identifies the substituted text as a pointer to some
+         * underlying type (e.g. Foo or AppLoader506::Foo). The corresponding short symbol will
+         * demangle to the string produced by demangling the underlying type with a * suffix. In
+         * theory the prefix for a pointer type lookup name might be defined by another pointer type
+         * lookup name (if, say, we needed to encode type Foo**). In practice, that case should not
+         * arise with Java method signatures.
+         */
+        private interface LookupName {
```
olpaw, line 491: These two interfaces could turned into sealed interfaces.
(Although not needed right now as you do not switch over them)

See https://docs.oracle.com/en/java/javase/17/language/sealed-classes-and-interfaces.html at _Declaring Sealed Interfaces_.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
```diff
@@ -441,30 +569,75 @@ public String mangle(Member member) {
...
+            mangleWriteSimpleName(name);
+        }
+
+        private void mangleWriteSubstitutablePrefixedName(String prefix1, String prefix2, String name) {
+            // this should only be called when inserting a sequence into a namespace
+            assert sb.charAt(sb.length() - 1) == 'N';
+            // we can substitute the composed prefix followed by the name
+            // or we can substitute all three individual symbols
+            LookupName lookup2 = new SimpleLookupName(prefix2);
+            LookupName lookup = new NamespaceLookupName(prefix1, lookup2);
```
olpaw, line 623: maybe
s/lookup2/simpleLookup
s/lookup/namespaceLookup

olpaw (approved):
Overall LGTM now. Only cosmetic suggestions.

Ready to merge if @zakkak reports that this does not cause performance regression.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: You are welcome. Of course, I'll be thanking myself in 6 months time when I come to review this!

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: I decided to use a map here anyway.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: I have sealed both interfaces.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/image/NativeImageBFDNameProvider.java
adinn: Yes, better. Done.

adinn:
@olpaw Remaining recommendations have been followed. Hopefully this will pass the gate and be ready for inclusion.

fniephaus (approved):

zakkak:
Thanks for the updates @adinn and sorry for not performing the perf runs in time @olpaw, it was a busy week.

Issue #7402, opened 2023-09-12
Support jdk.ObjectCount and jdk.ObjectCountAfterGC


roberttoyonaga:
**Feature request**

Please include the following information:

**Is your feature request related to a problem? Please describe.**
Not related to a problem.

**Describe the solution you'd like.**
Support for these events. They should provide count and memory size per object type after a GC has completed. 

**Describe who do you think will benefit the most.**
GraalVM users that use JFR to debug memory leaks, and allocation related issues.

**Describe alternatives you've considered.**
Heap dumps are an alternative approach to inspecting objects on the heap. However, jdk.ObjectCount and jdk.ObjectCountAfterGC have the advantage of continuously providing a data points over time.  It should also be quicker to emit these events than do a full heap dump. 

**Express whether you'd like to help contributing this feature**
I'll help contribute this

PR #7433, opened 2023-09-15
Add support for JFR jdk.ObjectCount and jdk.ObjectCountAfterGC events

roberttoyonaga:
### Summary
These events provide aggregate data about objects on the heap. They also useful for inspecting how the population of objects on the heap changes over time. For each class, the number of instances and total size is reported. The only difference between these two events is that `JFR jdk.ObjectCount` is periodic and guaranteed to run once per JFR chunk (not on every GC), while `JFR jdk.ObjectCountAfterGC` runs upon every GC. 

### Details
After a GC is completed, the heap is walked and live object data is gathered. This all happens during the GC safepoint. 

### Performance
This is expected to slightly slow down garbage collection and so is off-by-default in JIT mode. I ran an experiment comparing with/without these new events enabled. Hyperfoil was used to make 50 requests/s over 5s to a simple plaintext quarkus app running JFR. See [here](https://github.com/rh-jmc-team/jfr-native-image-performance) for the code. The results of 14 paired trials is shown below:

 **Increase in mean response time after events enabled**
  | Mean response time increase (us) |
-- | -- | 
Mean | 292.4 | 
Median | 226 | 
STDV | 371.092 |
MAX | 1238 |

 **Increase in 99th % response time after events enabled**
  | p99 response time increase (us) |
-- | -- | 
Mean | 6477.2 | 
Median | 6783 | 
STDV | 5112.92 |
MAX | 13566 |

(P99 means 99% of responses were faster)

As you can see, the performance hit is much worse for p99. This is expected because the change only happens during a GC.  RSS usage and start up time were unaffected by enabling these new events.

I also tried explicitly timing how long the `ObjectCountEventSupport.emitEvents()` takes to complete. It appears to take about 2.7-3.1 ms on my machine (based on results from only 4 GCs). 

Related issue: https://github.com/oracle/graal/issues/7402


christianhaeubl (commented):
Thanks for the PR, I added a few comments and questions.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/EveryChunkNativeGCPeriodicEvents.java
```diff
@@ -0,0 +1,53 @@
...
+public class EveryChunkNativeGCPeriodicEvents extends Event {
+
+    public static void emit() {
+        emitObjectCount();
+    }
+
+    @Uninterruptible(reason = "Set and unset should be atomic with invoked GC to avoid races.")
+    private static void emitObjectCount() {
+        if (JfrEvent.ObjectCount.shouldEmit()) {
+            ObjectCountEventSupport.setShouldSendRequestableEvent(true);
```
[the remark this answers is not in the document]
roberttoyonaga: That's a good idea. I've adopted your suggestion.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrTypeRepository.java
```diff
@@ -102,15 +102,20 @@ private TypeInfo collectTypeInfo(boolean flushpoint) {
         return typeInfo;
     }
 
+    boolean isClassGenerated(Class<?> clazz) {
```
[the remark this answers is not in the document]
roberttoyonaga: I've removed this now. Please see responses below.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrTypeRepository.java
```diff
@@ -102,15 +102,20 @@ private TypeInfo collectTypeInfo(boolean flushpoint) {
...
+        return clazz != null && clazz.getPackage() != null && clazz.getPackage().getName().equals("jdk.internal.reflect") && clazz.getName().contains("GeneratedSerializationConstructorAccessor");
```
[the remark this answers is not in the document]
roberttoyonaga: Previously, I needed to determine which classes were generated in order  to adjust the assertions. This was because there were cases where generated classes have the same package as non-generated classes, however their module is different (the unnamed module).

This means there was the possibility for  a single package name (which was being used as the key for deduplication) to correspond to multiple modules.  

In hotspot, no special handling is done for generated classes. Instead, each class has a pointer back to its package instance which then corresponds to a single module. This means there is a one-one relationship instead of one-many. 

To replicate the same behaviour here, we need a unique key which corresponds to exactly one package-module pair.  To do this I've decided to use the concatenation of the concatenate the package and module names.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrTypeRepository.java
```diff
@@ -246,10 +251,14 @@ private boolean isClassVisited(TypeInfo typeInfo, Class<?> clazz) {
         return typeInfo.classes.contains(clazz) || flushedClasses.contains(clazz);
     }
 
-    private boolean addPackage(TypeInfo typeInfo, Package pkg, Module module) {
+    private boolean addPackage(TypeInfo typeInfo, Package pkg, Module module, boolean generated) {
         if (isPackageVisited(typeInfo, pkg)) {
-            assert module == (flushedPackages.containsKey(pkg.getName()) ? flushedPackages.get(pkg.getName()).module : typeInfo.packages.get(pkg.getName()).module);
-            return false;
+            Module cached = (flushedPackages.containsKey(pkg.getName()) ? flushedPackages.get(pkg.getName()).module : typeInfo.packages.get(pkg.getName()).module);
```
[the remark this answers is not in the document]
roberttoyonaga: The logic here was a bit confusing  and has been removed now in favor of what's described in the above response.

Previously, this logic was so that we would only replace the cached module if it is the unnamed module. And we only want to bother replacing it with a named module. This effectively would result in only serializing named modules, or the unnamed module if there was no named module for the given package.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrSymbolRepository.java
```diff
@@ -77,8 +76,6 @@ public long getSymbolId(String imageHeapString, boolean previousEpoch, boolean r
             return 0;
         }
 
-        assert Heap.getHeap().isInImageHeap(imageHeapString);
```
roberttoyonaga, line 80: This assertion was failing due to `com.sun.proxy.jdk.proxy1`. I think this assertion is incorrect because its possible we may sometimes need to serialize symbols not in the image heap.  This was exposed now since the ObjectCount event causes many more symbols to be required. What do you think?

christianhaeubl: That would be bad, then we would need to change `JfrSymbolRepository` significantly. Do you have a stack trace?

roberttoyonaga: If the string is not in the image heap, would it be possible to derive the hash from the symbol string itself rather than from its memory address? 

I added an assertion to `JfrTypeRepository.writePackage`. Then with `mx native-unittest`:
```
21) com.oracle.svm.test.jfr.TestVirtualThreadsJfrStreaming#test
java.lang.AssertionError: Not in image heapcom.sun.proxy.jdk.proxy1
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jfr.JfrTypeRepository.writePackage(JfrTypeRepository.java:179)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jfr.JfrTypeRepository.writePackages(JfrTypeRepository.java:172)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jfr.JfrTypeRepository.write(JfrTypeRepository.java:68)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jfr.JfrChunkWriter.writeConstantPools(JfrChunkWriter.java:354)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jfr.JfrChunkWriter.writeCheckpointEvent(JfrChunkWriter.java:324)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jfr.JfrChunkWriter.writeFlushCheckpoint(JfrChunkWriter.java:296)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jfr.JfrChunkWriter.closeFile(JfrChunkWriter.java:222)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jfr.SubstrateJVM.setOutput(SubstrateJVM.java:363)
        at jdk.jfr@21/jdk.jfr.internal.JVM.setOutput(JVM.java:167)
        at jdk.jfr@21/jdk.jfr.internal.MetadataRepository.setOutput(MetadataRepository.java:295)
        at jdk.jfr@21/jdk.jfr.internal.PlatformRecorder.start(PlatformRecorder.java:267)
        at jdk.jfr@21/jdk.jfr.internal.PlatformRecording.start(PlatformRecording.java:123)
        at jdk.jfr@21/jdk.jfr.consumer.RecordingStream.startAsync(RecordingStream.java:379)
        at com.oracle.svm.test.jfr.JfrStreamingTest.startStream(JfrStreamingTest.java:87)
        at com.oracle.svm.test.jfr.JfrStreamingTest.startStream(JfrStreamingTest.java:73)
        at com.oracle.svm.test.jfr.TestVirtualThreadsJfrStreaming.test(TestVirtualThreadsJfrStreaming.java:68)
        at java.base@21/java.lang.reflect.Method.invoke(Method.java:580)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
        at com.oracle.mxtool.junit.MxJUnitWrapper.runRequest(MxJUnitWrapper.java:376)
        at org.graalvm.nativeimage.junitsupport/com.oracle.svm.junit.SVMJUnitRunner.run(SVMJUnitRunner.java:164)
        at org.graalvm.nativeimage.junitsupport/com.oracle.svm.junit.SVMJUnitRunner.main(SVMJUnitRunner.java:184)
```

roberttoyonaga: >  please try to figure out where the problematic Package object is created

It gets created the first time we try to get the package. In this case, that's in `JfrTypeRepository` while we're trying to write out the type data. 
`Class.getPackage()` -->  `BootLoader.definePackage(Class<?> c)` --> `Target_jdk_internal_loader_BootLoader.getDefinedPackage` --> `Target_java_lang_Package()`
       
> I suppose that Package object also doesn't live in the image heap?

That's correct. It seems like when `DynamicHubInitializer.registerPackage` is called at build time, `javaClass.getPackage()` returns null for all packages with name `com.sun.proxy.jdk.proxy1`, preventing the package from being registered. Maybe this is why the package is not in the image heap.      

Is something wrong with the behavior for handling proxy classes? Otherwise, should we modify the `JfrSymbolRepository` (I can do that in a separate PR)?

christianhaeubl: This sounds like a difference in behavior between Native Image and HotSpot. If HotSpot directly returns `null` for the package of proxy classes, then Native Image should do the same (i.e., it should not allocate a package at run-time in `Target_jdk_internal_loader_BootLoader.getDefinedPackage`).

Please try to implement a smaller reproducer that shows this difference in behavior and open a separate issue for this problem.

roberttoyonaga: Hi @christianhaeubl, I've opened an issue here: https://github.com/oracle/graal/issues/8796
It includes a small reproducer too.

PR #7518, opened 2023-09-29
Omit internal jdk.JavaMonitorWait JFR events

roberttoyonaga:
### Summary
This is a very minor change to replicate the behavior in OpenJDK by omitting internal `jdk.JavaMonitorWait` events emitted by the JFR periodic task.  This basically just reduces some noise, especially when the application is idle.

Bugzilla: https://bugs.openjdk.org/browse/JDK-8286707?jql=text%20~%20%22jfr%22
OpenJDK PR: https://github.com/openjdk/jdk/pull/8883

See before (sorted by thread name):
image

And after:
image

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/JavaMonitorWaitEvent.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/JavaMonitorWaitEvent.java
@@ -1,6 +1,6 @@
 /*
- * Copyright (c) 2022, 2022, Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2022, 2022, Red Hat Inc. All rights reserved.
+ * Copyright (c) 2022, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2023, Red Hat Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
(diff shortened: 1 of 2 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/JavaMonitorWaitEvent.java
```diff
@@ -39,7 +39,7 @@
 
 public class JavaMonitorWaitEvent {
     public static void emit(long startTicks, Object obj, long notifier, long timeout, boolean timedOut) {
-        if (HasJfrSupport.get()) {
+        if (HasJfrSupport.get() && !obj.getClass().getName().equals("jdk.jfr.internal.JVM$ChunkRotationMonitor")) {
```
fniephaus, line 42: I'd prefer this to avoid NPEs:
```suggestion
        if (HasJfrSupport.get() && !"jdk.jfr.internal.JVM$ChunkRotationMonitor".equals(obj.getClass().getName())) {
```

christianhaeubl (commented):
Thanks for the PR.

Instead of doing a string comparison, please directly compare the class objects (the String comparison will fail silently if the JDK code changes). You will need to create a `@TargetClass` for `jdk.jfr.internal.JVM$ChunkRotationMonitor`.

christianhaeubl:
Thanks, this PR will be merged to master in the next few days.

fniephaus (approved):
LGTM, thanks Robert!

christianhaeubl (approved):

PR #7553, opened 2023-10-05
[GR-49300] Break the initialization cycle in NIO/cgroups code.

graalvmbot:


jerboaa (approved):
Thanks @christianhaeubl! This is sufficient for our use-case. +1 from me.

christianhaeubl:
@jerboaa: I pushed another small fix, see https://github.com/oracle/graal/pull/7553/commits/d090636e662599c02e62c35208fafb6def1fd4f6.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_nio_Bits.java
```diff
@@ -53,3 +55,23 @@ final class Target_java_nio_Bits {
...
+    static long getMAX_MEMORY() {
+        return DirectMemoryAccessors.getDirectMemory();
+    }
+
+    static void setMAX_MEMORY(long value) {
+        /* Nothing to do. */
+    }
+
+    // Checkstyle: resume
+}
```
jerboaa, line 77: Instead of going through the accessor an alternative could be to leaverage the init hooks in JDK code and actually set the init level to `1` once we know the real `MaxDirectMemorySize` (base on the heap size). For example here: https://github.com/oracle/graal/pull/7553/files#diff-db72af86eee732c8a7e3e53319b32fdd92ee0ec00735f413a02858c9dc0aff4dR126. That would mimic what HotSpot itself is doing. FWIW, that's what my original code was intending to do.

Both approaches work.

christianhaeubl: The problem is that Native Image is already multi-threaded when this code is executed. Modifying the init level may have unexpected effects on other threads.

jerboaa: OK, thanks.

jerboaa:
@christianhaeubl Thanks for getting this fix in! Much appreciated.

PR #7554, opened 2023-10-05
[GR-46740] Fix native downcalls on JDK 22

graalvmbot:


zakkak:
@mur47x111 this breaks JDK 21 support, right? i.e. if I build GraalVM based on this branch using JDK 21 foreign calls will be broken.

zakkak:
Hi @fniephaus can you help answering the above?

mur47x111:
this is a question for @peter-hofer

peter-hofer:
@zakkak this change affects only downcalls of the "Project Panama" Foreign Function & Memory API in GraalVM Native Image. This API is a preview feature at this time and we currently plan on supporting its features in Native Image only on the latest JDK version that is supported by GraalVM. Our support in JDK 21 is also only in its early stages.
@christianwimmer @wirthi

zakkak:
Thank you for the clarification @peter-hofer.

Issue #7546, opened 2023-10-04
Support JFR -XX:FlightRecorderOptions


roberttoyonaga:
**Feature request**

Add support for `-XX:FlightRecorderOptions`. These options allow the user to configure things like JFR repository location and stack depth at app startup at the same time when `-XX:StartFlightRecording` is used.  [Erik Gahlin's response here](https://stackoverflow.com/questions/64138134/java-flight-recorder-comprehensive-list-of-vm-options-for-openjdk-8-openjdk-11) is a good breakdown of what can be done with `-XX:FlightRecorderOptions`.

**Is your feature request related to a problem? Please describe.**
This is a new feature request, not a problem.

**Describe the solution you'd like.**
We need to parse the options passed by the user and configure internal JFR appropriately.  The behavior should match OpenJDK.

**Describe who do you think will benefit the most.**
GraalVM native image users will benefit. 
 
**Describe alternatives you've considered.**
No real alternatives. This is just something that is missing. 

**Express whether you'd like to help contributing this feature**
I can help contribute this.

PR #7585, opened 2023-10-10
[GR-51629] Add support for JFR -XX:FlightRecorderOptions

roberttoyonaga:
### Summary
`-XX:FlightRecorderOptions` can be passed on the command line when launching your app. This PR adds support for the following JFR options in native image: 

- `globalbuffercount`
-  `globalbuffersize`
- `maxchunksize`
- `memorysize`
- `repositorypath`
- `stackdepth`
- `thread_buffer_size`
- `preserveRepository`. 

A few of these options are legacy options are are not very important (although the current OpenJDK still supports them). `maxchunksize`, `repositorypath`,  and`stackdepth` are probably the most useful options and advanced JFR users may find them important.  `preserveRepository` is a new option added to JDK22 via https://github.com/openjdk/jdk/pull/13111. 

See issue: https://github.com/oracle/graal/issues/7546 

### Details
In HotSpot, the VM-level JFR options may not be changed after JFR is initialized. This means we expect that `com.oracle.svm.core.jfr.JfrOptionSet` will not be needed after JFR is initialized. Thus the data flow should look something like the diagram below:
**HotSpot:**

**SubstrateVM:**

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/doc-files/FlightRecorderOptionsHelp.txt
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/doc-files/FlightRecorderOptionsHelp.txt
@@ -4,7 +4,7 @@ This option expects a comma separated list of key-value pairs. None of the optio
 
 globalbuffercount   (Optional) Number of global buffers. This option is a legacy
                     option: change the memorysize parameter to alter the number of
-                    global buffers. This value cannot be changed once JFR has been"
+                    global buffers. This value cannot be changed once JFR has been
                     initialized. (LONG, default determined by the value for
                     memorysize)
 
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/doc-files/FlightRecorderOptionsHelp.txt
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/doc-files/FlightRecorderOptionsHelp.txt
@@ -1,0 +1,46 @@
+Usage: -XX:FlightRecorderOptions=[option[=value][,...]]
+
+This option expects a comma separated list of key-value pairs. None of the options are mandatory. Possible option keys are as follows:
+
+globalbuffercount   (Optional) Number of global buffers. This option is a legacy
+                    option: change the memorysize parameter to alter the number of
+                    global buffers. This value cannot be changed once JFR has been"
+                    initialized. (LONG, default determined by the value for
+                    memorysize)
+
+globalbuffersize    (Optional) Size of the global buffers, in bytes. This option is a
+                    legacy option: change the memorysize parameter to alter the size
+                    of the global buffers. This value cannot be changed once JFR has
+                    been initialized. (STRING, default determined by the value for
+                    memorysize)
+
+maxchunksize        (Optional) Maximum size of an individual data chunk in bytes if
+                    one of the following suffixes is not used: 'm' or 'M' for
+                    megabytes OR 'g' or 'G' for gigabytes. This value cannot be
+                    changed once JFR has been initialized. (STRING, 12M)
+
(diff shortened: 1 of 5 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/doc-files/FlightRecorderOptionsHelp.txt
```diff
@@ -0,0 +1,46 @@
+Usage: -XX:FlightRecorderOptions=[option[=value][,...]]
+
+This option expects a comma separated list of key-value pairs. None of the options are mandatory. Possible option keys are as follows:
+
+globalbuffercount   (Optional) Number of global buffers. This option is a legacy
+                    option: change the memorysize parameter to alter the number of
+                    global buffers. This value cannot be changed once JFR has been"
```
galderz, line 7: Remove training `"`

galderz (commented):
Small typo. Have you tested this options somehow?

Even though present in OpenJDK, I have doubts about bringing in legacy options.

roberttoyonaga:
Thanks @galderz !
> Small typo.

Good catch, I'll fix that. 
> Have you tested this options somehow?

Yes I've tested them manually by launching JFR, passing FlightRecorderOptions and checking to make sure the specified options are respected. I'm not sure if it's possible to add a JUnit test for this because it requires passing the options when launching the executable (options must not be changed after initialization). This is a problem because JFR is started/initialized once for all the Native Image unit tests. 
> Even though present in OpenJDK, I have doubts about bringing in legacy options.

Yes, that's a fair point. I only decided to add them because we already implicitly handle those 'legacy" options, we just don't yet parse them from the command line.  I'm fine with removing those legacy options from this PR. Let's see what @christianhaeubl  thinks too.

ramiswailem:
https://github.com/ramiswailem/api/issues/130

fniephaus:
> I'm not sure if it's possible to add a JUnit test for this because it requires passing the options when launching the executable (options must not be changed after initialization). This is a problem because JFR is started/initialized once for all the Native Image unit tests.

We probably need a different kind of test. Take a look at, for example, how the [`debuginfotest` is implemented](https://github.com/oracle/graal/blob/7ea11579755b0a2c6db2ec46d516c56047163100/substratevm/mx.substratevm/mx_substratevm.py#L893). You could generate a simple native image, then run it with the right flags in the background and do what you need to do.

> I'm fine with removing those legacy options from this PR.

I'm inclined to say we don't support legacy options, unless there's actual user demand.

christianhaeubl:
> I'm inclined to say we don't support legacy options, unless there's actual user demand.

Most of those legacy options are more like experimental options (i.e., they can be used to provide workarounds or fine tune JFR), so I would suggest that we support them as well and change the documentation accordingly.

I will probably start integrating this PR later today, as it has some overlap with https://github.com/oracle/graal/pull/6743, which I am currently integrating. @roberttoyonaga: if you plan to add any test cases, please do that in a separate PR.

christianhaeubl:
@roberttoyonaga: I opened https://github.com/oracle/graal/pull/8254. I primarily simplified the [FlightRecorderOptionsHelp.txt](https://github.com/oracle/graal/pull/8254/files#diff-3bf6f54c98be192bb77329855bcb4b721f1c30c24575658fe599457e3f86c68f) a bit. Please let me know if you think that I removed anything relevant.

roberttoyonaga:
> I opened #8254. I simplified [FlightRecorderOptionsHelp.txt](https://github.com/oracle/graal/pull/8254/files#diff-3bf6f54c98be192bb77329855bcb4b721f1c30c24575658fe599457e3f86c68f) a bit. Please let me know if you think that I removed anything relevant.

Thanks @christianhaeubl. Your changes look good to me.

Issue #7631, opened 2023-10-19
Native Memory Tracking and Associated JFR events


roberttoyonaga:
**Feature request**
 
**Is your feature request related to a problem? Please describe.**
Not related to a problem.

**Describe the solution you'd like.**
Hotspot JFR has[ native memory tracking (NMT)](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/tooldescr007.html) events. It may be useful to have support for NMT in SubstrateVM.

In Hotspot, NMT hooks into malloc and also the virtual memory management code (where mmap is used). This allows for tracking how much memory each VM component needs for its own operations. Native memory is grouped into categories (ie GC, Metaspace, etc). The JFR NMT events are just a small feature built on top of the already existing NMT infrastructure. 

In SubstrateVM, the memory used by the java application and VM itself is not so clearly separated because memory used by the VM can be managed by the GC. So I can think of two approaches:
1. Record only off-heap memory. Perhaps adding tracking in places such as `org.graalvm.nativeimage.UnmanagedMemory#malloc`
	- This wouldn't provide a complete picture of how much memory each VM component is actually using. It would at least fill in the gaps left by heap dumps and JFR allocation events. 
2. Determine a way to distinguish VM-level allocations from application level ones. This would provide a more complete breakdown of what memory is needed for VM internal operations. I think this is probably more complicated to accomplish. 

In SubstrateVM there are three areas where the VM itself might contribute to memory usage: the java heap, off-heap, and the image heap.

To implement NMT in Native Image we first need to decide what "native memory" we want to track. Does "native memory" mean all memory used by the VM for its internal operations, or does it simply mean unmanaged memory?  In the Hotspost NMT implementation, these are basically the same thing -- memory used by the VM is not managed by the GC.  

In Hotspot, there are two NMT modes "summary" and "detailed". The "detailed" mode includes call site information,  whereas the "summary" only provides a memory amounts per category. As a first step, I think the goal should be scoped down to only the "summary" mode.  

**Describe who do you think will benefit the most.**
GraalVM users and contributors would benefit.  Some use cases would be:

1. RSS is higher than expected or exceeds specified heap size. NMT can shed light on memory usage contributing to RSS not visible in heap dumps or JFR allocation events.
2. A user or GraalVM developer may want to measure how much each VM component contribues to RSS. For example, is JFR too expensive to use in their native image executables for their use-case?

**Express whether you'd like to help contributing this feature**
[...]

PR #7669, opened 2023-10-25
Clean up off-heap memory usages

roberttoyonaga:
### Summary
This PR does a bit of clean up to set the stage for further work on native memory tracking (see issue: https://github.com/oracle/graal/issues/7631).  Specifically, it replaces usages of `LibC` malloc/calloc/realloc/free with calls to`UnmanagedMemory`.  From now on, I suggest that `LibC` not be used directly to manage memory.  This will allow for NMT code to be added to `UnmanagedMemorySupport` implementations. Additionally, this PR specifically marks locations where memory allocated outside Java code must be freed (`UnmanagedMemory#untrackedFree`). 

This refactoring serves a second important purpose. It helps mitigate the following pitfalls associated with "malloc headers" needed for NMT. [In hotspot](https://github.com/openjdk/jdk/blob/master/src/hotspot/share/services/mallocHeader.hpp#L37-L88), the "malloc header" is an extra bit of memory allocated contiguously with the main memory payload. It contains metadata such as size and type used for NMT.  I think a similar approach can be used in SubstrateVM using `@RawStructures` (similar to how the JFR buffers have a header and payload). A pointer to the data payload is returned from malloc and the header is transparent to the caller.  Here are the pitfalls I can see:

1. If `LibC#malloc()` is used to allocate memory instead of `UnmanagedMemorySupport`, we'll miss some data that should be tracked. 
2. If `LibC#free()` is used to free tracked memory, we'll be in trouble because only `UnmanagedMemorySupport` is aware of the extra header data.
3. If `UnmanagedMemory#free()` is called on untracked memory (off-heap memory allocated outside of Java), then we'll segfault when trying to access a header that doesn't exist. 

The clean up in this PR mitigates these pitfalls, but doesn't eliminate them. If the user is determined, they can use `ImageSingletons.lookup(LibCSupport.class)` to use LibC directly, and they can choose to use `UnmanagedMemory#free()` instead of `UnmanagedMemory#untrackedFree()` if they ignore the Javadoc.

### DIFF
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/UnmanagedMemory.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/UnmanagedMemory.java
@@ -146,4 +146,14 @@ public final class UnmanagedMemory {
     public static void free(PointerBase ptr) {
         ImageSingletons.lookup(UnmanagedMemorySupport.class).free(ptr);
     }
+
+    /**
+     * Temporarily the same as {@link UnmanagedMemory#free(PointerBase)}. Will later be different
+     * because it will not attempt to perform any NMT operations. This is crucial for releasing
+     * memory allocated by C libraries which will not have NMT "malloc headers". If
+     * {@link UnmanagedMemory#free(PointerBase)} is used instead, a segfault will occur.
+     */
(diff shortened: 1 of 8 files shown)


roberttoyonaga:
Closed by https://github.com/oracle/graal/pull/7883

Issue #7677, opened 2023-10-26
Unexpected ordering of entries in classpath starting with 23.1


zakkak:
**Describe the issue**
https://github.com/oracle/graal/commit/d5000f4146e66faaa4d79cd80c5b3274fcf12e14 (specifically [this change](https://github.com/oracle/graal/commit/d5000f4146e66faaa4d79cd80c5b3274fcf12e14#diff-9e053bec9eca0799896bd471363fa6aa1df70269e454b89dff1ea5fb844a480cL2153-R2191)) changes the ordering of the classpath resulting in unexpected behavior when loading resources or services.

**Steps to reproduce the issue**
See https://github.com/quarkiverse/quarkus-cxf/issues/1078#issuecomment-1779148123

**Describe GraalVM and your environment:**
 - GraalVM version (latest snapshot builds can be found [here](https://github.com/graalvm/graalvm-ce-dev-builds/releases)), or commit id if built from source: **d5000f4146e66faaa4d79cd80c5b3274fcf12e14** and **fe2bdb158c70e32f7fc49af4612616d4ffc772eb**
 - JDK major version: **21**
 - OS: **Fedora 38**
 - Architecture: **AMD64**

PR #7678, opened 2023-10-26
[GR-50675] Fix class-path order when Manifests include "Class-Path" attributes

zakkak:
`destination` is the classpath GraalVM creates from parsing the jar files passed to it. The ordering of the entries in the classpath is important. Jar files themselves should be added before the entries of their corresponding "Class-Path" attributes.

At the same time we want to process the contents of `META-INF/native-image` of the entries in the "Class-Path" before processing the contents of `META-INF/native-image` of the jar itself. This is required to ensure that the jar `META-INF/native-image` contents can override those of the entries in its "Class-Path".

As a result I chose to optimistically add the path to the classpath (aka `destination`) and remove it afterwards if deemed necessary.

Fixes https://github.com/oracle/graal/issues/7677

The fix should be backported to 23.1 as well

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -2110,12 +2110,13 @@ public class NativeImage {
 
         Path classpathEntryFinal = useBundle() ? bundleSupport.substituteClassPath(classpathEntry) : classpathEntry;
         if (!imageClasspath.contains(classpathEntryFinal) && !customImageClasspath.contains(classpathEntryFinal)) {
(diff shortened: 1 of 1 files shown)


zakkak:
Hi @olpaw, this is a kind reminder.

olpaw (approved):

zakkak:
Thank you for the review @olpaw. I updated the PR to include your suggestions.

olpaw (approved):

olpaw:
PR is in the merge queue.

zakkak:
Thank you @olpaw

zakkak:
@ezzarghili a note that this fix should be backported to 23.1 as well. Could you please confirm it's on the backport queue?

olpaw:
@zakkak I created a ticket for backporting this to 23.1.x.

Issue #5868, opened 2023-02-01
[GR-44056] Cannot invoke "java.util.Set.iterator()" because "triggers" is null.


ddobrin:
Reported today in Spring Boot first: [Spring Boot 3 Native Needs RomeTools and Kotlin Serialization JARs](https://github.com/spring-projects/spring-boot/issues/32915).

**Issue can be reproduced on Mac Intel and Mac M1 and Linux**.

The repro is [here](https://github.com/ddobrin/serverless-photosharing-workshop/tree/master/services/image-analysis/java)
Env: Spring Boot 3.0.2, GraalVM 22.3.

**Java version:**
  openjdk version "17.0.5" 2022-10-18 
  OpenJDK Runtime Environment GraalVM CE 22.3.0 (build 17.0.5+8-jvmci-22.3-b08)
  OpenJDK 64-Bit Server VM GraalVM CE 22.3.0 (build 17.0.5+8-jvmci-22.3-b08, mixed mode, sharing)

The service uses Google Native Java Client Libraries and accesses Vision and Firestore; uses the latest Google BOM.

Native image builds and works fine when adding rometools & kotlinx to the pom.xml:
...
com.rometools
rome
...
org.jetbrains.kotlinx
kotlinx-serialization-core

Without the added dependencies, it can be built with the Java agent, collect the generated *.json files, copy them to src/main/resources/META-INF and build the native image, which will fail with:
[6/7] Compiling methods... [********] (65.0s @ 6.35GB)

Fatal error: java.lang.NullPointerException: Cannot invoke "jdk.vm.ci.meta.JavaField.format(String)" because "field" is null
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.meta.HostedUniverse.lookup(HostedUniverse.java:369)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.meta.HostedUniverse.lookup(HostedUniverse.java:282)

These dependency requirements did not exist in Boot 3 M5, however libraries have evolved since.

fniephaus:
Hi @ddobrin, I'm having troubles following your instructions to reproduce the issue. When I check out the code, I seem to be able to build the application without a problem. What dependencies do I need to remove? I'm assuming with Java agent you mean the Trace Agent from Native Image? How do I run the application with the agent?

Also, could you please provide a longer stack trace? It seems that you are hitting this line with `field` being `null`:
[...]

PR #7736, opened 2023-11-06
Synchronize accesses to reachability handlers

zakkak:
Prevents data races in reachability handlers registration when using
`-H:-RunReachabilityHandlersConcurrently`.

The fact that https://github.com/oracle/graal/issues/5868 is not deterministic and it's related to enabling/disabling concurrency hints that it's probably caused by a data race. 
Adding a synchonized blocks in `com.oracle.svm.hosted.ReachabilityHandlerFeature#registerReachabilityHandler(org.graalvm.nativeimage.hosted.Feature.BeforeAnalysisAccess, java.lang.Object, java.lang.Object[])` seems to resolve the issue.

Closes https://github.com/oracle/graal/issues/5868

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ReachabilityHandlerFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ReachabilityHandlerFeature.java
@@ -32,6 +32,7 @@ import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
@@ -44,6 +45,7 @@ import com.oracle.graal.pointsto.meta.AnalysisType;
 import com.oracle.svm.core.SubstrateOptions;
 import com.oracle.svm.core.feature.AutomaticallyRegisteredFeature;
 import com.oracle.svm.core.feature.InternalFeature;
+import com.oracle.svm.core.util.ConcurrentIdentityHashMap;
 import com.oracle.svm.core.util.UserError;
 import com.oracle.svm.core.util.VMError;
 import com.oracle.svm.hosted.FeatureImpl.BeforeAnalysisAccessImpl;
@@ -52,8 +54,8 @@ import com.oracle.svm.hosted.FeatureImpl.DuringAnalysisAccessImpl;
 @AutomaticallyRegisteredFeature
 public class ReachabilityHandlerFeature extends ReachabilityHandler implements InternalFeature {
 
-    private final IdentityHashMap<Object, Set<Object>> activeHandlers = new IdentityHashMap<>();
-    private final IdentityHashMap<Object, Map<Object, Set<Object>>> triggeredHandlers = new IdentityHashMap<>();
+    private final Map<Object, Set<Object>> activeHandlers = new ConcurrentIdentityHashMap<>();
+    private final Map<Object, Map<Object, Set<Object>>> triggeredHandlers = new ConcurrentIdentityHashMap<>();
 
     public static ReachabilityHandlerFeature singleton() {
         return ImageSingletons.lookup(ReachabilityHandlerFeature.class);
@@ -88,7 +90,7 @@ public class ReachabilityHandlerFeature extends ReachabilityHandler implements I
         BeforeAnalysisAccessImpl access = (BeforeAnalysisAccessImpl) a;
         AnalysisMetaAccess metaAccess = access.getMetaAccess();
 
-        Set<Object> triggerSet = activeHandlers.computeIfAbsent(callback, c -> new HashSet<>());
+        var triggerSet = activeHandlers.computeIfAbsent(callback, c -> ConcurrentHashMap.newKeySet());
 
         for (Object trigger : triggers) {
             if (trigger instanceof Class) {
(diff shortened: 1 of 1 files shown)


christianwimmer (changes requested):
Instead of the explicit `synchronized` blocks, it would be better to use proper synchronized data structures. For `ReachabilityHandlerFeature.activeHandlers` and `triggeredHandlers` that would be wrapping using `Collections.synchronizedMap`, while for the nested set that does not require identity semantics we can just use `ConcurrentHashMap.createKeySet`.

zakkak:
Hi @christianwimmer I see there is `com.oracle.svm.core.util.ConcurrentIdentityHashMap` as well, shall I use it instead of `Collections.synchronizedMap` for better performance?

christianwimmer:
> Hi @christianwimmer I see there is `com.oracle.svm.core.util.ConcurrentIdentityHashMap` as well, shall I use it instead of `Collections.synchronizedMap` for better performance?

Yes, I forgot that we have that.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ReachabilityHandlerFeature.java
```diff
@@ -119,7 +122,7 @@ public void duringAnalysis(DuringAnalysisAccess a) {
                 Set<Object> triggers = activeHandlers.get(callback);
                 if (callback instanceof Consumer) {
                     if (isTriggered(access, triggers)) {
-                        triggeredHandlers.put(callback, null);
+                        triggeredHandlers.put(callback, NULL_MAP);
```
christianwimmer, line 125: Why not just use `Map.of()`? We want that map to be immutable anyways.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ReachabilityHandlerFeature.java
```diff
@@ -52,8 +54,9 @@
 @AutomaticallyRegisteredFeature
 public class ReachabilityHandlerFeature extends ReachabilityHandler implements InternalFeature {
 
-    private final IdentityHashMap<Object, Set<Object>> activeHandlers = new IdentityHashMap<>();
-    private final IdentityHashMap<Object, Map<Object, Set<Object>>> triggeredHandlers = new IdentityHashMap<>();
+    private static final Map<Object, Set<Object>> NULL_MAP = new IdentityHashMap<>();
+    private final ConcurrentIdentityHashMap<Object, ConcurrentHashMap.KeySetView<Object, Boolean>> activeHandlers = new ConcurrentIdentityHashMap<>();
```
christianwimmer, line 58: Why change the generic type from `Set` to `ConcurrentHashMap.KeySetView`? The field can just be defined as `Map<Object, Set<Object>> activeHandlers`.

Same for all the other type definitions.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ReachabilityHandlerFeature.java
```diff
@@ -88,7 +91,7 @@ private void registerReachabilityHandler(BeforeAnalysisAccess a, Object callback
         BeforeAnalysisAccessImpl access = (BeforeAnalysisAccessImpl) a;
         AnalysisMetaAccess metaAccess = access.getMetaAccess();
 
-        Set<Object> triggerSet = activeHandlers.computeIfAbsent(callback, c -> new HashSet<>());
+        ConcurrentHashMap.KeySetView<Object, Boolean> triggerSet = activeHandlers.computeIfAbsent(callback, c -> ConcurrentHashMap.newKeySet());
```
christianwimmer, line 94: that's actually the prototypical use case for `var triggerSet`, which we can now use in the native-image code base.

christianwimmer (changes requested):

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ReachabilityHandlerFeature.java
zakkak: I didn't think about it. It makes much more sense. Done, thanks.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ReachabilityHandlerFeature.java
zakkak: I guess I just like explicit types :)

Changed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/ReachabilityHandlerFeature.java
zakkak: Done.

christianwimmer:
@oubidar-Abderrahim can you please create the internal PR for this

christianwimmer (approved):

zakkak:
Thank you for the review @christianwimmer

zakkak:
This should also be backported to 23.1 and 23.0

zakkak:
@oubidar-Abderrahim is there any progress in regards to getting this merged? Thanks

christianwimmer:
> @oubidar-Abderrahim is there any progress in regards to getting this merged? Thanks

I'm in the process of running the final tests and benchmarks, should be merged soon. And it will be backported.

zakkak:
Great, thanks for the update @christianwimmer.

btomala:
We encountered the same issue. Thanks for fixing it. @christianwimmer when do we expect this will be released?

zakkak:
@btomala it should land in [GraalVM for JDK 21.0.2 in late January](https://www.graalvm.org/release-calendar/).

Issue #7739, opened 2023-11-06
Update JFR mirror event registration to account for new JDK class 


roberttoyonaga:
This change (https://github.com/openjdk/jdk/pull/16411) in OpenJDK will cause [`com.oracle.svm.hosted.jfr.JfrEventSubstitution#createMirrorEventsMapping`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jfr/JfrEventSubstitution.java#L229) to fail. Specifically, the introduction of [MirrorEvents.java ](https://github.com/openjdk/jdk/pull/16411/files#diff-b7ed5785d2fecc3aa9c0ca4d7de5b6fa5c200ae06729f653433033cff8912a5b) will require us to modify how we register JFR mirror events in SubstrateVM. 

I can help do this maintenance once labsJDK is up to date with the OpenJDK changes that will cause this problem.

zapster:
Thanks @roberttoyonaga, we are already working on integrating JDK 22+22. I've pushed the current state to #7748, but note that it is work in progress and the labsjdk builds are not yet public.

Regarding the mirror events change, for now we have only removed references to the legacy code, but you are right, we need to adopt `MirrorEvents.java` as well. We are still tackling other breaking JDK changes, but I hope that it wont take to long.

In any case, comments/suggestion/contributions to the WIP very much appreciated.

roberttoyonaga:
Hi @zapster! Thank you. I've added [a comment on the WIP internal PR mirror](https://github.com/oracle/graal/pull/7748/files#r1385000066) that explains in more detail what I mean.

PR #7748, opened 2023-11-07
[GR-49852] [GR-49613] Update JVMCI to 22+22-jvmci-b01.

graalvmbot:
Update to 22+22-jvmci-b01

* Adopt "[JDK-8239508](https://github.com/openjdk/jdk/commit/2d5829afbcc022297940cbf436b14725a9251b3) JFR: @RemoveFields". 
  * Already reported on Github: https://github.com/oracle/graal/issues/7739
* Adopt "JDK-8288899 java/util/concurrent/ExecutorService/CloseTest.java failed with "InterruptedException: sleep interrupted""
  *  PR with an innocent title that [changes the signature](https://github.com/openjdk/jdk/pull/14301#pullrequestreview-1673674131) of the public non-final `ForkJoinPool#invokeAll` method and thus is making JDK 22 source incompatible with previous JDKs. This is not only a problem for native-image but for all Java code that subclass `ForkJoinPool`.
  *  CSR (Compatibility & Specification Review) for this change:  [JDK-8309407](https://bugs.openjdk.org/browse/JDK-8309407)


substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jfr/JfrEventSubstitution.java
```diff
@@ -66,14 +67,20 @@ public class JfrEventSubstitution extends SubstitutionProcessor {
...
         ResolvedJavaType jdkJfrEventWriter = metaAccess.lookupJavaType(EventWriter.class);
         changeWriterResetMethod(jdkJfrEventWriter);
         typeSubstitution = new ConcurrentHashMap<>();
         methodSubstitutions = new ConcurrentHashMap<>();
         fieldSubstitutions = new ConcurrentHashMap<>();
-        mirrorEventMapping = createMirrorEventsMapping();
+        if (JavaVersionUtil.JAVA_SPEC < 22) {
+            mirrorEventMapping = createMirrorEventsMapping();
+        } else {
+            mirrorEventMapping = null;
```
roberttoyonaga, line 82: For JDK 22, I think you will still need to call `createMirrorEventsMapping()`. But `createMirrorEventsMapping()` will need `ReflectionUtil.lookupClass(false, "jdk.jfr.internal.MirrorEvents");` instead of `ReflectionUtil.lookupClass(false, "jdk.jfr.internal.instrument.JDKEvents");` depending on what `JavaVersionUtil.JAVA_SPEC` returns. 

Additionally, I think an `@Alias` will be required for `jdk.jfr.internal.instrument.JDKEvents#mirrorEventClasses` similar to what already exists in `Target_jdk_jfr_internal_instrument_JDKEvents` for JDK 21.

zapster: Thanks so much for the suggestions @roberttoyonaga! Interestingly, the mirrored events are not covered by our testing.

Since the JDK update is rather critical, we plan to merge this PR as is and deal with the remaining JFR changes in a follow-up PR. We should also add some testing so that we cover these cases in our gate jobs.

If you could prepare a PR, we would very much appreciate it as it would speed up things. I can let you know as soon as the labjdk builds are publicly available, so you can start on top of this PR without waiting for it to merge.

Otherwise, if you don't have time, I can prepare a PR as well.

zapster: The labsJDK 22+22 artifacts are available: https://github.com/graalvm/labs-openjdk/releases

roberttoyonaga: > If you could prepare a PR, we would very much appreciate it as it would speed up things.

Hi @zapster! Yes, I'll prepare a PR for these JFR updates

roberttoyonaga: Hi @zapster, on second thought, we may not actually need to do anything further to accommodate the changes related to `MirrorEvents.java`.  The new behaviour is that `jdk.jfr.internal.MirrorEvents` maintains a cache of all the mirror event classes instead of that cache needing to be built in `JDKEvents`. So this means the extra step of using `SecuritySupport#registerMirror` is unnecessary. We also now don't have to worry about doing this extra step at native image build-time before manually retransforming the JFR event classes in `JfrEventSubstitution` anymore.   

I made a PR that adds a test to verify mirror events are handled correctly. https://github.com/oracle/graal/pull/7771

zapster: Awesome, thanks a lot!

PR #7771, opened 2023-11-08
Add test to verify JFR mirror events

roberttoyonaga:
Adds a test to ensure mirror events are handled and set up correctly.  This should help catch changes in OpenJDK that might affect such events in SubstrateVM.


zapster:
@christianhaeubl could you please have a look and integrate it? See also https://github.com/oracle/graal/pull/7748#discussion_r1387028078

christianhaeubl (approved):
Thanks, I did some simplifications on top, see the linked PR above. This should get merged in the next few days.

PR #7807, opened 2023-11-13
Support seperated Antlr grammar.

oraluben:
Current `mx_truffle.create_parser` generates `grammar{Lexer,Parser}.java` from a single Antlr file. Antlr also supports defining lexer and parser separately, and sometimes it can be easier to maintain such two grammar files.

This pr allows `mx_truffle.create_parser` to handle both cases. I understand it's some truffle internal thing, but this method is easy to use and be integrated into `mx`-based workflow. I used to use it in my language impl, and with a mocked `mx.run_java` to support this. After mx@v7.0.0 it is no longer allowed, so I submit it and hope this change would benefit other truffle-based language.


oubidar-Abderrahim:
Thank you for contributing to GraalVM,
@chumer can you please review this PR? Thank you

oraluben:
can be workarounded by `mx.__dict__['_allowed_writes'].append('run_java')`

Issue #7786, opened 2023-11-09
[GR-51623] graalvm turkish system language compile bug


burhan-taskesen:
**Describe the issue**
When compiling a jar file with the native-image command, if your system language is Turkish, there will be problems in the reflection part. When converting the letter i to uppercase, it is translated as "I". Since this letter is available in Turkish but not in English, NoSuchMethodException is thrown. When you change the system language to English, the problem disappears. (macos arm64)

In my application, I call the method of a library shared with JNA and print the result to the screen. Here are the images:

**Describe GraalVM and your environment:**
 - GraalVM version (21.0.1 && 17.0.9)
 - JDK major version: 21.0.1 && 17.0.9
 - OS: macOS Sonoma
 - Architecture: arm64

![image](https://github.com/oracle/graal/assets/98235097/6db6323c-2480-4fe0-ac4c-2ed7c6200d71)
![image](https://github.com/oracle/graal/assets/98235097/314084e1-c407-4258-9d32-dea334d5c8c9)
![image](https://github.com/oracle/graal/assets/98235097/abf08cdb-23cb-48a0-a90a-5e5d12fc4bef)

selhagani:
Hi @burhan-taskesen , Can you share a concise reproducer please? Thank you.

petermz:
I think I know what's going on here: `String.toUpperCase` is called [here](https://github.com/oracle/graal/blob/ebc75fa3f205ee2d37b9fcc6a466687dcf307b58/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/methodhandles/MethodHandleFeature.java#L315) with no locale parameter, which uses system locale by default. This works most of the time, but in Turkish locale "i" is converted to LATIN CAPITAL LETTER I WITH DOT ABOVE.

The fix might be as simple as call `toUpperCase(Locale.ENGLISH)` instead.

burhan-taskesen:
Actually, I did not use any method of the "String" class in the code. So there is no place where I can try this suggestion. All my code does is call the interface method :/

<img width="806" alt="image" src="https://github.com/oracle/graal/assets/98235097/cb53016f-5dfc-493b-9d14-7cae93b7e868">
<img width="601" alt="image" src="https://github.com/oracle/graal/assets/98235097/7c8adf54-b1b8-4ec4-9ed5-cd92e1a809f6">

selhagani:
I see, in that case as requested previously, can you share a concise reproducer please?

[...]

PR #7924, opened 2023-11-29
[GR-51623] Call `String::to{Upper,Lower}Case` with `Locale.ENGLISH`

petermz:
`String::toUpper/LowerCase` methods use system locale by default. Most of the locales work fine, but some don't, one example is Turkish where `I` and `i` are different letters. This is causing build failures such as https://github.com/oracle/graal/issues/7786 .

I believe it is safer to always pass `Locale.ENGLISH` into these methods. To limit scope of this fix, I only fixed places where errors might occur. I didn't touch user-facing messages, assertions, and tests.


oubidar-Abderrahim:
Thank you for your contribution to Graal, we'll take a look into this shortly.

fniephaus (approved):
This LGTM, thanks for the PR!

fniephaus:
Note that I merged this via #8247 and that I also added a test to ensure this doesn't happen again:
https://github.com/oracle/graal/pull/8247/files#diff-ba10d62de621598019c54022b8ce83841a16b60d1e9dd7792e3a396ce0bbc32e

petermz:
Looks great!
Does this kind of test somehow apply automatically to all SubstrateVM
classes?

fniephaus:
> Does this kind of test somehow apply automatically to all SubstrateVM classes?

Yes, we internally run a test based on this also on the SVM code base.

PR #7941, opened 2023-12-01
Fix typos in org.graalvm.polyglot.Engine

zakkak:


chumer:
This is no longer relevant as since 23.1.2 and 24.0.0 we no longer produce this error.

PR #7945, opened 2023-12-04
[GR-44559] Finish ThreadMXBean implementation for Native Image.

### LINKED ISSUE #6101: [GR-44559] Finish `ThreadMXBean` implementation for Native Image.
[`SubstrateThreadMXBean`](https://github.com/oracle/graal/blob/7268f717047f68473e5bff76be666bc609fddcf0/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/management/SubstrateThreadMXBean.java) implements `com.sun.management.ThreadMXBean` for Native Image and is not fully implemented yet, for example:

https://github.com/oracle/graal/blob/7268f717047f68473e5bff76be666bc609fddcf0/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/management/SubstrateThreadMXBean.java#L149-L172

To improve JMX support (#4732), some of these methods throwing ` VMError.unsupportedFeature()` will be switched to return reasonable, but empty default values (e.g., `new ThreadInfo[0]`). While this allows some JMX tools to work, some key functionality of `ThreadMXBean` is still missing and needs to be implemented.

smthelusive:
The functionality of `ThreadMXBean` is implemented and working within native image [GR-44559](https://github.com/oracle/graal/issues/6101)

### Testing
Build native image with: `--enable-monitoring=jmxserver,jvmstat` 
Run it with:
`-Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.port=9996 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.enableThreadContentionMonitoring=true`

Tools like VisualVM show the Thread monitoring info:

Screenshot 2023-12-04 at 11 15 56

Example result of `dumpAllThreads` method call:

Screenshot 2023-12-04 at 11 19 58

Result of `getThreadInfo` by Thread id (the locks and monitors shouldn't be included):
Screenshot 2023-12-04 at 11 24 57

Deadlock detection returns an array of ids of deadlocked threads:
Screenshot 2023-12-04 at 11 28 51

Closes #6101

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
@@ -1,0 +1,158 @@
+/*
+ * Copyright (c) 2023, 2024, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.core.jdk;
+
+import com.oracle.svm.core.SubstrateUtil;
+import com.oracle.svm.core.jdk.management.SubstrateThreadMXBean;
+import com.oracle.svm.core.locks.Target_java_util_concurrent_locks_AbstractOwnableSynchronizer;
+import com.oracle.svm.core.monitor.JavaMonitor;
+import com.oracle.svm.core.thread.JavaThreads.JMXMonitoring;
+import com.oracle.svm.core.thread.PlatformThreads;
+
+import java.lang.management.ThreadInfo;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.concurrent.locks.AbstractOwnableSynchronizer;
+import java.util.concurrent.locks.LockSupport;
+
+/**
+ * Utils to support {@link SubstrateThreadMXBean} for providing threading information for JMX
+ * support. Include the {@link ThreadInfo} constructing utils, and deadlock detection utils.
+ */
+public class ThreadMXUtils {
+
+    public static class ThreadInfoConstructionUtils {
+
+        private static StackTraceElement[] getStackTrace(Thread thread, int maxDepth) {
+            StackTraceElement[] stackTrace = thread.getStackTrace();
+            return maxDepth == -1 || maxDepth >= stackTrace.length ? stackTrace : Arrays.copyOfRange(stackTrace, 0, maxDepth);
+        }
+
+        private record Blocker(Object blockObject, Thread ownerThread) {
+        }
+
+        private static Blocker getBlockerInfo(Thread thread) {
+            Object blocker = LockSupport.getBlocker(thread);
+
+            if (blocker instanceof JavaMonitor javaMonitor) {
+                return new Blocker(
+                                javaMonitor.getBlockedObject(),
+                                SubstrateThreadMXBean.getThreadById(javaMonitor.getOwnerThreadId()));
+            } else if (blocker instanceof AbstractOwnableSynchronizer synchronizer) {
+                return new Blocker(synchronizer,
+                                SubstrateUtil.cast(synchronizer, Target_java_util_concurrent_locks_AbstractOwnableSynchronizer.class)
+                                                .getExclusiveOwnerThread());
+            }
+            return new Blocker(blocker, null);
+        }
+
+        private static Object[] getLockedSynchronizers(Thread thread) {
+            return JMXMonitoring.getThreadLocks(thread);
+        }
+
+        private record LockedMonitors(Object[] monitorObjects, int[] monitorDepths) {
+        }
+
+        private static LockedMonitors getLockedMonitors(Thread thread, int stacktraceLength) {
+            List<JMXMonitoring.MonitorInfo> monitors = JMXMonitoring.getThreadMonitors(thread);
+            Object[] monitorObjects = monitors.stream().map(JMXMonitoring.MonitorInfo::originalObject).toArray();
+            int[] monitorDepths = monitors.stream().mapToInt(monitorInfo -> stacktraceLength < 0 ? -1 : stacktraceLength - monitorInfo.stacksize()).toArray();
+            return new LockedMonitors(monitorObjects, monitorDepths);
+        }
+
+        private static int getThreadState(Thread thread, boolean inNative) {
+            int state = PlatformThreads.getThreadStatus(thread);
+            if (inNative) {
+                // set the JMM thread state native flag to true:
+                state |= 0x00400000;
+            }
+            return state;
+        }
+
+        public static ThreadInfo getThreadInfo(Thread thread, int maxDepth,
+                        boolean withLockedMonitors, boolean withLockedSynchronizers) {
+            Blocker blocker = getBlockerInfo(thread);
+            StackTraceElement[] stackTrace = getStackTrace(thread, maxDepth);
+            LockedMonitors lockedMonitors = getLockedMonitors(thread, stackTrace.length);
+            boolean inNative = stackTrace.length > 0 && stackTrace[0].isNativeMethod();
+            Target_java_lang_management_ThreadInfo targetThreadInfo = new Target_java_lang_management_ThreadInfo(
+                            thread,
+                            getThreadState(thread, inNative),
+                            blocker.blockObject,
+                            blocker.ownerThread,
+                            JMXMonitoring.getThreadTotalBlockedCount(thread),
+                            JMXMonitoring.getThreadTotalBlockedTime(thread),
+                            JMXMonitoring.getThreadTotalWaitedCount(thread),
+                            JMXMonitoring.getThreadTotalWaitedTime(thread),
+                            stackTrace,
+                            withLockedMonitors ? lockedMonitors.monitorObjects : new Object[0],
+                            withLockedMonitors ? lockedMonitors.monitorDepths : new int[0],
+                            withLockedSynchronizers ? getLockedSynchronizers(thread) : new Object[0]);
+            return SubstrateUtil.cast(targetThreadInfo, ThreadInfo.class);
+        }
+    }
+
+    public static class DeadlockDetectionUtils {
+        /**
+         * Returns an array of thread ids of blocked threads within some given array of ThreadInfo.
+         *
+         * @param threadInfos array of ThreadInfo for the threads among which the deadlocks should
+         *            be detected
+         * @param byMonitorOnly true if we are interested only in the deadlocks blocked exclusively
+         *            on monitors
+         * @return array containing thread ids of deadlocked threads
+         */
+        public static long[] findDeadlockedThreads(ThreadInfo[] threadInfos, boolean byMonitorOnly) {
+            HashSet<Long> deadlocked = new HashSet<>();
+            for (ThreadInfo threadInfo : threadInfos) {
+                HashSet<Long> chain = new HashSet<>();
+                ThreadInfo current = threadInfo;
+                while (current != null && current.getLockInfo() != null && !deadlocked.contains(current.getThreadId())) {
+                    if (!chain.add(current.getThreadId())) {
+                        if (!byMonitorOnly || chain.stream().allMatch(DeadlockDetectionUtils::isBlockedByMonitor)) {
+                            deadlocked.addAll(chain);
+                        }
+                        chain.clear();
+                        break;
+                    }
+                    long currentLockOwnerId = current.getLockOwnerId();
+                    current = Arrays.stream(threadInfos).filter(ti -> ti.getThreadId() == currentLockOwnerId).findAny().orElse(null);
+                }
+            }
+            return deadlocked.stream().mapToLong(threadId -> threadId).toArray();
+        }
+
+        /**
+         * Anything that is deadlocked can be blocked either by monitor (the object related to
+         * JavaMonitor), or a lock (anything under AbstractOwnableSynchronizer).
+         *
+         * @return true if provided thread is blocked by a monitor
+         */
+        private static boolean isBlockedByMonitor(long threadId) {
+            return LockSupport.getBlocker(SubstrateThreadMXBean.getThreadById(threadId)) instanceof JavaMonitor;
+        }
+    }
+}
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/management/SubstrateThreadMXBean.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/management/SubstrateThreadMXBean.java
@@ -31,6 +31,8 @@ import java.util.Objects;
 
 import javax.management.ObjectName;
 
+import com.oracle.svm.core.jdk.ThreadMXUtils;
+import com.oracle.svm.core.thread.ThreadCpuTimeSupport;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.Platform;
 import org.graalvm.nativeimage.Platforms;
@@ -38,10 +40,12 @@ import org.graalvm.nativeimage.Platforms;
 import com.oracle.svm.core.Uninterruptible;
 import com.oracle.svm.core.jdk.UninterruptibleUtils.AtomicInteger;
 import com.oracle.svm.core.jdk.UninterruptibleUtils.AtomicLong;
+import com.oracle.svm.core.jdk.ThreadMXUtils.ThreadInfoConstructionUtils;
 import com.oracle.svm.core.thread.PlatformThreads;
-import com.oracle.svm.core.thread.ThreadCpuTimeSupport;
 
 import sun.management.Util;
+
+import static com.oracle.svm.core.Uninterruptible.CALLED_FROM_UNINTERRUPTIBLE_CODE;
 
 /**
  * This class provides a partial implementation of {@link com.sun.management.ThreadMXBean} for SVM.
@@ -55,9 +59,9 @@ public final class SubstrateThreadMXBean implements com.sun.management.ThreadMXB
     private final AtomicInteger peakThreadCount = new AtomicInteger(0);
     private final AtomicInteger threadCount = new AtomicInteger(0);
     private final AtomicInteger daemonThreadCount = new AtomicInteger(0);
-
     private boolean allocatedMemoryEnabled;
     private boolean cpuTimeEnabled;
+    private boolean contentionMonitoringEnabled;
 
     @Platforms(Platform.HOSTED_ONLY.class)
     SubstrateThreadMXBean() {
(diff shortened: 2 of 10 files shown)


roberttoyonaga:
Haven't look at this in detail yet, but full support for ThreadMXBean would be very nice to have! 

You may need to run `mx eclipseformat` for that failing gate to pass.

smthelusive:
Thank you for the tip @roberttoyonaga! Done 👍

roberttoyonaga (changes requested):
Hi @smthelusive ! It would be very nice to have this ThreadMXBean support. I left some comments/concerns below.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/MultiThreadedMonitorSupport.java
```diff
@@ -248,6 +248,9 @@ public void monitorEnter(Object obj, MonitorInflationCause cause) {
                 if (monitor == null) { // successful
                     JavaMonitorInflateEvent.emit(obj, startTicks, MonitorInflationCause.MONITOR_ENTER);
                     newMonitor.latestJfrTid = current;
+                    JavaThreads.JMXMonitoring.addThreadMonitor(newMonitor);
+                    newMonitor.blockedObject = obj;
+                    newMonitor.depth = Thread.currentThread().getStackTrace().length;
                     return;
```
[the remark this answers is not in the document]
smthelusive: Hi @roberttoyonaga! Thank you for reviewing my PR.

Let me clarify on this section a bit. I'm not actually setting a thread blocker here. It's been handled before me, and in the end I'm getting the blocker by reading `Thread.parkBlocker` field.
The problem is that when the thread is blocked upon a monitor, that field will contain an instance of `JavaMonitor`. `JavaMonitor` represents the monitor, but does not save any information about the original `Object` with that blocked monitor, so I need to save it when the new `JavaMonitor` is created.

I see that I did miss some things:
- I do it in fast path where we're creating a new `JavaMonitor`, however I'm not handling the use-case when we're creating new monitor for `additionalMonitors`, so I'll make a change;
- No need to set the `blockedObject` to `null` on `monitorExit` later in this class;
- I need to update the depth when re-entering the same monitor, and also on `monitorExit` when `monitor.acquisitions > 1`.

Please let me know if that makes sense.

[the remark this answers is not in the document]
roberttoyonaga: Ok I see. So you only ever need to write `blockedObject` upon monitor inflation. And you're relying on `LockSupport.getBlocker` to get the status of the thread before reading that field.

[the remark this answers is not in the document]
smthelusive: Yes, exactly 👍

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
```diff
@@ -0,0 +1,206 @@
...
+        private static ThreadInfo fromTarget(Target_java_lang_management_ThreadInfo targetThreadInfo) {
+            return ThreadInfo.class.cast(targetThreadInfo);
+        }
+    }
+
+    public static class DeadlockDetectionUtils {
+        private static final ReentrantLock lock = new ReentrantLock();
+        private static final HashSet<Long> deadlocked = new HashSet<>();
+        private static final HashSet<Long> chain = new HashSet<>();
+        private static ThreadInfo[] allThreadInfos = new ThreadInfo[0];
```
fniephaus, line 148: Is this global state really needed? Could you allocate the objects in `findDeadlockedThreads` and pass them to the other private helpers instead?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/management/SubstrateThreadMXBean.java
```diff
@@ -150,45 +152,61 @@ public int getDaemonThreadCount() {
...
     @Override
     public ThreadInfo getThreadInfo(long id) {
-        return null;
+        return getThreadInfo(id, -1);
     }
 
     @Override
     public ThreadInfo[] getThreadInfo(long[] ids) {
-        return new ThreadInfo[0];
+        return (ThreadInfo[]) Arrays.stream(ids).mapToObj(this::getThreadInfo).toArray();
```
fniephaus, line 175: Can you re-use `getThreadInfo(long[] ids, int maxDepth)` here?
```suggestion
        return getThreadInfo(ids, -1);
```

fniephaus (commented):
Thanks a lot for this PR, well done @smthelusive!
Since threading is rather important, the PR will need to be reviewed carefully. Please feel free to reach out if you have any questions or in case the PR (or you) get stuck. :wink: 

I'm going to run our internal gates on this. I will report any issues here...

fniephaus:
Here's a `ClassCastException` I ran into testing this:
```
java.lang.ClassCastException: java.util.concurrent.locks.ReentrantLock$NonfairSync cannot be cast to com.oracle.svm.core.locks.Target_java_util_concurrent_locks_AbstractOwnableSynchronizer
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.ThreadMXUtils.toTarget(ThreadMXUtils.java:56)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.ThreadMXUtils$ThreadInfoConstructionUtils.getBlockerInfo(ThreadMXUtils.java:87)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.ThreadMXUtils$ThreadInfoConstructionUtils.getThreadInfo(ThreadMXUtils.java:119)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.management.SubstrateThreadMXBean.getThreadInfo(SubstrateThreadMXBean.java:188)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.management.SubstrateThreadMXBean.lambda$getThreadInfo$2(SubstrateThreadMXBean.java:299)
	at java.base@21.0.1/java.util.stream.LongPipeline$1$1.accept(LongPipeline.java:177)
	at java.base@21.0.1/java.util.Spliterators$LongArraySpliterator.forEachRemaining(Spliterators.java:1223)
	at java.base@21.0.1/java.util.Spliterator$OfLong.forEachRemaining(Spliterator.java:777)
	at java.base@21.0.1/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base@21.0.1/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base@21.0.1/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:575)
	at java.base@21.0.1/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
	at java.base@21.0.1/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:616)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.management.SubstrateThreadMXBean.getThreadInfo(SubstrateThreadMXBean.java:300)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.management.SubstrateThreadMXBean.dumpAllThreads(SubstrateThreadMXBean.java:305)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.management.SubstrateThreadMXBean.findDeadlockedThreads(SubstrateThreadMXBean.java:283)
```

smthelusive:
Hi @fniephaus and @roberttoyonaga. Thank you both for the review! I've fixed the points that you mentioned and reworked some parts. 

Here is what changed: 

- Fix calling `Thread.getStackTrace` multiple times.
- Change the way to detect if the thread is running native code. `StackTraceElement` contains information whether the method is a native method. `IsolateThread` isn't necessary for this approach, which is nice.
- Optimise `SubstrateThreadMXBean.dumpAllThreads` flow, so it doesn't look up all threads twice.
- Reuse `getThreadInfo(long[] ids, int maxDepth)` as suggested.
- Collect monitor re-enters as stack. While testing without native image, I noticed that on monitor re-enter in the same Thread the Object is stored multiple times as a locked monitor, each time with different depth. So I changed the implementation and this also resulted in storing the depths properly.
- Handle all cases when `JavaMonitor` is created or reentered. This includes also handling the `MultiThreadedMonitorSupport.additionalMonitors` use-case for the objects that don't have the monitor offset.
- Not substitute the whole `AbstractOwnableSynchronizer` class.
- Lock monitors and locks when reading these collections and make that piece of code more reliable.
- Solve the problem with the stack frame offset when the monitor is entered. The problem: I want to know where in the stack the monitor was actually obtained in the user code. However, while handling the monitor creation, we are putting several extra frames on top of the stack and then remove them, so they are not in the stack at the moment when we're reporting anything. It is possible to detect where exactly the monitor actually happened by checking how many top frames have module/class names of our internal code, but it's expensive. A lot cheaper is to have a fixed offset, especially since it's a single use-case with the same flow, where we always have exactly 4 extra frames on the stack… I'm not a big fan of this solution, but couldn't come up with a better one.
- Simplified deadlock detection logic.
- Fixed the years in the headers.
- @fniephaus I couldn't reproduce the `ClassCastException`, but my changes should have fixed the potential of it happening. Could you please check?

Let me know if you have more suggestions, I will do my best to fix it. :)

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
```diff
@@ -0,0 +1,176 @@
...
+            return new Object[0];
+        }
+
+        private record LockedMonitors(Object[] monitorObjects, int[] monitorDepths) {
+        }
+
+        private static LockedMonitors getLockedMonitors(Thread thread, int stacktraceLength) {
+            ArrayDeque<JMXMonitoring.MonitorInfo> monitors = JMXMonitoring.getThreadMonitors(thread);
+            if (monitors != null) {
+                synchronized (monitors) {
```
roberttoyonaga, line 95: I believe there is still a race here.  I don't think it's enough to only synchronize here when we're reading the monitors. We also need to synchronize when we modify the stack of monitors in `removeThreadMonitor`  and `addThreadMonitor`.  For example, if `monitors` is modified in between when `monitorObjects` and `monitorDepths` is assigned, `monitorObjects` and `monitorDepths` won't be the same size.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
```diff
@@ -0,0 +1,176 @@
...
+                                SubstrateUtil.cast(synchronizer, Target_java_util_concurrent_locks_AbstractOwnableSynchronizer.class)
+                                                .getExclusiveOwnerThread());
+            }
+            return new Blocker(blocker, null);
+        }
+
+        private static Object[] getLockedSynchronizers(Thread thread) {
+            EconomicSet<AbstractOwnableSynchronizer> locks = JMXMonitoring.getThreadLocks(thread);
+            if (locks != null) {
+                synchronized (locks) {
```
roberttoyonaga, line 77: I believe there is still a race here. I don't think it's enough to only synchronize here when we're reading the locks. We also need to synchronize when we modify the set of locks . For example, if the set of locks is modified by the target thread in between when we assign `lockObjects` a size and when we finish iterating the set, then we may try to access an index that doesn't exist.

roberttoyonaga (commented):
Happy new year @smthelusive! Thank you for the changes. You've addressed most of my previous questions/concerns. I have left a few more below.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
```diff
@@ -0,0 +1,176 @@
...
+
+        private record LockedMonitors(Object[] monitorObjects, int[] monitorDepths) {
+        }
+
+        private static LockedMonitors getLockedMonitors(Thread thread, int stacktraceLength) {
+            ArrayDeque<JMXMonitoring.MonitorInfo> monitors = JMXMonitoring.getThreadMonitors(thread);
+            if (monitors != null) {
+                synchronized (monitors) {
+                    Object[] monitorObjects = monitors.stream().map(JMXMonitoring.MonitorInfo::originalObject).toArray();
+                    int[] monitorDepths = monitors.stream().mapToInt(monitorInfo -> stacktraceLength < 0 ? -1 : stacktraceLength - monitorInfo.stacksize()).toArray();
```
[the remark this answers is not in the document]
smthelusive: Hi @roberttoyonaga, happy new year! Thank you for your review. I will look into the comments and make the changes.
For this one, I made a drawing to illustrate what I'm doing here:
graal_pr

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
smthelusive: Hi @roberttoyonaga, I see.
So, only the thread that owns the monitors is modifying the monitors collection. The other thread is reading it here. I have made a change to get an unmodifiable copy of the monitors list, so then I can use it without worrying about changes in the original monitors collection. I think I shouldn't block the writes when I'm reading the collection.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
smthelusive: Hi @roberttoyonaga, indeed there was such possibility.
So, only one thread is adding or removing locks - it's the thread that owns them. Another thread is reading these values for reporting here. I did a small change here to use an iterator over the locks. This way I'm sure that any modifications of the underlying collection won't affect me when I'm reading the collection (and I think I shouldn't block the writes while reading it).

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/JavaThreads.java
```diff
@@ -428,6 +435,153 @@ static void setCurrentThread(Thread carrier, Thread thread) {
...
+                StackTraceElement[] stacktrace = currentThread.getStackTrace();
+
+                /*
+                 * During monitor creation we add several extra frames on top of the stack. At the
+                 * moment when monitor reporting via JMX is requested the stack doesn't contain them
+                 * anymore. The offset value below helps align with the real location where
+                 * synchronized block occurred.
+                 */
+                int stackSizeWithoutInternalLogic = -1;
+                if (MonitorInflationCause.MONITOR_ENTER.equals(cause)) {
```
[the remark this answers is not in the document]
smthelusive: I've made a change to check the module names instead.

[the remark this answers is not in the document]
roberttoyonaga: Thanks. This seems better now.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
roberttoyonaga: Thank you. I think this idea of getting a "snapshot" of the list is good. However, there is still the possibility that the list is modified while `List.copyOf()` is in-progress (race during construction).  I'm not sure how bad this could be because it depends on the implementation of  `List.copyOf()` , `ArrayDeque.push` `ArrayDeque.pop`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
roberttoyonaga: This is better than before, but doesn't prevent the race. While iterating the list here, the owning thread may concurrently remove items from the list. Will this cause problems?

smthelusive: Hi @roberttoyonaga,
when the values get removed, they are set to null (in an array), and on iterating, the null values are skipped. However, there might be a problem when the size shrinks, if that happens in between `hasNext()` and `next()` calls.
I've made a change. It's really convenient to continue using the `EconomicSet` since I can provide it with equivalence strategy that uses `System.identityHashCode`. So I decided to go for external synchronization. I'm not using the `ReadWriteLock` here because that adds an extra lock that the thread obtains while it's already busy adding a lock, so it will loop trying to add more lock entries. So the monitor it is.

I've also made a change for the monitors collection, where I'm using now `ConcurrentLinkedDeque`.

roberttoyonaga: > So the monitor it is

I think your new approach has solved the concurrency problem here. However, I think it's possible to easily optimize this with more fine-grained locking. You could maybe synchronize on the lock collection itself (`Target_java_lang_Thread#locks`) so multiple threads can work simultaneously on different collections.  Is there a reason you are using a [global lock](https://github.com/oracle/graal/pull/7945/files#diff-75f469d0228df9c99ef350ab92b31c4beccc00802f61903041bf45682a4b65c7R101) instead?

> I've also made a change for the monitors collection, where I'm using now ConcurrentLinkedDeque.

What is the purpose of this change? Changing to a `ConcurrentLinkedDeque` does not make the construction of the list copy operation atomic ([`return List.copyOf(targetThread.monitors);`](https://github.com/oracle/graal/pull/7945/files#diff-75f469d0228df9c99ef350ab92b31c4beccc00802f61903041bf45682a4b65c7R610)). So I still have the [same comment as before.](https://github.com/oracle/graal/pull/7945#discussion_r1463860189)

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/JavaThreads.java
```diff
@@ -430,6 +441,178 @@ static void setCurrentThread(Thread carrier, Thread thread) {
...
+                    calculateContentionOnThreadRunning(targetThread);
+                    break;
+            }
+        }
+
+        @Uninterruptible(reason = CALLED_FROM_UNINTERRUPTIBLE_CODE, mayBeInlined = true)
+        private static void setThreadBlockedToEnterLock(Target_java_lang_Thread thread) {
+            if (thread.lastStartedBlocked == NO_VALUE) {
+                thread.lastStartedBlocked = System.currentTimeMillis();
+                thread.blockedCount++;
```
roberttoyonaga, line 488: Are we guaranteed to always be dealing with the current thread? If so, maybe it's worth adding an assertion here. 

Otherwise, is it safe to be un-atomically modifying `thread.lastStartedBlocked`, `thread.blockedCount`, `thread.waitedCount`, etc ? Or do we require synchronization here as well?

smthelusive: For the thread states that I'm interested in, it's modified by the current thread only. 
However, there are use-cases such as when a new thread is initialized, then its state can be set by its parent just before the new thread is assigned as current. In this case, the thread state will be set to `NEW`, and I'm not updating anything for this state.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
smthelusive: Indeed, I shouldn't be using the global lock. Fixed that.

Regarding the `ConcurrentLinkedDeque`, I should have used the Iterator that it provides instead of `List.copyOf`. According to the documentation, this Iterator is weakly consistent:

> - they may proceed concurrently with other operations
> - they will never throw `ConcurrentModificationException`
> - they are guaranteed to traverse elements as they existed upon construction exactly once, and may (but are not guaranteed to) reflect any modifications subsequent to construction.

roberttoyonaga (approved):
Thank you for working on this @smthelusive! These changes look good to me now. 

It would be nice to also have some tests for this MXBean in the future as well.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/locks/Target_java_util_concurrent_locks_AbstractOwnableSynchronizer.java
```diff
@@ -0,0 +1,79 @@
...
+import com.oracle.svm.core.thread.JavaThreads;
+
+import java.util.concurrent.locks.AbstractOwnableSynchronizer;
+
+/**
+ * {@link java.util.concurrent.locks.AbstractOwnableSynchronizer} is substituted in order to track
+ * the locks per thread when JMX support is enabled.
+ */
+@SuppressWarnings({"unused"})
+@TargetClass(value = java.util.concurrent.locks.AbstractOwnableSynchronizer.class, onlyWith = JmxIncluded.class)
```
roberttoyonaga, line 41: One last minor comment. I've poorly named this:`JmxIncluded` is based on whether **remote** JMX support is included. I think we want to have this substitution even if remote JMX is excluded because we can still use the MBeanServer locally.

smthelusive:
Hi @fniephaus and @christianhaeubl! 
It looks like I need some more reviews for this PR, all the previous concerns were addressed :blush: 
Please let me know if you have any other comments, I will be happy to fix it.
Thank you!

fniephaus:
Hey @smthelusive, thanks for the update! I'm going to run the public and internal gates on your PR again. Hopefully, we soon have enough time to review your PR as well. We are currently busy integrating other contributions from @roberttoyonaga and others.

fniephaus:
This PR unfortunately got mirrored to #8430, sorry about that. I am adding fixes for the style gate so that other gates can run as well.

fniephaus:
Looks like our internal gate has found a problem with the PR:
```
Fatal error: java.lang.AssertionError: null
    at com.oracle.svm.core.thread.JavaThreads$JMXMonitoring.addThreadLock(JavaThreads.java:532)
    at java.util.concurrent.locks.AbstractOwnableSynchronizer.setExclusiveOwnerThread(AbstractOwnableSynchronizer.java:62)
    at java.util.concurrent.locks.ReentrantLock$NonfairSync.initialTryLock(ReentrantLock.java:226)
    at java.util.concurrent.locks.ReentrantLock$Sync.lock(ReentrantLock.java:152)
    at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:322)
    at jdk.internal.misc.InternalLock.lock(InternalLock.java:74)
    at java.io.PrintStream.write(PrintStream.java:788)
    at java.io.PrintStream.print(PrintStream.java:1002)
    at java.lang.ThreadGroup.uncaughtException(ThreadGroup.java:697)
    at java.lang.ThreadGroup.uncaughtException(ThreadGroup.java:690)
    at java.lang.Thread.dispatchUncaughtException(Thread.java:2901)
    at java.lang.VirtualThread.run(VirtualThread.java:311)
    at java.lang.VirtualThread$VThreadContinuation$1.run(VirtualThread.java:190)
```

This suggests that the implementation may not be working correctly with virtual threads. Could you take a look, please?

smthelusive:
Hi all, thank you for your reviews. 
Given the latest feedback in the [mirror PR](https://github.com/oracle/graal/pull/8430) I don't think that I can proceed with the current solution, hence closing this PR.
Thank you for your time, your help, and this opportunity to learn.

PR #7998, opened 2023-12-11
[GR-51017] Use JDK 17 for building Quarkus

zakkak:
Quarkus no longer supports JDK 11

See https://github.com/oracle/graal/actions/runs/7162366442/job/19499236553#step:12:2328

### DIFF
--- a/.github/workflows/quarkus.yml
+++ b/.github/workflows/quarkus.yml
@@ -128,6 +128,11 @@ jobs:
       with:
         name: graalvm
         path: graalvm.tgz
+    # Use Java 17 to build Quarkus as that's the lowest supported JDK version currently
+    - uses: actions/setup-java@v4
+      with:
+        distribution: 'temurin'
+        java-version: '17'
     - name: Build Quarkus
       run: |
         cd ${QUARKUS_PATH}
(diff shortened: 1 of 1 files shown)


.github/workflows/quarkus.yml
```diff
@@ -128,6 +128,12 @@ jobs:
       with:
         name: graalvm
         path: graalvm.tgz
+    # Use Java 17 to build Quarkus as that's the lowest supported JDK version currently
+    - uses: graalvm/setup-graalvm@v1
+      with:
+        version: 'latest'
```
fniephaus, line 134: `version` is deprecated. As much as I like using `setup-graalvm`, maybe it's better to use `setup-java` instead because here, we only care about compiling with javac?

zakkak: Mostly yes. We decided to use GraalVM just as a very basic test that Quarkus compiles with it.

zakkak: Updated to using Temurin builds of JDK 17

fniephaus: Thanks! Could we change this to use Oracle JDK? I can make the change unless you'd like to do it.

zakkak: Sure, please go ahead.

Issue #8020, opened 2023-12-14
Enabling JFR monitoring has side effects in Netty's initialization/configuration (and possibly in other libraries as well) that possibly skew the observed metrics.


zakkak:
**Describe the issue**
When enabling JFR monitoring (`--enable-monitoring=jfr`) we observe that some constants in `io.netty.util.internal.PlatformDependent` and `io.netty.util.internal.PlatformDependent0` are initialized to a different value.

**Steps to reproduce the issue**
```
cd /tmp
git clone --branch 2023-12-07-mandrel-it-tests-236 https://github.com/zakkak/issue-reproducers reproducer
cd reproducer
mvn package
java -agentlib:native-image-agent=config-output-dir=src/main/resources/META-INF/native-image -jar target/reproducer-1.0-SNAPSHOT.jar
mvn package
native-image -jar target/reproducer-1.0-SNAPSHOT.jar
./reproducer-1.0-SNAPSHOT | tee no-jfr.txt
native-image -jar target/reproducer-1.0-SNAPSHOT.jar --enable-monitoring=jfr
./reproducer-1.0-SNAPSHOT | tee jfr.txt
diff no-jfr.txt jfr.txt
```

This will result in a diff like the following:

```diff
9c9
< PlatformDependent0.DIRECT_BUFFER_CONSTRUCTOR = null
---
> PlatformDependent0.DIRECT_BUFFER_CONSTRUCTOR = private java.nio.DirectByteBuffer(long,long)
17c17
< PlatformDependent0.INTERNAL_UNSAFE = null
---
> PlatformDependent0.INTERNAL_UNSAFE = jdk.internal.misc.Unsafe@67141b8a
20c20
< PlatformDependent0.UNSAFE = sun.misc.Unsafe@41c5ea82
---
> PlatformDependent0.UNSAFE = sun.misc.Unsafe@4b7a3a11
43c43
< PlatformDependent.ALLOWED_LINUX_OS_CLASSIFIERS = [Ljava.lang.String;@1851c3b0
---
> PlatformDependent.ALLOWED_LINUX_OS_CLASSIFIERS = [Ljava.lang.String;@7e8b3bed
50,51c50,51
< PlatformDependent.USE_DIRECT_BUFFER_NO_CLEANER = false
< PlatformDependent.DIRECT_MEMORY_COUNTER = null
---
> PlatformDependent.USE_DIRECT_BUFFER_NO_CLEANER = true
[...]

PR #8037, opened 2023-12-15
Small JFR clean-up

roberttoyonaga:
### Summary
Some minor house cleaning in 3 areas:
1. Remove the "hack" in JfrEventSubstitution.java. It's no longer needed because JDK code has changed to make `EventWriter.reset()` public.
2. Added javadoc and logging in `Target_jdk_jfr_internal_JVM.java`. Currently, whenever a new JFR @Deprecated event is used an unclear error message is logged:
```
 [warn][jfr,setting] Exception occurred when setting value "forRemoval" for class jdk.jfr.internal.Control
Passed
```
One issue is that there may be too much useless logger output that drowns out meaningful logs. A possible solution is to simply log that a feature is unimplemented only once. Instead of logging every time an attempt to use it is made.  What do you think about this?
3. Avoid unnecessary module opening. [This caused issues with Quarkus](https://github.com/oracle/graal/issues/8020). I think this might simply be outdated code and is safe to remove now.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrFeature.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrFeature.java
@@ -149,8 +149,6 @@ public class JfrFeature implements InternalFeature {
 
     @Override
     public void afterRegistration(AfterRegistrationAccess access) {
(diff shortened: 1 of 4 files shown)


christianhaeubl (approved):
Thanks for the changes, I plan to merge them into master and 24.0.

With the extra logging that you added in `setMiscellaneous()`, two warnings will be printed per invocation (one because of the explicit logging, one because of the exception), which is not ideal. I will remove the exception throw so that only the more specific message gets printed.

PR #4783, opened 2022-08-02
[GR-40198] Provide public API for feature-based JNI / Resource / Proxy / Serialization registration.

graalvmbot:
This PR adds the missing APIs that will allow our users to not rely on our svm-internal APIs anymore.

PR #8041, opened 2023-12-16
[GR-50962] Add object reachability hooks.

graalvmbot:


kkriske:
`registerObjectReachableCallback` is currently only available in the `DuringSetupAccessImpl` class, which is not part of the public API.
Adding it to the `DuringSetupAccess` interface would allow end users to use this call instead of the object replacer as well.
@olpaw could this be considered ad a candidate to add to the public API? Tagging you because of the work you did on https://github.com/oracle/graal/pull/4783

kristofdho:
@christianwimmer In general cases where static objects need some additional work to trigger lazy loading of e.g. classes from files.
Something in this form:
```java
class Util {
    public static final ClassHolder EXAMPLE_CLASS = new ClassHolder("org.example.Example");
}

class ClassHolder {
    private final String className;
    private Class<?> classValue;

    public ClassHolder(String className) {
        this.className = className;
    }

    public Class<?> getClassValue() {
        if (classValue == null) {
            try {
                classValue = Class.forName(className);
                classValue = Object.class;
            }
        }
        return classValue;
    }
}
```

What we currently do, is initialize the `Util` class at build time, iterate over its fields, capture the ones of type `ClassHolder` and call `getClassValue` to trigger class loading and avoid reflection config.

I've now realized it would be easier to use object replacers, but then the `registerObjectReachableCallback` is intended to replace that specific usecase to avoid abusing the object replacers outside of their intended purpose. Hence the request to expose this to public API.

We have multiple similar, and more complex usecases, mostly aimed at reducing the need of reflection and/or resource config, because manual config, or even automated collection with the agent, is unfeasable.
One of these usecases is situated in Apache POI, which would require us to craft e.g. a test that uses all possible excel features to make sure there are no possible runtime failures due to missing reflection/resource.

PS: same person as @kkriske, but on this account I have access to the relevant resources.

Issue #4574, opened 2022-05-17
[native-image] Debug information not usable in gdb


ctiedt:
**Describe the issue**
When building a native image binary with the `-g` flag, the debugging information is not recognized by gdb. The specific error is `Dwarf Error: wrong version in compilation unit header (is 0, should be 2, 3, 4 or 5) [in module /home/clemens/Studium/MA/re22/helloworld]`.

**Steps to reproduce the issue**
Please include both build steps as well as run steps
1. Write a Java program (e.g. just a Hello World) and compile it to bytecode
2. Compile it to a native binary with `native-image -g HelloWorld`
3. Load the binary in gdb: `gdb helloworld`

**Describe GraalVM and your environment:**
 - GraalVM version: CE 22.1.0
 - JDK major version: 17
 - OS: Arch Linux
 - Architecture: AMD64
 - gdb: 12.1 (also tested on 10.2 and 11.2)

fniephaus:
Thanks for the bug report, @ctiedt. I tried to reproduce this but it seems that I can debug a helloworld with gdb 12.1 on archlinux. Maybe @adinn has an idea what could be causing the Dwarf Error that you are seeing?

adinn:
@ctiedt  Hi Clemens

If you can provide me with a copy of the broken the binary and a dump of the debug info I may be able to to debug this and see what the broken compile unit is (I would prefer not to try dumping debug info on my own, non-Alpine box). To produce the dump output please run these commands

    $ objdump --dwarf=info helloworld > helloworld.info.dw
    $ objdump --dwarf=abbrev helloworld > helloworld.abbrev.dw
    $ objdump -h helloworld > helloworld.headers.dw

I'm assuming that helloworld is the generated image and is in the current working directory. Also, that objdump is available on your build host.

I would be grateful if you could zip up the resulting output files, `helloworld.info.dw`, `helloworld.abbrev.dw`, `helloworld.headers.dw` plus the original binary `helloworld` and attach them to this issue.

ctiedt:
Hi Andrew, thank you very much for looking into this. The files you requested are attached below. All the `objdump` commands also gave a warning `Corrupt unit length (got 0x6c0000 expected at most 0xec2c9) in section .debug_info` that was not included in the dumps.

[helloworld-dbg.zip](https://github.com/oracle/graal/files/8709212/helloworld-dbg.zip)

adinn:
Hi Clemens,

[...]

Issue #3452, opened 2021-06-08
native-image generates corrupted debuginfo


avimalka:
**Describe the issue**
When generating native image with -g it's impossible to debug it with gdb as debuginfo section is corrupted.
When running gdb following message is showing:
```
Reading symbols from App...
Dwarf Error: wrong version in compilation unit header (is 0, should be 2, 3, 4 or 5) [in module /tmp/native-image/App]
(No debugging symbols found in App)
```
Also objdump show the following warning:
```
objdump: Warning: Debug info is corrupted, .debug_info header at 0xce has length 6c0000
```

**Steps to reproduce the issue**
native-image -g --no-fallback -H:DebugInfoSourceSearchPath=/tmp/graal-src.zip -jar App.jar

**Describe GraalVM and your environment:**
GraalVM version: 21.1.0
JDK major version: 11
OS: Linux Suse 12
Architecture: amd64

**More details**
objdump output: https://pastebin.com/AmLEWu1R
-H:Log=DebugTypeInfo:3,DebugCodeInfo:3,dwarf:3 output:  https://mega.nz/file/vIwTzIDB#k9NRGt5enSyb6hyzBfvvQ-o89MUepMHPuJ4hVc0Y2JQ (160MB)
Compiled file from my machine: https://mega.nz/file/qYo3QSbQ#X4Nz7szNDo1nKmny1EoULT3ZOadSDiO-z06Sgub-Mlk

adinn:
@avimalka Hi Avi,

I have had a look at the output log file, objdump output and the binary and I am fairly sure this problem is to do with the version of assembler you are using rather  than an error in the GraalVM Native generated debuginfo.

Can you provide me with a copy of the source code and class files in graal-src.zip and App.jar that you mention in the steps listed above? I would like to try building on my own machine to see if my version of gcc leads to the same problems.

I have provided an explanation of what I think is going wrong below.

----- 8< -------- 8< -------- 8< -------- 8< -------- 8< -------- 8< -------- 8< ---

The problem appears to be an extra pair of bytes in the info section.

The info dump shows 3 records before the one that has the wrong length:

      Compilation Unit @ offset 0x0:
       Length:        0x2a (32-bit)
       Version:       2
       Abbrev Offset: 0x0
       Pointer Size:  8
     <0><b>: Abbrev Number: 1 (DW_TAG_compile_unit)
        <c>   DW_AT_stmt_list   : 0x0
        <10>   DW_AT_low_pc      : 0x405200
[...]

PR #8046, opened 2023-12-18
 [GR-50985] Set alignment to 1 for Java DWARF debug info sections

esytnik:
This PR addresses cases when GraalVM sometimes generates corrupted debug info by ensuring that  debug sections have alignment set to 1.

Current we mistakenly generate debug info sections with alignment of 8 (for 64-bit OS) and this causes linker to generate corrupt debug info when it has several object files with debug info with different alignment. For example on Alpine Linux crt1.o and crti.o libraries have debug information section with alignment of 1. Linker writes those sections first and then pads to alignment of 8 before processing Graal-generated debug info section because it has alignment of 8.

Evidently calling `newUserDefinedSection` was an oversight as `newDebugSection` method already has a comment on this exact issue:
```
                                 // debugging information is mostly unaligned; padding can result in
                                 // corrupted data when the linker merges multiple debugging
                                 // sections from different inputs
```

See #https://github.com/oracle/graal/issues/4574 for more insight.

### DIFF
--- a/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/ELFObjectFile.java
+++ b/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/ELFObjectFile.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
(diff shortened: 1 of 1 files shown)


adinn:
Doh! Thank you for spotting this @esytnik !

adinn (approved):

adinn:
@olpaw @fniephaus Could you approve the workflow to check this patch and also help review and commit it?

adinn:
It should fix the problem reported in #4574

fniephaus (approved):
LGTM

esytnik:
> It should fix the problem reported in #4574

as well as https://github.com/oracle/graal/issues/3452

Issue #8075, opened 2023-12-21
java.lang.ExceptionInInitializerError when trying to create native image for program using XML processing


jerboaa:
### Describe the issue
When trying to generate a `native-image` from a Java program that happens to use the XML stack from the JDK (module `java.xml`), then there is a chance that a `ExceptionInInitializerError` happens during the `native-image` build because class loading during the native-image build isn't deterministic. Therefore, in programs in the wild where we have seen the issue, it sometimes happens. Sometimes not. Depending on class loading order and state of inline-before-analysis.

This is hard to reproduce with plain OpenJDK, since the public XML APIs carefully make sure that certain classes are being loaded in the correct order.

The reproducer provided simulates the issue we've observed.

### Steps to reproduce the issue

```
$ cat Test.java
import java.nio.file.Files;
import java.nio.file.Path;
import java.util.Arrays;

import javax.xml.parsers.DocumentBuilderFactory;

import org.w3c.dom.Document;

public class Test {

    private static final String CLASS = "com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl";
    
    private void loadFragmentScannerSingle() throws Exception {
        Class<?> fragmentClass = Class.forName(CLASS);
        System.out.println("Loaded: " + fragmentClass);
    }
    
    private void doXmlProcessing(Path xml) throws Exception {
        var dbf = DocumentBuilderFactory.newInstance();
        var db = dbf.newDocumentBuilder();
        Document doc;
        try (var is = Files.newInputStream(xml)) {
            doc = db.parse(is);
        }
        System.out.println("Child nodes: " + doc.getChildNodes().getLength());
        System.out.println("First child: " + doc.getChildNodes().item(0).getNodeName());
        loadFragmentScannerSingle();
    }
    
    public static void main(String[] args) throws Exception {
        if (args.length < 1) {
            System.out.println("Usage: Test <1|2 xml-file>");
            System.exit(1);
        }
        int number = Integer.parseInt(args[0]);
        Test t = new Test();
        if (number == 1) {
            t.loadFragmentScannerSingle();
        } else if (number == 2) {
            Path xml = Path.of(args[1]);
            t.doXmlProcessing(xml);
        } else {
            throw new IllegalArgumentException("Illegal value: " + Arrays.asList(args));
        }
    }

}
$ cat test.xml
<point>
    <x>1</x>
    <y>y</y>
</point>
$ javac Test.java
$ native-image Test
========================================================================================================================
[...]

PR #8076, opened 2023-12-21
Force xml class load order after JDK-8306055

jerboaa:
Here is a proposal to fix the issue described in #8075. By eagerly loading the `jdk.xml.internal.JdkXmlUtils` class which only contains static constants, the analysis is then free to scan the class space and (potentially) initialize classes in any order - since we make sure `jdk.xml.internal.JdkXmlUtils` has already been loaded in the host JVM.

Passes the simple reproducer and also seems to fix the issues we've seen in Quarkus CI:

```
$ native-image Test
GraalVM Native Image: Generating 'test' (executable)...
[1/8] Initializing...                                                                                    (6.6s @ 0.17GB)
 Java version: 23-beta+2-ea, vendor version: Mandrel-24.1.0-dev
 Graal compiler: optimization level: 2, target machine: x86-64-v3
 C compiler: gcc (redhat, x86_64, 13.2.1)
 1 user-specific feature(s):
 - com.oracle.svm.thirdparty.gson.GsonFeature
Build resources:
 - 26.49GB of memory (42.3% of 62.56GB system memory, determined at start)
 - 12 thread(s) (100.0% of 12 available processor(s), determined at start)
[2/8] Performing analysis...  [***]                                                                     (13.3s @ 0.50GB)
    4,217 reachable types   (75.8% of    5,562 total)
    6,392 reachable fields  (47.9% of   13,333 total)
   21,478 reachable methods (49.8% of   43,115 total)
    1,354 types,    87 fields, and   703 methods registered for reflection
       58 types,    58 fields, and    52 methods registered for JNI access
[3/8] Building universe...                                                                               (1.9s @ 0.45GB)
[4/8] Parsing methods...      [*]                                                                        (1.5s @ 0.59GB)
[5/8] Inlining methods...     [***]                                                                      (1.0s @ 0.55GB)
[6/8] Compiling methods...    [****]                                                                    (12.6s @ 0.46GB)
[7/8] Laying out methods...   [*]                                                                        (2.0s @ 0.56GB)
[8/8] Creating image...       [*]                                                                        (1.8s @ 0.64GB)
   9.03MB (44.48%) for code area:    12,776 compilation units
  10.94MB (53.90%) for image heap:  121,684 objects and 47 resources
 336.62kB ( 1.62%) for other data
  20.30MB in total
Top 10 origins of code area:                                Top 10 object types in image heap:
   4.49MB java.base                                            2.65MB byte[] for code metadata
   3.15MB java.xml                                             1.63MB byte[] for java.lang.String
 986.68kB svm.jar (Native Image)                               1.19MB java.lang.String
 112.55kB java.logging                                         1.00MB java.lang.Class
  63.88kB org.graalvm.nativeimage.base                       523.07kB heap alignment
  47.59kB jdk.proxy1                                         513.00kB int[][]
  45.84kB jdk.proxy3                                         410.61kB byte[] for general heap data
  26.79kB jdk.internal.vm.ci                                 362.40kB com.oracle.svm.core.hub.DynamicHubCompanion
  21.97kB org.graalvm.collections                            266.02kB java.util.HashMap$Node
  11.42kB jdk.proxy2                                         247.65kB java.lang.Object[]
  16.78kB for 4 more packages                                  2.20MB for 1105 more object types
Recommendations:
 HEAP: Set max heap for improved and more predictable memory usage.
 CPU:  Enable more CPU features with '-march=native' for improved performance.
                        1.9s (4.2% of total time) in 222 GCs | Peak RSS: 1.11GB | CPU load: 8.07
Produced artifacts:
 /home/sgehwolf/Documents/mandrel/bugs/upstream-graal/issue_xml_jdk22/testme/test (executable)
Finished generating 'test' in 42.3s.
$ ./test 2 test.xml 
Child nodes: 1
First child: point
Loaded: class com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl
```

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
@@ -36,6 +36,7 @@ import static com.oracle.svm.hosted.xml.XMLParsersRegistration.TransformerClasse
 
 import com.oracle.svm.core.feature.InternalFeature;
 import com.oracle.svm.core.jdk.JNIRegistrationUtil;
+import com.oracle.svm.util.ReflectionUtil;
 import com.oracle.svm.core.feature.AutomaticallyRegisteredFeature;
 
 @AutomaticallyRegisteredFeature
@@ -43,6 +44,14 @@ public class JavaxXmlClassAndResourcesLoaderFeature extends JNIRegistrationUtil 
 
     @Override
     public void beforeAnalysis(BeforeAnalysisAccess access) {
+        // Ensure that class loading of xml related classes work
+        // com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl
+        // implicitly depends on jdk.xml.internal.JdkXmlUtils. So initialization
+        // order needs to happen in the reverse order.
+        Class<?> jdkUtil = access.findClassByName("jdk.xml.internal.JdkXmlUtils");
+        if (jdkUtil != null) {
+            ReflectionUtil.newInstance(jdkUtil);
+        }
         access.registerReachabilityHandler(new SAXParserClasses()::registerConfigs,
                         method(access, "javax.xml.parsers.SAXParserFactory", "newInstance"));
 
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
@@ -44,10 +44,12 @@ public class JavaxXmlClassAndResourcesLoaderFeature extends JNIRegistrationUtil 
 
     @Override
     public void beforeAnalysis(BeforeAnalysisAccess access) {
-        // Ensure that class loading of xml related classes work
-        // com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl
-        // implicitly depends on jdk.xml.internal.JdkXmlUtils. So initialization
-        // order needs to happen in the reverse order.
+        /*
+         * Ensure that class loading of xml related classes work.
+         * com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl
+         * implicitly depends on jdk.xml.internal.JdkXmlUtils. So initialization
+         * order needs to happen in the reverse order.
+         */
         Class<?> jdkUtil = access.findClassByName("jdk.xml.internal.JdkXmlUtils");
         if (jdkUtil != null) {
             ReflectionUtil.newInstance(jdkUtil);
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
@@ -36,7 +36,7 @@ import static com.oracle.svm.hosted.xml.XMLParsersRegistration.TransformerClasse
 
 import com.oracle.svm.core.feature.InternalFeature;
 import com.oracle.svm.core.jdk.JNIRegistrationUtil;
-import com.oracle.svm.util.ReflectionUtil;
+import com.oracle.svm.hosted.classinitialization.ClassInitializationSupport;
 import com.oracle.svm.core.feature.AutomaticallyRegisteredFeature;
 
 @AutomaticallyRegisteredFeature
@@ -46,13 +46,13 @@ public class JavaxXmlClassAndResourcesLoaderFeature extends JNIRegistrationUtil 
     public void beforeAnalysis(BeforeAnalysisAccess access) {
         /*
          * Ensure that class loading of xml related classes work.
-         * com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl
-         * implicitly depends on jdk.xml.internal.JdkXmlUtils. So initialization
-         * order needs to happen in the reverse order.
+         * com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl implicitly depends
+         * on jdk.xml.internal.JdkXmlUtils. So initialization order needs to happen in the reverse
+         * order.
          */
         Class<?> jdkUtil = access.findClassByName("jdk.xml.internal.JdkXmlUtils");
         if (jdkUtil != null) {
-            ReflectionUtil.newInstance(jdkUtil);
+            ClassInitializationSupport.singleton().forceInitializeHosted(jdkUtil, "break circular class initialization in XML code", false);
         }
         access.registerReachabilityHandler(new SAXParserClasses()::registerConfigs,
                         method(access, "javax.xml.parsers.SAXParserFactory", "newInstance"));


oubidar-Abderrahim:
Hi @jerboaa, thank you for your contribution. I believe JDK-8306055 was already addressed in https://github.com/oracle/graal/commit/933c6602f620a203c80a86d4e9d789bb7d8fb56b 
cc @christianwimmer @zapster

jerboaa:
@oubidar-Abderrahim https://github.com/oracle/graal/commit/933c6602f620a203c80a86d4e9d789bb7d8fb56b doesn't fix this native-image build issue, AFAIK. Whether or not it shows up at native-image build time depends on the class loading order which the static analysis doesn't explicitly control at image build time. https://github.com/oracle/graal/commit/933c6602f620a203c80a86d4e9d789bb7d8fb56b only addresses the JdkCatalog issue at image runtime, AFAIK.

See #8075 for all the details.

jerboaa:
Gentle ping. Any more thoughts on this?

zapster:
@jerboaa thanks of the patch and sorry for the delay. I already had a look at some point but then forgot about it again.

hard-initializing `JdkXmlUtils` seems a bit rough, although in that particular case it is probably fine. I'd prefer something reachability handler based. Have you experimented in that direction?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
```diff
@@ -36,13 +36,22 @@
...
 
 @AutomaticallyRegisteredFeature
 public class JavaxXmlClassAndResourcesLoaderFeature extends JNIRegistrationUtil implements InternalFeature {
 
     @Override
     public void beforeAnalysis(BeforeAnalysisAccess access) {
+        // Ensure that class loading of xml related classes work
+        // com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl
+        // implicitly depends on jdk.xml.internal.JdkXmlUtils. So initialization
+        // order needs to happen in the reverse order.
```
zapster, line 50: General remark: for multi-line text, we prefer block comments (`/* */`) and full sentences with terminal punctuation, because that plays better with the formatter wrt rearranging line breaks. No blocker, just a hint for the future. :)

jerboaa:
> hard-initializing `JdkXmlUtils` seems a bit rough, although in that particular case it is probably fine. I'd prefer something reachability handler based. Have you experimented in that direction?

@zapster I had, but this turned out to be difficult. The issue at hand is class loading order in the hosted VM. I.e. when some reachability triggers, it might already be too late and `XMLDocumentFragmentScannerImpl` or some other XML class depending on `JdkXmlUtils` might have already been loaded as this is done in parallel and non-deterministic. Therefore, I failed to see how exactly this could be done in order for it to work reliably in all cases.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
jerboaa: OK, I'll fix.

zapster:
@jerboaa thanks. I'm also playing around with reachability handlers locally without luck.

What puzzles me about the original error is that the JDKs initialization protocol is somehow bypassed. In the test case, for example, instantiating `XMLDocumentFragmentScannerImpl` out of the blue is just not supposed to work. That is why this test fails on hotspot as well. So I'm wondering what causes the actual problem in the production code? Maybe there is a similar problem there that should be fixed instead? I looked at some of the stack traces from the linked issues, but nothing caught my eye. Do you have any more insights?

jerboaa:
> Maybe there is a similar problem there that should be fixed instead? I looked at some of the stack traces from the linked issues, but nothing caught my eye. Do you have any more insights?

It's related to the combination of build time initialization (that quarkus uses) and the static analysis stack in GraalVM which can - in some cases - trigger the wrong class loading order. This combination basically triggers - in the failure case - a case described in this comment:
https://github.com/quarkusio/quarkus/issues/37657#issuecomment-1858317205

There are all sorts of add-opens/add-exports at play as well.

The test got modelled around the failure case which can be triggered using hotspot (yet is an anti-pattern), but is hard to do with only using public JDK xml APIs (I failed to do that).

zapster:
@jerboaa thanks again.

> The test got modelled around the failure case which can be triggered using hotspot (yet is an anti-pattern), but is hard to do with only using public JDK xml APIs (I failed to do that).

Do you have by any chance a simple reproducer that works on hotspot but not on native image?

jerboaa:
> @jerboaa thanks again.
> 
> > The test got modelled around the failure case which can be triggered using hotspot (yet is an anti-pattern), but is hard to do with only using public JDK xml APIs (I failed to do that).
> 
> Do you have by any chance a simple reproducer that works on hotspot but not on native image?

Unfortunately not. The best we have is the synthesized one in https://github.com/oracle/graal/issues/8075.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
jerboaa: Should be fixed now.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
```diff
@@ -36,13 +36,24 @@
...
     public void beforeAnalysis(BeforeAnalysisAccess access) {
+        /*
+         * Ensure that class loading of xml related classes work.
+         * com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl
+         * implicitly depends on jdk.xml.internal.JdkXmlUtils. So initialization
+         * order needs to happen in the reverse order.
+         */
+        Class<?> jdkUtil = access.findClassByName("jdk.xml.internal.JdkXmlUtils");
+        if (jdkUtil != null) {
+            ReflectionUtil.newInstance(jdkUtil);
```
christianwimmer, line 55: Use `RuntimeClassInitialization.initializeAtBuildTime(jdkUtil)` because as far as I understand it you only care that the class gets initialized, there is no need to actually instantiate the class.

zapster: Unfortunately, that does not work because it would initialize it too late. For the same reason, reachability handlers and `--initialize-at-build-time` does not work.

zapster: Maybe something like:
```
        ClassInitializationSupport.singleton().forceInitializeHosted(jdkUtil, "break circular class initialization in XML code", false);
```

christianwimmer: Hm, that seems to be a difference between our old `ProvenSafeClassInitializationSupport` and the new default `AllowAllHostedUsagesClassInitializationSupport`. The old policy immediately initialized the class when it was registered, which is actually a good thing I think - we also have other places where we need to mitigate class initialization deadlocks, and that seems like a simple way to do it.

jerboaa: I've used `forceInitializeHosted()` now.

jerboaa:
I can also reference [JDK-8323740](https://bugs.openjdk.org/browse/JDK-8323740) in the comment if that's desired.

zapster:
> I can also reference [JDK-8323740](https://bugs.openjdk.org/browse/JDK-8323740) in the comment if that's desired.

I don't think it is necessary, but feel free to do so if you like.

zapster (dismissed):

zapster:
LGTM I'll get this integrated.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
zapster: @christianwimmer thinks we should investigate changing the `AllowAllHostedUsagesClassInitializationSupport` first. This might open up different opportunities to solve the original problem. I'll give that a try come back to you.

zapster: With the change suggested by @christianwimmer, one could use `--initialize-at-build-time=jdk.xml.internal.JdkXmlUtils` to mitigate the problem.
@jerboaa would that be acceptable for Quarkus as well?

jerboaa: I'll have to check and will get back to you later today or tomorrow. We don't really have a quarkus reproducer, though. So verification will be getting a fix integrated and then rely on many CI runs to see if the problem persists after a fix has been pushed.

zapster: Ok sounds good. FYI, #8185 is the `AllowAllHostedUsagesClassInitializationSupport` change I'm currently running through our CI, in case you want to have a peek.

zakkak: Quarkus initializes everything at build time (except those few classes that can't be initialized at build time or are causing issues) so `jdk.xml.internal.JdkXmlUtils` as well is registered for build-time initialization (although I expect this to be the case even outside Quarkus since it's a core JDK class).

I have also confirmed this by running one of the Quarkus tests that triggers the issue with `-H:+PrintClassInitialization`.

Having said that I believe the suggested solution would work out of the box for Quarkus.

cc @jerboaa

zapster: That would be great. So I will continue with #8185 and we see if the issues on your side disappear once it is merged.

jerboaa: Sounds good to me.

jerboaa:
Closing this PR since https://github.com/oracle/graal/pull/8193 is supposed to fix this.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
jerboaa: @zapster I'm rebooting this conversation. https://github.com/oracle/graal/pull/8185 or the replacement, https://github.com/oracle/graal/pull/8193 So I'm reopening this PR.

jerboaa:
> Closing this PR since #8193 is supposed to fix this.

Re-opening since the issue isn't fixed.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
```diff
@@ -36,13 +36,24 @@
...
     public void beforeAnalysis(BeforeAnalysisAccess access) {
+        /*
+         * Ensure that class loading of xml related classes work.
+         * com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl implicitly depends
+         * on jdk.xml.internal.JdkXmlUtils. So initialization order needs to happen in the reverse
+         * order.
+         */
+        Class<?> jdkUtil = access.findClassByName("jdk.xml.internal.JdkXmlUtils");
+        if (jdkUtil != null) {
+            ClassInitializationSupport.singleton().forceInitializeHosted(jdkUtil, "break circular class initialization in XML code", false);
```
christianwimmer, line 55: I created #8256 because then the registration of the individual class is at the same place that registers the whole package for initialization at build time, i.e., when we remove the package we will remember to also remove the class.

jerboaa: Sounds good. Thank you.

jerboaa: https://github.com/oracle/graal/pull/8256 fixed this.

jerboaa:
https://github.com/oracle/graal/pull/8256 was the actual fix for this (more thorough CI testing still needed on our end, though).

PR #8077, opened 2023-12-22
Support Java agent instrumentation in native image

ziyilin:
This PR supports Java agent instrumentation in native image.
### Scope
Java agent can instrument and transform runtime Java classes in two ways:
1. premain: Java agent is loaded before main class starts. The classes are transformed at their first loading time.
2. main: Java agent is attached to a running JVM. The classes are transformed after the agent attached time.
 
This PR only supports the first kind of Java agent instrumentation, i.e. the premain way.

### General Idea
The general idea of this PR is to compile the recorded transformed class into native image. 
It is implemented in three stages:
1. Interception stage: native-image-agent records all the transformed classes, dynamic generated classes, and the target agent's premain method in a run.
2. Build time stage: Instrumented classes are prepended to the beginning of imageCp, so they will override their original classes at build time.
3. Runtime stage: Add `premain` execution in `JavaMainWrapper` before executing the main method. The recorded agent premain method shall be executed at this moment. 

### Agent Developers' Work
Due to the complexity of agent instrumentation, this PR cannot automatically do all the work. It relies on the agent developer to provide the following input:
1. JDK class transformation. If a JDK class is modified by the agent, the agent developer should provide a substituted class (with GraalVM's `@TargetClass` APIs) of such class as well. 
2. Agent premain method implementation. This PR supports running premain method in native image to initiate the instrumentation. But the agent developer should make sure the code running in native image works well. For example, a premain method can do two things, 1) turn a global flag on, which activates all the instrumented code; and 2) do bytecode transformation.  In native image only the first task should be executed, the second task is not necessary and cannot be executed. 

### Simple Demo
Here is a simple demo project. 
[agent-demo.zip](https://github.com/oracle/graal/files/13748844/agent-demo.zip)

The following steps can show how this PR works:
1. Build this PR and assume the result is placed at `$GRAALVM_HOME`.
2. Unzip the file.
3. Run `mvn package` to build the demo project.
4. Run `java -cp application/target/application-1.0-SNAPSHOT.jar org.example.Main`. The result should be
  
    This is foo
    This is bar
    0

6. Run `collectInstrument.sh` to collect instrumentation configurations. Now the program is instrumented, the output should be
   
   before foo
   This is foo 
   before foo
   This is bar
   2

8. Run `build.sh` to build the demo native image.
9. Run the built native image. The result should be the same as step 6. 

### Opentelemetry Demo


christianwimmer:
> Instrumented classes are prepended to the beginning of imageCp, so they will override their original classes at build time.

How does that work for modules? As far as I know, It is not possible with the Java module system to provide classes for one package from two different .jar files.

ziyilin:
> > Instrumented classes are prepended to the beginning of imageCp, so they will override their original classes at build time.
> 
> How does that work for modules? As far as I know, It is not possible with the Java module system to provide classes for one package from two different .jar files.

Good question. Module system is not considered yet in this version of implementation. I have another version that dumps transformed classes as classdata files at pre-run time, then loads them at image build time to substitute the original ones. I think that will solve the module system problem. But I encountered some other issues with that solution, so turned to current easier solution. 

I'll see how to make it work with module system.

NicklasWallgren:
Has there been any progress toward adding support for Java agents?

@vjovanov @christianwimmer

CH3CHO:
Would the final solution support transforming JDK built-in classes, such as `Thread`, `ThreadExecutorPoolExecutor`, etc.? Because currently, we have them transformed to add some byte code for tracing and metrics collection, which won't work after switching to native image.

ziyilin:
There will be a major update of this PR very soon. I have been working on issues that are exposed when we apply this PR with Opentelemetry agent.
At this moment, we have successfully collected opentelemetry agent based monitoring data on the native version of Spring boot, MySql, Redis and Kafka demo applications.

ziyilin:
> Would the final solution support transforming JDK built-in classes, such as `Thread`, `ThreadExecutorPoolExecutor`, etc.? 

This issue will be addressed in the upcoming update of this PR. But it does not automatically apply your transformed JDK classes into native image, because GraalVM itself depends on those classes as well. You need to write static transformation class with APIs provided by this PR. The details will be published soon.

vjovanov:
@ziyilin I have opened an internal PR to merge, but to run the tests we first need to resolve conflicts.

zonk911:
Hi Folks, any news on that topic? I was hoping for that since a while.

fatekingsama:
Hi Folks, any news on that topic? I was hoping for that since a while.

vjovanov:
This feature didn't make it for GraalVM 25.0 due to our focus on improving the correctness and error reporting of reflection. We will re-prioritize this for the future release. 

We also need to re-think how this feature is going to be implemented, given that we might have an ability to interpret Java bytcode code at runtime in the future.

PR #8126, opened 2024-01-09
[GR-51205] Avoid needlessly rescanning black objects.

graalvmbot:


zakkak:
Hi @peter-hofer, do you have any measurements you could share regarding the runtime impact of this change?

peter-hofer:
@zakkak, the impact is really rather small, only two benchmarks see an improvement of execution time on the order of 1-3%.

PR #8230, opened 2024-01-23
[GR-52129] Augment RuntimeClassInitialization#initializeAtBuildTime javadoc

zakkak:
Document that using `""` as the package name results in all packages and
classes being registered for build time initialization.

### DIFF
--- a/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeClassInitialization.java
+++ b/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeClassInitialization.java
@@ -150,6 +150,10 @@ public final class RuntimeClassInitialization {
     /**
      * Registers all classes in provided packages as eagerly initialized during image-build time.
      * <p>
+     * Passing {@code ""} as a package, registers all packages and classes for initialization at
+     * build time. This might have unintended side-effects and should thus be used with great
+     * caution.
+     * <p>
      * All static initializers of {@code classes} will be executed during image-build time and
      * static fields that are assigned values will be available at runtime. {@code static final}
      * fields will be considered as constant.


fniephaus:
Thanks for the PR, @zakkak. Aren't you documenting more of a bug or do you actually make use of this? Registering everything for build-time initialization seems rather dangerous.

zakkak:
> Thanks for the PR, @zakkak. Aren't you documenting more of a bug or do you actually make use of this? Registering everything for build-time initialization seems rather dangerous.

This is actually a feature for Quarkus, one that we specifically requested to keep being able to use when the option to use `--initialize-at-build-time` without arguments was deprecated (https://github.com/oracle/graal/commit/5013ca7c9d88bd12de2ebd24e3196a40d9044c9c).

Although build-time initialization can be a double-edged sword Quarkus is still using it by default for all classes due to its benefits despite requiring more work to be done on the framework side.

Corresponding Quarkus Issue https://github.com/quarkusio/quarkus/issues/17887

fniephaus:
Understood. From a pure user's perspective, this is still concerning: what if users calculate the value dynamically and it happens to be `""`. Similar to `null` values, maybe we want to guard from accidental misuse somehow? What do you think, @vjovanov?

vjovanov:
I don't see another way without making a breaking change. I would not be worried as this is used only by the power users.

sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/hosted/RuntimeClassInitialization.java
```diff
@@ -150,6 +150,8 @@ public static void initializeAtRunTime(String... packages) {
     /**
      * Registers all classes in provided packages as eagerly initialized during image-build time.
      * <p>
+     * Passing {@code ""} as a package, registers all packages and classes for initialization at build time.
```
fniephaus, line 153: I'd be inclined to add a disclaimer like "Use at your own risk as this can have unintended side-effects.". Even if Quarkus leverage this, we shouldn't really recommend this to other users.

zakkak: I added "This might have unintended side-effects and should thus be used with great caution."

fniephaus (approved):
LGTM

PR #8251, opened 2024-01-25
[GR-48683] Add JFR event OldObjectSample.

graalvmbot:
Based on https://github.com/oracle/graal/pull/6743.


galderz (approved):
All good. Thanks for the thorough refactoring and in particular thanks for showing how the testing backdoor can be used. There are not many examples of that in the oracle/graal source tree unfortunately.

galderz:
There are some style issues that should be fixed.

galderz:
One final thing @christianhaeubl, in case we want to backport this work to another GraalVM version, I was wondering if you're changes and my PR could be squashed and we put us both as co-authors?

roberttoyonaga (approved):

christianhaeubl:
> ... in case we want to backport this work to another GraalVM version, I was wondering if you're changes and my PR could be squashed and we put us both as co-authors?

On our side, we prefer having individual commits.

I guess it would make backporting easier if we got rid of the merge commit that is between your and my commit. However, for that I would need to rebase this branch, which changes the hash of your commit and puts me as a co-author on your commit. As far as I know, this is not preferred by Red Hat.

galderz:
Ok, let's leave it as is then.

christianhaeubl:
@roberttoyonaga, @galderz  : I pushed one more commit where I did the following changes:
- resolve the conflicts between this PR and https://github.com/oracle/graal/pull/8254 (FlightRecorderOptions)
- changed the FlightRecorderOptions so that they match HotSpot on JDK21
    - removed `globalbuffercount`
    - renamed `repositorypath` to `repository`
    - renamed `thread_buffer_size` to `threadbuffersize`
- improve the error reporting if something fails during JFR argument parsing

Please double-check and let me know if you find any issues.

galderz:
All looks good, but I see CI still failing with style issues.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrManager.java
```diff
@@ -81,9 +82,18 @@ public static JfrManager get() {
...
-            parseFlightRecorderLogging(SubstrateOptions.FlightRecorderLogging.getValue());
+            parseFlightRecorderLogging();
+            parseFlightRecorderOptions();
+        };
+    }
+
+    public static RuntimeSupport.Hook startupHook() {
+        return isFirstIsolate -> {
+            parseFlightRecorderLogging();
+            parseFlightRecorderOptions();
```
roberttoyonaga, line 96: Why do have to do parse again here if we already did it in the previous hook?

christianhaeubl: Thanks, I removed it.

galderz:
Thanks @christianhaeubl!

PR #8266, opened 2024-01-29
Allow control of `throwDeniedThreadAccess` via `TruffleContext.threadAccessDeniedHandler`

### LINKED ISSUE #8033: Let languages control multi threaded access ... is not allowed for language(s) js
**Feature request**

The [Enso language](http://enso.org) implementation has managed to significantly simplify its interoperability with JavaScript in https://github.com/enso-org/enso/pull/7882 thanks to `OtherContextTruffleObject` concept available in recent versions of GraalVM.

I am especially thankful for removal of two thousand lines in https://github.com/enso-org/enso/pull/7882/files#diff-ccad61bee766711b658faf056e1a3f9381d9a05a9537edb2ac85774d039f0f6dL89 and related files.

**Is your feature request related to a problem? Please describe.**

However the functionality isn't completely on par - our original code implemented a GIL to avoid:
```
java.lang.IllegalStateException: Multi threaded access requested by thread main but is not allowed for language(s) js.
at com.oracle.truffle.polyglot.PolyglotEngineException.illegalState(PolyglotEngineException.java:133)
at com.oracle.truffle.polyglot.PolyglotContextImpl.throwDeniedThreadAccess(PolyglotContextImpl.java:1393)
at com.oracle.truffle.polyglot.PolyglotLanguageContext.checkThreadAccess(PolyglotLanguageContext.java:300)
```

However there doesn't seem to be a way for us to implement such GIL now. We can live with it for now (as Enso runs single-threaded), but Enso isn't single-threaded language and we will need a solution in the future.

**Describe the solution you'd like.**

We'd like to **control what happens** instead of `IllegalStateException: Multi threaded access ... is not allowed`. Options:
- *specify by an option* that the code shall block and wait until the access is allowed rather than throwing `IllegalStateException`
- *specify* `onEnter` and `onExit` callback that would allow to synchronize access by locking ourselves
- *specify* `Executor` that would be asked (by its `execute(Runnable)`) method to perform all interop operations sent to `OtherContextTruffleObject`
- make Graal.js multi-threading ready 

According to @kustosz (the creator of Enso language Truffle based interpreter) the current _"Multi threaded access requested by thread main but is not allowed"_ isn't intuitive for people embedding Truffle languages into their Java applications at all. 

I'd add: The behavior may be correct, but it is the least helpful of all possible behaviors. As Enso example shows, implementing the useful behavior requires quite a lot of code.

> GraalVM users, GraalVM contributors, developers of libraries and frameworks which depend on GraalVM, or somebody else?

People who embed Truffle languages will find any kind of improvement far more friendly than the current behavior.

We can obviously return those [proxy files](https://github.com/enso-org/enso/pull/7882/files#diff-ccad61bee766711b658faf056e1a3f9381d9a05a9537edb2ac85774d039f0f6dL89) back and implement everything ourselves. I'd however prefer not doing so.

**Additional context.**

Enso provides rich support for embedding other languages: [foreign section specification](https://github.com/enso-org/enso/blob/56cc9561b1d76835ad03efce80311ac505d3ae3b/docs/polyglot/polyglot-bindings.md#embedded-syntax-usage-syntax). Using other Truffle languages easily and smoothly straightens Enso as well as Truffle!

**Express whether you'd like to help contributing this feature**

One can simulate the problem in Enso test suite by removing [@Ignore annotation](https://github.com/enso-org/enso/pull/7882/files#diff-d1aa42f833254735760555fd5bfa3e89188839435cddc76e2f1282ba93ca33faR79) from the test that runs JavaScript in parallel, but I can work on standalone examples that would simulate the behavior using non-Enso languages as well.

The implementation in `OtherContextTruffleObject` is probably too complicated to be enhanced by an outsider like me. It is a core feature behind _native isolates_ (not sure what's the real name) and it is hard to contribute to it without having access to the _native isolates_. Easier if @tzezula does it ;-)

JaroslavTulach:
Let's fix #8033 by controlling what happens when `throwDeniedThreadAccess` is called.

The simplest way to address #8033 is to replace the current action in `throwDeniedThreadAccess` - e.g. throw `PolyglotEngineException.illegalState` with a **callback to embedder** to decide what to do - _don't throw, but retry_. However _calling foreign code_ at the moment of `throwDeniedThreadAccess` is slightly dangerous as internal lock on `PolyglotContextImpl` is being held. As such, let's start with the first step 82b14c347bba1254a993d8e041a478f5211ddd5f - let's make sure the exception is always **thrown without holding internal lock**(s). To do so, I am introducing an [internal checked exception](https://github.com/oracle/graal/pull/8266/files#diff-47943d4a28651c36b2200943e94e316dab289b539a4922bae9b46cae53b4d9c1R43) and propagating it up to the point where the **lock isn't held**. Only **then it is converted** to original `PolyglotEngineException.illegalState` exception and thrown. The fact that the exception is checked, instructs the `javac` to check all paths that may yield the `PolyglotThreadAccessException` and help me to never forget to handle the conversion. The unusual `loop: for (;;) {... continue loop.... break loop}` style is used as a preparation for the next step... 

The [next step](https://github.com/oracle/graal/pull/8266/commits/c4e3624dc79f440d624c22ab1d22a62dce5358a7) then defines `Builder.onDeniedThreadAccess(Consumer<RuntimeException> handler)` which allows the embedders to decide whether to throw the exception or **do something more intrinsic**.

The third step is to [use the API](https://github.com/oracle/graal/pull/8266#issuecomment-1916014216). Great, it [seems to work](https://github.com/enso-org/enso/pull/8885/files#r1470552562) as planned.


JaroslavTulach:
With `onDeniedThreadAccess` as proposed by this PR the embedders have a way to [act like an Ethernet on collision](https://github.com/enso-org/enso/pull/8885/commits/160f2b81617f459a1a2bf198f940c78e7e7db9ed#diff-a84ceb8a331a33ef55d986aac04b55c48c52f2daddf6bea30911823ab24ff3a7R24) and successfully orchestrate the two threads accessing the same context to run into completion.

More info on [Ethernet collision detection and behavior](https://en.wikipedia.org/wiki/Carrier-sense_multiple_access_with_collision_detection) - most important _"Calculate and wait the random backoff"_.

chumer:
I like the idea in principle.

What I like about this approach:
* Does not need any checks on the fast-path.
* Flexibility to implement your own strategy in case of multi-threaded access without allowing actual multi-threaded access.
* Careful considerations of current locking strategy.

What I don't like about this approach:
* It makes already quite complex control flow even more complex. I think some of the changes could be improved for better readability.
* It only works if you exclusively use single-threaded languages. Note that we force transition to multi-threaded as soon as language is initialized that depends on it to avoid late deoptimization. We would need to make sure we would need to change: 
```
     boolean transitionToMultiThreading = isSingleThreaded() && hasActiveOtherThread(true, false);
```
to something like this (just a sketch)

```
     boolean multiThreadedAccess = (isSingleThreaded() || multiThreadedCallback) &&  hasActiveOtherThread(true, false);
```
* Following up the previous problem I think we should change the API to be a callback whenever a multi-threaded access is performed.  (e.g. Builder#onMultiThreadedAccess)
* We lock ourselves in with the current approach of doing slow-path and fast-path thread transitions. If we ever want to have fast transitions (enter/leave) from multiple threads this will be harder to do as we need to keep supporting the callback. 
* Further investigation and testing needed for toggling behavior. What if multiple threads try to enter/leave concurrently. Can it happen that no thread ever wins?
* Its a complex API to use correctly. Lots of traps there. So we need extensive documentation of dos and don'ts. Ideally also examples on how to use it.
* We also need to extend the TruffleContext API in a complete implementation.

JaroslavTulach:
> I like the idea in principle.

OK, I'll write some unit tests and then we can continue the review. I am not sure I fully understand all the comments, for example:

> only works if you exclusively use single-threaded languages

Sure, I only care about Graal.js - all the other languages we are interested in are mutli threading ready. All I need is to make when there is a collision, it doesn't end up that fatal are right now and there is a way to recover. I don't get the comment about multi-threaded languages. We can sort it out when I prepare some unit tests for review. But I'd prefer to keep my PR as minimal as possible - solving complete Truffle corner cases is job for a professional, not for an outsider like me.

> quite complex control flow even more complex

Do you mean those `for (;;) { ... continue ... }` loops? They look more scary in the `diff` than in reality. I just had to shift the indentation and that makes the change look bigger than it is.

I'll find some time this week and add some test cases to this PR.

JaroslavTulach:
The first test added in addb748 verifies that all the ways that originally yielded `illegalState` exception can reach the `onDeniedThreadAccess` callback which can then throw some other exception.

The other test (added in f91faeb)  verifies that the `onDeniedThreadAccess` callback can _request a retry_ and then the threaded access succeeds.

@chumer, what additional tests you want me to provide?

JaroslavTulach:
Hello @chumer, is there anything else I shall do to get this API change in?

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
```diff
@@ -1180,6 +1182,23 @@ public Builder allowCreateThread(boolean enabled) {
             return this;
         }
 
+        /**
```
chumer, line 1185: The corresponding TruffleContext API is missing.

chumer (changes requested):
A great step forward. Some of the things I previously commented on were unfortunately not adressed. Added a few more comments.

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextImpl.java
```diff
@@ -837,131 +837,147 @@ Object[] enterThreadChanged(boolean enterReverted, boolean pollSafepoint, boolea
                     assert !mustSucceed;
                     throw PolyglotEngineException.illegalState("Context cannot be entered in polyglot thread's beforeEnter or afterLeave notifications.");
                 }
-                boolean needsInitialization = false;
-                synchronized (this) {
-                    PolyglotThreadInfo threadInfo = getCurrentThreadInfo();
+                PolyglotThreadAccessException threadAccessException = null;
+                LOOP: for (;;) {
```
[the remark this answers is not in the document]
JaroslavTulach: Fixed in 38926cf99e0

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
```diff
@@ -1180,6 +1182,23 @@ public Builder allowCreateThread(boolean enabled) {
...
+         * yielded by default. By installing this {@code handler} one can control what shall happen.
+         * Either to throw the provided exception or resolve the multithreaded situation and return
+         * to retry the thread access again.
+         *
+         * @param handler callback that either throws the provided {@code RuntimeException} or
+         *            returns to signal request for retry
+         * @return this builder
+         * @since 23.2
+         */
+        public Builder onDeniedThreadAccess(Consumer<RuntimeException> handler) {
```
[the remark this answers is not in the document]
JaroslavTulach: Done in 19f1f072290

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotThreadAccessException.java
```diff
@@ -0,0 +1,59 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.oracle.truffle.polyglot;
+
+final class PolyglotThreadAccessException extends Exception {
```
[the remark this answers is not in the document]
JaroslavTulach: Using empty `fillInStackTrace` since: 729c52a951975deaf296cd540fce30e576cede0c

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
JaroslavTulach: I am not really sure what's missing? Shall some methods be added to [TruffleContext](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/TruffleContext.html)?

truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/MultiThreadedLanguageTest.java
```diff
@@ -141,6 +142,86 @@ public void testNoThreadAllowed() {
...
+        MultiThreadedLanguage.isThreadAccessAllowed = (req) -> {
+            return false;
+        };
+
+        try {
+            context.initialize(MultiThreadedLanguage.ID);
+            fail();
+        } catch (PolyglotException e) {
+            assertTrue(e.getMessage().contains("SecurityException"));
+            assertTrue(e.isInternalError());
```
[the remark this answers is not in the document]
JaroslavTulach: This is a copy of a [previous testNoThreadAllowed](https://github.com/oracle/graal/pull/8266/files/22746a8e03bf693c77aa927d5e4377b86367d186#diff-dcb3fed13df86c2b8a0e30cd2a5f480d46d078cad7b79274e2db178be1562fc4L119). That test throws `IllegalStateException`.

Do you want me to modify this test (and code) to throw `SecurityException` directly?

truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/MultiThreadedLanguageTest.java
```diff
@@ -141,6 +142,86 @@ public void testNoThreadAllowed() {
...
+
+        // allow again so we can close
+        MultiThreadedLanguage.isThreadAccessAllowed = (req) -> {
+            return true;
+        };
+        context.close();
+    }
+
+    @Test
+    public void testNoThreadAllowedRetry() {
```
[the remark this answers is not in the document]
JaroslavTulach: Multi-threaded showcase written in dd0c464

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
```diff
@@ -1180,6 +1182,23 @@ public Builder allowCreateThread(boolean enabled) {
...
+         * Installs handler to control what happens on multiple thread access. When multiple threads
+         * are accessing a context which isn't ready for multithreaded access an exception is
+         * yielded by default. By installing this {@code handler} one can control what shall happen.
+         * Either to throw the provided exception or resolve the multithreaded situation and return
+         * to retry the thread access again.
+         *
+         * @param handler callback that either throws the provided {@code RuntimeException} or
+         *            returns to signal request for retry
+         * @return this builder
+         * @since 23.2
```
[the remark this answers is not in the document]
JaroslavTulach: The version should be 24.2 if this gets merged by August/September.

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
JaroslavTulach: Christian says to add the method to https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/TruffleContext.Builder.html

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/EngineAccessor.java
```diff
@@ -1150,7 +1150,14 @@ public Thread createThread(Object polyglotLanguageContext, Runnable runnable, Ob
                 threadContext = innerContext.getContext(threadContext.language);
             }
             PolyglotThread newThread = new PolyglotThread(threadContext, runnable, group, stackSize, beforeEnter, afterLeave);
-            threadContext.context.checkMultiThreadedAccess(newThread);
```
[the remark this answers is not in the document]
JaroslavTulach: Good idea. Done in acd67766bae

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
```diff
@@ -1180,6 +1182,23 @@ public Builder allowCreateThread(boolean enabled) {
             return this;
         }
 
+        /**
+         * Installs handler to control what happens on multiple thread access. When multiple threads
+         * are accessing a context which isn't ready for multithreaded access an exception is
+         * yielded by default. By installing this {@code handler} one can control what shall happen.
+         * Either to throw the provided exception or resolve the multithreaded situation and return
+         * to retry the thread access again.
+         *
```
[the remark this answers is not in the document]
JaroslavTulach: Example provided in 2604dd4efff this is how the documentation for the method looks like:

onDeniedThreadAccess

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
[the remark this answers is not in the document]
JaroslavTulach: Version (optimistically) set to 24.2 in 2604dd4efff

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
```diff
@@ -1180,6 +1183,30 @@ public Builder allowCreateThread(boolean enabled) {
...
+         * snippet: {@snippet file="org/graalvm/polyglot/Context.java" region="retry"}
+         *
+         *
+         * @param handler callback that either throws the provided {@code RuntimeException} or
+         *            returns to signal request for retry
+         * @return this builder
+         * @since 24.2
+         */
+        public Builder onDeniedThreadAccess(Consumer<RuntimeException> handler) {
+            this.onDeniedThreadAccess = handler;
```
chumer, line 1206: I would feel more comfortable if this is only TruffleContext API for now.

chumer (changes requested):

JaroslavTulach:
I tried to use this change in Enso. I am struggling. Enso still uses `24.0.0` - upgrading to `24.2.0-SNAPSHOT` doesn't seem to be easy at all (I have to give up otherwise I don't make it for GraalVM meetup). Backporting the change in this PR back to 24.0.0 isn't without issues either. But overall I think: **Enso needs just the `TruffleContext` change**. Following patch seems to trigger the right bits:
```diff
diff --git engine/runtime-fat-jar/src/main/java/module-info.java engine/runtime-fat-jar/src/main/java/module-info.java
index 87816c9723..9d577e04bb 100644
--- engine/runtime-fat-jar/src/main/java/module-info.java
+++ engine/runtime-fat-jar/src/main/java/module-info.java
@@ -9,6 +9,7 @@ open module org.enso.runtime {
   // works.
   requires org.enso.profiling;
   requires org.enso.ydoc;
+  requires org.graalvm.collections;
   requires org.graalvm.polyglot;
   requires org.graalvm.truffle;
   requires static org.slf4j;
diff --git engine/runtime-integration-tests/src/test/java/org/enso/interpreter/test/ForeignMethodInvokeTest.java engine/runtime-integration-tests/src/test/java/org/enso/interpreter/test/ForeignMethodInvokeTest.java
index 3cb3722f50..6a2822161e 100644
--- engine/runtime-integration-tests/src/test/java/org/enso/interpreter/test/ForeignMethodInvokeTest.java
+++ engine/runtime-integration-tests/src/test/java/org/enso/interpreter/test/ForeignMethodInvokeTest.java
@@ -13,7 +13,6 @@ import org.graalvm.polyglot.Context;
 import org.graalvm.polyglot.Value;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
 import org.junit.Test;

 public class ForeignMethodInvokeTest {
@@ -91,7 +90,6 @@ public class ForeignMethodInvokeTest {
     assertEquals(12, res2.getArrayElement(2).asInt());
   }

-  @Ignore
   @Test
   public void testParallelInteropWithJavaScript() throws Exception {
     var source =
diff --git engine/runtime-language-epb/src/main/java/org/enso/interpreter/epb/EpbContext.java engine/runtime-language-epb/src/main/java/org/enso/interpreter/epb/EpbContext.java
index 0eed392545..929818ed01 100644
--- engine/runtime-language-epb/src/main/java/org/enso/interpreter/epb/EpbContext.java
+++ engine/runtime-language-epb/src/main/java/org/enso/interpreter/epb/EpbContext.java
@@ -44,6 +44,7 @@ final class EpbContext {
       if (innerContext == null) {
         innerContext =
             env.newInnerContextBuilder()
+                .onDeniedThreadAccess(this::deniedThreadAccess)
                 .initializeCreatorContext(true)
                 .inheritAllAccess(true)
                 .config(INNER_OPTION, "yes")
@@ -52,6 +53,14 @@ final class EpbContext {
     }
   }

+  private void deniedThreadAccess(Throwable t) {
+    try {
+      Thread.sleep(10);
+    } catch (InterruptedException ex) {
+      // go on
+    }
+  }
+
   /**
    * @param node the location of context access. Pass {@code null} if not in a node.
    * @return the proper context instance for the current {@link
```
I am going to remove the API changes in `org.graalvm.polyglot.Context` API.

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
JaroslavTulach: Done in e6e076e

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
```diff
@@ -2154,3 +2181,30 @@ private void validateSandbox(SandboxPolicy useSandboxPolicy) {
         }
     }
 }
+
+final class Snippets {
```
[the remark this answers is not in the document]
JaroslavTulach: Removed together with the API change in e6e076e

However it is surprising you aren't using the official JDK `@snippet` tag and sticking to the errorprone _code in HTML_ style, I have to say.

truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/MultiThreadedLanguageTest.java
```diff
@@ -143,6 +148,157 @@ public void testNoThreadAllowed() {
...
+
+            // allow again so we can close
+            MultiThreadedLanguage.isThreadAccessAllowed = (req) -> {
+                return true;
+            };
+        }
+    }
+
+    @Test
+    public void testUsingSingleThreadedLanguageFromMultipleThreads() throws Throwable {
```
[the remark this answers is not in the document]
JaroslavTulach:
```bash
graal/truffle$ mx unittest MultiThreadedLanguageTest --repeat 10000
...........................................................................
Time: 16,952.849

OK (160000 tests)


10 longest running test classes after running 1 of 1 classes:
  3,021,411 ms    com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest
10 longest running tests:
        112 ms    testMultiThreading[VIRTUAL_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
         98 ms    testMultiThreading[PLATFORM_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
         16 ms    testAccessTruffleContextFromExclusivePolyglotThread[PLATFORM_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
          6 ms    testAccessTruffleContextFromExclusivePolyglotThread[VIRTUAL_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
          1 ms    testAccessTruffleContextPolyglotThread[PLATFORM_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
          0 ms    testAsssertionIfThreadStillActive[PLATFORM_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
          0 ms    testInterruptPolyglotThread[PLATFORM_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
          0 ms    testMultiThreadedAccessExceptionThrownToCreator[VIRTUAL_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
          0 ms    testSingleThreading[VIRTUAL_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
          0 ms    testInterruptPolyglotThread[VIRTUAL_THREADS](com.oracle.truffle.api.test.polyglot.MultiThreadedLanguageTest)
```

truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/MultiThreadedLanguageTest.java
```diff
@@ -143,6 +148,157 @@ public void testNoThreadAllowed() {
...
+                    retryCounter.notifyAll();
+                }
+            } catch (InterruptedException ex) {
+                throw new IllegalStateException(ex);
+            }
+        }).build()) {
+            var anotherThread = new Thread("Another thread") {
+                Value proxyValue;
+                Value result;
+                Throwable err;
```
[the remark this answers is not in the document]
JaroslavTulach: The test is gone since e6e076e2290da43d8cd331504fd85a801dbdbb8f that removed the `org.graalvm.polyglot` API changes.

truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/MultiThreadedLanguageTest.java
```diff
@@ -143,6 +148,157 @@ public void testNoThreadAllowed() {
...
+        }
+    }
+
+    @Test
+    public void testUsingSingleThreadedLanguageFromMultipleThreads() throws Throwable {
+        var retryCounter = new int[1];
+        try (Context ctx = Context.newBuilder(SingleThreadedLanguage.ID).onDeniedThreadAccess((t) -> {
+            try {
+                // wait a bit and try again
+                Thread.sleep(10);
```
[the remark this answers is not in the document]
JaroslavTulach: The test is gone since e6e076e2290da43d8cd331504fd85a801dbdbb8f that removed the `org.graalvm.polyglot` API changes.

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotThreadAccessException.java
```diff
@@ -0,0 +1,68 @@
...
+    PolyglotThreadAccessException(String message) {
+        super(message);
+    }
+
+    @Override
+    public Throwable fillInStackTrace() {
+        return this;
+    }
+
+    void rethrow(PolyglotContextImpl context) {
```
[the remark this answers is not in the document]
JaroslavTulach: Using your code in e775d5e. Yes, it works better. I had to change the test as now the exception message is propagated. Thank you.

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/Context.java
```diff
@@ -1180,6 +1183,30 @@ public Builder allowCreateThread(boolean enabled) {
...
+         * To implement <em>"retry later on collision"</em> strategy use for example following
+         * snippet: {@snippet file="org/graalvm/polyglot/Context.java" region="retry"}
+         *
+         *
+         * @param handler callback that either throws the provided {@code RuntimeException} or
+         *            returns to signal request for retry
+         * @return this builder
+         * @since 24.2
+         */
+        public Builder onDeniedThreadAccess(Consumer<RuntimeException> handler) {
```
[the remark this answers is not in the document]
JaroslavTulach: Renamed in 3661c02af4e7fc413c07206b9bfd38a4792d3fe6

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/impl/AbstractPolyglotImpl.java
```diff
@@ -799,7 +799,9 @@ public abstract Object createContext(Object receiver, SandboxPolicy sandboxPolic
                         Object hostAccess,
                         Object polyglotAccess,
                         boolean allowNativeAccess,
-                        boolean allowCreateThread, boolean allowHostClassLoading, boolean allowInnerContextOptions, boolean allowExperimentalOptions,
+                        boolean allowCreateThread,
+                        Consumer<RuntimeException> onDeniedThreadAccess,
```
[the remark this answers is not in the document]
JaroslavTulach: Changed to `Consumer<String>` in 3661c02af4e7fc413c07206b9bfd38a4792d3fe6

JaroslavTulach:
@chumer, I have addressed all your API change requests. Please review again. Right now I am struggling with endless loop in
```
graal/truffle$ mx build && mx unittest TruffleContextTest
```
It is a result of a bad merge. I'll address it soon. However I don't expect it to impact the observable API - e.g. you can review even now.

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotThreadAccessException.java
```diff
@@ -0,0 +1,68 @@
...
+        assert !Thread.holdsLock(context) : "Only rethrow without holding internal lock";
+        PolyglotEngineException ex = PolyglotEngineException.illegalState(getMessage());
+        if (context.config.onDeniedThreadAccess != null) {
+            try {
+                context.config.onDeniedThreadAccess.accept(ex);
+            } catch (RuntimeException another) {
+                throw new PolyglotEngineException(another);
```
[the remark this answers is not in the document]
JaroslavTulach: This code has disappeared at the end.

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/impl/AbstractPolyglotImpl.java
```diff
@@ -803,7 +803,9 @@ public abstract Object createContext(Object receiver, SandboxPolicy sandboxPolic
                         Object hostAccess,
                         Object polyglotAccess,
                         boolean allowNativeAccess,
-                        boolean allowCreateThread, boolean allowHostClassLoading, boolean allowInnerContextOptions, boolean allowExperimentalOptions,
+                        boolean allowCreateThread,
+                        Consumer<String> onDeniedThreadAccess,
```
chumer, line 807: No longer needed no?

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleContext.java
```diff
@@ -757,6 +757,7 @@ public final class Builder {
         private Boolean allowPolyglotAccess;
         private Boolean allowEnvironmentAccess;
         private ZoneId timeZone;
+        private Consumer<String> onDeniedThreadAccess;
```
chumer, line 760: please rename also private fields to threadAccessDeniedHandler and parameters everywhere:

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleContext.java
```diff
@@ -1036,6 +1037,23 @@ public Builder allowPolyglotAccess(boolean b) {
...
+         * yielded by default. By installing this {@code handler} one can control what shall happen.
+         * Either to throw exception (with the provided reason) or to resolve the multithreaded
+         * situation somehow and return to retry the thread access again.
+         *
+         * @param handler callback (that gets a reason as an input) that either throws an exception
+         *            or returns to signal a <b>request for retry</b>
+         * @return this builder
+         * @since 24.2
+         */
+        public Builder threadAccessDeniedHandler(Consumer<String> handler) {
```
chumer, line 1052: Since we keep this in TruffleContext we can go back to Consumer < RuntimeException >

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextConfig.java
```diff
@@ -256,7 +258,7 @@ private static Map<String, String> computeCommonOptions(Map<String, String> opti
     PolyglotContextConfig(PolyglotEngineImpl engine, SandboxPolicy sandboxPolicy, Boolean forceSharing,
                     OutputStream out, OutputStream err, InputStream in,
                     boolean hostLookupAllowed, Object polyglotAccess, boolean nativeAccessAllowed,
-                    boolean createThreadAllowed, boolean hostClassLoadingAllowed,
+                    boolean createThreadAllowed, Consumer<String> onDeniedThreadAccess, boolean hostClassLoadingAllowed,
```
chumer, line 261: old name here

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextConfig.java
```diff
@@ -278,6 +280,7 @@ private static Map<String, String> computeCommonOptions(Map<String, String> opti
         this.polyglotAccess = polyglotAccess;
         this.nativeAccessAllowed = nativeAccessAllowed;
         this.createThreadAllowed = createThreadAllowed;
+        this.onDeniedThreadAccess = onDeniedThreadAccess;
```
chumer, line 283: old name

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextConfig.java
```diff
@@ -109,6 +109,7 @@ final class PolyglotContextConfig {
     final Runnable onCancelled;
     final Consumer<Integer> onExited;
     final Runnable onClosed;
+    final Consumer<String> onDeniedThreadAccess;
```
chumer, line 112: old name

chumer (changes requested):
minor changes requested

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleContext.java
JaroslavTulach: I still assume that one day people demand this API in `org.graalvm.polyglot`. When that happens, it is better for the API to be a-like to this `TruffleContext` API.

I believe you were right there is no value in using `RuntimeException`. The message seems to work well.

Thus I prefer to keep `Consumer<String>` argument.

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextConfig.java
JaroslavTulach: Renamed in 3d2506c

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextConfig.java
JaroslavTulach: Renamed in 3d2506c

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextConfig.java
JaroslavTulach: Renamed in 3d2506c

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleContext.java
JaroslavTulach: Renamed in 3d2506c

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/TruffleContext.java
chumer: you are right String is better.

sdk/src/org.graalvm.polyglot/src/org/graalvm/polyglot/impl/AbstractPolyglotImpl.java
JaroslavTulach: Removed in 29a5199. However, as I [argued here](https://github.com/oracle/graal/pull/8266#discussion_r1767009663) I am convinced there will be a time when the `Consumer<String>` will get into `org.graalvm.polyglot` as well. Then `git show 29a5199 | patch -R -p1` will be needed. But of course, the code doesn't need to be changed by this PR.

chumer: I don't like to keep dead code. Thanks for removing.

JaroslavTulach:
Change log updated in 7149f99

JaroslavTulach:
Squashed in 85854c02d5b7bce0e436a31deffb762257c17d46

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotLanguageContext.java
```diff
@@ -619,69 +619,81 @@ void ensureCreated(PolyglotLanguage accessingLanguage) {
...
+                                creatingThread = null;
+                            }
+                            created = true;
                         }
                     }
+                    break LOOP;
+                }
+            } finally {
+                if (!wasCreated && eventsEnabled) {
+                    EngineAccessor.INSTRUMENT.notifyLanguageContextCreateFailed(context.engine, context.creatorTruffleContext, language.info);
```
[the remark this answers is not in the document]
jchalou: Please address my above comment. The behavior should be exactly the same if the threadDeniedAccessHandler is not set. Don't move the finally block outside of the synchronized block. You can put `threadAccessException == null` condition into the finally block or something, so that the LanguageContextCreateFailed event is not executed multiple times in case of retries. Moreover, your current solution is firing the LanguageContextCreateFailed event even if the language context is already created, which could be easily fixed by setting the `wasCreated` to true in that case, but instead please do the larger suggested change.

truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/MultiThreadedLanguage.java
```diff
@@ -52,11 +52,11 @@
 @TruffleLanguage.Registration(id = MultiThreadedLanguage.ID, name = MultiThreadedLanguage.ID)
 public class MultiThreadedLanguage extends TruffleLanguage<LanguageContext> {
 
-    static final String ID = "MultiThreadedLanguage";
+    public static final String ID = "MultiThreadedLanguage";
 
-    static final ThreadLocal<Function<Env, Object>> runinside = new ThreadLocal<>();
+    public static final ThreadLocal<Function<Env, Object>> runinside = new ThreadLocal<>();
```
jchalou, line 57: runinside does not have to be public

truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/ContextSharingTest.java
```diff
@@ -320,6 +320,7 @@ public void testThreadCheck() throws InterruptedException, ExecutionException {
 
     @Registration(id = SINGLE_THREAD, name = SINGLE_THREAD)
     public static class SingleThreadedLanguage extends TruffleLanguage<Env> {
+        public static final String ID = SINGLE_THREAD;
```
jchalou, line 323: not used

jchalou (changes requested):

truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/ContextSharingTest.java
JaroslavTulach: Fixed in 75e60df. No longer needed because we removed `org.graalvm.polyglot.Context.Builder` API and its associated test.

truffle/src/com.oracle.truffle.api.test/src/com/oracle/truffle/api/test/polyglot/MultiThreadedLanguage.java
JaroslavTulach: Fixed in 75e60df. No longer needed because we removed `org.graalvm.polyglot.Context.Builder` API and its associated test.

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotThreadAccessException.java
```diff
@@ -0,0 +1,78 @@
...
+                consumer.accept(getMessage());
+            } catch (IllegalStateException e) {
+                throw new PolyglotEngineException(e);
+            } catch (Throwable another) {
+                if (context.parent != null) { // inner context
+                    // inner context. no wrapping needed
+                    throw another;
+                } else {
+                    // embedder context
+                    throw PolyglotImpl.hostToGuestException(context.engine, another);
```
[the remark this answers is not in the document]
JaroslavTulach: Done in f55bbdb3649

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotLanguageContext.java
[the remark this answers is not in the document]
JaroslavTulach: Great catch! Addressed in b8e8590caa5.

The goal is to cause _no changes_ to the current behavior of the code except looping. According to diff in my IDE that hides the whitespace changes:

diff

We only wrap the existing code with the `loop: for (;;) {` and don't change any other behavior since b8e8590caa5.

jchalou (approved):

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotLanguageContext.java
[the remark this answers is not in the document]
jchalou: Thanks, I am also fine with triggering the notifyLanguageContextCreate and notifyLanguageContextCreateFailed multiple times in case of retries.

chumer (approved):

JaroslavTulach:
> Jaroslav, I am about to merge your work. In case you want your name on it, please resolve conflict, apply the following diff:

It is not enough to apply the patch. Conflict still remains.

jchalou:
> > Jaroslav, I am about to merge your work. In case you want your name on it, please resolve conflict, apply the following diff:
> 
> It is not enough to apply the patch. Conflict still remains. I had to add [a810ee8](https://github.com/oracle/graal/commit/a810ee85dc38c94acdfb2f50c7aff8f61b0d3500) - hopefully it is the last time I need to manually merge conflicts in `PolyglotContextImpl`.
> 
> Finally **squashed** in [b7f3b4c](https://github.com/oracle/graal/commit/b7f3b4c2b2bc5081f6d9b151116f481cf44591a5)

Thanks, the patch was not intended to resolve the conflict, these we just further changes that needed to be applied after the conflict was resolved.

JaroslavTulach:
I see the [threadAccessDeniedHandler](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/TruffleContext.Builder.html#threadAccessDeniedHandler(java.util.function.Consumer)) in the official Javadoc. Thank you for your help accepting my patch!

PR #8287, opened 2024-01-31
[GR-51491] Adapt to ICStubs removal.

graalvmbot:


fisk:
Hmm. In the generated code you show from before my change, I'm surprised that you seem to decode the receiver klass on AMD64 (which looks correct), but seemingly not decoding any compressed class pointer on AArch64, which to me seems a bit surprising. Why is that?

fisk:
> > Hmm. In the generated code you show from before my change, I'm surprised that you seem to decode the receiver klass on AMD64 (which looks correct), but seemingly not decoding any compressed class pointer on AArch64, which to me seems a bit surprising. Why is that?
> 
> 
> 
> The output comes from a run with `-XX:-UseCompressedClassPointers`.

Okay good.

PR #8316, opened 2024-02-05
[GR-47109] Support JFR event ObjectAllocationSample.

graalvmbot:
Based on https://github.com/oracle/graal/pull/6899. This PR contains larger cleanups and changes the code so that it is a lot closer to the original C++ sources (makes maintenance easier).

@roberttoyonaga : please double-check this PR and let me know if I removed anything relevant.


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/throttling/JfrAdaptiveSampler.java
```diff
@@ -0,0 +1,201 @@
...
+                /* Recheck under lock if the current window is still expired. */
+                if (activeWindow.isExpired(timestampNs)) {
+                    rotate(activeWindow);
+                }
+                JavaSpinLockUtils.unlock(this, LOCK_OFFSET);
+            }
+            return false;
+        }
+
+        return activeWindow.sample();
```
roberttoyonaga, line 82: hi @christianhaeubl this reverts the read-write lock changes. Doesn't this still have the same problems you were concerned about (https://github.com/oracle/graal/pull/6899#discussion_r1330104194).

A window may expire after a thread has checked expiry status but before it has sampled with respect to it.

roberttoyonaga:
Thank you for working on integrating this. Apart from my comment above, the sampling logic and tests all seem fine to me.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/throttling/JfrAdaptiveSampler.java
christianhaeubl: Yes, I removed the read/write lock because I don't think that this race can cause any relevant problems (that is probably also the reason why HotSpot doesn't guard against it).

christianhaeubl:
Thanks for double-checking.

roberttoyonaga (approved):

PR #8349, opened 2024-02-09
[GR-51851] Add initial support for native memory tracking.

graalvmbot:


roberttoyonaga:
> please review the core parts of the PR

Thank you @christianhaeubl . I'll look over this. 

> Regarding the NMT categories: is there any need to stick closely to the categories that HotSpot has?

I do not think there is any need to conform to exactly the same categories. If we need new ones, it should be fine to add them.  If there are categories that do not apply to SVM, we can ignore them. I do think it's worth retaining any Hotspot categories that apply in SVM to make it simpler for a user that may be relying/accustomed to them.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/memory/UntrackedNullableNativeMemory.java
```diff
@@ -0,0 +1,135 @@
...
+ * Manages native memory. This class explicitly does <b>NOT</b> support native memory tracking
+ * (NMT). It can therefore be used for the following use cases:
+ * <ul>
+ * <li>Allocate native memory that is later freed outside of Native Image code (e.g., in a C
+ * library).</li>
+ * <li>Free native memory that was allocated outside of Native Image code (e.g., in a C
+ * library).</li>
+ * </ul>
+ */
+public class UntrackedNullableNativeMemory {
```
roberttoyonaga, line 49: Why have `UntrackedNullableNativeMemory` wrap `UnmanagedMemorySupport` if they serve a similar purpose? Is it a concern that a caller could either call `UnmanagedMemorySupport` or `UntrackedNullableNativeMemory` directly to accomplish the same thing? They could even use `LibC` directly, if they are determined.

roberttoyonaga:
> Unfortunately, we can't add native memory tracking to UnmanagedMemory because it is public API.

Now that NMT hooks are added at a level higher than `UnmanagedMemory`, this means we need to worry about  people breaking convention and using `UnmanagedMemorySupport` to create mismatched allocation/frees with `NativeMemory` (not just `LibC`). 

Just an idea: What if we kept the NMT code at the level of `UnmanagedMemorySupport` and kept `UnmanagedMemory` unaltered. Then if callers use`UnmanagedMemory`, we  can assign a header and aggregate into a "default" category. By doing this, we reduce the the number of untracked mallocs and are less likely to have mismatched allocation/frees.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/nmt/NativeMemoryTrackingTests.java
```diff
@@ -0,0 +1,90 @@
...
+public class NativeMemoryTrackingTests {
+    private static final int ALLOCATION_SIZE = 1024 * 16;
+    private static final int RESERVE_SIZE = ALLOCATION_SIZE;
+    private static final int REALLOC_SIZE = ALLOCATION_SIZE / 2;
+
+    @Test
+    public void testMalloc() {
+        assertEquals(0, getUsedMemory());
+
+        Pointer ptr = NativeMemory.malloc(WordFactory.unsigned(ALLOCATION_SIZE), NmtCategory.Code);
```
roberttoyonaga, line 49: Is it impossible to be using the`NmtCategory.Code` category concurrently with the test?

christianhaeubl: Yes, as there is no JIT compilation happening.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/memory/UntrackedNullableNativeMemory.java
christianhaeubl: `UnmanagedMemory` and `UnmanagedMemorySupport` shouldn't be used internally. By not using it internally, it makes it easier to distinguish between code that bypasses NMT on purpose and code that accidentally uses the public API. Note that `UnmanagedMemorySupport` is platform-specific and only some implementations delegate to `LibC`.

christianhaeubl:
> Just an idea: What if we kept the NMT code at the level of `UnmanagedMemorySupport` and kept `UnmanagedMemory` unaltered. Then if callers use `UnmanagedMemory`, we can assign a header and aggregate into a "default" category. By doing this, we reduce the the number of untracked mallocs and are less likely to have mismatched allocation/frees.

We can't use NMT for our public API classes (`UnmanagedMemorySupport` and `UnmanagedMemory`) because we don't know how people use that API. At the moment, it is for example possible to allocate memory via `UnmanagedMemory` and to free that memory in C code. `UnmanagedMemory` can also be used to free memory that was allocated by C code. We would break that behavior if we added NMT to those classes.

roberttoyonaga (approved):
Thank you for working on integrating this @christianhaeubl

PR #8359, opened 2024-02-12
[GR-52034] Update IGV to netbeans 20 and JDK 21

graalvmbot:
Update the base of IGV to netbeans 20 and support JDK11 through JDK21.


visualizer/IdealGraphVisualizer/nbproject/platform.properties
```diff
@@ -207,23 +209,23 @@ disabled.modules=\
...
     org.netbeans.modules.localhistory,\
     org.netbeans.modules.localtasks,\
-    org.netbeans.modules.libs.gradle,\
     org.netbeans.modules.lsp.client,\
     org.netbeans.modules.markdown,\
     org.netbeans.modules.maven.checkstyle,\
     org.netbeans.modules.maven.coverage,\
     org.netbeans.modules.maven.grammar,\
     org.netbeans.modules.maven.graph,\
-    org.netbeans.modules.maven.htmlui,\
```
JaroslavTulach, line 226: Why are you enabling this module?

tkrodriguez: Which module do you mean?  I got complaints in the libraries panel and tried to make them go away.  I would have sworn I disabled all the gradle stuff not enabled more.  I take a look.  I don't think it's related to the problem I'm having though.

tkrodriguez: Sorry my brain just caught up with the fact that this is an exclusion list so `-` means adding.  I didn't explicitly add org.netbeans.modules.maven.htmlui so it must have been auto enabled.  I'll do another pass over this.

JaroslavTulach: I tried to run the tests and yes, there is a problem with `net.java.html` & co. dependencies. I've never seen it myself. I don't know what is wrong off-hand. Debugging is needed - I'd ask some NetBeans experts like Sváťa for help.

PR #8290, opened 2024-02-01
[GR-52025] Use update center for NetBeans 14 in IGV

JaroslavTulach:
Looks like IGV has been [updated to use NetBeans-14](https://github.com/oracle/graal/blob/f105a80046c34cc69b35c1e45227e119cfa2ee80/visualizer/IdealGraphVisualizer/nbproject/platform.properties#L343), however the system is still configured to use _update centers_ from NetBeans 12.5 - that's unfortunate as the latest versions of plugins aren't available. This PR aligns the IGV's base NetBeans version with its update centers.

CCing @tkrodriguez, @sdedic, @dbalek, @Ondrej-Douda

PR #8405, opened 2024-02-19
Less indented (but full) stacktrace in IGV

JaroslavTulach:
Truffle stack traces are huge and long. I've noticed the current IGV implementation truncates them at depth 50 cutting out the most interesting parts. While investigating how to fix that I realized the problem is in indentation. While the indentation of the stack frames is nice, it is far less important than **ability to see the whole stack**.

Less Indentation

This PR inserts `org.openide.nodes.Node` instances for all frames in, but **indents only last 24**. As a result all the frames are available and the most important ones are nicely indented. CCing @sdedic, @tkrodriguez

### DIFF
--- a/visualizer/IdealGraphVisualizer/SourceRepository/src/org/graalvm/visualizer/source/impl/ui/StackViewTopComponent.java
+++ b/visualizer/IdealGraphVisualizer/SourceRepository/src/org/graalvm/visualizer/source/impl/ui/StackViewTopComponent.java
@@ -642,7 +642,7 @@ public final class StackViewTopComponent extends TopComponent
                     return s;
                 }
             };
-            node.setDisplayName(provider.getProperties().getString(PROPNAME_NAME, "")); // NOI18N 
+            node.setDisplayName(provider.getProperties().getString(PROPNAME_NAME, "")); // NOI18N
             setActivatedNodes(new Node[]{node});
         }
     }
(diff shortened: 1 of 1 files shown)


visualizer/IdealGraphVisualizer/SourceRepository/src/org/graalvm/visualizer/source/impl/ui/StackViewTopComponent.java
```diff
@@ -702,10 +702,17 @@ private void updateComboContents(NodeStack locs) {
                 }
             }
         }
-        NodeStack.Frame head = locs.bottom();
-        ChainNode node = new ChainNode(head, head.getLookup().lookup(Node.class),
-                                       head.getNested() == null ? Children.LEAF : new ChainChildren(head));
+        final int indentLevel = 24;
```
JaroslavTulach, line 705: This constant can change, but no more than to [50 as limited here](https://github.com/oracle/graal/blob/f2b350f312f95ff366c149c05950ef19de72e14e/visualizer/IdealGraphVisualizer/SourceRepository/src/org/graalvm/visualizer/source/impl/ui/StackViewTopComponent.java#L149).

JaroslavTulach:
@tkrodriguez can this go in? Working with deep (Enso interpreter) stacktraces is really hard when the stack is cut at 50 (uninteresting) frames!

tkrodriguez:
Why indent them at all?  It certainly doesn't help with readability I think and indenting some and not all seems mildly confusing.

JaroslavTulach:
> Why indent them at all?

I think you have access to the history of the code prior open sourcing, so you can find out who implemented that. Maybe @sdedic knows. My (non verified) hypothesis is: _by indenting it looks more similar to VisualVM_:

VisualVM indentation

> It certainly doesn't help with readability 

It helps me realize where is the root and where the top most frame is, but I am not the right person to discuss esthetics. 

> ... indenting some and not all seems mildly confusing.

Please integrate whatever modification of my patch you want - even [setting the indentLevel](https://github.com/oracle/graal/pull/8405#discussion_r1494872708) to zero or simplifying the code to have no indentation at all.

In any case _not seeing the whole stack is a bug_ and while I can rely on my own build of IGV where the indentation is limited, I have colleagues who would like to use released version of IGV (just yesterday @Akirathan contacted me with IGV issues related to #8290). Please let us not be stopped by considering esthetics too much. I can live with whatever implementation, if it shows all the frames.

sdedic (approved):

JaroslavTulach:
@tkrodriguez, do you think you can merge this PR? If you want to disable indentation, feel free to change the [indentLevel constant](https://github.com/oracle/graal/pull/8405#discussion_r1494872708) to zero. Thank you.

tkrodriguez:
Sorry, this was merged as cb98bbf5fef since I cherry picked it into the other PR.  So it's part of 8f54b7699b7c1d444ab5a19b8e2a6661960f02d6

Issue #8401, opened 2024-02-19
TestThrottler#testDistributionHighRate fails transiently


fniephaus:
Hi @roberttoyonaga, could you take a look at this? The new test seems to fail sometimes, even here on GitHub:

```
1) com.oracle.svm.test.jfr.TestThrottler#testDistributionHighRate
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.oracle.svm.test.jfr.TestThrottler.expectNear(TestThrottler.java:311)
	at com.oracle.svm.test.jfr.TestThrottler.testDistribution(TestThrottler.java:306)
	at com.oracle.svm.test.jfr.TestThrottler.testDistributionHighRate(TestThrottler.java:236)
```

Example build: https://github.com/oracle/graal/actions/runs/7958635035/job/21723904758?pr=8400#step:14:1966

roberttoyonaga:
Hi @fniephaus! I looked into the problem and made a PR here https://github.com/oracle/graal/pull/8412

fniephaus:
Fixed by https://github.com/oracle/graal/pull/8412

PR #8412, opened 2024-02-20
[GR-52295] JFR Throttler Sampling Distribution Test Update

roberttoyonaga:
This PR addresses GH issue: https://github.com/oracle/graal/issues/8401

The test case [`testDistributionHighRate`](https://github.com/oracle/graal/blob/3a131e2f5779742c793d1711cbc04837252ac9d3/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestThrottler.java#L233) fails intermittently due to under-sampling below the allowed error tolerance (2%). This is because the `JfrAdaptiveSampler` [incorporates an element of randomness into the sampling algorithm](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/throttling/JfrAdaptiveSampler.java#L170). All the `testDistribution*` test cases suffer from the same problem because they rely on tolerance values to check the sampling distribution which has some randomness to it. In the corresponding tests in Hotspot (which are actually disabled), it seems like these tolerance values have been picked arbitrarily/empirically, not derived from anything in particular. If, I remove the randomness from the sampler, the test behaves deterministically and always passes. I don't think it makes much sense to remove the randomness here since the purpose of the distribution tests is to test the random sampling. 

This PR bumps up the tolerance for the failing test to 5%. Based on what I have observed, the results are never off by more than 3% (in about 300 runs). This may not be the desired solution since there is still a non-zero chance the RNG causes the test result to be beyond the tolerance. Another solution is to remove the `testDistribution*` test cases from the set throttler tests entirely, [as is done in OpenJDK](https://github.com/openjdk/jdk/blob/b419e9517361ed9d28f8ab2f5beacf5adfe3db91/test/hotspot/gtest/jfr/test_adaptiveSampler.cpp#L243).  If we want to omit those tests, I'll update this PR to do that instead.

### DIFF
--- a/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestThrottler.java
+++ b/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestThrottler.java
@@ -233,7 +233,7 @@ public class TestThrottler extends JfrRecordingTest {
     public void testDistributionHighRate() {
         int maxPopPerWindow = 2000;
         int expectedSamplesPerWindow = 50;
-        testDistribution(() -> maxPopPerWindow, expectedSamplesPerWindow, 0.02);
+        testDistribution(() -> maxPopPerWindow, expectedSamplesPerWindow, 0.05);
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


fniephaus (approved):
LGTM, thanks!

roberttoyonaga:
There's a gate error, but it doesn't seem related to the change in this PR. 
```
There was 1 failure:
1) jdk.graal.compiler.truffle.test.BytecodeOSRNodeTest#testOSRPolling
```

fniephaus:
Nope, that's another transient 🙁

PR #8424, opened 2024-02-21
[GR-52170] Document `-O`, `-march`, and build artifacts.

graalvmbot:
This PR adds documentation for the `-O` and `-march` options as well as for Native Image build artifacts. It also contains minor consistency improvements and a bug fix for `-march=list`.


czelabueno:
@fniephaus Finally, it will be helpful to have a good understanding of the optimization options with the documentation.

Is there a separate option to optimize the executables/files size that can be documented in [OptimizationsAndPerformance.md](docs/reference-manual/native-image/OptimizationsAndPerformance.md) ? Or is using UPX the best option so far?

fniephaus:
> Is there a separate option to optimize the executables/files size that can be documented in [OptimizationsAndPerformance.md](https://github.com/oracle/graal/pull/docs/reference-manual/native-image/OptimizationsAndPerformance.md) ? Or is using UPX the best option so far?

We have plans to add `-Os` to enable user to optimize for size, but that will still take some time. UPX is not really the best option, but currently often better than just zipping executables. The problem with UPX is that it adds additional startup time, which can be significant, and it can increase peak RSS / memory usage. If you don't care about these two metrics, UPX is probably ok.

czelabueno:
> The problem with UPX is that it adds additional startup time, which can be significant, and it can increase peak RSS / memory usage

Exactly, increase peak RSS and add startup latency can be counterproductive but I understand that an out-of-box option will soon be released. Those are good news.
Thanks!

PR #8430, opened 2024-02-22
[GR-44559] Finish ThreadMXBean implementation for Native Image.

graalvmbot:


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/locks/Target_java_util_concurrent_locks_AbstractOwnableSynchronizer.java
```diff
@@ -0,0 +1,78 @@
...
+     * Sets the thread that currently owns exclusive access. A {@code null} argument indicates that
+     * no thread owns access. This method does not otherwise impose any synchronization or
+     * {@code volatile} field accesses.
+     * 
+     * @param thread the owner thread
+     */
+    @Substitute()
+    protected void setExclusiveOwnerThread(Thread thread) {
+        if (thread == null && exclusiveOwnerThread != null) {
+            JavaThreads.JMXMonitoring.removeThreadLock(exclusiveOwnerThread,
```
christianhaeubl, line 59: How does HotSpot track contentions for ReentrantLocks?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
```diff
@@ -0,0 +1,158 @@
...
+        private static int getThreadState(Thread thread, boolean inNative) {
+            int state = PlatformThreads.getThreadStatus(thread);
+            if (inNative) {
+                // set the JMM thread state native flag to true:
+                state |= 0x00400000;
+            }
+            return state;
+        }
+
+        public static ThreadInfo getThreadInfo(Thread thread, int maxDepth,
```
christianhaeubl, line 95: What should this method return if the thread is no longer
a. alive
b. blocked

I think a lot of corner cases are missing here.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ThreadMXUtils.java
```diff
@@ -0,0 +1,158 @@
...
+                        boolean withLockedMonitors, boolean withLockedSynchronizers) {
+            Blocker blocker = getBlockerInfo(thread);
+            StackTraceElement[] stackTrace = getStackTrace(thread, maxDepth);
+            LockedMonitors lockedMonitors = getLockedMonitors(thread, stackTrace.length);
+            boolean inNative = stackTrace.length > 0 && stackTrace[0].isNativeMethod();
```
christianhaeubl, line 100: This is not correct - you would need to access the VM-internal thread state of the isolate thread but this isn't accessible easily for threads other than the current thread.

christianhaeubl (changes requested):
Thanks for the PR, I added a few comments and questions.

I did not do a detailed review yet but I had a brief look at the HotSpot code for the same functionality and it uses a completely different approach (thread info is dumped during a VM operation). I think we will need something similar to avoid a large performance impact.

smthelusive:
Hi @christianhaeubl, thank you for the review.
Given what you said that the whole approach should be different, how should I proceed with this PR? Does it make sense to continue polishing this solution?

Issue #8482, opened 2024-02-29
[GR-52454] Native Image shutdown hooks don't run upon SIGINT


roberttoyonaga:
In order for the SIGINT handler to be registered, [`--install-exit-handlers` must be passed at build time](https://github.com/oracle/graal/blob/2334a134d6e69e379bbdae3d5c39894c9c2947b7/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateExitHandlerFeature.java#L37). This means that stopping a running image with SIGINT doesn't result in shutdown hooks being run.  This affects features such as JFR because a JFR recording will not be dumped when the program exits. This is also a problem for NMT since no report will be printed. This is different than the behavior in regular Java mode. 

Is there a reason that the `--install-exit-handlers` is by default false? If not, should we make it default to true?

roberttoyonaga:
Perhaps one reason `--install-exit-handlers` is false by default is to avoid adding size to small images. One solution is to, keep the default behavior the same but always enable exit handlers if features such as JFR/NMT are included in the image build. Such features already add a significant amount of size so the overhead of `--install-exit-handlers` should be relatively small.

roberttoyonaga:
I've added a very small change in this PR: https://github.com/oracle/graal/pull/8498 to always add exit handlers if features that need them (JFR, NMT) are included in the image build.

PR #8498, opened 2024-03-01
[GR-52454] Include signal exit handlers in the image build if JFR or NMT is also included

roberttoyonaga:
Currently, Native Image shutdown hooks don't run upon SIGINT unless the hosted option`--install-exit-handlers` is set to true at build time. This is a problem for JFR and NMT that do important work such as dumping stats/recordings on exit.

See issue: https://github.com/oracle/graal/issues/8482

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateExitHandlerFeature.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateExitHandlerFeature.java
@@ -34,7 +34,7 @@ import com.oracle.svm.core.feature.AutomaticallyRegisteredFeature;
 public class SubstrateExitHandlerFeature implements InternalFeature {
     @Override
     public void beforeAnalysis(BeforeAnalysisAccess access) {
-        if (SubstrateOptions.InstallExitHandlers.getValue()) {
+        if (SubstrateOptions.InstallExitHandlers.getValue() || VMInspectionOptions.hasJfrSupport() || VMInspectionOptions.hasNativeMemoryTrackingSupport()) {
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateExitHandlerFeature.java
```diff
@@ -34,7 +34,7 @@
 public class SubstrateExitHandlerFeature implements InternalFeature {
     @Override
     public void beforeAnalysis(BeforeAnalysisAccess access) {
-        if (SubstrateOptions.InstallExitHandlers.getValue()) {
+        if (SubstrateOptions.InstallExitHandlers.getValue() || VMInspectionOptions.hasJfrSupport() || VMInspectionOptions.hasNativeMemoryTrackingSupport()) {
```
fniephaus, line 37: Typically, we let options turn on/off other options, so here `EnableMonitoringFeatures` would turn on `InstallExitHandlers` if JFR or NMT is turned on. This is arguably easier, but let me take another look.

roberttoyonaga: Should I adopt your suggestion and use `OptionKey.onValueUpdate()` to turn on `InstallExitHandlers`? Or should I leave it as it currently is?

fniephaus: I'm currently trying to find out whether we could/should enable `InstallExitHandlers` by default. In any case, I can make the change if any is need.

PR #8507, opened 2024-03-04
[WIP] [GR-45171] Basic JVMTI infrastructure.

graalvmbot:
This PR adds basic JVM Tool Interface (JVM TI) support to Native Image so that JVM TI agents can be used at run-time:
- If JVMTI should be used at run-time, it needs to be be enabled during the image build with `-H:+JVMTI`.
- When running a native executable that was built with JVM TI support, an agent can then be attached to the native executable with `-XX:JVMTIAgentLib=...` or `-XX:JVMTIAgentPath=...` (similar to `-agentlib:...` and `-agentpath:...` on HotSpot).

Restrictions:
- only a single agent is supported at the moment
- only the following JVM TI functions are supported at the moment
  - `Allocate`
  - `Deallocate`
  - `ForceGarbageCollection`
  - `SetEventCallbacks`
  - `SetEventNotificationMode`
  - `GetPotentialCapabilities`
  - `GetCapabilities`
  - `AddCapabilities`
  - `RelinquishCapabilities`
  - `GetTime`
  - `GetPhase` (only returns `JVMTI_PHASE_LIVE` and `JVMTI_PHASE_DEAD`)
  - `DisposeEnvironment`
  - `GetVersionNumber`
  - `GetErrorName`
- only the following JVM TI events are supported at the moment
  - `JVMTI_EVENT_VM_INIT`
  - `JVMTI_EVENT_VM_START`
  - `JVMTI_EVENT_VM_DEATH`
- no JVM TI capabilities are supported at the moment

Open for discussions:
- At the moment, Native Image calls `Agent_OnLoad` after VM startup (i.e., some Java code may already have been executed at this point). This does not really match with the JVM TI specification. The alternative would be to call `Agent_OnAttach` instead.
- Native Image triggers the events `JVMTI_EVENT_VM_INIT` and `JVMTI_EVENT_VM_START` directly after invoking `Agent_OnLoad`.

Open issues:
- The JNI implementation in Native Image allocates Java heap memory at the moment (especially for object handles). This needs to be changed before more complex functionality can be implemented (i.e., native memory needs to be used instead). Otherwise, it is impossible to implement certain JVM TI events (e.g., `JVMTI_EVENT_RESOURCE_EXHAUSTED`, `JVMTI_EVENT_GARBAGE_COLLECTION_*`).


christianhaeubl:
Closed in favor of https://github.com/oracle/graal/pull/9558

Issue #8486, opened 2024-03-01
[jfr] jdk.PhysicalMemory event doesn't set 'usedSize'


jerboaa:
**Describe the issue**
When enabling a native image with `--enable-monitoring=jfr` and then run the resulting image with JFR on, the emitted `jdk.PhysicalMemory` event unconditionally sets the value of `usedSize` to `0`. This should be fixed. Aside: if the value isn't available it should set it to `-1` to make it clear it's not supported or not possible. `0` might be a real value.

**Steps to reproduce the issue**
```
$ cat CheckOperatingSystemMXBean.java 
import com.sun.management.OperatingSystemMXBean;
import java.lang.management.ManagementFactory;

public class CheckOperatingSystemMXBean {

    public static void main(String[] args) throws Exception {
        OperatingSystemMXBean osBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
        System.out.println(String.format("Runtime.availableProcessors: %d", Runtime.getRuntime().availableProcessors()));
        System.out.println(String.format("OperatingSystemMXBean.getAvailableProcessors: %d", osBean.getAvailableProcessors()));
        System.out.println(String.format("OperatingSystemMXBean.getTotalPhysicalMemorySize: %d", osBean.getTotalPhysicalMemorySize()));
	int count = 0;
	while (count < 10) {
		Thread.sleep(1000);
		count++;
	}
    }

}
$ javac CheckOperatingSystemMXBean.java
$ native-image --enable-monitoring=jfr CheckOperatingSystemMXBean
$ ./checkoperatingsystemmxbean -XX:+FlightRecorder -XX:StartFlightRecording=filename=flight-native.jfr -XX:FlightRecorderLogging=jfr
[info][jfr] Added periodic task for EveryChunkPeriodEvents(11884)
[info][jfr] Added periodic task for EndChunkPeriodEvents(11903)
[info][jfr] Flight Recorder initialized
[info][jfr] Repository base directory: /tmp
[warn][jfr,setting] @Deprecated JFR events, and leak profiling are not yet supported.
[warn][jfr,setting] @Deprecated JFR events, and leak profiling are not yet supported.
[info][jfr        ] Started recording "1" (1) {dumponexit=true, filename=[...]/flight-native.jfr}
Runtime.availableProcessors: 12
OperatingSystemMXBean.getAvailableProcessors: 12
OperatingSystemMXBean.getTotalPhysicalMemorySize: 67167309824
[...]

PR #8513, opened 2024-03-04
[GR-52484] Partial support for physical memory usage (fixes JFR bug) and JFR tests for periodic native events

roberttoyonaga:
### Summary

- Add partial support for physical memory usage (if container usage is enabled and linux). 
- Fix bug when reporting physical memory usage in `jdk.PhysicalMemory` JFR event. It will now be set to a valid value or -1 (instead of 0).
- Add tests for `jdk.PhysicalMemory` and other periodic native JFR events.

Note: Eventually, it may be worth adding support in for the case `Containers.isContainerized()` is false. This will require different implementations for different OS. In hotspot, the implmentation for Linux attempts to read `/proc/meminfo `, and if that fails, it tries to read the struct returned from the sysinfo() call.

Related issue: https://github.com/oracle/graal/issues/8486

```
$ mx native-image --install-exit-handlers --enable-monitoring=jfr -m jdk.httpserver

$ ./jdk.httpserver -XX:StartFlightRecording=filename=rec.jfr 

$ $JAVA_HOME/bin/jfr print --events jdk.PhysicalMemory rec.jfr 
jdk.PhysicalMemory {
  startTime = 13:14:55.953 (2024-03-04)
  totalSize = 62.4 GB
  usedSize = 35.3 GB
}

jdk.PhysicalMemory {
  startTime = 13:14:57.352 (2024-03-04)
  totalSize = 62.4 GB
  usedSize = 35.3 GB
}


```

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heap/PhysicalMemory.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heap/PhysicalMemory.java
@@ -24,9 +24,15 @@
  */
 package com.oracle.svm.core.heap;
 
+import java.io.IOException;
+import java.lang.management.ManagementFactory;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
 import java.util.concurrent.locks.ReentrantLock;
 
 import org.graalvm.nativeimage.ImageSingletons;
+import org.graalvm.nativeimage.Platform;
 import org.graalvm.word.UnsignedWord;
 import org.graalvm.word.WordFactory;
 
@@ -39,6 +45,8 @@ import com.oracle.svm.core.thread.VMOperation;
 import com.oracle.svm.core.util.UnsignedUtils;
 import com.oracle.svm.core.util.VMError;
 
+import com.sun.management.OperatingSystemMXBean;
+
 /**
  * Contains static methods to get configuration of physical memory.
  */
@@ -49,6 +57,8 @@ public class PhysicalMemory {
         /** Get the size of physical memory from the OS. */
         UnsignedWord size();
     }
+
+    private static final long K = 1024;
 
     private static final ReentrantLock LOCK = new ReentrantLock();
     private static final UnsignedWord UNSET_SENTINEL = UnsignedUtils.MAX_VALUE;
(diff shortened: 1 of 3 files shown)


jerboaa:
OK. Got it. We could read `/proc/meminfo`'s `MemAvailable` on Linux in addition to the value from the OS bean and take the minimum. I take it the OS bean would be OK on Windows and Mac. That would cover most cases?

roberttoyonaga:
> OK. Got it. We could read `/proc/meminfo`'s `MemAvailable` on Linux in addition to the value from the OS bean and take the minimum.

If reading `/proc/meminfo`'s `MemAvailable` on Linux is unsuccessful, Hotspot tries to use the struct returned by `sysinfo()` instead. 

So the flow might instead be:
If containerized, return the container metrics. Otherwise try `/proc/meminfo`and compare with `OperatingSystemMXBean.getFreeMemorySize()` . If unsuccessful, try `sysinfo()` and compare with `OperatingSystemMXBean.getFreeMemorySize()`.

>  I take it the OS bean would be OK on Windows and Mac. That would cover most cases?

hmmm I'm not sure about that. Hotspot does some alternative special logic to cover windows. But maybe its because its in VM-level code and doesn't have access to JMX. I also don't have a windows/mac machine to test this on.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heap/PhysicalMemory.java
```diff
@@ -107,6 +112,32 @@ public static UnsignedWord size() {
         return cachedSize;
     }
 
+    /**
+     * Returns the amount of used physical memory in bytes, or -1 if not supported yet.
+     */
+    public static long usedSize() {
+        if (Containers.isContainerized()) {
+            return Containers.memoryUsageInBytes();
+        }
```
jerboaa, line 121: Unfortunately, `Containers.isContainerized()` will return true on most physical Linux machines the way it's implemented now. You'd have to do additional checks: e.g. `Containers.memoryLimitInBytes() > 0` to actually enter the non-containerized branches.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heap/PhysicalMemory.java
```diff
@@ -107,6 +112,32 @@ public static UnsignedWord size() {
...
+                    String memAvailable = line.replaceAll("\\D", "");
+                    if (!memAvailable.isEmpty()) {
+                        return size().rawValue() - Long.parseLong(memAvailable);
+                    }
+                }
+            } catch (IOException e) {
+                return -1;
+            }
+        }
+        return -1;
```
jerboaa, line 138: I believe a reasonable implementation would be to use the OS bean here for other platforms, rather than keep unimplemented. YMMV.

jerboaa:
> > /proc/meminfo's MemAvailable on Linux in addition to the value from the OS bean and take the minimum
> 
> Actually, I don't think we want to always take the minimum. In [my example above](https://github.com/oracle/graal/pull/8513#issuecomment-1979414907) it's better to report "available memory" because "free memory" is deceptively low. It might be better to copy Hotspot and report `/proc/meminfo MemAvailable` if possible, and fall back to reporting "free memory" otherwise.

Makes sense. One simplification could be to use the OS bean when containerized and on other platforms (it should do the right thing). Only for `Containers.isContainerized() == false` use the Linux specific available memory trick.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heap/PhysicalMemory.java
roberttoyonaga: Ok I see. I've adopted your suggestion and I'm now checking `Containers.memoryLimitInBytes()` instead.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heap/PhysicalMemory.java
roberttoyonaga: I've adopted your suggestion and am now using `OperatingSystemMXBean` if not on linux and `UseContainerSupport` is set.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/heap/PhysicalMemory.java
jerboaa: I don't think we need `Containers.memoryUsageInBytes()`. Here is what I think should work for the majority of cases: https://github.com/jerboaa/graal/commit/03b319579abcd26c0761c6f55f7811b71914f392

roberttoyonaga: ok I see what you mean. I think that's a good idea.

jerboaa (approved):
LGTM.

christianhaeubl:
Thanks for the PR, please squash your commits.

roberttoyonaga:
@christianhaeubl ok the commits have been squashed now.

christianhaeubl (approved):
Thanks!

PR #8539, opened 2024-03-07
Omit unnecessary AMD64 TEST instruction

petermz:
This is inspired by similar [change](https://bugs.openjdk.org/browse/JDK-8312213) in JDK. When a TEST instruction immediately follows one of AND/OR/XOR, it may be left out because the necessary flags (SF, ZF) are already set.

This patch might probably be structured better, and extended to support more cases. But first I'd just like to check if the approach taken is sane and overall makes sense, so I'd appreciate comments and criticisms.


fniephaus:
Thanks for the PR, @petermz! Could you please fix the [style issues](https://github.com/oracle/graal/actions/runs/8189897350/job/22395644449?pr=8539#step:14:1142)?

petermz:
Sure, fixed

PR #8542, opened 2024-03-08
Add Java-level JFR event tests and annotate JfrAdaptiveSampler methods with BasedOnJDKFile

roberttoyonaga:
### Summary:

1. **Add tests for Java-level JFR events.**  

These events we typically get "for free", but problems related to substitutions can result in them being emitted incorrectly, or not at all. 

4. **Add @BasedOnJdkFile to throttler classes.** 
 
I'm not sure if its better to annotate each individual method, or simply the whole class. Will this annotation somehow alert us if the relevant code in the JDK changes, or is it just a form of documentation?

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/throttling/JfrEventThrottler.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/throttling/JfrEventThrottler.java
@@ -27,6 +27,7 @@ package com.oracle.svm.core.jfr.throttling;
 
 import static com.oracle.svm.core.Uninterruptible.CALLED_FROM_UNINTERRUPTIBLE_CODE;
 
+import com.oracle.svm.core.util.BasedOnJDKFile;
 import com.oracle.svm.core.Uninterruptible;
 import com.oracle.svm.core.jfr.JfrTicks;
 import com.oracle.svm.core.thread.JavaSpinLockUtils;
@@ -47,10 +48,9 @@ public class JfrEventThrottler extends JfrAdaptiveSampler {
     private static final long TEN_PER_1000_MS_IN_HOURS = 36000;
     private static final long DAY = 24 * HOUR;
     private static final long TEN_PER_1000_MS_IN_DAYS = 864000;
-
-    private static final long DEFAULT_WINDOW_LOOKBACK_COUNT = 25;
-    private static final long LOW_RATE_UPPER_BOUND = 9;
-    private static final long WINDOW_DIVISOR = 5;
+    @BasedOnJDKFile("https://github.com/openjdk/jdk/blob/jdk-23+8/src/hotspot/share/jfr/recorder/service/jfrEventThrottler.cpp#L105-L106") private static final long DEFAULT_WINDOW_LOOKBACK_COUNT = 25;
+    @BasedOnJDKFile("https://github.com/openjdk/jdk/blob/jdk-23+8/src/hotspot/share/jfr/recorder/service/jfrEventThrottler.cpp#L112-L113") private static final long LOW_RATE_UPPER_BOUND = 9;
+    @BasedOnJDKFile("https://github.com/openjdk/jdk/blob/jdk-23+8/src/hotspot/share/jfr/recorder/service/jfrEventThrottler.cpp#L113-L114") private static final long WINDOW_DIVISOR = 5;
 
     private static final JfrSamplerParams DISABLED_PARAMS = new JfrSamplerParams();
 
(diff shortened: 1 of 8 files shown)


zapster:
> I'm not sure if its better to annotate each individual method, or simply the whole class.

I guess it depends. Individual methods give us higher granularity, so if only one method changes this might be simpler to adopt. On the other hand, if all of the cpp file in the JDK is just related to the Java file in native image, a single one might be fine as well.

One thing we should keep in mind when adding these annotations: It should be clear what needs to be done if the referenced JDK code snippet changes. If the is not clear from the native image code alone, please add further comments.

> Will this annotation somehow alert us if the relevant code in the JDK changes, or is it just a form of documentation?

There is no public utility yet that uses this information, but we are reviewing those annotation when updating to a new JDK version and take care that they stay up to date.

That said, this is a pretty new approach and we are still learning what works and what not. So with that in mind, I'd deal use this as computer readable documentation for now.

In that sense, your change is very much OK wrt to `BasedOnJDKFile`, thanks. Very much appreciated!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/throttling/JfrEventThrottler.java
```diff
@@ -47,10 +48,9 @@ public class JfrEventThrottler extends JfrAdaptiveSampler {
     private static final long TEN_PER_1000_MS_IN_HOURS = 36000;
     private static final long DAY = 24 * HOUR;
     private static final long TEN_PER_1000_MS_IN_DAYS = 864000;
-
-    private static final long DEFAULT_WINDOW_LOOKBACK_COUNT = 25;
-    private static final long LOW_RATE_UPPER_BOUND = 9;
-    private static final long WINDOW_DIVISOR = 5;
+    @BasedOnJDKFile("https://github.com/openjdk/jdk/blob/jdk-23+8/src/hotspot/share/jfr/recorder/service/jfrEventThrottler.cpp#L105-L106") private static final long DEFAULT_WINDOW_LOOKBACK_COUNT = 25;
```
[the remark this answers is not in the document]
roberttoyonaga: >  for some reason github decided to use my private email,

Yeah I saw that too, so I reset that commit and force pushed a new one. 

>i meant to do that for all static final fields.

Ok that makes sense. I added another commit.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/throttling/JfrEventThrottler.java
```diff
@@ -47,9 +48,11 @@ public class JfrEventThrottler extends JfrAdaptiveSampler {
     private static final long TEN_PER_1000_MS_IN_HOURS = 36000;
     private static final long DAY = 24 * HOUR;
     private static final long TEN_PER_1000_MS_IN_DAYS = 864000;
-
+    @BasedOnJDKFile("https://github.com/openjdk/jdk/blob/jdk-23+8/src/hotspot/share/jfr/recorder/service/jfrEventThrottler.cpp#L105-L106") //
     private static final long DEFAULT_WINDOW_LOOKBACK_COUNT = 25;
+    @BasedOnJDKFile("https://github.com/openjdk/jdk/blob/jdk-23+8/src/hotspot/share/jfr/recorder/service/jfrEventThrottler.cpp#L112-L113") //
```
zapster, line 53: This and the other two annotations for the constant above and below should use a single line reference. Currently they span two lines, see https://github.com/openjdk/jdk/blob/jdk-23+8/src/hotspot/share/jfr/recorder/service/jfrEventThrottler.cpp#L112-L113.

```suggestion
    @BasedOnJDKFile("https://github.com/openjdk/jdk/blob/jdk-23+8/src/hotspot/share/jfr/recorder/service/jfrEventThrottler.cpp#L112") //
```

Other annotations that should reference a single source line but use a range are probably affected as well.

roberttoyonaga: Ok I've updated the relevant annotations to point to a single line instead of a range. [I needed up update the regex here too.](https://github.com/oracle/graal/pull/8542/files#diff-8fcc9f49f232b75e70027a8ec6252e86ace325eb4cb9010238885b47bc49295cR76)

zapster: ah right. apparently, we did not yet use single line references. thanks for fixing.

zapster (approved):
looks good. I will integrate it asap.

roberttoyonaga:
Thank you!

PR #8591, opened 2024-03-18
[GR-52680] [GR-52382] Take down some redundant user guides.

graalvmbot:


zakkak:
Given that PGO is not available to GraalVM CE users, what's the alternative for them?

Does this PR imply that there are plans to remove `-H:+DashboardAll`, `-H:+DashboardJson`, and `-H:DashboardDump` as well at some point?

cc @olyagpl @fniephaus

olyagpl:
@zakkak, the `-H:+BuildReport` option remains available in GraalVM CE. We need to write a separate guide for non-PGO use case.

fniephaus:
> Given that PGO is not available to GraalVM CE users, what's the alternative for them?

There are some other options to tune performance like build time initialization as used in Quarkus. Even though PGO is not available in CE, it is part of Oracle GraalVM and can be used for [free in production](https://blogs.oracle.com/java/post/graalvm-free-license).

> Does this PR imply that there are plans to remove -H:+DashboardAll, -H:+DashboardJson, and -H:DashboardDump as well at some point?

The dashboard UI is no longer available and instead, users can leverage the [new build reports](https://www.graalvm.org/jdk22/reference-manual/native-image/optimizations-and-performance/PGO/build-reports/). We do intend to deprecate the dashboard options because the format is neither well specified nor designed. Instead, we want to redesign it and provide a proper JSON schema. This will be part of a larger initiative where we want to unify data exports in Native Image, among other exports such as build-output, build-artifacts, and embedded-resources.

Hope this helps!

zakkak:
> There are some other options to tune performance like build time initialization as used in Quarkus. Even though PGO is not available in CE, it is part of Oracle GraalVM and can be used for [free in production](https://blogs.oracle.com/java/post/graalvm-free-license).
> [...snip...]
> The dashboard UI is no longer available and instead, users can leverage the [new build reports](https://www.graalvm.org/jdk22/reference-manual/native-image/optimizations-and-performance/PGO/build-reports/).

Sorry for not making this more clear. I meant what's the alternative for CE users to get the equivalent of the new build reports, since PGO is not available in CE.

I tried passing `-H:+BuildReport -H:+BuildReportSamplerFlamegraph` to GraalVM for JDK 22 CE and I got:

```
Warning: The option '-H:+BuildReportSamplerFlamegraph' is experimental and must be enabled via '-H:+UnlockExperimentalVMOptions' in the future.
Warning: The option '-H:+BuildReport' is experimental and must be enabled via '-H:+UnlockExperimentalVMOptions' in the future.
Warning: Please re-evaluate whether any experimental option is required, and either remove or unlock it. The build output lists all active experimental options, including where they come from and possible alternatives. If you think an experimental option should be considered as stable, please file an issue.
Error: Could not find option 'BuildReport' from 'user'. Use -H:PrintFlags= to list all available options.
Could not find option 'BuildReportSamplerFlamegraph' from 'user'. Use -H:PrintFlags= to list all available options.
```

> We do intend to deprecate the dashboard options because the format is neither well specified nor designed. Instead, we want to redesign it and provide a proper JSON schema. This will be part of a larger initiative where we want to unify data exports in Native Image, among other exports such as build-output, build-artifacts, and embedded-resources.

This sounds good. Although not well specified or designed as you say they still contain valuable info (see https://quarkus.io/blog/mandrel-23-0-image-size-increase/) and we would like to have an alternative before seeing them go away :)
I will keep an eye out for it.

> Hope this helps!

It does, thanks Fabio!

fniephaus:
> Sorry for not making this more clear. I meant what's the alternative for CE users to get the equivalent of the new build reports, since PGO is not available in CE.

Build reports and PGO are two different features, both are not available in CE. In Oracle GraalVM for JDK 22, we have added an (optional) visualization of PGO data in the reports, which can be turned on with `-H:+BuildReportSamplerFlamegraph`.

Generally, the information displayed in the build reports can be requested in JSON files, for example the first page of a report shows the data from the build-output JSON export. So CE users can get and script the data, while the build reports provide them with useful visualizations of that data.

> This sounds good. Although not well specified or designed as you say they still contain valuable info (see https://quarkus.io/blog/mandrel-23-0-image-size-increase/) and we would like to have an alternative before seeing them go away :)

I'm aware, which is why we haven't deprecated the dashboard options just yet. We're using the build reports to redesign the JSON format for what is included in the dashboard dumps. Once we have a design we are happy with, we share a proposal and add a new option. The goal is to provide this data via a unified data export mechanism through appropriate public API. We just can come up with a good design over night. :wink:

Issue #8629, opened 2024-03-22
Track virtual memory with native memory tracking (NMT)


roberttoyonaga:
**Feature request**

Please include the following information:

**Is your feature request related to a problem? Please describe.**
Currently, NMT only includes malloc tracking. Unlike OpenJDK, virtual memory tracking is missing. 

**Describe the solution you'd like.**
Add virtual memory accounting code to the native memory tracking infrastructure. It should be mostly similar `VirtualMemoryTracker` in Hotspot. Virtual memory tracking should record committed and reserved memory. The records should be grouped by memory category (Java heap, Image heap, etc) similar to the malloc tracking. Virtual memory data should be reported in the same way as malloc information.  

**Describe who do you think will benefit the most.**
Users and developers will benefit from more complete NMT information. 

**Describe alternatives you've considered.**
There isn't a clear alternative way to get insights into committed/reserved memory.  This information is currently not recorded anywhere. 

**Express whether you'd like to help contributing this feature**
I can help with this.

PR #8630, opened 2024-03-22
[GR-56600] Track virtual memory with native memory tracking (NMT) 

roberttoyonaga:
### Summary
Related Issue: https://github.com/oracle/graal/issues/8629

This pull request adds virtual memory tracking to NMT. Previously, NMT only tracked mallocs and was missing information about virtual memory. 

The main component of this PR is the `VirtualMemoryTracker` class which  is based on `virtualMemoryTracker.cpp` in Hotspot. Since commits and uncommits may partially (or not at all) overlap with existing committed regions, it is necessary to keep an internal model of virtual memory in use.  `NativeMemoryTracking` delegates virtual memory tracking duties to `VirtualMemoryTracker` which maintains this internal model.  This is the same as how `memTracker.hpp` delegates to `virtualMemoryTracker.cpp` in Hotspot. `NativeMemoryTracking` remains the the main interface to record and retrieve tracking data, while`VirtualMemoryTracker` and associated classes are only used by NMT internally.

Tracking virtual memory usage before the image heap is mapped requires special handling. We can't use the `NativeMemoryTracking` image singleton yet so we must instead use unmanaged memory and stack memory to temporarily record virtual memory usage. Once the image heap is mapped, the temporary records can be properly committed to tracking. `NmtPreImageHeapData` is used to enqueue records of pre-image-heap reserves/commits. 

**More details:**
- All virtual memory accounting operations are protected by a spinlock. I don't think this should have an impact on performance since changes in virtual memory usage should be infrequent. `memTracker.hpp` uses the same approach to protect virtual memory accounting. 
- Virtual memory peak tracking is also added. 
- Updated NMT report printing to make it more similar to OpenJDK:
```
Native memory tracking

Total
        (reserved=30200KB, committed=29688KB)
        (malloc=32KB, count=77)
        (peak malloc=10518KB, count at peak=528)
        (mmap: reserved=30168KB, committed=29656KB)
        (mmap: peak reserved=30168KB, peak committed=29656KB)
...
JFR
        (reserved=16KB, committed=16KB)
        (malloc=16KB, count=2)
        (peak malloc=10495KB, count at peak=453)
        (mmap: reserved=0KB, committed=0KB)
        (mmap: peak reserved=0KB, peak committed=0KB)
...
```

### DIFF
--- a/substratevm/src/com.oracle.svm.core.foreign/src/com/oracle/svm/core/foreign/TrampolineSet.java
+++ b/substratevm/src/com.oracle.svm.core.foreign/src/com/oracle/svm/core/foreign/TrampolineSet.java
@@ -36,6 +36,7 @@ import org.graalvm.word.PointerBase;
 import org.graalvm.word.UnsignedWord;
 import org.graalvm.word.WordFactory;
 
+import com.oracle.svm.core.nmt.NmtCategory;
 import com.oracle.svm.core.os.VirtualMemoryProvider;
 import com.oracle.svm.core.util.VMError;
 
@@ -111,7 +112,7 @@ final class TrampolineSet {
         VirtualMemoryProvider memoryProvider = VirtualMemoryProvider.get();
         UnsignedWord pageSize = allocationSize();
         /* We request a specific alignment to guarantee correctness of getAllocationBase */
-        Pointer page = memoryProvider.commit(WordFactory.nullPointer(), pageSize, VirtualMemoryProvider.Access.WRITE | VirtualMemoryProvider.Access.FUTURE_EXECUTE);
+        Pointer page = memoryProvider.commit(WordFactory.nullPointer(), pageSize, VirtualMemoryProvider.Access.WRITE | VirtualMemoryProvider.Access.FUTURE_EXECUTE, NmtCategory.Internal);
         if (page.isNull()) {
             throw new OutOfMemoryError("Could not allocate memory for trampolines.");
         }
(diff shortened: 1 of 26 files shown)


roberttoyonaga:
Added a couple commits to provide windows support and improve NMT report printing.

christianhaeubl:
@roberttoyonaga : thanks for the PR, however I think that the complexity of this PR outweighs its benefits at the moment.

We can probably get almost the same functionality with a far simpler change:
- `VirtualMemoryProvider` is primarily used for the Java heap (image heap + runtime heap)
- For the Java heap, the garbage collector knows in detail how much memory is reserved and committed (see `HeapAccounting`). We probably don't expose all the needed information at the moment but it should be rather simple to add.

So, I think that something like the above would be the better approach for now. However, I would still suggest to keep your branch around as things may change in the future.

roberttoyonaga:
Hi @christianhaeubl, ok I'll work on another simplified PR. I'll get rid of the `NmtVirtualMemoryTracker` class and retrieve the committed/reserved statistics from the GC directly. The reason I didn't take that approach originally, is because I wanted to avoid making assumptions based on the current implementation. The more general approach here provides more flexibility, but I agree that it might be unnecessarily complicated for how SVM uses virtual memory currently.

roberttoyonaga:
Hi @christianhaeubl,  I've made a PR for the simplified version here: https://github.com/oracle/graal/pull/9505

christianhaeubl:
I am closing this PR in favor of https://github.com/oracle/graal/pull/9505.

PR #8668, opened 2024-03-29
[GR-52942] Add JFR events and track peak usage for native memory tracking.

graalvmbot:


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/NativeMemoryUsagePeakEvent.java
```diff
@@ -0,0 +1,47 @@
...
+import jdk.jfr.StackTrace;
+
+/** Similar to the JFR event jdk.NativeMemoryUsage, except that it tracks the peak usage. */
+@Experimental
+@Name("jdk.NativeMemoryUsagePeak")
+@Label("Native Memory Usage Peak")
+@Description("Native memory peak usage for a given memory type in the JVM.")
+@Category({"Java Virtual Machine", "Memory"})
+@StackTrace(false)
+public class NativeMemoryUsagePeakEvent extends Event {
```
roberttoyonaga, line 43: All of the other events with the "jdk" prefix have been ported over from OpenJDK. Maybe there should be some way of differentiating these NMT events. If we want to continue using the "jdk" prefix, do you think it would be a good idea to include a note in the event description indicating these events are built-in specific to Native Image?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/NativeMemoryUsageTotalPeakEvent.java
```diff
@@ -28,14 +28,17 @@
...
 import jdk.jfr.Label;
 import jdk.jfr.Name;
 import jdk.jfr.StackTrace;
 
-@Name("svm.NativeMemoryUsageTotalPeak")
+/** Similar to the JFR event jdk.NativeMemoryUsageTotal, except that it tracks the peak usage. */
+@Experimental
+@Name("jdk.NativeMemoryUsageTotalPeak")
 @Label("Native Memory Usage Total Peak")
-@Description("Information about native memory peak usage of committed virtual memory and malloc.")
```
roberttoyonaga: I initially specified that the peak usage is reported with regard to committed virtual memory + mallocs, since we report both committed and reserved memory in the other NMT events. Do you think it's not necessary? I'm fine leaving out those details if it's already obvious.

roberttoyonaga:
Thank you for working on integrating these changes @christianhaeubl ! 
Mostly everything looks good to me. I just left a few small questions above.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/NativeMemoryUsagePeakEvent.java
fniephaus: We've also marked the new events as experimental for now, that's how I would currently look for them. But yes, a short note like `(GraalVM Native Image only)` or so in the description sounds reasonable to me.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/NativeMemoryUsageTotalPeakEvent.java
christianhaeubl: I think we should report both committed and reserved for the Native Image-specific events as well. I will change the code accordingly.

PR #8753, opened 2024-04-11
[GR-46564] Support more than one image CodeInfo and reduce frame info redundancy with global method ids.

graalvmbot:


SergejIsbrecht:
@peter-hofer , 

could you maybe explain what the impact of this change will be? Will it reduce the native application binary size?

peter-hofer:
@SergejIsbrecht , indeed this change can reduce image size by up to 5%.

PR #8805, opened 2024-04-19
[GR-53528] Update labsjdk to 23+21-jvmci-b01

graalvmbot:
Update labsjdk to 23+21-jvmci-b01

* adopt https://github.com/openjdk/jdk/pull/18688
* adopt https://github.com/openjdk/jdk/pull/16208
  * need to handle new warnings in our code base
* adopt https://github.com/openjdk/jdk/pull/18527


jerboaa:
@zapster Hi! Is there an ETA when this merges? We've failing CI due to this for ~2 weeks now.

zapster:
@jerboaa there have been no JDK 23+19 and JDK 23+20 EA builds from https://jdk.java.net/23/, so we could also not update LabsJDK. But now things should be back at the normal weekly frequency. We are already working on JDK 23+22 (#8910).

PR #8836, opened 2024-04-25
Fix potential github action smells

ceddy4395:
Hey! 🙂
I want to contribute the following changes to your workflow:
- Avoid starting new workflow whilst the previous one is still running
- Run tests on multiple OS's
- Stop running workflows when there is a newer commit in branch
- Avoid uploading artifacts on forks 

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,10 +91,11 @@ permissions:
 jobs:
   build-graalvm-linux:
     name: /${{ matrix.env.PRIMARY }} ${{ matrix.env.GATE_TAGS }} JDK${{ matrix.env.JDK_VERSION }}
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-20.04, ubuntu-22.04]
         include:
           # /compiler
           - env:
@@ -225,6 +226,8 @@ jobs:
   build-graalvm-windows:
     name: /substratevm on Windows
     runs-on: windows-2022
+    matrix: 
+      os: [windows-2022, windows-2019]
     env:
       MX_PYTHON: 'python'
     steps:
--- a/.github/workflows/quarkus.yml
+++ b/.github/workflows/quarkus.yml
@@ -51,6 +51,10 @@ on:
   - cron: '0 3 * * *'
   workflow_dispatch:
 
+concurrency:
+  group: ${{github.workflow}} - ${{github.ref}}
+  cancel-in-progress: true
+
 env:
   COMMON_MAVEN_ARGS: "-e -B --settings .github/mvn-settings.xml --fail-at-end"
   DB_NAME: hibernate_orm_test
@@ -180,7 +184,7 @@ jobs:
         run: find . -type d -name '*-reports' -o -wholename '*/build/reports/tests/functionalTest' | tar -czf test-reports.tgz -T -
       - name: Upload failure Archive (if maven failed)
         uses: actions/upload-artifact@v4
-        if: failure()
+        if: failure() && github.repository == 'oracle/graal'
         with:
           name: test-reports-native-${{matrix.category}}
           path: 'test-reports.tgz'
--- a/.github/workflows/micronaut.yml
+++ b/.github/workflows/micronaut.yml
@@ -60,6 +60,10 @@ env:
 permissions:
   contents: read # to fetch code (actions/checkout)
 
+concurrency: 
+  group: ${{github.workflow}}
+  cancel-in-progress: true
+
 jobs:
   build-graalvm-and-micronaut:
     name: Native Tests


.github/workflows/main.yml
```diff
@@ -91,10 +91,11 @@ permissions:
 jobs:
   build-graalvm-linux:
     name: /${{ matrix.env.PRIMARY }} ${{ matrix.env.GATE_TAGS }} JDK${{ matrix.env.JDK_VERSION }}
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-20.04, ubuntu-22.04]
```
fniephaus, line 98: Thanks for suggesting this but considering that build do consume quite some resources (11 jobs, 3h in total, almost 6000 workflow runs), I think it would be better if we keep testing on just one version of Ubuntu here.

.github/workflows/main.yml
```diff
@@ -225,6 +226,8 @@ jobs:
   build-graalvm-windows:
     name: /substratevm on Windows
     runs-on: windows-2022
+    matrix: 
+      os: [windows-2022, windows-2019]
```
fniephaus, line 230: The matrix is not used. Also, Native Image needs Visual Studio 2022 or later, so it won't build on `windows-2019`.

.github/workflows/quarkus.yml
```diff
@@ -51,6 +51,10 @@ on:
   - cron: '0 3 * * *'
   workflow_dispatch:
 
+concurrency:
+  group: ${{github.workflow}} - ${{github.ref}}
```
fniephaus, line 55: Why `- ${{github.ref}}` here and not in `micronaut.yml`? There are also other workflows that may need a `concurrency` rule such as `spring.yml`.

fniephaus (changes requested):
Thanks for the contribution. it was a conscious decision to keep the number of OS version to a minimum, and `windows-2019` won't even work. The `concurrency` policy may be useful to have. Please take a look at my comments.

.github/workflows/quarkus.yml
ceddy4395: I've added the `${{github.ref}}` because it can also be triggered by pull requests. Therefore, I was assuming that if there are two pr's running this workflow, one should not cancel the other. 
If this is not the case, the `${{github.ref}}` can be removed.

.github/workflows/main.yml
ceddy4395: Understood! I'll take it out

.github/workflows/main.yml
ceddy4395: Ah yes I see, thank you for the feedback, let me take it out :smile:

.github/workflows/quarkus.yml
fniephaus: In `main.yml`, we've got:
```
group: "workflow = ${{ github.workflow }}, ref = ${{ github.event.ref }}, pr = ${{ github.event.pull_request.id }}"
```

Any reason not to use this consistently? My point was also about being consistent across all other workflows (e.g., micronaut.yml and spring.yml)

.github/workflows/main.yml
```diff
@@ -91,7 +91,7 @@ permissions:
 jobs:
   build-graalvm-linux:
     name: /${{ matrix.env.PRIMARY }} ${{ matrix.env.GATE_TAGS }} JDK${{ matrix.env.JDK_VERSION }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
fniephaus, line 94: Can we keep `ubuntu-20.04` for now?

ceddy4395: Yes of course, mistake on my side! 
Out of curiosity, any specific reason not to use 22.04?

fniephaus: Yes: we want to test with the oldest glibc available (ubuntu-20.04 is the oldest ubuntu currently supported by GHA), so that we know things are still working there. The moment the glibc version gets bumped, we will know because the builds will fail and then will we move to 22.04. The same is true for any other dev and non-dev dependency.

ceddy4395: I see, thanks for the clarification :smile:

.github/workflows/quarkus.yml
ceddy4395: Ahh I see, sorry for the misunderstanding. 
I've changed it so that it is consistent with the `main.yml`. Same for the `reachabilit-metadata.yml`, `spring.yml` and `micronaut.yml`. 
I did not change the `cdt-inspect.yml` because this is only triggered through a schedule.

fniephaus:
I'm closing this due to inactivity. Please feel free to comment in case it should be reopened.

PR #8859, opened 2024-04-29
[GR-53747] Enable debug mode for native-image and native-image-configure

lgxbslgx:
Hi all,

When I use the command `mx -d -v native-image HelloWorld` to try to debug the compiler driver, the process doesn't stop and doesn't listen at port `8000`. After diving into the code, I find that the command `native-image` of the `mx` can't identify and process the JDK arguments. And I find the `native-image-configure` also has the same problem. This patch fixes them.

Thanks for taking the time to review. This is my first time contributing to GraalVM, maybe I need some guidance. Thanks again.

Best Regards,
-- Guoxiong

### DIFF
--- a/substratevm/mx.substratevm/mx_substratevm.py
+++ b/substratevm/mx.substratevm/mx_substratevm.py
@@ -2024,12 +2024,10 @@ def native_image_on_jvm(args, **kwargs):
             args.append("-D" + key + "=" + value)
 
     arg = [executable]
-    debug_args = []
     jdk = get_jdk()
     if jdk.debug_args and not mx.is_debug_disabled():
-        debug_args = jdk.debug_args
-    for cur_arg in debug_args:
-        args.insert(0, '--vm.' + cur_arg[1:])
+        for cur_arg in jdk.debug_args:
+            args.insert(0, '--vm.' + cur_arg[1:])
     jacoco_args = mx_gate.get_jacoco_agent_args(agent_option_prefix='-J')
     if jacoco_args is not None:
         arg += jacoco_args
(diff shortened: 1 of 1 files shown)


substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -2024,6 +2024,9 @@ def native_image_on_jvm(args, **kwargs):
             args.append("-D" + key + "=" + value)
 
     arg = [executable]
+    jdk_args = get_jdk().processArgs([])
+    for cur_arg in jdk_args:
+        args.insert(0, '--vm.' + cur_arg[1:])
```
fniephaus, line 2029: This is a neat hack, just afraid this accidentally pulls in more options that are not working as expected. Instead, we could just do what `add_debug_args()` does within mx. We will follow up

fniephaus:
Note: this allows users to debug the `native-image-configure` and `native-image` launchers. The latter is actually the driver process, not the actual builder process. For debugging the builder process (the one performing the static analysis and compilation), you can use `--debug-attach`.

substratevm/mx.substratevm/mx_substratevm.py
lgxbslgx: I updated the code just now. Current code only adds the debug arguments.

Maybe we should mark `add_debug_args()` as not local method so that it can be re-used. But such refactor is better to be done at another PR. What do you think about it?

fniephaus: Thanks! I actually just [mirrored your PR](https://github.com/oracle/graal/pull/8861) internally and pushed a similar commit: https://github.com/oracle/graal/pull/8861/commits/1cc9f382a1be175ea3e1c81e82280113eb93d97a

lgxbslgx: > Thanks! I actually just [mirrored your PR](https://github.com/oracle/graal/pull/8861) internally and pushed a similar commit: [1cc9f38](https://github.com/oracle/graal/commit/1cc9f382a1be175ea3e1c81e82280113eb93d97a)

Thanks. I am not familiar to the dev flow of Graal now, but it is strange to has two similar PRs.

fniephaus: No problem. We need to mirror external contributions in our internal infrastructure and because I added a commit, the two branches have diverged so the mirror created another PR. Are you ok with my changes in https://github.com/oracle/graal/pull/8861?

fniephaus: Ok, great. Thank you! I will take it from here :)

fniephaus:
Thanks for contributing! Merged via #8861.

PR #8868, opened 2024-05-01
Minor typo fix in docs

arvyy:


fniephaus:
Thanks for the PR. We can unfortunately only merge it when you have signed the OCA. Are you willing to sign it? If not, we can fix the typo independently from your PR.

arvyy:
No, I won't be signing an OCA as long as oracle account creator keeps asking me personal information like my home address and phone number.

Issue #8814, opened 2024-04-22
Dump a native memory tracking (NMT) report on signal


roberttoyonaga:
**Feature request**

Please include the following information:

**Is your feature request related to a problem? Please describe.**
Not related to a problem/bug. 

**Describe the solution you'd like.**
It would be great to be able to dump an NMT report upon receiving a signal (similar to heap dumps). 

In regular Java mode, NMT reports can be requested via jcmd. However, since Native Image doesn't have jcmd, a report is only written upon program completion. The usefulness of this is limited because the report will only show a snapshot of the memory usage at the time shutdown hooks are run (the data is kept as an instantaneous snapshot).

When NMT is enabled alongside JFR, JFR events will continuously expose NMT data. This solves the aforementioned problem of only getting a snapshot upon shutdown. However, there is no guarantee that JFR is included in the image along with NMT. JFR also has higher overhead than NMT so users may not want to include it. 

SIGUS1 and SIGUSR2 are already taken by heap dumps and compilation dumps respectively. One possibility is to dump an NMT report whenever a heap dump is created. This makes sense because the NMT report fills in memory information not covered by heap dumps. Another possibility is to use SIGUSR1 or SIGUSR2 depending on whether they are already used by other features. This approach is a little messy and only works if all 3 features aren't being used. 

**Describe who do you think will benefit the most.**
GraalVM users and developers can benefit from this. 

**Describe alternatives you've considered.**
Another approach might be to create a SubstrateVM specific MXBean for NMT. This MXBean would be capable of initiating an NMT report dump through remote JMX. Users would need to include the remote JMX feature in their image. Then they could remotely request NMT reports from JDK Mission Control, VisualVM, or some other method of JMX connection. 

I don't think this is the best solution since it requires including another feature in the image. 

**Express whether you'd like to help contributing this feature**
I can help work on this, if others think it is a useful addition.

fernando-valdez:
Thanks for requesting this new feature. 
We will keep track in this internal ticket: GR-53596

roberttoyonaga:
@christianhaeubl how do you feel about the proposed ideas?

christianhaeubl:
Dumping NMT data on `SIGUSR1` sounds reasonable. In the long run, I think that we should try to move away from the signal-based approach (there is no good way to specify which data should be dumped).

roberttoyonaga:
[...]

PR #8876, opened 2024-05-02
[GR-53596] Support dumping NMT reports on SIGUSR1

roberttoyonaga:
### Summary
This PR adds support for dumping NTM reports on SIGUSR1.  SIGUSR1 is shared with the heap dump feature. If both heap dump and NMT features are included in the image, then both types of report will be dumped.

This change will allow users to sample NMT data at arbitrary points in application execution. Previously, you could only get a report once the application finishes.

Related GH issue: https://github.com/oracle/graal/issues/8814


roberttoyonaga:
Dumping NMT reports through JCMD is preferred over this method using signals. If https://github.com/oracle/graal/pull/9232 is integrated, we should not integrate this PR.

PR #8888, opened 2024-05-07
Print exception details/stacktrace when failed to add a library

### LINKED ISSUE #8887: Print exception details/stacktrace when failed to add a library
https://github.com/oracle/graal/blob/481625b5da2b293cc672fd4b9348612483b6e76c/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibraries.java#L148-L150

I would be helpful to print exception details/stacktrace when failed to add a library.

Eng-Fouad:
Resolves #8887

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibraries.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibraries.java
@@ -146,6 +146,7 @@ public abstract class NativeLibraries {
             String canonical = builtin ? file.getName() : file.getCanonicalPath();
             return addLibrary(canonical, builtin);
         } catch (IOException e) {
+            e.printStackTrace();
             return false;
         }
     }


selhagani:
Hey @Eng-Fouad, Thank you for your contribution!

@wirthi Could you please assign this to someone that can review this PR. Thanks a lot!

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/NativeLibraries.java
```diff
@@ -146,6 +146,7 @@ private boolean loadLibrary0(File file, boolean builtin) {
             String canonical = builtin ? file.getName() : file.getCanonicalPath();
             return addLibrary(canonical, builtin);
         } catch (IOException e) {
+            e.printStackTrace();
```
peter-hofer, line 149: We don't know where the library is supposed to be and this method is used for probing, so adding this might print several unhelpful stack traces even when the library can ultimately be loaded. I believe the behavior is the same in OpenJDK, see `jdk.internal.loader.NativeLibraries#loadLibrary(java.lang.Class<?>, java.io.File)`.

Eng-Fouad: @peter-hofer At least print it with debug level if possible instead of hiding the exception details entirely.

Eng-Fouad:
Printing the stack trace with debug level would be really helpful.

PR #8988, opened 2024-05-29
Add premain for static agent support

ziyilin:
Java agent always has a premain method to initialize the agent. SVM needs the premain as well to support agent in native image.

At compile time, -H:PremainClasses= option is used to set the premain classes.
At runtime, premain runtime options are set along with main class' arguments in the format of -XX-premain:[class]:[options]

This PR is part of https://github.com/oracle/graal/pull/8077

### DIFF
--- a/substratevm/CHANGELOG.md
+++ b/substratevm/CHANGELOG.md
@@ -3,6 +3,7 @@
 This changelog summarizes major changes to GraalVM Native Image.
 
 ## GraalVM for JDK 24 (Internal Version 24.2.0)
+* Together with Alibaba, we added java agent support for native image as experimental feature. 
 * (GR-54476): Issue a deprecation warning on first use of a legacy `graal.` prefix (see GR-49960 in [Compiler changelog](../compiler/CHANGELOG.md)).
   The warning is planned to be replaced by an error in GraalVM for JDK 25.
 * (GR-48384) Added a GDB Python script (`gdb-debughelpers.py`) to improve the Native Image debugging experience.
--- a/substratevm/mx.substratevm/mx_substratevm.py
+++ b/substratevm/mx.substratevm/mx_substratevm.py
@@ -1703,6 +1703,66 @@ def cinterfacetutorial(args):
     native_image_context_run(_cinterfacetutorial, args)
 
 
+@mx.command(suite.name, 'javaagenttest', 'Runs tests for java agent with native image')
+def agenttest(args):
+    def build_and_test_agent_image(native_image, args):
+        args = [] if args is None else args
+        build_dir = join(svmbuild_dir(), 'javaagenttest')
+
+        # clean / create output directory
+        if exists(build_dir):
+            mx.rmtree(build_dir)
+        mx_util.ensure_dir_exists(build_dir)
+        test_build_output = mx.dependency('com.oracle.svm.test').classpath_repr()
+
+        # create agent jar file
+        agent1 = join(build_dir, 'testagent1.jar')
+        mx.run([mx.get_jdk().jar, 'cmf', join(test_build_output, 'resources','agent1', 'MANIFEST.MF'),
+                agent1,
+                join('com', 'oracle', 'svm', 'test', 'agent', 'Agent.class'),
+                join('com', 'oracle', 'svm', 'test', 'agent', 'Agent$DemoTransformer.class'),
+                join('com', 'oracle', 'svm', 'test', 'agent', 'AgentPremainHelper.class')],
+               cwd = test_build_output)
+
+        agent2 = join(build_dir, 'testagent2.jar')
+        mx.run([mx.get_jdk().jar, 'cmf', join(test_build_output, 'resources','agent2', 'MANIFEST.MF'),
+                agent2,
+                join('com', 'oracle', 'svm', 'test', 'agent', 'Agent2.class'),
+                join('com', 'oracle', 'svm', 'test', 'agent', 'AgentPremainHelper.class')],
+               cwd = test_build_output)
+
+        # build test with agent
+        test_cp = os.pathsep.join([classpath('com.oracle.svm.test'), agent1, agent2])
+
+
+        nativeagent_premain_options=['-XXpremain:com.oracle.svm.test.agent.Agent:test.agent=true',
+                                     '-XXpremain:com.oracle.svm.test.agent.Agent2:test.agent2=true']
+
+        binary_path = join(build_dir, 'agenttest1')
+        native_image([
+                         '-cp', test_cp,
+                         '-J-ea', '-J-esa',
+                         '-o', binary_path,
+                         '-H:+ReportExceptionStackTraces',
+                         '-H:PremainClasses=com.oracle.svm.test.agent.Agent,com.oracle.svm.test.agent.Agent2',
+                         '-H:Class=com.oracle.svm.test.agent.AgentTest',
+                     ]  + args)
+        mx.run([binary_path] + nativeagent_premain_options)
+
+        # Switch the premain sequence of agent1 and agent2
+        binary_path = join(build_dir, 'agenttest2')
+        native_image([
+                         '-cp', test_cp,
+                         '-J-ea', '-J-esa',
+                         '-o', binary_path,
+                         '-H:+ReportExceptionStackTraces',
+                         '-H:PremainClasses=com.oracle.svm.test.agent.Agent2,com.oracle.svm.test.agent.Agent',
+                         '-H:Class=com.oracle.svm.test.agent.AgentTest',
+                     ]  + args)
+        mx.run([binary_path] + nativeagent_premain_options)
+
+    native_image_context_run(build_and_test_agent_image, args)
+
 @mx.command(suite.name, 'clinittest', 'Runs the ')
 def clinittest(args):
     def build_and_test_clinittest_image(native_image, args):
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
@@ -1,0 +1,297 @@
+/*
+ * Copyright (c) 2024, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024, 2024, Alibaba Group Holding Limited. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.svm.core;
+
+import com.oracle.svm.core.c.NonmovableArrays;
+import com.oracle.svm.core.code.CodeInfo;
+import com.oracle.svm.core.code.CodeInfoAccess;
+import com.oracle.svm.core.code.CodeInfoTable;
+import com.oracle.svm.core.jdk.ModuleNative;
+import com.oracle.svm.core.util.VMError;
+import org.graalvm.nativeimage.Platform;
+import org.graalvm.nativeimage.Platforms;
+
+import java.lang.instrument.ClassDefinition;
+import java.lang.instrument.ClassFileTransformer;
+import java.lang.instrument.Instrumentation;
+import java.lang.instrument.UnmodifiableClassException;
+import java.lang.instrument.UnmodifiableModuleException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.jar.JarFile;
+
+/**
+ * Java agent can do instrumentation initialization work in premain phase. This class supports
+ * registering such premain methods at native image build time and invoking them at native image
+ * runtime. <br>
+ * JVM supports two kind of premain methods:
+ * <ol>
+ * <li>{@code public static void premain(String agentArgs, Instrumentation inst)}</li>
+ * <li>{@code public static void premain(String agentArgs)}</li>
+ * </ol>
+ * For the first one, at registration time we will set the second parameter with an instance of
+ * {@link NativeImageNoOpRuntimeInstrumentation} class which does not do any actual work as no
+ * instrumentation can do in native code at runtime.
+ * <p>
+ * <b>Be noticed</b>, the original agent premain method may not work well at native image runtime
+ * even if the input {@link Instrumentation} class is replaced with
+ * {@link NativeImageNoOpRuntimeInstrumentation}.
+ * </p>
+ * <p>
+ * It is the agent developers' responsibility to implement a native version of their agent premain
+ * method. It can be implemented in two ways:
+ * <ul>
+ * <li>Isolate code by checking current runtime. For example: <code>
+ *         <pre>
+ *         if ("runtime".equals(System.getProperty("org.graalvm.nativeimage.imagecode"))) {
+ *           // native image runtime
+ *         } else{
+ *           // JVM runtime
+ *         }
+ *         </pre>
+ *     </code> Alternatively: Instead of directly getting property,
+ * <code>ImageInfo.inImageRuntimeCode()</code> can also be used to check if current runtime is
+ * native image runtime, but it requires extra dependency.</li>
+ * <li>Use {@link com.oracle.svm.core.annotate.TargetClass} API to implement a native image version
+ * premain.</li>
+ * </ul>
+ * </p>
+ */
+public class PreMainSupport {
+
+    private static final String PREMAIN_OPTION_PREFIX = "-XXpremain:";
+
+    /**
+     * A record for premain method.
+     *
+     * @param className the full qualified class name that declares the premain method
+     * @param method the premain method
+     * @param args the arguments of premain method
+     */
+    record PremainMethod(String className, Method method, Object[] args) {
+    }
+
+    private Map<String, String> premainOptions = new HashMap<>();
+    // Order matters
+    private List<PremainMethod> premainMethods = new ArrayList<>();
+
+    @Platforms({Platform.HOSTED_ONLY.class})
+    public void registerPremainMethod(String className, Method executable, Object... args) {
+        premainMethods.add(new PremainMethod(className, executable, args));
+    }
+
+    /**
+     * Retrieve premain options from input args. Keep premain options and return the rest args as
+     * main args. Multiple agent options should be given in separated {@code -XX-premain:} leading
+     * arguments. The premain options format: <br>
+     * -XX-premain:[full.qualified.premain.class]:[premain options]
+     * -XX-premain:[full.qualified.premain.class2]:[premain options] <br>
+     *
+     * @param args original arguments for premain and main
+     * @return arguments for main class
+     */
+    public String[] retrievePremainArgs(String[] args) {
+        List<String> mainArgs = new ArrayList<>();
+
+        for (String arg : args) {
+            if (arg.startsWith(PREMAIN_OPTION_PREFIX)) {
+                String premainOptionKeyValue = arg.substring(PREMAIN_OPTION_PREFIX.length());
+                String[] pair = premainOptionKeyValue.split(":");
+                if (pair.length == 2) {
+                    premainOptions.put(pair[0], pair[1]);
+                }
+            } else {
+                mainArgs.add(arg);
+            }
+        }
+        return mainArgs.toArray(new String[0]);
+    }
+
+    public void invokePremain() {
+        for (PremainMethod premainMethod : premainMethods) {
+
+            Object[] args = premainMethod.args;
+            if (premainOptions.containsKey(premainMethod.className)) {
+                args[0] = premainOptions.get(premainMethod.className);
+            }
+            try {
+                // premain method must be static
+                premainMethod.method.invoke(null, args);
+            } catch (Throwable t) {
+                if (t instanceof InvocationTargetException) {
+                    t = t.getCause();
+                }
+                VMError.shouldNotReachHere("Fail to execute " + premainMethod.className + ".premain", t);
+            }
+        }
+    }
+
+    /**
+     * This class is a dummy implementation of {@link Instrumentation} interface. It serves as the
+     * registered premain method's second parameter. At native image runtime, no actual
+     * instrumentation work can do. So all the methods here are empty.
+     */
+    public static class NativeImageNoOpRuntimeInstrumentation implements Instrumentation {
+
+        private static final Set<String> systemModules = Set.of("org.graalvm.nativeimage.builder", "org.graalvm.nativeimage", "org.graalvm.nativeimage.base", "com.oracle.svm.svm_enterprise",
+                        "org.graalvm.word", "jdk.internal.vm.ci", "jdk.graal.compiler", "com.oracle.graal.graal_enterprise");
+
+        @Override
+        public void addTransformer(ClassFileTransformer transformer, boolean canRetransform) {
+        }
+
+        @Override
+        public void addTransformer(ClassFileTransformer transformer) {
+        }
+
+        @Override
+        public boolean removeTransformer(ClassFileTransformer transformer) {
+            return false;
+        }
+
+        @Override
+        public boolean isRetransformClassesSupported() {
+            return false;
+        }
+
+        @Override
+        public void retransformClasses(Class<?>... classes) throws UnmodifiableClassException {
+            throw new UnsupportedOperationException("Native image doesn't support retransform class at runtime.");
+        }
+
+        @Override
+        public boolean isRedefineClassesSupported() {
+            return false;
+        }
+
+        @Override
+        public void redefineClasses(ClassDefinition... definitions) throws ClassNotFoundException, UnmodifiableClassException {
+            throw new UnsupportedOperationException("Native image doesn't support redefine class at runtime.");
+        }
+
+        @Override
+        public boolean isModifiableClass(Class<?> theClass) {
+            return false;
+        }
+
+        @Override
+        public Class<?>[] getAllLoadedClasses() {
+            ArrayList<Class<?>> userClasses = new ArrayList<>();
+            CodeInfo imageCodeInfo = CodeInfoTable.getFirstImageCodeInfo();
+            while (imageCodeInfo.isNonNull()) {
+                Class<?>[] classes = NonmovableArrays.heapCopyOfObjectArray(CodeInfoAccess.getClasses(imageCodeInfo));
+                if (classes != null) {
+                    for (Class<?> clazz : classes) {
+                        if (clazz != null) {
+                            Module module = clazz.getModule();
+                            if (module == null ||
+                                            module.getName() == null ||
+                                            !isSystemClass(module)) {
+                                userClasses.add(clazz);
+                            }
+                        }
+                    }
+                }
+                imageCodeInfo = CodeInfoAccess.getNextImageCodeInfo(imageCodeInfo);
+            }
+            return userClasses.toArray(new Class<?>[0]);
+        }
+
+        private static boolean isSystemClass(Module module) {
+            return systemModules.contains(module.getName());
+        }
+
+        @Override
+        public Class<?>[] getInitiatedClasses(ClassLoader loader) {
+            throw new UnsupportedOperationException("Native image has flat classloader hierarchy. Can't get classes by classloader. " +
+                            "Try to call getAllLoadedClasses() to get all loaded classes.");
+        }
+
+        @Override
+        public long getObjectSize(Object objectToSize) {
+            return -1;
+        }
+
+        @Override
+        public void appendToBootstrapClassLoaderSearch(JarFile jarfile) {
+            throw new UnsupportedOperationException("Native image doesn't support modification of the bootstrap classloader search path at run time." +
+                            " Please avoid calling this method in Native Image by checking \"runtime\".equals(System.getProperty(\"org.graalvm.nativeimage.imagecode\"))");
+        }
+
+        @Override
+        public void appendToSystemClassLoaderSearch(JarFile jarfile) {
+            throw new UnsupportedOperationException("Native image doesn't support modification of the system classloader search path at run time." +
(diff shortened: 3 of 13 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,234 @@
...
+ * <li>Use {@link com.oracle.svm.core.annotate.TargetClass} API to implement a native image version
+ * premain.</li>
+ * </ul>
+ * </p>
+ */
+public class PreMainSupport {
+
+    private static final String PREMAIN_OPTION_PREFIX = "-XX-premain:";
+
+    class PremainMethod {
```
vjovanov, line 85: This could be a record, right?

vjovanov: Should be `static` if it is a class.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,234 @@
...
+        for (PremainMethod premainMethod : premainMethods) {
+            try {
+                Object[] args = premainMethod.args;
+                if (premainOptions.containsKey(premainMethod.className)) {
+                    args[0] = premainOptions.get(premainMethod.className);
+                }
+                // premain method must be static
+                premainMethod.method.invoke(null, args);
+            } catch (Throwable t) {
+                VMError.shouldNotReachHere("Fail to execute " + premainMethod.className + ".premain", t);
```
vjovanov, line 144: This should be a user-facing error, right? User code can throw, so I don't see this as something that is a `VMError`. What does the JVM do in this case?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,234 @@
...
+            }
+        }
+    }
+
+    /**
+     * This class is a dummy implementation of {@link Instrumentation} interface. It serves as the
+     * registered premain method's second parameter. At native image runtime, no actual
+     * instrumentation work can do. So all the methods here are empty.
+     */
+    public static class SVMRuntimeInstrumentImpl implements Instrumentation {
```
vjovanov, line 154: Maybe `NativeImageNoOpRuntimeInstrumentation`?

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/InstrumentFeature.java
```diff
@@ -0,0 +1,118 @@
...
+            } catch (NoSuchMethodException e) {
+                try {
+                    premain = clazz.getDeclaredMethod("premain", String.class);
+                } catch (NoSuchMethodException e1) {
+                    VMError.shouldNotReachHere("Can't find premain method in " + premainClass, e1);
+                }
+            }
+            preMainSupport.registerPremainMethod(premainClass, premain, args.toArray(new Object[0]));
+        } catch (ClassNotFoundException e) {
+            VMError.shouldNotReachHere("Can't register agent premain method, because the declaring class " + premainClass + " is not found", e);
```
vjovanov, line 113: This is a `UserError` as the user can make a mistake. We should also add a sentence on how to recover from that mistake to the error message.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,234 @@
...
+ * even if the input {@link Instrumentation} class is replaced with
+ * {@link SVMRuntimeInstrumentImpl}.
+ * </p>
+ * <p>
+ * It is the agent developers' responsibility to implement a native version of their agent premain
+ * method. It can be implemented in two ways:
+ * <ul>
+ * <li>Isolate code by checking current runtime. For example: <code>
+ *         <pre>
+ *         String vm = System.getProperty("java.vm.name");
```
vjovanov, line 68: We should guide the users to use the properties from [ImageInfo](https://github.com/oracle/graal/blob/2449afbb28dadf2d9a70a866da06d664f1d1a539/sdk/src/org.graalvm.nativeimage/src/org/graalvm/nativeimage/ImageInfo.java#L67). Then they can see if the agent is applied at build time or at runtime. We also need to make sure that this property is set before premain is called for the Native Image builder.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,234 @@
...
+    public void registerPremainMethod(String className, Method executable, Object... args) {
+        premainMethods.add(new PremainMethod(className, executable, args));
+    }
+
+    /**
+     * Retrieve premain options from input args.
+     * Keep premain options and return the rest args as main args.
+     * The premain options format:
+     * <br>
+     * -XX-premain:[full.qualified.premain.class]:[premain options]
```
vjovanov, line 111: I am not sure if this should be our final format for the flag. I think we can use it until we complete the feature and decide what is the user interface in the end.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/InstrumentFeature.java
```diff
@@ -0,0 +1,118 @@
...
+ * This feature supports instrumentation in native image.
+ */
+@AutomaticallyRegisteredFeature
+public class InstrumentFeature implements InternalFeature {
+    private ClassLoader cl;
+    private PreMainSupport preMainSupport;
+
+    public static class Options {
+        @Option(help = "Specify premain-class list. Multiple classes are separated by comma, and order matters.")//
+        public static final HostedOptionKey<LocatableMultiOptionValue.Strings> PremainClasses = new HostedOptionKey<>(null);
```
vjovanov, line 55: This is also an internal option. In the end we should probably pick the `premain` functions from the agent JARs.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: In JVM, if the `premain` method fails, the JVM starting process shall be terminated. See https://github.com/openjdk/jdk21u/blob/2971cb5769121b47ac8c8db1078d67680a19341f/src/java.instrument/share/native/libinstrument/InvocationAdapter.c#L623-L629

I use `VMError` because it's a fatal error during VM starts up. But the cause is indeed an user error.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/InstrumentFeature.java
```diff
@@ -0,0 +1,118 @@
...
+     */
+    @Override
+    public List<Class<? extends Feature>> getRequiredFeatures() {
+        return List.of(ReflectionFeature.class);
+    }
+
+    @Override
+    public void afterRegistration(AfterRegistrationAccess access) {
+        FeatureImpl.AfterRegistrationAccessImpl a = (FeatureImpl.AfterRegistrationAccessImpl) access;
+        cl = a.getImageClassLoader().getClassLoader();
```
sor-bae, line 71: I'm not sure this is still required, but I had to provide an agent JAR twice as `-J-javaagent` and `-cp` for NI build configurations in agent JARs to be picked up by the NI builder. So just checking, users can provide an agent JAR only once as `-J-javaagent` for their premain classes to be registered?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: > This could be a record, right?
Yes, in my current implementation it's recorded by native-image-agent.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: Fixed. This property can be obtained in `premain`.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/InstrumentFeature.java
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: Agreed. I added a `//todo` for future changing.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/InstrumentFeature.java
ziyilin: In this PR, we are not automatically picking agent from `-cp` or `-J-javaagent`.  It will be handled in the following PRs. But I'm sure user doesn't need to provide agent jar multiple times.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: > This could be a record, right?

Yes, in my current implementation it's recorded by native-image-agent.

vjovanov: When I say a `record` I mean a [java record](https://openjdk.org/jeps/359).

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/InstrumentFeature.java
```diff
@@ -0,0 +1,117 @@
...
+            } catch (NoSuchMethodException e) {
+                try {
+                    premain = clazz.getDeclaredMethod("premain", String.class);
+                } catch (NoSuchMethodException e1) {
+                    UserError.abort(e1, "Can't register agent premain method, because can't find the premain method from the given class %s. Please check your -H:%s setting.", premainClass, Options.PremainClasses.getName());
+                }
+            }
+            preMainSupport.registerPremainMethod(premainClass, premain, args.toArray(new Object[0]));
+        } catch (ClassNotFoundException e) {
+            UserError.abort(e,"Can't register agent premain method, because the given class %s is not found. Please check your -H:%s setting.", premainClass, Options.PremainClasses.getName());
```
vjovanov, line 114: It is better to use `SubstrateOptionsParser.commandArgument` to print the option name. This keeps the correct behavior  when options become public or change the type.

vjovanov:
To keep this feature working as expected we need to add a few tests into: `com.oracle.svm.test`

The tests can be executed from similarly to the tests for class initialization: `https://github.com/oracle/graal/blob/master/substratevm/mx.substratevm/mx_substratevm.py#L1721`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: I see. Yes, this is a java record. I have fixed it.

ziyilin:
> To keep this feature working as expected we need to add a few tests into: `com.oracle.svm.test`
> 
> The tests can be executed from similarly to [the tests for class initialization](https://github.com/oracle/graal/blob/master/substratevm/mx.substratevm/mx_substratevm.py#L1721).

I have added a test task [agenttest](https://github.com/oracle/graal/pull/8988/files#diff-eab1ebfe7b30eac56f68f32f6eff8466a51b176a60fcd6b0309d974f48579057R1721). It now only checks the `premain` method, doesn't do any class transformation yet.

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -1717,6 +1717,66 @@ def cinterfacetutorial(args):
     native_image_context_run(_cinterfacetutorial, args)
 
 
+@mx.command(suite.name, 'agenttest', 'Runs the ')
```
vjovanov, line 1720: I feel it would be more accurate to call it `javaagenttest`.
Also, we should complete the usage message here. Something like:
```
usage_msg='Runs tests for java agents with native image'
```

vjovanov:
We need to add the `CHANGELOG.md` entry that claims that this is an experimental feature for now.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/agent/Agent2.java
```diff
@@ -0,0 +1,53 @@
...
+        if (agentArgs != null && !agentArgs.isBlank()) {
+            String[] argPairs = agentArgs.split(",");
+            for (String argPair : argPairs) {
+                String[] pair = argPair.split("=");
+                if (pair.length == 2) {
+                    System.setProperty(pair[0], pair[1]);
+                }
+            }
+        }
+        System.setProperty("instrument.enable", "true");
```
vjovanov, line 46: I would add a set of assertions that checks that the values for `Instrumentation` are correct at image run time. This way we assure nobody changes those values and surprises the users.

ziyilin:
> We need to add the `CHANGELOG.md` entry that claims that this is an experimental feature for now.

Not sure which version and the internal code, I added the entry at the top of  the CHANGELOG.md.

substratevm/mx.substratevm/mx_substratevm.py
ziyilin: Fixed

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/agent/Agent2.java
ziyilin: The check of Instrumentation class is put in the Agent.

vjovanov: I would still add a check for all the relevant methods that can be invoked on the `Instrumentation` good examples are: `Instrumentation#isModifiableClass`, `Instrumentation#isRetransformClassesSupported`, `Instrumentation#isModifiableModule`, `Instrumentation#getAllLoadedClasses`, etc.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        @Override
+        public boolean isModifiableClass(Class<?> theClass) {
+            return false;
+        }
+
+        @Override
+        public Class<?>[] getAllLoadedClasses() {
+            return new Class<?>[0];
```
vjovanov, line 191: I think we should discuss this. How is this method typically used? 

Should we maybe return all the classes that were reached? I think that we need an explanation for this feature.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        @Override
+        public Class<?>[] getInitiatedClasses(ClassLoader loader) {
+            return new Class<?>[0];
+        }
+
+        @Override
+        public long getObjectSize(Object objectToSize) {
+            return 0;
```
vjovanov, line 201: We should either return the actual value by looking into Native Image internals, or something completely unusable such as `Long.MIN_VALUE`. Is this information used to transform classes?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,231 @@
...
+        public boolean isNativeMethodPrefixSupported() {
+            return false;
+        }
+
+        @Override
+        public void setNativeMethodPrefix(ClassFileTransformer transformer, String prefix) {
+        }
+
+        @Override
+        public void redefineModule(Module module, Set<Module> extraReads, Map<String, Set<Module>> extraExports, Map<String, Set<Module>> extraOpens, Set<Class<?>> extraUses,
```
vjovanov, line 222: I feel this should be a hard failure that will explain to the users what they should do (preferable). 

Now, the question is how many agents will call this even when transformation is not allowed. I suspect not many.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,231 @@
...
+                // premain method must be static
+                premainMethod.method.invoke(null, args);
+            } catch (Throwable t) {
+                VMError.shouldNotReachHere("Fail to execute " + premainMethod.className + ".premain", t);
+            }
+        }
+    }
+
+    /**
+     * This class is a dummy implementation of {@link Instrumentation} interface. It serves as the
```
vjovanov, line 147: We need to mention in the release notes that this is experimental and that the behavior of some methods might change. We should test only the methods for which we are completely certain they will stay the same.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,231 @@
...
+ * even if the input {@link Instrumentation} class is replaced with
+ * {@link NativeImageNoOpRuntimeInstrumentation}.
+ * </p>
+ * <p>
+ * It is the agent developers' responsibility to implement a native version of their agent premain
+ * method. It can be implemented in two ways:
+ * <ul>
+ * <li>Isolate code by checking current runtime. For example: <code>
+ *         <pre>
+ *         if (ImageInfo.inImageRuntimeCode()) {
```
vjovanov, line 68: I would prefer to use the system property directly here, or at least mention it. This will not require extra dependencies to use this feature.

substratevm/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 This changelog summarizes major changes to GraalVM Native Image.
 
 ## GraalVM for JDK 24 (Internal Version 24.2.0)
+* Together with Alibaba, we added java agent support for native image as experimental feature.
```
vjovanov, line 6: I will polish this line when the PR is done.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: This method is called to declare extra exports/opens/uses/reads packages from module.
It is not rarely used. The agent may need extra access to certain modules.
We can call APIs from [ModuleNative](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/ModuleNative.java) to make checks only.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: In observability scenario, user can call this method to collect all jars loaded by the application.
I think return the reached classes in native image here is reasonable.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: This method is rarely used. 
It is not used by class transformation.
I've searched the source of [opentelemetery java instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation) and didn't find any usage as well.
In OpenJDK, it returns `oop->size() * wordSize`. What is the similar function in substratevm? We can do the similar calculation so it will be perfectly compatible with JVM. But considering the rarely usage, `Long.MIN_VALUE` may be also acceptable for now.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
trask: I'd suggest `throw new UnsupportedOperationException()` based on the javadoc for the method https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/Instrumentation.html#redefineClasses-java.lang.instrument.ClassDefinition...-

> [UnsupportedOperationException](https://docs.oracle.com/javase/8/docs/api/java/lang/UnsupportedOperationException.html) - if the current configuration of the JVM does not allow redefinition ([isRedefineClassesSupported()](https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/Instrumentation.html#isRedefineClassesSupported--) is false) or the redefinition attempted to make unsupported changes

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
trask: I'd suggest `0` in case anyone is summing over these values, otherwise I think they'll get negative overflow(?) by using MIN_VALUE

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
trask: > Should we maybe return all the classes that were reached?

this sounds good if it's possible

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,231 @@
...
+            }
+        }
+    }
+
+    /**
+     * This class is a dummy implementation of {@link Instrumentation} interface. It serves as the
+     * registered premain method's second parameter. At native image runtime, no actual
+     * instrumentation work can do. So all the methods here are empty.
+     */
+    public static class NativeImageNoOpRuntimeInstrumentation implements Instrumentation {
```
trask, line 151: I think no-op is ok for everything, under the assumption that agents can't do "runtime instrumentation" things in native images

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: Fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: I set it to -1 as it's both invalid and won't get overflowed easily.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,273 @@
...
+
+        @Override
+        public boolean isModifiableClass(Class<?> theClass) {
+            return false;
+        }
+
+        @Override
+        public Class<?>[] getAllLoadedClasses() {
+            synchronized (this) {
+                if (allLoadedClasses == null) {
```
vjovanov, line 206: I would not cache here as it takes up RSS.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,273 @@
...
+                for (Module m : entry.getValue()) {
+                    ModuleNative.addExports(module, entry.getKey(), m);
+                }
+            }
+            // Ignore the extraOpens, extraUses and extraProvides as they are not supported by
+            // ModuleNative.
+        }
+
+        @Override
+        public boolean isModifiableModule(Module module) {
```
vjovanov, line 266: @ivan-ristovic can you please look at this implementation?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,273 @@
...
+                            "Try to call getAllLoadedClasses() to get all loaded classes.");
+        }
+
+        @Override
+        public long getObjectSize(Object objectToSize) {
+            return -1;
+        }
+
+        @Override
+        public void appendToBootstrapClassLoaderSearch(JarFile jarfile) {
```
vjovanov, line 225: I would throw here as it is not really supported as well as to `appendToSystemClassLoaderSearch`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,273 @@
...
+ * native image runtime, but it requires extra dependency.</li>
+ * <li>Use {@link com.oracle.svm.core.annotate.TargetClass} API to implement a native image version
+ * premain.</li>
+ * </ul>
+ * </p>
+ */
+public class PreMainSupport {
+
+    // todo: Not finally decided, could be changed in the future
+    private static final String PREMAIN_OPTION_PREFIX = "-XX-premain:";
```
vjovanov, line 89: I would go with `-XXpremain:` to be consistent with other arguments. 

Once we make a flag for build-time transformation  this should be enabled automatically for enabled agents. Option will be there only for passing the arguments.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/InstrumentFeature.java
```diff
@@ -0,0 +1,120 @@
...
+ */
+@AutomaticallyRegisteredFeature
+public class InstrumentFeature implements InternalFeature {
+    private ClassLoader cl;
+    private PreMainSupport preMainSupport;
+
+    public static class Options {
+        // todo: Not finally decided, could be changed in the future
+        @Option(help = "Specify premain-class list. Multiple classes are separated by comma, and order matters.")//
+        public static final HostedOptionKey<AccumulatingLocatableMultiOptionValue.Strings> PremainClasses = new HostedOptionKey<>(
```
vjovanov, line 57: This should be an experimental option until we add an API option for this.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/InstrumentFeature.java
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * This feature supports instrumentation in native image.
+ */
+@AutomaticallyRegisteredFeature
+public class InstrumentFeature implements InternalFeature {
+    private ClassLoader cl;
+    private PreMainSupport preMainSupport;
+
+    public static class Options {
+        // todo: Not finally decided, could be changed in the future
```
vjovanov, line 55: Please remove all todos.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ivan-ristovic: LGTM. This is the same implementation as in `java.instrument` API, and our run-time module system has no specific quirks that require divergence from the JDK.

substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/InstrumentFeature.java
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: fixed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/hub/ClassForNameSupport.java
```diff
@@ -213,4 +215,15 @@ public RuntimeConditionSet getConditionFor(Class<?> jClass) {
             return conditionalClass.getConditions();
         }
     }
+
+    public Class<?>[] getAllLoadedClasses() {
```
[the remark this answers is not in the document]
ziyilin: Fixed. I put the `getAllLoadedClasses`  implementation to `PremainSupport`, because it doesn't need anything from `ClassForNameSupport` any more.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        @Override
+        public long getObjectSize(Object objectToSize) {
+            return -1;
+        }
+
+        @Override
+        public void appendToBootstrapClassLoaderSearch(JarFile jarfile) {
+            throw new UnsupportedOperationException("Native image doesn't support modify classloader search path at runtime.");
```
vjovanov, line 247: Maybe `Native image doesn't support modification of the classloader search path at run time. Please avoid calling this method in Native Image by checking  "runtime".equals(System.getProperty("org.graalvm.nativeimage.imagecode"))`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        @Override
+        public void appendToSystemClassLoaderSearch(JarFile jarfile) {
+            throw new UnsupportedOperationException("Native image doesn't support modify classloader search path at runtime.");
```
vjovanov, line 252: Similar here, as a matter of fact I would add the action sentence to all `UnsupportedOperationException`s

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
```diff
@@ -0,0 +1,296 @@
...
+                                            module.getName() == null ||
+                                            !isSystemClass(module)) {
+                                userClasses.add(clazz);
+                            }
+                        }
+                    }
+                }
+                imageCodeInfo = CodeInfoAccess.getNextImageCodeInfo(imageCodeInfo);
+            }
+            userClasses.trimToSize();
```
vjovanov, line 226: Now I see we don't need the `trimToSize` as the list is thrown away.

vjovanov:
This PR makes a good foundation for the agent support. After the final changes I will run all the tests, polish the `CHANGELOG.md` entry, and start with merging.

vjovanov (dismissed):

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: removed

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/PreMainSupport.java
ziyilin: Fixed

vjovanov (approved):
LGTM. Any last comments from other reviewers?

cemo:
@vjovanov will you merge this PR soon?

vjovanov:
The PR is in the queue and should be merged soon. We will also have follow-up PRs that push this feature forward.

vjovanov:
Merged: https://github.com/oracle/graal/commit/f16e655d85a2e541e0fc25d782ac6605709f4a28

Thank you @ziyilin!

Issue #8993, opened 2024-05-29
epsilon and G1 seem not to work with JFR in NI


buuuuuuug:
**Describe the issue**
epsilon GC and G1 GC seems cannot work properly with JFR enabled in NI mode by throwing an error msg of ` Pool jdk.types.GCName must contain at least one element` or `Pool jdk.types.GCCause must contain at least one element` sometimes on application start.
while serial GC works just fine.
is there anything that i possibly miss?

**Steps to reproduce the issue**
Please include both build steps as well as run steps
1. **[git clone https://github.com/buuuuuuug/jfr-vth-pin.git]**
2. **[gradle nativeCompile]**
3. **[modify build.gradle file by `buildArgs.add('--gc=G1')` or `buildArgs.add('--gc=sepsilon')`.]**
4. **[./build/native/nativeCompile/jfr-vth-pin -XX:StartFlightRecording="filename=recording.jfr,settings=continue+vth.jfc"]**

**Describe GraalVM and your environment:**
- GraalVM Oracle 21.02
- JDK major version: [21]
- OS: [macOS Monterey] [Ubuntu 22.04] [Rocky Linux 9.2]
- Architecture: [AMD64]

**More details**

```
    ~/IdeaProjects/jfr-vth-pin git:[main]
./build/native/nativeCompile/jfr-vth-pin -XX:StartFlightRecording="filename=recording.jfr,settings=continue+vth.jfc"
  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/

 :: Spring Boot ::                (v3.3.0)

2024-05-29T20:53:30.966+08:00  INFO 10553 --- [jfr-vth-pin] [           main] c.chaney.jfrvthpin.JfrVthPinApplication  : Starting AOT-processed JfrVthPinApplication using Java 21.0.2 with PID 10553 (/home/chaney/IdeaProjects/jfr-vth-pin/build/native/nativeCompile/jfr-vth-pin started by chaney in /home/chaney/IdeaProjects/jfr-vth-pin)
[...]

PR #9001, opened 2024-05-30
[GR-55269] Don't include GC related JFR serializers unless Serial GC is used.

roberttoyonaga:
Resolves Bug: https://github.com/oracle/graal/issues/8993

If epsilon or G1 GC is used, JFR does not emit any GC collection events. However, we were previously still serializing GC-related constant pools (which are empty in those cases). JFR consumer code, which parses chunks for jfr-streaming, expects these pools to be non-empty. 

This PR is just a small change to exclude those serializers when appropriate.


christianhaeubl:
Thanks, this PR will get merged in the next few days.

Issue #7962, opened 2023-12-05
Introduce a Field `type` in `reflect-config.json` that Includes all Type's Metadata in the Image


vjovanov:
### TL;DR

We'd like to include all the type's metadata for types that are registered for reflection as the default behavior in `reflect-config.json`. The newly included metadata would contain method signatures, inner classes, field metadata, record components, signers, and permitted subclasses. The transitively included classes would be included in the image, but not registered for reflection.

This is inspired by [#7753](https://github.com/oracle/graal/issues/7753) where we show that roughly 25% of registration entries can be removed when we use the proposed approach here.

Since this is a breaking change, we propose a new schema for inclusion:
```json
{
    "type": "app.App"
}
```
as this is anyhow required for [proxies and lambdas](https://github.com/oracle/graal/issues/7476). The new entry would be equivalent to
```json
{
    "name": "app.App",
    "queryAllDeclaredMethods": true,
    "queryAllDeclaredConstructors": true,
    "queryAllDeclaredFields": true,
    "queryAllPublicMethods": true,
    "queryAllPublicConstructors": true,
    "queryAllPublicFields": true, 
    "allDeclaredClasses": true,
    "allPublicClasses": true,
    "allRecordComponents": true,
    "allNestMembers": true,
    "allSigners": true,
    "allPermittedSubclasses": true
}
```

It should be possible to disable the metadata from a `type` by disabling certain properties. For example, we can exclude method metadata with:
```json
{
    "type": "app.App",
    "queryAllDeclaredMethods": false,
    "queryAllPublicMethods": false
}
```

Since this change will propagate quickly through the ecosystem, we intend to backport this change to GraalVM for JDK 17 and JDK 21.

The format with `name` will be kept functional in the foreseeable future due to the number of existing projects that use it. However, the agent will be changed to produce the format with `type`.

### Goals

1. Reduce the load of maintaining reachability metadata.
2. Allow the development of a programming model aimed to include metadata from user code. 
3. Simplify reasoning about reflection metadata.

### Non-Goals

1. Increase image size significantly--10% or more.
2. Reduce any of the other performance metrics for Native Image.

### Discussion points

1. Should `Class.forName("literal.Type")` include all class metadata in the image? Likely not, this would make the programming model unpredictable and bloat the image.

2. Should we include the default constructors with the `type`? Likely not, although it would reduce the amount of metadata by 10%. 

[...]

Issue #7753, opened 2023-11-07
Reevaluate the impact of requiring metadata entries for each class member query


vjovanov:
### TL;DR

The current reflection in Native Image requires metadata entries for each query on `java.lang.Class` instances. For example, calling `clazz.getDeclaredMethods()` will require a JSON entry `"allDeclaredMethods": true"` for that class. This fine-grained approach can be cumbersome as it yields many entries in reflection metadata.

We will investigate the impact of including metadata of all `java.lang.Class` members for classes registered for reflection. This would include names and signatures of all methods, properties of all fields, and all annotations. 

The impact will be evaluated on, Spring Petclinic, Micronaut Shopcart, and a few prominent metadata-repository projects. The following metrics will be taken into account: Max RSS, Startup time, Image Size, and the number of required metadata entries.

### Goals
1. Evaluate if we can reduce the amount of required metadata without noticeably increasing the image size.
2. Improve statistics collection for reflection metadata.
 
### Non-goals
1. Change reflection metadata to bloat the produced native images. 
2. Change JNI, Serialization, or dynamic proxies.
3. Include bodies of all methods into the reflection metadata.

loicottet:
Impact of adding all classes, methods and constructors (no fields) from registered classes as queried (not invoked):

Micronaut shopcart:

- Image size: +1.5-2% (avg 1.7%)
- Startup: +0-5% (avg 2.4%, all within the margin of error)
- Max RSS ±2% (avg. 0.3%)

Spring petclinic:

- Image size: +1-1.5% (average 1.1%)
- App startup: +5-18% (average 9.6%)
- Max RSS: ±1.5% (average +0.5%)

The startup time may have been affected by some methods added as reachable to enable running the benchmark with the additional configuration.

Evaluation of the impact on configuration file entries will follow soon. Evaluating the impact of including all elements as invoked/accessed will take more time, if it is at all possible, since the analysis and program behaviour will be affected in more drastic ways.

vjovanov:
We can do that as well, @loicottet can you share the code when ready on a PR and notify us here? 

The reason we did not evaluate Quarkus is simply because Quarkus uses almost no reflection. We have the tika benchmark in our set of benchmarks, so we should report that here as well.

maxandersen:
I would expect micronaut and quarkus would be similar in amount of reflections ?

[...]

PR #9043, opened 2024-06-04
[GR-50385] Include all reflection queries for registered classes

graalvmbot:


zakkak:
Hi @loicottet, could you please add a description to the PR?

If I am not mistaken this is related to https://github.com/oracle/graal/issues/7753. It also seems to bypass the proposal in https://github.com/oracle/graal/issues/7962, is that correct?

The reason this was opened and merged that fast I assume was to get it in before the [Rampdown Phase One (aka feature freeze)](https://openjdk.org/projects/jdk/23/), but that leaves us with no time to properly test it and seems risky for a change in the defaults (that we don't seem to have a way to revert through an option). Are there any plans on further patches on top of it that would allow us to avoid this new behavior if needed?

cc @maxandersen

zakkak:
> It also seems to bypass the proposal in #7962, is that correct?

Correcting myself. The proposal in #7962 (to add a new way to register classes that will trigger automatic registration of all metadata) seems to have been implemented in https://github.com/oracle/graal/pull/8369 and this PR seems like a follow up.

loicottet:
The description of the PR should have been copied when mirroring, sorry about that. This PR's point is to "flip the switch" on #8369, so that using `"type"` actually registers all reflective queries having no impact on reachability.

zakkak:
> The description of the PR should have been copied when mirroring, sorry about that.

Thanks for the prompt reply @loicottet. It looks like this is broken in general (I see descriptions missing in most if not all recent mirrored PRs), could you please have someone look into it?

PR #9137, opened 2024-06-18
[GR-54502] Update mx version.

graalvmbot:


common.json
```diff
@@ -4,7 +4,7 @@
     "Jsonnet files should not include this file directly but use ci/common.jsonnet instead."
   ],
 
-  "mx_version": "7.25.14",
+  "mx_version": "7.26.0",
```
zakkak, line 7: Hi @ansalond 

mx 7.26.0 does not appear upstream

https://github.com/graalvm/mx/blob/2fae8cfc6610e827d28eac075819b3d7e9148751/src/mx/_impl/mx.py#L18176

AFAIK there were some mirroring issues, could you please have a look?

Issue #9199, opened 2024-06-26
Images with JFR will not build due to out of date substitution


roberttoyonaga:
**Describe the issue**
Due to [a recent JDK change](https://github.com/openjdk/jdk/pull/19799/files#diff-b7f754817501bf15b4cb8004fc066e46fbd97b30b15ef406aa09fd00ca2e5f14L51), images that include JFR will not build with GraalVM build from the latest LabsJDK. 

This is because of the now out-of-date substitution: https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/Target_jdk_jfr_internal_JVM_ChunkRotationMonitor.java#L31

PR #9202, opened 2024-06-26
Fix JFR substitutions for JDK Compatibility

roberttoyonaga:
Related Issue: https://github.com/oracle/graal/issues/9199

[A recent JDK change](https://github.com/openjdk/jdk/pull/19799) has caused some JFR substitutions to go out-of-date. Native Image binaries with JFR currently will not build with the latest LabsJDK.  This PR fixes those substitutions.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDK23OrEarlier.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDK23OrEarlier.java
@@ -1,6 +1,5 @@
 /*
- * Copyright (c) 2023, 2023, Oracle and/or its affiliates. All rights reserved.
- * Copyright (c) 2023, 2023, Red Hat Inc. All rights reserved.
+ * Copyright (c) 2024, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -23,12 +22,15 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+package com.oracle.svm.core.jdk;
 
-package com.oracle.svm.core.jfr;
+import jdk.graal.compiler.serviceprovider.JavaVersionUtil;
 
-import com.oracle.svm.core.annotate.TargetClass;
-import com.oracle.svm.core.jdk.JDK23OrEarlier;
+import java.util.function.BooleanSupplier;
 
-@TargetClass(className = "jdk.jfr.internal.JVM$ChunkRotationMonitor", onlyWith = {HasJfrSupport.class, JDK23OrEarlier.class})
-public final class Target_jdk_jfr_internal_JVM_ChunkRotationMonitor {
+public class JDK23OrEarlier implements BooleanSupplier {
+    @Override
+    public boolean getAsBoolean() {
+        return JavaVersionUtil.JAVA_SPEC <= 23;
+    }
 }
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/Target_jdk_jfr_internal_JVM.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/Target_jdk_jfr_internal_JVM.java
@@ -42,6 +42,7 @@ import com.oracle.svm.core.annotate.TargetElement;
 import com.oracle.svm.core.heap.PhysicalMemory.PhysicalMemorySupport;
 import com.oracle.svm.core.jdk.JDK22OrLater;
 import com.oracle.svm.core.jdk.JDK23OrLater;
+import com.oracle.svm.core.jdk.JDK24OrLater;
 import com.oracle.svm.core.jfr.traceid.JfrTraceId;
 import com.oracle.svm.core.util.VMError;
 import com.oracle.svm.util.ReflectionUtil;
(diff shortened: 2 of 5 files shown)


roberttoyonaga:
@zapster Can you please have a look at this one when you have time?

roberttoyonaga:
Looks like the GHA gates are failing due to:
```
1. ERROR in /home/runner/work/graal/graal/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/Target_jdk_jfr_internal_JVM.java (at line 202)
	/** See {@link JVM#nanosNow}. */
Compiling com.oracle.svm.core with ecj-daemon(JDK 21): Failed due to error: -1
	                   ^^^^^^^^
Javadoc: nanosNow cannot be resolved or is not a field
----------
----------
2. ERROR in /home/runner/work/graal/graal/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/JavaMonitorWaitEvent.java (at line 39)
	import jdk.jfr.internal.HiddenWait;
	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
The import jdk.jfr.internal.HiddenWait cannot be resolved
----------
3. ERROR in /home/runner/work/graal/graal/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/JavaMonitorWaitEvent.java (at line 43)
	if (HasJfrSupport.get() && obj != null && !HiddenWait.class.equals(obj.getClass())) {
	                                           ^^^^^^^^^^
HiddenWait cannot be resolved to a type
----------
3 problems (3 errors)
```

Maybe the GHAs are not using the latest LabsJDK?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/Target_jdk_jfr_internal_JVM.java
```diff
@@ -198,6 +199,12 @@ public static long getTicksFrequency() {
         return JfrTicks.getTicksFrequency();
     }
 
+    @Substitute
+    @TargetElement(onlyWith = JDK24OrLater.class)
+    public static long nanosNow() {
+        return JfrTicks.currentTimeNanos();
```
zapster, line 205: Is the accuracy of this high enough? This method only offers milliseconds resolution.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDK23OrEarlier.java
```diff
@@ -0,0 +1,36 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.core.jdk;
+
+import jdk.graal.compiler.serviceprovider.JavaVersionUtil;
+
+import java.util.function.BooleanSupplier;
+
+public class JDK23OrEarlier implements BooleanSupplier {
```
zapster, line 31: We will never need this. On master, we either support JDK 21 or JDK 24, so `JDK21OrEarlier` and `JDK23OrLater` is sufficient. I plan to rename `JDK23OrLater` to `JDK24OrLater` or maybe even `JDKLatest` to avoid such renamings in the future.

zapster:
Thanks @roberttoyonaga for your contribution! We have a similar adoption of this change in our JDK update PR #9181. In general, it always makes sense to look for such JDK update PRs when it comes to recent JDK change adoptions and potentially start discussions there.

In our adoption, we also ported the `javaTimeSystemUTC` method, because it has a higher resolution.  (See also my inline comment.) From the only existing usage in `jdk.jfr.internal.MetadataRepository#awaitEpochMilliShift`, having only millisecond resolution seems fine, which makes me think we should rather use your approach. However, I'm worried that `JVM.nanosNow()` might be used by other methods in the future, which require sub-millisecond accuracy. Such new usages would be hard to detect. What do you think?

cc @christianhaeubl

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDK23OrEarlier.java
roberttoyonaga: >Thanks @roberttoyonaga for your contribution! We have a similar adoption of this change in our JDK update PR https://github.com/oracle/graal/pull/9181. In general, it always makes sense to look for such JDK update PRs when it comes to recent JDK change adoptions and potentially start discussions there.

@zapster Ok I wasn't aware of the updates PR. I'll remember for that for the future. 

 >In our adoption, we also ported the javaTimeSystemUTC method, because it has a higher resolution. (See also my inline comment.) From the only existing usage in jdk.jfr.internal.MetadataRepository#awaitEpochMilliShift, having only millisecond resolution seems fine, which makes me think we should rather use your approach. However, I'm worried that JVM.nanosNow() might be used by other methods in the future, which require sub-millisecond accuracy. Such new usages would be hard to detect. What do you think?

I think porting javaTimeSystemUTC is good. Yes, I think it's fine now, but it would be hard to catch problems if it starts being used for nanosecond accuracy in the future. We should go with your approach.  Thanks!

roberttoyonaga:
Closing this PR because the implementation is already done by another.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDK23OrEarlier.java
roberttoyonaga: @zapster Actually, on second thought, I think it's important that the chunk timestamps use the same clock as `nanosNow()`. See [the comment here](https://github.com/openjdk/jdk/pull/19799/files#diff-b7f754817501bf15b4cb8004fc066e46fbd97b30b15ef406aa09fd00ca2e5f14R174).  We can keep using `javaTimeSystemUTC`, but I think we ought to also change [how we timestamp chunks](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrChunkFileWriter.java#L166),  probably by changing the clock that backs `JfrTicks.currentTimeNanos()` to `javaTimeSystemUTC`.

roberttoyonaga: I opened a new PR for this: https://github.com/oracle/graal/pull/9210

zapster: sounds reasonable, thank you

Issue #9201, opened 2024-06-26
[GR-54994] Avoid duplicate JFR ThreadSleep events


roberttoyonaga:
**Describe the issue**
Native image JFR is now emitting duplicate ThreadSleep events. ThreadSleep events are being emitted from [borrowed JDK java-level code](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Thread.java#L476) **and** also [in SubstrateVM code](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/PlatformThreads.java#L991). This results in 2 events with the same timestamp and almost identical stacktraces.

We need to remove emission of ThreadSleep events from SVM code.

Example below:

![image](https://github.com/oracle/graal/assets/35396843/f2e9dd6f-f0ad-4adf-a631-bd1407cc40c6)

PR #9210, opened 2024-06-27
[GR-54994] Remove duplicate ThreadSleep JFR events and use nanosNow() to timestamp JFR chunks.

roberttoyonaga:
Related issues: https://github.com/oracle/graal/issues/9201 and  https://github.com/oracle/graal/issues/9199

This pull request:

1. Updates JfrTicks to use the new `PlatformTimeUtils.nanosNow()` method. This ensures that [borrowed java-level JDK code](https://github.com/openjdk/jdk/pull/19799/files#diff-c5bdd08f039200bb96baffb6b5ad5716d506c0a53e1f768bf18cb1d6570a4027R340) that uses `JVM.nanosNow()` is using the same clock that we timestamp chunk files with. See [the comment here](https://github.com/openjdk/jdk/pull/19799/files#diff-b7f754817501bf15b4cb8004fc066e46fbd97b30b15ef406aa09fd00ca2e5f14R174). Currently we use `System.currentTimeMillis()` to timestamp chunks, which is a different clock and only has ms precision. This might cause inconsistencies if JDK code uses `JVM.nanosNow()` and compares it to times retrieved from `System.currentTimeMillis()` .  
2.  Removes the duplicate emission of `jdk.ThreadSleep` events from SVM code. Previously the events were being emitted from borrowed Java-level JDK code and then again from SubstrateVM code. 

Before
image

After: 
image

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrTicks.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrTicks.java
@@ -26,6 +26,7 @@ package com.oracle.svm.core.jfr;
 
 import com.oracle.svm.core.Uninterruptible;
 import com.oracle.svm.core.util.TimeUtils;
+import com.oracle.svm.core.util.PlatformTimeUtils;
 
 /**
  * Utility class to manage ticks for event timestamps based on an initial start point when the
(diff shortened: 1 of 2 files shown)


christianhaeubl:
Thanks, this PR will be merged in the next few days.

PR #9211, opened 2024-06-28
[GR-54918] Rename JDK23OrLater to JDKLatest and remove obsolete predicates

graalvmbot:
Currently, exactly two JDK releases are supported: JDK Latest (currently 24) and JDK 21. All other version checks are not needed and cause confusion. Thus, this PR does the following:

* replace `JDK22OrEarlier` with `JDK21OrEarlier`
* replace `JDK22OrLater` with `JDK23OrLater`
* rename `JDK23OrLater` to `JDKLatest`

`JDKLatest` is implemented in a way that we don't need to update the condition for every JDK version bump (boils down to `JAVA_SPEC > 21`).

There are also predicates for JDK 11 and JDK 17. Those are deprecated and unused, but kept for backwards compatibility reasons.


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKLatest.java
```diff
@@ -28,9 +28,13 @@
 
 import jdk.graal.compiler.serviceprovider.JavaVersionUtil;
 
-public class JDK22OrEarlier implements BooleanSupplier {
+/**
+ * Denotes the latest supported JDK version. It corresponds to the highest key in the
+ * {@code JVMCI_MIN_VERSIONS} map in {@link jdk.graal.compiler.hotspot.JVMCIVersionCheck}.
```
zakkak, line 33: Would it make sense to move this map in a package that is accessible here and actually reference the highest key instead of hardcoding the check to `21`?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKLatest.java
```diff
@@ -28,9 +28,13 @@
...
+ */
+public class JDKLatest implements BooleanSupplier {
     @Override
     public boolean getAsBoolean() {
-        return JavaVersionUtil.JAVA_SPEC <= 22;
+        return JavaVersionUtil.JAVA_SPEC > 21;
```
zakkak, line 38: Nitpick: I understand that `21` is used here to avoid having to update the code with each STS JDK release, but this doesn't match the javadoc, nor represents the latest JDK.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKLatest.java
zapster: The package is accessible, it is just the field that is `private`. The LTS version change so seldomly that IMHO it is fine to have a few references to it.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/JDKLatest.java
zapster: From my perspective, that is just an implementation detail. `21` is the most stable version we have and it only changes once every few years.

Issue #8915, opened 2024-05-13
Add support for JCMD and the Attach API


roberttoyonaga:
### TL;DR
Support executing commands issued from`jcmd` and support connection to SubstrateVM through the Attach API.

### Motiviation
Native Image currently doesn't support `jcmd`. Currently, we work around this by relying on catching and handling Linux signals for many things.  Examples include, generating heap dumps, dumping threads, generating NMT reports. JFR also cannot dynamically be controlled at runtime (unless JMX is used).  Signals are not ideal because there are a limited number of usable signals. They are also not ideal because arguments cannot be specified.  `jcmd` would solve these problems because it would allow for an unlimited number of commands and also allows for arguments. 

Implementing the Attach API also opens the door to the possibility of other Java tools like `jmap`, `jstack`, etc being supported in the future. 

### Goals
**Describe the solution you'd like.**
- Implement the Attach API in SubstrateVM
- Lay groundwork for issuing commands via `jcmd`. Focus on NMT, heap dumps, and JFR commands for now. 

### Non-Goals
- Support Java agents using dynamic attach.
- Support `jmap`, `jstack`, or other tools that use the Attach API
- Support all `jcmd` commands. The goal is to put the piping in place and support a few commands to start with.

roberttoyonaga:
I'm weighing the benefits of implementing this as uninterruptible vs interruptible. The the implementation in OpenJDK is in native code: " [The rationale for a pure native implementation is to be able to execute diagnostic commands even in critical situations like an out-of-memory condition.](https://openjdk.org/jeps/137) "  

However, implementing this interruptibly would lower complexity by  allowing for using `java.nio.channels.SocketChannel` to handle unix domain sockets and `jdk.internal.misc.Signal` for handling SIGQUIT (which kicks off the Attach API handshake).

Currently, handling signals for dumping thread stacks, or head dumps is interruptible. I am leaning toward implementing the Attach API and JCMD support interruptibly in Java. Are there any concerns about this?

christianhaeubl:
> The implementation in OpenJDK is in native code.

If OpenJDK implements a feature fully in native code, it is pretty likely that we will need uninterruptible code (at least in the long run - we got bitten by that a couple times in the past). When implementing a prototype, it is definitely easier to use interruptible code though. Maybe, you can roughly sketch the steps that the attach API needs to execute, then it is easier to estimate if we have a reasonable chance with interruptible code.

roberttoyonaga:
[...]

PR #9232, opened 2024-07-03
[GR-56601] Add support for JCMD and the Attach API

roberttoyonaga:
### Summary
This pull request adds initial support for the Attach-API and JCMD.  Specifically, other processes may now use the "attach" mechanism to communicate with Native Images using JCMD. This has been implemented as an optional feature that is excluded from the image by default (similar to JFR or heap dumping). 

For now the following JCMD command types are supported: 

- GC.heap_dump
- JFR.start
- JFR.stop
- JFR.check
- JFR.dump
- Thread.dump_stacks
- Thread.dump_to_file
- VM.native_memory
- VM.runtime_compilation
- help

Support for more commands should be straightforward to add in the future. 

Related issue: https://github.com/oracle/graal/issues/8915

### Motivation
Native Image has a growing number of features such as NMT, heap dumps, thread dumps, and compilation dumps that rely on handling Unix signals to perform operations requested at runtime. This is not ideal because there is a limited number of Unix signals that are available to be used. It is also not ideal because arguments cannot be easily used with signals. Supporting JCMD would be a better alternative to signals because we can support an unlimited number of commands and also accept arguments. 

JFR in OpenJDK heavily uses JCMD to accept user requests at runtime. Currently, there is no way to stop/start/dump recordings spontaneously at runtime (apart from using the remote JMX feature - which has significant overhead with respect to image size). This is inconvenient if a user wants to diagnose a problem without shutting down or restarting their application.  For example, if a user wants to access JFR recordings from their application, they need to terminate their application so that a JFR dump is created. With JCMD they are able to dump recordings whenever they desire without needing to restart their entire application.

### Usage
1. Build this PR with `mx build`
2. Build your executable with the`attach` monitoring option. This will automatically include JCMD as well. Make sure your application runs long enough for you to enter some requests through JCMD.
`mx native-image --enable-monitoring=nmt,jfr,jvmstat,threaddump,attach YourApp`
3. Run your app. Then find its PID.
`./yourapp`

4. In another terminal:
`$JAVA_HOME/bin/jcmd <your app's PID> help`
`$JAVA_HOME/bin/jcmd <your app's PID> help JFR.stop`
`$JAVA_HOME/bin/jcmd <your app's PID> help help`
`$JAVA_HOME/bin/jcmd <your app's PID> VM.native_memory summary`

### Limitations

Attach-API only works on posix so far. This is not a regression since headp dump, thread dump, compilation dump, nmt dump, and JFR, all don't work on Windows anyway (since they require unix signals to operate).

The Windows implementation will be largely independent of code in this posix implementation. So there will be no need to go back and rework anything here. It would just be additional.

### Other notes
- JfrArgumentParser.java is mostly just code factored out of JfrManager so that it can be reused by the JFR DCMDs.
- DumpThreadStacksSupport.java is mostly just code factored out from DumpThreadStacksOnSignalFeature.java
- Error handling in JfrManager has been changed so that we don't throw RuntimeExceptions when handling DCMDs with invalid input. Instead we just throw DcmdExceptions and fail gracefully. Avoid bad input crashing the Attach server.


roberttoyonaga:
The GHA `GraalVM Gate / /sulong style,fullbuild,sulongBasic JDK21 (pull_request) `  is failing due to some Truffle copyright headers failing the style check.  This seems unrelated to the code in this PR.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/dcmd/DcmdFeature.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.graalvm.nativeimage.ImageSingletons;
+
+import com.oracle.svm.core.VMInspectionOptions;
+import com.oracle.svm.core.feature.AutomaticallyRegisteredFeature;
+import com.oracle.svm.core.feature.InternalFeature;
+import com.oracle.svm.core.thread.ThreadDumpToFileDcmd;
+
+@AutomaticallyRegisteredFeature
+public class DcmdFeature implements InternalFeature {
```
christianhaeubl, line 37: Can diagnostic commands also be triggered via some other way than the Attach API? If not, then I will merge the `attach` and `dcmd` packages.

christianhaeubl (commented):
Thanks for the PR, I started reviewing and integrating it so please don't push any further changes for now.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/dcmd/DcmdParseException.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.svm.core.dcmd;
+
+/** For validation of input arguments. */
+public class DcmdParseException extends Exception {
```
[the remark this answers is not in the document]
roberttoyonaga: Yes I agree, `IllegalArgumentException` is probably good enough here

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/dcmd/JfrDumpDcmd.java
```diff
@@ -0,0 +1,314 @@
...
+import static com.oracle.svm.core.jfr.JfrArgumentParser.parseJfrOptions;
+import static com.oracle.svm.core.jfr.JfrArgumentParser.parseMaxSize;
+import static java.util.concurrent.TimeUnit.DAYS;
+import static java.util.concurrent.TimeUnit.HOURS;
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+public class JfrDumpDcmd extends AbstractDcmd {
```
christianhaeubl, line 70: Why can't we use `jdk.jfr.internal.dcmd.DCmdDump` instead?

roberttoyonaga: Maybe we could. But `jdk.jfr.internal.dcmd.DCmdDump` expects to be provided with a `jdk.jfr.internal.dcmd.ArgumentParser` and also writes back results differently (`reportOperationComplete`). In OpenJDK, each individual DCMD is responsible for sending back their own specific responses.  In this PR, only the AttachListener classes know about the sockets and the DCMDs just return a simple string for the AttachListener to deal with.

PR #9250, opened 2024-07-05
[GR-54128] Re-import libsvm_container code and add change detection annotations

graalvmbot:
This change on top of https://github.com/oracle/graal/pull/8989 re-imports the `libsvm_container` code from the JDK and adds annotations to keep track of upstream changes. In particular:

* re-import libsvm_container code from JDK 24+5
* add `@BasedOnJDKFile` annotation for `libsvm_container` files imported from JDK 
* add `mx gate --tags check_libcontainer_annotations` gate job
* add `mx reimport-libcontainer-files --jdk-repo path/to/jdk/open`


zapster:
@jerboaa while this follow-up PR to https://github.com/oracle/graal/pull/8989 is still in progress, it should give you an idea how we gonna deal with keeping the `libsvm_container` up to date. Since that PR was lying around for some time now, I had the opportunity to re-import the code several times and it always was seamless and fast (<1h). With the `@BasedOnJDKFile` annotations we will never miss any changes to the relevant code. (And yes, we already noticed https://github.com/openjdk/jdk/pull/18201 :wink:)

I hope that this will address most of the concerns you shared in https://github.com/oracle/graal/pull/8989#issuecomment-2138179493.

zapster:
(Fun fact: I just noticed the behavior fixed by https://github.com/openjdk/jdk/pull/18201 yesterday when testing this change and was a bit puzzled. So thanks for the fix. :slightly_smiling_face:)

jerboaa:
Thanks for the heads-up.

PR #9257, opened 2024-07-05
[GR-55135] Update labsjdk to 24+5-jvmci-b01

graalvmbot:
Update labsjdk to 24+5-jvmci-b01

* adopt https://github.com/openjdk/jdk/pull/19892


substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestContainerEvent.java
```diff
@@ -47,6 +47,7 @@ public class TestContainerEvent extends JfrRecordingTest {
     @Test
     public void test() throws Throwable {
         Assume.assumeTrue("Container support not enabled or available", Container.isSupported());
+        Assume.assumeTrue("Container detection currently not working GR-55178", false);
```
zapster, line 50: We are missing adoption of https://github.com/openjdk/jdk/pull/18201, which will be done in course of https://github.com/oracle/graal/pull/9250.

PR #9316, opened 2024-07-17
[GR-54333] New guide: Optimize a Native Executable for File Size.

graalvmbot:


docs/reference-manual/native-image/guides/optimize-file-size.md
```diff
@@ -0,0 +1,105 @@
...
+
+2. (Optional) Run the application:
+    ```bash
+    ./target/fortune
+    ```
+    The application will return a random saying. 
+
+3. Check the file size which should be around 13M:
+    ```bash
+    du -sh target/fortune*
```
zakkak, line 49:
```suggestion
    du -h target/fortune
```

Isn't the use of `-s` (and the glob) misleading here? The actual size of the native image is the size of the binary only.

docs/reference-manual/native-image/guides/optimize-file-size.md
```diff
@@ -0,0 +1,105 @@
...
+    16K    target/fortune-1.0-SNAPSHOT.jar
+    9.8M   target/fortune-optimized
+    1.9M   target/fortune-optimized-build-report.html
+    ```
+    The file size decreased from 13M to 9.8M! 
+
+How much the file size can be reduced by the `-Os` option varies between applications, and depends on how much Native Image applies inlining and other optimizations that increase size in the default `-O2` mode.
+
+The build report generated in the previous step, _fortune-optimized-build-report.html_, tells exactly what was included in your native executable.
+It is an HTML file that you can open in a regular web browser.
```
zakkak, line 91: Why not generate this report for both cases, so that readers can compare them?

olyagpl: It makes sense, I agree. It'd be good to update the public demo first. I'll follow up on this suggestion in the next PR. Thanks for reviewing!

PR #9359, opened 2024-07-24
Improve Logging Accuracy in onFinalize Method

logresearch:
This PR addresses an issue in the onFinalize method where the logging statement could potentially mislead the reader into believing that the execution was successful, even in the presence of an error. The logging sequence has been revised to ensure that the log message accurately reflects the method’s state.

## Description
Previously, the onFinalize method logged “Executed successfully: “ before checking for errors. This sequence could result in misleading logs when errors are present, as the success message would still be logged.

### Changes Made

- The check for errors (error != null) and the corresponding AssertionError throw now precede the success logging.
- The log message “Executed successfully: “ is printed only if no errors are detected, ensuring that the success message aligns with the method’s actual state.

### Impact
This change ensures that log messages accurately represent the state of the method, providing clearer and more reliable logging output for error tracking and debugging purposes.

### DIFF
--- a/truffle/src/com.oracle.truffle.tck.instrumentation/src/com/oracle/truffle/tck/instrumentation/DebugALot.java
+++ b/truffle/src/com.oracle.truffle.tck.instrumentation/src/com/oracle/truffle/tck/instrumentation/DebugALot.java
@@ -124,12 +124,12 @@ public class DebugALot extends TruffleInstrument implements SuspendedCallback {
 
     @Override
     protected void onFinalize(Env env) {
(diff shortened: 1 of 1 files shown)


fniephaus:
Thanks for your contribution. Could you please sign the OCA so we can start integrating this?

fniephaus (approved):

entlicher:
Your change is in: https://github.com/oracle/graal/commit/978c81e6180874ee7676c2f13d30eccff82a1e25
Thanks for your contribution.

PR #9505, opened 2024-08-14
[GR-57310] NMT simplified virtual memory tracking.

roberttoyonaga:
### Summary
This is a simplified version of https://github.com/oracle/graal/pull/8630 
This PR adds virtual memory tracking, but makes the assumption that virtual memory is used neatly (no overlapping committed regions etc).  Please see the javadoc in `NativeMemoryTracking` for more info.

**Notes:**
There's 3 approaches we could take with this simplified version. 

1. The approach in this PR. This is a middle ground between 2 and 3. Instrument at the caller level (ie in `HeapChunkProvider`). This makes the caller responsible for correct instrumentation.
2. Get rid of "track" methods and `NmtVirtualMemoryInfo`. Instead, directly query `HeapAccounting` whenever we need to report virtual memory. This is probably less extensible than 1 or 3. 
3. Lower level instrumentation of `VirtualMemoryProvider` implementations. Similar to the [original PR](https://github.com/oracle/graal/pull/8630 ). This would ensure we don't miss any allocations if virtual memory becomes used in other places in the future.  However, I don't think this is necessary right now since virtual memory is used simply in SVM currently. We can avoid instrumenting in platform specific code.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapChunkProvider.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/HeapChunkProvider.java
@@ -24,6 +24,7 @@
  */
 package com.oracle.svm.core.genscavenge;
 
+import com.oracle.svm.core.VMInspectionOptions;
 import org.graalvm.nativeimage.Platform;
 import org.graalvm.nativeimage.Platforms;
 import org.graalvm.word.Pointer;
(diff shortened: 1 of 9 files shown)


roberttoyonaga:
Realized I needed to make a small change to consider all image heap regions when determining the size. I'm done making changes now.

christianhaeubl:
Thanks for the PR. I had to change a few things, so I opened another PR based on your changes: https://github.com/oracle/graal/pull/9549.

PR #9549, opened 2024-08-22
[GR-57310] Add Native Memory Tracking for the image heap and the Java heap.

graalvmbot:
Based on https://github.com/oracle/graal/pull/9505.


roberttoyonaga (approved):
This looks good to me! Thank you for working on integrating this.

roberttoyonaga:
Just a question, in the original PR, I iterate through all the [`ImageHeapInfo` and sum up the used sizes](https://github.com/oracle/graal/pull/9505/files#diff-5529ee695c43822faedba53ac065f9a6f14a89b1f6253178c66a78318257dbd2R480-R500). Is the reason for the chained `ImageHeapInfo` related to Layered Native Images? Is there a functional difference between how we get the Image Heap sizes in this PR and the original one?

christianhaeubl:
> Is the reason for the chained `ImageHeapInfo` related to Layered Native Images

Yes.

> Is there a functional difference between how we get the Image Heap sizes in this PR and the original one?

Yes, there is a bit of a difference. Your code computes the actually used number of image heap bytes (i.e., it only takes the parts into account that are really filled with objects). This number is always smaller than the actually committed memory (e.g., certain image heap parts are empty/unused and only exist for alignment reasons).

Issue #9536, opened 2024-08-20
Large quantities of virtual threads may cause JFR performance issues


roberttoyonaga:
**Describe the issue**
In SubstrateVM, JFR eagerly registers virtual threads when they are mounted and during a chunk rotation. This means large numbers of virtual threads will result in a lot of work, even if the threads don't end up emitting any events. In Hotspot, virtual threads are registered lazily upon event emission.  

**Steps to reproduce the issue**
- Create a large number of virtual threads
- Configure JFR to emit fewer events.

**Describe GraalVM and your environment:**
The problem is in the latest GraalVM master and all previous versions that support virtual threads in JFR.

**Solution**
Adopt a lazy approach to registration similar to Hotspot.

PR #9570, opened 2024-08-26
[GR-57064] Lazy virtual thread JFR registration and epoch generation caching

roberttoyonaga:
### Summary
In Hotspot, JFR registers virtual threads lazily when they emit an event. Currently, in SubstrateVM, we register virtual threads eagerly whenever they are mounted. This can be problematic if many virtual threads mount/unmount while not emitting many events. This is especially bad since SubstrateVM uses a [global mutex](https://github.com/oracle/graal/blob/7345dd6076979a24fa0f39a2cf52a32f40524e83/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrThreadRepository.java#L116) to protect the "thread constant pool repository".  This mutex must be  acquired every time we register **or even just check** thread registration status. In Hotspot, there is no such global "thread constant pool repository", and no global synchronization -- instead a thread "checkpoint event" is written to thread local JFR buffers lazily when needed.  

This PR does 2 things:
1. Make virtual thread registration lazy instead of eager. Register virtual threads only when they need to emit an event, not upon mounting.
2.  Avoid  locking global "thread constant pool repo" mutex every time we need to check virtual thread registration status. Do this by caching the "epoch generation" in the `Target_java_lang_VirtualThread.java` object (similar to [`Target_java_lang_Thread.jfrExcluded`](https://github.com/oracle/graal/blob/7345dd6076979a24fa0f39a2cf52a32f40524e83/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/thread/Target_java_lang_Thread.java#L86). Before locking the thread repo mutex, first compare "epoch generation" in case the virtual thread is already registered. A very similar techniqie is used in Hotspot to avoid writing thread checkpoint events unless necessary. 

#### Results
**Scenario 1**: Many virtual thread mounts/unmounts while not emitting JFR events.
Used [test app 1](https://gist.github.com/roberttoyonaga/ce00cbe0ecd2a7ce0bbf0adf267db5f8). The new lazy method takes ~1105ms to complete, while the old eager approach takes ~1140ms. 
 
**Scenario 2**: Many JFR events emitted concurrently by many virtual threads.
Used [test app 2](https://gist.github.com/roberttoyonaga/fbef15e61bd1c555b249699b83d82e3c). The registration with cached epoch generation checking takes ~42.600s to complete, while registration with global locking takes ~73.650s. 

Related issue: https://github.com/oracle/graal/issues/9536

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
@@ -26,6 +26,7 @@ package com.oracle.svm.core.jfr;
 
 import java.util.List;
 
+import com.oracle.svm.core.SubstrateUtil;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.IsolateThread;
 import org.graalvm.nativeimage.Platform;
@@ -42,6 +43,7 @@ import com.oracle.svm.core.jfr.oldobject.JfrOldObjectProfiler;
 import com.oracle.svm.core.jfr.oldobject.JfrOldObjectRepository;
 import com.oracle.svm.core.jfr.sampler.JfrExecutionSampler;
 import com.oracle.svm.core.jfr.throttling.JfrEventThrottling;
+import com.oracle.svm.core.jfr.traceid.JfrTraceIdEpoch;
 import com.oracle.svm.core.log.Log;
 import com.oracle.svm.core.sampler.SamplerBufferPool;
 import com.oracle.svm.core.sampler.SamplerBuffersAccess;
@@ -49,6 +51,7 @@ import com.oracle.svm.core.sampler.SamplerStatistics;
 import com.oracle.svm.core.sampler.SubstrateSigprofHandler;
 import com.oracle.svm.core.thread.JavaThreads;
 import com.oracle.svm.core.thread.JavaVMOperation;
+import com.oracle.svm.core.thread.Target_java_lang_VirtualThread;
 import com.oracle.svm.core.thread.VMThreads;
 import com.oracle.svm.core.util.VMError;
 
@@ -303,6 +306,7 @@ public class SubstrateJVM {
     @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
     public static long getThreadId(Thread thread) {
         if (HasJfrSupport.get()) {
+            maybeRegisterVirtualThread(thread);
             return JavaThreads.getThreadId(thread);
         }
         return 0;
@@ -311,9 +315,27 @@ public class SubstrateJVM {
     @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
     public static long getCurrentThreadId() {
         if (HasJfrSupport.get()) {
+            maybeRegisterVirtualThread(Thread.currentThread());
             return JavaThreads.getCurrentThreadId();
         }
         return 0;
+    }
+
+    /**
+     * Register virtual threads if they aren't registered already. Platform threads are registered
+     * eagerly when started and at chunk rotations.
+     */
+    @Uninterruptible(reason = "Epoch should not change while checking generation.")
+    private static void maybeRegisterVirtualThread(Thread thread) {
+        // Do quick preliminary checks to avoid global locking unless necessary.
+        if (JavaThreads.isVirtual(thread)) {
+            Target_java_lang_VirtualThread tjlv = SubstrateUtil.cast(thread, Target_java_lang_VirtualThread.class);
+            int currentEpochGen = JfrTraceIdEpoch.getInstance().currentEpochGeneration();
+            if (tjlv.jfrGeneration != currentEpochGen) {
+                getThreadRepo().registerThread(thread);
+                tjlv.jfrGeneration = currentEpochGen;
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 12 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
```diff
@@ -311,11 +315,29 @@ public static long getThreadId(Thread thread) {
...
         }
         return 0;
     }
 
+    /**
+     * Register virtual threads if they aren't registered already. Platform threads are registered
+     * eagerly when started and at chunk rotations.
+     */
+    @Uninterruptible(reason = "Epoch should not change while checking generation.")
+    private static void maybeRegisterVirtualThread(Thread thread) {
```
christianhaeubl, line 329: Doesn't this break once we leave uninterruptible code?
- we start in epoch 0
- thread A calls `getThreadId(...)`, which registers a `vthread` for epoch 0 in `maybeRegisterVirtualThread`
- thread A leaves the uninterruptible code and gets blocked at a safepoint
- thread B changes the epoch from 0 to 1
- after the safepoint, thread A continues execution and uses the thread id that was returned by `getThreadId(...)` even though the vthread was only registered for epoch 0

One example where I think that this might happen: `JavaMonitorQueuedSynchronizer.ConditionNode.notifierJfrTid`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/traceid/JfrTraceIdEpoch.java
```diff
@@ -43,6 +43,7 @@ public class JfrTraceIdEpoch {
     private static final long EPOCH_1_BIT = 0b10;
 
     private boolean epoch;
```
christianhaeubl, line 45: `epoch` can probably be removed (i.e., can be computed from the least significant bit of `epochGeneration`).

christianhaeubl (commented):
Thanks for the PR, I added a few comments.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
roberttoyonaga: I think the situation you describe is actually ok.  Once the epoch changes, new event emissions from "thread A" will fail the epoch generation check, and registration will be re-done. Since `getThreadID()` is called for every event emission, the important thing is that gathering event info with `getThreadID()` and the writing of the event data to the JFR buffer are done within the same block of uninterruptible code, which does indeed happen (ex. all `emit0` methods). 

Calls to `getThreadID()` outside of the event emission path (ex. with `JavaMonitorQueuedSynchronizer.ConditionNode.notifierJfrTid`) result in unnecessary registration checks, but shouldn't break anything.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/traceid/JfrTraceIdEpoch.java
roberttoyonaga: Ok, good idea. I'll remove that.

roberttoyonaga:
Truffle gate error seems to be unrelated

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SubstrateJVM.java
christianhaeubl: Ok, so you essentially assume that `getThreadID()` will be called again for thread A once the thread ID is actually used in a JFR event. However, I don't think that this assumption is valid. Here is one example:
- Thread B blocks at `ConditionNode`.
- Thread A notifies thread B and sets `ConditionNode.notifierJfrTid`.
- The JFR epoch changes.
- Once thread B wakes up, it uses the value of `ConditionNode.notifierJfrTid` (i.e., the ID of thread A) when emitting a `JavaMonitorWaitEvent`. However, thread A is not registered for the current epoch.

roberttoyonaga: Ok I see. I mistakenly thought we always used `JfrNativeEventWriter.putThread` when dealing with threads. But in that case we use a the plain TID directly. 

I think we'll have to switch to always using `putThread` whenever dealing with the JFR `Thread` type.  What do you think of this?

roberttoyonaga: The downside is that we'll have to remember to give careful consideration to emitting  JFR `Thread` data. 
I'm not sure there's a better place to put the registration check though. And we'll need to check at some point in order to copy Hotspot and do lazy registration.

roberttoyonaga: I've just pushed a commit to update all the places we write thread IDs as JFR event data to use either `putThread(Thread)` or `putThread(long)`. This should ensure that threads are always registered if they get referenced from event data. The downside is that we need to remember to use `putThread`. 

Another unfortunate thing is that we can only do checks with `putThread(Thread)` because with only the TID we have no way of knowing whether the thread is virtual and cannot access `Target_java_lang_VirtualThread.jfrGeneration`.

christianhaeubl:
Thanks!

I did one pass over the PR and opened https://github.com/oracle/graal/pull/11070. For now, I think it is better/safer if we still register all virtual threads eagerly. At the moment, we have a few JFR events where we only store the thread id. We would need to change all that code so that we keep more information around, which in my opinion is not worth the effort at the moment.

roberttoyonaga:
> I did one pass over the PR and opened https://github.com/oracle/graal/pull/11070. For now, I think it is better/safer if we still register all virtual threads eagerly. At the moment, we have a few JFR events where we only store the thread id. We would need to change all that code so that we keep more information around, which in my opinion is not worth the effort at the moment.

Yes that sounds fine to me as well.  Most of improvement I measured was gained through avoiding global locking, not the lazy registration anyway.

PR #9621, opened 2024-09-05
[GR-57740] Update labsjdk to 24+13-jvmci-b01

graalvmbot:
Update labsjdk to 24+13-jvmci-b01

* adopt https://github.com/openjdk/jdk/pull/20429
* adopt https://github.com/openjdk/jdk/pull/20664
* adopt https://github.com/openjdk/jdk/pull/19213
* adopt https://github.com/openjdk/jdk/pull/20713
* adopt https://github.com/openjdk/jdk/pull/20588


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/Target_java_lang_Module.java
```diff
@@ -51,9 +51,15 @@ public boolean isNativeAccessEnabled() {
     @TargetElement(onlyWith = JDK21OrEarlier.class)
     public native void ensureNativeAccess(Class<?> owner, String methodName);
 
-    @Alias
+    /**
+     * Allow all native access. This is a workaround for JDK-8331671 until we have a better
+     * solution. (GR-57608)
+     */
```
zapster, line 57: fyi @zakkak @jerboaa

jerboaa: @zapster OK. Thanks for the heads-up. I'd have thought `--enable-native-access=<graal-module>` to be added where it needs to be. Like all the `--add-opens/add-exports`. Is that the longer-term plan?

zapster: yes, we do that already. see the changes in the driver code in this pr. the problem with this method is that it unconditionally pulls in all the error handling code. to give you an idea, this results in doubling the number of reachable types in simple `mx helloworld`. the planned solution is to make sure that all the code paths leading here are force inlined so that the module is constant and we can fold the error/warning branch away.

jerboaa: I see.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1323,6 +1324,8 @@ private int completeImageBuild() {
         if (config.modulePathBuild && !finalImageClasspath.isEmpty()) {
             imageBuilderJavaArgs.add(DefaultOptionHandler.addModulesOption + "=ALL-DEFAULT");
         }
+        assert !enableNativeAccessModules.isEmpty();
+        imageBuilderJavaArgs.add("--enable-native-access=" + String.join(",", enableNativeAccessModules));
```
jerboaa, line 1328: @zapster For our builds this produces warnings like the following:

```
WARNING: Unknown module: com.oracle.svm.svm_enterprise specified to --enable-native-access
WARNING: Unknown module: com.oracle.graal.graal_enterprise specified to --enable-native-access
```

Could we get something like https://github.com/jerboaa/graal/commit/74a2e81efb6c71367a41ee6e864d8ac3c4d4e416 added to this PR? That would fix it for us. Thanks!

jerboaa (commented):
This unbreaks our builds with JDK 24+13. Awesome! Could we get the suggested addition added, perhaps so that we don't get the warnings?

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
zapster: Thanks for the suggestion. I don't want to add it to this PR though, as I want to get the update merged asap. I don't want to risk breaking things with last minute changes. Also I'm not 100% sure this is the solution we want. I first want to explore alternatives, e.g., routing all native accesses through a single module and remove all others from the native access list. But this needs more experimentation (and probably discussion).

jerboaa: OK, that's fair enough. Please keep me in the loop if possible. This will fail in our CI.

zapster: will do!

zapster: PR is up #9654

PR #9645, opened 2024-09-10
[GR-52034] Convert IGV to maven build and update to netbeans 22

graalvmbot:
This updates IGV to the latest netbeans 22 platform which officially supports JDK 17 through 22.  This also converts the build to use maven which means you can use any IDE that supports maven to work on the IGV sources.  The diff is huge because of renames so It's not easily reviewable except as a patch I think.  I've tested it fairly well on my local machine but it would be nice to have some more guinea pigs.  Checking out the branch and launching `mx igv` in the compiler suite should work to test it out.


tkrodriguez:
@JaroslavTulach here the workign update to the latest netbeans platform.  Maybe you can try it out and see if it resolves the problems you were having.

Also, the switch to maven has exposed some problems with the mx support in NetBeans interfering with open the maven based project.  When I try to open IdealGraphVisualizer as a maven project from NetBeans 22, it properly recognizes the main maven pom.xml but it seems to think the modules are mx projects.
opendialog1
and opening the project shows that it's missing all the modules.

projects1

disabling the java-mx-project by editing `$HOME/Library/Application Support/NetBeans/22/config/Modules/org-netbeans-modules-java-mx-project.xml`allows it to open correctly.

projects2

Any suggestions for working around this besides disabling that module?

JaroslavTulach:
Wow, 1135 changed files. That's out of any review possibilities.

> use maven

Good step. Using Maven is the best way to develop a "NetBeans Platform" application with any IDE of one's choice. Good move!

JaroslavTulach:
> @JaroslavTulach here the workign update to the latest netbeans platform. Maybe you can try it out and see if it resolves the problems you were having.

My biggest problem was too old APIs I had to use in my module in order to work in IGV. Getting NetBeans 22 APIs is great. Btw. version 23 should have already been released or is going to be out in few weeks.


> Also, the switch to maven has exposed some problems with the mx support in NetBeans interfering with open the maven based project. When I try to open IdealGraphVisualizer as a maven project from NetBeans 22, it properly recognizes the main maven pom.xml but it seems to think the modules are mx projects.

We need to fix NetBeans `java.mx.projects` support! 

> disabling the java-mx-project by editing `$HOME/Library/Application Support/NetBeans/22/config/Modules/org-netbeans-modules-java-mx-project.xml`allows it to open correctly.

There is a special support in `java.mx.projects` module to recognize "visualizer" modules. More related issue is https://github.com/apache/netbeans/pull/2789 that code needs to be updated to recognize the new "Maven structure". 

> Any suggestions for working around this besides disabling that module?

Please don't workaround. Submit a patch for NetBeans to fix the issue.

JaroslavTulach:
Btw. there is detection of [.mxignore](https://github.com/apache/netbeans/blame/0374e1f4e791393cdcebeba4580177aee3154332/java/java.mx.project/src/org/netbeans/modules/java/mx/project/SuiteFactory.java#L85) file - maybe you can place it (into the root?) of maven project and convince the `java.mx.projects` module to ignore the Maven projects.

visualizer/Readme.md
```diff
@@ -3,7 +3,8 @@
 ## Prerequisites
 
 - JDK 21 is the recommended Java runtime platform for IGV, but any release between 17 and 22 is
-  supported by the NetBeans 22 platform.
+  supported by the NetBeans 22 platform.  It's recommended that the build is performed using JDK 17
+  so that it can be used on all supported releases.
```
JaroslavTulach, line 7: Why shouldn't the build be supported on all releases? If you use compile with `--release` `javac` option, it is going to work on older JDKs regardless the JDK you compile with. Should there be a problem with that see [frgaal retrofit compiler](https://frgaal.org/#usage-with-maven). It is easy to configure from a `pom.xml` and guarantees JDK independence.

visualizer/Readme.md
```diff
@@ -15,36 +16,45 @@ Certains kind of edits, like editing the NetBeans module exlusions or editing th
 components, will require using NetBeans.
 
 There is a known issue with opening the Maven based project in NetBeans because of a bug with the
-`mx` project support built into NetBeans itself.`  Because 
+`mx` project support built into NetBeans itself.  The `mx` support seems to interfere with
+recognizing the maven modules so at the moment it's necessary to disable the `java-mx-support`
+module in NetBeans 22.  On the mac this can be done by editing the file `$HOME/Library/Application
```
JaroslavTulach, line 21: Have you tried the `.mxignore` file?

JaroslavTulach: As far as I can say:
```
/graalvm/graal/visualizer/IdealGraphVisualizer$ echo >.mxignore
/graalvm/graal/visualizer/IdealGraphVisualizer$ netbeans-22/bin/netbeans --open .
```
Fixes the problem and I can open the Maven pom projects as all its children.

tkrodriguez: Thanks for figuring this out.  I'll add it to the repo and update the readme.

visualizer/Readme.md
tkrodriguez: I'll adjust the build to target release 17 and remove this language.

PR #9600, opened 2024-09-02
fix: EspressoFinalReference NoSuchMethodError

Weaxs:
bug fix: https://github.com/oracle/graal/issues/9599

```
Exception in thread "main" org.graalvm.polyglot.PolyglotException: com.oracle.truffle.espresso.meta.EspressoError: should not reach here: Error injecting while injecting classes to support finalization in the host (version 17)
	at com.oracle.truffle.espresso.meta.EspressoError.shouldNotReachHere(EspressoError.java:73)
	at com.oracle.truffle.espresso.ref.FinalizationSupport.<clinit>(FinalizationSupport.java:72)
	at com.oracle.truffle.espresso.runtime.EspressoEnv.<init>(EspressoEnv.java:139)
	at com.oracle.truffle.espresso.runtime.EspressoContext.<init>(EspressoContext.java:213)
	at com.oracle.truffle.espresso.EspressoLanguage.createContext(EspressoLanguage.java:192)
	at com.oracle.truffle.espresso.EspressoLanguage.createContext(EspressoLanguage.java:85)
	at com.oracle.truffle.api.LanguageAccessor$LanguageImpl.createEnvContext(LanguageAccessor.java:270)
	at com.oracle.truffle.polyglot.PolyglotLanguageContext.ensureCreated(PolyglotLanguageContext.java:652)
	at com.oracle.truffle.polyglot.PolyglotLanguageContext.ensureInitialized(PolyglotLanguageContext.java:731)
	at com.oracle.truffle.polyglot.PolyglotContextImpl.eval(PolyglotContextImpl.java:1681)
	at com.oracle.truffle.polyglot.PolyglotContextDispatch.eval(PolyglotContextDispatch.java:60)
	at org.graalvm.polyglot.Context.eval(Context.java:402)
	at aaaaaa.ScriptTest.script(ScriptTest.java:138)
	at aaaaaa.ScriptTest.main(ScriptTest.java:51)
	Suppressed: Attached Guest Language Frames (0)
Caused by: java.lang.NoSuchMethodException: no such constructor: com.oracle.truffle.espresso.ref.EspressoFinalReference.<init>(StaticObject,StaticObject,ReferenceQueue)void/newInvokeSpecial
	at java.base/java.lang.invoke.MemberName.makeAccessException(MemberName.java:974)
	at java.base/java.lang.invoke.MemberName$Factory.resolveOrFail(MemberName.java:1117)
[...]

PR #9662, opened 2024-09-11
[GR-57845] Fix symbolic references to StaticObject in ClassAssembler

graalvmbot:


gilles-duboscq:
This is the integration of #9600

PR #9670, opened 2024-09-12
[GR-56098] Backport to 23.1: reachability-metadata.json parser

loicottet:
The goal of this PR is to enable the 23.1 version of GraalVM to correctly parse reachability metadata files emitted by GraalVM 24.1 and future releases. These files are parsed and matched to the existing capabilities of GraalVM 23.1 on a best effort basis, meaning that existing workflows won't be disrupted, but features introduced in subsequent versions will be ignored. For example, GraalVM 24.1 can register members (methods, fields and constructors) of proxy classes for reflection, which was not possible before, and does it via the reflection metadata instead of a separate proxy metadata file. When encountering such metadata, the 23.1 parser will now correctly register the proxy class for runtime instantiation, but won't enable reflective access to the specified members.
The PR is split in two commits to facilitate reviewing. The first one is a simple copy-paste of the parser from master (most of the contents of the com.oracle.svm.core.configure package) and the second contains the modifications required to make the parser fit with the features of 23.1.


zakkak:
@fniephaus @ezzarghili this be opened in https://github.com/graalvm/graalvm-community-jdk21u instead?

fniephaus:
> shouldn't this be opened in https://github.com/graalvm/graalvm-community-jdk21u instead?

Yes, I think so.

vjovanov:
This PR is moved [here](https://github.com/graalvm/graalvm-community-jdk21u/pull/23) as @zakkak suggested.

PR #9671, opened 2024-09-12
[GR-56098] Backport to 23.0: reachability-metadata.json parser

loicottet:
The goal of this PR is to enable the 23.0 version of GraalVM to correctly parse reachability metadata files emitted by GraalVM 24.1 and future releases. These files are parsed and matched to the existing capabilities of GraalVM 23.0 on a best effort basis, meaning that existing workflows won't be disrupted, but features introduced in subsequent versions will be ignored. For example, GraalVM 24.1 can register members (methods, fields and constructors) of proxy classes for reflection, which was not possible before, and does it via the reflection metadata instead of a separate proxy metadata file. When encountering such metadata, the 23.0 parser will now correctly register the proxy class for runtime instantiation, but won't enable reflective access to the specified members.
The PR is split in two commits to facilitate reviewing. The first one is a simple copy-paste of the parser from master (most of the contents of the com.oracle.svm.core.configure package) and the second contains the modifications required to make the parser fit with the features of 23.0.


vjovanov:
@dchuyko this PR needs to be backported to 17. Which repo are you currently using for the community backports for JDK 17?

dchuyko:
> @dchuyko this PR needs to be backported to 17. Which repo are you currently using for the community backports for JDK 17?

We use our internal one. https://github.com/graalvm/graalvm-for-jdk17-community-backports is archived. https://github.com/graalvm/graalvm-community-jdk21u could be a source of truth.

vjovanov:
If you have problems with back porting this PR, please let us know.

vjovanov:
This PR is no longer valid, closing.

PR #9779, opened 2024-09-29
Avoid adding same (AArch64|AMD64)SaveRegistersOp object twice to basic block

smarr:
This is not necessarily a performance issue. Or at least, I didn't check whether this actually leads to the instructions being emitted twice. Though, it violates basic assumptions at the LIR level, for instance, that individual instructions are separate objects.

The indirectly called `emitSaveRegisters(.)` methods already store the instructions.
But just to make sure, I added an assert that they are indeed already added.

This is code that hasn't been touched in years. Though, @tkrodriguez and @rschatz touched related bits (5 and 11 years ago).

For me, it causes issues when trying to revive @zapster's Trace Register Allocator (https://github.com/smarr/truffle/tree/trace-ra) for our work on optimizing the code produced for bytecode loops (without PE).

/cc @davleopo

### DIFF
--- a/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/AArch64HotSpotNodeLIRBuilder.java
+++ b/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/AArch64HotSpotNodeLIRBuilder.java
@@ -105,7 +105,6 @@ public class AArch64HotSpotNodeLIRBuilder extends AArch64NodeLIRBuilder implemen
             assert stub.getLinkage().getDescriptor().getTransition() != HotSpotForeignCallDescriptor.Transition.SAFEPOINT : stub;
             Register[] savedRegisters = getGen().getRegisterConfig().getAllocatableRegisters().toArray();
             AArch64SaveRegistersOp saveOp = getGen().emitSaveAllRegisters(savedRegisters);
-            append(saveOp);
             result.setSaveOnEntry(saveOp);
         }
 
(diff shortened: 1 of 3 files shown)


zapster:
That seems to be a regression from e7abe6206b46a8803bcad1995de1b6d687b6bc9d. @tkrodriguez can you have a look?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/AArch64HotSpotNodeLIRBuilder.java
```diff
@@ -105,7 +105,7 @@ protected void emitPrologue(StructuredGraph graph) {
             assert stub.getLinkage().getDescriptor().getTransition() != HotSpotForeignCallDescriptor.Transition.SAFEPOINT : stub;
             Register[] savedRegisters = getGen().getRegisterConfig().getAllocatableRegisters().toArray();
             AArch64SaveRegistersOp saveOp = getGen().emitSaveAllRegisters(savedRegisters);
-            append(saveOp);
+            assert getGen().lastIs(saveOp);
```
zapster, line 108: not sure the assertion is worth it. either `emitSaveAllRegisters` is supposed to append it (I think it should), or not.

tkrodriguez: It might be worth adding a more general assert in `LIRGenerator.append` like
```
assert !lirForBlock.contains(op) : "added " + op + " twice";
```

tkrodriguez:
I think this looks good.  Would you mind squashing into a single commit?  In case you aren't familiar with it, do something like:

```
git rebase -i 04b0554ffef16f42f879a3298d3c657db03127ab^
```
and then change `pick` to `fixup` for the last 2 commits.  Then `git push --force` the result.

smarr:
> I think this looks good. Would you mind squashing into a single commit? In case you aren't familiar with it, do something like:
> 
> ```
> git rebase -i 04b0554ffef16f42f879a3298d3c657db03127ab^
> ```
> 
> and then change `pick` to `fixup` for the last 2 commits. Then `git push --force` the result.

I am quite partial to https://www.git-tower.com/help/videos/surviving-with-git/squashing-commits 😉 
Squashed and commit comments fixed up.

tkrodriguez (approved):
Looks good.  I'll push it though our gate.

Issue #8237, opened 2024-01-24
Unable to collect GC data with NotificationEmitter in native build


viniciusxyz:
**A favor I ask of the team**

This is actually the second time I have opened this issue, the first was #7803 opened on November 11, 2023, but it was closed due to lack of interpretation, I ask that you please be aware that this issue does not refer to compilation problems and yes, the lack of data in the NotificationEmitter at runtime to collect GC metrics, this occurs in both G1GC and SerialGC in both GraalVM CE and Oracle GraalVM

**Describe the problem**

I'm developing applications and I'm missing some statistics when using Micrometer, being more specific the details and duration of GC pauses, I started validating the lib code and noticed that the problem is actually that it seems that NotificationEmitter is not working launching the notification events and this only happens in the native image, these metrics are very important for us to be able to put the applications into production, so I would like help to solve this problem.

**Steps to reproduce the issue**

1. git clone --depth 1 https://github.com/viniciusxyz/graalvm-issue-7803-notification-emmiter.git 
2. Build app `mvn clean package -Pnative`
3. Run app `./target/main-notification-emitter`

When running the native image, only the notifier addition log will be displayed, and when running with hotspot, when System.gc() is called, a log is displayed from the emission of the NotificationEmitter event

**Describe GraalVM and its environment:**
 - GraalVM version: GraalVM CE 21+35.1 (build 21+35-jvmci-23.1-b15)
 - JDK major version: 21
 - OS: Windows 11
 - Architecture: AMD64

**More details**

Print execution with hotspot:

![image](https://github.com/oracle/graal/assets/19265464/3e2454fc-9e6b-4441-b89c-b5e6dca1631c)

Print from execution with native compilation:

![image](https://github.com/oracle/graal/assets/19265464/6ed46e6f-fa2b-4f58-8389-dbcc366efb20)

All compilation configuration is in the [pom file](https://github.com/viniciusxyz/graalvm-issue-7803-notification-emmiter/blob/master/pom.xml#L58) of the project passed in the example

To prove that the problem does not depend on whether it is graalvm ce or oracle graalvm, the print follows with the same behavior:

![image](https://github.com/oracle/graal/assets/19265464/88788a33-6446-486f-9ce0-9be252b90aa7)

[...]

PR #9799, opened 2024-10-01
[GR-51307] Add GC NotificationEmitter Support

roberttoyonaga:
### Summary
This PR adds support for the emission of GC notifications from GC MXBeans. Previously, SubstratVM's GC MXBeans (`CompleteGarbageCollectorMXBean` and `IncrementalGarbageCollectorMXBean`) implemented `NotificationEmitter`, but actually did not support notifications.

### Details
A new feature `ServiceFeature` has been added. Use `--enable-monitoring=gcnotif` to include the feature at build time and add support for GC notifications. 

The ImageSingleton `ServiceSupport` starts up a dedicated service thread which asynchronously handles creating and sending notifications.  Notification "requests" are created during GC safepoints. They temporarily hold the information later needed to create and send GC notifications. During the GC safepoint, the thread executing the VM operation enqueues a request in `GcNotifier` and signals the service thread that a request is ready. That process should be relatively quick and allocation free. Once the safepoint is over, the service thread uses the queued requests to create notifications that are sent by the appropriate GC MXBean. 

The new classes `GcNotifier` and `ServiceThread` are meant to serve almost the same purposes as the `GCNotifier` and `ServiceThread` classes in Hotspot respectively. 

The various GC MXBeans now extend `AbstractGarbageCollectorMXBean` which contains the code for creating and sending notifications from request data.  This class serves a similar purpose to `com.sun.management.internal.GarbageCollectorExtImpl`. Additionally, this class extends `sun.management.NotificationEmitterSupport` from which the listener registration/deregistration logic is reused. 

### Other notes
I'm not sure about the naming of  `ServiceSupport` and `ServiceFeature`. Maybe `NotificationSupport` makes more sense? I decided on the "service" prefix because, in Hotspot, the GC notification code is in the /hotspot/share/services directory. Also, maybe we will want to add other services in the future.

Related Issues:  https://github.com/oracle/graal/issues/8237   https://github.com/oracle/graal/issues/13995


substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/HasGcNotificationSupport.java
```diff
@@ -0,0 +1,38 @@
...
+package com.oracle.svm.core.genscavenge.service;
+
+import java.util.function.BooleanSupplier;
+
+import com.oracle.svm.core.VMInspectionOptions;
+
+public class HasGcNotificationSupport implements BooleanSupplier {
+    @Override
+    public boolean getAsBoolean() {
+        return VMInspectionOptions.hasGcNotificationSupport();
```
christianhaeubl, line 36: Just a note: I was hoping that we could include this functionality unconditionally but it seems that this is not an option because it pulls in too much JDK code.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceFeature.java
```diff
@@ -0,0 +1,65 @@
...
+package com.oracle.svm.core.genscavenge.service;
+
+import com.oracle.svm.core.feature.AutomaticallyRegisteredFeature;
+import com.oracle.svm.core.feature.InternalFeature;
+import com.oracle.svm.core.jdk.RuntimeSupport;
+import com.oracle.svm.core.VMInspectionOptions;
+import org.graalvm.nativeimage.ImageSingletons;
+
+@AutomaticallyRegisteredFeature
+public class ServiceFeature implements InternalFeature {
```
christianhaeubl, line 36: `ServiceThreadFeature`, `ServiceThreadStartupHook`, `ServiceThreadShutdownHook`, `ServiceThreadSupport`

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceSupport.java
```diff
@@ -0,0 +1,79 @@
...
+import org.graalvm.nativeimage.Platforms;
+import org.graalvm.word.UnsignedWord;
+
+public class ServiceSupport {
+    private GcNotifier gcNotifier;
+    private ServiceThread serviceThread;
+
+    @Platforms(Platform.HOSTED_ONLY.class)
+    public ServiceSupport() {
+        gcNotifier = new GcNotifier();
```
christianhaeubl, line 43: `GcNotifier` should be a conditionally registered image singleton. If the service thread infrastructure is used for other use cases in the future, there is guarantee that a `GcNotifier` is actually present.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceFeature.java
```diff
@@ -0,0 +1,65 @@
...
+        RuntimeSupport.getRuntimeSupport().addStartupHook(new ServiceStartupHook());
+        RuntimeSupport.getRuntimeSupport().addShutdownHook(new ServiceShutdownHook());
+        ImageSingletons.add(ServiceSupport.class, new ServiceSupport());
+    }
+}
+
+final class ServiceStartupHook implements RuntimeSupport.Hook {
+    @Override
+    public void execute(boolean isFirstIsolate) {
+        if (isFirstIsolate) {
```
christianhaeubl, line 54: Please remove this check - any isolate could have a service thread.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceSupport.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    void initialize() {
+        serviceThread.start();
+    }
+
+    /**
+     * Called from GC code. Allocation Free.
+     */
+    public void beforeCollection(long startTime) {
```
christianhaeubl, line 59: The GC should directly call the `gcNotifier` (instead of calling into the `ServiceThreadSupport`).

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceThread.java
```diff
@@ -0,0 +1,87 @@
...
+import org.graalvm.nativeimage.Platform;
+import org.graalvm.nativeimage.Platforms;
+
+/**
+ * This class is the dedicated thread that handles services. For now, the only service is GC
+ * notifications.
+ */
+public class ServiceThread extends Thread {
+    private final UninterruptibleUtils.AtomicBoolean atomicNotify;
+    private final VMSemaphore semaphore;
```
christianhaeubl, line 42: Please use `VMMutex` and `VMCondition` instead (we only use `VMSemaphore` where it is absolutely necessary to do so, as it is much more complex to create on some platforms).

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/VMInspectionOptions.java
```diff
@@ -60,11 +60,13 @@ public final class VMInspectionOptions {
     private static final String MONITORING_JMXSERVER_NAME = "jmxserver";
     private static final String MONITORING_THREADDUMP_NAME = "threaddump";
     private static final String MONITORING_NMT_NAME = "nmt";
+    private static final String MONITORING_GC_NOTIFICATIONS = "gcnotif";
```
christianhaeubl, line 63: I will need to discuss the option name internally. Does HotSpot use the term `gcnotif` in any way (e.g., is there an option to disable GC notifications)?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -1084,7 +1084,8 @@ protected void onValueUpdate(EconomicMap<OptionKey<?>, Object> values, Integer o
 
     @Fold
     public static final boolean needsExitHandlers() {
-        return ConcealedOptions.InstallExitHandlers.getValue() || VMInspectionOptions.hasJfrSupport() || VMInspectionOptions.hasNativeMemoryTrackingSupport();
+        return ConcealedOptions.InstallExitHandlers.getValue() || VMInspectionOptions.hasJfrSupport() || VMInspectionOptions.hasNativeMemoryTrackingSupport() ||
+                        VMInspectionOptions.hasGcNotificationSupport();
```
christianhaeubl, line 1088: Why is this change necessary?

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/GcNotifier.java
```diff
@@ -0,0 +1,162 @@
...
+     * then return immediately.
+     */
+    public void sendNotification() {
+        assert !VMOperation.isInProgressAtSafepoint();
+
+        /*
+         * We don't need to drain all requests since signals queue at the semaphore level. The rest
+         * of the requests will be handled in subsequent calls to this method by the sevice thread.
+         */
+        if (!updateCurrentRequest()) {
```
christianhaeubl, line 129: Please adapt this logic once you switch to `VMMutex` and `VMCondition`.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceThread.java
```diff
@@ -0,0 +1,87 @@
...
+import com.oracle.svm.core.locks.VMSemaphore;
+import com.oracle.svm.core.log.Log;
+import org.graalvm.nativeimage.Platform;
+import org.graalvm.nativeimage.Platforms;
+
+/**
+ * This class is the dedicated thread that handles services. For now, the only service is GC
+ * notifications.
+ */
+public class ServiceThread extends Thread {
```
christianhaeubl, line 40: Does HotSpot use the service thread or a separate notification thread for the GC notifications?

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceThread.java
```diff
@@ -0,0 +1,87 @@
...
+    private final UninterruptibleUtils.AtomicBoolean atomicNotify;
+    private final VMSemaphore semaphore;
+    private volatile boolean stopped;
+    private GcNotifier gcNotifier;
+
+    @Platforms(Platform.HOSTED_ONLY.class)
+    @SuppressWarnings("this-escape")
+    public ServiceThread(GcNotifier gcNotifier) {
```
christianhaeubl, line 48: Please set a name for this thread, and potentially also a thread group (please check in the HotSpot code).

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractGarbageCollectorMXBean.java
```diff
@@ -0,0 +1,134 @@
...
+            gcInfoBuilder = SubstrateUtil.cast(gib, GcInfoBuilder.class);
+        }
+        return gcInfoBuilder;
+    }
+
+    /**
+     * Use the data taken from the request queue to populate MemoryUsage. The service thread calls
+     * this method.
+     */
+    public void createNotification(GcNotificationRequest request) {
```
christianhaeubl, line 73: Parts of this code are probably based on OpenJDK code. If so, please annotate with `@BasedOnJDKFile`.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/GcNotificationRequest.java
```diff
@@ -0,0 +1,104 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.svm.core.genscavenge.service;
```
christianhaeubl, line 27: Please move everything in this package to a sub package in the GC independent project `com.oracle.svm.core
com.oracle.svm.core`.

christianhaeubl (commented):
Thanks for the PR, I added a larger number of comments.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/VMInspectionOptions.java
roberttoyonaga: Hotspot doesn't use the name `gcnotif`. I just decided on it arbitrarily. As far as I can tell, GC notifications are always are enabled (as long as there are some listeners registered).

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceThread.java
roberttoyonaga: Hotspot uses `NotificationThread`, which is different than `ServiceThread`. We should rename this class "NotificationThread" instead. Unless we plan to have one dedicated thread handle service and notifications.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/GcNotifier.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * This class keeps a circular queue of requests to later use when emitting GC notifications. The
+ * requests are pre-allocated so that they can be used during a GC.
+ *
+ * This class is responsible for enqueuing requests during GC safepoints and later dequeuing
+ * requests to generate notifications when the service thread handles signals outside of safepoints.
+ */
+public class GcNotifier {
+    // 5 is probably more than enough, although making the queue larger isn't a problem either.
```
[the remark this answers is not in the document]
roberttoyonaga: Hotspot uses a linked list instead of a circular buffer. So they never drop notifications. I opted for a pre-allocated buffer for simplicity and so we don't have to deal with native memory.  I think it'll be rare that there'll be many successive GCs before the queue can be processed. However, I can change it to a linked list if you think it'll be a problem.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceFeature.java
roberttoyonaga: Based on https://github.com/oracle/graal/pull/9799#discussion_r1795872505, is it better to rename to `NotificationThreadFeature` , `NotificationThreadStartupHook`, etc ? I guess it depends on whether we plan to combine service and notifications together which might make sense if there aren't many services/notifications.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractGarbageCollectorMXBean.java
```diff
@@ -0,0 +1,134 @@
...
+                        new MBeanNotificationInfo(
+                                        new String[]{GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION},
+                                        "javax.management.Notification",
+                                        "GC Notification")
+        };
+    }
+
+    @Override
+    public GcInfo getLastGcInfo() {
+        return gcInfo;
```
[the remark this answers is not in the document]
roberttoyonaga: Oh that's a good point. And it's worse if the queue has more items in it.  It looks like Hotspot always returns info from the most recent GC. I'll change this method to do the same.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/GcNotificationRequest.java
roberttoyonaga: I tried doing that originally, but it seems I cannot import from `com.oracle.svm.core.genscavenge` from inside `com.oracle.svm.core`.  Is there a way to work around this? 

The build error:
`error: package com.oracle.svm.core.genscavenge does not exist`

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceSupport.java
roberttoyonaga: Hmm I'm not sure I fully understand.  Do you mean that it should be conditionally registered here inside `ServiceSupport`?  Why would GcNotifier not be present if the `ServiceFeature` has been included?

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceFeature.java
christianhaeubl: Sounds good, let's use `Notification*` for now.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceSupport.java
christianhaeubl: The service/notification thread may be used by multiple features (this is not the case at the moment but it is very likely that this will be the case in the future). Not all features are necessarily present in a Native Image, so we should implement generic infrastructure (like the service/notification thread) in a way that it still works even if some features are not available.

If the GC notifier is an image singleton, this can be achieve easily by calling `ImageSingletons.contains(...)` in the notification thread to check if the GC notifier is present. If so, the thread can do the corresponding processing work.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/GcNotificationRequest.java
christianhaeubl: You will need to split the implementation into a GC independent and a GC-specific part. This will also make it easier to implement this for all Native Image GCs in the future:
- GC independent parts in `com.oracle.svm.core`: notification thread, notification logic, and all the GC independent logic that accesses the memory pools. You will probably need introduce an interface and implementation that interface on the GC-specific side.
- GC specific parts in `com.oracle.svm.core.genscavenge`: essentially everything that would be different for a completely different GC implementation (e.g., the logic in `GenScavengeMemoryPoolMXBeans`, how many memory pools a GC uses, or how the memory pools are named)

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractGarbageCollectorMXBean.java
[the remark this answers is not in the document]
roberttoyonaga: I've fixed the approach here by always caching the GC info during the collection. This is basically [what Hotspot is doing](https://github.com/openjdk/jdk/blob/jdk-24%2B19/src/hotspot/share/services/memoryManager.cpp#L302) too.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/GcNotifier.java
```diff
@@ -0,0 +1,162 @@
...
+    private static final int QUEUE_SIZE = 5;
+    private static volatile int rear;
+    private static volatile int front;
+
+    GcNotificationRequest[] requests;
```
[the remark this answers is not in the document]
roberttoyonaga: I've created a class `CircularQueue` that extends `RingBuffer`, since `RingBuffer` only seems to retain the last N items but doesn't actually track the first/last items added.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceFeature.java
roberttoyonaga: I've  changed all prefixes to "Notification" and removed all uses of "Service" to be consistent.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceSupport.java
roberttoyonaga: Ok I've made `GcNotifier` an image singleton that is registered if `GcNotificationFeature` is included at build time with `--enable-monitoring=gcnotif`. 

`NotificationThreadFeature` is included depending on whether `GcNotificationFeature` or other future `*NotificationFeature`s exist. 

`NotificationThread` will check, in it's loop, whether each type of notification feature exists before trying to send any notifications of that type.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/ServiceThread.java
roberttoyonaga: I replaced the usage of  `VMSemaphore` with `VMMutex` and `VMCondition`. However, I experienced deadlock in `NotificationThread.signal()` when trying to lock the mutex during GC. So instead I'm calling `VMCondition.signal()` without holding the mutex. So there can be a rare (and probably harmless) race described in [this comment](https://github.com/oracle/graal/pull/9799/files#diff-2849899c2604e9a87436de2bf07ffbabb5c0d878c9de12eafd33d435b7eea879R79-R85).

Do you think there's a better way to do this?

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/GcNotifier.java
roberttoyonaga: This is now handled by checking `NotificationThread.hasRequests()` before blocking in the notification thread loop.

substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/service/GcNotificationRequest.java
roberttoyonaga: I put the files into `com.oracle.svm.core.notification` and  `com.oracle.svm.core.gc` (the naming could maybe be improved).

`com.oracle.svm.core.gc` contains the interface `MemoryPoolMXBeansProvider` for accessing GC-specific memory pool info. 

I also put `AbstractGarbageCollectorMXBean` inside `com.oracle.svm.core.gc`. Different GCs can override methods like `gcThreadCount()` and  `isIncremental()` so that the notification infrastructure knows how to handle notifications for different collectors.

roberttoyonaga:
Thank you for the review @christianhaeubl.  I think I've addressed all your comments. Please have another look when you have time. Thanks!

christianhaeubl:
Thanks. I started integrating this PR. I will let you know if I run into any major issues.

christianhaeubl:
Just to give a brief update: before I can integrate this, I first need to finish a bunch of larger changes affecting the Native Image MX bean infrastructure. So, integration will be delayed a bit.

roberttoyonaga:
> Just to give a brief update: before I can integrate this, I first need to finish a bunch of larger changes affecting the Native Image MX bean infrastructure. So, integration will be delayed a bit.

No problem. Thank you for the update. Please let me know if you find there's anything I can help with on my end.

roberttoyonaga:
Hi @christianhaeubl, have you had a chance to start integrating this? Should I merge master to fix conflicts, or not add any new commits?

Abdullah4345 (approved):

sluongng:
I recently started experimenting with using Graal Native Image for Google's Bazel build system https://github.com/bazelbuild/bazel/issues/7884#issuecomment-4936179830

And the missing GC bean was highlighted as a big miss for making the switch https://github.com/bazelbuild/bazel/pull/30285

Is there a GitHub issue that I can track the progress of this multi-year work?
Or is this PR the best place for now?

roberttoyonaga:
> I recently started experimenting with using Graal Native Image for Google's Bazel build system [bazelbuild/bazel#7884 (comment)](https://github.com/bazelbuild/bazel/issues/7884#issuecomment-4936179830)
> 
> And the missing GC bean was highlighted as a big miss for making the switch [bazelbuild/bazel#30285](https://github.com/bazelbuild/bazel/pull/30285)
> 
> Is there a GitHub issue that I can track the progress of this multi-year work? Or is this PR the best place for now?

Thank you for your input @sluongng! I have just made an issue to directly track this: https://github.com/oracle/graal/issues/13995
However, most of the discussion has been taking place on this PR and on https://github.com/oracle/graal/issues/8237

roberttoyonaga:
> No need to update this PR. The larger changes to the Native Image MX bean infrastructure are still work-in-progress on our side.

@christianhaeubl do you know if this work is still blocked?

christianhaeubl:
I actively started working on that topic again but there are still a lot of open points before this can be merged.

PR #9802, opened 2024-10-02
[GR-56599] Create type units for types during debuginfo generation for referencing across multiple debug info files.

graalvmbot:
Each type is currently generated within a compilation unit, with no way to reference it from outside the `.debug_info` section. During debuginfo generation at image runtime, we have to create a separate debuginfo file that references types generated during image build. This would require to reproduce debuginfo for types that was already created at image build.

One way to reference types in different objectfiles is to put them into so-called type units and create type signatures that can be used as references. Each type is placed in its own type unit and type references will be changed to refer to the type units. During the runtime debuginfo generation, we can then reference existing types by the type signature.

Type units were introduced in DWARF 4 in their own debugging section (.debug_types) and moved to the .debug_info section in DWARF 5 to reduce overhead. As a change to DWARF 5 requires little changes and makes creating type units more convenient, we will also update the DWARF version of the .debug_info section to DWARF 5.


substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/StructureTypeEntry.java
```diff
@@ -45,9 +48,25 @@ public abstract class StructureTypeEntry extends TypeEntry {
      */
     protected final List<FieldEntry> fields;
 
+    /**
+     * The type signature of this types' layout.
+     */
+    protected long layoutTypeSignature;
```
olpaw, line 54: You should mention why we need layoutTypeSignature at all.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/StructureTypeEntry.java
```diff
@@ -45,9 +48,25 @@ public abstract class StructureTypeEntry extends TypeEntry {
...
+    protected long indirectLayoutTypeSignature;
```
olpaw, line 55: Do we really need a compressed type variant `layoutTypeSignature`?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/TypeEntry.java
```diff
@@ -38,12 +34,24 @@
...
     /**
      * The name of this type.
      */
     protected final String typeName;
 
+    /**
+     * The type signature of this type.
+     */
+    protected long typeSignature;
+    protected long indirectTypeSignature;
```
olpaw, line 53: A better name might be `typeSignatureForCompressed`

olpaw: Ideally we would not need a separate field at all for the compressed variants and be able to derive a compressed variant typeSignature from the one and only `typeSignature`. But we can fix this in a followup PR.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/ForeignTypeEntry.java
```diff
@@ -57,6 +58,10 @@ public DebugInfoProvider.DebugTypeInfo.DebugTypeKind typeKind() {
         return DebugInfoProvider.DebugTypeInfo.DebugTypeKind.FOREIGN;
     }
 
+    public void setLayoutTypeSignature(long typeSignature) {
```
olpaw, line 61: Add a comment where this is used

olpaw:
Also:
* How does this  change to dwarf5 + type signatures affect debug-info size?

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf/constants/DwarfUnitHeader.java
```diff
@@ -0,0 +1,44 @@
...
+package com.oracle.objectfile.elf.dwarf.constants;
+
+/**
+ * All the Dwarf unit headers needed in the debug info section.
+ */
+public enum DwarfUnitHeader {
+    DW_UT_compile(0x1),
+    DW_UT_type(0x2);
+
+    private final int value;
```
olpaw, line 35: Some of these are of type `byte`  others, like here `int`, why not use byte also here?

olpaw (changes requested):
Looks good overall.

Please process the comments and we should be able to merge this soon.

dominikmascherbauer:
> Hi @dominikmascherbauer
> 
> the PR is in good shape already. I only have a few questions before we can merge this:
> 
>     * Did you check if this causes any regressions in debuginfo generation time of the image builder?

The time for debug info generation stays about the same, I checked that with debug info files with sizes of up to a few hundred MB.

>     * Is it guaranteed that the classloader contributes to the type signatures for everything except primitive types?

Yes, the classloader is looked up for all types except primitives (and the single instance of the header type which represents the object header).
The type signature then is the hash of the loader name and the type name, except for built-in classloaders (those are ignored in the hash).

>     * How does this change to dwarf5 + type signatures affect debug-info size?

Overall, this achieves a reduction of around 30% in debug info size, mostly due to the dwarf5 format changes of the .debug_loc and .debug_range sections.
Also, the loading time of debug info in gdb is reduced by about 15%.

substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/debugentry/StructureTypeEntry.java
dominikmascherbauer: No, the only time we use it currently is in the type unit for the compressed type, so we can move it there.

olpaw (approved):

PR #9832, opened 2024-10-07
[GR-58381] Synchronize accesses to class init state

graalvmbot:


zakkak:
Hello @c-refice, could you please provide a description for this PR?

I am wondering if this patch would help with the (theoretical) issue described in https://github.com/quarkusio/quarkus/issues/40737#issuecomment-2160681274:

>  if we add concurrency in class initialization there is no way for GraalVM to detect whether all threads have completed or not. That said if a class initializer spawns a thread and returns, the spawned thread can potentially run in parallel with the analysis (e.g. the reachable callbacks) and image building in general which can be problematic...

c-refice:
> I am wondering if this patch would help with the (theoretical) issue described in https://github.com/quarkusio/quarkus/issues/40737#issuecomment-2160681274:

From what I understood the issue is with a native image build, whereas this change is specific to HotSpot compilations in Graal JIT.

zakkak:
Thank you @alina-yur and @c-refice for the clarifications

PR #9846, opened 2024-10-08
Garbage collect on allocation exceeding max memory pool size

roberttoyonaga:
### Summary 
Garbage collection is currently done on the allocation slow path after the max eden size was already exceeded by a **previous** allocation.  For example, if new TLAB allocationA causes the max size to be exceeded, the pool max size will remain exceeded until new TLAB allocationB happens.  It would be better to do GC when the allocation that actually causes the threshold to be exceeded happens. 

This is because, on the slow path, [`maybeCollectOnAllocation`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/GCImpl.java#L157) is called too early. `maybeCollectOnAllocation` calls [`shouldCollectOnAllocation`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractCollectionPolicy.java#L90) which checks whether the eden size threshold is exceeded.
The problem is that `maybeCollectOnAllocation` is called before `increaseEdenUsedBytes` is called, so `shouldCollectOnAllocation` doesn't take into account the size of the current allocation. Is there a reason for this, or is this a bug? 

I've just made a small change so that `shouldCollectOnAllocation` takes into account the size of the current allocation that is requiring a new TLAB (and potentially causing the memory pool max to be exceeded). 

See https://github.com/oracle/graal/pull/6930 for more context.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractCollectionPolicy.java
+++ b/substratevm/src/com.oracle.svm.core.genscavenge/src/com/oracle/svm/core/genscavenge/AbstractCollectionPolicy.java
@@ -87,11 +87,11 @@ abstract class AbstractCollectionPolicy implements CollectionPolicy {
     }
 
     @Override
-    public boolean shouldCollectOnAllocation() {
+    public boolean shouldCollectOnAllocation(UnsignedWord allocationSize) {
         if (sizes == null) {
             return false; // updateSizeParameters() has never been called
         }
-        UnsignedWord edenUsed = HeapImpl.getAccounting().getEdenUsedBytes();
+        UnsignedWord edenUsed = HeapImpl.getAccounting().getEdenUsedBytes().add(allocationSize);
         return edenUsed.aboveOrEqual(edenSize);
     }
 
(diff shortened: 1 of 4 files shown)


christianhaeubl:
Note that your change is not enough to fix the issue that the actual usage of memory pools may exceed the maximum value temporarily. Unfortunately, the problem is more fundamental and would need a larger change. That is also the reason why I merged https://github.com/oracle/graal/pull/6930 instead of implementing a proper fix right away.

Just to give a rough idea:
- At the moment, memory pools don't really have a maximum size (i.e., unlike other GCs, the Native Image serial GC doesn't reserve a fixed address space for each memory pool).
- Due to that, very large objects (e.g., objects that are larger than the size of eden) are allocated in eden anyways.
- Also, the check in `shouldCollectOnAllocation` is not atomic enough to actually prevent the maximum value from being exceeded temporarily.

roberttoyonaga:
I'm closing this because it seems like it's not really a problem that the memory pool "thresholds" are temporarily exceeded.

PR #9957, opened 2024-10-24
Only `--enable-native-access=org.graalvm.truffle` when svmt present

zakkak:
Since https://github.com/oracle/graal/pull/9892 got merged, passing the parameter `--enable-native-access=org.graalvm.truffle` to GraalVM builds not shipping the truffle api results in the following warning:

```
WARNING: Unknown module: org.graalvm.truffle specified to --enable-native-access
```

This patch checks whether the Truffle Runtime SVM component is part of the GraalVM distribution to figure out whether the parameter should be included or not.

Not sure if that's enough, or whether we should check for other components as well.

Fixes https://github.com/graalvm/mandrel/issues/805

cc @chumer @tzezula


zakkak:
This seems to be superseded by https://github.com/oracle/graal/pull/9964. Please free to close this issue if I am correct. Thank you @tzezula.

PR #9963, opened 2024-10-24
[GR-56601] Initial support for jcmd.

graalvmbot:
This PR is based on https://github.com/oracle/graal/pull/9232 (there is not much code left from the original PR though) and contains the following changes:
- Added [jcmd](https://docs.oracle.com/en/java/javase/23/docs/specs/man/jcmd.html) support on Linux and macOS (see `JCmdFeature`).
    - the `jcmd` support builds on top of the attach API (see `PosixAttachApiSupport` and the files in `com.oracle.svm.core.attach` and `com.oracle.svm.core.posix.attach`)
    - the attach API uses `SIGQUIT`/`SIGBREAK` and domain sockets for communication
    - Native Image announces via a JVM performance counter that supports the attach API (see changes in `com.oracle.svm.core.jvmstat.SystemCounters`). `jcmd` can access the performance counters to detect which process are running.
    - Native Image already partially supported `SIGQUIT`/`SIGBREAK`. This PR extends the existing feature and renames it from `DumpRuntimeCompilationOnSignalFeature` to `SigQuitFeature`.
- Implemented 15 diagnostic commands (see `com.oracle.svm.core.dcmd.*`) that can be triggered via `jcmd`
- Various JFR changes and simplifications so that `-XX:StartFlightRecording` and `jcmd <pid> JFR.start` use the same code and have the same behavior.
- Minor JFR fixes for `JVMSupport.makeFilename` and the JFR startup.
- Minor changes to the heap dumping infrastructure so that we support everything that is needed for `jcmd <pid> GC.heap_dump dump.hprof`.
- A few Posix infrastructure changes to get rid of some code duplication (primarily related to write operations on files/sockets).


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jcmd/JCmdFeature.java
```diff
@@ -0,0 +1,74 @@
...
+@AutomaticallyRegisteredFeature
+public class JCmdFeature implements InternalFeature {
+    @Override
+    public boolean isInConfiguration(IsInConfigurationAccess access) {
+        return VMInspectionOptions.hasJCmdSupport();
+    }
+
+    @Override
+    public List<Class<? extends Feature>> getRequiredFeatures() {
+        return List.of(SigQuitFeature.class, AttachApiFeature.class);
```
roberttoyonaga, line 72: Should the `DCmdFeature` also be included in this list?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/dcmd/AbstractJfrDCmd.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    protected String getSyntaxAndExamples() {
+        Target_jdk_jfr_internal_dcmd_AbstractDCmd cmd = createDCmd();
+        String[] lines = getHelp(cmd);
+        return String.join(System.lineSeparator(), lines);
+    }
+
+    private static String[] getHelp(Target_jdk_jfr_internal_dcmd_AbstractDCmd cmd) {
```
roberttoyonaga, line 64: The JFR.start DCMD from the JDK shows extended JFR configuration options (that aren't supported) as part of it's help text. It's not a big problem since you return a descriptive error message `Warning! The .jfc option/setting 'xyz' doesn't exist or is not supported.` The other option would be to print custom help text, copy-pasting the parts we support in Native Image. However, maybe that's not really worth the duplication and maintenance of keeping it up to date.

roberttoyonaga (approved):
Thank you for doing this PR @christianhaeubl! This looks really good to me.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jcmd/JCmdFeature.java
christianhaeubl: Good point, it is indeed cleaner if `DCmdFeature` is included here as well.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/dcmd/AbstractJfrDCmd.java
christianhaeubl: Right, there is a bit too much output. Most extended configuration options are already hidden due to the substitution of `Target_jdk_jfr_internal_dcmd_DCmdStart.jfcOptions()`. I don't think that it is worth to duplicate the help text, so I think we will just live with the extra output for now.

fniephaus (approved):

Issue #10008, opened 2024-11-01
[Native Image] GraalVM for JDK 24 ignores manifest entry `Enable-Native-Access`


zakkak:
### Describe the Issue

GraalVM for JDK 24 ignores manifest entry `Enable-Native-Access` when compiling to native image, resulting in warnings. According to [JEP 472](https://openjdk.org/jeps/472):

> You can add `Enable-Native-Access: ALL-UNNAMED` to the manifest of an executable JAR file, i.e., a JAR file that is launched via [java -jar](https://docs.oracle.com/en/java/javase/22/docs/specs/man/java.html#synopsis). (The only supported value for the Enable-Native-Access manifest entry is `ALL-UNNAMED`; other values cause an exception to be thrown.)

I would expect `native-image` to detect this entry and not throw warnings.

### Using the latest version of GraalVM can resolve many issues.

- [X] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

### GraalVM Version

native-image 24 2025-03-18
GraalVM Runtime Environment GraalVM CE 24-dev+21.1 (build 24+21-jvmci-b01)
Substrate VM GraalVM CE 24-dev+21.1 (build 24+21, serial gc)

8a88c4486d4c0463bb17de5099d9ac52203b920a

### Operating System and Version

Linux fedora 6.11.5-300.fc41.x86_64

### Build Command

$JAVA_HOME/bin/native-image --initialize-at-build-time=. -jar target/reproducer-1.0-SNAPSHOT.jar

### Expected Behavior

Native build should build without any warnings.

### Actual Behavior

Native build prints warnings:

```
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::loadLibrary has been called by com.aayushatharva.brotli4j.Brotli4jLoader in an unnamed module (file:/home/zakkak/code/issue-reproducers/target/reproducer-1.0-SNAPSHOT.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```

### Steps to Reproduce

```
cd /tmp
git clone --branch 2024-11-01-jep472-manifest-ignored https://github.com/zakkak/issue-reproducers reproducers
cd reproducers
mvn package
# run in JVM (with no warnings)
$JAVA_HOME/bin/java \
  -jar target/reproducer-1.0-SNAPSHOT.jar
# generate native-image
[...]

PR #10009, opened 2024-11-01
[GR-60382] Handle Enable-Native-Access manifest attribute if present

zakkak:
Closes https://github.com/oracle/graal/issues/10008

### DIFF
--- a/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
+++ b/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
@@ -1055,6 +1055,7 @@ public class NativeImage {
     void handleManifestFileAttributes(Path jarFilePath, Attributes mainAttributes) {
         handleMainClassAttribute(jarFilePath, mainAttributes);
         handleModuleAttributes(mainAttributes);
+        handleEnableNativeAccessAttribute(mainAttributes);
     }
 
     void handleMainClassAttribute(Path jarFilePath, Attributes mainAttributes) {
@@ -1075,6 +1076,16 @@ public class NativeImage {
         String addExportsValues = mainAttributes.getValue("Add-Exports");
         if (addExportsValues != null) {
             handleModuleExports(addExportsValues, NativeImageClassLoaderOptions.AddExports);
+        }
+    }
+
+    void handleEnableNativeAccessAttribute(Attributes mainAttributes) {
+        String nativeAccessAttrName = mainAttributes.getValue("Enable-Native-Access");
+        if (nativeAccessAttrName != null) {
+            if (!ALL_UNNAMED.equals(nativeAccessAttrName)) {
+                throw NativeImage.showError("illegal value \"" + nativeAccessAttrName + "\" for " + nativeAccessAttrName + " manifest attribute. Only " + ALL_UNNAMED + " is allowed");
+            }
+            addImageBuilderJavaArgs("--enable-native-access=" + ALL_UNNAMED);
         }
     }
 


zapster:
thanks for the contribution @zakkak. I'll have a look as soon as time allows.

zapster:
Sorry for the delay, @zakkak. A lot of side tracking this week. I agree that #10008 is something we need to fix. In the current state, `--enable-native-access` is only about the image builder itself. At image run time, all native access check are currently disabled (aka all native access is allowed to all modules). This is about to change. I've pushed the current proposal to #10076. It is not yet fully ready, but the basics should be there. I'll need to do further testing, especially with external projects, though. Feedback from your sides is also very welcome.

That said, I believe that #10076 will break this PR, or at least partially defeat its purpose. I suggest we wait until #10076 lands and then revisit this and #10008. Would that work for you?

zakkak:
Thank you for the update @zapster .

> Would that work for you?

Yes, that's fine with us. As said in https://github.com/quarkusio/quarkus/issues/44257#issuecomment-2455114293 this is still a warning.

zakkak:
> I'll need to do further testing, especially with external projects, though. Feedback from your sides is also very welcome.

Quarkus is not using modules, so I am not sure how much we can test this on our side other than passing `--enable-native-access=ALL-UNNAMED`. If there is anything specific you would like me to try, let me know.

zapster:
> Quarkus is not using modules, so I am not sure how much we can test this on our side

That is totally fine. Out of interest, are you already passing `--enable-native-access=ALL-UNNAMED` to the image builder, or are do you intend to set this via the manifest attribute (#10008)?

zakkak:
No, we currently don't do anything about it. For some cases setting it via the manifest attribute would be enough, but we will definitely need to pass `--enable-native-access` in other cases such as testing, dev-mode, etc.

The Quarkus issue tracking this is https://github.com/quarkusio/quarkus/issues/44257.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1078,6 +1079,13 @@ void handleModuleAttributes(Attributes mainAttributes) {
         }
     }
 
+    void handleEnableNativeAccessAttribute(Attributes mainAttributes) {
+        String enableNativeAccess = mainAttributes.getValue("Enable-Native-Access");
+        if (enableNativeAccess != null) {
```
zapster, line 1084: We should check here that `enableNativeAccess` is `ALL-UNNAMED` because that is the only eligible values and if it is not, issue a error message similar to the JDK.

zakkak: Done

zapster (approved):
Thanks for the update, the change looks good now. I'll integrate it asap, but I'm still busy with post-branch-off tasks, so it could take a bit. However, I'll make sure that this is backported to the GraalVM for 24 release branch.

zakkak:
Great, thank you @zapster :pray:

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1078,6 +1079,16 @@ void handleModuleAttributes(Attributes mainAttributes) {
         }
     }
 
+    void handleEnableNativeAccessAttribute(Attributes mainAttributes) {
+        String enableNativeAccess = mainAttributes.getValue("Enable-Native-Access");
```
olpaw, line 1083:
```suggestion
        String nativeAccessAttrName = "Enable-Native-Access";
        String enableNativeAccess = mainAttributes.getValue(nativeAccessAttrName);
```

olpaw (changes requested):
Small changes requested

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1078,6 +1079,16 @@ void handleModuleAttributes(Attributes mainAttributes) {
...
+        if (enableNativeAccess != null) {
+            if (!"ALL-UNNAMED".equals(enableNativeAccess)) {
+                NativeImage.showError("illegal value \"" + enableNativeAccess + "\" for Enable-Native-Access manifest attribute. Only ALL-UNNAMED is allowed");
```
zapster, line 1086:
```suggestion
                throw NativeImage.showError("illegal value \"" + enableNativeAccess + "\" for Enable-Native-Access manifest attribute. Only " + ALL_UNNAMED + " is allowed");
```

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1078,6 +1079,16 @@ void handleModuleAttributes(Attributes mainAttributes) {
         }
     }
 
+    void handleEnableNativeAccessAttribute(Attributes mainAttributes) {
+        String enableNativeAccess = mainAttributes.getValue("Enable-Native-Access");
+        if (enableNativeAccess != null) {
+            if (!"ALL-UNNAMED".equals(enableNativeAccess)) {
```
zapster, line 1085:
```suggestion
            if (!ALL_UNNAMED.equals(enableNativeAccess)) {
```

zapster (changes requested):
Since there are anyways pending changes, I've added another minor suggestion.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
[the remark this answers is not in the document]
olpaw: @zakkak, odd that the message starts with lowercase (`illegal value ...`) but that's exactly what
```
src/java.base/share/classes/sun/launcher/resources/launcher.properties
```
uses. So you did the right thing.
Using `git grep 'Error: [A-Z]'` vs. `git grep 'Error: [a-z]'` shows there is zero consistency in this regard.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
```diff
@@ -1078,6 +1079,16 @@ void handleModuleAttributes(Attributes mainAttributes) {
...
+                NativeImage.showError("illegal value \"" + enableNativeAccess + "\" for Enable-Native-Access manifest attribute. Only ALL-UNNAMED is allowed");
+            }
+            addImageBuilderJavaArgs("--enable-native-access=" + enableNativeAccess);
```
zapster, line 1088:
```suggestion
            addImageBuilderJavaArgs("--enable-native-access=" + ALL_UNNAMED);
```

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
zakkak: Out of curiosity, what's the point of `throw` here since `showError` never actually returns. Looking in other `showError` usages the use of `throw` doesn't seem to be consistent.

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
zakkak: Good catch, interesting that it works even with `enableNativeAccess`

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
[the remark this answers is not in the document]
zakkak: Done

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
zakkak: Done

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
zakkak: Done

substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java
zapster: It helps me to see that execution will stop here without knowing the implementation of  `showError`. Of course one can look it up, but I find it more readable. But it is a matter of taste, mostly.

olpaw (approved):

zapster (approved):

PR #10135, opened 2024-11-21
Implement Truffle compiler control based on HotSpot's CompileBroker compilation activity

### LINKED ISSUE #10133: Implement Truffle compiler control based on HotSpot's CompileBroker compilation activity
Truffle compilations run in "hosted" mode, i.e. the Truffle runtimes triggers compilations independently of HotSpot's [`CompileBroker`](https://github.com/openjdk/jdk/blob/8f22db23a50fe537d8ef369e92f0d5f9970d98f0/src/hotspot/share/compiler/compileBroker.hpp). But the results of Truffle compilations are still stored as ordinary nmethods in HotSpot's code cache (with the help of the JVMCI method `jdk.vm.ci.hotspot.HotSpotCodeCacheProvider::installCode()`). The regular JIT compilers are controlled by the `CompileBroker` which is aware of the code cache occupancy. If the code cache runs full, the `CompileBroker` temporary pauses any subsequent JIT compilations until the code cache gets swept (if running with `-XX:+UseCodeCacheFlushing -XX:+MethodFlushing` which is the default) or completely shuts down the JIT compilers if running with `-XX:+UseCodeCacheFlushing`.

Truffle compiled methods can contribute significantly to the overall code cache occupancy and they can trigger JIT compilation stalls if they fill the code cache up. But the Truffle framework itself is neither aware of the current code cache occupancy, nor of the compilation activity of the `CompileBroker`. If Truffle tries to install a compiled method through JVMCI and the code cache is full, it will silently fail. Currently Truffle interprets such failures as transient errors and basically ignores it. Whenever the corresponding method gets hot again (usually immediately at the next invocation), Truffle will recompile it again just to fail again in the nmethod installation step, if the code cache is still full.

When the code cache is tight, this can lead to situations, where Truffle is unnecessarily and repeatedly compiling methods which can't be installed in the code cache but produce a significant CPU load. Instead, Truffle should poll HotSpot's `CompileBroker` compilation activity and pause compilations for the time the `CompileBroker` is pausing JIT compilations (or completely shutdown Truffle compilations if the `CompileBroker` shut down the JIT compilers).

The corresponding JVMCI change is tracked under [JDK-8344727: [JVMCI] Export the CompileBroker compilation activity mode for Truffle compiler control](https://bugs.openjdk.org/browse/JDK-8344727).

This topic has initially been discussed on the Truffle channel in the Graal Slack: https://graalvm.slack.com/archives/CNQSB2DHD/p1724699423251739

simonis:
Truffle compilations run in "hosted" mode, i.e. the Truffle runtimes triggers compilations independently of HotSpot's [`CompileBroker`](https://github.com/openjdk/jdk/blob/8f22db23a50fe537d8ef369e92f0d5f9970d98f0/src/hotspot/share/compiler/compileBroker.hpp). But the results of Truffle compilations are still stored as ordinary nmethods in HotSpot's code cache (with the help of the JVMCI method `jdk.vm.ci.hotspot.HotSpotCodeCacheProvider::installCode()`). The regular JIT compilers are controlled by the `CompileBroker` which is aware of the code cache occupancy. If the code cache runs full, the `CompileBroker` temporary pauses any subsequent JIT compilations until the code cache gets swept (if running with `-XX:+UseCodeCacheFlushing -XX:+MethodFlushing` which is the default) or completely shuts down the JIT compilers if running with `-XX:+UseCodeCacheFlushing`.

Truffle compiled methods can contribute significantly to the overall code cache occupancy and they can trigger JIT compilation stalls if they fill the code cache up. But the Truffle framework itself is neither aware of the current code cache occupancy, nor of the compilation activity of the `CompileBroker`. If Truffle tries to install a compiled method through JVMCI and the code cache is full, it will silently fail. Currently Truffle interprets such failures as transient errors and basically ignores it. Whenever the corresponding method gets hot again (usually immediately at the next invocation), Truffle will recompile it again just to fail again in the nmethod installation step, if the code cache is still full.

When the code cache is tight, this can lead to situations, where Truffle is unnecessarily and repeatedly compiling methods which can't be installed in the code cache but produce a significant CPU load. Instead, Truffle should poll HotSpot's `CompileBroker` compilation activity and paus compilations for the time the `CompileBroker` is pausing JIT compilations (or completely shutdown Truffle compilations if the `CompileBroker` shut down the JIT compilers).

The corresponding JVMCI change is tracked under [JDK-8344727: [JVMCI] Export the CompileBroker compilation activity mode for Truffle compiler control](https://bugs.openjdk.org/browse/JDK-8344727).

This PR fixes the problem by checking HotSpot's compilation activity mode in `OptimizedTruffleRuntime::submitForCompilation()` before actually submitting a compilation task to a compile queue. If the compilation activity mode is `RUN_COMPILATION` the task is submitted as before without any changes. However, if the compilation activity mode is `STOP_COMPILATION` (i.e. the `CompileBroker` has temporarily stopped JIT compilations, we flush the current compile queue (because compiled methods can not be installed anyway) and return `null`. We also start a timer which can be configured with the new `StoppedCompilationRetryDelay` parameter (defaults to 1000ms). After `StoppedCompilationRetryDelay` we submit a new compilation task, even if the compilation activity mode is not `RUN_COMPILATION`. This can help to trigger a code cache cleanup in situations when there are no JIT compilations, because code cache sweeping is only triggered when new nmethods are installed. Finally, when the compilation activity mode is `SHUTDOWN_COMPILATION`, we simply shutdown the compilation queue and issue a warning to inform users about the code cache shortage.

I've manually tested the change by running an octane benchmark with a very small code cache. Before the change, we could see the following results:
```
RayTrace: 484
----
Score (version 9): 484

[engine] Truffle runtime statistics for engine 1
    Compilations                : 2255
      Success                   : 549
      Temporary Bailouts        : 1703
        jdk.vm.ci.code.BailoutException: Code installation failed: code cache is full: 1702
      Permanent Bailouts        : 0
      Failed                    : 0
      Interrupted               : 3
    Invalidated                 : 0
    Queues                      : 2480
    Dequeues                    : 30
        Target inlined into only caller: 30
    Splits                      : 156
    Compilation Accuracy        : 1.000000
    Queue Accuracy              : 0.987903
    Compilation Utilization     : 2.948298
    Remaining Compilation Queue : 195
    Time to queue               : count=2480, sum=187025989722, min=   60029, average= 75413705.53, max=181007679, maxTarget=cross
    Time waiting in queue       : count=2254, sum=1385581671, min=      54, average=   614721.24, max=156182217, maxTarget=Function.prototype.apply <split-322>

    JVMCI CompileBroker Time:
       Compile:          0,000 s
       Install Code:     0,000 s (installs: 0, CodeBlob total size: 0, CodeBlob code size: 0)

    JVMCI Hosted Time:
       Install Code:    27,227 s (installs: 565, CodeBlob total size: 5916736, CodeBlob code size: 1713432)

  nmethod code size         :  6834896 bytes
  nmethod total size        : 14834520 bytes
CodeCache: size=3896Kb used=3458Kb max_used=3586Kb free=437Kb
 bounds [0x00007ffff436d000, 0x00007ffff473b000, 0x00007ffff473b000]
 total_blobs=1378 nmethods=566 adapters=714
 compilation: disabled (not enough contiguous free space left)
              stopped_count=26, restarted_count=25
 full_count=3200

real	3m11,398s
user	11m19,363s
sys	0m9,387s
```

As you can see, out of 2255 compilations, 1702 have been to no purpose, because they failed in the final installation step because of a full code cache (i.e. `BailoutException: Code installation failed: code cache is full: 1702`). The other interesting observation is that although the benchmark terminated after 3min11s wall clock time, it actually consumed 11m19s cpu time, because the Truffle compiler threads where continuously doing useless work.

With my changes applied, the picture looks as follows:
```
RayTrace: 531
----
Score (version 9): 531


[engine] Truffle runtime statistics for engine 1
    Compilations                : 128
      Success                   : 111
      Temporary Bailouts        : 17
        jdk.vm.ci.code.BailoutException: Code installation failed: code cache is full: 16
        org.graalvm.compiler.core.common.CancellationBailoutException: Compilation cancelled.: 1
      Permanent Bailouts        : 0
      Failed                    : 0
      Interrupted               : 0
    Invalidated                 : 0
    Queues                      : 410
    Dequeues                    : 283
        Compilation temporary disabled due to full code cache.: 277
        Target inlined into only caller: 6
    Splits                      : 156
    Compilation Accuracy        : 1.000000
    Queue Accuracy              : 0.309756
    Compilation Utilization     : 0.069104
    Remaining Compilation Queue : 0
    Time to queue               : count= 410, sum=24979066360, min=   88714, average= 60924552.10, max=186708180, maxTarget=blend
    Time waiting in queue       : count= 128, sum=  22955820, min=       6, average=   179342.35, max=  724639, maxTarget=:anonymous <split-344>

    JVMCI CompileBroker Time:
       Compile:          0,000 s
       Install Code:     0,000 s (installs: 0, CodeBlob total size: 0, CodeBlob code size: 0)

    JVMCI Hosted Time:
       Install Code:     0,637 s (installs: 124, CodeBlob total size: 1156760, CodeBlob code size: 321496)

  nmethod code size         :  2460456 bytes
  nmethod total size        :  4833272 bytes
CodeCache: size=3896Kb used=3464Kb max_used=3537Kb free=431Kb
 bounds [0x00007ffff436d000, 0x00007ffff473b000, 0x00007ffff473b000]
 total_blobs=1557 nmethods=745 adapters=714
 compilation: disabled (not enough contiguous free space left)
              stopped_count=5, restarted_count=4
 full_count=33

real	3m14,756s
user	3m22,013s
sys	0m1,044s
```

As you can see, we compile considerably fewer methods and we only have 16 compilation failures due to a full code cache. At the same time we can see that from the 410 methods which have been enqueued for compilation, 277 have been dequeued because of `Compilation temporary disabled due to full code cache`. Again, the wall clock time of the benchmark was 3m14s but this time, the overall cpu time was just slightly higher with 3m22s because we haven't done such a huge amount of unnecessary compilations any more. Also notice how HotSpot's code cache `full_count`, which is incremented every time when a nmethod can't be installed because of a full code cache, dropped from 3200 to 33.

These example were run with `-XX:+UseCodeCacheFlushing -XX:+MethodFlushing`. If the JVM is configured with `-XX:-UseCodeCacheFlushing`, the benefits of this change are even higher.

Fixes #10133

### DIFF
--- a/truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilerRuntime.java
+++ b/truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilerRuntime.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * The Universal Permissive License (UPL), Version 1.0
--- a/truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilerRuntime.java
+++ b/truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilerRuntime.java
@@ -359,4 +359,31 @@ public interface TruffleCompilerRuntime {
      */
     boolean isSuppressedFailure(TruffleCompilable compilable, Supplier<String> serializedException);
 
+    /**
+     * Represents HotSpot's compilation activity mode which is one of:
+     * {@code stop_compilation = 0}, {@code run_compilation = 1} or {@code shutdown_compilation = 2}
+     * Should be in sync with the {@code CompilerActivity} enum in {@code hotspot/share/compiler/compileBroker.hpp}
+     */
+    enum CompilationActivityMode {
+        STOP_COMPILATION,
+        RUN_COMPILATION,
+        SHUTDOWN_COMPILATION;
+
+        static public CompilationActivityMode fromInteger(int i) {
+            return switch (i) {
+                case 0 -> STOP_COMPILATION;
+                case 1 -> RUN_COMPILATION;
+                case 2 -> SHUTDOWN_COMPILATION;
+                default -> throw new RuntimeException("Invalid CompilationActivityMode " + i);
+            };
+        }
+    }
+
+    /**
+     * Returns the current host compilation activity mode which is one of:
+     * {@code STOP_COMPILATION}, {@code RUN_COMPILATION} or {@code SHUTDOWN_COMPILATION}
+     */
+     default CompilationActivityMode getCompilationActivityMode() {
+         return CompilationActivityMode.RUN_COMPILATION;
+     }
 }
--- a/truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilerRuntime.java
+++ b/truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilerRuntime.java
@@ -358,32 +358,4 @@ public interface TruffleCompilerRuntime {
      * silent.
      */
     boolean isSuppressedFailure(TruffleCompilable compilable, Supplier<String> serializedException);
-
-    /**
-     * Represents HotSpot's compilation activity mode which is one of:
-     * {@code stop_compilation = 0}, {@code run_compilation = 1} or {@code shutdown_compilation = 2}
-     * Should be in sync with the {@code CompilerActivity} enum in {@code hotspot/share/compiler/compileBroker.hpp}
-     */
-    enum CompilationActivityMode {
-        STOP_COMPILATION,
-        RUN_COMPILATION,
-        SHUTDOWN_COMPILATION;
-
-        static public CompilationActivityMode fromInteger(int i) {
-            return switch (i) {
-                case 0 -> STOP_COMPILATION;
-                case 1 -> RUN_COMPILATION;
-                case 2 -> SHUTDOWN_COMPILATION;
-                default -> throw new RuntimeException("Invalid CompilationActivityMode " + i);
-            };
-        }
-    }
-
-    /**
-     * Returns the current host compilation activity mode which is one of:
-     * {@code STOP_COMPILATION}, {@code RUN_COMPILATION} or {@code SHUTDOWN_COMPILATION}
-     */
-     default CompilationActivityMode getCompilationActivityMode() {
-         return CompilationActivityMode.RUN_COMPILATION;
-     }
 }
--- a/truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
+++ b/truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
@@ -84,6 +84,7 @@ import com.oracle.truffle.api.nodes.NodeVisitor;
 import com.oracle.truffle.api.nodes.RootNode;
 import com.oracle.truffle.compiler.TruffleCompilable;
 import com.oracle.truffle.runtime.OptimizedRuntimeOptions.ExceptionAction;
+import com.oracle.truffle.runtime.OptimizedTruffleRuntime.CompilationActivityMode;
 
 import jdk.vm.ci.meta.JavaConstant;
 import jdk.vm.ci.meta.SpeculationLog;
@@ -858,58 +859,60 @@ public abstract class OptimizedCallTarget implements TruffleCompilable, RootCall
         return compilationFailed;
     }
 
-    final boolean isCompilationStopped() {
-        OptimizedTruffleRuntime.CompilationActivityMode compilationActivityMode = runtime().getCompilationActivityMode();
+    final CompilationActivityMode getCompilationActivityMode() {
+        CompilationActivityMode compilationActivityMode = runtime().getCompilationActivityMode();
         long sct = runtime().stoppedCompilationTime().get();
-        if (compilationActivityMode == OptimizedTruffleRuntime.CompilationActivityMode.STOP_COMPILATION) {
+        if (compilationActivityMode == CompilationActivityMode.STOP_COMPILATION) {
             if (sct != 0 && System.currentTimeMillis() - sct > engine.stoppedCompilationRetryDelay) {
                 runtime().stoppedCompilationTime().compareAndSet(sct, 0);
-                // Try again every StoppedCompilationRetryDelay milliseconds to potentially trigger a code cache sweep.
-                compilationActivityMode = OptimizedTruffleRuntime.CompilationActivityMode.RUN_COMPILATION;
+                // Try again every StoppedCompilationRetryDelay milliseconds to potentially trigger
+                // a code cache sweep.
+                compilationActivityMode = CompilationActivityMode.RUN_COMPILATION;
             }
         }
 
         switch (compilationActivityMode) {
-            case RUN_COMPILATION : {
+            case RUN_COMPILATION: {
                 // This is the common case - compilations are not stopped.
-                return false;
+                return CompilationActivityMode.RUN_COMPILATION;
             }
-            case STOP_COMPILATION : {
+            case STOP_COMPILATION: {
                 if (sct == 0) {
                     runtime().stoppedCompilationTime().compareAndSet(0, System.currentTimeMillis());
                 }
                 // Flush the compilations queue for now. There's still a chance that compilation
                 // will be re-enabled eventually, if the hosts code cache can be cleaned up.
                 Collection<OptimizedCallTarget> targets = runtime().getCompileQueue().getQueuedTargets(null);
-                // If there's just a single compilation target in the queue, the chance is high that it is
-                // the one we've just added after the StoppedCompilationRetryDelay ran out, so keep it to
-                // potentially trigger a code cache sweep.
+                // If there's just a single compilation target in the queue, the chance is high that
+                // it is the one we've just added after the StoppedCompilationRetryDelay ran out, so
+                // keep it to potentially trigger a code cache sweep.
                 if (targets.size() > 1) {
                     for (OptimizedCallTarget target : targets) {
                         target.cancelCompilation("Compilation temporary disabled due to full code cache.");
                     }
                 }
-                return true;
+                return CompilationActivityMode.STOP_COMPILATION;
             }
-            case SHUTDOWN_COMPILATION : {
+            case SHUTDOWN_COMPILATION: {
                 // Compilation was shut down permanently because the hosts code cache ran full and
                 // the host was configured without support for code cache sweeping.
                 TruffleLogger logger = engine.getLogger("engine");
                 // The logger can be null if the engine is closed.
-                if (logger != null && runtime().logShutdownCompilations().compareAndExchange(true, false)) {
-                    logger.log(Level.WARNING, "Truffle compilations permanently disabled because of full code cache. "  +
-                            "Increase the code cache size using '-XX:ReservedCodeCacheSize=' and/or run with '-XX:+UseCodeCacheFlushing -XX:+MethodFlushing'.");
+                if (logger != null && engine.logShutdownCompilations().compareAndExchange(true, false)) {
+                    logger.log(Level.WARNING, "Truffle compilations permanently disabled because of full code cache. " +
+                                    "Increase the code cache size using '-XX:ReservedCodeCacheSize=' and/or run with '-XX:+UseCodeCacheFlushing -XX:+MethodFlushing'.");
                 }
-                try {
-                    runtime().getCompileQueue().shutdownAndAwaitTermination(100 /* milliseconds */);
-                } catch (RuntimeException re) {
-                    // Best effort, ignore failure
+                // Flush the compilation queue and mark all methods as not compilable.
+                for (OptimizedCallTarget target : runtime().getCompileQueue().getQueuedTargets(null)) {
+                    target.cancelCompilation("Compilation permanently disabled due to full code cache.");
+                    target.compilationFailed = true;
                 }
-                return true;
+                return CompilationActivityMode.SHUTDOWN_COMPILATION;
             }
-            default : CompilerDirectives.shouldNotReachHere("Invalid compilation activity mode: " + compilationActivityMode);
+            default:
+                CompilerDirectives.shouldNotReachHere("Invalid compilation activity mode: " + compilationActivityMode);
         }
-        return false;
+        return CompilationActivityMode.RUN_COMPILATION;
     }
 
     /**
@@ -931,16 +934,21 @@ public abstract class OptimizedCallTarget implements TruffleCompilable, RootCall
                 return false;
             }
 
+            CompilationActivityMode cam = getCompilationActivityMode();
+            if (cam != CompilationActivityMode.RUN_COMPILATION) {
+                if (cam == CompilationActivityMode.SHUTDOWN_COMPILATION) {
+                    // Compilation was shut down permanently.
+                    compilationFailed = true;
+                }
+                return false;
+            }
+
             CompilationTask task = null;
             // Do not try to compile this target concurrently,
             // but do not block other threads if compilation is not asynchronous.
             synchronized (this) {
                 if (!needsCompile(lastTier)) {
                     return true;
-                }
-
-                if (isCompilationStopped()) {
-                    return false;
                 }
 
                 ensureInitialized();
--- a/truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
+++ b/truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
@@ -43,6 +43,7 @@ package com.oracle.truffle.runtime;
 import java.lang.ref.Reference;
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -51,7 +52,9 @@ import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.function.Supplier;
+import java.util.logging.Level;
 
+import com.oracle.truffle.api.TruffleLogger;
 import org.graalvm.options.OptionKey;
 import org.graalvm.options.OptionValues;
 
@@ -855,6 +858,60 @@ public abstract class OptimizedCallTarget implements TruffleCompilable, RootCall
         return compilationFailed;
     }
 
+    final boolean isCompilationStopped() {
+        OptimizedTruffleRuntime.CompilationActivityMode compilationActivityMode = runtime().getCompilationActivityMode();
+        long sct = runtime().stoppedCompilationTime().get();
+        if (compilationActivityMode == OptimizedTruffleRuntime.CompilationActivityMode.STOP_COMPILATION) {
+            if (sct != 0 && System.currentTimeMillis() - sct > engine.stoppedCompilationRetryDelay) {
+                runtime().stoppedCompilationTime().compareAndSet(sct, 0);
+                // Try again every StoppedCompilationRetryDelay milliseconds to potentially trigger a code cache sweep.
+                compilationActivityMode = OptimizedTruffleRuntime.CompilationActivityMode.RUN_COMPILATION;
+            }
+        }
+
+        switch (compilationActivityMode) {
+            case RUN_COMPILATION : {
+                // This is the common case - compilations are not stopped.
+                return false;
+            }
+            case STOP_COMPILATION : {
+                if (sct == 0) {
+                    runtime().stoppedCompilationTime().compareAndSet(0, System.currentTimeMillis());
+                }
+                // Flush the compilations queue for now. There's still a chance that compilation
+                // will be re-enabled eventually, if the hosts code cache can be cleaned up.
+                Collection<OptimizedCallTarget> targets = runtime().getCompileQueue().getQueuedTargets(null);
+                // If there's just a single compilation target in the queue, the chance is high that it is
+                // the one we've just added after the StoppedCompilationRetryDelay ran out, so keep it to
+                // potentially trigger a code cache sweep.
+                if (targets.size() > 1) {
+                    for (OptimizedCallTarget target : targets) {
+                        target.cancelCompilation("Compilation temporary disabled due to full code cache.");
+                    }
+                }
+                return true;
+            }
+            case SHUTDOWN_COMPILATION : {
+                // Compilation was shut down permanently because the hosts code cache ran full and
+                // the host was configured without support for code cache sweeping.
+                TruffleLogger logger = engine.getLogger("engine");
+                // The logger can be null if the engine is closed.
+                if (logger != null && runtime().logShutdownCompilations().compareAndExchange(true, false)) {
+                    logger.log(Level.WARNING, "Truffle compilations permanently disabled because of full code cache. "  +
+                            "Increase the code cache size using '-XX:ReservedCodeCacheSize=' and/or run with '-XX:+UseCodeCacheFlushing -XX:+MethodFlushing'.");
+                }
+                try {
+                    runtime().getCompileQueue().shutdownAndAwaitTermination(100 /* milliseconds */);
+                } catch (RuntimeException re) {
+                    // Best effort, ignore failure
+                }
+                return true;
+            }
+            default : CompilerDirectives.shouldNotReachHere("Invalid compilation activity mode: " + compilationActivityMode);
+        }
+        return false;
+    }
+
     /**
      * Returns <code>true</code> if the call target was already compiled or was compiled
      * synchronously. Returns <code>false</code> if compilation was not scheduled or is happening in
@@ -866,6 +923,7 @@ public abstract class OptimizedCallTarget implements TruffleCompilable, RootCall
         if (!needsCompile(lastTier)) {
             return true;
         }
+
         if (!isSubmittedForCompilation()) {
             if (!acceptForCompilation()) {
                 // do not try to compile again
@@ -879,6 +937,10 @@ public abstract class OptimizedCallTarget implements TruffleCompilable, RootCall
             synchronized (this) {
                 if (!needsCompile(lastTier)) {
                     return true;
+                }
+
+                if (isCompilationStopped()) {
+                    return false;
                 }
 
                 ensureInitialized();
--- a/truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedRuntimeOptions.java
+++ b/truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedRuntimeOptions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * The Universal Permissive License (UPL), Version 1.0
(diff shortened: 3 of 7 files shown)


truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
```diff
@@ -912,10 +913,45 @@ private void notifyCompilationFailure(OptimizedCallTarget callTarget, Throwable
     protected void onEngineCreated(EngineData engine) {
     }
 
+    private long stoppedCompilationTime = 0;
+    private boolean logShutdownCompilations = true;
```
chumer, line 917: submitForCompilation is not thread-safe. We need to handle these fields to be modified from multpile threads. So we would need to synchronize accessing these fields.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
```diff
@@ -912,10 +913,45 @@ private void notifyCompilationFailure(OptimizedCallTarget callTarget, Throwable
...
+            }
+        } else {
+            // Compilation was shut down permanently because the hosts code cache ran full and
+            // the host was configured without support for code cache sweeping.
+            assert compilationActivityMode == CompilationActivityMode.SHUTDOWN_COMPILATION;
+            TruffleLogger logger = optimizedCallTarget.engine.getLogger("engine");
+            // The logger can be null if the engine is closed.
+            if (logger != null && logShutdownCompilations) {
+                logShutdownCompilations = false;
+                logger.log(Level.WARNING, "Truffle host compilations permanently disabled because of full code cache. "  +
```
chumer, line 945: We do not really call Truffle compilations host compilations, that is typically reserved for compilations of HotSpot directly. Maybe we should just change it to Truffle compilations to avoid confusion.
Sidenote: it would not be a problem to show this warning once per engine.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
```diff
@@ -912,10 +913,45 @@ private void notifyCompilationFailure(OptimizedCallTarget callTarget, Throwable
...
 
+    private long stoppedCompilationTime = 0;
+    private boolean logShutdownCompilations = true;
+
     @SuppressWarnings("try")
     public CompilationTask submitForCompilation(OptimizedCallTarget optimizedCallTarget, boolean lastTierCompilation) {
-        Priority priority = new Priority(optimizedCallTarget.getCallAndLoopCount(), lastTierCompilation ? Priority.Tier.LAST : Priority.Tier.FIRST);
-        return getCompileQueue().submitCompilation(priority, optimizedCallTarget);
+        BackgroundCompileQueue queue = getCompileQueue();
+        CompilationActivityMode compilationActivityMode = getCompilationActivityMode();
```
chumer, line 922: Maybe we should move this logic into OptimizedCallTarget#compile instead and have a more efficent check, similar to how we already check for OptimizedCallTarget.compilationTask? Currently we would take a lock for every OptimizedCallTarget.call that triggers a compilation if compilation is paused.

We should also use the OptimizedCallTrage.EngineData class to store any state associated with the retry time. I know it most likely will be the same for all engines, but its a principle we work by to avoid that engines can affect each other. So we also avoid sharing locks.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
```diff
@@ -912,10 +913,45 @@ private void notifyCompilationFailure(OptimizedCallTarget callTarget, Throwable
...
+            stoppedCompilationTime = 0;
+            Priority priority = new Priority(optimizedCallTarget.getCallAndLoopCount(), lastTierCompilation ? Priority.Tier.LAST : Priority.Tier.FIRST);
+            return queue.submitCompilation(priority, optimizedCallTarget);
+        } else if (compilationActivityMode == CompilationActivityMode.STOP_COMPILATION) {
+            if (stoppedCompilationTime == 0) {
+                stoppedCompilationTime = System.currentTimeMillis();
+            }
+            // Flush the compilations queue. There's still a chance that compilation will be re-enabled
+            // eventually, if the hosts code cache can be cleaned up.
+            for (OptimizedCallTarget target : queue.getQueuedTargets(optimizedCallTarget.engine)) {
```
chumer, line 934: This just cancels compilation for one engine only, whereas the JVMCI indication is for all engines. Either we move this code to the engine level or we pass `null` to filter for all engines in the queue.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
```diff
@@ -912,10 +913,45 @@ private void notifyCompilationFailure(OptimizedCallTarget callTarget, Throwable
...
+            assert compilationActivityMode == CompilationActivityMode.SHUTDOWN_COMPILATION;
+            TruffleLogger logger = optimizedCallTarget.engine.getLogger("engine");
+            // The logger can be null if the engine is closed.
+            if (logger != null && logShutdownCompilations) {
+                logShutdownCompilations = false;
+                logger.log(Level.WARNING, "Truffle host compilations permanently disabled because of full code cache. "  +
+                    "Increase the code cache size using '-XX:ReservedCodeCacheSize=' and/or run with '-XX:+UseCodeCacheFlushing -XX:+MethodFlushing'.");
+            }
+            try {
+                queue.shutdownAndAwaitTermination(100 /* milliseconds */);
```
chumer, line 949: Can you describe a bit what you are trying to achieve here? why would it help to block the entire interpreter thread if this happens?

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
```diff
@@ -912,10 +913,45 @@ private void notifyCompilationFailure(OptimizedCallTarget callTarget, Throwable
...
+            } catch (RuntimeException re) {
```
chumer, line 950: Why would this fail? If you want to catch InterruptedException and ignore that, that would be fine. But I do not like catching wildcard exceptions and ignore them, I would prefer to rethrow.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
```diff
@@ -912,10 +913,45 @@ private void notifyCompilationFailure(OptimizedCallTarget callTarget, Throwable
...
+    private long stoppedCompilationTime = 0;
+    private boolean logShutdownCompilations = true;
+
     @SuppressWarnings("try")
     public CompilationTask submitForCompilation(OptimizedCallTarget optimizedCallTarget, boolean lastTierCompilation) {
-        Priority priority = new Priority(optimizedCallTarget.getCallAndLoopCount(), lastTierCompilation ? Priority.Tier.LAST : Priority.Tier.FIRST);
-        return getCompileQueue().submitCompilation(priority, optimizedCallTarget);
+        BackgroundCompileQueue queue = getCompileQueue();
+        CompilationActivityMode compilationActivityMode = getCompilationActivityMode();
+        if (compilationActivityMode == CompilationActivityMode.RUN_COMPILATION ||
```
chumer, line 923: Minor, but I think we could write this as a switch and easier to read retry handling (probably opinionated):

```
CompilationActivityMode compilationActivityMode = getCompilationActivityMode();
if (compilationActivityMode != CompilationActivityMode.RUN_COMPILATION) {
   if ((stoppedCompilationTime != 0 && System.currentTimeMillis() - stoppedCompilationTime > optimizedCallTarget.getOptionValue(StoppedCompilationRetryDelay)) {
        compilationActivityMode = CompilationActivityMode.RUN_COMPILATION;
  }
}

switch (compilationActivityMode) {
 ...
  default : CompilerDirectives.shouldNotReachHere("invalid activity mode")
}

```

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotTruffleRuntime.java
```diff
@@ -692,4 +695,30 @@ public boolean isLibGraalCompilationEnabled() {
         return compilationSupport instanceof LibGraalTruffleCompilationSupport;
     }
 
+    static MethodHandle getCompilationActivityMode;
+    static {
+        try {
+            MethodType mt = MethodType.methodType(int.class);
+            getCompilationActivityMode = MethodHandles.lookup().findVirtual(HotSpotJVMCIRuntime.class, "getCompilationActivityMode", mt);
+        } catch (NoSuchMethodException | IllegalAccessException e) {
+            // Older JVMCI runtimes might not support `getCompilationActivityMode()`
```
chumer, line 704: The method handle field final can and should be final if you refactor this code.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotTruffleRuntime.java
```diff
@@ -692,4 +695,30 @@ public boolean isLibGraalCompilationEnabled() {
...
+     * {@code stop_compilation = 0}, {@code run_compilation = 1} or {@code shutdown_compilation = 2}
+     */
+    @Override
+    public CompilationActivityMode getCompilationActivityMode() {
+        int activityMode = 1; // Default is to run compilations
+        if (getCompilationActivityMode != null) {
+            try {
+                activityMode = (int) getCompilationActivityMode.invokeExact(HotSpotJVMCIRuntime.runtime());
+            } catch (Throwable t) {
+                // Ignore and go with the default mode
```
chumer, line 719: We are not catching errors like this. Please rethrow, we want to see and detect failures like this.
If this would repeatedly fail it would trigger an error each time which would be just visible in the performance.

truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilerRuntime.java
```diff
@@ -359,4 +359,31 @@ default ResolvedJavaType resolveType(MetaAccessProvider metaAccess, String class
...
+                default -> throw new RuntimeException("Invalid CompilationActivityMode " + i);
+            };
+        }
+    }
+
+    /**
+     * Returns the current host compilation activity mode which is one of:
+     * {@code STOP_COMPILATION}, {@code RUN_COMPILATION} or {@code SHUTDOWN_COMPILATION}
+     */
+     default CompilationActivityMode getCompilationActivityMode() {
```
chumer, line 386: There is no need to have the getCompilationActivityMode in this interface. This interface is intended for methods that get called by the Graal compiler. You can move this abstract specification directly into OptimizedTruffleRuntime and implemented more concretely in HotSpotTruffleRutnime.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedRuntimeOptions.java
```diff
@@ -158,6 +158,14 @@ public ExceptionAction apply(String s) {
     // TODO: GR-29949
     public static final OptionKey<Long> CompilerIdleDelay = new OptionKey<>(10000L);
 
+    @Option(help = "Before the Truffle runtime submits an OptimizedCallTarget for compilation, it checks for the " +
+                   "compilation activity mode in the host VM. If the activity mode is 'STOP_COMPILATION' because " +
+                   "of a full code cache, no new compilation requests are submitted and the compilation queue is flushed. " +
+                   "After 'StoppedCompilationRetryDelay' milliseconds new compilations will be submitted again " +
+                   "(which might trigger a sweep of the code cache and a reset of the compilation activity mode in the host JVM).",
```
chumer, line 165: Well this sounds like it would be supported in all runtimes. But it is not. SVM does not support this yet. So we should mention this only works for HotSpot right now. Note that OptimizedRuntimeOptions are shared across all optimizing runtimes this includes SVM.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedRuntimeOptions.java
```diff
@@ -158,6 +158,14 @@ public ExceptionAction apply(String s) {
     // TODO: GR-29949
     public static final OptionKey<Long> CompilerIdleDelay = new OptionKey<>(10000L);
 
+    @Option(help = "Before the Truffle runtime submits an OptimizedCallTarget for compilation, it checks for the " +
```
chumer, line 161: Minor: The help text is extensive but seems a bit implementation heavy. I think it is enough for a user of this option to know what happens that is observable to the user. In other words specific enum names like STOP_COMPILATION won't help understanding the semantics of this option.

chumer (changes requested):

truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilerRuntime.java
```diff
@@ -359,4 +359,31 @@ default ResolvedJavaType resolveType(MetaAccessProvider metaAccess, String class
...
+                case 1 -> RUN_COMPILATION;
+                case 2 -> SHUTDOWN_COMPILATION;
+                default -> throw new RuntimeException("Invalid CompilationActivityMode " + i);
+            };
+        }
+    }
+
+    /**
+     * Returns the current host compilation activity mode which is one of:
+     * {@code STOP_COMPILATION}, {@code RUN_COMPILATION} or {@code SHUTDOWN_COMPILATION}
```
dougxc, line 384: I would not list the enum constants in this javadoc - just one more location you need to remember to update if enum constants are added (or deleted/renamed).

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedRuntimeOptions.java
dougxc: Also, this option is ignored if not running on HotSpot right? Might be worth pointing that out.

dougxc: > Also, this option is ignored if not running on HotSpot right? Might be worth pointing that out.

Sorry - just saw now that @chumer already pointed this out.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
simonis: In this case synchronization is not really required for correctness (the worst thing that can happen is that we schedule a compilation too much if compilations are stopped or we log more than once if compilation has been shutdown) but I've now changed both fields to `AtomicLong` and `AtomicBoolean` to also make the code formally correct.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
simonis: Your observation is correct, but I don't understand your comment "*we pass null to filter for all engines in the queue*"? From looking at the code of `getQueuedTargets()`, passing a `null` argument for the engine wouldn't return any compilation at all:
```java
        for (CompilationTask.ExecutorServiceWrapper wrapper : array) {
            OptimizedCallTarget target = wrapper.compileTask.targetRef.get();
            if (target != null && target.engine == engine) {
                queuedTargets.add(target);
            }
        }
```
Did I misunderstood you or do you propose that I change `getQueuedTargets()` to treat a `null` engine as a filter for "all engines"?

chumer: > Did I misunderstood you or do you propose that I change getQueuedTargets() to treat a null engine as a filter for "all engines"?

yes that is what I meant. But of course if you have a better idea or want to do this in an extra API that is probably fine too

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
simonis: I've moved the compilation activity mode handling to `OptimizedCallTarget::compile()` as suggested.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
simonis: Done

simonis:
Hi @chumer, @dougxc,

Thanks a lot for your quick and very helpful review. I've just pushed a new version of my changes which hopefully addresses your requests (except for the right reaction on compilation shutdowns for which I've asked for additional help).

I've now moved the compilation mode handling one lever up into `OptimizedCallTarget` as suggseted by @chumer and refactored the code into a new function which makes it a little bit cleaner.

Please let me know what you think?

Thank you and best regards,
Volker

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedRuntimeOptions.java
```diff
@@ -158,6 +158,14 @@ public ExceptionAction apply(String s) {
...
+                   "compilation activity mode in the host VM. If the activity mode is 'STOP_COMPILATION' because " +
+                   "of a full code cache, no new compilation requests are submitted and the compilation queue is flushed. " +
+                   "After 'StoppedCompilationRetryDelay' milliseconds new compilations will be submitted again " +
+                   "(which might trigger a sweep of the code cache and a reset of the compilation activity mode in the host JVM).",
+                   usageSyntax = "<ms>", category = OptionCategory.EXPERT)
```
dougxc, line 167: I think you should still point out in this text that this (currently) only has an effect when running on HotSpot.

simonis: Hm, not sure if you saw the sentence I've added: "*The option is only supported on the HotSpot Truffle runtime. On runtimes which don't support it the option has no effect*".

Or am I missing something?

dougxc: Sorry, I was confused by the GitHub view which was pinned to a specific commit for some reason.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
```diff
@@ -855,6 +858,60 @@ final boolean isCompilationFailed() {
         return compilationFailed;
     }
 
+    final boolean isCompilationStopped() {
+        OptimizedTruffleRuntime.CompilationActivityMode compilationActivityMode = runtime().getCompilationActivityMode();
```
chumer, line 862: Optional: no need to qualify the class name. It makes enough sense without.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
chumer: The fastest way to disable a compilation is using OptimizedCallTarget.compilationFailed.
Every time we see shutdown with a call target we could disable compilation.
Best idea I have.
I put up an example what i would do in another comment (no idea how to link draft comments here)

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
```diff
@@ -855,6 +858,60 @@ final boolean isCompilationFailed() {
...
+                if (logger != null && runtime().logShutdownCompilations().compareAndExchange(true, false)) {
+                    logger.log(Level.WARNING, "Truffle compilations permanently disabled because of full code cache. "  +
+                            "Increase the code cache size using '-XX:ReservedCodeCacheSize=' and/or run with '-XX:+UseCodeCacheFlushing -XX:+MethodFlushing'.");
+                }
+                try {
+                    runtime().getCompileQueue().shutdownAndAwaitTermination(100 /* milliseconds */);
+                } catch (RuntimeException re) {
+                    // Best effort, ignore failure
+                }
+                return true;
```
chumer, line 908: Here is my suggestion:

Note instead of runtime we use the engine to log to avoid missing warnings. (warnings can be suppressed per engine)


```java
      if (engine.logShutdownCompilations().get()) {
                // The logger can be null if the engine is closed.
                if (engine.logShutdownCompilations().compareAndExchange(true, false)) {
                       TruffleLogger logger = engine.getLogger("engine");
                       if (logger != null) {
                          // log for each engine once
                      }
                }
      }
      // do not try to compile this call target again
      callTarget.compilationFailed = true;
      return true;
```

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
```diff
@@ -855,6 +858,60 @@ final boolean isCompilationFailed() {
...
+                if (sct == 0) {
+                    runtime().stoppedCompilationTime().compareAndSet(0, System.currentTimeMillis());
+                }
+                // Flush the compilations queue for now. There's still a chance that compilation
+                // will be re-enabled eventually, if the hosts code cache can be cleaned up.
+                Collection<OptimizedCallTarget> targets = runtime().getCompileQueue().getQueuedTargets(null);
+                // If there's just a single compilation target in the queue, the chance is high that it is
+                // the one we've just added after the StoppedCompilationRetryDelay ran out, so keep it to
+                // potentially trigger a code cache sweep.
+                if (targets.size() > 1) {
```
chumer, line 887: > the one we've just added 

I assume if this assumption is false, the compilation would fail since it has been stopped, right?
You basically cancel the compilations to avoid having them all fail?
What kind of failure is to be expected if it is stopped? A non permanent bailout? Any other failure could trigger OptimizedCallTarget.compilationFailed to be set to true, which would no longer call this method.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
```diff
@@ -881,6 +939,10 @@ public final boolean compile(boolean lastTierCompilation) {
                     return true;
                 }
 
+                if (isCompilationStopped()) {
```
chumer, line 942: I would move this check right before the synchronized block. Doing this in the synchronized block does not give us any advantage as its global anyway, but it is costly as its done on every call target call in the interpreter.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
chumer: On a second thought, it might be a good idea to return the enum here, so the compilationFailed flag can be set in OptimizedCallTarget.compile and not in a method called `isCompilationStopped` which is a bit surprising.

chumer (changes requested):

chumer:
@simonis Thank you for the changes!. did another turn. looks like its going to be the last one 👍

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
simonis: Bit I'll have to import it then. Nevertheless, I agree that the extra import is worth it in order to simplify this code.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
simonis: This code is based on the characteristics of the current HotSpot code cache sweeping. After [JDK-8290025: Remove the Sweeper](https://bugs.openjdk.org/browse/JDK-8290025) (i.e. since JDK 20), code cache sweeping can only be triggered by a full GC if certain conditions like the code cache allocation rate or the code cache fill state meet certain criteria. Each time before a new nmethod will be inserted into the code cache, these criteria are checked (see [`CodeCache::gc_on_allocation()`](https://github.com/openjdk/jdk/blob/ed03f0d9d10518242a3dc6e3685f1bdb0550c723/src/hotspot/share/code/codeCache.cpp#L776)) and if the criteria is met, a full GC will be triggered which may in turn trigger a sweep of the code cache. The nmethod age (and thus the condition on when nmethods will be removed from the code cache) is determined by the number of last such sweeping cycles they have been active.

The current algorithm is not perfect. E.g. once compilation gets disabled, there will be no more code cache checks due to `CodeCache::gc_on_allocation()`. Also, the code cache allocation rate will be zero after compilation has been disabled. Finally, if the application is not running full GCs because of a full heap for a long time (which is not uncommon for a well-tuned application), no more code cache sweeping will be triggered at all.

I plan to address some of these shortcoming, but for now it looked beneficial in my experiments, to periodically trigger a Truffle compilation (and thus a call to `CodeCache::gc_on_allocation()`), even if compilation is disabled. That's why I introduced the `StoppedCompilationRetryDelay` mechanism in the first place.

But in my experiments I observed that once I enter a new compilation into the queue because the `StoppedCompilationRetryDelay`  passed, in most cases it was almost immediately removed from the queue again before it could be picked up by a compiler thread because the next invocation of `OptimizedCallTarget::compile()` was just faster (i.e. it took the `STOP_COMPILATION` path and drained the queue). This means that the `StoppedCompilationRetryDelay` mechanism, in contrast to its design, almost never triggered a compilation.

The assumption in the comment, that a single entry in the queue was probably entered because `StoppedCompilationRetryDelay` timed out is only wrong at the beginning, when compilation gets stopped (i.e. the compilation activity mode changed from `RUN_COMPILATION` to `STOP_COMPILATION`). In that case we might unnecessarily leave at most one compilation task in the compilation queue, which we could actually remove. But that doesn't do any harm. The compilation will just fail because of the full code cache with the only draw-back that we've unnecessarily wasted some CPU cycles for the compilation.

simonis:
> @simonis Thank you for the changes!. did another turn. looks like its going to be the last one 👍 I know such features are notoriously hard to test. Do you have an idea how we could do at least a smoke test for this?

I can imagine something like the following:
- start the JVM with a very small code cache
- In a loop
  - trigger the compilation of a Truffle method
  - call `jcmd Compiler.codecache` on the current process
  - from the output, we can determine how many times compilation was dis-/enabled and how many nmethod installations have failed because of a full code cache.
  
It would be a little tricky to exclude all the JIT compilations and only allow truffle compilations so the results don't get influenced by JIT compilations.

Is this something acceptable as Truffle test? Do you already have similar tests (i.e. simple tests which repeatedly trigger the compilation of a single method, tests which call external tools and parse their output). I'm just asking because the HotSpot JTreg tests have frameworks and tooling for such kind of tests. I'm not familiar with the Truffle tests and don't want to re-invent the wheel so I'd be happy if you could point me to some tests which I could use a template to start with.

simonis:
@chumer, could you please have a look at the updated version?

chumer (approved):
Looks good. Thanks a lot for this contribution @simonis !
I will take it from here to integrate.

chumer:
Integrated: https://github.com/oracle/graal/pull/10358

simonis:
Fixes #10133

PR #10232, opened 2024-12-04
Include `Charset` for `sun.jnu.encoding` by default

fmeum:
The `Charset` specified by `sun.jnu.encoding` is required e.g. for `java.(n)io` path encoding/decoding as the image inherits `sun.jnu.encoding` from the build environment. On Windows, this will typically be a legacy code page charset, which is otherwise not added by default.


olpaw:
> On Windows, this will typically be a legacy code page charset, which is otherwise not added by default.

Makes sense to me. @fmeum did you check how much this increases the size of a helloworld image on Windows?

olpaw (changes requested):
@fmeum I also need a test case that would fail without this change.
See also https://github.com/oracle/graal/pull/10232#issuecomment-2544929492

PR #10301, opened 2024-12-12
[GR-60395] Allow combination of captureCallState and critical(true).

graalvmbot:


substratevm/src/com.oracle.svm.core.foreign/src/com/oracle/svm/core/foreign/Target_jdk_internal_foreign_abi_NativeEntryPoint.java
```diff
@@ -58,7 +58,8 @@ public static Target_jdk_internal_foreign_abi_NativeEntryPoint make(ABIDescripto
                     MethodType methodType,
                     boolean needsReturnBuffer,
                     int capturedStateMask,
-                    boolean needsTransition) {
+                    boolean needsTransition,
+                    boolean usingAddressPairs) {
```
zakkak, line 62: @fangerer is the lack of a `@BasedOnJDKFile` annotation on this method intentional?

fangerer: Yes, since the substitution class is annotated with `@TargetClass`, we track changes in this class anyway.

zakkak: Good, thank you for the explanation

Issue #10304, opened 2024-12-13
[Native Image] JDK-8336768 breaks `-H:+ForeignAPISupport`


zakkak:
### Describe the Issue

[JDK-8336768](https://bugs.openjdk.org/browse/JDK-8336768) changed the signature of `jdk.internal.foreign.abi.NativeEntryPoint#make` breaking the substitution in `com.oracle.svm.core.foreign.Target_jdk_internal_foreign_abi_NativeEntryPoint#make`.

### Using the latest version of GraalVM can resolve many issues.

- [X] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

### GraalVM Version

ca7e8b3abff44bff9b5e233f36bb10acd023e510

### Operating System and Version

Linux fedora 6.11.5-300.fc41.x86_64

### Build Command

[...]

PR #10305, opened 2024-12-13
Adapt JDK-8336768: Allow captureCallState and critical linker options to be combined

zakkak:
Closes: https://github.com/oracle/graal/issues/10304

> [!NOTE]
> If merged, this should be backported to 24.2 as well


mur47x111:
I assume this needs to be backported to JDK24, right?

mur47x111:
Also could you please add a test case for us to notice this earlier? Thanks!

fangerer:
fyi: this PR is obsolete since #10301 is already merged.

zakkak:
> Also could you please add a test case for us to notice this earlier? Thanks!

I will have a look at it.

Issue #6253, opened 2023-03-21
Cryptic parsing error messages with `--link-at-build-time` and incomplete classpath


zakkak:
**Describe the issue**
Parsing errors may be cryptic when using `--link-at-build-time` and incomplete classpath.
For instance if a `Class A` cannot be initialized due to a `NoClassDefFoundError` caused by a missing dependency, the user will be presented with an `UnresolvedElementException` stating that some method of `Class A` is unresolved, but it won't provide any additional info on the cause.

**Steps to reproduce the issue**
```
cd /tmp
git clone --branch 2023-03-21-graal-4661-parsing-error https://github.com/zakkak/issue-reproducers reproducer
cd reproducer
mvn package
export JAVA_HOME=/path/to/graalvm
# Attempt to generate native-image
$JAVA_HOME/bin/native-image \
  --no-fallback \
  -H:+ReportExceptionStackTraces \
  --link-at-build-time \
  --initialize-at-build-time=. \
  -jar target/reproducer-1.0-SNAPSHOT.jar
```

**Describe GraalVM and your environment:**
 - GraalVM version (latest snapshot builds can be found [here](https://github.com/graalvm/graalvm-ce-dev-builds/releases)), or commit id if built from source: **e9893c06ff637ec859c18dcba53598ca51933af1**
 - JDK major version: **20**
 - OS: **Fedora 37**
 - Architecture: **AMD64**

**More details**

The reproducer will result in build errors like the following:

```
Error: Unsupported features in 2 methods
Detailed message:
Error: Discovered unresolved method during parsing: net.sf.oval.expression.ExpressionLanguageJavaScriptImpl.<init>(). This error is reported at image build time because class net.sf.oval.expression.ExpressionLanguageRegistry is registered for linking at image build time by command line.
Error encountered while parsing net.sf.oval.expression.ExpressionLanguageRegistry.getExpressionLanguage(ExpressionLanguageRegistry.java:78) 
Parsing context:
   at net.sf.oval.AbstractCheck.isActive(AbstractCheck.java:151)
   at net.sf.oval.Validator.checkConstraint(Validator.java:763)
   at net.sf.oval.Validator._validateStaticInvariants(Validator.java:670)
   at net.sf.oval.Validator.validateInvariants(Validator.java:1382)
   at net.sf.oval.Validator.validate(Validator.java:1308)
   at gdv.xport.feld.Feld.validateInvariants(Feld.java:623)
[...]

PR #9856, opened 2024-10-10
Add opt-in link-at-build-time option to fail fast on NoClassDefFound

zakkak:
This is a resurrection of https://github.com/oracle/graal/pull/6725 as https://github.com/oracle/graal/issues/6253 turns out to not be completely fixed after all (see https://github.com/oracle/graal/issues/6253#issuecomment-2402138210).

According to @fniephaus in https://github.com/oracle/graal/pull/6753#issuecomment-1632386911

> We have an idea that no longer requires a static map, but I haven't had enough time to look into this yet.

So I asked for more details to see if I can implement that idea and get a proper fix in.

In the meantime, #6253 (originally reported in #4661) is a long standing issue for Quarkus users and we would really like a fix for it (even a work around like the one I propose). Note that by default Quarkus initializes and links all classes at build time, so this is not a corner case for Quarkus.

This PR works around #6253, by providing users the option to _fail fast_ if a class is missing and they are linking at build time. The option is intentionally made an opt-in in order to not alter the current behavior till a _proper_ fix is available and it should be completely removed once we get that in.

To try this out you may run:

```
cd /tmp
git clone --branch 2024-10-09-oracle-graal-6253-reproducer https://github.com/zakkak/issue-reproducers reproducers
cd reproducers
mvn package
# run in JVM
java \
  -jar target/reproducer-1.0-SNAPSHOT.jar
# generate native-image with --link-at-build-time
native-image \
  --no-fallback \
  --link-at-build-time  \
  -H:+ReportExceptionStackTraces \
  -jar target/reproducer-1.0-SNAPSHOT.jar
# Build fails
```

<details>
<summary>Passing `-H:+LinkAtBuildTimeFailFast` results in the following output:</summary>

```
Error: Error during parsing of method com.microsoft.sqlserver.jdbc.SQLServerConnection.setKeyVaultProvider(java.lang.String, java.lang.String). This error is reported at image build time because class com.microsoft.sqlserver.jdbc.SQLServerConnection is registered for linking at image build time by command line and command line.
Error encountered while parsing com.microsoft.sqlserver.jdbc.SQLServerConnection.connectInternal(SQLServerConnection.java:2205) 
Parsing context:
   at com.microsoft.sqlserver.jdbc.SQLServerConnection.connect(SQLServerConnection.java:1971)
   at com.microsoft.sqlserver.jdbc.SQLServerDriver.connect(SQLServerDriver.java:1263)
   at root method.(Unknown Source)

[...]

PR #10320, opened 2024-12-16
[GR-60554] Improve causalilty reporting for unresolved elements.

graalvmbot:


zakkak:
Hi @dougxc do I get it right that this PR resolves the same issue that https://github.com/oracle/graal/pull/9856/files tries to resolve?

Could you please mirror the internal description as well?

dougxc:
> Hi @dougxc do I get it right that this PR aim is to resolve #6253?

It certainly goes a long way in helping. There's also https://bugs.openjdk.org/browse/JDK-8346282 which might allow Native Image to further improve transparency of linkage errors. @cstancu may be able to comment better on that.

In terms of this comment:

> we would need to preserve resolution and linking errors via JVMCI, where they are currently just ignored.

The one instance of this I'm aware of is [being fixed](https://github.com/openjdk/jdk/pull/22767/files#diff-b782878562668748c5c59acc2e937f7c24de4529b8a74bd3a4eae83fa0e07846L801). If there are other cases, please let me know.

zakkak:
Thank you @dougxc.

> The one instance of this I'm aware of is [being fixed](https://github.com/openjdk/jdk/pull/22767/files#diff-b782878562668748c5c59acc2e937f7c24de4529b8a74bd3a4eae83fa0e07846L801). If there are other cases, please let me know.

Sure

PR #10344, opened 2024-12-18
[GR-58574] Add SVM interpreter and JDWP support for Native Image

graalvmbot:
New option to enable JDWP support for an image: -H:+JDWP.
The JDWP implementation is split into two parts in order to reduce application image size.

- svm.interpreter: Adds a Java bytecode interpreter to SVM, origins from Espresso.
- jdwp.resident: Part of the JDWP implementation that lives in an application image.
- jdwp.server: Part of JDWP that handles the communication to a debugger. Can run either as native-image (libsvmjdwp) or on HotSpot (development only).
- jdwp.common: Shared bits.


He-Pin:
Finally have this!

PR #10348, opened 2024-12-19
Update ideal-graph-visualizer.md

JensLincke:
add missing jdk installtation step, since it is very picky to use jdk 21


oubidar-Abderrahim:
Hi @olyagpl do you think this is still relevant? if yes, please review the PR, Thank you

olyagpl:
We explicitly support JDK 24 so I think this is obsolete. 
Thank you!

PR #10358, opened 2024-12-19
[GR-60636] Implement Truffle compiler control based on HotSpot's CompileBroker compilation activity.

graalvmbot:


simonis:
Fixes #10133

PR #10364, opened 2024-12-20
[GR-60548] Cleanly separate word box types by class loader.

graalvmbot:


compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/HotSpotTTYStreamProvider.java
```diff
@@ -69,7 +68,7 @@ public PrintStream getStream() {
     }
 
     static {
-        Word.ensureInitialized();
+        /* Calling this method ensures that the static initializer has been executed. */
```
toinouH, line 71: Was the deletion intentional here ?

Issue #7095, opened 2023-08-01
Trim JFR event stack traces


roberttoyonaga:
**Feature request**

**Is your feature request related to a problem? Please describe.**
Currently, JFR stacktraces include all frames up to the point the stack walk is done.
![image](https://github.com/oracle/graal/assets/35396843/92aefaa2-9740-4621-8f7b-873ffb7c69ef)
This means that the last few frames in the stacktrace always display internal code and are probably not interesting to the user. This makes flamegraphs more cluttered than necessary and does not match the behavior of JFR in Java mode (shown below).  
![image](https://github.com/oracle/graal/assets/35396843/6270ed44-d929-446e-83cb-2ab81c3e3760)

**Describe the solution you'd like.**
It might be useful if the stacktraces were trimmed for each event where stacktraces are enabled such that they are similar to the ones obtained running in Java mode. The infrastructure to do this is already in place so the work should be very minimal. Setting the stacktrace `skipcount` from the point of event emission for each event should do the trick.  

**Describe who do you think will benefit the most.**
JFR users. 

**Describe alternatives you've considered.**
1). Continue collecting the full stacktrace. This might be useful for those developing GraalVM. Collecting the full stacktrace also avoids having to update a hard-coded `skipcount` whenever the underlying JFR implementation changes (but this is probably a minor benefit). 

2). Only trim the last ~3 frames that pertain to the actual stacktrace generation. Keep the other internal frames to assist with GraalVM development and debugging.  

I think any of these options are reasonable, but that it's probably worth bringing up for discussion. 

**Express whether you'd like to help contributing this feature**
I can do this, if we decide this change is desirable.

jbachorik:
If I can chime in - for a common profiler user those extra frames are quite distracting as they are the internal implementation of collecting the stacktrace. When this data is aggregated one can easily end up with the bottleneck pointing to the stacktrace collection code which is both incorrect and undesired.

I can imagine, for the Graal developers, there would be a config knob (eg. a system property) to set which would enable recording unredacted stack, but for the common use case it would be really helpful to have the frame where the stacktrace was collected to become the leaf of the stacktrace.

PR #10385, opened 2024-12-27
[GR-61151] Trim internal JFR stacktraces to match OpenJDK

roberttoyonaga:
### Summary:
OpenJDK JFR does not record VM level frames. Previously, Native Image JFR reported internal SubstrateVM frames in stacktraces. This is bad because it makes it confusing/difficult for users to see where events are coming from since the ends of every trace look the same. For example, this adds noise to flamegraphs. It can also mislead to users to believing that SubstrateVM internals are the cause of event emissions if they are accustomed to the way stacktraces are reported in OpenJDK.  

This PR adds trimming of internal JFR stacktraces by default. I've also added a runtime option to disable this (and revert to the old behavior) to aid in Native Image development/debugging. Now, any SubstrateVM frames in samples taken by the CPU profiler or regular events are filtered out. 

Use `-XX:-JfrTrimInternalStackTraces` to disable the filtering at runtime.

Related Issue: https://github.com/oracle/graal/issues/7095

### Note:
In this PR, the filtering is done at the time of stacktrace serialization. This happens at a chunk rotation. The problem with this approach is that we have already computed the stacktraces up to the maximum depth specified by the user (default is 64). So, it the point of sampling, we may need to truncate at the max depth, then later we may further reduce the stacktrace size by filtering during serialization. This means we would have truncated some frames unnecessarily. In reality, this is probably not much of an issue since there's only 3-5 internal frames being filtered out for most events.  I've considered the below alternatives:

1. Use "skipCount" instead of filtering. This would require maintaining a hardcoded skipcount for each event type. For some events (such ObjectAllocationSample) multiple paths could be taken leading up to event emission so a single event may require different hardcoded skipcounts depending on the code path taken. I don't like this approach because it becomes harder to maintain as more events are added and the internal code changes.  
2. Filter earlier at the time of stack-walking (long before stacktrace serialization). This would allow us to avoid truncating traces unless truly necessary. The problem with this is that we'd need to synchronize access to a `FrameInfoCursor` instance. This isn't possible because we need to do the stack-walking inside the JFR sampler signal handler. It's also not possible to allocate a new `FrameInfoCursor`  whenever we need to use it. One solution is to preallocate an instance of `FrameInfoCursor` for each thread and use `JfrExecutionSampler.preventSamplingInCurrentThread()` to prevent races if preempted for signal handling.  But this seems overly complex and unnecessary compared to the current solution.


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/sampler/SamplerJfrStackTraceSerializer.java
```diff
@@ -167,8 +169,12 @@ private static int visitFrame(JfrNativeEventWriterData data, CodeInfo codeInfo,
         int numStackTraceElements = 0;
         FRAME_INFO_CURSOR.initialize(codeInfo, ip, false);
         while (FRAME_INFO_CURSOR.advance()) {
+            FrameInfoQueryResult frame = FRAME_INFO_CURSOR.get();
+            if (SubstrateJVM.getStackTraceRepo().getTrimInternalStackTraces() &&
```
fniephaus, line 173: Since it's constant, I prefer moving the condition before the loop, even if the compiler can do that for us ;)

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/sampler/SamplerJfrStackTraceSerializer.java
```diff
@@ -167,8 +169,12 @@ private static int visitFrame(JfrNativeEventWriterData data, CodeInfo codeInfo,
...
+                            UninterruptibleUtils.String.startsWith(frame.getSourceClassName(), SUBSTRATEVM_PREFIX)) {
+                continue;
+            }
             if (data.isNonNull()) {
-                FrameInfoQueryResult frame = FRAME_INFO_CURSOR.get();
                 serializeStackTraceElement(data, frame);
```
fniephaus, line 178: Should this additional check be within the `if (data.isNonNull()) {` branch or not?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrStackTraceRepository.java
```diff
@@ -71,6 +71,7 @@ public class JfrStackTraceRepository implements JfrRepository {
     private final JfrStackTraceEpochData epochData1;
 
     private int stackTraceDepth;
+    private boolean trimInternalStackTraces;
```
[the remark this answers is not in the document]
roberttoyonaga: Yes it's a bit unfortunate. It gets set during startup once the runtime options are parsed. We also need to change this during the JFR tests.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/sampler/SamplerJfrStackTraceSerializer.java
roberttoyonaga: I think this condition must be evaluated loop iteration because the outcome may change whenever we advance the `FRAME_INFO_CURSOR` to check the next frame.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/sampler/SamplerJfrStackTraceSerializer.java
roberttoyonaga: Yes, maybe moving it inside `if (data.isNonNull())` would be a small optimization

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/sampler/SamplerJfrStackTraceSerializer.java
roberttoyonaga: Hmm I'm not sure we can move the conditional out of the loop since we must recheck it with each new advancement of `FRAME_INFO_CURSOR`.   `SubstrateJVM.getStackTraceRepo().getTrimInternalStackTraces()` is constant but we can't even move that part outside since we still need to execute the loop to potentially serialize frames.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/sampler/SamplerJfrStackTraceSerializer.java
roberttoyonaga: Actually, I don't think it should be moved inside the `if (data.isNonNull())` check. We must increment `numStackTraceElements` regardless of the if (data.isNonNull())` check, but must not increment if we are trimming the current stacktrace.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/sampler/SamplerJfrStackTraceSerializer.java
fniephaus: Sorry, I was only talking about moving `SubstrateJVM.getStackTraceRepo().getTrimInternalStackTraces()` outside the loop.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrStackTraceRepository.java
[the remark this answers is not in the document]
fniephaus: Would it make sense to turn the option from a `RuntimeOptionKey` into a `HostedOptionKey` and make sure trimming is installed at build time? Is it really necessary for the option to be mutable at run time?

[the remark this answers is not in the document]
roberttoyonaga: Yes, I think it could make sense to set this option at build time with a `HostedOptionKey`. The 2 downsides would be:

1. We'd need to remove the test that checks trimming stacktraces, since we won't be able to toggle the option dynamically at runtime.
2. Users will need to rebuild their executables if they discover that they need full stacktraces. This is probably not much of a problem since I think most users won't need that.

Do you think it's worth it?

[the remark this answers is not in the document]
fniephaus: RE 1: I guess you could unconditionally disable trimming in `native_unittests_task` in `mx_substratevm.py`. That should do the try on the gate. When you want to run `mx native-unittest` on the cmd line, you need to pass it as an additional build flag.
RE 2: Agreed, normal users are probably not too interested in this. Disabling trimming only really seems useful to measure JFR overhead, something only we really care about, right?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/sampler/SamplerJfrStackTraceSerializer.java
```diff
@@ -167,8 +170,12 @@ private static int visitFrame(JfrNativeEventWriterData data, CodeInfo codeInfo,
         int numStackTraceElements = 0;
         FRAME_INFO_CURSOR.initialize(codeInfo, ip, false);
         while (FRAME_INFO_CURSOR.advance()) {
+            FrameInfoQueryResult frame = FRAME_INFO_CURSOR.get();
+            if (VMInspectionOptions.JfrTrimInternalStackTraces.getValue() &&
```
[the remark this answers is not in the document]
roberttoyonaga: Thank you for your feedback. How do you feel about using `skipCount` to trim only internal JFR frames on SubstrateVM level events?   This would be just the top ~3 frames and not affect CPU profiler samples or JDK events. This would avoid filtering out any middle frames, and also avoid unnecessarily shortening stacktraces 
due to filtering at serialization time. Limiting the trimming to only JFR frames while keeping other internal frames would also make it easier to maintain. JDK events use a [hardcoded skipcount](https://github.com/openjdk/jdk/blob/jdk-25%2B4/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformEventType.java#L101-L111) too.

This wouldn't exactly match what users see in OpenJDK, but would at least get rid of the top few redundant internal JFR frames.  

Alternatively, like you mentioned, we could filter from the top of the stack down until we hit a known method. And only do this for VM-level events while walking the stack.  This involves more steps since it must use `FrameInfoCursor` while stack walking, but it's nice that it avoids needing to manually set a skipcount.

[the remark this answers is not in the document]
christianhaeubl: I think it makes sense to give the hardcoded skipcount a try as it is the simplest approach. If that turns out to be a maintenance burden, we can always implement a more complex approach later on.

We will need fairly good tests for the stack traces of the different JFR events though. Otherwise, it is probably very easy to introduce a regression.

roberttoyonaga:
Note: `truffle JDK21` GHA failing due to failure downloading Eclipse. Seems unrelated to this PR.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/sampler/SamplerJfrStackTraceSerializer.java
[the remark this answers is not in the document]
roberttoyonaga: I've implemented the skipcount approach and added tests for all the events affected to ensure we don't end up skipping too many frames. I left the `JfrTrimInternalStackTraces` hosted option set to true by default.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrEvent.java
```diff
@@ -32,13 +32,16 @@
...
 import com.oracle.svm.core.thread.JavaThreads;
+import com.oracle.svm.core.VMInspectionOptions;
 
 /**
  * This file contains the VM-level events that Native Image supports on all JDK versions. The event
  * IDs depend on the JDK version (see metadata.xml file) and are computed at image build time.
  */
 public final class JfrEvent {
-    public static final JfrEvent ThreadStart = create("jdk.ThreadStart");
+    private static final int defaultInternalSkipCount = VMInspectionOptions.JfrTrimInternalStackTraces.getValue() ? 3 : 0;
```
christianhaeubl, line 42: Can we apply the skip count to all JFR events or are there any events that need to be excluded explicitly?

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestThreadParkEvents.java
```diff
@@ -79,6 +80,11 @@ private static void validateEvents(List<RecordedEvent> events) {
         boolean parkUnparkFound = false;
 
         for (RecordedEvent event : events) {
+            List<RecordedFrame> frames = event.getStackTrace().getFrames();
+            assertTrue(frames.size() > 0);
+            System.out.println(frames.getFirst().getMethod().getName());
```
christianhaeubl, line 85: Is this on purpose or is it a leftover from debugging?

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestAllocationRequiringGCEvent.java
```diff
@@ -54,6 +54,9 @@ public void test() throws Throwable {
 
     private static void validateEvents(List<RecordedEvent> events) {
         assertTrue(events.size() > 0);
+        for (RecordedEvent event : events) {
+            checkStackTraceTrimming(event, "emit");
```
christianhaeubl, line 58: Shouldn't we trim the "emit" methods as well?

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrEvent.java
roberttoyonaga: I think it would be fine to apply the skipcount to all events defined in SubstrateVM. The periodic ones and others that don't contain stacktraces would just ignore the skipcount.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestThreadParkEvents.java
roberttoyonaga: Yes, that should have been replaced with `checkStackTraceTrimming` . Thanks for spotting that.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestAllocationRequiringGCEvent.java
roberttoyonaga: I wanted to trim the "emit" methods as well, but when I increment the skipcount, it seems to trim 2 frames. Maybe this is due to inlining? I'm not entirely sure.

roberttoyonaga: So I decided it was better to trim slightly less and keep more information in the stacktrace.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrEvent.java
christianhaeubl: Ok, then let's use the skipCount for all events. Please remove the static field `defaultInternalSkipCount` and move its logic directly into `getSkipCount()`.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestAllocationRequiringGCEvent.java
christianhaeubl: > it seems to trim 2 frames

Interesting - it seems that the `skipCount` logic is broken at the moment as it is applied on the level of physical stack frames (instead of virtual frames). We need to fix that, otherwise, `skipCount` will be pretty much unpredictable.

We would probably need to do something like the following:
- remove the `skipCount` from `JfrStackWalker` (at that point in time, we only know about the physical stack frames)
- modify `JfrStackWalker` so that it stores the requested skip count at the very beginning of the sampler data (i.e., before any recorded IPs)
- when serializing the raw IPs into JFR stack traces, we need to apply the skip count on the level of Java frames

Please let me know if you can look into that.

roberttoyonaga: Yes, I can look into this. I'll add that change to this pull request (unless you think it's cleaner separately).

roberttoyonaga: We also probably need to delay when we do truncation to the time of serialization. Otherwise we may truncate unnecessarily.

christianhaeubl: Thanks!

I think that 2 different kinds of truncation is needed for the bottom-most frames:
- `JfrStackWalker`: truncate so that we have at most `skipCount + getMaxDepth()` physical frames (each physical frame corresponds to at least 1 virtual frame)
- during serialization: apply skipCount to remove the top frames and truncate the bottom-most frames so that we have at most `getMaxDepth()` virtual frames.

roberttoyonaga: Right, I see. I'll work on adding this. Thanks!

roberttoyonaga:
Hi @christianhaeubl, I've implemented the fix for the skipcount bug we talked about. Please have another look when you have time.  Thank you!

roberttoyonaga:
Hi @christianhaeubl, just pinging you here to keep this fix on your radar.

christianhaeubl:
Thanks, I started integrating this PR. I'll let you know if I run into any issues.

PR #10444, opened 2025-01-09
Run the Quarkus workflow weekly instead of nightly

zakkak:
Reduce the frequency we run the Quarkus tests as they are quite resource hungry and lengthy.


fniephaus (approved):
LGTM, thank you!

PR #10510, opened 2025-01-18
Use offsets from `.bgv` file to select appropriate text

JaroslavTulach:
Enso is trying to [generate IGV graph](https://github.com/enso-org/enso/pull/12061#issuecomment-2598666306) for our internal `IR`:

- https://github.com/enso-org/enso/pull/12061 

We had issues to provide proper `nodeSourceLocation` information. I had to debug visualizer code to find out what's wrong. While doing that I made few fixes. 

-  we were providing `-1` as line location - IGV was then doing nothing - with 75d41ee8179929bdda1c5fed3220e6ef357092ca it at least opens the file
- I was getting `NullPointerException`s so I am adding checks and `assert` to catch that earlier
- Last, but not least: we do provide _start and end_ offsets for Enso nodes. Let's use them!

Location used

With these changes IGV picks up `start` and `end` offset from `location` property and is capable to select appropriate source range. CCing @tkrodriguez, @sdedic, @Akirathan

### DIFF
--- a/visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
+++ b/visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
@@ -23,6 +23,9 @@
 
 package org.graalvm.visualizer.source.impl.ui;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import org.graalvm.visualizer.source.Location;
 import org.graalvm.visualizer.source.ui.Trackable;
 import org.netbeans.api.actions.Openable;
@@ -34,14 +37,20 @@ import org.openide.filesystems.FileObject;
 import org.openide.text.Line;
 import org.openide.util.NbBundle;
 
-import javax.swing.SwingUtilities;
 import javax.swing.text.Document;
 import javax.swing.text.JTextComponent;
+import javax.swing.text.StyledDocument;
+import org.openide.text.Annotatable;
+import org.openide.text.Annotation;
+import org.openide.text.NbDocument;
+import org.openide.util.Mutex;
+import org.openide.util.Task;
+import org.openide.util.TaskListener;
 
 /**
  *
  */
-public class LocationOpener implements Openable, Trackable {
+public final class LocationOpener implements Openable, Trackable {
     private final Location location;
 
     public LocationOpener(Location location) {
@@ -57,31 +66,48 @@ public class LocationOpener implements Openable, Trackable {
     }
 
     private void openOrView(boolean focus) {
-        FileObject toOpen;
-        toOpen = location.getOriginFile();
+        FileObject toOpen = location.getOriginFile();
         if (toOpen == null) {
             return;
         }
-        int line = location.getLine();
         EditorCookie cake = toOpen.getLookup().lookup(EditorCookie.class);
+        if (cake == null) {
+            return;
+        }
 
-        Line l;
-        try {
-            l = cake.getLineSet().getOriginal(line - 1);
-        } catch (IndexOutOfBoundsException ex) {
-            // expected, the source has changed
-            return;
+        Task task = cake.prepareDocument();
+        class WhenShowing implements TaskListener, Runnable {
+            @Override
+            public void taskFinished(Task task) {
+                task.removeTaskListener(this);
+                Mutex.EVENT.postReadRequest(this);
+            }
+
+            @Override
+            public void run() {
+                final StyledDocument doc = cake.getDocument();
+                if (doc == null) {
+                    return;
+                }
+                List<Annotatable> select = findLinesOrParts(cake.getLineSet(), doc, location.getLine() - 1, location.getOffsetsOrNull());
+                if (select.size() == 1 && select.get(0) instanceof Line) {
+                    Line line = (Line) select.get(0);
+                    line.show(Line.ShowOpenType.REUSE, focus ? Line.ShowVisibilityType.FRONT : Line.ShowVisibilityType.FRONT);
+                    CurrentNodeAnnotation.highlight(select);
+                } else if (select.size() >= 1 && select.get(0) instanceof Line.Part) {
+                    Line.Part part = (Line.Part) select.get(0);
+                    part.getLine().show(Line.ShowOpenType.REUSE, focus ? Line.ShowVisibilityType.FRONT : Line.ShowVisibilityType.FRONT, part.getColumn());
+                    CurrentNodeAnnotation.highlight(select);
+                } else {
+                    // neither line nor offsets
+                    cake.open();
+                    StatusDisplayer.getDefault().setStatusText(Bundle.ERR_LineNotFound());
+                }
+            }
+
         }
-        if (l == null) {
-            cake.open();
-            StatusDisplayer.getDefault().setStatusText(Bundle.ERR_LineNotFound());
-            return;
-        }
-        if (SwingUtilities.isEventDispatchThread()) {
-            l.show(Line.ShowOpenType.REUSE, focus ? Line.ShowVisibilityType.FRONT : Line.ShowVisibilityType.FRONT);
-        } else {
-            SwingUtilities.invokeLater(() -> l.show(Line.ShowOpenType.REUSE, focus ? Line.ShowVisibilityType.FRONT : Line.ShowVisibilityType.FRONT));
-        }
+        WhenShowing select = new WhenShowing();
+        task.addTaskListener(select);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
```diff
@@ -65,22 +66,24 @@ private void openOrView(boolean focus) {
...
+        Mutex.EVENT.readAccess(() -> {
             l.show(Line.ShowOpenType.REUSE, focus ? Line.ShowVisibilityType.FRONT : Line.ShowVisibilityType.FRONT);
-        } else {
-            SwingUtilities.invokeLater(() -> l.show(Line.ShowOpenType.REUSE, focus ? Line.ShowVisibilityType.FRONT : Line.ShowVisibilityType.FRONT));
+        });
+    }
+
+    private Line findLine(EditorCookie cake, int line) {
+        try {
+            return cake.getLineSet().getOriginal(line - 1);
```
JaroslavTulach, line 82: Possibly IGV should honor https://www.graalvm.org/graphio/javadoc/jdk/graal/compiler/graphio/GraphLocations.html#locationOffsetStart(java.lang.Object) when the line isn't positive.

visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
```diff
@@ -63,24 +67,48 @@ private void openOrView(boolean focus) {
             return;
         }
         int line = location.getLine();
+        int[] offsets = location.getOffsetsOrNull();
```
JaroslavTulach, line 70: Text based languages (like Truffle ones) are likely to fill offsets rather than line numbers. Passing the offsets to the `LocationOpener` in faba84a and using the offsets rather than line when present. More an experiment for your consideration than bullet proof code right now.

visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
```diff
@@ -63,24 +67,44 @@ private void openOrView(boolean focus) {
...
         }
+
+        Task task = cake.prepareDocument();
+
+        TaskListener[] whenShowing = new TaskListener[1];
+        whenShowing[0] = (ev) -> {
+            task.removeTaskListener(whenShowing[0]);
+            Mutex.EVENT.postReadRequest(() -> {
+                if (offsets == null) {
+                    StatusDisplayer.getDefault().setStatusText(Bundle.ERR_LineNotFound());
```
tkrodriguez, line 83: I'm not sure if I'm understanding the control flow here but it looks to me that if the location had a valid line but no offsets you'd report `ERR_LineNotFound`.  Am I missing some subtlety?

visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
tkrodriguez: I don't have much opinion since I never use this functionality so you should make it work in some rational way.  If the two ways of addressing the text disagree then it's unclear what you should do.  Naively I might assume that the line number is more to be trusted given that offsets seem a bit squishy given differences in line terminations between windows and unix, plus the lack of actual terminator characters in a UI element that displays text.  How are these mismatches expected to be dealt with?

visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
JaroslavTulach: - I think you are right Tom, 8955e696540584b49748dd247290bee1b0aa892a would address that
- I have to admit I am not sure if `JEditorPane.select` is the right way to highlight the selection
- maybe there is something better like attaching some annotation to a [Line.Part](https://bits.netbeans.org/dev/javadoc/org-openide-text/org/openide/text/Line.Part.html)
- I can try to implement whatever you guys think is the best way to handle the highlight

visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
sdedic: re. different newline terminators: the information about the OS the code was compiled in is lost during the compilation step, and cannot be even dumped to BGV. Maybe a hint for `truffle` to represent positions using line:column pair rather than offsets.

From the user's standpoint I agree that the whole line is too coarse; it's far better to point at a specific position, if we have such data.

I suggest to make *an option* for this: if the IGV knows (the user configures it) the line terminator use on the compiling machine, it can then recompute the offset to this machine's newline style and to line:column coordinates.

visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
```diff
@@ -62,25 +66,48 @@ private void openOrView(boolean focus) {
...
+            lineNumber = NbDocument.findLineNumber(cake.getDocument(), offsets[0]);
+        }
+
+        Task task = cake.prepareDocument();
+        Line line = findLine(cake, lineNumber);
+
+        TaskListener[] whenShowing = new TaskListener[1];
+        whenShowing[0] = (ev) -> {
+            task.removeTaskListener(whenShowing[0]);
+            Mutex.EVENT.postReadRequest(() -> {
```
sdedic, line 82: after the document loads, please check that the offsets are still within document length; if not, fall back to just the line.

JaroslavTulach:
Thank you for the reviews. Yup, line endings may cause some misalignment. Let's try:
- if there are no offset information, use line (just like we did so far)
- if there is no line information, use offset (possibly accepting misalignment of Windows)
- if there are both information present
    - compute the line from offset
        - if the lines are different - use just line (to keep current behavior)
        - if the lines are the same - `select` the column region on the line 

That way we favor existing behavior and enhance it when additional information is present. I'll write some tests and rewrite the logic now.

visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
JaroslavTulach: Done in 9ee80e5

JaroslavTulach:
> Maybe a hint for truffle to represent positions using line:column pair rather than offsets.

Shouldn't we rather specify that _offsets_ in the IGV treat new line as a single character? That's how Truffle `Source` does it and also how NetBeans/Swing `javax.swing.text.Document` does it. Things will be simpler then.

Given Enso is the first system trying to use the offsets, we can make some small adjustments to the specification to make things easier.

visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
JaroslavTulach: I hope we'll find a way that works for current as well as future use-cases without a configuration option. I hope https://github.com/oracle/graal/pull/10510#issuecomment-2629232980 is a sound approach to achieve so.

JaroslavTulach:
With c4f059d I like what I see, CCing @Akirathan:

Selection of just *

Looks like with this PR the IGV _can work for [Enso IR graphs](https://github.com/enso-org/enso/pull/12061) well enough_. I hope that with https://github.com/oracle/graal/pull/10510#issuecomment-2629232980 we have a _good compromise to keep the existing behavior_ unmodified.

Can you do some testing on your side and possibly integrate this enhancement? It would increase our (Enso compiler team) productivity when exploring [Enso IR graphs](https://github.com/enso-org/enso/pull/12061) a lot. Btw. We'd be in a need of a new IGV release then...

JaroslavTulach:
With ff264d9 we properly select multi line selection: 

multi line

JaroslavTulach:
I've just squashed the _highlighting parts of this PR_ into a32bff9 with the hope it all will be found acceptable for integration.

visualizer/IdealGraphVisualizer/SourceRepository/src/main/java/org/graalvm/visualizer/source/impl/ui/LocationOpener.java
JaroslavTulach: I cannot find the info explicitly spelled in the [javadoc of Truffle Source](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/source/Source.html), but I believe a _newline_ separator is always treated as a single character when counting _offset_ in the source.

NetBeans does treat _newline_ as a single character as well.

As such I believe this protocol shall do the same: count _newline_ as a single character when it comes to offsets.

JaroslavTulach:
With 293308c I see

line & offset

e.g. both line and offset being selected at once.

I am happy with such a state. @tkrodriguez, can you integrate this change?

JaroslavTulach:
@tkrodriguez, can you restart [this job](https://github.com/oracle/graal/actions/runs/13582152834/job/37969970378?pr=10510#step:14:1951) and once it is green, merge my PR? Thank you in advance for that or for additional review comments.

tkrodriguez:
Yes I can integrate it. I've been out of the office unexpectedly but I'll prepare an internal PR.  Sorry for the delay.

tkrodriguez:
Would you mind rebasing onto latest master?

tkrodriguez:
Also I think your new assert broke `BinaryReaderMockTest`.  `MockElements.methodDeclaringClass` always returns null and it's not clear to me how to implement it.

JaroslavTulach:
> Would you mind rebasing onto latest master?

I don't know what _rebasinc onto master_ is good for, but done in b294b4a10c1ad8cb8b413569b834a15d1057d9b8

> assert broke BinaryReaderMockTest

OK, I'll take a look.

PR #10522, opened 2025-01-21
[GR-54697] Implement debuginfo generation at image-runtime.

graalvmbot:
On Linux native-image supports generating images with dwarf-debuginfo. This allows advanced users to debug image execution with GDB.

Since debuginfo is generated as part of the image build process it is only providing debug support for the AOT compiled methods in an image. For methods that get compiled at image run-time (as part of truffle runtime compilation) there is currently no debuginfo available.

To fix this problem the process of generating debuginfo must also happen at image runtime whenever a new method gets compiled and is about to be registered in the run-time code-cache. This requires code that is currently only used at build-time (hosted) to be made usable also for inclusion in the image itself. The relevant parts of the hosted code need to be refactored as needed.

Once runtime-debuginfo is generated GDB needs to be notified about it. This needs to happen via the [GDB JIT Interface](https://sourceware.org/gdb/current/onlinedocs/gdb.html/JIT-Interface.html). Note that it also required that GDB needs to be notified that run-time debuginfo is no more valid in case a run-time compiled method gets removed from the run-time code-cache (e.g. due to deoptimization).

The existing debuginfo testing needs to be extended to also test the run-time debuginfo generation.

Caveat: For GDB to be able to resolve type references from run-time debuginfo to AOT debuginfo, currently a modified version of GDB is required. https://github.com/dominikmascherbauer/binutils-gdb
We are working on getting the required changes upstream.


olpaw:
Even if we don't get the GDB changes upstream in time this PR should be merged soon. The Debuginfo generation speedup due to parallelization plus the fix to make our DWARF5 work with GDB 16 or later justifies having this on master.

olpaw:
@adinn this PR will likely be merged mid or end of next week. If you have any questions or comments, this would be a good time. cc @dominikmascherbauer

olpaw (approved):

olpaw:
> Caveat: For GDB to be able to resolve type references from run-time debuginfo to AOT debuginfo, currently a modified version of GDB is required. https://github.com/dominikmascherbauer/binutils-gdb
We are working on getting the required changes upstream.

This is not required anymore as we now use GDB opaque types to refer to already at-build-time-defined types from the run-time debuginfo generator.

PR #10513, opened 2025-01-20
[GR-61425] Update labsjdk to 25+6-jvmci-b01

graalvmbot:
Update labsjdk to 25+6-jvmci-b01

Adopted:
[JDK-8347287](https://bugs.openjdk.org/browse/JDK-8347287) - JFR: Remove use of Security Manager - [PR](https://github.com/openjdk/jdk/pull/22787)
[JDK-8347129](https://bugs.openjdk.org/browse/JDK-8347129) - cpuset cgroups controller is required for no good reason - [PR](https://github.com/openjdk/jdk/pull/23037)
[JDK-8346460](https://bugs.openjdk.org/browse/JDK-8346460) - NotifyFramePop should return JVMTI_ERROR_DUPLICATE - [PR](https://github.com/openjdk/jdk/pull/22798)
[JDK-8346143](https://bugs.openjdk.org/browse/JDK-8346460) - add ClearAllFramePops function to speedup debugger single stepping in some cases - [PR](https://github.com/openjdk/jdk/pull/22744)

PR #10529, opened 2025-01-21
Fix substratevm after 'JDK-8347287: JFR: Remove use of Security Manager'

simonis:
[JDK-8347287: JFR: Remove use of Security Manage](https://bugs.openjdk.org/browse/JDK-8347287) which has removed the SecurityManager from the JFR files has screwed up the substratevm and the native image build.

This pull requests fixes the problem in such a way that substratevm/native image work again with both versions of the JDK, before and after [JDK-8347287](https://bugs.openjdk.org/browse/JDK-8347287).

`mx --primary-suite=substratevm --components=ni,nju,lg gate -t 'native unittests'` which were the only JFR tests I could find succeed after this PR with both versions of the JDK before and after [JDK-8347287](https://bugs.openjdk.org/browse/JDK-8347287). Please let me know if there are other tests which I can run.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SecuritySupportSafePathPresent.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SecuritySupportSafePathPresent.java
@@ -24,31 +24,40 @@
  */
 package com.oracle.svm.core.jfr;
 
-import java.util.List;
-
 import com.oracle.svm.core.annotate.Alias;
-import com.oracle.svm.core.annotate.RecomputeFieldValue;
-import com.oracle.svm.core.annotate.Substitute;
 import com.oracle.svm.core.annotate.TargetClass;
 import com.oracle.svm.core.annotate.TargetElement;
-import com.oracle.svm.core.util.VMError;
 
-import jdk.jfr.internal.SecuritySupport;
+import java.util.function.BooleanSupplier;
 
-@TargetClass(value = jdk.jfr.internal.SecuritySupport.class, onlyWith = HasJfrSupport.class)
-public final class Target_jdk_jfr_internal_SecuritySupport {
-    // Checkstyle: stop
-    @Alias @TargetElement(onlyWith = SecuritySupportSafePathPresent.class) @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.Reset) //
-    static Target_jdk_jfr_internal_SecuritySupport_SafePath JFC_DIRECTORY;
-    @Alias @TargetElement(onlyWith = SecuritySupportSafePathPresent.class) @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.Reset) //
-    static Target_jdk_jfr_internal_SecuritySupport_SafePath JAVA_IO_TMPDIR;
-    // Checkstyle: resume
-
-    @Substitute @TargetElement(onlyWith = SecuritySupportSafePathPresent.class)
-    public static List<Target_jdk_jfr_internal_SecuritySupport_SafePath> getPredefinedJFCFiles() {
-        throw VMError.shouldNotReachHere("Paths from the image build must not be embedded into the Native Image.");
+/**
+ * A predicate that returns {@code true} if {@code boolean jdk.jfr.internal.SecuritySupport.SafePath} exists.
+ */
+final class SecuritySupportSafePathPresent implements BooleanSupplier {
+    @Override
+    public boolean getAsBoolean() {
+        try {
+            Class.forName("jdk.jfr.internal.SecuritySupport$SafePath");
+            return true;
+        } catch (ClassNotFoundException e) {
+            return false;
+        }
     }
+}
 
-    @Alias @TargetElement(onlyWith = SecuritySupportSafePathPresent.class)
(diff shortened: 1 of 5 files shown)


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/SecuritySupportSafePathPresent.java
```diff
@@ -0,0 +1,63 @@
...
+}
+
+@TargetClass(className = "jdk.jfr.internal.SecuritySupport$SafePath", onlyWith = SecuritySupportSafePathPresent.class)
+final class Target_jdk_jfr_internal_SecuritySupport_SafePath {
+    public Target_jdk_jfr_internal_SecuritySupport_SafePath(String p) {
+        originalConstructor(p);
+    }
+    @Alias
+    @TargetElement(name = TargetElement.CONSTRUCTOR_NAME)
+    native void originalConstructor(String p);
```
roberttoyonaga, line 62: Why do we need `originalConstructor` instead of just
```
@Alias
Target_jdk_jfr_internal_SecuritySupport_SafePath(String p) {
}
```

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/Target_jdk_jfr_internal_Repository.java
```diff
@@ -24,22 +24,46 @@
...
+
 import com.oracle.svm.core.SubstrateUtil;
 import com.oracle.svm.core.annotate.Alias;
+import com.oracle.svm.core.annotate.RecomputeFieldValue;
 import com.oracle.svm.core.annotate.Substitute;
 import com.oracle.svm.core.annotate.TargetClass;
+import com.oracle.svm.core.annotate.TargetElement;
 
-import jdk.jfr.internal.SecuritySupport.SafePath;
+import jdk.jfr.internal.SecuritySupport;
```
roberttoyonaga, line 37: I think there's a few places like here where `import jdk.jfr.internal.SecuritySupport;` is no longer needed.

roberttoyonaga (approved):
Thank you. This PR looks good to me. I left a few minor notes.  Yes, `mx native-unittest` will run all the JFR tests in `com.oracle.svm.test.jfr`. Those are the only ones.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/Target_jdk_jfr_internal_SecuritySupport.java
```diff
@@ -30,24 +30,25 @@
...
-    @Substitute
-    public static List<SafePath> getPredefinedJFCFiles() {
+    @Substitute @TargetElement(onlyWith = SecuritySupportSafePathPresent.class)
+    public static List<Target_jdk_jfr_internal_SecuritySupport_SafePath> getPredefinedJFCFiles() {
         throw VMError.shouldNotReachHere("Paths from the image build must not be embedded into the Native Image.");
     }
 
-    @Alias
-    static native SafePath getPathInProperty(String prop, String subPath);
+    @Alias @TargetElement(onlyWith = SecuritySupportSafePathPresent.class)
```
roberttoyonaga, line 52: I think all these `@TargetElement(onlyWith = SecuritySupportSafePathPresent.class)` can be applied to the  `@TargetClass` instead to make things a bit simpler

roberttoyonaga (changes requested):
Actually, I think that `SubstrateJVM.getDumpPath()` now needs to accommodate the case where `SecuritySupportSafePathPresent` is false. In that case `Target_jdk_jfr_internal_SecuritySupport.getPathInProperty` won't be available. In the [latest version](https://github.com/openjdk/jdk/blob/jdk-25%2B6/src/hotspot/share/jfr/recorder/repository/jfrEmergencyDump.cpp#L168) of Hotspot JFR code, it looks like the default path is the current working directory.

christianhaeubl:
Thanks for the PR but we already have an internal PR with similar changes. We do update Native Image to the latest OpenJDK tag approximately once per week. So, I am closing this PR.

zapster:
Thanks for you contribution @simonis!

I have not yet looked at your PR, but some general comments:

> This pull requests fixes the problem in such a way that substratevm/native image work again with both versions of the JDK, 

The Graal (native image and compiler) master branch is always only compatible to exactly one JDK version, namely the version that is specified in [common.json](https://github.com/oracle/graal/blob/master/common.json#L48-L54). Supporting more than one version in one codebase is not maintainable (nor testable) and also defies our goal of aligning development more closely to https://github.com/openjdk/jdk. (Currently, there is still JDK 21 support, but that is hopefully going away soon, so lets ignore this.) We adopt every new promoted JDK build, so the compatible JDK version of Graal is bumped once a week.

So a new JDK change that requires Graal adoption will be dealt with in the PR that updates the JDK. JDK-8347287 went in with JDK 25+6 and there is already a PR (#10513) that adopts this change, at least to the extent that all gates pass.

So bottom line: for JDK adoptions always look out for our adoption PRs. Of course, our adoptions might not cover every single detail. In that case, we are grateful for follow-up PRs from the community after the JDK update has been merged.

PR #10534, opened 2025-01-22
[GR-61076] Only set TrackNodeSourcePosition if OptimizationLevel is O0.

graalvmbot:


zakkak:
Hello @boris-spas, could you please elaborate on the reasons behind this change? Wouldn't this change come at the cost of losing accuracy when not using `-O0`?

boris-spas:
> could you please elaborate on the reasons behind this change?

This is essentially a bugfix, this PR is the intended behaviour - when a user set -O0, we override a few options (e.g. to make debugging easier). Before this PR we overrode these options for every -O<value>, which is not the intended behaviour as these flags should have their default values (or what ever the user also set).

> Wouldn't this change come at the cost of losing accuracy when not using -O0?

Not sure what you mean TBH, losing accuracy of what?

zakkak:
Thanks for the clarifications.

> Not sure what you mean TBH, losing accuracy of what?

I misinterpreted the patch and thought that the options were always set to `true` before this PR. As a result, I thought that this patch would result in less accurate debug info when not using `-O0`.

PR #10549, opened 2025-01-23
[GR-61448] Include Compilation ID in truffle compilation logs.

graalvmbot:


truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/libgraal/TruffleToLibGraalCalls2.java
```diff
@@ -0,0 +1,51 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.oracle.truffle.runtime.hotspot.libgraal;
+
+import static com.oracle.truffle.compiler.hotspot.libgraal.TruffleToLibGraal.Id.GetCompilationId;
+
+import com.oracle.truffle.compiler.hotspot.libgraal.TruffleToLibGraal;
+
+final class TruffleToLibGraalCalls2 {
```
JohnTortugo, line 47: @jchalou - Out of curiosity: why do we need a separate file for this method?

jchalou: This is because Truffle has to be compatible also with libgraal that does not have the `getCompilationId` method entry point. `LibGraalTruffleCompilationSupport` registers the native methods is a way that either registers all the native methods in a class or none at all, and so the ones that are not required to be there have to be in their separate classes (for now, we do plan to introduce some more sophisticated methodology in the future).

PR #10551, opened 2025-01-23
[GR-61527] Backport to 24.2: Add MaximumCompilations option.

graalvmbot:


compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/truffle/test/GR42688Test.java
```diff
@@ -0,0 +1,247 @@
...
+import com.oracle.truffle.api.nodes.Node;
+import com.oracle.truffle.api.nodes.RootNode;
+import com.oracle.truffle.compiler.TruffleCompilerListener;
+import com.oracle.truffle.runtime.AbstractCompilationTask;
+import com.oracle.truffle.runtime.OptimizedCallTarget;
+import com.oracle.truffle.runtime.OptimizedTruffleRuntime;
+import com.oracle.truffle.runtime.OptimizedTruffleRuntimeListener;
+
+/*
+ * Test that reproduces the not-yet fixed bug GR-42688 and makes sure the issue is alleviated by MaximumCompilations limit.
```
JohnTortugo, line 56: @jchalou , @chumer - Where can I see more information about bug `GR-42688` ? Thank you. /cc @simonis

simonis: @jchalou , thanks for the additional information. I have a few follow-up questions if you don't mind:

1. It looks like [GR-42688 was fixed in TruffleRuby](https://github.com/oracle/truffleruby/commit/0435e2cc)? Does that mean that the issue you've explained can be fixed in a Truffle language implementation? I.e. could it be fixed in GraalJS in the same way?
2. Is there a chance to log these specific deopts/recompilations or at least detect them in a compilation log?
3. Does this issue also happen, if we restrict the compiler to Tier 1 compilations?

Thank you and best regards,
Volker

jchalou: 1. Language implementations definitely can influence how prone they are to the bug. One pattern that makes the bug possible is a node specialization with a guard that uses an unshared cached node which also has multiple specializations. The reason is that such patterns need more than one execution to "stabilize", because last cached entry is executed first and its cached node might not be specialized to the values the previous cache entries have already seen and so the compiled code might deopt for already seen inputs (and then never get a chance to respecialize due to the bug). The GR42688Test is based on exactly that. You can see it was inspired by the TruffleRuby nodes. After TruffleRuby made the cached MetaClassNode shared, the bug is no longer reproducible using that particular specialization.
2. It is very complicated to detect that the bug actually occured. You can quite easily find out that a certain calltarget is compiled repeatedly with the same compiled code size. However, whether that is because some inlined calltarget is not being respecialized, because it has its own compiled code and is never executed in the interpreter, that can't be determined from the logs. You can trace inlining and find out that compiled code for some inlined calltarget does exist while the caller is being repeatedly recompiled, but whether the recompilations are due to the bug or some other bug, that is not possible to find out. You would have to study the compilation graphs. We are working on a deoptimization cycle detection tool which will save you the need to look for suspicious repeated compilations and it will also help with determining where in the AST the deoptimizations come from, but still, more detailed inspection of the compilation graphs might be needed to determine the exact cause of the repeated deopt.
3. Restricting the compiler to Tier 1 compilations successfully prevents the bug because Tier 1 does only trivial inlining.

JohnTortugo: Thank you @jchalou for detailed explanation. I imagine that this issue can arise also in the form of `transferToInterpreter` calls. Is that possible? I'm investigating an issue where a method was recompiled thousands of times producing the same code size and the reason of the recompilations seems to be that an specialization didn't happen. I shared some logs [here](https://gist.github.com/JohnTortugo/5bdb7501cfb479a702696baad2b4c17f) if you're interested in taking a look.

jchalou: @JohnTortugo The cause of the repeated compilations is different than the imprecise deopt in your case, because those compilations are Tier 1 compilations without any inlining. It may be caused by a particular repeated `transferToInterpreter` call though. If you run with the option `-Dpolyglot.engine.TraceTransferToInterpreter=true`, you may be able to identify which one. You can also analyze the compilation graphs.  If you send me a reproducer, I can investigate it for you. If you are familiar with IGV (ideal graph visualizer), I can also help you to investigate. The way to do it is t run with `-XX:+TraceDeoptimization` and then based on the compilation id of some repeated compilation, find the corresponding deopt log entry and note the debug_id of the entry. You can then find the corresponding node in the compilation graphs in the IGV dump of the compilation. To take dumps of the `set` method compilations, run with `-Dgraal.Dump=Truffle:3 -Dgraal.MethodFilter=set -Dgraal.DumpPath=<directory to dump to>`. You can also send me the dumps and I can analyze them for you. Which Truffle language are the compilations from? JavaScript? Please let me know. Thanks.

Issue #10742, opened 2025-02-24
[Native Image] JFR: Exception occurred during execution of flush task


dekelpilli:
### Describe the Issue

I was trying to use JFR to help debug an issue I was having, but the recording would remain empty, with the following log line appearing over and over again while the app was running:
```
[warn][jfr,system ] Exception occurred during execution of flush task
```

I've set up this repo to reproduce this issue - https://github.com/dekelpilli/graalvm-jfr-bug. It's worth noting that in its current state, the error does not always appear, but re-compiling will often make it appear again. 

### Using the latest version of GraalVM can resolve many issues.

- [x] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

### GraalVM Version

java version "23.0.1" 2024-10-15
Java(TM) SE Runtime Environment Oracle GraalVM 23.0.1+11.1 (build 23.0.1+11-jvmci-b01)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 23.0.1+11.1 (build 23.0.1+11-jvmci-b01, mixed mode, sharing)

Also, as this is using clojure:
```
$ clojure --version
Clojure CLI version 1.12.0.1517
```

### Operating System and Version

Darwin Mac 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6031 arm64

### Troubleshooting Confirmation

- [x] I tried the suggestions in the [troubleshooting guide](https://www.graalvm.org/latest/reference-manual/native-image/guides/troubleshoot-run-time-errors/).

### Run Command
```
./target/graal-memory-test -XX:StartFlightRecording=filename=recording.jfr
```
### Expected Behavior

`recording.jfr` is populated with JFR data during the application's runtime.

### Actual Behavior

The file is created but remains empty, and the application logs the following a few times per second:
```
[warn][jfr,system ] Exception occurred during execution of flush task
```

### Steps to Reproduce

From the [graalvm-jfr-bug repo](https://github.com/dekelpilli/graalvm-jfr-bug), `./compile.sh && ./run.sh`. This requires Clojure to be installed.

If this correctly creates JFR data, simply run this again until it doesn't. In my experience, it ends up working about 50% of the time.

I wasn't able to attach them to this issue, but I've committed bundles of both a working and a non-working compilation:
[...]

PR #10749, opened 2025-02-24
[GR-63550] Fix bug in JFR type repository

roberttoyonaga:
This is a fix for the bug reported by: https://github.com/oracle/graal/issues/10742

The problem was that, for each class used during the epoch, we only visited its module's classloader (if the class' package and it's module were both non null).  However, later on when serializing classes, we attempt to write the class' classloader.  This classloader is never explicitly visited so may not have been processed.


zakkak:
Hi @roberttoyonaga, can you please make the commit message more descriptive? Thanks

roberttoyonaga:
I updated the commit message. Note that one GHA is failing with: `OSError: [Errno 28] No space left on device`. Seems unrelated to this change.

christianhaeubl:
Thanks, I will integrate this later this week.

zapster:
> I updated the commit message. Note that one GHA is failing with: OSError: [Errno 28] No space left on device. Seems unrelated to this change.

Yes, it is unrelated. It has been fixed in the mean time on master.

Issue #3472, opened 2021-06-14
Support for Shenandoah GC


rkennke:
The Graal compiler currently does not support the Hotspot Shenandoah GC. Shenandoah GC is relevant in situations where latency is a concern, because it drastically reduces GC pause times. This is not only relevant in large-heap settings where GC tends to take a long time due to the size of live data that needs to be marked and relocated, but also in resource-constrained environment where even relatively little STW-work can balloon.

A first step in supporting Shenandoah GC in Graal compiler would be to implement the basic Shenandoah barriers:
- Load-reference barrier to be employed after loading references
- SATB-barrier to be employed before reference writes (very similar to G1's SATB barrier)
- CAS-obj barrier that handle Unsafe.compareAndSwapReference() and related methods
- Arraycopy barrier to deal with arraycopies of Object[]
- Reference.get() barrier

This would be enough to support Graal compiler in JDK 11. However, additional features have been added between JDK 11 and JDK 17, which require extra work:
- Concurrent class-unloading requires so-called nmethod-barriers: little pieces of GC code to be called when calling into 'armed' nmethods (and a mechanism to arm nmethods).
- Concurrent thread-scanning which requires stack-watermark support.

Both these features are not Shenandoah specific. For example, ZGC would also make use of them.

Much of the first step (JDK 11 support) is implemented in #2426.

mcraj017:
@dougxc please advise

rkennke:
Notice that this is mostly informational for now. We recently had a discussion with @thomaswue were we agreed that I file this issue to collect TODOs and a rough outline, also with an eye on possibly supporting Shenandoah in native-image.

dougxc:
I'm closing this issue due to lack of activity. Please re-open it (and reassign) when/if there are plans to make progress on adding Shenandoah to Native Image.

bric3:
@dougxc I don't think this issue is related to Shenandoah on the native image but on the Graal Compiler.

reneleonhardt:
It's sad that there is no progress after 3 years, switching from G1 to Shenandoah reduced our spring-web latency massively.

JohnTortugo:
Please, re-open this issue. I'm resuming @rkennke work on this patch.

rkennke:
I cannot seem to assign this back to me. I'm currently working on it, WIP PR is here: https://github.com/oracle/graal/pull/10904

zakkak:
[...]

PR #10904, opened 2025-03-21
Shenandoah support

rkennke:
This implements the barriers that are needed to run with Shenandoah GC in the Graal compiler. (Issue: https://github.com/oracle/graal/issues/3472)

There are 3 basic kinds of barriers needed for Shenandoah:
- SATB barriers (aka pre-write-barrier, also needed for Reference.get() support). Those are pretty similar to G1's SATB barriers. Those barriers are inserted before reference stores, or in the case of Reference.get(), after the load of the referent. The SATB barriers are inserted in the node graph, and expanded to assembly in the respective backends. (Compared to the G1 backend, we implemented a slight improvement, where we move the mid-path-section into an out-of-line stub, similar to the slow-path. This should improve performance by helping static branch prediction. We may want to change G1 barriers in a similar fashion.)
- Load-reference-barriers (LRB). Those are conceptually similar to ZGC's read-barriers, but differ in the implementation. Those barriers, too, are inserted as nodes, and expanded to assembly in the backends.
- Card-marking barriers. Those are only needed when running with generational Shenandoah, and are similar to Serial and Parallel GC's card-marking barriers. However, in contrast to Serial and Parallel GC, those Shenandoah card-barriers are again inserted as nodes, and expanded to assembly in the backends. (We may want to adapt this code in Serial and Parallel and ditch their snippets-based implementation.)

Notice that none of the barriers are implemented as snippets (like Serial/Parallel's card-barriers) or in the backend-only (as ZGC's read-barriers). We needed a way to efficiently deal with compressed-oops, which is not (easily) possible to do in the backend. In the node-graph this is pretty easy: insert the LRB with preceding and succeeding uncompress/compress after any load and before the (potential) uncompress (i.e. turn load->uncompress into load -> (uncompress -> lrb -> compress) -> uncompress) and then let the optimizer optimize away the trailing compress -> uncompress pairs.

In order to support this, we needed a few additions:
- The compression nodes now have a method that allows to add them without using unique(). If we used unique(), then the uncompress before the LRB would be matched with the original uncompress after the load, and we would cut out the LRB.
- We moved the barrier insertion for Shenandoah from the mid-tier to the low-tier. This is needed because we can't insert barriers to FloatingReadNodes. We moved the barrier insertion to after fixing the read-nodes, at which point this is safe to do. Other GCs keep adding their barriers in the mid-tier. The mechanics is that BarrierSet defaults to mid-tier, but implementations can override this to add barriers in low-tier (instead, or additionally).

X86 port contributed by @JohnTortugo.

Testing:
- Renaissance
- SPECjvm2008
- SPECjbb2015
- DaCapo

(We have run those workloads for correctness testing only, we have not (yet) conducted a performance study.)

### DIFF
--- a/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/core/phases/LowTier.java
+++ b/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/core/phases/LowTier.java
@@ -28,6 +28,7 @@ import static jdk.graal.compiler.phases.common.DeadCodeEliminationPhase.Optional
 
 import jdk.graal.compiler.core.common.GraalOptions;
 import jdk.graal.compiler.graph.Graph;
+import jdk.graal.compiler.nodes.GraphState;
 import jdk.graal.compiler.options.Option;
 import jdk.graal.compiler.options.OptionKey;
 import jdk.graal.compiler.options.OptionType;
@@ -47,6 +48,7 @@ import jdk.graal.compiler.phases.common.ProfileCompiledMethodsPhase;
 import jdk.graal.compiler.phases.common.PropagateDeoptimizeProbabilityPhase;
 import jdk.graal.compiler.phases.common.RemoveOpaqueValuePhase;
 import jdk.graal.compiler.phases.common.TransplantGraphsPhase;
+import jdk.graal.compiler.phases.common.WriteBarrierAdditionPhase;
 import jdk.graal.compiler.phases.schedule.SchedulePhase;
 import jdk.graal.compiler.phases.schedule.SchedulePhase.SchedulingStrategy;
 import jdk.graal.compiler.phases.tiers.LowTierContext;
@@ -86,6 +88,8 @@ public class LowTier extends BaseTier<LowTierContext> {
 
         appendPhase(new FixReadsPhase(true,
                         new SchedulePhase(GraalOptions.StressTestEarlyReads.getValue(options) ? SchedulingStrategy.EARLIEST : SchedulingStrategy.LATEST_OUT_OF_LOOPS_IMPLICIT_NULL_CHECKS)));
+
+        appendPhase(new WriteBarrierAdditionPhase(GraphState.StageFlag.LOW_TIER_BARRIER_ADDITION));
 
         appendPhase(canonicalizerWithoutGVN);
 
--- a/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/GraalHotSpotVMConfig.java
+++ b/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/GraalHotSpotVMConfig.java
@@ -609,6 +609,44 @@ public class GraalHotSpotVMConfig extends GraalHotSpotVMConfigAccess {
     public final long zBarrierSetRuntimeLoadBarrierOnOopArray = getZGCAddressField("ZBarrierSetRuntime::load_barrier_on_oop_array");
     public final int zPointerLoadShift = getConstant("ZPointerLoadShift", Integer.class, -1, osArch.equals("aarch64"));
 
+    /*
+     * Shenandoah GC support.
+     */
+    /**
+     * Indicates whether or not the HotSpot VM has been built with Shenandoah support. If not, then
+     * we don't expect the Shenandoah symbols to be present in JVMCI.
+     */
+    public final boolean hasShenandoahGC = getStore().getConstants().containsKey("INCLUDE_SHENANDOAHGC") && getConstant("INCLUDE_SHENANDOAHGC", Boolean.class);
+
+    /*
+     * Various Shenandoah GC constants.
+     */
+    public final int shenandoahGCStateOffset = getConstant("ShenandoahThreadLocalData::gc_state_offset", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahSATBIndexOffset = getConstant("ShenandoahThreadLocalData::satb_mark_queue_index_offset", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahSATBBufferOffset = getConstant("ShenandoahThreadLocalData::satb_mark_queue_buffer_offset", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahCardTableOffset = getConstant("ShenandoahThreadLocalData::card_table_offset", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahGCRegionSizeBytesShift = getFieldValue("CompilerToVM::Data::shenandoah_region_size_bytes_shift", Integer.class, "int", -1, hasShenandoahGC);
+    public final long shenandoahGCCSetFastTestAddress = getFieldValue("CompilerToVM::Data::shenandoah_in_cset_fast_test_addr", Long.class, "address", -1L, hasShenandoahGC);
+
+    public final int shenandoahGCStateHasForwarded = getConstant("ShenandoahHeap::HAS_FORWARDED", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahGCStateMarking = getConstant("ShenandoahHeap::MARKING", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahGCStateEvacuation = getConstant("ShenandoahHeap::EVACUATION", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahGCStateUpdateRefs = getConstant("ShenandoahHeap::UPDATE_REFS", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahGCStateWeakRoots = getConstant("ShenandoahHeap::WEAK_ROOTS", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahGCStateYoungMarking = getConstant("ShenandoahHeap::YOUNG_MARKING", Integer.class, -1, hasShenandoahGC);
+    public final int shenandoahGCStateOldMarking = getConstant("ShenandoahHeap::OLD_MARKING", Integer.class, -1, hasShenandoahGC);
+
+    /*
+     * Shenandoah barrier slow-paths.
+     */
+    public final long shenandoahLoadBarrierStrong = getAddress("ShenandoahRuntime::load_reference_barrier_strong", -1L, hasShenandoahGC);
+    public final long shenandoahLoadBarrierStrongNarrow = getAddress("ShenandoahRuntime::load_reference_barrier_strong_narrow", -1L, hasShenandoahGC);
+    public final long shenandoahLoadBarrierWeak = getAddress("ShenandoahRuntime::load_reference_barrier_weak", -1L, hasShenandoahGC);
+    public final long shenandoahLoadBarrierWeakNarrow = getAddress("ShenandoahRuntime::load_reference_barrier_weak_narrow", -1L, hasShenandoahGC);
+    public final long shenandoahLoadBarrierPhantom = getAddress("ShenandoahRuntime::load_reference_barrier_phantom", -1L, hasShenandoahGC);
+    public final long shenandoahLoadBarrierPhantomNarrow = getAddress("ShenandoahRuntime::load_reference_barrier_phantom_narrow", -1L, hasShenandoahGC);
+    public final long shenandoahWriteBarrierPre = getAddress("ShenandoahRuntime::write_barrier_pre", -1L, hasShenandoahGC);
+
     // aarch64 specific nmethod entry barrier support
     // @formatter:off
     public final int BarrierSetAssembler_nmethod_patching_type = getFieldValue("CompilerToVM::Data::BarrierSetAssembler_nmethod_patching_type", Integer.class, "int", -1, osArch.equals("aarch64"));
--- a/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/HotSpotShenandoahBarrierSet.java
+++ b/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/HotSpotShenandoahBarrierSet.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package jdk.graal.compiler.hotspot;
+
+import jdk.graal.compiler.core.common.CompressEncoding;
+import jdk.graal.compiler.debug.GraalError;
+import jdk.graal.compiler.hotspot.nodes.HotSpotCompressionNode;
+import org.graalvm.word.LocationIdentity;
+
+import jdk.graal.compiler.core.common.memory.BarrierType;
+import jdk.graal.compiler.core.common.type.AbstractObjectStamp;
+import jdk.graal.compiler.core.common.type.Stamp;
+import jdk.graal.compiler.hotspot.replacements.HotSpotReplacementsUtil;
+import jdk.graal.compiler.nodes.ValueNode;
+import jdk.graal.compiler.nodes.gc.shenandoah.ShenandoahBarrierSet;
+import jdk.vm.ci.meta.JavaKind;
+import jdk.vm.ci.meta.ResolvedJavaField;
+import jdk.vm.ci.meta.ResolvedJavaType;
+
+/**
+ * Specialization of {@link ShenandoahBarrierSet} that adds support for read barriers on handle
+ * locations and compressed references.
+ */
+public class HotSpotShenandoahBarrierSet extends ShenandoahBarrierSet {
+    private final CompressEncoding oopEncoding;
+
+    public HotSpotShenandoahBarrierSet(ResolvedJavaType objectArrayType, ResolvedJavaField referentField, GraalHotSpotVMConfig config) {
+        super(objectArrayType, referentField);
+        this.oopEncoding = config.getOopEncoding();
+        this.useLoadRefBarrier = config.getFlag("ShenandoahLoadRefBarrier", Boolean.class);
+        this.useSATBBarrier = config.getFlag("ShenandoahSATBBarrier", Boolean.class);
+        this.useCASBarrier = config.getFlag("ShenandoahCASBarrier", Boolean.class);
+        this.useCardBarrier = config.getFlag("ShenandoahCardBarrier", Boolean.class);
+    }
+
+    @Override
+    protected BarrierType barrierForLocation(BarrierType currentBarrier, LocationIdentity location, JavaKind storageKind) {
+        if (location instanceof HotSpotReplacementsUtil.OopHandleLocationIdentity) {
+            return BarrierType.READ;
+        }
+        return super.barrierForLocation(currentBarrier, location, storageKind);
+    }
+
+    @Override
+    public BarrierType readBarrierType(LocationIdentity location, ValueNode address, Stamp loadStamp) {
+        if (location instanceof HotSpotReplacementsUtil.OopHandleLocationIdentity) {
+            GraalError.guarantee(loadStamp instanceof AbstractObjectStamp, "expect object, got: %s", loadStamp);
+            return BarrierType.READ;
+        }
+        return super.readBarrierType(location, address, loadStamp);
+    }
+
+    @Override
+    public BarrierType writeBarrierType(LocationIdentity location) {
+        if (location instanceof HotSpotReplacementsUtil.OopHandleLocationIdentity) {
+            return BarrierType.FIELD;
+        }
+        return BarrierType.NONE;
+    }
+
+    @Override
+    protected ValueNode maybeUncompressReference(ValueNode value, boolean narrow) {
+        if (value != null && narrow) {
+            return HotSpotCompressionNode.uncompressWithoutUnique(value.graph(), value, oopEncoding);
+        }
+        return value;
+    }
+
+    @Override
+    protected ValueNode maybeCompressReference(ValueNode value, boolean narrow) {
+        if (value != null && narrow) {
+            return HotSpotCompressionNode.compressWithoutUnique(value.graph(), value, oopEncoding);
+        }
+        return value;
+    }
+}
--- a/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahBarrierSetLIRGenerator.java
+++ b/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahBarrierSetLIRGenerator.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package jdk.graal.compiler.hotspot.aarch64.shenandoah;
+
+import jdk.graal.compiler.asm.aarch64.AArch64Address;
+import jdk.graal.compiler.core.aarch64.AArch64LIRGenerator;
+import jdk.graal.compiler.core.aarch64.AArch64ReadBarrierSetLIRGenerator;
+import jdk.graal.compiler.core.common.LIRKind;
+import jdk.graal.compiler.core.common.memory.BarrierType;
+import jdk.graal.compiler.core.common.memory.MemoryExtendKind;
+import jdk.graal.compiler.core.common.memory.MemoryOrderMode;
+import jdk.graal.compiler.core.common.spi.ForeignCallLinkage;
+import jdk.graal.compiler.hotspot.GraalHotSpotVMConfig;
+import jdk.graal.compiler.hotspot.meta.HotSpotHostForeignCallsProvider;
+import jdk.graal.compiler.hotspot.meta.HotSpotProviders;
+import jdk.graal.compiler.lir.LIRFrameState;
+import jdk.graal.compiler.lir.Variable;
+import jdk.graal.compiler.lir.aarch64.AArch64AddressValue;
+import jdk.graal.compiler.lir.gen.LIRGeneratorTool;
+import jdk.graal.compiler.lir.gen.ShenandoahBarrierSetLIRGeneratorTool;
+import jdk.graal.compiler.nodes.gc.shenandoah.ShenandoahLoadRefBarrierNode;
+import jdk.vm.ci.aarch64.AArch64Kind;
+import jdk.vm.ci.meta.AllocatableValue;
+import jdk.vm.ci.meta.PlatformKind;
+import jdk.vm.ci.meta.Value;
+
+/**
+ * Lowers Shenandoah barriers to AArch64 LIR.
+ */
+public class AArch64HotSpotShenandoahBarrierSetLIRGenerator implements ShenandoahBarrierSetLIRGeneratorTool, AArch64ReadBarrierSetLIRGenerator {
+    public AArch64HotSpotShenandoahBarrierSetLIRGenerator(GraalHotSpotVMConfig config, HotSpotProviders providers) {
+        this.config = config;
+        this.providers = providers;
+    }
+
+    private final GraalHotSpotVMConfig config;
+    private final HotSpotProviders providers;
+
+    private static ForeignCallLinkage getReadBarrierStub(LIRGeneratorTool tool, ShenandoahLoadRefBarrierNode.ReferenceStrength strength, boolean narrow) {
+        return switch (strength) {
+            case STRONG ->
+                narrow ? tool.getForeignCalls().lookupForeignCall(HotSpotHostForeignCallsProvider.SHENANDOAH_LOAD_BARRIER_NARROW)
+                                : tool.getForeignCalls().lookupForeignCall(HotSpotHostForeignCallsProvider.SHENANDOAH_LOAD_BARRIER);
+            case WEAK ->
+                narrow ? tool.getForeignCalls().lookupForeignCall(HotSpotHostForeignCallsProvider.SHENANDOAH_LOAD_BARRIER_WEAK_NARROW)
+                                : tool.getForeignCalls().lookupForeignCall(HotSpotHostForeignCallsProvider.SHENANDOAH_LOAD_BARRIER_WEAK);
+            case PHANTOM ->
+                narrow ? tool.getForeignCalls().lookupForeignCall(HotSpotHostForeignCallsProvider.SHENANDOAH_LOAD_BARRIER_PHANTOM_NARROW)
+                                : tool.getForeignCalls().lookupForeignCall(HotSpotHostForeignCallsProvider.SHENANDOAH_LOAD_BARRIER_PHANTOM);
+        };
+    }
+
+    @Override
+    public Value emitLoadReferenceBarrier(LIRGeneratorTool tool, Value obj, Value address, ShenandoahLoadRefBarrierNode.ReferenceStrength strength, boolean narrow, boolean notNull) {
+        PlatformKind platformKind = obj.getPlatformKind();
+        LIRKind kind = LIRKind.reference(platformKind);
+        Value result = tool.newVariable(tool.toRegisterKind(kind));
+        ForeignCallLinkage callTarget = getReadBarrierStub(tool, strength, narrow);
+        AllocatableValue object = tool.asAllocatable(obj);
+        AArch64AddressValue loadAddress = ((AArch64LIRGenerator) tool).asAddressValue(address, AArch64Address.ANY_SIZE);
+        tool.getResult().getFrameMapBuilder().callsMethod(callTarget.getOutgoingCallingConvention());
+        tool.append(new AArch64HotSpotShenandoahLoadRefBarrierOp(config, providers, tool.asAllocatable(result), object, loadAddress, callTarget, strength, notNull));
+        return result;
+    }
+
+    @Override
+    public void emitPreWriteBarrier(LIRGeneratorTool lirTool, Value address, AllocatableValue expectedObject, boolean nonNull) {
+        AllocatableValue temp = lirTool.newVariable(LIRKind.value(AArch64Kind.QWORD));
+        // If the assembly must load the value then it needs a temporary to store it.
+        AllocatableValue temp2 = expectedObject.equals(Value.ILLEGAL) ? lirTool.newVariable(LIRKind.value(AArch64Kind.QWORD)) : Value.ILLEGAL;
+
+        // Load the address into a register
+        AllocatableValue addressValue = lirTool.newVariable(address.getValueKind());
+        lirTool.emitMove(addressValue, address);
+
+        ForeignCallLinkage callTarget = lirTool.getForeignCalls().lookupForeignCall(HotSpotHostForeignCallsProvider.SHENANDOAH_WRITE_BARRIER_PRE);
+        lirTool.getResult().getFrameMapBuilder().callsMethod(callTarget.getOutgoingCallingConvention());
+        lirTool.append(new AArch64HotSpotShenandoahSATBBarrierOp(config, providers, addressValue, expectedObject, temp, temp2, callTarget, nonNull));
+    }
+
+    @Override
+    public void emitCardBarrier(LIRGeneratorTool lirTool, Value address) {
+        AArch64AddressValue addr = ((AArch64LIRGenerator) lirTool).asAddressValue(address, AArch64Address.ANY_SIZE);
+        AllocatableValue tmp = lirTool.newVariable(LIRKind.value(AArch64Kind.QWORD));
+        lirTool.append(new AArch64HotSpotShenandoahCardBarrierOp(config, providers, addr, tmp));
+    }
+
+    @Override
+    public void emitCompareAndSwapOp(LIRGeneratorTool tool, boolean isLogic, Value address, MemoryOrderMode memoryOrder, AArch64Kind memKind, Variable result,
+                    AllocatableValue allocatableExpectedValue, AllocatableValue allocatableNewValue, BarrierType barrierType) {
+        AllocatableValue tmp1 = tool.newVariable(LIRKind.value(AArch64Kind.QWORD));
+        tool.append(new AArch64HotSpotShenandoahCompareAndSwapOp(config, providers, memKind, memoryOrder, isLogic, result, allocatableExpectedValue, allocatableNewValue, tool.asAllocatable(address),
+                        tmp1));
+    }
+
+    @Override
+    public Value emitAtomicReadAndWrite(LIRGeneratorTool tool, LIRKind readKind, Value address, Value newValue, BarrierType barrierType) {
+        // We insert the necessary barriers in the node graph, at that level it
+        // is easier to handle compressed object references. No need to do anything
+        // special here.
+        return tool.emitAtomicReadAndWrite(readKind, address, newValue, BarrierType.NONE);
+    }
+
+    @Override
+    public Variable emitBarrieredLoad(LIRGeneratorTool tool, LIRKind kind, Value address, LIRFrameState state, MemoryOrderMode memoryOrder, BarrierType barrierType) {
+        // We insert the necessary barriers in the node graph, at that level it
+        // is easier to handle compressed object references. No need to do anything
+        // special here.
+        return tool.getArithmetic().emitLoad(kind, address, state, memoryOrder, MemoryExtendKind.DEFAULT);
+    }
+}
--- a/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahCardBarrierOp.java
+++ b/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahCardBarrierOp.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package jdk.graal.compiler.hotspot.aarch64.shenandoah;
+
+import jdk.graal.compiler.asm.Label;
+import jdk.graal.compiler.asm.aarch64.AArch64Address;
+import jdk.graal.compiler.asm.aarch64.AArch64MacroAssembler;
+import jdk.graal.compiler.hotspot.GraalHotSpotVMConfig;
+import jdk.graal.compiler.hotspot.meta.HotSpotProviders;
+import jdk.graal.compiler.hotspot.replacements.HotSpotReplacementsUtil;
+import jdk.graal.compiler.lir.LIRInstructionClass;
+import jdk.graal.compiler.lir.SyncPort;
+import jdk.graal.compiler.lir.aarch64.AArch64AddressValue;
+import jdk.graal.compiler.lir.aarch64.AArch64LIRInstruction;
+import jdk.graal.compiler.lir.asm.CompilationResultBuilder;
+import jdk.vm.ci.code.Register;
+import jdk.vm.ci.meta.AllocatableValue;
+
+import static jdk.graal.compiler.lir.LIRInstruction.OperandFlag.COMPOSITE;
+import static jdk.graal.compiler.lir.LIRInstruction.OperandFlag.REG;
+import static jdk.vm.ci.aarch64.AArch64.zr;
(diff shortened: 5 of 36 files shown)


tkrodriguez:
If you have any questions about how to structure your changes feel free to ping me over slack as I was the primary author of the new LIR support for barriers.

rkennke:
> If you have any questions about how to structure your changes feel free to ping me over slack as I was the primary author of the new LIR support for barriers.

Thanks, Tom! I will do that whenever I get stuck or have questions. So far I'm making progress. Structurally, Shenandoah will be look like a mix of ZGC (for the load-barrier, even though I am modeling it as a Node that consumes the loaded value, instead of replacing the ReadNode altogether) and G1 (for the SATB parts), and likely Serial/Parallel for the card-table parts.

tkrodriguez:
Sounds good.  There's still some more work to finish out the switch to LIR only barriers but I think supporting G1 and ZGC covers the required strategies in a fairly pragmatic way.

dougxc:
@tkrodriguez can you please take another look at this. Once done, we can ask @davleopo and @gergo- to look at it.

tkrodriguez:
I'll take a look.

tkrodriguez:
Are the gate failures actual problems?  It would be good to see a clean gate.  Also, we should squash the history before committing.

rkennke:
> Are the gate failures actual problems? It would be good to see a clean gate.

I don't know what those problems are. I fixed everything that looked related to my changes. Those failures look like infra problems, some volumes seem to have run out of memory or something. I doubt that it is related.

> Also, we should squash the history before committing. Ok, I can do that - tomorrow.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/core/phases/LowTier.java
```diff
@@ -87,6 +88,8 @@ public LowTier(OptionValues options) {
         appendPhase(new FixReadsPhase(true,
                         new SchedulePhase(GraalOptions.StressTestEarlyReads.getValue(options) ? SchedulingStrategy.EARLIEST : SchedulingStrategy.LATEST_OUT_OF_LOOPS_IMPLICIT_NULL_CHECKS)));
 
+        appendPhase(new WriteBarrierAdditionPhase());
```
tkrodriguez, line 91: This is missing from EconomyLowTier.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/gc/shenandoah/ShenandoahBarrierSet.java
```diff
@@ -0,0 +1,342 @@
...
+
+    private void addLoadReferenceBarrier(FixedAccessNode node, AddressNode address, BarrierType barrierType) {
+        GraalError.guarantee(node != null, "input value must not be null");
+        StructuredGraph graph = node.graph();
+        boolean narrow = node.stamp(NodeView.DEFAULT) instanceof NarrowOopStamp;
+        ValueNode uncompressed = maybeUncompressReference(node, narrow);
+        ShenandoahLoadRefBarrierNode lrb = graph.add(new ShenandoahLoadRefBarrierNode(uncompressed, address, barrierType, narrow));
+        ValueNode compValue = maybeCompressReference(lrb, narrow);
+        ValueNode newUsage = uncompressed != node ? uncompressed : lrb;
+        node.replaceAtUsages(compValue, InputType.Value, usage -> usage != newUsage);
```
tkrodriguez, line 234: I think this can just be `node.replaceAtUsages(compValue, usage -> usage != newUsage);`.  compValue is a regular Java value so it can only have `Value` usages.  So you don't need the changes in Node.java either, which I wasn't a fan of.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/core/phases/LowTier.java
rkennke: Oops. Added it there.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/gc/shenandoah/ShenandoahBarrierSet.java
rkennke: I am not sure what lead me to think checking for the InputType was necessary, tbh. Maybe during development something else was wrong elsewhere and I would see non-Value input for some reason. Anyhow, I changed/reverted that part. Thanks for pointing out!

rkennke:
@tkrodriguez There seem to be GHA failures that report missing SerialWriteBarriers in SubstrateVM. Could this be related to moving the barrier addition phase from mid- to low-tier? Is there a dependency on that in SubstrateVM?

rkennke:
> Yes this is something I mentioned in our slack discussions. Moving `WriteBarrierAdditionPhase` after `LowTierLoweringPhase` creates problems for GCs that still use snippets since they expect to be lowered by LowTierLoweringPhase. In the long term I would like it to be done there but I'm not sure how to bridge the gap. I've got an internal PR based on your branch that I'm testing and I was going to look into this.
> 
> The options are to conditionalize the placement of `WriteBarrierAdditionPhase` based on methods from `BarrierSet` but that's not available when constructing the suites. We could have early and late `WriteBarrierAdditionPhase` to handle each case during the transition but that's a bit ugly to me. Or we could beef up `WriteBarrierAdditionPhase` to perform any required lowering itself, though that might be a bit complicated. It would also complicate stuff like `BarrierSetVerificationPhase` and some barrier elimination that's part of enterprise.
> 
> I'm going to try putting `appendPhase(new PlaceholderPhase<>(WriteBarrierAdditionPhase.class));` in both MidTier and LowTier and do the placeholder replacement based on the BarrierSet.

As far as I can see, it's only the SerialWriteBarrierNode which depends on snippets (is that right?). That should be relatively straightforward to implement without snippets and would look almost exactly like ShenandoahCardBarrierNode implementation - even a little simpler. I could work on implementing that, if you think that'd help.

tkrodriguez:
It would be easy to convert the HotSpot serial barrier to LIR but native image uses snippets for its barriers and its serial barrier is non-trivial.  So we'll need to live with this mixed model for a little while I think.  I'm beginning to think we might just need an early and late phase.  I think the way barriers for vector writes work, we might have to do barrier addition for them before LowTierLowering, or at least before VectorLoweringPhase, which is before FixReadsPhase.

It might be too much to try to resolve all these issues in this PR.  Since Shenandoah is currently HotSpot only maybe it would be best to special case its barrier insertion.  I'll will play some more with this to see what would be best.

rkennke:
> It would be easy to convert the HotSpot serial barrier to LIR but native image uses snippets for its barriers and its serial barrier is non-trivial. So we'll need to live with this mixed model for a little while I think. I'm beginning to think we might just need an early and late phase. I think the way barriers for vector writes work, we might have to do barrier addition for them before LowTierLowering, or at least before VectorLoweringPhase, which is before FixReadsPhase.
> 
> It might be too much to try to resolve all these issues in this PR. Since Shenandoah is currently HotSpot only maybe it would be best to special case its barrier insertion. I'll will play some more with this to see what would be best.

I implemented barrier addition to trigger in both mid- and low-tier, and the BarrierSet implementation gets to choose which one (or both, if it wishes) is appropriate. This choice defaults to mid-tier, and can be overridden in implementations, like I did in ShenandoahBarrierSet. This way we get a clean gate :-)

tkrodriguez:
Thanks.  I'll see whether that works ok in our full gate.  I'd tried something slightly different but the `StageFlags` makes it hard to be super flexible about when these phases run.  I might be tempted to keep only the phase that actually does the work in the final suite.

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/hotspot/test/DeferredBarrierAdditionTest.java
```diff
@@ -98,7 +99,7 @@ protected void testHelper(final String snippetName, final int expectedBarriers,
             new GuardLoweringPhase().apply(graph, midContext);
             new MidTierLoweringPhase(createCanonicalizerPhase()).apply(graph, midContext);
             graph.getGraphState().setAfterFSA();
-            new WriteBarrierAdditionPhase().apply(graph, midContext);
+            new WriteBarrierAdditionPhase(GraphState.StageFlag.MID_TIER_BARRIER_ADDITION).apply(graph, midContext);
```
tkrodriguez, line 102: Maybe it's better for the default constructor to just assume `MID_TIER_BARRIER_ADDITION`?  Part of me wants to drop the `MID_TIER_` from that name as well since eventually we'll be back to only having one `WriteBarrierAdditionPhase`.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/gc/shenandoah/ShenandoahPreWriteBarrierNode.java
```diff
@@ -0,0 +1,87 @@
...
+        super(TYPE, address, expectedObject, true);
+        assert doLoad == (expectedObject == null) : Assertions.errorMessageContext("adr", address, "expectedO", expectedObject, "doLoad", doLoad);
+        this.doLoad = doLoad;
+    }
+
+    public ValueNode getExpectedObject() {
+        return getValue();
+    }
+
+    public boolean doLoad() {
```
tkrodriguez, line 62: `doLoad` appears to be dead

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphState.java
```diff
@@ -62,22 +62,22 @@ public final class GraphState {
...
+                    StageFlag.LOW_TIER_BARRIER_ADDITION);
     private static final EnumSet<StageFlag> ENTERPRISE_MID_TIER_MANDATORY_STAGES = EnumSet.of(
                     StageFlag.OPTIMISTIC_ALIASING,
                     StageFlag.GUARD_LOWERING,
                     StageFlag.VALUE_PROXY_REMOVAL,
                     StageFlag.SAFEPOINTS_INSERTION,
                     StageFlag.MID_TIER_LOWERING,
                     StageFlag.FSA,
-                    StageFlag.NODE_VECTORIZATION,
-                    StageFlag.BARRIER_ADDITION);
```
tkrodriguez, line 80: I think this should have been changed to `StageFlag.LOW_TIER_BARRIER_ADDITION` instead of being deleted.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/phases/common/WriteBarrierAdditionPhase.java
```diff
@@ -37,10 +37,16 @@
...
+    public WriteBarrierAdditionPhase(StageFlag stage) {
+        this.stage = stage;
+    }
+
     @Override
     public Optional<NotApplicable> notApplicableTo(GraphState graphState) {
         return NotApplicable.ifAny(
-                        NotApplicable.ifApplied(this, StageFlag.BARRIER_ADDITION, graphState),
+                        NotApplicable.ifApplied(this, stage, graphState),
                         NotApplicable.unlessRunAfter(this, StageFlag.MID_TIER_LOWERING, graphState),
```
tkrodriguez, line 50: This should probably be `stage == StageFlag.MID_TIER_BARRIER_ADDITION ? StageFlag.MID_TIER_LOWERING : StageFlag.FIXED_READS`.

tkrodriguez:
Overall I think this looks good.  I've only lightly reviewed the actual shenandoah parts since I don't really know anything about how the collector works.  A high level JavaDoc comment on each of your newly added classes would be appreciated.

@davleopo @gergo- I think it's in good shape for review.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/phases/common/WriteBarrierAdditionPhase.java
rkennke: I've been thinking the same, but we don't always do FIXED_READS, right?

tkrodriguez: True.  Then maybe just `LOW_TIER_LOWERING` then.  It's hard to get these right sometimes.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/GraalHotSpotVMConfig.java
```diff
@@ -436,6 +436,14 @@ public int threadLastJavaFpOffset() {
     public final int g1CardQueueBufferOffset = getConstant("G1ThreadLocalData::dirty_card_queue_buffer_offset", Integer.class, -1, !g1LowLatencyPostWriteBarrierSupport);
     public final int g1CardTableBaseOffset = getConstant("G1ThreadLocalData::card_table_base_offset", Integer.class, -1, g1LowLatencyPostWriteBarrierSupport);
 
+    public final boolean hasShenandoahGC = getStore().getConstants().containsKey("INCLUDE_SHENANDOAHGC") && getConstant("INCLUDE_SHENANDOAHGC", Boolean.class);
```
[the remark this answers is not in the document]
rkennke: It does work because I added it in upstream OpenJDK recently. As soon as labsjdk picks it up, it should work with that, too.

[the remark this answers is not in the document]
rkennke: See also: https://github.com/openjdk/jdk/pull/25325

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahBarrierSetLIRGenerator.java
```diff
@@ -0,0 +1,100 @@
...
+        AArch64AddressValue loadAddress = ((AArch64LIRGenerator) tool).asAddressValue(address, AArch64Address.ANY_SIZE);
+        tool.getResult().getFrameMapBuilder().callsMethod(callTarget.getOutgoingCallingConvention());
+        tool.append(new AArch64HotSpotShenandoahLoadRefBarrierOp(config, providers, tool.asAllocatable(result), object, loadAddress, callTarget, strength, notNull));
+        return result;
+    }
+
+    @Override
+    public void emitPreWriteBarrier(LIRGeneratorTool lirTool, Value address, AllocatableValue expectedObject, boolean nonNull) {
+        AllocatableValue temp = lirTool.newVariable(LIRKind.value(AArch64Kind.QWORD));
+        // If the assembly must load the value then it's needs a temporary to store it
```
davleopo, line 82: grammar error in comment

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahLoadRefBarrierOp.java
```diff
@@ -0,0 +1,233 @@
...
+    private final GraalHotSpotVMConfig config;
+
+    @Def({REG}) protected AllocatableValue result;
+    @Use({REG}) protected AllocatableValue object;
+    @Alive({COMPOSITE}) protected AArch64AddressValue loadAddress;
+
+    protected final ForeignCallLinkage callTarget;
+
+    ShenandoahLoadRefBarrierNode.ReferenceStrength strength;
+    boolean notNull;
```
davleopo, line 120: please add javadoc to non private members added in this PR and if possible also to private ones if they are not trivial

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahLoadRefBarrierOp.java
```diff
@@ -0,0 +1,233 @@
...
+        this.result = result;
+        this.object = object;
+        this.loadAddress = loadAddress;
+        this.callTarget = callTarget;
+        this.strength = strength;
+        this.notNull = notNull;
+    }
+
+    @Override
+    public void emitCode(CompilationResultBuilder crb, AArch64MacroAssembler masm) {
```
davleopo, line 139: how likely are these implementations to change? should we tie them to some jdk code with `SyncPort` cc @mur47x111

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/gc/shenandoah/ShenandoahLoadRefBarrierNode.java
```diff
@@ -0,0 +1,98 @@
...
+ * Shenandoah load-reference barriers. Those are added after reference-loads, and are used to
+ * canonicalize references during concurrent evacuation. During concurrent evacuation we might see
+ * both from- and to-space references to the same objects, and this barrier ensures that we only see
+ * to-space references. (a.k.a. To-space invariant).
+ */
+@NodeInfo(cycles = CYCLES_64, size = SIZE_64)
+public final class ShenandoahLoadRefBarrierNode extends ValueNode implements LIRLowerable {
+    public static final NodeClass<ShenandoahLoadRefBarrierNode> TYPE = NodeClass.create(ShenandoahLoadRefBarrierNode.class);
+
+    public enum ReferenceStrength {
```
davleopo, line 55: javadoc please

davleopo:
Thanks very nice work, looks pretty ready I did not see any blockers from my side.  I added some style/unification comments. One question is if we should already add necessary `SyncPort` annotations to the jdk code in question to capture changes asap.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/HotSpotShenandoahBarrierSet.java
```diff
@@ -0,0 +1,97 @@
...
+        if (location instanceof HotSpotReplacementsUtil.OopHandleLocationIdentity) {
+            return BarrierType.READ;
+        }
+        return super.barrierForLocation(currentBarrier, location, storageKind);
+    }
+
+    @Override
+    public BarrierType readBarrierType(LocationIdentity location, ValueNode address, Stamp loadStamp) {
+        if (location instanceof HotSpotReplacementsUtil.OopHandleLocationIdentity) {
+            assert loadStamp instanceof AbstractObjectStamp : loadStamp;
```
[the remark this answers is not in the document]
dougxc: Also keep in mind that these assertions are only compiled into libgraal if assertions are enabled when building libgraal. That is, when running a JVM with libgraal, the value of `-ea` and `-esa` does not affect libgraal. For that reason, consider using `GraalError.guarantee`.

davleopo:
I forgot to mention here is the [SyncPort class](https://github.com/oracle/graal/blob/master/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/SyncPort.java) we use it to denote hot spot ports. When the upstream implementation changes we are informed in our internal testing.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/amd64/shenandoah/AMD64HotSpotShenandoahLoadRefBarrierOp.java
```diff
@@ -0,0 +1,222 @@
...
+import static jdk.graal.compiler.lir.LIRInstruction.OperandFlag.REG;
+import static jdk.vm.ci.code.ValueUtil.asRegister;
+
+/**
+ * X86 backend for the Shenandoah load-reference barrier.
+ */
+public class AMD64HotSpotShenandoahLoadRefBarrierOp extends AMD64LIRInstruction {
+    public static final LIRInstructionClass<AMD64HotSpotShenandoahLoadRefBarrierOp> TYPE = LIRInstructionClass.create(AMD64HotSpotShenandoahLoadRefBarrierOp.class);
+
+    enum GCStateBitPos {
```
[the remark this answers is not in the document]
tkrodriguez: Should these be values that we fetch from the host JVM?  Or at least validate their equivalence in GraalHotSpotVMConfig?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahLoadRefBarrierOp.java
mur47x111: Right, we use `@SyncPort` to monitor changes in OpenJDK. For example, if this handcrafted assembly needs to be synchronized with the code in [this section of shenandoahBarrierSetAssembler_aarch64.cpp](https://github.com/openjdk/jdk/blob/a2743bab4fd203b0791cf47e617c1a95b05ab3cc/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp#L232-L309) (please double check), we include the following annotation:
```
    // @formatter:off
    @SyncPort(from = "https://github.com/openjdk/jdk/blob/a2743bab4fd203b0791cf47e617c1a95b05ab3cc/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp#L232-L309",
              sha1 = "4ed44f985dfdca39bf93c6d306a378be4bf88fe7")
    // @formatter:on
```
and then use 
```
HOTSPOT_PORT_SYNC_CHECK=true mx build
```
to generate the correct hash.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahLoadRefBarrierOp.java
```diff
@@ -0,0 +1,233 @@
...
+        public int getValue() {
+            return this.value;
+        }
+    }
+
+    private final HotSpotProviders providers;
+    private final GraalHotSpotVMConfig config;
+
+    @Def({REG}) protected AllocatableValue result;
+    @Use({REG}) protected AllocatableValue object;
```
mur47x111, line 114: Note that here `object` and `result` may be allocated to the same register, which seems fine

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahLoadRefBarrierOp.java
```diff
@@ -0,0 +1,233 @@
...
+                        masm.lsr(64, rtmp2, objectRegister, HotSpotReplacementsUtil.shenandoahGCRegionSizeBytesShift(config));
+                        masm.ldr(8, rtmp2, AArch64Address.createRegisterOffsetAddress(8, rtmp1, rtmp2, false));
+                        masm.cbnz(8, rtmp2, slowPath);
+                        masm.jmp(done);
+                    }
+                });
+            }
+            // Call runtime slow-path LRB in out-of-line slow-path.
+            crb.getLIR().addSlowPath(this, () -> {
+                try (AArch64MacroAssembler.ScratchRegister tmp1 = masm.getScratchRegister(); AArch64MacroAssembler.ScratchRegister tmp2 = masm.getScratchRegister()) {
```
mur47x111, line 199: @tkrodriguez will we generate multiple copies of this in the same compilation unit?

tkrodriguez: Yes, `addSlowPath` always makes a unique copy of the slow path since it's dispatched with branches.  Each copy is slightly distinct because of register usage.  For the barriers, we could make these more compact by having custom slow path stubs with entry points specialized by register signature but that would only save a few instructions but maybe it would add up.

tkrodriguez:
`mx unittest -XX:+UseShenandoahGC jdk.graal.compiler.hotspot.test.GCBarrierEmissionTest` shows several problems.  On aarch there's a cbnz with size 8 but it should be 32 I think.
```
java.lang.AssertionError: 8
	at jdk.graal.compiler/jdk.graal.compiler.asm.aarch64.AArch64MacroAssembler.cbnz(AArch64MacroAssembler.java:1768)
	at jdk.graal.compiler/jdk.graal.compiler.hotspot.aarch64.shenandoah.AArch64HotSpotShenandoahLoadRefBarrierOp.lambda$emitCode$0(AArch64HotSpotShenandoahLoadRefBarrierOp.java:192)
	at jdk.graal.compiler/jdk.graal.compiler.lir.LIRInstruction$LIRInstructionSlowPath.emitSlowPathCode(LIRInstruction.java:78)
	at jdk.graal.compiler/jdk.graal.compiler.lir.asm.CompilationResultBuilder.emitSlowPath(CompilationResultBuilder.java:589)
```
and `jdk.graal.compiler.hotspot.test.GCBarrierEmissionTest#getAndSetBarrier` produces an unschedulable graph.

This test might be useful for you to look more closely at.  I'd originally written it as a testbed for the ZGC port.  It's designed so that you can compare the output of C2 and Graal for all the core barrier patterns.  See https://github.com/oracle/graal/blob/master/compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/hotspot/test/GCBarrierEmissionTest.java#L53.  I have a little utility to split each nmethod dump into it's own file so you can do pairwise comparison if you are interested in that.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahLoadRefBarrierOp.java
rkennke: > Note that here `object` and `result` may be allocated to the same register, which seems fine

Yes, that is intentional. Thanks.

tkrodriguez:
Thanks for updates.  There are various style problems that need to be fixed.  Unused imports, guarantee should use `%s` and missing copyrights.  This is my script to run the minimal style check locally in parallel which speeds up the whole process

```
#!/bin/bash
export ECLIPSE_EXE=/Users/tkrodrig/Downloads/eclipse-4.26/Eclipse.app/Contents/MacOS/eclipse
set -e

# Use tmp build output for ecj so we don't have to recompile
tmpecjdir=/tmp/mxgatestyle-ecj.$$
trap "rm -fr $tmpecjdir" 2 0

# ensure it's built first
mx build
cat <<EOF | parallel -j 5 ::::
mx gate --task-filter SpotBugs
JDT=builtin MX_ALT_OUTPUT_ROOT=$tmpecjdir mx gate --task-filter BuildWithEcj
mx gate --task-filter CodeFormatCheck
mx gate --task-filter Checkstyle
mx unittest CheckGraal
EOF
```

The CodeFormatCheck step requires a downloaded eclipse 4.26 binary to work but you can probably skip that.  The BuildWithEcj runs separately because we don't want to overwrite the javac compiled files.

rkennke:
@tkrodriguez The finagle-http benchmark failed because of missing stuff in the CAS-ref barriers. I implemented those parts now. I also took a stab at the truffle thing you mentioned, at least on arm64, but I am not even sure that it is correct. Is there any way to test it? The x86 part of that is still missing, and I'll be off to vacation for the next two weeks. I'll try to squeeze it in later tonight, or maybe tomorrow, but if I can't, it'll have to wait. Same for the gate failures (at least part of them are waiting for https://github.com/openjdk/jdk/pull/25552 from OpenJDK upstream anyway).

tkrodriguez:
No rush on any of this, but I just wanted keep you posted on what I'm seeing with our testing.

As far as the truffle part, you can at least minimally exercise it with the truffle unit tests in the compiler suite.  More extensive testing requires running the language tests.  I've got tasks in our CI to exercise shenandoah that will test that out.  So you can make an initial implementation that passes the basic unit tests and we can stress it in our gate.

The implementation should just be a directly invokeable entry point into the normal load barrier LIR op.  So it should be mostly just be refactoring the op and maybe some manual tmp selection that's compatible with the live registers on entry.  I can help you work out any kinks.  It's an annoying part of the implementation.

rkennke:
> No rush on any of this, but I just wanted keep you posted on what I'm seeing with our testing.
> 
> As far as the truffle part, you can at least minimally exercise it with the truffle unit tests in the compiler suite. More extensive testing requires running the language tests. I've got tasks in our CI to exercise shenandoah that will test that out. So you can make an initial implementation that passes the basic unit tests and we can stress it in our gate.
> 
> The implementation should just be a directly invokeable entry point into the normal load barrier LIR op. So it should be mostly just be refactoring the op and maybe some manual tmp selection that's compatible with the live registers on entry. I can help you work out any kinks. It's an annoying part of the implementation.

I made a blind/best-effort implementation for the Truffle frame-setup parts, based on what I could understand from the context and the ZGC implementation. It's pretty straightforward. An open question in the x86 parts is: I need two temporary registers, and I am not sure which ones would be free to use. I used r9 and r11, they are caller-saved in normal calling conventions, but I am not sure if this is the correct assumption in that code. Also, I haven't tested it, yet. Let me know if you want anything else changed. I might not get to it before June 23, though...

rkennke:
@tkrodriguez @mur47x111 @dougxc @davleopo I merged latest master, the missing symbol problem disappeared, as expected. Then I fixed a problem with a unit-test complaining about guarantee() calls with string concats, which made the gate tests clean. I also squashed all changesets into a single commit (again). I think I addressed all review comments so far, it should be ready for re-review now.

tkrodriguez:
Thanks.  I'll launch our internal gate and see where we stand on testing.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahCompareAndSwapOp.java
```diff
@@ -0,0 +1,218 @@
...
+
+            emitCompareAndSwap(masm, accessKind, address, result, expected, newVal, memoryOrder, setConditionFlags);
+            // EQ flag set iff success. result holds value fetched.
+
+            masm.jmp(done);
+
+            if (setConditionFlags) {
+                masm.bind(resultNullFailure);
+                // Clear zero flag to indicate failure.
+                masm.subs(32, zr, zr, 1);
```
tkrodriguez: This doesn't appear to be a valid instruction.  GCBarrierEmissionTest and UnsafeSubstitutionsTest fail because of this on aarch64.

rkennke: Ugh. I pushed a fix - please re-test. Thank you!

rkennke:
@tkrodriguez what is the status of this testing?

tkrodriguez:
Mostly the gate looks good.  There are problems with the amd64 truffle barrier that I'm fixing.  I don't know if I can push here but I can provide the patch once it's working if not.  Then I can launch the full benchmark suites too.

rkennke:
> Mostly the gate looks good. There are problems with the amd64 truffle barrier that I'm fixing. I don't know if I can push here but I can provide the patch once it's working if not. Then I can launch the full benchmark suites too.

Ok, thank you! What is the status of it? If you have a patch, I will happily intergrate it into this PR.

rkennke:
@tkrodriguez @thomaswue ping - what is the status of this PR? I'm waiting for a fix for aarch64/truffle from Tom, and it also is waiting for an approval and hopefully can be integrated soon (in time for jdk25?)

thomaswue:
@rkennke Sorry for not being enough responsive here. I will check where this stands.

tkrodriguez:
Sorry I haven't got back to you.  I'll finish the truffle changes today and push them.  @davleopo and @gergo- maybe you could do another review pass?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahBarrierSetLIRGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+    public void emitCompareAndSwapOp(LIRGeneratorTool tool, boolean isLogic, Value address, MemoryOrderMode memoryOrder, AArch64Kind memKind, Variable result,
+                    AllocatableValue allocatableExpectedValue, AllocatableValue allocatableNewValue, BarrierType barrierType) {
+        AllocatableValue tmp1 = tool.newVariable(LIRKind.value(AArch64Kind.QWORD));
+        tool.append(new AArch64HotSpotShenandoahCompareAndSwapOp(config, providers, memKind, memoryOrder, isLogic, result, allocatableExpectedValue, allocatableNewValue, tool.asAllocatable(address),
+                        tmp1));
+    }
+
+    @Override
+    public Value emitAtomicReadAndWrite(LIRGeneratorTool tool, LIRKind readKind, Value address, Value newValue, BarrierType barrierType) {
+        // We insert the necessary barriers in the node graph, at that level it
```
davleopo, line 120: minor: this comment is a bit mislead IMHO - I guess it means the necessary barriers have been added to GraalIR in the frontend of the compiler already?

davleopo:
LGTM

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahCardBarrierOp.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    // @formatter:off
+    @SyncPort(from = "https://github.com/openjdk/jdk/blob/a2743bab4fd203b0791cf47e617c1a95b05ab3cc/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp#L373-L392",
+              sha1 = "1c3e544b6fdec2f4ca0f07b2a1d5261d55754cb9")
+    // @formatter:on
+    protected void emitCode(CompilationResultBuilder crb, AArch64MacroAssembler masm) {
+        try (AArch64MacroAssembler.ScratchRegister tmp2 = masm.getScratchRegister();
+                        AArch64MacroAssembler.ScratchRegister tmp3 = masm.getScratchRegister()) {
```
gergo-, line 77: I would suggest using the same naming as the HotSpot stub, i.e., rename `tmp2` to `rScratch1` and `tmp3` to `rScratch2`.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahCompareAndSwapOp.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    @Override
+    // @formatter:off
+    @SyncPort(from = "https://github.com/openjdk/jdk/blob/a2743bab4fd203b0791cf47e617c1a95b05ab3cc/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp#L471-L606",
+              sha1 = "553a2fb0d37f39016eda85331e8cd2421153cbfe")
+    // @formatter:on
+    public void emitCode(CompilationResultBuilder crb, AArch64MacroAssembler masm) {
+        Register address = asRegister(addressValue);
+        Register result = asRegister(resultValue);
```
gergo-, line 85: I like this more meaningful name than the original's `tmp2`, but please add a comment indicating the renaming.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahCompareAndSwapOp.java
```diff
@@ -0,0 +1,219 @@
...
+        // Step 1. Fast-path.
+        //
+        // Try to CAS with given arguments. If successful, then we are done.
+
+        emitCompareAndSwap(masm, accessKind, address, result, expected, newVal, memoryOrder, true);
+        // EQ flag set iff success. result holds value fetched.
+
+        // If expected equals null but result does not equal null, the
+        // step2 branches to done to report failure of CAS. If both
+        // expected and tmp2 equal null, the following branches to done to
```
gergo-, line 104: `tmp2` -> `result`

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahCompareAndSwapOp.java
```diff
@@ -0,0 +1,219 @@
...
+        Register result = asRegister(resultValue);
+        Register expected = asRegister(expectedValue);
+        Register newVal = asRegister(newValue);
+
+        Register tmp1 = asRegister(tmp1Value);
+        Label step2 = new Label();
+        Label done = new Label();
+        GraalError.guarantee(accessKind == AArch64Kind.QWORD || accessKind == AArch64Kind.DWORD, "must be 64 or 32 bit access");
+        int size = (accessKind == AArch64Kind.QWORD) ? 64 : 32;
+
```
gergo-, line 94: Please also add a brief comment that the control flow is different from the HotSpot stub, namely that step 4 is inlined in the slow path below.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahLoadRefBarrierOp.java
```diff
@@ -0,0 +1,206 @@
...
+            masm.ldr(8, rscratch1, gcState);
+            if (strength != ShenandoahLoadRefBarrierNode.ReferenceStrength.STRONG) {
+                // This is needed because in a short-cut cycle we may get a trailing
+                // weak-roots phase but no evacuation/update-refs phase, and during that,
+                // we need to take the LRB to report null for unreachable weak-refs.
+                // This is true even for non-cset objects.
+                // Two tests because HAS_FORWARDED | WEAK_ROOTS currently is not representable
+                // as a single immediate.
+                masm.tst(64, rscratch1, config.shenandoahGCStateHasForwarded);
+                masm.branchConditionally(AArch64Assembler.ConditionFlag.NE, slowPath);
```
gergo-, line 146: `AArch64MacroAssembler` has `tbz` and `tbnz` macro-instructions, could those be used here?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/amd64/shenandoah/AMD64HotSpotShenandoahCompareAndSwapOp.java
```diff
@@ -0,0 +1,275 @@
...
+        AMD64BaseAssembler.OperandSize opSize = AMD64BaseAssembler.OperandSize.get(accessKind);
+
+        // Remember expectedReg for retry logic below
+        mov(masm, opSize, tmp1, expectedReg);
+
+        // Step 1. Fast-path.
+        //
+        // Try to CAS with given arguments. If successful, then we are done.
+
+        // There are two ways to reach this label. Initial entry into the
```
gergo-, line 131: Misplaced comment from the AArch64 version?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/gc/shenandoah/ShenandoahBarrierSet.java
```diff
@@ -0,0 +1,362 @@
...
+            default:
+                throw new GraalError("unexpected barrier type: " + barrierType);
+        }
+    }
+
+    private void addLoadReferenceBarrier(FixedAccessNode node, AddressNode address, BarrierType barrierType) {
+        GraalError.guarantee(node != null, "input value must not be null");
+        StructuredGraph graph = node.graph();
+        boolean narrow = node.stamp(NodeView.DEFAULT) instanceof NarrowOopStamp;
+        ValueNode uncompressed = maybeUncompressReference(node, narrow);
```
gergo-, line 245: It seems really important here that the `maybeUncompressReference` is done without unique because we don't want to reuse an existing uncompress node that is a usage of `node`. The implementation does this correctly, but you could make the requirement more explicit like this:
```
        Graph.Mark beforeUncompression = graph.getMark();
        ValueNode uncompressed = maybeUncompressReference(node, narrow);
        if (uncompressed != node) {
            GraalError.guarantee(graph.isNew(beforeUncompression, uncompressed), "we must not reuse an existing uncompress node");
        }
```

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/phases/common/WriteBarrierAdditionPhase.java
```diff
@@ -37,19 +37,29 @@
...
+        this.stage = stage;
+    }
+
     @Override
     public Optional<NotApplicable> notApplicableTo(GraphState graphState) {
         return NotApplicable.ifAny(
-                        NotApplicable.ifApplied(this, StageFlag.BARRIER_ADDITION, graphState),
-                        NotApplicable.unlessRunAfter(this, StageFlag.MID_TIER_LOWERING, graphState),
+                        NotApplicable.ifApplied(this, stage, graphState),
+                        NotApplicable.unlessRunAfter(this, stage == StageFlag.BARRIER_ADDITION ? StageFlag.MID_TIER_LOWERING : StageFlag.LOW_TIER_LOWERING, graphState),
```
gergo-, line 54: I think this would look nicer if we renamed `BARRIER_ADDITION` to `MID_TIER_BARRIER_ADDITION` for symmetry. The constructor should check that the stage flag passed in is one of `MID_TIER_BARRIER_ADDITION` or `LOW_TIER_BARRIER_ADDITION`.

gergo- (commented):
Sorry for not joining earlier review rounds. I left some style/documentation comments, looks good overall.

tkrodriguez:
So I have working truffle entry point read barrier changes but that exposed a deeper problem.  The current changes rely on running after FixReadsPhase or not having any floating reads but the economy configuration doesn't run FixReadsPhase.  We're also not strict about never emitting FloatingReadNodes in the mode so some reads end up without a barrier because they are floating.  We're moving towards enforcing no floating reads in economy but it's not how it works yet.

So I'm looking at moving the read barrier into the LIR so it matches the ZGC implemenation.  This would also remove all the changes to where WriteBarrierAdditionPhase is run.  The way I'm approaching this is changing the oop reading node to always return uncompressed oops when a read barrier is used, which avoids any dancing around with compression in the barrier itself.  It is largely straightforward and is mostly passing the unit tests.  I'm chasing a crash or two but I should have something later today that we can evaluate.

rkennke:
> So I have working truffle entry point read barrier changes but that exposed a deeper problem. The current changes rely on running after FixReadsPhase or not having any floating reads but the economy configuration doesn't run FixReadsPhase. We're also not strict about never emitting FloatingReadNodes in the mode so some reads end up without a barrier because they are floating. We're moving towards enforcing no floating reads in economy but it's not how it works yet.
> 
> So I'm looking at moving the read barrier into the LIR so it matches the ZGC implemenation. This would also remove all the changes to where WriteBarrierAdditionPhase is run. The way I'm approaching this is changing the oop reading node to always return uncompressed oops when a read barrier is used, which avoids any dancing around with compression in the barrier itself. It is largely straightforward and is mostly passing the unit tests. I'm chasing a crash or two but I should have something later today that we can evaluate.

That sounds great, thank you! Let me know if I can help with anything!

tkrodriguez:
That idea didn't really pan out as it would have required changes to FloatingReadNode that I didn't really want to make.  So I'm looking into enforcing the rule that there should be no floating reads in economy mode.  That's likely to be true soon as part of another PR but it's not being enforced there.  It's mostly straightforward to enforce and we can resolve any conflicts in how it's done once we're ready to merge.  I kind of wanted to push this kind of enforcement anyway, and this gives me a good reason.

tkrodriguez:
I have a set of changes that disallow floating reads from reaching the backend which makes the read barrier strategy for shenandoah work in economy.  It's not completely clear whether that will be pushed as part of this PR or if we might want to separate it.  Could you rebase to the latest master?  I need to update the CI tasks which recently changed.  Once you've rebased I can update my internal PR and mirror that to github with my fixes on top.  Then we can finish any required work there.  Sound good?

rkennke:
> I have a set of changes that disallow floating reads from reaching the backend which makes the read barrier strategy for shenandoah work in economy. It's not completely clear whether that will be pushed as part of this PR or if we might want to separate it. Could you rebase to the latest master? I need to update the CI tasks which recently changed. Once you've rebased I can update my internal PR and mirror that to github with my fixes on top. Then we can finish any required work there. Sound good?

Sounds good! I rebased my branch to latest master.

tkrodriguez:
So I've taken your changes and applied some fixes including the changes to have only fixed reads in backend, and the combined PR is at https://github.com/oracle/graal/pull/11941.  The fix reads change will be addressed separately for clarity under https://github.com/oracle/graal/pull/11942 and I will rebase once that merges.  I can address Gergos final comments there but it seems like it's all working.  I had a full clean gate including all benchmarks with shenanadoah last week though I had seem a crash or two during early testing so there might still be some problems lurking.  Once it's all clean and ready I'll do a final squash since the history is getting very ugly.  Does that sound all good?

rkennke:
> So I've taken your changes and applied some fixes including the changes to have only fixed reads in backend, and the combined PR is at #11941. The fix reads change will be addressed separately for clarity under #11942 and I will rebase once that merges. I can address Gergos final comments there but it seems like it's all working. I had a full clean gate including all benchmarks with shenanadoah last week though I had seem a crash or two during early testing so there might still be some problems lurking. Once it's all clean and ready I'll do a final squash since the history is getting very ugly. Does that sound all good?

Yep, perfect! Thank you for your help!

tkrodriguez:
I have rebased to master and incorporated most of Gergos suggestions but wasn't sure about the two I left unresolved.  The new state is over at https://github.com/oracle/graal/pull/11941.  @rkennke maybe you could provide guidance on the last 2 unresolved comments and my last set of review comment commits?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahLoadRefBarrierOp.java
rkennke: tbz and tbnz have a limit on how far away they can branch, and since we don't know where the slow stub is placed, we need to use the 'long' version.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/amd64/shenandoah/AMD64HotSpotShenandoahCompareAndSwapOp.java
rkennke: Right, this whole block (except the first 3 lines) can be removed.

rkennke:
> I have rebased to master and incorporated most of Gergos suggestions but wasn't sure about the two I left unresolved. The new state is over at #11941. @rkennke maybe you could provide guidance on the last 2 unresolved comments and my last set of review comment commits?

I commented on Gergos comments, I don't think we can use tbz/tbnz because of the jump offset limitation. The second part of the comment can be deleted. Your commit looks fine. Thank you for taking care of all this!

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/amd64/shenandoah/AMD64HotSpotShenandoahCompareAndSwapOp.java
tkrodriguez: By `this whole block` you mean the block comment starting with `// There are two ways to reach this label. Initial entry into the`?

rkennke: Correct.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/hotspot/aarch64/shenandoah/AArch64HotSpotShenandoahLoadRefBarrierOp.java
gergo-: OK thanks. If you have reason to touch the code again, it would be great to have a comment mentioning this limit.

tkrodriguez:
This was merged under https://github.com/oracle/graal/commit/3c7b5cb092a606d2d9c421dbe225bff16789fb2d

reneleonhardt:
@rkennke Thank you very much for all your work over half a year!

simonis:
@fniephaus, @alina-yur could you please add this PR to the [GraalVM Community Roadmap](https://github.com/orgs/oracle/projects/6/views/1)? I think it is worth being mentioned there.

Also, it seems that the [GraalVM Community Roadmap](https://github.com/orgs/oracle/projects/6/views/1) view is a little bit out of sync, as it still mentions "GraalVM 25" which was released in September 2025 under "Upcoming Releases". Maybe you can fix that as well?

alina-yur:
Hi @simonis, congrats on reaching the first milestone :) 

The topic is already in the roadmap as [#12237](https://github.com/orgs/oracle/projects/6/views/1?pane=issue&itemId=130712659&issue=oracle%7Cgraal%7C12237). I linked this PR there and moved the item to "In Progress". I also moved the 25 items to released, thank you for the suggestions.

simonis:
> Hi @simonis, congrats on reaching the first milestone :)
> 
> The topic is already in the roadmap as [#12237](https://github.com/orgs/oracle/projects/6/views/1?pane=issue&itemId=130712659&issue=oracle%7Cgraal%7C12237). I linked this PR there and moved the item to "In Progress". 

@alina-yur, [#12237](https://github.com/orgs/oracle/projects/6/views/1?pane=issue&itemId=130712659&issue=oracle%7Cgraal%7C12237) is about porting the Shenandoah GC to Substrate VM / Native Imaeg. This change here, which indeed might be badly named, is about adding Shenandoah support to the Graal JIT compiler running on HotSpot (similar to [[GR-27475] Add ZGC support](https://github.com/oracle/graal/pull/6170) which is on the roadmap as  [Add support for ZGC on HotSpot](https://github.com/orgs/oracle/projects/6/views/7?pane=issue&itemId=10947341&issue=oracle%7Cgraal%7C5050). So could you please add it to the roadmap as something like "Add support for Shenandoah on HotSpot" with the status "Done".

> I also moved the 25 items to released, thank you for the suggestions.

Thanks for moving 25 to released. Do you know why the ordering is so strange. E.g. for me "GraalVM 25" is sorted between "GraalVM for JDK 17" and "22.3.0 Release" instead of appearing on top of the list?

PR #10974, opened 2025-04-03
Kiwi1969 zos

kiwi1969:
Changes to avoid "Unsupported OS name z/OS" when trying to run on Z/OS with IBMJ9 VM


kiwi1969:
Still experimenting with these changes

PR #11038, opened 2025-04-15
[GR-64209] Adopt attach API changes and fix @BasedOnJDKFile annotations.

graalvmbot:
Updates the attach API from `JDK-24+18` to `JDK-25+18` (except for any functionality related to version 2 and `streaming`, which are both not supported at the moment).


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/attach/AttachListenerThread.java
```diff
@@ -132,20 +148,245 @@ private static void handleException(AttachOperation op, Throwable e) {
...
+    }
+
+    private static AttachOperation createAttachOperation(AttachSocketChannel channel, ArrayList<String> values) {
+        /* Parse the name and the options. */
+        String name;
+        String nameAndOptions = values.getFirst();
+
+        int optionStart = nameAndOptions.indexOf(' ');
+        if (optionStart != -1) {
+            name = nameAndOptions.substring(0, optionStart);
```
roberttoyonaga, line 287: It doesn't hurt to do this check, but I think we can just assume there will never be any options specified by the client. Because at this point it has [already asked the server "getversion"](https://github.com/alexmenkov/jdk/blob/master/src/jdk.attach/linux/classes/sun/tools/attach/VirtualMachineImpl.java#L127) and knows we don't support options, so [it won't supply any](https://github.com/openjdk/jdk/pull/23405/files#diff-f90f733fa4c70b9b92e40856d5ec7a287f9586fbde49148ce986316e29aceabcR544).

roberttoyonaga (approved):
This looks good to me! Thanks @christianhaeubl

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/attach/AttachListenerThread.java
christianhaeubl: Good point - I will probably keep the check for now though (it makes the diff against HotSpot a bit smaller).

PR #11131, opened 2025-05-05
Docs: Fix `--exclude-config` documentation

zakkak:
`--exclude-config` seems to take a space-separated pair instead of a
comma-separated one.

Thanks to @gsmet for noticing it.

cc @fniephaus @olyagpl

### DIFF
--- a/docs/reference-manual/native-image/BuildOptions.md
+++ b/docs/reference-manual/native-image/BuildOptions.md
@@ -90,7 +90,7 @@ Run `native-image --help` for help on build options.
 
 Run `native-image --help-extra` for help on additional options.
 
-* `--exclude-config`: exclude configuration for a comma-separated pair of class path/module path pattern and resource pattern. For example: `--exclude-config foo.jar,META-INF\/native-image\/.*.properties` ignores all properties files in _META-INF/native-image_ in all JAR files named _foo.jar_.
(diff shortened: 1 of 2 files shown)


gsmet (approved):

olyagpl:
Thanks @zakkak, I'll take it from here.

fniephaus:
@zakkak should we fix the option rather than the docs? We typically use commas as delimiter I think.

zakkak:
> @zakkak should we fix the option rather than the docs? We typically use commas as delimiter I think.

That would be more consistent (although I didn't check) but it would break existing code so we would need to support both for some transition period. I am OK with either aproach.

(Sorry for the delayed reply, I wrote the reply and forgot to press the "Comment" button)

zakkak:
> Agreed. Why don't we add comma splitting and keep the docs as they are (aka not document the space delimiter). We can leave a comment in the code to remove it in the future?

Sounds good to me.

fniephaus:
Oh, I checked the code and the flag indeed consumes two arguments: a jar name and a regular expression. It's a bit odd, also to use a comma instead. Maybe we could've used an `=` instead or so, but now that it's there and used, maybe it's indeed easiest to just fix the docs.

WDYT, @zakkak?

zakkak:
I am really OK with any approach as long as the documentation matches the implementation :)

fniephaus (approved):
LGTM. @olyagpl could you please integrate this?

PR #11149, opened 2025-05-07
[GR-64787] Enable --install-exit-handlers by default.

graalvmbot:
Enable `--install-exit-handlers` by default and deprecate the option.

Image size increase is negligible (10 new types and 53 new methods), and there is only 2.5% extra startup instructions (68498) for "Hello, World!" on Linux.


substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/SubstrateOptions.java
```diff
@@ -587,6 +587,11 @@ protected void onValueUpdate(EconomicMap<OptionKey<?>, Object> values, Boolean o
                 }
             }
         };
+
+        @APIOption(name = "install-exit-handlers", deprecated = "Enabled by default for executables. For shared libraries (when --shared is used), use the experimental option -H:+InstallJavaExitHandlersForSharedLibrary.")//
+        @Option(help = "Provide java.lang.Terminator exit handlers", deprecated = true, deprecationMessage = "Enabled by default for executables. For shared libraries (when --shared is used), use the experimental option -H:+InstallJavaExitHandlersForSharedLibrary.")//
+        protected static final HostedOptionKey<Boolean> InstallExitHandlers = new HostedOptionKey<>(false);
```
zakkak, line 593: Shouldn't this be true?

```suggestion
        protected static final HostedOptionKey<Boolean> InstallExitHandlers = new HostedOptionKey<>(true);
```

zakkak (commented):
Hi @vjovanov , Quarkus is already setting up its own signal handlers so we would like to have the option to disable this (at least for some transition period). Is it safe to assume `-H:-InstallExitHandlers` will survive at least till GraalVM for JDK 29? Thanks.

zakkak (approved):
LGMT. Thanks @vjovanov

Issue #10239, opened 2024-12-05
[Espresso] Support running without native access.


gilles-duboscq:
**Feature request**

**Is your feature request related to a problem? Please describe.**
Espresso requires native access to run the JDK's native libraries. Native access is also requires due to the implementation of Unsafe.
As a result, to run espresso in a polyglot context it requires all access.

**Describe the solution you'd like.**
*Some* of the JDK's native methods could be implemented without having to call native code.
Unsafe can be virtualized in a way that it doesn't require native access.

**Describe who do you think will benefit the most.**
Users who want to run espresso in a polyglot context without having to enable all access or native access. 

**Describe alternatives you've considered.**
* The security manager can be used to impose some restrictions in the espresso context but it is being removed. 
* Sulong can be used to run the JDK's native code but that still requires native access in order to call into system libraries. It is also not always straightforward to produce bitcode for the JDK for all platforms.

**Additional context.**
The I/O part of the JDK would be taken care of by #4625.
The following JDK 21 modules contain native code and are relevant for espresso:
* `java.base`
* `java.desktop`
* `java.management`
* `java.rmi`
* `jdk.jdi`
* `jdk.management`

`java.desktop` is too vast to realistically be implemented without calling into the native code.

The following JDK 21 modules contain native code but are unlikely to be used in a no-native espresso context:
* `java.security.jgss` (kerberos)
* `java.smartcardio` (ISO/IEC 7816 smart cards)
* `jdk.crypto.cryptoki` (PKCS#11 tokens)

The following JDK 21 modules contain native code but are not relevant for espresso:
* `jdk.hotspot.agent`
* `jdk.jdwp.agent`
* `jdk.jpackage`

**Express whether you'd like to help contributing this feature**
If you'd like to contribute, please read the [contribution guide](https://www.graalvm.org/community/contributors/).

kumargu:
thanks @gilles-duboscq for putting this up!


> Unsafe can be virtualized in a way that it doesn't require native access.
I am noob in this area, but this seems a lot of work. 

 Would you be able to give a high level estimate of the quantity of work? (sorry for asking the hard part of it so early, but i am evaluating if we (Opensearch) would be able to use it any time in Q1 2025).

kumargu:
tagging Opensearch's meta issue for referencing https://github.com/opensearch-project/OpenSearch/issues/1687

gilles-duboscq:
[...]

Issue #4625, opened 2022-06-07
[Espresso] Support guest Java filesystems backed by the Truffle/polyglot VFS API.


mukel:
**Feature request** 

**Is your feature request related to a problem? Please describe.**

Espresso uses the Java classes and native libraries from it's `--java.JavaHome`, it uses the IO implementation provided there.
Having a guest Java filesystem for both java.io and java.nio backed by the Truffle/polyglot VFS API would allow to sandbox IO within an Espresso context.

**Describe the solution you'd like.**
An option to enable and replace the default OpenJDK IO classes by a custom implementation backed by the Truffle VFS e.g. something like `--java.IO=truffle`.

**Describe who do you think will benefit the most.**

Embedders of Espresso and developers trying to overcome the `SecurityManager` deprecation.

**Describe alternatives you've considered.**

- Using the guest Security Manager (it works, but is being deprecated/removed soon).
- Hooking a few native methods for the vanilla Java classes (has different implementations per platform + breaking changes since native methods are usualy not part of the API).
- Implement a Java filesystem (legacy java.io and java.nio) that is Truffle aware, with some substitutions/instrinsics on the Espresso, all written in Java.

The result would be to have `TruffleFileSystem/TruffleFileSystemProvider/TrufflePath` instead of `UnixFileSystem/UnixFileSystemProvider/UnixPath` as the default implementation, similar on other platforms.

**Additional context.**

Is also important in this mode (Truffle IO) mostly used by embedders to block the default Java classes to be used/loaded.
Native libraries are also a concern, since cannot be dlopen-ed from memory.
The same idea applies for sockets but there's no Truffle socket API yet.

**Express whether you'd like to help contributing this feature**
Internally tracked as `GR-38920`.
If you'd like to contribute, please read the [contribution guide](https://www.graalvm.org/community/contributors/).

PR #11159, opened 2025-05-08
[GR-64626] Implement boot libraries in Java for Espresso.

graalvmbot:


rakachan:
This is the preliminary PR that kickstarts the work on #10239 and #4625.
For trying it out, run espresso with `--experimental-options --java.UseEspressoLibs=true`

Currently, this is only runs Hello World.

For impl details:
* We are re-using the Espresso substitution mechanisms to implement guest native methods. These substitutions get associated with a given "library" through the `group` annotation value. This allows associating `java.lang._` methods to `libjava` and `sun.nio._` methods to `libnio` for example.
* The `EspressoLibsNativeAccess` class intercepts the various requests to load native libraries, and provides handles to our own substitution groups. The VM is in control of how `libjava` is loaded.
* By providing our own "native" implementation for `NativeLibraries.load` and `NativeLibraries.findBuiltinLib`, we can intercept the loading of known libraries (Such as `libnio`, `libzip`, etc), and link against our own.
* Note: Though created with the same mechanisms, these native-"Substitutions" are not like the original substitutions: They do not replace the guest call unconditionally, and are required to be native-linked (in `Method.findJniCallTarget`) before they can be executed. As such, methods that are not `native` can not be substituted this way.

PR #11180, opened 2025-05-13
Introduce unmaintained category

ShalokShalom:
This introduces an unmaintained category and moves Yona too it (3 years without update)


oubidar-Abderrahim:
Hi, Thank you for opening this PR, unfortunately since the OCA was not signed yet, we cannot review or follow up on this.
Once you have a valid OCA, please create a new PR, thank you

Issue #11103, opened 2025-04-29
[Native Image] Unable to use Attach API from native-image client application


dtetreau-Oracle:
### Describe the Issue

I am attempting to use the [VirtualMachine Attach API](https://docs.oracle.com/en/java/javase/24/docs/api/jdk.attach/com/sun/tools/attach/VirtualMachine.html) from inside a Java program compiled with GraalVM 24 native-image. I need this API because I am writing a monitoring tool which needs the ability to turn on JMX management features and trigger JFR captures for Java applications which may not have been started with JMX-enabling command line flags.

When run on a normal JVM, the code works normally and I am able to call e.g. [VirtualMachine.attach](https://docs.oracle.com/en/java/javase/24/docs/api/jdk.attach/com/sun/tools/attach/VirtualMachine.html#attach(java.lang.String)) and [startLocalManagementAgent](https://docs.oracle.com/en/java/javase/24/docs/api/jdk.attach/com/sun/tools/attach/VirtualMachine.html#startLocalManagementAgent()), which ultimately lets me open up a JMX connection against the target host. However, when I compile my application to a native image using the currently-latest GraalVM 24 version (`graalvm-24-jdk-24.0.1-1.el9.x86_64` and `graalvm-24-native-image-24.0.1-1.el9.x86_64`) then the VirtualMachine.attach call fails.

```
java.lang.UnsatisfiedLinkError: Can't load library: attach | java.library.path = [/usr/lib64, /lib64, /lib, /usr/lib]
    at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.NativeLibraries.loadLibraryRelative(NativeLibraries.java:141)
    at java.base@24.0.1/java.lang.ClassLoader.loadLibrary(ClassLoader.java:100)
    at java.base@24.0.1/java.lang.Runtime.loadLibrary0(Runtime.java:822)
    at java.base@24.0.1/java.lang.System.loadLibrary(System.java:1663)
    at jdk.attach@24.0.1/sun.tools.attach.VirtualMachineImpl.<clinit>(VirtualMachineImpl.java:404)
    at jdk.attach@24.0.1/sun.tools.attach.AttachProviderImpl.attachVirtualMachine(AttachProviderImpl.java:56)
    at jdk.attach@24.0.1/com.sun.tools.attach.VirtualMachine.attach(VirtualMachine.java:201)
```

[...]

PR #11199, opened 2025-05-14
[GR-65559] Add JNI registration Attach API

dev-jonghoonpark:
related issue: https://github.com/oracle/graal/issues/11103

I resolved an issue where using the attach method in an application built with native-image resulted in the error:
`java.lang.UnsatisfiedLinkError: Can't load library: attach`

---

Additionally, during implementation, I discovered that in applications built with native-image, `VirtualMachine.list()` returns an empty list.
(This was not addressed in this PR.)

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/JNILibraryInitializer.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jni/JNILibraryInitializer.java
@@ -68,7 +68,7 @@ public class JNILibraryInitializer implements NativeLibrarySupport.LibraryInitia
     public boolean fillCGlobalDataMap(Collection<String> staticLibNames) {
         List<String> libsWithOnLoad = Arrays.asList("net", "java", "nio", "zip", "sunec", "jaas", "sctp", "extnet",
                         "j2gss", "j2pkcs11", "j2pcsc", "prefs", "verify", "awt", "awt_xawt", "awt_headless", "lcms",
-                        "fontmanager", "javajpeg", "mlib_image");
+                        "fontmanager", "javajpeg", "mlib_image", "attach");
         // TODO: This check should be removed when all static libs will have JNI_OnLoad function
         ArrayList<String> localStaticLibNames = new ArrayList<>(staticLibNames);
         localStaticLibNames.retainAll(libsWithOnLoad);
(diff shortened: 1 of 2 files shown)


roberttoyonaga:
I tested out and the funcitonality seems to be working. I used this procedure:
1. Build your branch
2. `javac`  [SimpleShortAttachClient.java](https://gist.github.com/roberttoyonaga/9b7fc8bf90b870365909e5e6a462cf38#file-simpleshortattachclient-java)
4. `mx native-unittest SimpleShortAttachClient`
5. `java -m jdk.httpserver` (In another terminal)
6. `jcmd`  and find the PID of jdk.httpserver
7. `jcmd  PID ManagementAgent.status`  verify the management agent is off
8. .`/simpleshortattachclient PID` (This should cause the Native Image binary to attach and start the management agent in the other JVM)
9. `jcmd  PID ManagementAgent.status` verify the management agent is on


```
$ jcmd
2205600 org.jetbrains.idea.maven.server.RemoteMavenServer36
2225971 jdk.httpserver
2125315 com.intellij.idea.Main
2225998 jdk.jcmd/sun.tools.jcmd.JCmd

$ jcmd  2225971 ManagementAgent.status
2225971:
Agent: disabled

$ ./simpleshortattachclient 2225971
PID.. 2226067
{sun.java.command=jdk.httpserver, sun.jvm.flags=, sun.jvm.args=-Djdk.module.main=jdk.httpserver}
{sun.java.command=jdk.httpserver, sun.jvm.flags=, sun.jvm.args=-Djdk.module.main=jdk.httpserver}


$ jcmd  2225971 ManagementAgent.status
2225971:
Agent: enabled

Connection Type: remote
Protocol       : rmi
...
```

dev-jonghoonpark:
Thank you for sharing the detailed testing procedure.
I also used a similar approach during my development.

corey-kdm:
Is there any chance this will get reviewed and (hopefully) merged?  I just ran into this issue myself.

zapster:
Hi @dev-jonghoonpark and thanks for your contribution! I'll take care of this next week or so.

roberttoyonaga:
I wonder if this should be hidden behind a hosted option like the existing `--enable-monitoring=attach` or something similar.

roberttoyonaga:
`VirtualMachine.attach` calls [`AttachProviders.provider`](https://github.com/openjdk/jdk/blob/jdk-26%2B1/src/jdk.attach/share/classes/com/sun/tools/attach/VirtualMachine.java#L194), which uses [`ServiceLoader`](https://github.com/openjdk/jdk/blob/master/src/jdk.attach/share/classes/com/sun/tools/attach/spi/AttachProvider.java#L233).  Do we need to register the `AttachProvider` service in META-INF/services/ or some other way? This seems not to be necessary based on my demo app above, but it's probably best to make sure.

zapster:
> I wonder if this should be hidden behind a hosted option like the existing `--enable-monitoring=attach` or something similar.

I don't think this is necessary. Due to the reachability handler, the library is only added if the code is reached. If not, the image size stays the same.

zapster:
> Do we need to register the `AttachProvider` service in META-INF/services/ or some other way? 

The OpenJDK implementations are already in the `META-INF/services` of the `jdk.attach` module and thus discovered automatically. If users want to add their own provider, they will basically do the same: have an entry in the `META-INF/services` of their module which will then be picked up by native image. No need to do anything special here.

olpaw (approved):

PR #11275, opened 2025-05-28
[GR-32941] Allocate TLABs from chunks instead of as chunks

graalvmbot:


roberttoyonaga:
Hi, this is great! Just a few questions:

Is it the case that this change tries to reduce the overall memory usage (by allowing TLABS to share heap chunks) but will not actually affect the frequency that the slow path is taken? 

Am I correct in understanding that even when a new TLAB is able to be created without allocating a new heap chunk there is still contention when making space in the current heap chunk? 

There is no dynamic resizing of TLABS, whenever more space is required, the old TLAB must be retired and a new one (with a different) size is created, or allocation is done directly on the heap?

PR #11369, opened 2025-06-10
[GR-64503] TRegex: update pre-defined character classes in OracleDBFlavor.

graalvmbot:


JohnTortugo:
@djoooooe - can you please elaborate what's the issue being addressed here?

How is the open source Graal community supposed to view a PR like this? The issue mentioned in the title of the PR seems to be private to Oracle. There is no description in the PR. There is no record if this PR was reviewed or not. How was it tested? The commit message is just one line... etc, etc.

PR #11400, opened 2025-06-12
[GR-65899] JFR substitutions for JDK 25+26.

roberttoyonaga:
This PR adds some substitutions related to the the new JFR features recently integrated into OpenJDK:
JEP 520: [JFR Method Timing & Tracing](https://openjdk.org/jeps/520)
JEP 509: [JFR CPU-Time Profiling (Experimental)](https://openjdk.org/jeps/509)

Without them we get warnings like this:
```
[warn][jfr,system ] Exception occurred during execution of event jdk.MethodTiming(13364). jdk.jfr.internal.JVM.drainStaleMethodTracerIds()[J [symbol: Java_jdk_jfr_internal_JVM_drainStaleMethodTracerIds or Java_jdk_jfr_internal_JVM_drainStaleMethodTracerIds__]
[warn][jfr,setting] Exception occurred when setting value "false". jdk.jfr.internal.JVM.setCPUThrottle(DZ)V [symbol: Java_jdk_jfr_internal_JVM_setCPUThrottle or Java_jdk_jfr_internal_JVM_setCPUThrottle__DZ] ==> expected: <true> but was: <false>

```


christianhaeubl (approved):
Thanks - this PR will get integrated in the next few days.

PR #11409, opened 2025-06-14
Make IGV start on GraalVM 24.0.1

JaroslavTulach:
I have problems launching IGV on GraalVM 24.0.1. It yields following error:
```
Executing: /bin/sh -c '/home/devel/NetBeansProjects/graalvm/graal/visualizer/IdealGraphVisualizer/application/target/idealgraphvisualizer/bin/idealgraphvisualizer' '--userdir' '/home/devel/NetBeansProjects/graalvm/graal/visualizer/IdealGraphVisualizer/application/target/userdir' '-J-Dnetbeans.logger.console=true' '-J-ea' '--branding' 'idealgraphvisualizer' '--jdkhome' '/home/devel/bin/graalvm-24'
WARNING: package com.sun.tools.javadoc.main not in jdk.javadoc
Error occurred during initialization of VM
java.lang.Error: A command line option has attempted to allow or enable the Security Manager. Enabling a Security Manager is not supported.
	at java.lang.System.initPhase3(java.base@24.0.1/System.java:1947)
```
The problem is that JDK 24 has removed [SecurityManager support](https://openjdk.org/jeps/486) and as the NetBeans Platform launcher needs to be adjusted to that. Such an adjustment has happened in NetBeans 26. Hence upgrading.

### DIFF
--- a/visualizer/IdealGraphVisualizer/pom.xml
+++ b/visualizer/IdealGraphVisualizer/pom.xml
@@ -117,7 +117,7 @@
         <module>ViewerApi</module>
     </modules>
     <properties>
-        <netbeans.version>RELEASE220</netbeans.version>
+        <netbeans.version>RELEASE260</netbeans.version>
         <brandingToken>idealgraphvisualizer</brandingToken>
         <swinglayouts.version>1.0.2</swinglayouts.version>
         <nbmmvnplugin.version>4.8</nbmmvnplugin.version>
--- a/visualizer/IdealGraphVisualizer/Data/src/main/nbm/manifest.mf
+++ b/visualizer/IdealGraphVisualizer/Data/src/main/nbm/manifest.mf
@@ -2,4 +2,4 @@ Manifest-Version: 1.0
 OpenIDE-Module: org.graalvm.visualizer.data
 OpenIDE-Module-Localizing-Bundle: org/graalvm/visualizer/data/Bundle.properties
 OpenIDE-Module-Specification-Version: 1.5
-
+OpenIDE-Module-Hide-Classpath-Packages: jdk.graal.compiler.graphio.**


visualizer/IdealGraphVisualizer/pom.xml
```diff
@@ -117,7 +117,7 @@
         <module>ViewerApi</module>
     </modules>
     <properties>
-        <netbeans.version>RELEASE220</netbeans.version>
+        <netbeans.version>RELEASE260</netbeans.version>
```
JaroslavTulach, line 120: With this change I can built regularly:
```bash
/graal/visualizer/IdealGraphVisualizer$ JAVA_HOME=/graalvm-21/ mvn install
```
and then execute on GraalVM 24.0.1 (which was failing previously):
```bash
/graal/visualizer/IdealGraphVisualizer$ JAVA_HOME=/graalvm-24 mvn -f application/ \
   -Dexec.vmArgs= -Dexec.appArgs= \
  nbm:cluster-app nbm:run-platform
```

visualizer/IdealGraphVisualizer/View/src/main/java/org/graalvm/visualizer/view/impl/GraphTypeModeSelector.java
```diff
@@ -45,7 +46,12 @@
 public class GraphTypeModeSelector implements ModeSelector {
     @Override
     public Mode selectModeForOpen(TopComponent tc, Mode mode) {
-        if (tc.getLookup().lookup(InputGraph.class) == null) {
+        try {
+            if (tc.getLookup().lookup(InputGraph.class) == null) {
```
JaroslavTulach: Without this change I couldn't open any file in the editor. I was getting exception like this one

obrazek

JaroslavTulach:
> @tkrodriguez: weird unit test failures when running them on 24b

- building and running unit tests needs to be done on JDK21
- looks like JDK24 increased the security checks again
- and IGV modules can no longer access Graal code in the JDK
- I haven't investigated any details, but probably some `--add-opens` or `--add-exports` when launching IGV would help


With following command line the execution gets further:
```
/graal/visualizer/IdealGraphVisualizer$ ./application/target/idealgraphvisualizer/bin/idealgraphvisualizer -J--add-exports=jdk.graal.compiler/jdk.graal.compiler.graphio.parsing.model=ALL-UNNAMED -J--add-exports=jdk.graal.compiler/jdk.graal.compiler.graphio.parsing=ALL-UNNAMED -J-Dnetbeans.logger.console=true --userdir ~/tmp/igvud3 --jdkhome ~/bin/graalvm/
```
why IGV needs to access graphio.parsing from a JDK and how you want to make sure it gets the right version on different JVMs, is beyond ambitions of this PR...

tkrodriguez:
I'm already looking into this and as you say it's more complicated than just bumping to RELEASE260.  We don't want to access jdk.graal.compiler.graphio.parsing.model from the JDK as there is already a module in IGV that contains this code.  The presence of jargraal in GraalVM seems to break this and I'm not sure how to stop that.  An Oracle JDK24, which contains only libgraal, doesn't have this problem but the actual unit tests fail other ways.  The IGV application seems to run ok for me though.  I guess your exception fix above is probably related to the jargraal problem though.

tkrodriguez:
I've got a working update at https://github.com/oracle/graal/pull/11421 including a workaround for the GraalVM JDK 24 IGV problem.  Could you test it out?

JaroslavTulach:
> We don't want to access `jdk.graal.compiler.graphio.parsing.model` from the JDK as there is already a module in IGV that contains this code. 

I see. I didn't realize it before. In such case, the 3637a2d is the right fix.

> I've got a working update at https://github.com/oracle/graal/pull/11421 including a workaround for the GraalVM JDK 24 IGV problem. Could you test it out?

Done. It seems to be broken: https://github.com/oracle/graal/pull/11421#issuecomment-2981222946 - probably 3637a2d  is necessary.

tkrodriguez:
Thanks for that fix.  I'll try that out.  If it can resolve our GraalVM 24 jargraal issue that would be great.

tkrodriguez:
That seems to have improved things, but I still need to disable JVMCI or I get exceptions like this when processing filters:
```
Error(s) encountered when executing filter Coloring:
java.lang.IllegalStateException: javax.script.ScriptException: javax.script.ScriptException: org.graalvm.polyglot.PolyglotException: TypeError: instantiate on jdk.graal.compiler.graphio.parsing.model.Properties$RegexpPropertyMatcher failed due to: Message not supported.
	at org.graalvm.visualizer.filter.CustomFilter.applyWith0(CustomFilter.java:311)
	at org.graalvm.visualizer.filter.CustomFilter.applyWith(CustomFilter.java:273)
	at org.graalvm.visualizer.filter.profiles.impl.FilterBridge.applyWith(FilterBridge.java:222)
```
Maybe other modules need to have the OpenIDE-Module-Hide-Classpath-Packages stuff?

tkrodriguez:
I just added it to every manifest.mf and that seems to have resolved the problem.  I'm respinning my changes with all your suggestions.  Thanks!

JaroslavTulach:
Long time merged, I assume.

The-Alchemist:
i can confirm it starts on GraalVM 25.

```
java -version
openjdk version "25.0.4.1" 2026-08-18
OpenJDK Runtime Environment GraalVM CE 25.3.4.1-dev+1.1 (build 25.0.4.1+1-jvmci-25.3-b22)
OpenJDK 64-Bit Server VM GraalVM CE 25.3.4.1-dev+1.1 (build 25.0.4.1+1-jvmci-25.3-b22, mixed mode, sharing)
```

PR #11570, opened 2025-07-04
Fix JFR GC events

roberttoyonaga:
We are currently writing JFR `jdk.GarbageCollection` events incorrectly. The event format has an ["event thread" data field](https://github.com/openjdk/jdk/blob/f2d2eef988c57cc9f6194a8fd5b2b422035ee68f/src/hotspot/share/jfr/metadata/metadata.xml#L375), but we are not writing it. This causes the subsequent fields to have incorrect placement and be written wrong as well. 

**Before the fix:** notice many fields are blank or zero
image
**After the fix:**
image

I've added a test to catch such issues in the future.


christianhaeubl:
Thanks, will be integrated in the next few days.

christianhaeubl (approved):

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jfr/TestGarbageCollectionEvents.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (c) 2022, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2022, 2022, Red Hat Inc. All rights reserved.
```
christianhaeubl, line 3: I will change the copyright years of this file to 2025 if this is fine with you.

roberttoyonaga: Yes please! I forgot to update that. Thanks

roberttoyonaga:
@christianhaeubl should I now open a backport PR to the GraalVM for JDK25 release branch?

christianhaeubl:
Not necessary, I already did that (but may take a few days until it gets merged and therefore mirrored to GitHub).

PR #11597, opened 2025-07-09
More accurate assertions

koutheir:
Fixed incorrect assertion message.


koutheir:
I'm still working on this PR.

Issue #11045, opened 2025-04-16
Truffle OptimizedCallTarget profile counters do not decay


JohnTortugo:
## Describe GraalVM and your environment 
 - GraalVM version: **mainline**
 - CE or EE: **CE, possibly EE as well**
 - JDK version: **JDK21 or Tip**
 - OS and OS Version: **macOS Catalina, Linux**
 - Architecture: **amd64, aarch64**
 - The output of `java -Xinternalversion`:
```
OpenJDK 64-Bit Server VM (25-internal-adhoc.dcsl.labs-openjdk) for bsd-aarch64 JRE (25-internal-adhoc.dcsl.labs-openjdk), built on 2025-04-14T21:24:30Z with clang Apple LLVM 16.0.0 (clang-1600.0.26.3)
```

## Describe the issue
- **Problem 1:** The profile for a call target does not decay. Specifically, the CallCount and CallAndLoopCount fields are never decremented and therefore once a method is considered Hot it will be forever be considered hot. 
  - **Consequence:** In an application that lives long enough most methods will eventually become "Hot" which defeats the definition of what "Hot" is.

- **Problem 2:** If a call target reached the hotness threshold at any point in the past, every time the call target is executed and there is no compiled version of it present in the code cache, Truffle will recompile the call target.
  - Consequence 1 : HotSpot flushes a nmethod associated with a Truffle compilation from the code cache. The very next time the call target is invoked Graal will compile the method and insert it into the code cache again. This happens in a cycle in some cases because the code cache is under pressure - not necessarily “Full”.
  - Consequence 2: If the call target is considered Tier2 hot at the moment of its invocation, Truffle will compile the method at Tier 1 on its next invocation and then at Tier 2 on the subsequent invocation.

- **Problem 3:** A concern that I have regarding the MaximumCompilations option is that if the application executes long enough many methods will reach this threshold and never be compiled again before the application restarts. The counter is increased in every compilation of a method, independent of the reason that made the previous compiled version of the method obsolete. I.e., we should have some policy to reset this threshold. 
  - Consequence 1: Application performance will drastically degrade overtime because as more and more methods reach the maximum compilation threshold, therefore less and less methods are compiled.

## Steps to reproduce the issue
- There is an attached example that uses GraalJS to illustrate the issues mentioned above.
- Just build and run the attached code with the latest version of Graal/GraalJS.
[...]

PR #11610, opened 2025-07-12
Detect cold CallTarget invalidation and reset its profile

JohnTortugo:
Closes: #11045 

Please review this patch that modifies Truffle to: 

1. Reset a CallTarget's profile when the `nmethod` associated with it was invalidated by HotSpot because the Code Cache's method flushing heuristics deemed the nmethod to be _cold_.

1. Limit the number of compilations that a CallTarget can have within a time period. This doesn't change the current behavior, instead it just makes the constraint more flexible. I added a new parameter, `MaximumCompilationsWindow`, that the user can set to specify a time window within which the number of compilations will be limited (currently the window is whole duration of application execution).

Tests: 
 - OSX AArch64 with LabsJDK "tip" using "mx gate"
 - Linux x65 with LabsJDK "tip" using "mx gate"
 - GitHub Actions
- We have been running an internal system with _similar_ changes for months and we didn't notice any regression.

### DIFF
--- a/truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
+++ b/truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
@@ -218,7 +218,10 @@ public final class HotSpotOptimizedCallTarget extends OptimizedCallTarget {
     protected void notifyDeoptimized(VirtualFrame frame) {
         String reason = null;
         try {
-            if (getInvalidationReasonDescriptionMethodRef != null) {
+            // This method may be called with {@code installedCode ==
+            // INVALID_CODE} when the call target is not a root compilation and
+            // the parent compilation was deoptimized.
+            if (this.installedCode != INVALID_CODE && getInvalidationReasonDescriptionMethodRef != null) {
                 reason = (String) getInvalidationReasonDescriptionMethodRef.invoke(this.installedCode);
             }
         } catch (Exception e) {
@@ -231,6 +234,7 @@ public final class HotSpotOptimizedCallTarget extends OptimizedCallTarget {
     private int getInvalidationReason() {
         try {
             if (getInvalidationReasonMethodRef != null) {
+                assert installedCode != INVALID_CODE : "Cannot get invalidation reason of INVALID_CODE";
                 return (int) getInvalidationReasonMethodRef.invoke(this.installedCode);
             }
         } catch (Exception e) {


JohnTortugo:
Can someone please take a look? /cc @chumer @dougxc ?

dougxc:
Looks reasonable to me but a more detailed review should be done by @tzezula or @chumer.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
```diff
@@ -174,9 +178,23 @@ private void tetherSpeculationLog(HotSpotNmethod nmethod) throws Error, Internal
...
+     *
+     * @return whether the currently installed code is valid/executable.
+     */
     @Override
     public boolean isValid() {
-        return installedCode.isValid();
+        boolean isValid = installedCode.isValid();
+        if (!isValid && installedCode != INVALID_CODE) {
+            if (getInvalidationReason() == ((HotSpotTruffleRuntime) runtime()).getColdMethodInvalidationReason()) {
+                invalidateExistingCode();
```
chumer: Can't say I am a fan of this solution. How do we know isValid() is called often enough?
Also it is surprising that isValid() resets the compilation profile and even can call listener methods (listeners could call into isValid() too!)

What did you consider as alternatives here?
Can you explain when / how often this is invoked? Can we call this maybe between compilations on compiler threads?

I don't think we should use this approach.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
```diff
@@ -195,4 +213,51 @@ public SpeculationLog getCompilationSpeculationLog() {
         return HotSpotTruffleRuntimeServices.getCompilationSpeculationLog(this);
     }
 
+    @Override
+    protected void notifyDeoptimized(VirtualFrame frame) {
+        runtime().getListener().onCompilationDeoptimized(this, frame, getInvalidationReasonDescription());
+    }
+
+    @TruffleBoundary
```
chumer: Why is this code behind a boundary? If this code is called during PE (not sure it is) then we need to better optimize this code, aka better profiling depending on where its used. If its never called in PE we should remove the boundary.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntimeListener.java
```diff
@@ -294,8 +294,25 @@ default void onCompilationInvalidated(OptimizedCallTarget target, Object source,
      *
      * @param target the call target whose compiled code was just deoptimized
      * @param frame
+     * @param reason optional reason why the deoptimization happened.
      */
-    default void onCompilationDeoptimized(OptimizedCallTarget target, Frame frame) {
```
chumer, line 298: This interface needs to evolve in a compatible way. Should be easy enough keeping the old method deprecated, each delegating to the other. See for example onCompilationStarted.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
```diff
@@ -537,6 +540,7 @@ public final RootNode getRootNode() {
     public final void resetCompilationProfile() {
         this.callCount = 0;
         this.callAndLoopCount = 0;
+        this.timeOfFirstCompilationInWindow = Instant.now();
```
chumer: > Limit the number of compilations that a CallTarget can have within a time period. 

tbh I can't quite follow what the rationale for this feature is. When would you use this?
The maximum number of compilation is intended for compiler bugs, and should never be hit (breakpoint toggling in the debugger for example). If we extend this to time based limit, that defeats the purpose to detect those bugs.

So I think I need a motivating example here to understand.

Also I think setting:
```
 private Instant timeOfFirstCompilationInWindow;
```
in resetCompilationProfile() seems not very intuitive. Is this really the right place to do it?

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
```diff
@@ -1024,6 +1022,58 @@ public final boolean compile(boolean lastTierCompilation) {
...
+        if (engine.maximumCompilations < 0) {
+            return false;
+        }
+
+        // If there is a window specified for the maximum number of compilations we check if we
+        // should reset the number of compilations because we overflowed the window period.
+        if (engine.maximumCompilationsWindowInMinutes > 0) {
+            long ageInMinutes = ChronoUnit.MINUTES.between(timeOfFirstCompilationInWindow, Instant.now());
+            if (ageInMinutes >= engine.maximumCompilationsWindowInMinutes) {
+                // This compilation would have been blocked if the window hadn't overflowed,
```
chumer: Minor: use multi line comments for long text.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
JohnTortugo: > tbh I can't quite follow what the rationale for this feature is. When would you use this?
The maximum number of compilation is intended for compiler bugs, and should never be hit (breakpoint toggling in the debugger for example). If we extend this to time based limit, that defeats the purpose to detect those bugs.

My usage scenario may be a little edgey because 1) the application that caused me to bring up this change is very large and the code cache is conservatively small, therefore it causes many method flushes; 2) the recompilation of cold methods (the other issue that this PR addresses) together with (1) was causing many recompilations; 3) new guest language code is added/modified/removed from the system frequently.

I understand that the maximum recompilation limit is just a failsafe and isn't intended to be the main way to limit recompilations _and I'm not trying to make it so_. What I'm looking for is to make it more flexible, so that the system can remain 'stable' until I have a chance to act on the problem. By stable I mean that there aren't too many recompilations of the call target happening (i.e., a recompilation rate) and also that the method isn't run interpreted until I restart the application.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
JohnTortugo: I had to make this a TruffleBoundary because an unit test was failing because a compilation was getting too big due to recursive inlining during PE. Since the test was triggering that I suspected that it could eventually affect an end user as well.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
JohnTortugo: > How do we know isValid() is called often enough?

Do you mean often enough so that we detect "soon" that the call target was invalidated? My understanding is that there is no requirement for how "soon" we detect the invalidation. We just need to eventually detect it and since the `isValid` method is always called before a new compilation I decided to implement things like this.

> ... listeners could call into isValid() too!

The existing code is invalidated before `onCompilationDeoptimized` is called, which prevents a recursion.

> Can we call this maybe between compilations on compiler threads?

I'm not sure what's your suggestion here. Do you mean iterate on all current compiled call targets to see if any was invalidated?

> Can't say I am a fan of this solution.

Please, let me know your suggestions!

JohnTortugo: The goal here is to reset the CT (Call Target) profile when its INSTALLED_CODE was invalidated by HotSpot because it was cold. This needs to happen before we decide to compile the CT again. Realistically speaking this kind of invalidation can only happen between executions, not during an execution.

JVMCI is not able to call a method in Truffle or GraalJIT. Therefore, it cannot downcall to notify that the CT was invalidated. Thus, the identification that the CT was invalidated is a passive event that happens in Truffle.

Approaches that we can use to detect that the CT was invalidated:
- Check before every execution of the CT in interpreter mode; before we check if the CT should be compiled.
  * Pros:
    - Relatively simple to implement.
    - This is a direct way to achieve what we want: prevent recompilation of cold method.
  * Cons:
    - Executed every time that we want to execute the CT.

- Use a separate thread to detect all invalidations.
  * Blockers:
    - Will require maintaining a list of all compiled non-invalidated CT.

- When we are installing a new code.
  * Blockers:
    - This is too late. There is already a new compilation of the CT.

JohnTortugo: What do you think?

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
chumer: So because the compilation queue is flushed often, you want to limit the maximum compilations per certain time frame? Wouldn't it be better to just reset the successfulCompilationsCount on a flush? Or do I miss something here?

If we never reset the successfulCompilationsCount we will also trigger deopt loop detection which is not limited by a time window.

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/truffle/test/MaximumCompilationsTest.java
```diff
@@ -90,4 +94,125 @@ public void onCompilationFailed(OptimizedCallTarget target, String reason, boole
...
+    public void testMaxTwoCompilationsPerMinute() throws InterruptedException {
+        Assume.assumeTrue(Truffle.getRuntime() instanceof OptimizedTruffleRuntime);
+        OptimizedTruffleRuntime optimizedTruffleRuntime = (OptimizedTruffleRuntime) Truffle.getRuntime();
+        AtomicBoolean compilationResult = new AtomicBoolean();
+
+        try (Engine eng = Engine.newBuilder().allowExperimentalOptions(true).//
+                        option("engine.BackgroundCompilation", "false").//
+                        option("engine.CompileImmediately", "true").//
+                        option("engine.MaximumCompilations", "2").//
+                        option("engine.MaximumCompilationsWindow", "1").build()) {
```
chumer: Typically we use seconds not minutes. This came to a surprise to me. Any reason this is not seconds?
We could then also spend less time on this test. 90 seconds for such a test is excessive.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
chumer: As you say, ideally we would have a callback from HotSpot on invalidate/flush which allows to take action. We do have such a callback on SubstrateVM for example. Its very unfortunate we don't have that, because we could also use it for better logging.

Thinking loud, it seems the second best solution could be to install code for a call target that calls resetCompilationProfile() which then invalidates itself. This means we wouldn't need to check anything and we would just reset once on the next invocation and not on follow-up invocations. Does that seem feasible to you?

Could we schedule a call for this? It seems like we should do a bit of brain storming on a solution here. (feel free to ping me on slack)

JohnTortugo: > Could we schedule a call for this? It seems like we should do a bit of brain storming on a solution here. (feel free to ping me on slack)

Absolutely. I'll ping you on slack.

JohnTortugo:
@chumer, I did a refactoring of the code following your suggestions. Basically, I changed two things:

- I dropped the changes related to resetting number of compilations based on time. The approach now is to reset `successfullRecompilationCount` when the method was invalidated because it was cold.
- I reverted the changes in `HotspotOptimizedCallTarget` to detect that the method was invalidated. I implemented the idea that you suggested: verifying the invalidation reason of the call target when it's about to be recompiled.

I tested the latest changes using the approach described in the issue ticket and everything worked fine. `mx unittest` and `mx gate` also didn't raise any issue.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
```diff
@@ -97,24 +98,35 @@ public HotSpotOptimizedCallTarget(EngineData engine) {
...
+        invalidateInstalledCodeWithReasonMethodRef = initialize(HotSpotNmethod.class, "invalidate", boolean.class, int.class);
+        invalidateInstalledCodeWithoutReasonMethodRef = initialize(InstalledCode.class, "invalidate", boolean.class);
+        getInvalidationReasonMethodRef = initialize(HotSpotNmethod.class, "getInvalidationReason");
+        getInvalidationReasonDescriptionMethodRef = initialize(HotSpotNmethod.class, "getInvalidationReasonDescription");
+    }
+
+    private static Method initialize(Class<?> clx, String methodName, Class<?>... args) {
         try {
-            method = InstalledCode.class.getDeclaredMethod("invalidate", boolean.class);
+            return clx.getDeclaredMethod(methodName, args);
```
chumer: Afaics these are public APIs sow why do you use getDeclaredMethod instead of getMethod?

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
```diff
@@ -97,24 +98,35 @@ public HotSpotOptimizedCallTarget(EngineData engine) {
...
-        setSpeculationLog = method;
-        method = null;
+        setSpeculationLog = initialize(HotSpotNmethod.class, "setSpeculationLog", HotSpotSpeculationLog.class);
+        invalidateInstalledCodeWithReasonMethodRef = initialize(HotSpotNmethod.class, "invalidate", boolean.class, int.class);
+        invalidateInstalledCodeWithoutReasonMethodRef = initialize(InstalledCode.class, "invalidate", boolean.class);
+        getInvalidationReasonMethodRef = initialize(HotSpotNmethod.class, "getInvalidationReason");
+        getInvalidationReasonDescriptionMethodRef = initialize(HotSpotNmethod.class, "getInvalidationReasonDescription");
+    }
+
+    private static Method initialize(Class<?> clx, String methodName, Class<?>... args) {
```
chumer: nitpick: initialize is not a great method name. I would probably use findMethod.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
```diff
@@ -195,4 +211,37 @@ public SpeculationLog getCompilationSpeculationLog() {
...
 
+    @Override
+    protected void notifyDeoptimized(VirtualFrame frame) {
+        String reason = null;
+        try {
+            if (getInvalidationReasonDescriptionMethodRef != null) {
+                reason = (String) getInvalidationReasonDescriptionMethodRef.invoke(this.installedCode);
+            }
+        } catch (Exception e) {
+            // ignore
```
chumer: We are never ignoring all Exception in Truffle. This is bad practice without good reason. If you want to ignore a particular exception make sure you capture it. I have seen many cases of this, please make sure you address them all.

JohnTortugo:
Gentle ping. @chumer - Please, take a look when you can.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntimeListenerDispatcher.java
```diff
@@ -54,7 +54,7 @@
  * {@link TruffleCompilerListener} events to {@link OptimizedTruffleRuntimeListener} events.
  */
 @SuppressWarnings("serial")
-final class OptimizedTruffleRuntimeListenerDispatcher extends CopyOnWriteArrayList<OptimizedTruffleRuntimeListener> implements OptimizedTruffleRuntimeListener, TruffleCompilerListener {
+public final class OptimizedTruffleRuntimeListenerDispatcher extends CopyOnWriteArrayList<OptimizedTruffleRuntimeListener> implements OptimizedTruffleRuntimeListener, TruffleCompilerListener {
```
chumer: Not a huge fan of making this public. But I don't have a better idea.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
chumer: nitpick: clx => clazz (our common name for that parameter)

image

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
```diff
@@ -1491,4 +1496,15 @@ public enum CompilationActivityMode {
...
     }
+
+    /**
+     * This method is intended to give subclasses a last say on whether the compilation should be
+     * aborted or not.
+     *
+     * @param callTarget - The OptimizedCallTarget for the compilation.
+     * @return true if the compilation should be aborted.
+     */
+    protected boolean shouldAbortCompilation(OptimizedCallTarget callTarget) {
```
chumer: Would it make sense to use OptimizedCallTarget#prepareForCompilation instead? You can return false too to delay the current compilation. I think if we already have a call-back we should use it.
I think you could override HotSpotOptimizeddCallTarget#prepareForCompilation

Also I find it problematic that a method is called shouldAbortCompilation does also reset the compilation profile, its somewhat surprising. Maybe we can find a different name if we keep it.
 I think its fine if prepareForCompilation does that.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotTruffleRuntime.java
```diff
@@ -748,4 +765,25 @@ private static CompilationActivityMode resolveHotSpotActivityMode(int i) {
...
+     * code cache. Otherwise, we can enter a recompilation cycle because Truffle will always see the
+     * method as hot (because the profile counters never reset). To handle this case we reset the CT
+     * profile whenever its prior compilation was invalidated because it was cold.
+     */
+    @Override
+    protected boolean shouldAbortCompilation(OptimizedCallTarget callTarget) {
+        HotSpotOptimizedCallTarget hsCallTarget = (HotSpotOptimizedCallTarget) callTarget;
+        if (hsCallTarget.getInvalidationReason() == this.coldMethodInvalidationReason) {
+            hsCallTarget.resetCompilationProfile();
+            hsCallTarget.resetInstalledCode();
```
chumer: How do we know that this is not racy with compilations installing code at the same time?
Or is this something we would accept?

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotOptimizedCallTarget.java
JohnTortugo: Thank you for pointing this out. I'll refactor it.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/hotspot/HotSpotTruffleRuntime.java
JohnTortugo: My view here is that no two threads will run this method concurrently, because the call only happens while `compilationTask` in the CT is `!= null`. Also, my understanding is that even if for some reason the system ends up invalidating a "good" compilation because of a race condition: 1) for Truffle it shouldn't matter because it will at the same time be installing new code and 2) HotSpot will just drop the previous compilation from the CC because it will eventually be cold. Please, let me know if this doesn't make sense.

I moved this method to HotSpotOptimizedCallTarget::prepareForCompilation as you suggested in another comment.

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedTruffleRuntime.java
JohnTortugo: I moved the code to `HotSpotOptimizeddCallTarget#prepareForCompilation`.

JohnTortugo:
@chumer - I believe I addressed all your comments. Please, take a look again when you can.

chumer:
@jchalou  please give this another pair of eyes. If its fine feel free to integrate.

chumer:
@JohnTortugo any chance we could have a smoke test for this behavior? That is reasonably complex? Would be great to be notified if this breaks.

PR #11769, opened 2025-07-23
[GR-69121] Skip emission of internal JFR ThreadPark events

roberttoyonaga:
When synchronization such as monitor enter or wait is used, `Unsafe.park` is used under-the-hood to block threads. This results in not only `jdk.JavaMonitorWait` and `jdk.JavaMonitorEnter` events being emitted, but secondary `jdk.ThreadPark` events being emitted as well. The `jdk.ThreadPark` events are only emitted due to the implementation details of the [`com.oracle.svm.core.monitor.JavaMonitor`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/JavaMonitorQueuedSynchronizer.java#L397) and [`com.oracle.svm.core.monitor.JavaMonitorQueuedSynchronizer`](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/monitor/JavaMonitorQueuedSynchronizer.java#L139C17-L139C33) classes. 

 In Hotspot, the Java `Unsafe` class is not needed to block threads so `jdk.JavaMonitorWait` and `jdk.JavaMonitorEnter` events are emitted without corresponding `jdk.ThreadPark` events.

We should reduce noise and copy Hotspot by skipping the emission of these internal  `jdk.ThreadPark` events.

**Hotspot**
image
image
Notice that there is no `jdk.ThreadPark` corresponding to the  `jdk.JavaMonitorEnter`  event.

**Native Image before the fix**
image
image
Notice that there is a `jdk.ThreadPark` corresponding to the  `jdk.JavaMonitorEnter`  event

**Native Image after the fix**
image
image

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/ThreadParkEvent.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/events/ThreadParkEvent.java
@@ -37,10 +37,21 @@ import com.oracle.svm.core.jfr.JfrNativeEventWriterData;
 import com.oracle.svm.core.jfr.JfrNativeEventWriterDataAccess;
 import com.oracle.svm.core.jfr.JfrTicks;
 import com.oracle.svm.core.jfr.SubstrateJVM;
+import com.oracle.svm.core.monitor.JavaMonitor;
 
 public class ThreadParkEvent {
     public static void emit(long startTicks, Object obj, boolean isAbsolute, long time) {
         if (HasJfrSupport.get()) {
+            /*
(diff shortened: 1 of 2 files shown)


roberttoyonaga:
@christianhaeubl  Can you please have a look at this when you have time?

selhagani:
Hi @roberttoyonaga, 

Thank you for your kind contribution, we appreciate it!

christianhaeubl (approved):

christianhaeubl:
Thanks, will probably be integrated later this week.

PR #12005, opened 2025-08-19
[GR-68873] Use correct Python version in GitHub style gate

graalvmbot:


DSouzaM:
Manually mirrored description:

> This fixes the Truffle style gate that has been failing for some time.
> Using apt to install pip was installing pip and the style packages (like pylint) to run on system Python, which was not the same as MX_PYTHON (3.8). By installing the packages using MX_PYTHON we get the intended behaviour.

PR #12068, opened 2025-08-28
Minor cleanups in Truffle code

blaumeise20:
I improved a few small things I noticed while working on Truffle.

@gilles-duboscq


oubidar-Abderrahim:
Hi, Thank you for your contribution. @gilles-duboscq could you please review or assign someone to do so, thanks

truffle/src/com.oracle.truffle.runtime/src/com/oracle/truffle/runtime/OptimizedCallTarget.java
```diff
@@ -829,7 +829,7 @@ private boolean frameDescriptorEquals(FrameDescriptor parentFrameDescriptor) {
         return false;
     }
 
-    private Object executeRootNode(VirtualFrame frame, CompilationState tier) {
+    private final Object executeRootNode(VirtualFrame frame, CompilationState tier) {
```
chumer, line 832: there is no need to make private methods final.please remove.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/phases/inlining/GraphManager.java
```diff
@@ -64,30 +64,29 @@ final class GraphManager {
...
                 try (Scope hostScope = context.debug.scope("CreateGraph", context.graph);
                                 Indent indent = context.debug.logAndIndent("evaluate %s", context.graph);) {
-                    partialEvaluator.doGraphPE(context, plugin, graphCacheForInlining);
-                    context.graph.getAssumptions().record(new TruffleAssumption(context.getNodeRewritingAssumption(partialEvaluator.getProviders())));
-                    StructuredGraph graphAfterPE = copyGraphForDebugDump(context);
-                    postPartialEvaluationSuite.apply(context.graph, context);
-                    entry = new Entry(context.graph, plugin, graphAfterPE, useSize ? NodeCostUtil.computeGraphSize(context.graph) : -1);
-                    context.debug.dump(DebugContext.INFO_LEVEL, context.graph, "After PE Tier");
+                    TruffleAssumption nodeRewritingAssumption = new TruffleAssumption(context.getNodeRewritingAssumption(partialEvaluator.getProviders()));
+                    entry = pe(context, nodeRewritingAssumption, DebugContext.INFO_LEVEL);
```
jchalou, line 89: Extracting common method cannot be done in this way because there are very strict rules on the usage of the DebugContext#dump method. For example, it is checked that the first argument (dumpLevel) is a constant, not a variable. So we would have to work around this somehow and the code wouldn't look all that nice. So we decided to drop the relevant commit. The rest of this PR will be merged through our internal CI.

jchalou: Actually, we decided to lift the restrictions in this case, so this PR will be merged as is (with some minor adjustments on top of it).

Issue #12236, opened 2025-09-25
Project Terminus: Self-hosting Native Image


dougxc:
## TL;DR

Terminus will make it possible for Native Image to be self-hosted. That is, Native Image will be able to produce an image of itself such that it runs as a native executable without any need for an underlying JVM such as HotSpot.

This will be achieved by using separate heaps at build time.
* The host heap will be used for host types and data structures.
* An Espresso context with its own heap will be used for guest types and data structures.

This avoids complications due to state being mixed between host classes and guest classes.

This is a work in progress but aspects of the current design are outlined here:
* Load both user and SubstrateVM runtime classes into an Espresso context. JDK classes are then loaded twice, once in host and once in Espresso. Each has its own copy of static fields.
* Perform closed world analysis exclusively on Espresso loaded types.
* No more ability to convert an arbitrary `AnalysisType` to a `java.lang.Class` instance.
* Run user features within the Espresso context. Provide necessary data structures and access to the host system for external features via Espresso interoperability.
* TBD: Audit internal features for "guest" vs "host" build time code and run it in the right context.
* Change host code to use JVMCI instead of core reflection when accessing Espresso context types (e.g., `Class`, `Method`, `Field` are replaced with `ResolvedJavaType`, `ResolvedJavaMethod`, `ResolvedJavaField`).

"Terminus" is the Roman god of borders. This project is about drawing a much stronger "border" between the native image generator and the application it's building.

## Goals
* Build Native Image as a native executable that does not need a host JVM.

Paullo612:
Sounds cool.

Does this mean that it eventually will be possible то load `lib/modules` from Windows build to Espresso on Linux and do a cross-compilation? Can't wait to stop my Windows CI runner!

dougxc:
I don't think this will enable cross-compilation as native image still needs to use the platform linker for linking in the platform-specific native libs.

Paullo612:
Looks like we can use `clang-cl` and `lld-link` for this. Chromium is already able to cross-compile itself from Linux to Windows. See documentation [here](https://chromium.googlesource.com/chromium/src/+/lkgr/docs/win_cross.md). [Here](https://github.com/mstorsjo/msvc-wine?tab=readme-ov-file#use-with-clanglld-in-msvc-mode) is another related project.

dougxc:
Ok, sounds promising. We will find out what's possible as the project progresses.

zakkak:
[...]

PR #12245, opened 2025-09-26
Substitution fails during NI build when `java.prefs` module is not present

Akirathan:
In [enso](https://github.com/enso-org/enso), we are building NI from a small JDK created by `jlink`. On latest [GraalVM CE 25.0.0](https://github.com/graalvm/graalvm-ce-builds/releases/tag/jdk-25.0.0), it is no longer possible to build NI from small JDK without `java.prefs` module on MacOS.

## Minimal reproducer
Makefile:
```make
NI_BUILDER_MODULES = "org.graalvm.nativeimage.builder,org.graalvm.nativeimage.driver,org.graalvm.nativeimage.librarysupport,org.graalvm.nativeimage.objectfile,org.graalvm.nativeimage.pointsto"
NI_BASE_MODULES = "org.graalvm.nativeimage,org.graalvm.nativeimage.base,com.oracle.graal.graal_enterprise,com.oracle.svm.svm_enterprise"
JDK_MODULES = "java.naming,java.net.http,java.rmi,jdk.attach,jdk.charsets,jdk.crypto.ec,jdk.httpserver,jdk.localedata,jdk.jdwp.agent,jdk.security.auth"
ADD_MODULES = "$(NI_BUILDER_MODULES),$(NI_BASE_MODULES),$(JDK_MODULES)"

MODULE_JARS = "$$JAVA_HOME/lib/svm/bin/../../graalvm/svm-driver.jar:$$JAVA_HOME/lib/svm/bin/../builder/native-image-base.jar:$$JAVA_HOME/lib/svm/bin/../builder/espresso-svm.jar:$$JAVA_HOME/lib/svm/bin/../builder/objectfile.jar:$$JAVA_HOME/lib/svm/bin/../builder/pointsto.jar:$$JAVA_HOME/lib/svm/bin/../builder/svm-enterprise.jar:$$JAVA_HOME/lib/svm/bin/../builder/svm.jar:$$JAVA_HOME/lib/svm/bin/../builder/svm-configure.jar:$$JAVA_HOME/lib/svm/bin/../builder/svm-capnproto-runtime.jar:$$JAVA_HOME/lib/svm/bin/../builder/svm-foreign.jar:$$JAVA_HOME/lib/svm/bin/../library-support.jar"

NI_EXE = small_jdk/lib/svm/bin/native-image

all: clean native_image

small_jdk:
	jlink --output small_jdk --module-path "$(MODULE_JARS)" --add-modules "$(ADD_MODULES)"
	cp -r $$JAVA_HOME/lib/graalvm $$JAVA_HOME/lib/svm $$JAVA_HOME/lib/static $$JAVA_HOME/lib/truffle small_jdk/lib

Main.class: Main.java
	javac Main.java

native_image: Main.class small_jdk
	$(NI_EXE) Main -o main

clean:
	rm -rf small_jdk
	rm -f Main.class
	rm -f main
```

Main.java:
```java
public class Main {
        System.out.println("Hello from Main");
    }
}
```

Running `make` on GraalVM CE 25.0.0, the following error is reported:
```
GraalVM Native Image: Generating 'main' (executable)...
[1/8] Initializing...
                                                                                    (0.0s @ 0.08GB)
Error: Substitution target for com.oracle.svm.core.posix.darwin.Target_java_util_prefs_FileSystemPreferences is not loaded. Use field `onlyWith` in the `TargetClass` annotation to make substitution only active when needed.
```

Running `make` on GraalVM CE 24.0.1 is OK.

Note that @JaroslavTulach was not possible to reproduce this error on Linux AMD64. Seems to affect only Mac (I have aarch64, but it probably also affects x64).

## Context
- Encountered in https://github.com/enso-org/enso/pull/14019#discussion_r2381732260
- The process how to create small JDK was recommended to us by @fniephaus in https://graalvm.slack.com/archives/CN9KSFB40/p1715117172381849?thread_ts=1715083564.880179&cid=CN9KSFB40
- Tracked as an issue in https://github.com/enso-org/enso/issues/14051

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/darwin/DarwinSubstitutions.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/darwin/DarwinSubstitutions.java
@@ -28,6 +28,7 @@ import static com.oracle.svm.core.posix.headers.darwin.DarwinTime.NoTransitions.
 import static com.oracle.svm.core.posix.headers.darwin.DarwinTime.NoTransitions.mach_timebase_info;
 
 import java.util.Objects;
+import java.util.function.BooleanSupplier;
 
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.Platform;
@@ -122,7 +123,7 @@ final class DarwinTimeUtil {
  * Native functions don't exist on Darwin because this whole class is used, and should exist, only
  * on Linux. See <code>java.util.prefs.Preferences#factory</code>.
  */
-@TargetClass(className = "java.util.prefs.FileSystemPreferences")
+@TargetClass(className = "java.util.prefs.FileSystemPreferences", onlyWith = IsJavaUtilPrefsPresent.class)
 final class Target_java_util_prefs_FileSystemPreferences {
     @Delete
     private static native int[] lockFile0(String fileName, int permission, boolean shared);
@@ -132,6 +133,14 @@ final class Target_java_util_prefs_FileSystemPreferences {
 
     @Delete
     private static native int chmod(String fileName, int permission);
+}
+
+final class IsJavaUtilPrefsPresent implements BooleanSupplier {
+    @Override
+    public boolean getAsBoolean() {
+        var prefsMod = ModuleLayer.boot().findModule("java.prefs");
+        return prefsMod.isPresent();
+    }
 }
 
 /**


Akirathan:
Because of this issue, we had to add `java.prefs` module to small JDK used for NI builds as a workaround. See https://github.com/enso-org/enso/pull/14019#discussion_r2375203210. I have confirmed that with this PR applied, `java.prefs` can be removed from the small JDK.

JaroslavTulach:
> * PR author: [pavel.marek@enso.org](mailto:pavel.marek@enso.org) (@Akirathan)

- Pavel is making this contribution on behalf of Enso
- the same company on behalf of I am making my contributions
- looking at my mailbox I see Enso signed the company agreement in ~2022
- such a company wide OCA agreement shall cover all Enso employees
- as such this contribution by Pavel shall be OCA covered as well

JaroslavTulach:
- The original code in question was integrated by @vjovanov as part of #11106
- Vojin, what do you think about these changes proposed by Pavel?

vjovanov:
Looks good, thanks for fixing this issue. 

I am wondering if we should do this systematically for all JDK substitutions. For example, we could ask the JDK substitutions to be targeted with a module. This would require us to change a lot of substitutions, but it would solve this problem for all other `jlink` related issues. @zapster WDYT?

zapster:
> I am wondering if we should do this systematically for all JDK substitutions. For example, we could ask the JDK substitutions to be targeted with a module.

If running on a custom `jlink`ed JDK is something we want to actively support, then it sounds reasonable. However, I'm not sure how much that aligns with the goals of Project Terminus https://github.com/oracle/graal/issues/12236. Maybe to go one step back, what issue are you solving with the "small jdk"?

> This would require us to change a lot of substitutions, but it would solve this problem for all other `jlink` related issues. @zapster WDYT?

I believe it is possible to implement this via a special case in the `AnnotationSubstitutionProcessor`. We would need to have a map from classes to jdk modules. We would need create it at/before native-image build time and bake into the `native-image` executable. I think list could be created automatically, e.g., via a `javac` annotation processor, assuming that the `native-image` executable is built with a "full jdk", which is the case in the reproducer outlined above. Then in `AnnotationSubstitutionProcessor#findTargetClass` we simply ask the map and check if the module is present, right after we evaluated the `onlyWith` class. That way, we would not need to change any substitution.

That said, I'd like to learn more about the motivation for jlinking. There might be other issues since the `native-image` executable does not expect that the base modules differ at native-image build and native-image run time.

JaroslavTulach:
Hello Josef, thanks for considering Enso use-case...

> what issue are you solving with the "small jdk"?

Enso is server side application. We don't want `java.desktop` & co. in our GraalVM. Neither HotSpot nor _native image_ mode (we use both). We are using hint from Fabio to built `jlink` version of GraalVM...

> * The process how to create small JDK was recommended to us by @fniephaus in https://graalvm.slack.com/archives/CN9KSFB40/p1715117172381849?thread_ts=1715083564.880179&cid=CN9KSFB40

...yes, it would be great to let `jlink` be supported by GraalVM more "officially".

vjovanov:
OK, let us merge this PR for now until we know the shape of GraalVM 25.1 better. Then we can re-iterate.

I'll put this to merge and backport to 25.0.1.

JaroslavTulach:
> We would need to have a map from classes to jdk modules.

Btw. [Element.getModuleOf](https://docs.oracle.com/en/java/javase/25/docs/api/java.compiler/javax/lang/model/util/Elements.html#getModuleOf(javax.lang.model.element.Element)) maybe the method you are looking for, @zapster

PR #12263, opened 2025-09-30
Extend wrapped JS function coverage in core runtime types

Atur256:
This pull request expands the set of host-accessible wrappers for internal JavaScript functions across key runtime types: `JSObject`, `JSNumber`, `JSString`, `JSSymbol`, and `JSValue`. These additions improve Java interop and simplify invocation of JavaScript behavior from host code.

Additionally, the class `JSError` has been renamed to `ThrownFromJavaScript` to better reflect its actual role. The original name implied alignment with the native JavaScript `Error` object, but the class more accurately represents any object thrown from JavaScript, regardless of its type.

### DIFF
--- a/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
+++ b/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
@@ -119,4 +119,137 @@ public final class JSNumber extends JSValue {
     public int hashCode() {
         return javaDouble().hashCode();
     }
+
+    @JS.Coerce
+    @JS(value = "return isFinite(number);")
+    public native static boolean isFinite(JSNumber number);
+
+    @JS.Coerce
+    @JS(value = "return isFinite(number);")
+    public native static boolean isFinite(Number number);
+
+    @JS.Coerce
+    @JS(value = "return Number.isInteger(number);")
+    public native static boolean isInteger(JSNumber number);
+
+    @JS.Coerce
+    @JS(value = "return Number.isInteger(number);")
+    public native static boolean isInteger(Number number);
+
+    @JS.Coerce
+    @JS(value = "return isNaN(number);")
+    public native static boolean isNaN(JSValue number);
+
+    @JS.Coerce
+    @JS(value = "return isNaN(number);")
+    public native static boolean isNaN(Number number);
+
+    @JS.Coerce
+    @JS(value = "return Number.isSafeInteger(number);")
+    public native static boolean isSafeInteger(JSValue number);
+
+    @JS.Coerce
+    @JS(value = "return Number.isSafeInteger(number);")
+    public native static boolean isSafeInteger(Number number);
+
+    @JS.Coerce
+    @JS(value = "return parseFloat(number);")
+    public native static float parseFloat(Number number);
+
+    @JS.Coerce
+    @JS(value = "return parseFloat(number);")
+    public native static float parseFloat(String number);
+
+    @JS.Coerce
+    @JS(value = "return parseInt(number);")
+    public native static int parseInt(Number number);
+
+    @JS.Coerce
+    @JS(value = "return parseInt(number);")
+    public native static int parseInt(String number);
+
+    @JS.Coerce
+    @JS(value = "return parseInt(number, radix);")
+    public native static int parseInt(String number, int radix);
+
+    @JS.Coerce
+    @JS(value = "return Number.EPSILON;")
+    public native static double EPSILON();
+
+    @JS.Coerce
+    @JS(value = "return Number.MAX_SAFE_INTEGER;")
+    public native static double MAX_SAFE_INTEGER();
+
+    @JS.Coerce
+    @JS(value = "return Number.MAX_VALUE;")
+    public native static double MAX_VALUE();
+
+    @JS.Coerce
+    @JS(value = "return Number.MIN_SAFE_INTEGER;")
+    public native static double MIN_SAFE_INTEGER();
+
+    @JS.Coerce
+    @JS(value = "return Number.MIN_VALUE;")
+    public native static double MIN_VALUE();
+
+    @JS.Coerce
+    @JS(value = "return Number.NaN;")
+    public native static double NaN();
+
+    @JS.Coerce
+    @JS(value = "return Number.NEGATIVE_INFINITY;")
+    public native static double NEGATIVE_INFINITY();
+
+    @JS.Coerce
+    @JS(value = "return Number.POSITIVE_INFINITY;")
+    public native static double POSITIVE_INFINITY();
+
+    @JS.Coerce
+    @JS(value = "return this.toExponential();")
+    public native String toExponential();
+
+    @JS.Coerce
+    @JS(value = "return this.toExponential(fractionDigits);")
+    public native String toExponential(int fractionDigits);
+
+    @JS.Coerce
+    @JS(value = "return this.toFixed();")
+    public native String toFixed();
+
+    @JS.Coerce
+    @JS(value = "return this.toFixed(digits);")
+    public native String toFixed(int digits);
+
+    @JS.Coerce
+    @JS(value = "return this.toLocaleString();")
+    public native String toLocaleString();
+
+    @JS.Coerce
+    @JS(value = "return this.toLocaleString(locales);")
+    public native String toLocaleString(String locales);
+
+    @JS.Coerce
+    @JS(value = "return this.toLocaleString(locales, options);")
+    public native String toLocaleString(String locales, JSObject options);
+
+    @JS.Coerce
+    @JS(value = "return this.toPrecision();")
+    public native String toPrecision();
+
+    @JS.Coerce
+    @JS(value = "return this.toPrecision(precision);")
+    public native String toPrecision(int precision);
+
+    @Override
+    @JS.Coerce
+    @JS(value = "return this.toString();")
+    public native String toString();
+
+    @JS.Coerce
+    @JS(value = "return this.toString(radix);")
+    public native String toString(int radix);
+
+    @JS.Coerce
+    @JS(value = "return this.valueOf();")
+    public native double valueOf();
 }
--- a/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
+++ b/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
@@ -49,14 +49,14 @@ package org.graalvm.webimage.api;
  * wrapped into a <code>JSObject</code> instance. The <code>JSObject</code> allows the Java code to
  * access the fields of the underlying JavaScript object using the <code>get</code> and
  * <code>set</code> methods.
- *
+ * <p>
  * The Java {@link JSObject} instance that corresponds to the JavaScript object is called a <i>Java
  * mirror</i>. Vice versa, the JavaScript instance is a <i>JavaScript mirror</i> for that
  * {@link JSObject} instance.
  *
  *
  * <h3>Anonymous JavaScript objects</h3>
- *
+ * <p>
  * Here are a few examples of creating and modifying anonymous JavaScript objects:
  *
  * <pre>
@@ -68,7 +68,7 @@ package org.graalvm.webimage.api;
  * System.out.println(pair.get("x"));
  * System.out.println(pair.get("y"));
  * </pre>
- *
+ * <p>
  * In this example, using the {@code JSObject} methods, the user can access the <code>x</code> and
  * <code>y</code> fields.
  *
@@ -76,32 +76,32 @@ package org.graalvm.webimage.api;
  * pair.set("x", 5.4);
  * System.out.println(pair.get("x"));
  * </pre>
- *
+ * <p>
  * The code above sets a new value for the <code>x</code> field, and then prints the new value.
  *
  *
  * <h3>Anonymous JavaScript functions</h3>
- *
+ * <p>
  * A {@code JSObject} can be a Java-side wrapper for a JavaScript {@code Function} object. The
  * JavaScript {@code Function} value can be returned by the JavaScript code of the method annotated
  * with the {@link JS} annotation.
- *
+ * <p>
  * The Java program can then call the underlying function by calling the
  * {@link JSObject#call(Object, Object...)} method. If the underlying JavaScript object is not
  * callable, then calling {@code call} will result in an exception.
- *
+ * <p>
  * The {@code call} method has the following signature:
  *
  * <pre>
  * public Object call(Object thisArgument, Object... arguments);
  * </pre>
- *
+ * <p>
  * The {@code invoke} method has the following signature:
  *
  * <pre>
  * public Object invoke(Object... arguments);
  * </pre>
- *
+ * <p>
  * The difference is that the method {@code call} takes an object for specifying {@code this} of the
  * JavaScript function, while {@code invoke} uses the underlying {@code JSObject} as the value of
  * {@code this}.
@@ -130,13 +130,13 @@ package org.graalvm.webimage.api;
  *
  *
  * <h3>Declaring JavaScript classes in Java</h3>
- *
+ * <p>
  * The second purpose of {@link JSObject} is to declare JavaScript classes as classes in Java code,
  * in a way that makes the Java objects look-and-feel like native JavaScript objects. Users should
  * subclass the {@link JSObject} class when they need to define a JavaScript class whose fields and
  * methods can be accessed conveniently from Java, without the {@link JSObject#get(Object)} and
  * {@link JSObject#set(Object, Object)} methods.
- *
+ * <p>
  * Directly exposing a Java object to JavaScript code means that the JavaScript code is able to
  * manipulate the data within the object (e.g. mutate fields, add new fields, or redefine existing
  * fields), which is not allowed by default for regular Java classes. Extending {@link JSObject}
@@ -144,7 +144,7 @@ package org.graalvm.webimage.api;
  * One of the use-cases for these functionalities are JavaScript frameworks that redefine properties
  * of JavaScript objects with custom getters and setters, with the goal of enabling data-binding or
  * reactive updates.
- *
+ * <p>
  * In a subclass of {@link JSObject}, every JavaScript property directly corresponds to the Java
  * field of the same name. Consequently, all these properties point to native JavaScript values
  * rather than Java values, so bridge methods are generated that are called for each property access
@@ -154,7 +154,7 @@ package org.graalvm.webimage.api;
  * corresponding Java field. For this reason, the bridge methods also generate check-casts on every
  * access: if the JavaScript property that corresponds to the Java field does not contain a
  * compatible value, a {@link ClassCastException} is thrown.
- *
+ * <p>
  * There are several restrictions imposed on {@link JSObject} subclasses:
  * <ul>
  * <li>Only public and protected fields are allowed to ensure encapsulation.</li>
@@ -179,7 +179,7 @@ package org.graalvm.webimage.api;
  *     }
  * }
  * </pre>
- *
+ * <p>
  * The preceding Java class is effectively translated to the corresponding JavaScript class:
  *
  * <pre>
@@ -194,7 +194,7 @@ package org.graalvm.webimage.api;
  *     }
  * }
  * </pre>
- *
+ * <p>
  * The {@code Point} class can be used from Java as if it were a normal Java class:
  *
  * <pre>
@@ -203,7 +203,7 @@ package org.graalvm.webimage.api;
  * System.out.println(p.y);
  * System.out.println(p.absolute());
  * </pre>
- *
+ * <p>
  * All accesses to the fields {@code x} and {@code y} are rewritten to accesses on the corresponding
  * JavaScript properties. JavaScript code may assign values to these properties that violate the
  * type of the corresponding Java fields, but a subsequent Java read of such a field will result in
@@ -219,7 +219,7 @@ package org.graalvm.webimage.api;
  *
  *
  * <h3>Passing {@code JSObject} subclasses between JavaScript and Java</h3>
- *
+ * <p>
  * When an object of the {@link JSObject} subclass is passed from Java to JavaScript code using the
  * {@link JS} annotation, the object is converted from its Java representation to its JavaScript
  * representation. Changes in the JavaScript representation are reflected in the Java representation
@@ -236,7 +236,7 @@ package org.graalvm.webimage.api;
  * reset(p0, 0.0, 0.0);
  * System.out.println(p0.x + ", " + p0.y);
  * </pre>
- *
+ * <p>
  * A {@link Class} object that represents {@link JSObject} can also be passed to JavaScript code.
  * The {@link Class} object is wrapped in a proxy, which can be used inside a {@code new} expression
  * to instantiate the object of the corresponding class from JavaScript.
@@ -250,14 +250,14 @@ package org.graalvm.webimage.api;
  * Point p1 = create(Point.class, 1.25, 0.5);
  * System.out.println(p1.x + ", " + p1.y);
  * </pre>
- *
+ * <p>
  * Note that creating an object in JavaScript and passing it to Java several times does not
  * guarantee that the same mirror instance is returned each time -- each time a JavaScript object
  * becomes a Java mirror, a different instance of the mirror may be returned.
  *
  *
  * <h3>Importing existing JavaScript classes</h3>
- *
+ * <p>
  * The {@link JSObject} class allows access to properties of any JavaScript object using the
  * {@link JSObject#get(Object)} and {@link JSObject#set(Object, Object)} methods. In situations
  * where the programmer knows the relevant properties of a JavaScript object (for example, when
@@ -265,14 +265,14 @@ package org.graalvm.webimage.api;
  * "imported" to Java. To do this, the user declares a {@link JSObject} subclass that serves as a
  * facade to the JavaScript object. This subclass must be annotated with the {@link JS.Import}
  * annotation.
- *
+ * <p>
  * The name of the declared class Java must match the name of the JavaScript class that is being
  * imported. The package name of the Java class is not taken into account -- the same JavaScript
  * class can be imported multiple times from within separate packages.
- *
+ * <p>
  * The exposed JavaScript fields must be declared as {@code protected} or {@code public}. The
  * constructor parameters must match those of the JavaScript class, and its body must be empty.
- *
+ * <p>
  * Here is an example of a class declared in JavaScript:
  *
  * <pre>
@@ -283,7 +283,7 @@ package org.graalvm.webimage.api;
  *     }
  * }
  * </pre>
- *
+ * <p>
  * To import this class in Java code, we need the following declaration in Java:
  *
  * <pre>
@@ -296,19 +296,19 @@ package org.graalvm.webimage.api;
  *     }
  * }
  * </pre>
- *
+ * <p>
  * The fields declared in the {@code Rectangle} class are directly mapped to the properties of the
  * underlying JavaScript object. If the type of the property of the underlying JavaScript object
  * does not match the type of the field declared in Java, then a field-read in Java will throw a
  * {@link ClassCastException}.
- *
+ * <p>
  * The {@code Rectangle} class can be instantiated from Java as follows:
  *
  * <pre>
  * Rectangle r = new Rectangle(640, 480);
  * System.out.println(r.width + "x" + r.height);
  * </pre>
- *
+ * <p>
  * A JavaScript object whose {@code constructor} property matches the imported JavaScript class can
  * be converted to the declared Java class when the JavaScript code passes a value to Java. Here is
  * a code example that creates the {@code Rectangle} object in JavaScript, and passes it to Java:
@@ -317,13 +317,13 @@ package org.graalvm.webimage.api;
  * &#64;JS("return new Rectangle(width, height);")
  * Rectangle createRectangle(int width, int height);
  * </pre>
- *
+ * <p>
  * Another way to convert a JavaScript object to a Java facade is to call the
  * {@link JSObject#as(Class)} method to cast the {@link JSObject} instance to the proper subtype.
  *
  *
  * <h3>Exporting Java classes to JavaScript</h3>
- *
+ * <p>
  * The users can annotate the exported classes with the {@link JS.Export} annotation to denote that
  * the {@link JSObject} subclass should be made available to JavaScript code. Exported classes can
  * be accessed using the JavaScript VM-instance API, using the `exports` property.
@@ -344,7 +344,7 @@ package org.graalvm.webimage.api;
  *     }
  * }
  * </pre>
- *
+ * <p>
  * The exported class can then be used from JavaScript code as follows:
  *
  * <pre>
@@ -395,7 +395,7 @@ public class JSObject extends JSValue {
     /**
      * Sets the value of the key passed as the argument in the JavaScript object.
      *
-     * @param key the object under which the value should be placed in the JavaScript object
+     * @param key      the object under which the value should be placed in the JavaScript object
      * @param newValue the value that should be placed under the given key in the JavaScript object
      */
     @JS("this[key] = newValue;")
@@ -413,7 +413,7 @@ public class JSObject extends JSValue {
      * Invoke the underlying JavaScript function, if this object is callable.
      *
      * @param args The array of Java arguments, which is converted to JavaScript and passed to the
-     *            underlying JavaScript function
+     *             underlying JavaScript function
      * @return The result of the JavaScript function, converted to the corresponding Java value
      */
     @JS("return this.apply(this, conversion.extractJavaScriptArray(args[runtime.symbol.javaNative]));")
@@ -424,8 +424,8 @@ public class JSObject extends JSValue {
      * in the function, if this object is callable.
      *
      * @param thisArg The value for the binding of {@code this} inside the JavaScript function
-     * @param args The array of Java arguments, which is converted to JavaScript and passed to the
-     *            underlying JavaScript function
+     * @param args    The array of Java arguments, which is converted to JavaScript and passed to the
+     *                underlying JavaScript function
      * @return The result of the JavaScript function, converted to the corresponding Java value
      */
     @JS("return this.apply(thisArg, conversion.extractJavaScriptArray(args[runtime.symbol.javaNative]));")
@@ -547,4 +547,108 @@ public class JSObject extends JSValue {
     public boolean equalsJavaScript(JSObject that) {
         return referenceEquals(this, that).asBoolean();
     }
+
+    @JS.Coerce
+    @JS(value = "return Object.create(proto);")
+    public static native JSObject create(JSObject proto);
+
+    @JS.Coerce
+    @JS(value = "return Object.create(proto, properties);")
+    public static native JSObject create(JSObject proto, JSObject properties);
+
+    @JS.Coerce
+    @JS(value = "return Object.defineProperties(obj, prop);")
+    public static native JSObject defineProperties(JSObject obj, JSObject prop);
+
+    @JS.Coerce
+    @JS(value = "return Object.defineProperty(obj, prop, descriptor);")
+    public static native JSObject defineProperty(JSObject obj, JSString prop, JSObject descriptor);
+
+    @JS.Coerce
+    @JS(value = "return Object.defineProperty(obj, prop, descriptor);")
+    public static native JSObject defineProperty(JSObject obj, String prop, JSObject descriptor);
+
+    @JS.Coerce
+    @JS(value = "return Object.entries(obj);")
+    public static native JSObject entries(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "Object.freeze(obj);")
+    public static native void freeze(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return Object.fromEntries(iterable);")
+    public static native JSObject fromEntries(JSObject iterable);
+
+    @JS.Coerce
+    @JS(value = "return Object.getOwnPropertyDescriptor(obj, prop);")
+    public static native JSObject getOwnPropertyDescriptor(JSObject obj, String prop);
+
+    @JS.Coerce
+    @JS(value = "return Object.getOwnPropertyNames(obj);")
+    public static native JSObject getOwnPropertyNames(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return Object.groupBy(items, callback);")
+    public native static JSObject groupBy(JSObject items, JSValue callback);
+
+    @JS.Coerce
+    @JS(value = "return Object.hasOwn(obj, prop);")
+    public static native boolean hasOwn(JSObject obj, String prop);
+
+    @JS.Coerce
+    @JS(value = "return Object.is(value1, value2);")
+    public static native boolean is(JSValue value1, JSValue value2);
+
+    @JS.Coerce
+    @JS(value = "return Object.isExtensible(obj);")
+    public static native boolean isExtensible(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return Object.isFrozen(obj);")
+    public static native boolean isFrozen(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return Object.isSealed(obj);")
+    public static native boolean isSealed(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return Object.keys(obj);")
+    public static native JSObject keys(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return Object.preventExtensions(obj);")
+    public static native JSObject preventExtensions(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return Object.seal(obj);")
+    public static native JSObject seal(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return Object.setPrototypeOf(obj, proto);")
+    public static native JSObject setPrototypeOf(JSObject obj, JSObject proto);
+
+    @JS.Coerce
+    @JS(value = "return Object.values(obj);")
+    public static native JSObject values(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return this.hasOwnProperty(prop);")
+    public native boolean hasOwnProperty(String prop);
+
+    @JS.Coerce
+    @JS(value = "return this.isPrototypeOf(object);")
+    public native boolean isPrototypeOf(JSObject object);
+
+    @JS.Coerce
+    @JS(value = "return this.propertyIsEnumerable(prop);")
+    public native boolean propertyIsEnumerable(String prop);
+
+    @JS.Coerce
+    @JS(value = "return this.toLocaleString();")
+    public native String toLocaleString();
+
+    @JS.Coerce
+    @JS(value = "return this.valueOf();")
+    public native JSValue valueOf();
 }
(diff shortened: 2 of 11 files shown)


sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
```diff
@@ -119,4 +119,137 @@ public boolean equals(Object that) {
...
         return javaDouble().hashCode();
     }
+
+    @JS.Coerce
+    @JS(value = "return isFinite(number);")
+    public native static boolean isFinite(JSNumber number);
+
+    @JS.Coerce
+    @JS(value = "return isFinite(number);")
+    public native static boolean isFinite(Number number);
```
axel22: Technically, not every `Number` data type is autocoerced to `JSNumber` (i.e. JavaScript `Number`). There are many `Number` subtypes in Java for which a proxy will be created on the JavaScript side.

I guess `isFinite` returns `true` in all non-Number cases in JavaScript (?), but is this the intent?
Maybe a `double number` is what we want in the signature here.

axel22: Similar comment for `isInteger`, `isNaN`, `isSafeInteger`, `parseInt` and `parseFloat`.

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,304 @@ public boolean equals(Object that) {
     public int hashCode() {
         return javaString().hashCode();
     }
+
+    @JS.Coerce
+    @JS(value = "return String.fromCharCode();")
+    public static native JSString fromCharCode();
```
axel22: While it technically works without arguments, I think that a single-argument overload (plus the varargs version you have later) is what would be best here. Same for `fromCodePoint`.

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,304 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return String.fromCharCode(...codeUnits);")
```
axel22, line 87: Was conversion to a JS array not necessary here?

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSSymbol.java
```diff
@@ -87,4 +87,83 @@ public boolean equals(Object that) {
     public int hashCode() {
         return javaString().hashCode();
     }
+
+    @JS(value = "return Symbol.for(key);")
+    public static native JSSymbol forKey(String key);
+
+    @JS.Coerce
+    @JS(value = "return Symbol.keyFor(sym);")
+    public static native JSValue keyFor(JSSymbol sym);
```
axel22: Does this return a `JSString` always, so it could return a `String` here?

Atur256: It could also return `undefined`, which is why `JSValue` is currently used here.

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
Atur256: Conversion wasn’t necessary in this case because primitive arrays are apparently coerced automatically. At least the int array is coerced to a `Int32Array` in JavaScript in this function.

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/testdispatcher/JSAnnotationTests.java
```diff
@@ -61,6 +65,14 @@ public static void main(String[] args) {
...
         } else if (checkClass(HtmlApiExamplesTest.class, className)) {
             HtmlApiExamplesTest.main(remainingArgs);
+        } else if (checkClass(JSNumberTest.class, className)) {
+            JSNumberTest.main(null);
+        } else if (checkClass(JSStringTest.class, className)) {
+            JSStringTest.main(null);
+        } else if (checkClass(JSSymbolTest.class, className)) {
+            JSSymbolTest.main(null);
+        } else if (checkClass(JSObjectTest.class, className)) {
+            JSObjectTest.main(null);
```
patrick96: Pass `remainingArgs` for all these tests, even if the arguments are not used right now. Passing `null` to the `main` method is really unexpected

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
```diff
@@ -119,4 +119,136 @@ public boolean equals(Object that) {
...
+    @JS(value = "return Number.isInteger(number);")
+    public native static boolean isInteger(JSNumber number);
+
+    @JS.Coerce
+    @JS(value = "return Number.isInteger(number);")
+    public native static boolean isInteger(double number);
+
+    @JS.Coerce
+    @JS(value = "return isNaN(number);")
+    public native static boolean isNaN(JSValue number);
```
patrick96: What's the rationale for using `JSValue` here and nit `JSNumber`?

Same question for `isSafeInteger`

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
```diff
@@ -119,4 +119,136 @@ public boolean equals(Object that) {
...
+    @JS(value = "return Number.isSafeInteger(number);")
+    public native static boolean isSafeInteger(double number);
+
+    @JS.Coerce
+    @JS(value = "return parseFloat(number);")
+    public native static float parseFloat(double number);
+
+    @JS.Coerce
+    @JS(value = "return parseFloat(number);")
+    public native static float parseFloat(String number);
```
patrick96: `parseFloat` returns JS number, so we should use `double` here.

Also, we should probably prefer `Number.parseFloat` (even though they're the same function) just for consistency.

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
```diff
@@ -119,4 +119,136 @@ public boolean equals(Object that) {
...
+    @JS(value = "return Number.isSafeInteger(number);")
+    public native static boolean isSafeInteger(JSValue number);
+
+    @JS.Coerce
+    @JS(value = "return Number.isSafeInteger(number);")
+    public native static boolean isSafeInteger(double number);
+
+    @JS.Coerce
+    @JS(value = "return parseFloat(number);")
+    public native static float parseFloat(double number);
```
patrick96: I don't think we need this method

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
```diff
@@ -119,4 +119,136 @@ public boolean equals(Object that) {
...
+    @JS.Coerce
+    @JS(value = "return parseInt(number);")
+    public native static int parseInt(double number);
+
+    @JS.Coerce
+    @JS(value = "return parseInt(number);")
+    public native static int parseInt(String number);
+
+    @JS.Coerce
+    @JS(value = "return parseInt(number, radix);")
```
patrick96: Prefer `Number.parseInt` for consistency

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
```diff
@@ -119,4 +119,136 @@ public boolean equals(Object that) {
...
+    public native static int parseInt(String number, int radix);
```
patrick96: `parseInt` can return `NaN` and it can also return integer values that don't fit into a 32bit int.

I see two approaches here:

* Return a `long` value and throw `IllegalArgumentException` when `parseInt` returns `NaN`
* Just return `JSNumber` here

@axel22  What do you think?

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
```diff
@@ -119,4 +119,136 @@ public boolean equals(Object that) {
...
+    @JS(value = "return parseFloat(number);")
+    public native static float parseFloat(double number);
+
+    @JS.Coerce
+    @JS(value = "return parseFloat(number);")
+    public native static float parseFloat(String number);
+
+    @JS.Coerce
+    @JS(value = "return parseInt(number);")
+    public native static int parseInt(double number);
```
patrick96: Depending on the approach we take for the return value, this method may not be necessary

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
```diff
@@ -119,4 +119,136 @@ public boolean equals(Object that) {
...
+    @JS(value = "return parseInt(number, radix);")
+    public native static int parseInt(String number, int radix);
+
+    @JS.Coerce
+    @JS(value = "return Number.EPSILON;")
+    public native static double EPSILON();
+
+    @JS.Coerce
+    @JS(value = "return Number.MAX_SAFE_INTEGER;")
+    public native static double MAX_SAFE_INTEGER();
```
patrick96: Should we return a `long` here and for `MIN_SAFE_INTEGER`?

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
```diff
@@ -119,4 +119,136 @@ public boolean equals(Object that) {
...
+    @JS.Coerce
+    @JS(value = "return this.toPrecision(precision);")
+    public native String toPrecision(int precision);
+
+    @JS.Coerce
+    @JS(value = "return this.toString(radix);")
+    private native String toJSString(int radix);
+
+    public String toString(int radix) {
+        return "JavaScript<" + typeof() + "; " + toJSString(radix) + ">";
```
patrick96: I think this should just return `Number.toString(radix)`.

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/api/JSNumberTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+        assertEquals("JavaScript<number; 11111111>", hex.toString(2));
+        assertEquals("JavaScript<number; ff>", hex.toString(16));
+        assertEquals("JavaScript<number; 377>", hex.toString(8));
+        assertEquals("JavaScript<number; -132>", neg.toString(5));
+    }
+
+    public static void testValueOf() {
+        assertEquals(42.0, JSNumber.of(INT).valueOf(), DELTA);
+        assertEquals(DOUBLE_SMALL, JSNumber.of(DOUBLE_SMALL).valueOf(), 0.000001);
```
patrick96: Does `DELTA` here not work?

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/api/JSNumberTest.java
```diff
@@ -0,0 +1,210 @@
...
+        assertFalse(JSNumber.isSafeInteger(JSNumber.of(DOUBLE_BIG)));
+        assertFalse(JSNumber.isSafeInteger(JSNumber.of(NAN)));
+        assertFalse(JSNumber.isSafeInteger(JSNumber.of(1e100)));
+    }
+
+    public static void testConstants() {
+        assertEquals(2.220446e-16, JSNumber.EPSILON(), 1e-16);
+        assertEquals(9.007199e+15, JSNumber.MAX_SAFE_INTEGER(), 0.000001e+15);
+        assertEquals(1.797693e+308, JSNumber.MAX_VALUE(), 0.000001e+308);
+        assertEquals(-9.007199e+15, JSNumber.MIN_SAFE_INTEGER(), 0.000001e+15);
```
patrick96: Use the value `-9007199254740991d` here directly. Similar for `MAX_SAFE_INTEGER`

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/api/JSNumberTest.java
```diff
@@ -0,0 +1,210 @@
...
+package com.oracle.svm.webimage.jtt.api;
+
+import org.graalvm.webimage.api.JSNumber;
+import org.graalvm.webimage.api.JSObject;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class JSNumberTest {
```
patrick96, line 36: Shouldn't all of the equality checks here use a delta of 0? We should be making sure that our API returns very specific values

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSValue.java
```diff
@@ -55,6 +55,12 @@ public abstract class JSValue {
     JSValue() {
     }
 
+    @SuppressWarnings("unchecked")
+    public static <R> R checkedCoerce(Object value, Class<R> cls) {
+        if (value instanceof JSValue jsResult) return jsResult.as(cls);
```
patrick96: Use curly braces

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSValue.java
```diff
@@ -55,6 +55,12 @@ public abstract class JSValue {
     JSValue() {
     }
 
+    @SuppressWarnings("unchecked")
+    public static <R> R checkedCoerce(Object value, Class<R> cls) {
```
patrick96, line 71: Add documentation to method

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSSymbol.java
patrick96: In that case I think we should modify this to return a Java string and `null` if the JS functions returned `undefined`

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSSymbol.java
```diff
@@ -87,4 +87,83 @@ public boolean equals(Object that) {
...
+    @JS(value = "return Symbol.toStringTag;")
+    public static native JSSymbol toStringTag();
+
+    @JS.Coerce
+    @JS(value = "return Symbol.unscopables;")
+    public static native JSSymbol unscopables();
+
+    @JS.Coerce
+    @JS(value = "return sym.valueOf();")
+    public static native JSSymbol valueOf(Object sym);
```
patrick96: Why is this method needed?

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/api/JSSymbolTest.java
```diff
@@ -0,0 +1,225 @@
...
+        testToStringTag();
+        testUnscopables();
+        testValueOf();
+    }
+
+    public static void testAsyncDispose() {
+        JSSymbol sym = JSSymbol.asyncDispose();
+
+        assertEquals("JavaScript<symbol; Symbol(nodejs.asyncDispose)>", sym.toString());
+        assertEquals(JSSymbol.class, sym.getClass());
```
patrick96: This is check is not needed here or in any of the tests below. `sym.getClass()` is always `JSSymbol.class` since the type of the variable is `JSSymbol`, which is a final class

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSSymbol.java
```diff
@@ -87,4 +87,83 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return a === b;")
+    public static native boolean isSameSymbol(JSSymbol a, JSSymbol b);
+
+    @JS.Coerce
+    @JS(value = "return sym.description;")
+    public static native String description(Object sym);
```
patrick96: Why is this not an instance method?

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/api/JSSymbolTest.java
```diff
@@ -0,0 +1,225 @@
...
+import org.graalvm.webimage.api.JSSymbol;
+import org.graalvm.webimage.api.JSValue;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertTrue;
+
+public class JSSymbolTest {
```
patrick96, line 37: I don't think we need to have tests for any of the well-known symbols. There doesn't seem to be a spec that defines what `description` or `keyFor` should produce for them. And just comparing against things like `nodejs.asyncDispose` ties us to the behavior of a specific node version.

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
         return javaString().hashCode();
     }
+
+    @JS.Coerce
+    @JS(value = "return String.fromCharCode(codeUnit);")
+    public static native JSString fromCharCode(int codeUnit);
+
+    @JS.Coerce
+    @JS(value = "return String.fromCharCode(...codeUnits);")
+    public static native JSString fromCharCode(int... codeUnits);
```
patrick96, line 88: Could use use `char... codeUnits` instead here? The function only accepts numbers from 0 to 65535 anyway.

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return String.fromCodePoint(codeUnit);")
+    public static native JSString fromCodePoint(int codeUnit);
+
+    @JS.Coerce
+    @JS(value = "return String.fromCodePoint(...codeUnits);")
+    public static native JSString fromCodePoint(int... codeUnits);
```
patrick96:
```suggestion
    public static native JSString fromCodePoint(int... codePoints);
```

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+
+    @JS(value = "return String.raw(template);")
+    public static native JSString raw(JSObject template);
+
+    @JS(value = """
+            const args = [];
+            for (let i = 0; i < substitutions.length; i++) {
+                args.push(substitutions[i]);
+            }
+            return String.raw(template, ...args);
```
patrick96: Does `...Array.from(substitutions)` work?

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+    @JS(value = "return this.localeCompare(compareString);")
+    public native int localeCompare(JSString compareString);
+
+    @JS.Coerce
+    @JS(value = "return this.localeCompare(compareString, locales);")
+    public native int localeCompare(JSString compareString, JSString locales);
+
+    @JS.Coerce
+    @JS(value = "return this.localeCompare(compareString, locales, options);")
+    public native int localeCompare(JSString compareString, JSString locales, Object options);
```
patrick96: This should be `JSObject` here and above

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return this.match(regexp);")
+    public native JSObject match(Object regexp);
```
patrick96:
```suggestion
    public native JSObject match(JSObject regexp);
```

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return this.matchAll(regexp);")
+    public native JSObject matchAll(Object regexp);
```
patrick96:
```suggestion
    public native JSObject matchAll(JSObject regexp);
```

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+    @JS(value = "return this.padStart(targetLength, padString);")
+    public native JSString padStart(int targetLength, JSString padString);
+
+    @JS.Coerce
+    @JS(value = "return this.repeat(count);")
+    public native JSString repeat(int count);
+
+    @JS.Coerce
+    @JS(value = "return this.replace(pattern, replacement);")
+    public native JSString replace(Object pattern, Object replacement);
```
patrick96:
```suggestion
    public native JSString replace(JSValue pattern, JSValue replacement);
```

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return this.replaceAll(pattern, replacement);")
+    public native JSString replaceAll(Object pattern, Object replacement);
```
patrick96:
```suggestion
    public native JSString replaceAll(JSValue pattern, JSValue replacement);
```

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return this.search(regexp);")
+    public native int search(Object regexp);
```
patrick96:
```suggestion
    public native int search(JSObject regexp);
```

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+    @JS(value = "return this.slice(indexStart, indexEnd);")
+    public native JSString slice(int indexStart, int indexEnd);
+
+    @JS.Coerce
+    @JS(value = "return this.split(separator);")
+    public native JSObject split(String separator);
+
+    @JS.Coerce
+    @JS(value = "return this.split(separator);")
+    public native JSObject split(JSObject separator);
```
patrick96:
```suggestion
    public native JSObject split(JSValue separator);
```

The separator could also be `JSString` or `JSUndefined`

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return this.split(separator, limit);")
+    public native JSObject split(String separator, int limit);
+
+    @JS.Coerce
+    @JS(value = "return this.split(separator, limit);")
+    public native JSObject split(JSObject separator, int limit);
```
patrick96:
```suggestion
    public native JSObject split(JSValue separator, int limit);
```

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
patrick96, line 1: This is missing the `substring` method

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+    @JS(value = "return this.trim();")
+    public native JSString trim();
+
+    @JS.Coerce
+    @JS(value = "return this.trimEnd();")
+    public native JSString trimEnd();
+
+    @JS.Coerce
+    @JS(value = "return this.trimRight();")
+    public native JSString trimRight();
```
patrick96: Do we need this? It's non-standard

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return this.trimStart();")
+    public native JSString trimStart();
+
+    @JS.Coerce
+    @JS(value = "return this.trimLeft();")
+    public native JSString trimLeft();
```
patrick96: Do we need this? It's non-standard

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,310 @@ public boolean equals(Object that) {
...
+
+    @JS.Coerce
+    @JS(value = "return this.valueOf();")
+    public native JSString valueOf();
```
patrick96: This method is not needed. It just returns the same string

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/api/JSStringTest.java
```diff
@@ -0,0 +1,537 @@
...
+
+import java.util.Map;
+import java.util.function.Function;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class JSStringTest {
```
patrick96, line 42: All uses of `.as(String.class)` can be replaced with `asString()`.

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
patrick96, line 1: This is missing the `assign`, `getOwnPropertySymbols`, `getPrototypeOf`, and `getOwnPropertyDescriptors` methods

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
```diff
@@ -547,4 +547,104 @@ public <T> T as(Class<T> cls) {
...
+    @JS(value = "return Object.create(proto);")
+    public static native JSObject create(JSObject proto);
+
+    @JS.Coerce
+    @JS(value = "return Object.create(proto, properties);")
+    public static native JSObject create(JSObject proto, JSObject properties);
+
+    @JS.Coerce
+    @JS(value = "return Object.defineProperties(obj, prop);")
+    public static native JSObject defineProperties(JSObject obj, JSObject prop);
```
patrick96:
```suggestion
    public static native JSObject defineProperties(JSObject obj, JSObject props);
```

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
```diff
@@ -547,4 +547,104 @@ public <T> T as(Class<T> cls) {
...
+    @JS(value = "return Object.defineProperty(obj, prop, descriptor);")
+    public static native JSObject defineProperty(JSObject obj, String prop, JSObject descriptor);
+
+    @JS.Coerce
+    @JS(value = "return Object.entries(obj);")
+    public static native JSObject entries(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "Object.freeze(obj);")
+    public static native void freeze(JSObject obj);
```
patrick96: We should also return `JSObject` here

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
```diff
@@ -547,4 +547,104 @@ public <T> T as(Class<T> cls) {
...
+    @JS(value = "return Object.getOwnPropertyNames(obj);")
+    public static native JSObject getOwnPropertyNames(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return Object.groupBy(items, callback);")
+    public native static JSObject groupBy(JSObject items, JSValue callback);
+
+    @JS.Coerce
+    @JS(value = "return Object.hasOwn(obj, prop);")
+    public static native boolean hasOwn(JSObject obj, String prop);
```
patrick96, line 597: Create overloads for `JSSymbol` and `JSString`

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
```diff
@@ -547,4 +547,104 @@ public <T> T as(Class<T> cls) {
...
+    @JS(value = "return Object.values(obj);")
+    public static native JSObject values(JSObject obj);
+
+    @JS.Coerce
+    @JS(value = "return this.isPrototypeOf(object);")
+    public native boolean isPrototypeOf(JSObject object);
+
+    @JS.Coerce
+    @JS(value = "return this.propertyIsEnumerable(prop);")
+    public native boolean propertyIsEnumerable(String prop);
```
patrick96, line 649: Create overloads for `JSSymbol` and `JSString`

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/api/JSObjectTest.java
```diff
@@ -0,0 +1,482 @@
...
+        set(nameDescriptor, JSString.of("Bob"), false, true);
+        JSObject properties = JSObject.create();
+        properties.set("name", nameDescriptor);
+
+        JSObject obj = JSObject.create(proto, properties);
+        boolean failed = false;
+        try {
+            obj.set("name", "Alice");
+        } catch (Exception e) {
+            failed = true;
```
patrick96: Use `assertThrows` here and in the rest of this class

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/api/JSObjectTest.java
```diff
@@ -0,0 +1,482 @@
...
+        }
+        String result = JSValue.checkedCoerce(obj.get("name"), String.class);
+        JSValue fun = (JSValue) obj.get("greet");
+        String greeting = apply(fun, null, "World");
```
patrick96: Isn't this the same as in `testPrototypeInheritance`? If so, this part can be removed from this test

web-image/src/com.oracle.svm.webimage.jtt/src/com/oracle/svm/webimage/jtt/api/JSObjectTest.java
```diff
@@ -0,0 +1,482 @@
...
+    @JS.Coerce
+    @JS("return it.toString();")
+    private static native String objToString(Object it);
+
+    @JS(value = """
+                    const arr = [];
+                    for (let i = 0; i < values.length; i++) {
+                        arr.push(values[i]);
+                    }
+                    return arr;
```
patrick96: Does `return Array.from(values)` work here?

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
Atur256: Using int... matches JavaScript’s fromCharCode, which takes numbers, not characters. While char... would work for values ≤ 65535, it forces explicit casting for int values.

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
Atur256: These were all functions that I was not able to implement for various reasons and therefore left out.

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSSymbol.java
```diff
@@ -87,4 +83,84 @@ public boolean equals(Object that) {
...
+    @JS(value = "return Symbol.for(key);")
+    public static native JSSymbol forKey(String key);
+
+    @JS.Coerce
+    @JS(value = "return Symbol.keyFor(sym);")
+    private static native JSValue keyForRaw(JSSymbol sym);
+
+    public static String keyFor(JSSymbol sym) {
+        JSValue result = keyForRaw(sym);
+        return result.typeof().equals("undefined") ? null : result.asString();
```
patrick96: Prefer `instanceof JSUndefined`. Or maybe we should even add a method to `JSValue` that tells us whether this is an undefined value

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
patrick96: In that case, we should have an overload that accepts `char...`

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSString.java
```diff
@@ -82,4 +82,304 @@ public boolean equals(Object that) {
...
+    @JS(value = "return String.fromCharCode(...codeUnits);")
+    public static native JSString fromCharCode(int... codeUnits);
+
+    @JS.Coerce
+    @JS(value = "return String.fromCodePoint(...codePoints);")
+    public static native JSString fromCodePoint(int... codePoints);
+
+    @JS(value = """
+            const sub = Array.from(substitutions);
+            console.log(sub);
```
patrick96: Forgotten log statement

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
patrick96: That's fine in that case. We can add them once we fix the underlying issues

web-image/src/com.oracle.svm.hosted.webimage/src/com/oracle/svm/hosted/webimage/codegen/oop/ClassWithMirrorLowerer.java
```diff
@@ -558,7 +558,7 @@ protected void lowerClassEnd() {
 
         // Store the mapping from the imported JavaScript class constructor to the Java facade class
         // under which the JavaScript class was imported.
-        if (isImportedClass) {
+//        if (isImportedClass) {
```
patrick96: Undo this again, we will implement this properly on our side, as it require some more work.

web-image/src/com.oracle.svm.hosted.webimage/src/com/oracle/svm/hosted/webimage/codegen/runtime/jsconversion.js
```diff
@@ -529,7 +529,7 @@ class Conversion {
         const facades = runtime.findFacadesFor(obj.constructor);
         const rawJavaHub = cls[runtime.symbol.javaNative];
         const internalJavaClass = rawJavaHub[runtime.symbol.jsClass];
-        if (facades.has(internalJavaClass)) {
+        if (obj.constructor === ({}).constructor || facades.has(internalJavaClass)) {
```
patrick96: Also undo this

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java
```diff
@@ -537,9 +537,12 @@ public double[] asDoubleArray() {
     @SuppressWarnings("unchecked")
     @Override
     public <T> T as(Class<T> cls) {
-        Object facade = extractFacadeClass(cls);
+        if (cls.isAssignableFrom(this.getClass())) {
```
patrick96: Also undo this

sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java
patrick96, line 1: This file fails a bunch of our internal style checks:

```
Checkstyle ends with 26 errors.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSObject.java:593: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:125: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:129: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:133: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:137: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:141: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:145: 'static' modifier out of order with the JLS suggestions.
gate: 27 Oct 2025 05:42:29(+01:13) ABORT: Checkstyle [0:00:03.179260] [disk (free/total): 258.5GB/926.4GB]
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:149: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:153: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:157: 'static' modifier out of order with the JLS suggestions.
gate: 27 Oct 2025 05:42:29(+01:13) END:   Checkstyle [0:00:03.179850] [disk (free/total): 258.5GB/926.4GB]
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:185: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:185: Name 'EPSILON' must match pattern '^[a-z][a-zA-Z0-9]*$'.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:189: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:189: Name 'MAX_SAFE_INTEGER' must match pattern '^[a-z][a-zA-Z0-9]*$'.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:193: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:193: Name 'MAX_VALUE' must match pattern '^[a-z][a-zA-Z0-9]*$'.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:197: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:197: Name 'MIN_SAFE_INTEGER' must match pattern '^[a-z][a-zA-Z0-9]*$'.
gate: 27 Oct 2025 05:42:29(+01:13) ABORT: Gate [0:01:13.733226] [disk (free/total): 258.5GB/926.4GB]
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:201: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:201: Name 'MIN_VALUE' must match pattern '^[a-z][a-zA-Z0-9]*$'.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:205: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:205: Name 'NaN' must match pattern '^[a-z][a-zA-Z0-9]*$'.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:209: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:209: Name 'NEGATIVE_INFINITY' must match pattern '^[a-z][a-zA-Z0-9]*$'.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:213: 'static' modifier out of order with the JLS suggestions.
/Users/graal/slave/e/main/sdk/src/org.graalvm.webimage.api/src/org/graalvm/webimage/api/JSNumber.java:213: Name 'POSITIVE_INFINITY' must match pattern '^[a-z][a-zA-Z0-9]*$'.
```

Please fix the ordering issues (it should be `static native` and not `native static`). And for the names, we should also follow the regular method name style, e.g. `maxSafeInteger` or `nan`.

PR #12277, opened 2025-10-02
[GR-70136] Add macro for LLVM backend

graalvmbot:
This makes the LLVM backend easier to use, by bundling all necessary options to run it into a single, easy-to-use macro. The macro performs the following actions:
* Sets the `-H:CompilerBackend=llvm` option
* Registers the LLVM backend-specific options to the option checker
* Includes the required jars on the module path


substratevm/src/com.oracle.svm.core.graal.llvm/src/com/oracle/svm/core/graal/llvm/LLVMBackend.md
```diff
@@ -11,7 +11,7 @@ mx --dynamicimports /substratevm build
 export JAVA_HOME=$(mx --dynamicimports /substratevm graalvm-home)
 ```
 
-To enable the LLVM backend, pass the `-H:CompilerBackend=llvm` option to the `native-image` command. 
+To enable the LLVM backend, pass the `--tool:llvm` option to the `native-image` command.
```
koutheir, line 14: @loicottet, I think you meant `--tool:llvm-backend` instead of `--tool:llvm`. Right?

loicottet: Yes, this has a fix on the way

Issue #12253, opened 2025-09-28
[Native Image] JFR native memory usage increases steadily


joggeli34:
### Describe the Issue

We use the build flag `--enable-monitoring=jfr,nmt` as we like to use pyroscope for profiling. This works in general well, but we see a constant increase of memory that is not on the heap.

On the left, the container memory. On the right the heap:
<img width="1082" height="312" alt="Image" src="https://github.com/user-attachments/assets/392569bf-4067-4b94-89fd-c8255ac67fe8" />

With the native memory tracking, we see that the JFR native memory is big and it seems to be growing the longer we let the application run:
```
Total
    (reserved=33774140KB, committed=762642KB)
    (malloc=219708KB, count=24817)
    (peak malloc=288450KB, count at peak=57367)
    (mmap: reserved=33554432KB, committed=542933KB)
    (mmap: peak reserved=33554432KB, peak committed=753190KB)
Java Heap
    (reserved=33461440KB, committed=450453KB)
    (malloc=0KB, count=0)
    (peak malloc=0KB, count at peak=0)
    (mmap: reserved=33461440KB, committed=450453KB)
    (mmap: peak reserved=33461440KB, peak committed=660710KB)
JFR
    (reserved=198613KB, committed=198613KB)
    (malloc=198613KB, count=24706)
    (peak malloc=232462KB, count at peak=62502)
    (mmap: reserved=0KB, committed=0KB)
    (mmap: peak reserved=0KB, peak committed=0KB)
```

We use the pyroscope-java, which creates a lot of JFR recordings with this code:
https://github.com/grafana/pyroscope-java/blob/main/agent/src/main/java/io/pyroscope/javaagent/JFRJDKProfilerDelegate.java

Otherwise we use default setting for the JFR, so we add no additional settings as start parameter.
By my understanding, the JFR default settings should use a lot less memory and should not lead to constant growing memory. Or is there a setting I missed to limit that?

### Using the latest version of GraalVM can resolve many issues.

- [x] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

### GraalVM Version

java 25 2025-09-16 LTS
Java(TM) SE Runtime Environment Oracle GraalVM 25+37.1 (build 25+37-LTS-jvmci-b01)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 25+37.1 (build 25+37-LTS-jvmci-b01, mixed mode, sharing)

### Operating System and Version

[...]

PR #12290, opened 2025-10-03
[GR-70360] Fix JFR memory leak in Java event TLBs

roberttoyonaga:
This PR fixes a native memory leak in JFR. The problem was reported originally by @joggeli34
 here https://github.com/oracle/graal/issues/12253

The problem is that JFR Java event buffers are not being retired/reinstated correctly. Instead of these buffers being reused between recordings, they are leaked. The reason for the leak is because we prematurely reset the `javaBuffer` thread local when stopping a recording. This is is bad because we need to reuse this thread local to reinstate the buffer in the next recording when a new Java `EventWriter` is created. Instead, a new buffer gets created and the retired one is lost.

This results in a leak of the size [(pageSize > 8 * 1024 ? pageSize : 8 * 1024)](https://github.com/openjdk/jdk/blob/jdk-26%2B18/src/jdk.jfr/share/classes/jdk/jfr/internal/Options.java#L67) * THREAD_COUNT  per new recording. 

The leak only happens when a recording is stopped and a new recording is started while the same threads are still running. This is the scenario when Java event buffers would be retired/reinstated.  
The leak was discovered by @joggeli34 because they are using [Pyroscope](https://github.com/grafana/pyroscope-java) which starts/stops new recordings every 10s. This frequent new recording creation exacerbates the leak. 

I used NMT to investigate this by adding more fine grained JFR NMT categories. Notice that committed memory in  category JFR1 increases overtime.
image
image
image
image

After the fix, the memory usage does not increase. 

This bug fix should be backported to GraalVM for JDK 25.

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrThreadLocal.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/JfrThreadLocal.java
@@ -217,7 +217,7 @@ public class JfrThreadLocal implements ThreadListener {
     @Uninterruptible(reason = "Locking without transition requires that the whole critical section is uninterruptible.")
     private static void flushToGlobalMemoryAndRetireBuffer(JfrBuffer buffer) {
         assert VMOperation.isInProgressAtSafepoint();
-        if (buffer.isNull()) {
+        if (buffer.isNull() || JfrBufferAccess.isRetired(buffer)) {
             return;
         }
 


christianhaeubl (approved):
Thanks, changes look good. Will be integrated in the next few days.

roberttoyonaga:
@christianhaeubl should I make a backport PR to the JDK 25 release branch?

christianhaeubl:
No, that should not be necessary - thanks (I already requested backporting internally, so this should land in the next possible release).

PR #12371, opened 2025-10-15
Fix jniobjecthandles

asselyam:
In ObjectHandlesImpl all methods specific to weak handles (like createWeak, isWeak, and destroyWeak) were removed. It no longer knows the difference between a strong and a weak handle. Its only job is to store an object at a given index and retrieve it.

In JNIGlobalHandles the single ObjectHandlesImpl instance was replaced with two separate instances: strongGlobalHandles and weakGlobalHandles. The 64-bit global handle space was split into two distinct ranges (strong and weak) by using a bit flag (bit 62). Two new methods, encodeStrong and encodeWeak, were created to correctly place a handle's value into its designated range.


asselyam:
Hello Dr. @christianhaeubl, per our discussion, I've opened this PR with the strong/weak handle refactoring. I'd appreciate your review when you have a moment. Thank you!

PR #12455, opened 2025-10-29
Implement native object handles feature

asselyam:
Corrected the implementation of the first steps and removed the duplicated code:

ObjectHandlesImpl is now only a storage class. Each instance contain either weak or strong handles.
JNIGlobalHandles uses 2 instances of ObjectHandlesImpl (one for strong handles, one for weak handles)
HandleWeakReference is removed


asselyam:
@christianhaeubl Dear Dr. Häubl, i corrected the code, compile it and tested. If I understand right i create a new PR for every task or everzthing should be in one PR but in different branches?

PR #12500, opened 2025-11-07
[GR-71287] Reduce JFR RSS contribution by substituting method tracing and timing classes

roberttoyonaga:
[JEP 520](https://github.com/openjdk/jdk/pull/25306) introduced JFR method tracing and timing. This is a feature we do not support yet because it involves significant changes in Hotspot.  If we substitute related classes we can reduce code size and image heap. 

Using a [basic Quarkus quickstart app](https://github.com/quarkusio/quarkus-quickstarts/tree/main/getting-started) this change results in a 0.92 MB code size decrease and a  1.31 MB image heap size decrease. 

***Before***
image
***After***
image

I see a similar reduction when using a simple hello world app:

***Before***
image

***After***
image

If a user tries to use JFR method tracing and timing from the command line they will receive a warning log then JFR will continue working as normal:
```
$  ./simplebusywait -XX:StartFlightRecording=filename=rec.jfr,method-timing=::java.util.HashMap::resize
[warn][jfr,start] The .jfc option/setting 'method-timing' doesn't exist or is not supported.
```

If they try to use method tracing and timing by using a JFR configuration file (.jfc) instead, they will also get a warning:
```
$ ./simplebusywait -XX:StartFlightRecording=filename=rec.jfr,settings=custom.jfc
[warn][jfr,setting] Method timing and tracing is not supported yet.
```

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/Target_jdk_jfr_internal_settings_MethodSetting.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jfr/Target_jdk_jfr_internal_settings_MethodSetting.java
@@ -31,6 +31,7 @@ import com.oracle.svm.core.annotate.TargetClass;
 import jdk.jfr.internal.PlatformEventType;
 import java.util.List;
 
+@SuppressWarnings({"unused", "static-method"}) //
 @TargetClass(className = "jdk.jfr.internal.settings.MethodSetting")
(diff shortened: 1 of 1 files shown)


christianhaeubl:
Thanks, your PR should get integrated in the next few days.

christianhaeubl (approved):

roberttoyonaga:
@christianhaeubl Can this be backported to the GraalVM for JDK 25 branch?  I can open the backport request if you agree.

christianhaeubl:
Should be safe to backport, feel free to open a the backport request.

PR #12503, opened 2025-11-07
Refactor ObjectHandlesImpl to use native memory

asselyam:
This change:
- Updates the 'buckets' field type to WordPointer.
- Eliminates the use of Unsafe in favor of WordBase pointer arithmetic (Pointer.add and WordPointer.addressOf).
- Utilizes Pointer.writeObject/readObject for managing Java object references in native memory.


asselyam:
@christianhaeubl  Dear Dr. Haeubl here is the updated code. I couldn't run mx native-unittest because of the GraalVM compiler bug. I sent you in Slack the output i got and i was wondering if you could check my code and give some recommendations. I was trying to solve this for the last couple of days but dont fully understand. Thank you!

PR #12537, opened 2025-11-13
Native memory handles

asselyam:
- Use of native memory instead of the 2 dimensional `Object[][]` for `ObjectHandlesImpl.buckets`. 
- References to Java objects into the native memory.


asselyam:
@christianhaeubl  Dear Dr. Häubl here is the last update. Sorry I cannot add myself you as a reviewer

PR #12554, opened 2025-11-17
knowing the result

aUsABuisnessman:
}
	const cacheKeysToGet = [
	  .dexBlacklistCacheKeys,
	  .dexRestrictionCacheKeys,
	] ;
	
	if (!cacheKeysToGet.length) {
		return dexKeys;
	}

	const result = await this.dexAdapterService.
dexHelper.cache.mget(
	cacheKeysToGet,
	);
	  (_,i) => result[i], // knowing the result is not NULL is sufficent


oubidar-Abderrahim:
Hi, I will close this PR since the OCA was not signed yet. Once you have a valid OCA you can create a new PR

PR #12557, opened 2025-11-18
Allow explicit marking of local variables as the key for merge explosion with `CompilerDirectives.mergeExplodeKey`

blaumeise20:
This PR adds a method `CompilerDirectives.mergeExplodeKey` which can be used to explicitly mark variables to be used as the merge key in `MERGE_EXPLODE` loop explosion.

Currently, the entire frame state is used for merging. Taking Espresso as an example, if you would jump to the same `bci` but assign a different value to another variable, it will result in two separate code paths. This can lead to unexpected graph size explosion and bugs. With the method introduced in this PR, you can mark a single variable to be used as the key. Truffle will merge on this key only, but verify that other variables stay the same. If merging should occur but there is a mismatch, the compilation will bail out.

Future steps include allowing multiple variables to be marked simultaneously (which has to be extended into irreducible loop handling), and creating an option for explicit PHI nodes, so local variables can be used as counters without the need for objects.


blaumeise20:
@gilles-duboscq

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
```diff
@@ -627,4 +627,30 @@ static final class ShouldNotReachHere extends AssertionError {
...
+     * }
+     * </pre>
+     *
+     * Only a single variable can currently be annotated with this method, and at every iteration,
+     * the value must be a constant integer.
+     *
+     * @param i the variable that should be used as the merge key.
+     * @return the unchanged value
+     */
+    @SuppressWarnings("unused")
```
gilles-duboscq: Why do we need this?

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
```diff
@@ -627,4 +627,30 @@ static final class ShouldNotReachHere extends AssertionError {
...
+     * while (bci != END_BCI) {
+     *     // ...
+     * }
+     * </pre>
+     *
+     * Only a single variable can currently be annotated with this method, and at every iteration,
+     * the value must be a constant integer.
+     *
+     * @param i the variable that should be used as the merge key.
+     * @return the unchanged value
```
gilles-duboscq, line 842: Please add `@since 25.1`

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
```diff
@@ -514,6 +515,17 @@ public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Rec
                 return true;
             }
         });
+
+        r.register(new RequiredInvocationPlugin("mergeExplodeKey", int.class) {
+            @Override
+            public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver, ValueNode value) {
+                if (canDelayIntrinsification) {
+                    return false;
```
gilles-duboscq: I think this deserves a comment explaining why we do this.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/LoopExplosionKeyNode.java
```diff
@@ -0,0 +1,52 @@
...
+
+import jdk.graal.compiler.graph.Node;
+import jdk.graal.compiler.graph.NodeClass;
+import jdk.graal.compiler.nodeinfo.NodeInfo;
+import jdk.graal.compiler.nodes.calc.FloatingNode;
+import jdk.graal.compiler.nodes.spi.Canonicalizable;
+import jdk.graal.compiler.nodes.spi.CanonicalizerTool;
+
+@NodeInfo(nameTemplate = "LoopExplosionKey", cycles = CYCLES_0, size = SIZE_0)
+public final class LoopExplosionKeyNode extends FloatingNode implements Canonicalizable {
```
gilles-duboscq: What happens if the directive is used in a non-merge-explode method and this node remains in the graph?
I think this should either be a very explicit compilation failure or performance warning. If it's only a warning then we need to remove such nodes.

@chumer what do you think? Should using `CompilationDirective.mergeExplodeKey` outside of the context of a `@ExplodeLoop(kind = LoopExplosionKind.MERGE_EXPLODE)` method be a compilation error or a performance warning?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -660,26 +653,25 @@ public final void decode(EncodedGraph encodedGraph) {
...
-                    nodeReference.node = loopScope.getNode(nodeReference.orderId);
+                    nodeReference.node = initialLoopScope.getNode(nodeReference.orderId);
                     if (nodeReference.node == null || !nodeReference.node.isAlive()) {
                         throw GraalError.shouldNotReachHere("Could not decode the EncodedNodeReference"); // ExcludeFromJacocoGeneratedReport
                     }
                     nodeReference.orderId = EncodedGraph.EncodedNodeReference.DECODED;
                 }
             }
-
-            graph.maybeMarkUnsafeAccess(encodedGraph);
```
gilles-duboscq, line 746: Why did you remove this?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -727,55 +723,75 @@ protected final LoopScope createInitialLoopScope(MethodScope methodScope, FixedW
...
                 }
+            }
+        }
+    }
 
+    private MethodScope processMethodScope(MethodScope methodScope) {
+        while (methodScope.currentLoopScope != null) {
+            /* Process nodes of loop. */
+            while (!methodScope.currentLoopScope.nodesToProcess.isEmpty()) {
+                // TODO(blaumeise20): move this
```
gilles-duboscq: If we leave such a TODO here we'll need a clearer comment (move where?, why/when?)

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -1131,9 +1141,47 @@ protected void handleLoopExplosionBegin(MethodScope methodScope, LoopScope loopS
         FrameState frameState = loopBegin.stateAfter();
 
         if (methodScope.loopExplosion.mergeLoops()) {
-            LoopExplosionState queryState = new LoopExplosionState(frameState, null);
-            LoopExplosionState existingState = loopScope.iterationStates.get(queryState);
+            if (loopScope.trigger == LoopScopeTrigger.START) {
+                List<Integer> mergeKey = null;
```
gilles-duboscq: Maybe this could be move to a method that looks for merge keys?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -1169,9 +1217,43 @@ protected void handleLoopExplosionBegin(MethodScope methodScope, LoopScope loopS
...
+            for (int i : loopScope.loopExplosionMergeKeySlots) {
+                ValueNode node = frameState.values().get(i);
+                if (!node.isConstant() || node.asJavaConstant().getJavaKind() != JavaKind.Int) {
+                    throw new PermanentBailoutException("Graal implementation restriction: Method with %s loop explosion must have a loop variable of type int. %s",
+                                    LoopExplosionPlugin.LoopExplosionKind.MERGE_EXPLODE, node);
+                }
+                values.add(node);
+            }
+        } else {
+            values = frameState.values().stream().toList();
```
gilles-duboscq:
```suggestion
            values = frameState.values().snapshot();
```

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
gilles-duboscq:
```suggestion
                List<Integer> mergeKeys = null;
```

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -1169,9 +1217,43 @@ protected void handleLoopExplosionBegin(MethodScope methodScope, LoopScope loopS
...
+    }
+
+    protected LoopExplosionKey createLoopExplosionKey(LoopScope loopScope, FrameState frameState) {
+        List<ValueNode> values;
+        if (loopScope.loopExplosionMergeKeySlots != null) {
+            values = new ArrayList<>(loopScope.loopExplosionMergeKeySlots.size());
+            for (int i : loopScope.loopExplosionMergeKeySlots) {
+                ValueNode node = frameState.values().get(i);
+                if (!node.isConstant() || node.asJavaConstant().getJavaKind() != JavaKind.Int) {
+                    throw new PermanentBailoutException("Graal implementation restriction: Method with %s loop explosion must have a loop variable of type int. %s",
```
gilles-duboscq:
```suggestion
                    throw new PermanentBailoutException("Graal implementation restriction: merge keys must partial evaluate to int constants at the loop header. %s",
```

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -1169,9 +1217,43 @@ protected void handleLoopExplosionBegin(MethodScope methodScope, LoopScope loopS
...
+        }
+    }
+
+    protected LoopExplosionKey createLoopExplosionKey(LoopScope loopScope, FrameState frameState) {
+        List<ValueNode> values;
+        if (loopScope.loopExplosionMergeKeySlots != null) {
+            values = new ArrayList<>(loopScope.loopExplosionMergeKeySlots.size());
+            for (int i : loopScope.loopExplosionMergeKeySlots) {
+                ValueNode node = frameState.values().get(i);
+                if (!node.isConstant() || node.asJavaConstant().getJavaKind() != JavaKind.Int) {
```
gilles-duboscq, line 1563: Should we accept all integer types (i.e., also accept longs)?
What do you think @chumer?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -2285,27 +2373,23 @@ private void findLoopExits(Loop loop) {
...
-                    continue outer;
+                while (current instanceof FixedWithNextNode c) {
+                    current = c.next();
                 }
+                if (current instanceof EndNode end && methodScope.loopExplosionMerges.contains(end.merge())) {
+                    // we found the place for the loop exit introduction since the subsequent
+                    // merge has a frame state
+                    loop.exits.removeIf(x -> x.merge() == merge);
+                    loop.exits.add(end);
+                    continue;
```
gilles-duboscq: this `continue` doesn't look necessary

espresso/src/com.oracle.truffle.espresso/src/com/oracle/truffle/espresso/nodes/BytecodeNode.java
```diff
@@ -755,6 +755,7 @@ private Object executeBodyFromBCI(VirtualFrame frame, int startBCI, int startTop
             livenessAnalysis.onStart(frame, skipLivenessActions);
         }
 
+        curBCI = CompilerDirectives.mergeExplodeKey(curBCI);
```
gilles-duboscq, line 773: This could be moved to a separate commit

gilles-duboscq (commented):
Please also check the 2 failed gates

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
blaumeise20: I intentionally called it this way because the entire list is the "key" comprised of multiple values and there aren't multiple separate "keys", just like a multi-column `PRIMARY KEY` in a database. But I'll change it if you think plural makes more sense here.

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
blaumeise20: I forgot to remove it after doing some testing...

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
blaumeise20: This value is never cleared and it is already set in `recordGraphElements`, so the call is redundant.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
blaumeise20: I copied it from a different place (`isPartialEvaluationConstant`), but I can add a comment to explain.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
gilles-duboscq: That's fair, let's leave it as-is then

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
```diff
@@ -627,4 +627,31 @@ static final class ShouldNotReachHere extends AssertionError {
...
+     *
+     * @since 25.1
```
chumer, line 844: Please add an entry about this in the truffle changelog.md. Including a migration recommendation that if you use ExplodeLoop MERGE that you should also use this new API.

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
```diff
@@ -627,4 +627,31 @@ static final class ShouldNotReachHere extends AssertionError {
...
+     * int bci = 0;
+     * // ...
+     * bci = CompilerDirectives.mergeExplodeKey(bci);
+     * while (bci != END_BCI) {
+     *     // ...
+     * }
+     * </pre>
+     *
+     * Only a single variable can currently be annotated with this method, and at every iteration,
+     * the value must be a constant integer.
```
chumer: @gilles-duboscq mentioned something about it fails if not all values are constant based off the key. This should be documented here.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
```diff
@@ -514,6 +515,22 @@ public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Rec
...
+        r.register(new RequiredInvocationPlugin("mergeExplodeKey", int.class) {
+            @Override
+            public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver, ValueNode value) {
+                if (canDelayIntrinsification) {
+                    /*
+                     * We make sure to keep the original call while parsing the method so we don't
+                     * lose the node when encoding the graph. At the time of decoding,
+                     * `canDelayIntrinsification` will be false, so the actual node is created.
+                     */
+                    return false;
```
chumer: That is a good thought, but I don't think it is necessary here to delay the intrinisification.
We delay it for isPartialEvaluationConstant because the interpretation will be different if the argument is a constant and it is only really a PE constant during decoding. For this intrinsic it does not make a difference as we unconditionally wrap the value into LoopExplosionKeyNode. So I think it would actually be better always do this during parsing already, as a method once parsed is decoded many times, so we avoid the overhead of applying the intrinsic here. Probably a minor effect on compile speed, but small things add up and its also simpler this way.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/LoopExplosionKeyNode.java
chumer: I would go with a compilation error. We can do a similar thing as in VerifyFrameDoesNotEscapePhase.
Register the phase in TruffleTier.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -727,55 +723,74 @@ protected final LoopScope createInitialLoopScope(MethodScope methodScope, FixedW
...
                 }
+            }
+        }
+    }
 
+    private MethodScope processMethodScope(MethodScope methodScope) {
+        while (methodScope.currentLoopScope != null) {
+            /* Process nodes of loop. */
+            while (!methodScope.currentLoopScope.nodesToProcess.isEmpty()) {
+                // TODO(blaumeise20): use MethodScope instead of LoopScope here and set currentLoopScope directly
```
chumer: TODOs must never link to user names but either link Jira or Github issues.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -1169,11 +1185,79 @@ protected void handleLoopExplosionBegin(MethodScope methodScope, LoopScope loopS
...
+                        if (loopScope.initialCreatedNodes[j] == keyNode) {
+                            loopScope.initialCreatedNodes[j] = keyNode.value;
+                        }
+                    }
+                }
+
+                keyNode.replaceAndDelete(keyNode.value);
+            }
+        }
+        if (mergeKey != null && mergeKey.size() > 1) {
```
chumer: Please add a comment that if mergeKey is null we do what we previously did and that is there for compatibility reasons if the merge key is not yet used.

chumer (changes requested):

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
gilles-duboscq: I think this refers to the case that thows "Method with %s loop explosion has differing values in non-key locals".
Other values don't have a to be constant, they only need to have the same (ssa) value.
Maybe "Other variables must have the same value at every iteration (e.g., partial evaluate to the same constant or remain unchanged)"

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
blaumeise20: I couldn't manage to keep the node alive until decoding. It would always vanish, even though I explicitly made `canonicalize` return `this`. This was the only solution I found that worked. Do you have any idea why the node could have been deleted?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
chumer: Just ints are fine for now imo.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
chumer: Out of my expertise. Maybe @davleopo knows?

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
blaumeise20: What issue ID should I specify?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
blaumeise20: I saw quite a few TODOs with a username or even no assignment at all, so I thought it would be fine... What should I put here instead?

gilles-duboscq: Well it's still unclear to me as part of which change you want to do this and why.
FYI there is:
* GR-71753 Allow multiple explicit keys in MERGE_EXPLODE
* GR-71752 Allow marking variables explicitly for phi node creation in MERGE_EXPLODE

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
blaumeise20: I'll add it in the method below where the actual values are read from the frame state, this here is just the check for multiple marked variables.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
blaumeise20: Thanks! The way this is currently handled isn't particularly beautiful and I was thinking of actually setting the `currentLoopScope` inside `processNextNode`. I already got a few changes (not done yet) that I would add as part of the next step. Should I specify `GR-71752` then?

chumer: > I saw quite a few TODOs with a username 

It is only a strongly enforced rule for the Truffle team, not necessarily for the compiler team. I think it is good practice regardless.

gilles-duboscq: > Should I specify GR-71752 then

That should be fine

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
gilles-duboscq: You can use GR-71613 ("Allow explicitly selecting the value used for merging in MERGE_EXPLODE")

gilles-duboscq:
It's a bit hard to understand what changed when you both rebase and make some changes. That means we have to re-review everything. I'd suggest to either make 2 pushes (one of the rebase and one for the extra changes) or push the extra changes as not-yet-squashed fixup commits.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -1131,9 +1140,16 @@ protected void handleLoopExplosionBegin(MethodScope methodScope, LoopScope loopS
...
         if (methodScope.loopExplosion.mergeLoops()) {
-            LoopExplosionState queryState = new LoopExplosionState(frameState, null);
-            LoopExplosionState existingState = loopScope.iterationStates.get(queryState);
+            if (loopScope.trigger == LoopScopeTrigger.START) {
+                loopScope.loopExplosionMergeKeySlots = computeIndicesForMergeKey(loopScope, frameState);
+            }
+            LoopExplosionKey key = createLoopExplosionKey(loopScope, frameState);
+            LoopExplosionState existingState = methodScope.iterationStates.get(key);
             if (existingState != null) {
+                if (!frameStateEquals(frameState, existingState.state)) {
```
gilles-duboscq: Should we check the framestates only if `loopScope.loopExplosionMergeKeySlots != null`? just to avoid the compile-time penalty for loops not using explicit merge keys.

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
```diff
@@ -790,4 +790,33 @@ static final class ShouldNotReachHere extends AssertionError {
         }
 
     }
+
+    /**
+     * Marks a local variable as part of the key used for merging exploded loop iterations. This
```
gilles-duboscq:
```suggestion
     * Marks a local variable as part of the key used for merging {@linkplain LoopExplosionKind#MERGE_EXPLODE exploded loop} iterations. This
```

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
blaumeise20: Any updates on this?

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/truffle/test/MergeExplodeKeyTest.java
```diff
@@ -0,0 +1,222 @@
...
+import com.oracle.truffle.api.frame.FrameSlotKind;
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.nodes.ExplodeLoop;
+import com.oracle.truffle.api.nodes.ExplodeLoop.LoopExplosionKind;
+import com.oracle.truffle.api.nodes.RootNode;
+
+import jdk.vm.ci.code.BailoutException;
+
+/**
+ * Tests the `CompilerDirectives.mergeExplodeKey` system for explicit merge explode keys.
```
davleopo: can we use a `@link` to `mergeExplodeKey` here?

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/truffle/test/MergeExplodeKeyTest.java
```diff
@@ -0,0 +1,222 @@
...
+        public static final byte JMP = 6;
+        public static final byte IFZERO = 7;
+        public static final byte RETURN = 8;
+    }
+
+    public static class Program extends RootNode {
+        protected final String name;
+        @CompilationFinal(dimensions = 1) protected final byte[] bytecodes;
+        protected final int stackOffset;
+        protected final boolean markTopAsKey;
```
davleopo, line 86: please add doc for the fields - especially `markTopAsKey` will have a special anticipated usage based on the name I assume

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/truffle/test/MergeExplodeKeyTest.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        @Override
+        @ExplodeLoop(kind = LoopExplosionKind.MERGE_EXPLODE)
+        public Object execute(VirtualFrame frame) {
+            int top = -1;
+            int bci = 0;
+            bci = CompilerDirectives.mergeExplodeKey(bci);
+            if (markTopAsKey) {
+                /* Testing error on multiple key variables. */
```
davleopo: so this is future work then? should this be marked as TODO?

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/truffle/test/MergeExplodeKeyTest.java
```diff
@@ -0,0 +1,222 @@
...
+                        } else {
+                            bci = bci + 2;
+                            continue;
+                        }
+                    }
+                    case Bytecode.RETURN: {
+                        int value = getInt(frame, top);
+                        return value;
+                    }
+
```
davleopo, line 193: there are a few spurious new lines, can we remove them?

truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilable.java
```diff
@@ -227,8 +227,6 @@ default int getSuccessfulCompilationCount() {
      * mean it will always be inlined, as inlining is a complex process that takes many factors into
      * account. If this method returns <code>false</code>, it will never be inlined. This typically
      * means that compilation with this compilable as the root has failed.
-     *
```
davleopo, line 230: unrelated change? can this be removed?

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
```diff
@@ -790,4 +791,33 @@ static final class ShouldNotReachHere extends AssertionError {
         }
 
     }
+
+    /**
+     * Marks a local variable as part of the key used for merging
```
davleopo: not sure how approachable "key used for merging" is here - Ill let @chumer  comment on the docs, but I think we need some more doc here how this can and cannot be used. The docs should also mention why its an `ìnt` etc

truffle/CHANGELOG.md
```diff
@@ -61,9 +61,9 @@ This changelog summarizes major changes between Truffle versions relevant to lan
...
 * GR-71887: Bytecode DSL: Added a `ClearLocal` operation for fast clearing of local values.
-* GR-71088 Added `CompilerDirectives.EarlyEscapeAnalysis` annotation that runs partial escape analysis early before partial evaluation enabling partial-evaluation-constant scalar replacements.
 * GR-71402: Added `InteropLibrary#isHostObject` and `InteropLibrary#asHostObject` for accessing the Java host-object representation of a Truffle guest object. Deprecated `Env#isHostObject`, `Env#isHostException`, `Env#isHostFunction`, `Env#isHostSymbol`, `Env#asHostObject`, and `Env#asHostException` in favor of the new InteropLibrary messages.
 * GR-71402: Added `InteropLibrary#hasStaticScope` and `InteropLibrary#getStaticScope` returning the static scope representing static or class-level members associated with the given meta object.
+* GR-71613: Added `CompilerDirectives.mergeExplodeKey` method for explicitly marking a local variable as a key for `@ExplodeLoop(MERGE_EXPLODE)` methods. It is recommended to migrate all merge exploded loops to use this method to catch unintended graph size explosions.
```
davleopo, line 69: should we define a migration period (in terms of releases) and after that issue warnings @chumer ?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
davleopo: what do you mean by `vanish` ? Can you elaborate - or alternatively I can have  a look if you give me a cmd line. I assume your new test would reveal this?

davleopo: I also dont see a reason to delay it for now actually.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/phases/VerifyMergeExplodeKeyDoesNotRemainPhase.java
```diff
@@ -0,0 +1,47 @@
...
+package jdk.graal.compiler.truffle.phases;
+
+import jdk.graal.compiler.graph.GraalGraphError;
+import jdk.graal.compiler.nodes.LoopExplosionKeyNode;
+import jdk.graal.compiler.nodes.StructuredGraph;
+import jdk.graal.compiler.nodes.util.GraphUtil;
+import jdk.graal.compiler.phases.BasePhase;
+import jdk.graal.compiler.truffle.TruffleTierContext;
+
+/**
```
davleopo, line 29: Is this phase really necessary? can we not check the graph state of the `this.graph()` at the point of interest and fail in the verify method? or alternatively make the node lir lowerable and fail if it still exists during lir generation?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/LoopExplosionKeyNode.java
```diff
@@ -0,0 +1,53 @@
...
+import jdk.graal.compiler.nodes.spi.CanonicalizerTool;
+
+@NodeInfo(nameTemplate = "LoopExplosionKey", cycles = CYCLES_0, size = SIZE_0)
+public final class LoopExplosionKeyNode extends FloatingNode implements Canonicalizable, IterableNodeType {
+    public static final NodeClass<LoopExplosionKeyNode> TYPE = NodeClass.create(LoopExplosionKeyNode.class);
+
+    @Input ValueNode value;
+
+    public LoopExplosionKeyNode(ValueNode value) {
+        super(TYPE, value.stamp);
```
davleopo: using stamp here and not `stamp.unrestricted` could potentially have some effect for constants during canonicalization if the stamp of this node is a constant node - and the input is a constant as well. Might relate to your problem regarding delay intrinsification. Let me know if you get it working, if not ill have a closer look.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -107,12 +110,21 @@ protected class MethodScope {
         /** The kind of loop explosion to be performed during decoding. */
         public final LoopExplosionPlugin.LoopExplosionKind loopExplosion;
 
+        public LoopScope currentLoopScope;
```
davleopo, line 119: please add javadoc

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -107,12 +110,21 @@ protected class MethodScope {
...
 
         /** All merges created during loop explosion. */
         public final EconomicSet<Node> loopExplosionMerges;
 
+        /**
+         * Information about already processed loop iterations for state merging during loop
+         * explosion. Only used when {@link MethodScope#loopExplosion} is
+         * {@link jdk.graal.compiler.nodes.graphbuilderconf.LoopExplosionPlugin.LoopExplosionKind#MERGE_EXPLODE}.
+         */
+        public final EconomicMap<LoopExplosionKey, LoopExplosionState> iterationStates;
```
davleopo, line 132: can you generate some statistics about the data here - do you ever clear this? for very large compiles I see this potentially be a memory bottle neck. Can you run some statistics on average filling, etc.

Also is this keeping inner loop iterations alive for complex explosion patterns ? we have certain workloads where states have >10k node inputs - please consider such extreme outliers.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -488,53 +443,72 @@ public Node[] materializeCreatedNodes() {
         public String toString() {
             return loopDepth + "," + loopIteration + (loopBeginOrderId == -1 ? "" : "#" + loopBeginOrderId) + " triggered by " + trigger;
         }
+    }
 
+    protected static final class LoopIterationQueue {
```
davleopo, line 455: javadoc please

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -488,53 +443,72 @@ public Node[] materializeCreatedNodes() {
...
+         * is enabled.
          */
-        public boolean hasIterationsToProcess() {
-            return nextIterationFromLoopEndDuplication != null && !nextIterationFromLoopEndDuplication.isEmpty() ||
-                            nextIterationFromLoopExitDuplication != null && !nextIterationFromLoopExitDuplication.isEmpty() ||
-                            nextIterationsFromUnrolling != null && !nextIterationsFromUnrolling.isEmpty();
+        public final Deque<LoopScope> fromUnrolling;
+
+        public LoopIterationQueue(LoopExplosionPlugin.LoopExplosionKind loopExplosion) {
+            this.fromLoopExitDuplication = loopExplosion.duplicateLoopExits() || loopExplosion.mergeLoops() ? new ArrayDeque<>(2) : null;
```
davleopo, line 481: might be a good time to think about default sizes here - we probably have not visited them for a long time. are the values `2` for the initial size good ones? please also generate some statistics here

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -488,53 +443,72 @@ public Node[] materializeCreatedNodes() {
...
             }
             return null;
         }
     }
 
-    protected static class LoopExplosionState {
-        public final FrameState state;
-        public final MergeNode merge;
-        public final int hashCode;
+    protected static final class LoopExplosionKey {
```
davleopo, line 757: javadoc, also for the fields please

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -583,6 +557,23 @@ public boolean equals(Object obj) {
...
             return hashCode;
         }
+
+        @Override
+        public String toString() {
+            return bci + "@[" + values.stream().map(v -> v.toString()).collect(Collectors.joining(", ")) + "]";
+        }
+    }
+
+    protected static final class LoopExplosionState {
```
davleopo, line 841: javadoc, also for the fields please

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -852,8 +869,8 @@ protected LoopScope processNextNode(MethodScope methodScope, LoopScope loopScope
                      * loop exit of the inner loop.
                      */
                     LoopScope outerScope = loopScope.outer;
-                    int nextIterationNumber = outerScope.nextIterationFromLoopExitDuplication.isEmpty() ? outerScope.loopIteration + 1
-                                    : outerScope.nextIterationFromLoopExitDuplication.getLast().loopIteration + 1;
+                    int nextIterationNumber = outerScope.nextIterations.fromLoopExitDuplication.isEmpty() ? outerScope.loopIteration + 1
```
davleopo: minor: few places with transitive field accesses - might be good candidates for methods on the scope

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -1169,11 +1185,84 @@ protected void handleLoopExplosionBegin(MethodScope methodScope, LoopScope loopS
...
-            loopScope.iterationStates.put(explosionState, explosionState);
+            LoopExplosionKey explosionKey = createLoopExplosionKey(loopScope, frameState);
+            LoopExplosionState explosionState = new LoopExplosionState(frameState, merge, loopScope.loopDepth);
+            methodScope.iterationStates.put(explosionKey, explosionState);
         }
     }
 
+    protected List<Integer> computeIndicesForMergeKey(LoopScope loopScope, FrameState frameState) {
+        List<Integer> mergeKey = null;
+        NodeInputList<ValueNode> values = frameState.values();
```
davleopo: this is lacking support for VirtualObjectMappings in the state if escape analysis was ran before graph decoding. We have a PR in the pipeline to add that to the current merge explode logic - you should be able to adapt your code then before merging as the fix will probably land before your PR

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -1169,11 +1185,84 @@ protected void handleLoopExplosionBegin(MethodScope methodScope, LoopScope loopS
...
+        for (int i = 0; i < values.size(); i++) {
+            if (values.get(i) instanceof LoopExplosionKeyNode keyNode) {
+                if (mergeKey == null) {
+                    mergeKey = new ArrayList<>();
+                }
+                mergeKey.add(i);
+
+                for (int j = 0; j < loopScope.createdNodes.length; j++) {
```
davleopo: comment why your are going to the value

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -1169,11 +1185,84 @@ protected void handleLoopExplosionBegin(MethodScope methodScope, LoopScope loopS
...
+                    if (loopScope.createdNodes[j] == keyNode) {
+                        loopScope.createdNodes[j] = keyNode.value;
+                    }
+                }
+
+                if (loopScope.initialCreatedNodes != null) {
+                    for (int j = 0; j < loopScope.initialCreatedNodes.length; j++) {
+                        if (loopScope.initialCreatedNodes[j] == keyNode) {
```
davleopo: twice the same pattern move to utility method

davleopo:
thanks @blaumeise20  - very nice work. I will make sure this is put through some proper internal PE time benchmarking and get back to you if we see any problems. I did not spot any algorithmic problems at the moment but the state equality code can become extremely hot so we need to make sure we are not doing anything algorithmically bad for extreme outlier graph shapes.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
blaumeise20: When creating those nodes during parsing, they disappear from the graph and get replaced by their original value, meaning I never see the node during decoding. This was the only way I found to fix it.

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/truffle/test/MergeExplodeKeyTest.java
blaumeise20: They were intentional to group together bytecodes with similar semantics, but I can remove them if you want.

truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilable.java
blaumeise20: Remove as in undo change? For context I put this into the "formatting cleanup" commit.

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/truffle/test/MergeExplodeKeyTest.java
davleopo: no its fine if it helps you, I just did not understand the semantic of this new line here for example

truffle/src/com.oracle.truffle.compiler/src/com/oracle/truffle/compiler/TruffleCompilable.java
davleopo: ok - normally we want to keep changes like this for dedicated issues - they are pure code cleanups. but never mind, keep it.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
gilles-duboscq: Maybe it's because they get a constant stamp (see `Stamp.asConstant`)? Did you check whether the [suggestion](https://github.com/oracle/graal/pull/12557#discussion_r2672190157) to give them an unrestricted stamp works?

davleopo:
@blaumeise20 we triggered internal benchmarks and I did some follow up analysis on pathological patterns - looks good. No PE time benchmark regressions.

truffle/src/com.oracle.truffle.api/src/com/oracle/truffle/api/CompilerDirectives.java
chumer: Yes this is missing documentation about error conditions. What happens if it was used wrongly? Bailout?
Should be documented.

truffle/CHANGELOG.md
chumer: Would it be beneficial to the implementation to break the original support without the key? If yes, then the answer is also yes. Otherwise I would probably not bother for now, unless a reason comes up.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/substitutions/TruffleGraphBuilderPlugins.java
blaumeise20: This did in fact solve the problem.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/truffle/phases/VerifyMergeExplodeKeyDoesNotRemainPhase.java
blaumeise20: I'll do it with the LIRLowerable approach.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
blaumeise20: I just moved the field since the map was created only once, and then passed along to every loop scope, so it was never cleared in the first place. But I'll note this as a thing to check for later.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
blaumeise20: I ran the tests in `jdk.graal.compiler.truffle.test` and it seems like 80% of the time the size of the queue doesn't exceed 1 element, sometimes it has 2 and very rarely more than that. I think those defaults should be fine.

truffle/CHANGELOG.md
blaumeise20: It's probably fine for now, but definitely something to consider in the future.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -492,129 +449,406 @@ public Node[] materializeCreatedNodes() {
...
+                collectSlotPaths(slotPaths, loopScope, frameState, values.get(i), currentPath, new ArrayList<>(2));
+            }
+            if (slotPaths.isEmpty()) {
+                return null;
+            }
+            return new LoopExplosionMergeKeyFilter(slotPaths);
+        }
 
-                    h = h * 31 + value.hashCode();
+        private static void collectSlotPaths(List<int[]> slotPaths, LoopScope loopScope, FrameState frameState, ValueNode value, ArrayList<Integer> currentPath,
```
davleopo, line 570: the naming of that method is a bit adhoc and please add javadoc what is being done here (maybe even how)

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/nodes/GraphDecoder.java
```diff
@@ -492,129 +449,406 @@ public Node[] materializeCreatedNodes() {
...
+                if (loopScope.initialCreatedNodes != null) {
+                    replaceKeyMarkerWithInner(loopScope.initialCreatedNodes, keyNode);
                 }
+                keyNode.replaceAndDelete(keyNode.value);
+                return;
             }
-            this.hashCode = h;
-            this.considerVirtualStateAfterPEA = effectiveConsiderVirtualStateAfterPEA;
+            if (!(value instanceof VirtualObjectNode virtualObject) || activeObjects.contains(virtualObject)) {
+                return;
```
davleopo: all those early returns should have a comment I believe

truffle/CHANGELOG.md
```diff
@@ -61,12 +61,12 @@ This changelog summarizes major changes between Truffle versions relevant to lan
 * GR-71088 Added `CompilerDirectives.EarlyEscapeAnalysis` annotation that runs partial escape analysis early before partial evaluation enabling partial-evaluation-constant scalar replacements. 
 * GR-71870 Truffle DSL no longer supports mixed exclusive and shared inlined caches. Sharing will now be disabled if mixing was used. To resolve the new warnings it is typically necessary to use either `@Exclusive` or `@Shared` for all caches.
 * GR-71887: Bytecode DSL: Added a `ClearLocal` operation for fast clearing of local values.
-* GR-71088 Added `CompilerDirectives.EarlyEscapeAnalysis` annotation that runs partial escape analysis early before partial evaluation enabling partial-evaluation-constant scalar replacements.
```
gilles-duboscq, line 64: I suppose that this removal is not intentional

blaumeise20: It is intentional, because the line is duplicated three lines above. It probably came from some rebasing on the PR that added it.

gilles-duboscq:
Regarding the espresso changes: it looks like this might cause issues with continuations since `shouldResumeContinuation` can change from `true` to `false` and should also be considered in the key. I think we should wait for the multi-key support before using this in espresso.

Another issue seems to be that the new code doesn't detect when a `MERGE_EXPLODE` method has more than 1 top-level loop. Could you add a test for that?

PR #12574, opened 2025-11-20
Set the correct org.graalvm.version when building unchained standalones

eregon:
* It would use the SDK suite version, which does not match if BOOTSTRAP_GRAALVM is set.

### DIFF
--- a/sdk/mx.sdk/mx_sdk_vm_ng.py
+++ b/sdk/mx.sdk/mx_sdk_vm_ng.py
@@ -460,7 +460,7 @@ class NativeImageBuildTask(mx.BuildTask):
             '--parallelism=' + str(self.parallelism),
             '--link-at-build-time',
             # we want "25.0.0-dev" and not "dev" (the default used in NativeImage#prepareImageBuildArgs)
-            '-Dorg.graalvm.version={}'.format(_suite.release_version()),
+            '-Dorg.graalvm.version={}'.format(get_bootstrap_graalvm_version()),
         ] + mx_sdk_vm_impl.svm_experimental_options(experimental_build_args)
         if os.environ.get('JVMCI_VERSION_CHECK'):
             # Propagate this env var when running native image from mx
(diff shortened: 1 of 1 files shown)


gilles-duboscq:
It's a bit unclear to me what version `org.graalvm.version` refers to.
This PR seems to assume it's the version of native image (including JIT compiler and truffle support).
But polyglot and truffle also seem to think it's their version although when using a bootstrap vm that wouldn't work.

@chumer, should `PolyglotEngineImpl#getVersion` stop using `HomeFinder.getInstance().getVersion()` and use `META-INF/graalvm/org.graalvm.truffle/version` or `META-INF/graalvm/org.graalvm.polyglot/version` instead?

chumer:
> @chumer, should PolyglotEngineImpl#getVersion stop using HomeFinder.getInstance().getVersion() and use META-INF/graalvm/org.graalvm.truffle/version or META-INF/graalvm/org.graalvm.polyglot/version instead?

Yes I think it should. We didn't get to cleaning this up yet. I also want to fully eliminate the HomeFinder. Happy to take contributions.

eregon:
> It's a bit unclear to me what version `org.graalvm.version` refers to.

From the name of property it seems rather clear it's the GraalVM version (legacy-style), as you say Native Image & Compiler version.
There is also the `GRAALVM/release` file which has e.g. `GRAALVM_VERSION="25.0.1"`, which HomeFinder reads if the property is not already set, so that's further indication that the meaning of `org.graalvm.version` is the GraalVM version.

Could we merge this as a first step since it seems clearly a bug fix?

* For the case where GraalVM and Truffle versions match it doesn't matter.
* For the case they don't match, it reported the wrong version and now reports the right version.

> should `PolyglotEngineImpl#getVersion` stop using `HomeFinder.getInstance().getVersion()` and use `META-INF/graalvm/org.graalvm.truffle/version` or `META-INF/graalvm/org.graalvm.polyglot/version` instead?

I can take a look at that, sounds easy enough.

> I also want to fully eliminate the HomeFinder.

I'm not interested in doing that cleanup though.

gilles-duboscq:
I agree we can make truffle use its own version as a follow-up. I'll integrate this.

PR #12621, opened 2025-12-01
[WIP] Change object handle implementation.

asselyam:


asselyam:
@christianhaeubl Dear Dr. Häubl here is the updated code

Issue #12676, opened 2025-12-08
[Native Image] Possible race condition in InterceptJSInvokeTypeFlow causes intermittent build failures


willcohen:
### Describe the Issue

Intermittent build failure (~80% failure rate) in `native-image --tool:svm-wasm` with "Should not reach here" error. The failure is non-deterministic - running the same build multiple times produces different results.

### Using the latest version of GraalVM can resolve many issues.

- [x] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

### GraalVM Version

java 26 2026-03-17
Java(TM) SE Runtime Environment Oracle GraalVM 26-dev+13.1 (build 26+13-jvmci-b01)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 26-dev+13.1 (build 26+13-jvmci-b01, mixed mode, sharing)

### Operating System and Version

25.1.0 Darwin Kernel Version 25.1.0: Mon Oct 20 19:34:03 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T8112 arm64

### Build Command

I unfortunately cannot provide a minimal repro -- it appears to trigger after a certain level of complexity, but I can share a repro based on a working project:

```bash
git clone https://github.com/willcohen/wasmts
cd wasmts
mvn clean compile

###### Run 10 times - typically ~2 succeed, ~8 fail
for i in {1..10}; do
  native-image --tool:svm-wasm -cp "target/classes:target/lib/*" \
    net.willcohen.wasmts.API target/wasmts 2>&1 | grep -E "(Finished|Failed)"
done
```

### Expected Behavior

A successful build!

### Actual Behavior

My output this last run was:
```
Finished generating 'wasmts' in 30.9s.
Finished generating 'wasmts' in 27.4s.
Failed generating 'wasmts' after 6.6s.
Finished generating 'wasmts' in 26.7s.
Failed generating 'wasmts' after 8.0s.
Failed generating 'wasmts' after 7.1s.
Failed generating 'wasmts' after 8.8s.
Finished generating 'wasmts' in 31.5s.
Failed generating 'wasmts' after 9.3s.
Failed generating 'wasmts' after 9.5s.
```

### Steps to Reproduce

git clone https://github.com/willcohen/wasmts
cd wasmts
mvn clean compile

######## Run 10 times - typically ~2 succeed, ~8 fail
for i in {1..10}; do
[...]

PR #12677, opened 2025-12-08
Prevent race condition with web-image analysis

willcohen:
Addresses #12676. web-image usage now succeeding locally accordingly.


willcohen:
Closing in favor of #12707 @patrick96.

PR #12763, opened 2025-12-31
Add JNI validation and handle checks

asselyam:
Added a new boolean hosted option to SubstrateOptions that determines if the JNI validation is enabled
Created a new class named ValidatedJNIFunctions by copying the JNIFunctions class and modifying the copy in the following way:

-removed all methods except for the ones that are annotated with @CEntryPoint
-changed all methods that are annotated with @CEntryPoint so that they just forward the call to the functions with the same name in JNIFunctions
-modified the JNIFunctionTablesFeature so that it uses ValidatedJNIFunctions instead of JNIFunctions if your new hosted option is enabled

Implemented the validation logic to validation methods created in ValidatedJNIFunctions Class by adding logic from (https://github.com/openjdk/jdk/blob/master/src/hotspot/share/prims/jniCheck.cpp)
Also implemented validation for JNIObjectHandle


asselyam:
Dear Dr. Haübl @christianhaeubl, here are the validated jni function changes

Issue #8350, opened 2024-02-09
Support for Cosmopolitan Libc


sgammon:
**Feature request**

Please include the following information:

**Is your feature request related to a problem? Please describe.**
Currently, I must run a Native Image build for each operating system I want to target with my software.

**Describe the solution you'd like.**
GraalVM supports targeting `glibc` and `musl` as LibC backends. If GraalVM additionally supported [Cosmopolitan Libc](https://github.com/jart/cosmopolitan), GraalVM users could compile once and ship to all major operating systems, and even micro-architectures, with one binary: Linux, macOS, Windows, FreeBSD, OpenBSD, NetBSD and even BIOS.

**Describe who do you think will benefit the most.**
Firstly, GraalVM developers, because iteration on software could be much faster if native builds only needed to occur once and remained testable on multiple operating systems. Secondly, consumers of software built by GraalVM developers, because a distribution option would become available which only involves one single binary.

Consider that CI systems, local development/testing, and nearly all popular forms of software distribution (downloads, archives, things like group policy) multiply in complexity with the number of output targets needed. By reducing these targets to just one (where appropriate), complexity of distribution of GraalVM native programs drops to meet Java's build-once-run-anywhere standard.

For use cases which should remain optimized to a target architecture, the pluggable nature of GraalVM's LibC support should still allow targeting `glibc` or `musl` (for example, tailored-for-purpose server binaries).

**Describe alternatives you've considered.**

- **Containers:** Docker containers and other cross-platform virtualization methods are already supported by GraalVM, but incur a performance and tooling penalty; binaries cannot be run directly, only wrapped in a container, which must be built, downloaded, and executed with underlying OS support.

- **WASM:** Although GraalVM can execute WASM, WebAssembly isn't ready for this task because GraalVM can't yet [target WASM](https://github.com/oracle/graal/issues/3391) for `native-image`.

**Additional context.**
I tried to build a `native-image` using Cosmopolitan, but wasn't able to get far, and I think contributing this feature is outside the scope of my skillset.

**Express whether you'd like to help contributing this feature**
I would be happy to help with a stacktrace, reproducer, or with contributions in Java. [Justine Tunney](justine.lol), the author of Cosmo, is very open to contributions and there is an active community.

fernando-valdez:
Thanks for creating this feature request. I will evaluate your feature request and will let you know if I have any questions.

fernando-valdez:
Created internal ticket: GR-52220

wirthi:
[...]

PR #12865, opened 2026-01-26
[GR-73086] additional libc backend

ahgamut:
This patch adds support to build `native-image` binaries with Cosmopolitan Libc. More specifically:

- `CosmoLibC` added as a subclass of `LibCBase`
- `--libc=cosmo` added to `SubstrateOptions`
- POSIX headers and related C directives duplicated with modifications
- stub functions added to `libchelper`
- added `BooleanSupplier` classes to check for libc usage
- implementation of `-fpatchable-function-entry` flag (as in `gcc`/`clang`) to support debugging

This enables building binaries with `native-image` that can be converted to Actually Portable Executables.

1. build the `cosmocc` compiler suite with [this patch](https://github.com/ahgamut/cosmopolitan/tree/header-stubs) of Cosmopolitan Libc
2. run `make static-libs-image` using [this patch](https://github.com/elide-dev/labs-openjdk/tree/cosmo) of labs JDK
3. build `native-image` as usual with this patch on the host system
4. use the built `native-image` with `--libc=cosmo` and provide necessary flags to build binaries

Example build recipes for Github Actions are given [here](https://github.com/elide-dev/envmap-sample).


sgammon:
relates to oracle/graal#8350

oubidar-Abderrahim:
Thank you for your contribution @ahgamut 
@wirthi could you please assign/review this PR?

olpaw:
@sgammon thank you for this very interesting PR. We’re currently investigating whether merging it is feasible and practical. As far as I can tell, adding Cosmopolitan support in NI depends on changes in other projects that we don’t control. Is Justine Tunney willing to integrate your Cosmo changes as-is? From what I see, these are largely stub implementations of C library functionality that Cosmo is currently missing, but that LabsJDK (and the NI build) depends on at build time. Ideally, there would be an official Cosmopolitan release that includes your changes, which we could then build on. 

Regarding the LabsJDK changes, I have another concern: they further increase our delta to OpenJDK and therefore increase our maintenance effort. Would you be willing to minimize these changes as much as possible and help resolve any merge conflicts that might arise in the future? 

Also, could you elaborate on the need for the AArch64-specific register changes that Cosmo seems to require? Where can I read more about the rationale behind those changes? 

Finally, if we integrate Cosmo support, we’ll likely need a setup to safeguard against regressions—e.g., a GitHub Actions–based CI job would be appreciated.

ahgamut:
> Is Justine Tunney willing to integrate your Cosmo changes as-is?

At the moment, the Cosmo changes are just [headers and stubs](https://github.com/ahgamut/cosmopolitan/tree/header-stubs) for well known functions, so I would hope yes. I've submitted a PR here: https://github.com/jart/cosmopolitan/pull/1474

> Regarding the LabsJDK changes, I have another concern: they further increase our delta to OpenJDK and therefore increase our maintenance effort. Would you be willing to minimize these changes as much as possible?

The labs JDK changes as far as I remember are mostly just `ifdef`s to account for Cosmopolitan Libc. The only major change is with respect to the `r28` register on `aarch64`, which brings us to

> Also, could you elaborate on the need for the AArch64-specific register changes that Cosmo seems to require? Where can I read more about the rationale behind those changes?

on `aarch64`, `native-image` uses `r28` to note the current isolate/thread. However, Cosmopolitan Libc currently uses [`r28` to remember thread-local storage](https://github.com/jart/cosmopolitan/blob/9ba2b50805cae693631c15aed8576c8f6c6f6faa/ape/specification.md?plain=1#L330), so there was a clash. I just chose to use `r25` because it is also callee-saved.

> Finally, if we integrate Cosmo support, we’ll likely need a setup to safeguard against regressions—e.g., a GitHub Actions–based CI job would be appreciated.

We have example CI actions [here](https://github.com/elide-dev/envmap-sample). I think in terms of CI we'd just need an action that creates `static-libs` for cosmo libc, like labs JDK currently provides for `musl`, and then a simplified version of [this build script](https://github.com/elide-dev/envmap-sample/blob/main/.github/workflows/job.fixed-envmap-example.yml).

olpaw:
> However, Cosmopolitan Libc currently uses [r28 to remember thread-local storage](https://github.com/jart/cosmopolitan/blob/9ba2b50805cae693631c15aed8576c8f6c6f6faa/ape/specification.md?plain=1#L330), so there was a clash. I just chose to use r25 because it is also callee-saved.
 
Sounds like we could possibly make this the default for all aarch64 images (not just cosmo). @lewurm do you agree?

lewurm:
> > However, Cosmopolitan Libc currently uses [r28 to remember thread-local storage](https://github.com/jart/cosmopolitan/blob/9ba2b50805cae693631c15aed8576c8f6c6f6faa/ape/specification.md?plain=1#L330), so there was a clash. I just chose to use r25 because it is also callee-saved.
> 
> Sounds like we could possibly make this the default for all aarch64 images (not just cosmo). @lewurm do you agree?

We have a tendency to choose the same register assignment as HotSpot to ease sharing of e.g. LIR instructions. After a quick look I don't see an unclean usages of the thread registers however, so it should be fine (and if not, it's certainly fixable).


> The labs JDK changes as far as I remember are mostly just `ifdef`s to account for Cosmopolitan Libc. The only major change is with respect to the `r28` register on `aarch64`, which brings us to

I'm confused about that (context https://github.com/elide-dev/labs-openjdk/commit/7042bab8ffb3a238cce9d330c580424a992edf80 ): Do you need to change it so that you can _build_ `libjvm` with the cosmo toolchain?  We still need a working labsjdk build, so I'm afraid to make that actually work on any AArch64 platform that will require some more work (we can discuss this on a labs JDK PR however; can you create one?).

olpaw:
Thanks for https://github.com/graalvm/labs-openjdk/pull/28

I left my comments: https://github.com/graalvm/labs-openjdk/pull/28#issuecomment-3870900290

olpaw:
@ahgamut regarding

> 1. build the cosmocc compiler suite with [this patch](https://github.com/ahgamut/cosmopolitan/tree/header-stubs) of Cosmopolitan Libc

I was able to do that. But how to proceed from here?

I see that now the `o` folder got created and a lot of tests ran. But how do I use the result of this step to complete the next step:

> 2. run make static-libs-image using [this patch](https://github.com/elide-dev/labs-openjdk/tree/cosmo) of labs JDK

I need more detailed instructions than what you provide in top comment https://github.com/oracle/graal/pull/12865#issue-3854470523.

- [ ] Please provide **precise instructions** for all the steps you mentioned above.

olpaw:
> At the moment, the Cosmo changes are just [headers and stubs](https://github.com/ahgamut/cosmopolitan/tree/header-stubs) for well known functions, so I would hope yes. I've submitted a PR here: https://github.com/jart/cosmopolitan/pull/1474

Sent a ping on the Cosmo PR https://github.com/jart/cosmopolitan/pull/1474#issuecomment-3958243748

olpaw:
> 2. run make static-libs-image using [this patch](https://github.com/elide-dev/labs-openjdk/tree/cosmo) of labs JDK

Using
```
export CC=/path/to/cosmo-repo/tool/cosmocc/bin/x86_64-unknown-cosmo-cc
export CXX=/path/to/cosmo-repo/tool/cosmocc/bin/x86_64-unknown-cosmo-c++
```
and
```
bash configure --without-{alsa,cups,fontconfig,x} --with-freetype=bundled
```
gives me
```
...
checking for jni library path... default
configure: Using default toolchain gcc (GNU Compiler Collection)
configure: Will use user supplied compiler CC=/home/pwoegere/OLabs/issues/GR-73086/cosmopolitan/tool/cosmocc/bin/x86_64-unknown-cosmo-cc
checking resolved symbolic links for CC... /home/pwoegere/OLabs/issues/GR-73086/cosmopolitan/tool/cosmocc/bin/cosmocross
configure: Using gcc C compiler version 14.1.0 [x86_64-unknown-cosmo-cc (GCC) 14.1.0]
checking whether the C compiler works... no
configure: error: in '/home/pwoegere/OLabs/issues/GR-73086/labs-openjdk':
configure: error: C compiler cannot create executables
See 'config.log' for more details
configure exiting with result code 77
```

In other words, how does this work in detail? See https://github.com/oracle/graal/pull/12865#issuecomment-3958198869

olpaw (changes requested):
See comments

ahgamut:
Updated the commits to avoid merge conflicts. 

When opening this PR, I provided a [repo](https://github.com/elide-dev/envmap-sample) containing all scripts and Github Actions used to build example binaries (the [`Envmap` example](https://www.graalvm.org/latest/reference-manual/native-image/guides/build-static-executables/), and the [`google-java-format` JAR](https://github.com/google/google-java-format/)) via this patch. 

Both example binaries appear to run without error even after the latest rebase [here](https://github.com/elide-dev/envmap-sample/actions/runs/22571425339) and [here](https://github.com/elide-dev/envmap-sample/actions/runs/22571846643). I will re-run the actions in a while to confirm.

ahgamut:
@olpaw I have described more details of the build steps below.

---

> 1) build the cosmocc compiler suite with [this patch](https://github.com/ahgamut/cosmopolitan/tree/header-stubs) of Cosmopolitan Libc

[this comment](https://github.com/oracle/graal/pull/12865#issuecomment-3958198869) mentions cosmocc can be built using the above branch.

However, I found building the labs JDK required some additional headers and static libraries (at least `libz`, `libpng`, `libjpeg`, and `libfreetype`) that are not provided alongside cosmocc by default. I build these libraries using my [`superconfigure`](https://github.com/ahgamut/superconfigure) repo, which provides scaffolding for `autotools`-style builds. 

> Similar to the build script attempted [here](https://github.com/oracle/graal/pull/12865#issuecomment-3959261255), it may be possible to build the static libs without having to rely on the `superconfigure` repo, by using the `bundled` option. I have not tested that build process yet. 

Relevant scripts: [`patch-cosmo-repo.bash`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/patch-cosmo-repo.bash#L32), [`build-cosmocc.bash`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-cosmocc.bash), and [`build-deps-labs.bash`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-deps-labs.bash).

> 2) run make static-libs-image using [this patch](https://github.com/elide-dev/labs-openjdk/tree/cosmo) of labs JDK

After collecting `cosmocc` and headers/libraries for (`libz`, `libjpeg`, `libpng`, `freetype`), I attempt to build the `static-libs-image` target with the labs JDK, using the patch under discussion at https://github.com/graalvm/labs-openjdk/pull/28. The goal is to have static libs similar to what is currently provided for `musl` in the [labs JDK Releases page](https://github.com/graalvm/labs-openjdk/releases/tag/jvmci-25.1-b15). 

I provide some additional arguments at the configuration stage (notably `--disable-precompiled-headers` and `--enable-dtrace=no`) to simpify the build.  After calling `make static-libs-image` I just collect the relevant archives from the build (along with `libz.a` from earlier) for use when building with `native-image`.

Relevant scripts: [`build-labs-jdk.bash`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-labs-jdk.bash). 

> [!NOTE]
> A detail I failed to mention earlier that is provided in the repo: static libraries within `native-image` (`liblibchelper.a`, the `libjvm.a` stub, and `libsvm_container.a`) also need to be built with cosmopolitan libc. I assume it is possible to build these via a specific target in the `mx` tool (as is done for `glibc` and `musl`), but for now I just use the `--alt-cc` option with `mx`.
>
> Relevant scripts: [`build-graal-helpers.bash`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-graal-helpers.bash#L47)
    
> 3) build native-image as usual with this patch on the host system

If I understand correctly, this step currently does not require cosmocc or any of the previous build artifacts. When writing the patch I used `mx --env ni-ce build` on my host system so I could have the diagnostic agent for testing.

Relevant scripts: [`build-native-image.bash`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-native-image.bash).

> 4) use the built native-image with --libc=cosmo and provide necessary flags to build binaries

Once `native-image` has been built with this patch, and cosmocc along with the other build artifacts have been collected, we can use all of them together to build `native-image` binaries with Cosmopolitan Libc.

For now, the `liblibchelper.a`, the `libjvm.a` stub, and `libsvm_container.a` built using `cosmocc` from earlier need to be copied to a [specific folder, namely `lib/svm/clibraries/linux-$ARCH/cosmo`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-envmap-example.bash#L52) before running a build. Once those files are in the correct spot, we can build static binaries via `native-image --libc=cosmo` and required flags.

Relevant scripts: [`build-envmap-example.bash`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-envmap-example.bash), [`build-google-java-format.bash`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-google-java-format.bash).

olpaw:
Thanks for the updates. I will look into it in the next days (hopefully).

olpaw:
Thanks for the info in https://github.com/oracle/graal/pull/12865#issuecomment-3983811544. I'm now able to successfully build the static libs with cosmocc. Next step, using these artifacts for review and test this PR.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoAccess.java
```diff
@@ -255,19 +256,23 @@ public static UnsignedWord getNativeMetadataSize(CodeInfo info) {
...
     }
 
     @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
     public static long relativeIP(CodeInfo info, CodePointer ip) {
         assert contains(info, ip);
         CodeInfoImpl impl = cast(info);
-        UnsignedWord baseOffset = ((UnsignedWord) ip).subtract((UnsignedWord) impl.getCodeStart());
+        // TODO (ahgamut): move nopsBeforeEntry to getRelativeIPOffset?
+        UnsignedWord actualCodeStart = ((UnsignedWord) cast(info).getCodeStart()).subtract(cast(info).getNopsBeforeEntry());
+        UnsignedWord baseOffset = ((UnsignedWord) ip).subtract((UnsignedWord) actualCodeStart);
```
olpaw:
```suggestion
        UnsignedWord baseOffset = ((UnsignedWord) ip).subtract(actualCodeStart);
```
That cast is redundant.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoAccess.java
```diff
@@ -183,7 +183,8 @@ public static boolean isAliveState(int state) {
     /** @see CodeInfoImpl#getCodeStart */
     @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
     public static CodePointer getCodeStart(CodeInfo info) {
-        return cast(info).getCodeStart();
+        UnsignedWord actualCodeStart = ((UnsignedWord) cast(info).getCodeStart()).subtract(cast(info).getNopsBeforeEntry());
```
olpaw: You cannot just arbitrarily modify CodeInfoAccess to make it suitable for Cosmo.

This has to be done in a way that does not modify behavior if `--libc=cosmo` is not used.

e.g. define a `getActualCodeStart(CodeInfoImpl impl) { ... }` that behaves different depending on whether this builds for Cosmo or not.

olpaw (changes requested):
The PR has to be written so that when NI does **not** build with `--libc=cosmo` NI works exactly as our master.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/cosmo/package-info.java
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+@Platforms({Platform.LINUX.class})
```
olpaw, line 26: Ok for now, but eventually you would want cosmo support to be available on all platforms. In the case of cosmo there is really no reason why this should not work.

olpaw:
If I manually copy the cosmo-built static jdk libs into `$GRAALVM_HOME/lib/static/linux-amd64/cosmo` I get a bit further and then run into another issue:
```
[2/8] Performing analysis...  [******]                                                                  (2.9s @ 0.45GiB)
    3,330 types,   3,762 fields, and  15,422 methods found reachable
    1,109 types,      36 fields, and     329 methods registered for reflection
       55 resource accesses registered with 171.30KiB total size

Error: Invalid library path /home/pwoegere/OLabs/issues/GR-73086/graal/sdk/mxbuild/linux-amd64/GRAALVM_0C98F50C75_JAVA25/graalvm-0c98f50c75-java25-25.1.0-dev/lib/svm/clibraries/linux-amd64/cosmo
Caused by: java.nio.file.NoSuchFileException: /home/pwoegere/OLabs/issues/GR-73086/graal/sdk/mxbuild/linux-amd64/GRAALVM_0C98F50C75_JAVA25/graalvm-0c98f50c75-java25-25.1.0-dev/lib/svm/clibraries/linux-amd64/cosmo
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.asIOException(UnixException.java:115)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newDirectoryStream(UnixFileSystemProvider.java:456)
	at java.base/java.nio.file.Files.newDirectoryStream(Files.java:445)
	at java.base/java.nio.file.Files.list(Files.java:3422)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.c.NativeLibraries.getAllStaticLibs(NativeLibraries.java:509)
...
```
This does not surprise me because you do not seem to have a cosmo variant defined in
```
"multitarget": {
    "libc": ["glibc", "musl", "default"],
},
```
of `com.oracle.svm.native.libchelper`, `com.oracle.svm.native.jvm.posix` `com.oracle.svm.native.libcontainer` in `substratevm/mx.substratevm/suite.py`

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoAccess.java
ahgamut: Agreed. Apologies for not catching the typo.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoAccess.java
ahgamut: > This has to be done in a way that does not modify behavior if --libc=cosmo is not used.

Yes, and I thought it will not modify behavior as long as `-H:PatchableFunctionEntry` is the default of `0,0`. 

My belief was that `PatchableFunctionEntry` can be applied to builds regardless of `--libc=cosmo`, but if not, we can keep it specific to that for now.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/cosmo/package-info.java
ahgamut: Yes, there is some local work where I tried this, if you like we can add that. I thought the first patch should just get builds in Linux working, as I have not tried building on the other platforms as much.

ahgamut:
@olpaw Thanks for the review! Apologies for the mistakes, I am still not fully familiar with the codebase. I will rebase my patch against the most recent commit and attempt to fix these soon.

Regarding the build issue mentioned in https://github.com/oracle/graal/pull/12865#issuecomment-4136317240 I am embarrassed I did not catch this issue myself. Can you tell me what build commands I can run to trigger this, with reference to the [`build-envmap-example.bash`](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-envmap-example.bash) script I am using? 

Edit: the build script I have copies the libs to `$GRAALVM_HOME/lib/svm/clibraries/linux-x86_64/cosmo` (see [here](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-envmap-example.bash#L52)) and not to `$GRAALVM_HOME/lib/static/linux-amd64/cosmo`.
Perhaps the error occurs due to something else. Let me rebase the patch and confirm.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoAccess.java
ahgamut: Fixed after rebasing the PR.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/code/CodeInfoAccess.java
ahgamut: Fixed after rebasing the PR. 

Now providing non-default `-H:PatchableFunctionEntry` will only be allowed with `--libc=cosmo`. If `PatchableFunctionEntry` is the default of `0,0`. It should build NI as usual.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/cosmo/package-info.java
ahgamut: I would prefer to leave this as-is for now, and submit a separate PR later.

ahgamut:
I've rebased this PR and added commits based on the reviews. The build steps should still be the same.

- redundant cast has been fixed
- a `getActualCodeStart` function to avoid repeating the calculation
- `getActualCodeStart` will differ from `getCodeStart` only if `--libc=cosmo` and nonzero `-H:PatchableFunctionEntry` is provided

olpaw:
> @olpaw Thanks for the review! Apologies for the mistakes, I am still not fully familiar with the codebase. I will rebase my patch against the most recent commit and attempt to fix these soon.
> 
> Regarding the build issue mentioned in [#12865 (comment)](https://github.com/oracle/graal/pull/12865#issuecomment-4136317240) I am embarrassed I did not catch this issue myself. Can you tell me what build commands I can run to trigger this

I just ran
```
mx native-image --verbose --native-compiler-path=/home/pwoegere/OLabs/issues/GR-73086/cosmocc-jart-master-bundle/bin/x86_64-unknown-cosmo-cc --libc=cosmo -jar /home/pwoegere/OLabs/hello.jar
```
from `graal/substratevm`.

olpaw:
> I've rebased this PR and added commits based on the reviews. The build steps should still be the same.
> 
>     * redundant cast has been fixed
> 
>     * a `getActualCodeStart` function to avoid repeating the calculation
> 
>     * `getActualCodeStart` will differ from `getCodeStart` only if `--libc=cosmo` and nonzero `-H:PatchableFunctionEntry` is provided

Thanks! I will have another look soon.

ahgamut:
Rebased to remove minimal conflicts. 

Now using https://github.com/oracle/graal/pull/12865/commits/5d1f08970d60d043e3bef0f59df6b50a02be282a we can just pass the location of all necessary native libs via `CLibraryPath` when building with `--libc=cosmo`.

Therefore @olpaw I think [this build command](https://github.com/oracle/graal/pull/12865#issuecomment-4198685936) should be similar to [the example here](https://github.com/elide-dev/envmap-sample/blob/main/scripts/build-envmap-example.bash#L59)

Perhaps something like:

```sh
export LABSJDK_LIBS="/path/to/labsjdk/cosmo-x86_64/static-libs"
export GRAAL_HELPER_LIBS="/path/to/graal/cosmo-x86_64-libs" # contains libclibchelper.a

export COSMOCC_BIN="/home/pwoegere/OLabs/issues/GR-73086/cosmocc-jart-master-bundle/bin/"
mx native-image --verbose \
    -march=compatibility \
    --static \
    --native-compiler-path="$COSMOCC_BIN/x86_64-unknown-cosmo-cc" \
    -H:CLibraryPath="$LABSJDK_LIBS,$GRAAL_HELPER_LIBS" \
    -H:PatchableFunctionEntry=18,16 \ # 7,6 if on aarch64
    --libc=cosmo \
    -jar /home/pwoegere/OLabs/hello.jar \
    -o hello.x86_64
```

PR #12905, opened 2026-02-03
[GR-73283] Fix JCMD Thread.dump_to_file

roberttoyonaga:
If JCMD issues `jcmd <PID> Thread.dump_to_file` the VM crashes with `Fatal error reported via JNI: JVM_CreateThreadSnapshot called:  Unimplemented`

This is due to the fact that the underlying implementation has changed: https://github.com/openjdk/jdk/pull/25425
Now (as of JDK 25), the `jdk.internal.vm.ThreadDumper` calls down to Hotspot to populate a `ThreadSnapshot` instance with thread data. We don't support this call yet. Before, all the thread data needed for dumping was obtained within Java code. 

This is a temporary solution. A more permanent solution would be to substitute `ThreadSnapshot.create` and implement the code that populates `ThreadSnapshot` with thread data in SubstrateVM.  I have opened a GitHub issue for that: https://github.com/oracle/graal/issues/12906

[JDK 25+26 ](https://github.com/openjdk/jdk/blob/jdk-25%2B26/src/java.base/share/classes/jdk/internal/vm/ThreadDumper.java#L181) was when the change was introduced

### DIFF
--- a/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/dcmd/Target_jdk_internal_vm_ThreadSnapshot.java
+++ b/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/dcmd/Target_jdk_internal_vm_ThreadSnapshot.java
@@ -79,7 +79,7 @@ final class Target_jdk_internal_vm_ThreadSnapshot {
         // Setting the blocker info in cases other than PARK_BLOCKER is non-trivial.
         Object blocker = LockSupport.getBlocker(thread);
         if (blocker != null && !(blocker instanceof JavaMonitor) && !(blocker instanceof JavaMonitorQueuedSynchronizer.JavaMonitorConditionObject)) {
-            snapshot.blockerTypeOrdinal = ((Enum<?>) (Object) Target_jdk_internal_vm_ThreadSnapshot_BlockerLockType.parkBlocker).ordinal();
+            snapshot.blockerTypeOrdinal = SubstrateUtil.cast(Target_jdk_internal_vm_ThreadSnapshot_BlockerLockType.PARK_BLOCKER, Enum.class).ordinal();
             snapshot.blockerObject = blocker;
         }
 
@@ -94,6 +94,8 @@ final class Target_jdk_internal_vm_ThreadSnapshot {
 
 @TargetClass(className = "jdk.internal.vm.ThreadSnapshot", innerClass = "BlockerLockType")
 final class Target_jdk_internal_vm_ThreadSnapshot_BlockerLockType {
+    // Checkstyle: stop
     @Alias //
-    static Target_jdk_internal_vm_ThreadSnapshot_BlockerLockType parkBlocker;
+    static Target_jdk_internal_vm_ThreadSnapshot_BlockerLockType PARK_BLOCKER;
+    // Checkstyle: resume
 }


oubidar-Abderrahim:
Thank you @roberttoyonaga for this PR
@christianhaeubl could you please review this?

roberttoyonaga:
This should be backported to GraalVM for JDK 25 as well

roberttoyonaga:
Just pushed a commit to add some more tests. Hopefully this will help us catch such problems earlier in the future.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/dcmd/DCmdFeature.java
```diff
@@ -61,7 +61,6 @@ public void afterRegistration(AfterRegistrationAccess access) {
             dcmdSupport.registerCommand(new JfrDumpDCmd());
         }
 
-        dcmdSupport.registerCommand(new ThreadDumpToFileDCmd());
```
christianhaeubl, line 64: Thanks for the PR.

I don't think that we should disable this command as that will result in a different error message (i.e., command not supported). Instead, I think it would be better to implement a minimal fix that substitutes the native method and only fills in the stack trace and the thread name.

roberttoyonaga: Yes okay I can work on that. 
By removing this here, my original intention was to not advertise that this command is supported  when the user executes `jcmd <PID> help` or `jcmd <PID>`.

roberttoyonaga: I've added a few more commits that substitute the call down to Hotspot and partially fill in the thread information.  I did not implement getting the owned locks or tracing the owners of blocking monitors.  I've added some more tests too. Please have another look when you have time.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/dcmd/HelpDCmd.java
```diff
@@ -41,7 +41,8 @@ public HelpDCmd() {
         super("help", "For more information about a specific command use 'help <command>'. With no argument this will show a list of available commands. 'help -all' will show help for all commands.",
                         Impact.Low, new DCmdOption<?>[]{COMMAND_NAME}, new DCmdOption<?>[]{PRINT_ALL},
                         new String[]{
-                                        "$ jcmd <pid> help Thread.dump_to_file"
+                                        "$ jcmd <pid> help " +
+                                                        "JFR.start"
```
christianhaeubl, line 45: I changed this back to `Thread.dump_to_file`. JFR support (and therefore also the command `JFR.start`) is not always included.

christianhaeubl:
Thanks, I started integrating this PR.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/jcmd/ThreadDumpToFileTest.java
```diff
@@ -0,0 +1,96 @@
...
+        parked.await();
+        while (!parkedThread.getState().equals(Thread.State.WAITING)) {
+            Thread.sleep(10);
+        }
+
+        String dump = getThreadDump();
+        LockSupport.unpark(parkedThread);
+        parkedThread.join();
+
+        assertTrue("Dump should contain the blocked thread's name", dump.contains("test-park-blocker-thread"));
```
christianhaeubl, line 72: I removed this test class and added the logic to `JCmdTest`.

substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/dcmd/Target_jdk_internal_vm_ThreadSnapshot.java
```diff
@@ -0,0 +1,101 @@
...
+    @Alias //
+    private Object blockerObject;
+
+    @Alias //
+    private Target_jdk_internal_vm_ThreadSnapshot() {
+    }
+
+    @Substitute
+    private static Target_jdk_internal_vm_ThreadSnapshot create(Thread thread) {
+        Target_jdk_internal_vm_ThreadSnapshot snapshot = new Target_jdk_internal_vm_ThreadSnapshot();
```
christianhaeubl, line 69: I had to change some of this logic to a VM operation. Otherwise, the values may be inconsistent.

christianhaeubl (commented):
I did a few cleanups and fixes on top, see the comments above and https://github.com/oracle/graal/pull/12926. Please have a look.

Besides that, I also squashed all commits. Please let know if that is fine for you.

roberttoyonaga:
https://github.com/oracle/graal/pull/12926 looks good to me. Thank you

christianhaeubl:
This was merged to master. I also requested backporting to GraalVM 25.0 (just for reference, the backporting issue is GR-73332).

roberttoyonaga:
Thank you!

PR #12924, opened 2026-02-09
[GR-73213] Provide Steps How to Build Base GraalVM CE from Source for External Users

graalvmbot:


olyagpl:
@simonis, would you be interested in reviewing this PR? We would appreciate your feedback.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
+
+1. Create a workspace directory, and enter it:
+    ```bash
+    mkdir workspace; cd workspace
+    ```
+2. Clone the `mx` repository:
+    ```bash
+    git clone https://github.com/graalvm/mx
+    ```
+3. Add `mx` to your `PATH` environment variable:
```
simonis: You should probably mention that the required `mx` version should correspond to one specified by `mx_version` in the file `common.json` of the Graal repository. The following code snippet proved useful for synchronizing to the correct `mx` version (assuming you have both the `graal` and the `mx` repository in the same directory):

```bash
$ grep mx_version /path/to/graal/common.json
  "mx_version": "7.68.10",
$ git clone https://github.com/graalvm/mx
$ cd mx/
$ git checkout 7.68.10
```

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
-```
+- Python version as defined in the [Graal CI](https://github.com/oracle/graal/blob/master/ci/common.jsonnet#L147) (currently Python 3.8). Install the required version and set the `MX_PYTHON` environment variable.
+- Developer packages for your operating system are required. The following versions are tested by the Oracle-internal Graal CI but newer versions might also work.
+  - **Linux**: [`devtoolset` 12](https://docs.redhat.com/en/documentation/red_hat_developer_toolset/12/html-single/user_guide/index) which includes
+    - `gcc` 12.2 (the minimum accepted version is 10.0; version 14.2 is also known to compile successfully)
+    - `make` 4.3
+    - `binutils` 2.36
+  - **macOS**: Sonoma and XCode 15
+  - **Windows**: Visual Studio 2022 17.13
+- OpenJDK or Oracle JDK on your system to be used as the "boot" image. Say you want to build a GraalVM version 25, you need to set `JAVA_HOME` to OpenJDK or Oracle JDK 25 or 24. You can install it with SDKMAN!.
```
simonis: This is not accurate. You only need an OpenJDK as boot JDK for building the LabsJDK. I'd remove this point here and add a corresponding point to the section which describes how to build a LabsJDK (mentioning that the boot JDK version must be not lower than LabsJDK version - 1).
If somebody opts to download LabsJDK binaries, OpenJDK isn't required at all.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
+# gcc toolset 12
+source /opt/rh/gcc-toolset-12/enable
 ```
 
-## Registering custom components
-Suites can register new, custom components calling`mx_sdk.register_graalvm_component()`.
+### Related Documentation
+
+- [Static, Dynamic, and Versioned Imports in mx](https://github.com/graalvm/mx/blob/master/docs/dynamic-imports.md)
+- [Building the JDK](https://github.com/graalvm/labs-openjdk/blob/master/doc/building.md)
```
simonis: Better write 'Building the LabsJDK' because OpenDJK has it's own build Readme.

vm/README.md
simonis: It would be nice to add a section on creating the IDE configuration. Something similar to [what I wrote here](https://github.com/simonis/GraalJsTest/blob/main/Notes.md#ide-support). Feel free to copy-paste from there if you think it is useful :)

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
+    The build will be available at:
+    ```bash
+    mx --env ce graalvm-home
+    ```
+    The uncompressed archive of the build will be available at:
+    ```bash
+    mx --env ce paths $(mx --env ce graalvm-dist-name)
+    ```
+    Since additional processing may occur during archiving, you should use the uncompressed archive.
+
```
simonis, line 123: I don't like to clutter my source repository with build artifacts. To avoid this, you can set the `MX_ALT_OUTPUT_ROOT` environment variable to redirect all build artifact to that directory. It would be nice to document that here (or in the FAQs below) as well.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
+    git checkout `cat /path/to/graal/common.json | grep '"labsjdk-ce-latest"' | sed 's:.*\(jvmci-[^"]*\)".*:\1:g'`
+    ```
+    For example, at the time of writing this guide, the tag is `jvmci-25.1-b14`:
+    ```
+    git status
+    HEAD detached at jvmci-25.1-b14
+    ```
+3. Build a JVMCI-enabled JDK. First configure the build and then build the image with `make`:
+    ```bash
+    bash configure --disable-dtrace
```
simonis: I don't think `--disable-dtrace` is required (at least I tried without and it worked) and it is confusing. So either explain why it is required or remove it.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
+    ```bash
+    git clone https://github.com/oracle/graal.git
+    ```
+ 2. Enter the _graal/vm_ directory:
+    ```bash
+    cd graal/vm
+    ```
+3. Build the base GraalVM:
+    ```bash
+    mx --env ce --sources=sdk:GRAAL_SDK,truffle:TRUFFLE_API,compiler:GRAAL,substratevm:SVM --debuginfo-dists --base-jdk-info=labsjdk:25 build --targets=GRAALVM
```
simonis: This seems overly complex and incomprehensible. I think it should simply read: `mx --env ce build`. That works, and it builds a complete GraalVM CE distro **including** the `native-image` tool. You can then also remove the previous statements which claim that `native-image` won't be build (i.e. "*(without the `native-image` tool)*").

I think it is fine to document additional, more sophisticated build command lines here, after the basic `mx --env ce build`, but then please describe each additional argument and explain what it does.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
 
-The VM suite allows you to build custom GraalVM distributions, as well as installable components.
-It defines a base GraalVM distribution that contains the JVMCI-enabled JDK, the GraalVM SDK, Truffle, and the GraalVM component installer.
-More components are added by dynamically importing additional suites.
-This can be done either by:
-1. running `mx --dynamicimports <suite...> build`
-2. setting the `DEFAULT_DYNAMIC_IMPORTS` or `DYNAMIC_IMPORTS` environment variables before running `mx build`
-3. running `mx --env <env file> build`
+The VM suite provides the tooling required to build GraalVM from source.
+It defines a base GraalVM package that contains the GraalVM SDK, the Graal JIT compiler, SubstrateVM (without the `native-image` tool), and the Truffle framework.
```
simonis: Remove "*(without the `native-image` tool)*". See below for more details.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
 
-This will show a list of components, launchers and libraries to be built.
-It is recommended to verify this output before running `build`.
+## Build GraalVM from Source
 
-### Example: build the base GraalVM CE image
-The base GraalVM CE image includes:
+GraalVM Community Edition (CE) is an open-source project based on OpenJDK.
+The base package includes:
 - SubstrateVM (without the `native-image` tool)
```
simonis: Remove "*(without the `native-image` tool)*". See below for more details.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
+`mx` works with suites. A **suite** is an `mx` construct that defines:
+  - what are the dependencies of a project
+  - where are the source files
+  - how they are built
+  - what are the results of the build process
+  - how to execute tests and programs
+
+Some Graal projects contain only one suite, while others contain multiple suites.
+
+_How do you check how many suites a Graal project repository contains?_
```
simonis: Make the question a further subsection (i.e. `#### How do you..`) instead of setting it in italic.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
+Search for files defining a suite named _suite.py_ in the root directory, using `mx` or `find`:
```
simonis: Should be `` `suite.py` `` instead of `` _suite.py_ ``.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
+You should see several suites.
 
-3.
-```bash
-$ export DEFAULT_DYNAMIC_IMPORTS=/substratevm,/tools,/sulong,/graal-nodejs
-$ export EXCLUDE_COMPONENTS=nju,nic,ni,nil,llp
-$ mx build
-```
-or:
+_What is a primary suite?_
```
simonis: Make the question a further subsection (i.e. #### What is..) instead of setting it in italic.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
-                                This can be a comma-separated list of disabled launchers or `true` to disable all native launchers.
-```
-And the following environment variables:
-```
-  NATIVE_IMAGES                 Same as '--native-images'
-  FORCE_BASH_LAUNCHERS          Same as '--force-bash-launchers'
-```
-
-### Example: force bash launchers
+Note that the build dependencies of each component are specified in the README file of the corresponding repository.
```
simonis: I think that's wrong. As far as I know the dependencies are defined in the corresponding `suite.py` files (but please check with somebody who really knows :)

vm/README.md
JohnTortugo: May be also worth adding that you can use `mx fetch-jdk` to download a base JDK for Graal.

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
 
-3.
-```bash
-$ export DEFAULT_DYNAMIC_IMPORTS=/substratevm,/tools,/sulong,/graal-nodejs
-$ export EXCLUDE_COMPONENTS=nju,nic,ni,nil,llp
-$ mx build
-```
-or:
+_What is a primary suite?_
+A **primary suite** is the one in the current directory, or the one passed as an `mx` argument.
```
JohnTortugo, line 164: Would be nice to have a definition of what a "suite" means. It's not true that the "current directory" is a suite for all directories (.e.g, docs/)

vm/README.md
```diff
@@ -1,148 +1,200 @@
...
-## Versioned dynamic imports
-Dynamic imports typically require the user to locate and clone the dynamically imported suites.
-There is also no indication of which version of those suites would work.
-To avoid this issue, the `vm` suite uses "versioned dynamic imports".
+Below is an example of a minimal configuration script for Linux:
 
-The `mx.vm/suite.py` file contains references to all the suites that might be imported to compose a GraalVM.
-Unlike usual suite imports, they are marked as `dynamic`, which means they are only considered if they are part of the dynamically imported suites.
-However, when they are included, they have URLs and versions which allow mx to automatically clone the correct version.
+```bash
```
JohnTortugo: I don't like this script at all. IMO, It seems too specific to be part of a documentation. I'm not onboard that you even need a "Minimal configuration script example", this should be clear from the rest of the text.

vm/README.md
olyagpl: Good point!

vm/README.md
olyagpl: That's a valuable input. I tested building on OpenJDK, then on Oracle JDK, but yes, OpenJDK would be sufficient as long as the version matches or isn't lower than -1. Also, moving this precondition down makes sense.
The current guidelines are basically derived from the buildspec  for a CI job that makes a GraalVM CE build.

vm/README.md
olyagpl: Will be done!

vm/README.md
olyagpl: Can the default be different on Oracle Linux? 
Without that flag I hit this: 
```
make/hotspot/src/hotspot/share/runtime/java.cpp:581:(.text+0xe6c): undefined reference to `__dtrace_hotspot___vm__shutdown'
/opt/rh/gcc-toolset-14/root/usr/libexec/gcc/x86_64-redhat-linux/14/ld: make/hotspot/src/hotspot/share/runtime/java.cpp:581:(.text+0xf98): undefined reference to `__dtrace_hotspot___vm__shutdown'
   ... (rest of output omitted)
   ```
 However, in the configure.log it's listed:
Screenshot 2026-02-11 at 23 21 26

vm/README.md
olyagpl: I kind of agree that this example doesn't quite fit into FAQs and is task-specific.

vm/README.md
olyagpl: For example, testing on mac, -`-disable-dtrace` wasn't required. I'm removing this flag.

vm/README.md
olyagpl: I checked the  expected output with `--graalvm-show` and the list is identical for both:
`mx --env ce graalvm-show`
and 
`mx --env ce --sources=sdk:GRAAL_SDK,truffle:TRUFFLE_API,compiler:GRAAL,substratevm:SVM graalvm-show`

vm/README.md
olyagpl: Removed.

dougxc:
@simonis I think it's good to have the document updated but in my view, the best way to have the process documented (in a testable way) is to augment the existing GitHub Actions in the graal repo to include the building and testing that is of interest to the community. One idea would be to reproduce the internal CI jobs (defined in the various `ci/` and `*.jsonnet`) as GA workflows. Even better would be to have the latter automatically derived from the former. I think @ezzarghili also agrees with this - it's just a matter of finding the resources to make it happen.

simonis:
> @simonis I think it's good to have the document updated but in my view, the best way to have the process documented (in a testable way) is to augment the existing GitHub Actions in the graal repo to include the building and testing that is of interest to the community. One idea would be to reproduce the internal CI jobs (defined in the various `ci/` and `*.jsonnet`) as GA workflows. Even better would be to have the latter automatically derived from the former. I think @ezzarghili also agrees with this - it's just a matter of finding the resources to make it happen.

@dougxc I'm completely fine with that approach. But I think it should also be possible to run the tests locally (especially for developers who are new to the project which I think is the audience for this document) and ideally both approaches should run the same set of tests.

I must also confess, that I'm not very familiar with GHA. So if it's possible to easily trigger and run them in personal forks of the graal repository, then we should document that in the Readme.

vm/README.md
olyagpl: Added in this commit https://github.com/oracle/graal/pull/12924/changes/391464f47cb38a365ebd608c7bc9f4e6b1d700c9

synecdoche:
@dougxc @simonis I'm strongly in favor of both GHA examples that demonstrate common build and test scenarios as well as documentation that shows how a new programmer might do the most common thing: build the code, make a change, and run some tests, all on their own machine. The GHA approach is useful so one can see not only how to build and test, but also that it does so successfully _somewhere_. However, GH-hosted actions is its own particular, peculiar environment, and new developers are likely to bump into a different set of issues on their machines.

I do have some experience wrangling a GHA workflow or two, so let me know if I can help with the external GHA effort. As Doug mentioned, this would be of interest to the community too. And of course for internal stuff, there is [self hosted GHA runners](https://docs.github.com/en/actions/concepts/runners/self-hosted-runners).

Issue #12641, opened 2025-12-02
[Native Image] Basic XML Signature usage fails with missing resource bundle


jerboaa:
### Describe the Issue

When running the following program as native-image it fails at runtime with a `java.lang.ExceptionInInitializerError`:

```
import javax.xml.crypto.dsig.Transform;
import javax.xml.crypto.dsig.XMLSignatureFactory;
import javax.xml.crypto.dsig.spec.TransformParameterSpec;

public class TestXMLSigTransform {
    
    public static final XMLSignatureFactory FAC = XMLSignatureFactory.getInstance("DOM");

    public static void main(String[] args) throws Exception {
        TransformParameterSpec params = null;
        Transform tr = FAC.newTransform(Transform.ENVELOPED, params);
        System.out.println(tr);
    }

}
```

### Using the latest version of GraalVM can resolve many issues.

- [x] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

### GraalVM Version

GraalVM build from master. Revision cca0ccd49b411efcf790be0f8e7805a36ad41d1c

### Operating System and Version

Linux x86_64

### Troubleshooting Confirmation

- [ ] I tried the suggestions in the [troubleshooting guide](https://www.graalvm.org/latest/reference-manual/native-image/guides/troubleshoot-run-time-errors/).

### Run Command

./testxmlsigtransform 

### Expected Behavior

Same as when run with `java TestXMLSigTransform.java`:

```
$ java TestXMLSigTransform.java
org.jcp.xml.dsig.internal.dom.DOMTransform@d130cf06
```

### Actual Behavior

```
./testxmlsigtransform 
Exception in thread "main" java.lang.ExceptionInInitializerError
	at java.xml.crypto@25.0.1/org.jcp.xml.dsig.internal.dom.XMLDSigRI$ProviderService.newInstance(XMLDSigRI.java:124)
	at java.xml.crypto@25.0.1/javax.xml.crypto.dsig.TransformService.getInstance(TransformService.java:246)
	at java.xml.crypto@25.0.1/org.jcp.xml.dsig.internal.dom.DOMXMLSignatureFactory.newTransform(DOMXMLSignatureFactory.java:377)
	at TestXMLSigTransform.main(TestXMLSigTransform.java:11)
	at java.base@25.0.1/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)
[...]

PR #12949, opened 2026-02-11
Fix resource bundle registration for java.xml.crypto TransformService

jerboaa:
This fix proposes to add the required resource bundle when `TransformService.getInstance()` from the `java.xml.crypto` module is reachable. Thoughts?

Testing:
- [x] Passes the reproducer from the issue. I've also tested this with a "larger" app on a 25u backport.

Closes: #12641

### DIFF
--- a/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
+++ b/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/xml/JavaxXmlClassAndResourcesLoaderFeature.java
@@ -33,6 +33,7 @@ import static com.oracle.svm.hosted.xml.XMLParsersRegistration.SAXParserClasses;
 import static com.oracle.svm.hosted.xml.XMLParsersRegistration.SchemaDVFactoryClasses;
 import static com.oracle.svm.hosted.xml.XMLParsersRegistration.StAXParserClasses;
 import static com.oracle.svm.hosted.xml.XMLParsersRegistration.TransformerClassesAndResources;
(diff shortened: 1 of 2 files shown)


loicottet:
The approach looks good to me, but I would register the reachability handler for the actual methods where the resource bundle gets looked up (`I18n.init`), so we're sure to not register it when not needed.

jerboaa:
> The approach looks good to me, but I would register the reachability handler for the actual methods where the resource bundle gets looked up (`I18n.init`), so we're sure to not register it when not needed.

Using `I18n.init()` now. Please take another look. Thanks for the review!

loicottet:
LGTM, I'll merge the PR. Thanks for the fix!

jerboaa:
Thanks!

jerboaa:
@loicottet Any update on getting this merged? Thank you!

loicottet:
The fix is failing in our internal CI, I'm working on fixing it.

jerboaa:
> The fix is failing in our internal CI, I'm working on fixing it.

OK. Thanks for the update.

loicottet:
Fixed by #12949

zakkak:
> Fixed by #12949

Hi @loicottet , I guess this was meant for https://github.com/oracle/graal/issues/12641

BTW, it looks like the bot did something "wrong" and the commits don't reference this PR:

* https://github.com/oracle/graal/commit/a65cc1cbeda9c5db7fc96f553bf35b8e5b18839d
* https://github.com/oracle/graal/commit/2e556f118f8b1303e858e2cf2a53ceff02d09b0c

loicottet:
Something weird happened indeed, our CI showed this PR as the mirror for the internal one, but it did not reflect my fixes and I can't find an up-to-date mirror PR. In any case, #12641 is indeed solved now.

PR #12982, opened 2026-02-16
JNI Validation Performance Test

asselyam:
Performance test was added to measure the execution time between enabled JNIValidation option and disabled


asselyam:
Dear Dr. @christianhaeubl here is the updated version of the thesis with additional performace check

PR #12965, opened 2026-02-12
fix: update ASM to 9.8

porunov:
Updates ASM to 9.8 to support Java 25 class-predefinition-support.

Notice, the latest ASM version is 9.9.1 that supports Java 26, but this PR syncs ASM to 9.8 for consistency with [mx suite](https://github.com/graalvm/mx/blob/6934051ffc792c8f5b9a6fd35aeb3e257ad67a3f/mx.mx/suite.py#L46-L91).

The intent of this PR is to be backported to [GraalVM 25](https://github.com/oracle/graal/tree/release/graal-vm/25.0) that will fix `class-predefinition-support`. 

Fixes #12723

Please, let me know if the development process requires to target non-`master` branch and I can rebase it against another branch in that case.

Issue #12723, opened 2025-12-16
[GR-72123][Native Image] Error Report - Included ASM 9.7.1 is not supporting Java 25 compiled classes


porunov:
### Describe the Issue

Current included ASM 9.7.1 is not supporting Java 25 compiled classes. ASM added support for Java 25 starting from the version 9.8.
Links to included shaded ASM dependency: 
https://github.com/oracle/graal/blob/e439da27aa3936ad13999f9cc02fad308960e5d3/truffle/mx.truffle/suite.py#L1454-L1456

https://github.com/oracle/graal/blob/ba0bea74ed63b28bcba714edfdfdcd63a73587a7/espresso/mx.espresso/suite.py#L463-L465

Due to this, users are unable to use native compilation when the project's target Java version is 25 and `experimental-class-define-support` is enabled.

Proposed solution: Updating ASM to version 9.8 or later. 

### Please confirm if you have tested with the latest available version.

- [x] I tried with the latest version of GraalVM. [Latest GraalVM Version](https://www.graalvm.org/downloads/)

### Share the `svm_err_b_*.md` File

# Complete report attached as a file:

[svm_err_b_20251216T135154.850_pid1528.md](https://github.com/user-attachments/files/24193029/svm_err_b_20251216T135154.850_pid1528.md)

# GraalVM Native Image Error Report

## Build Output

```
========================================================================================================================
GraalVM Native Image: Generating 'api' (executable)...
========================================================================================================================
For detailed information and explanations on the build output, visit:
https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/BuildOutput.md
------------------------------------------------------------------------------------------------------------------------

[1/8] Initializing...                                                                                    (0.0s @ 0.75GB)
```

## Stack Trace

```java
com.oracle.svm.core.util.VMError$HostedError: InternalFeature defined by com.oracle.svm.hosted.ClassPredefinitionFeature unexpectedly failed with a(n) java.lang.IllegalArgumentException
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.util.VMError.shouldNotReachHere(VMError.java:86)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.FeatureHandler.handleFeatureError(FeatureHandler.java:304)
	at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.FeatureHandler.forEachFeature(FeatureHandler.java:95)
[...]

PR #13016, opened 2026-02-19
Update ASM to 9.8

porunov:
Updates ASM to 9.8 to be in sync with `mx-suite` and `graalvm-community-jdk25u` repositories regarding ASM version.

Notice, the latest ASM version is 9.9.1 that supports Java 26, but this PR syncs ASM to 9.8 for consistency with [mx suite](https://github.com/graalvm/mx/blob/6934051ffc792c8f5b9a6fd35aeb3e257ad67a3f/mx.mx/suite.py#L46-L91).

This PR is the same as in previously closed PR #12965 but no longer related to previous issue #12723 which was fixed in `graalvm-community-jdk25u`.
The only purpose of this PR is to keep versions in sync between repositories, but not particular functionality issues.
If there are any reasons to keep ASM 9.7.1 in `oracle/graal` repository - let me know and we can close this PR in this case.

### DIFF
--- a/espresso/mx.espresso/suite.py
+++ b/espresso/mx.espresso/suite.py
@@ -461,7 +461,7 @@ suite = {
             "javaCompliance" : "17+",
             "spotbugs" : "false",
             "shadedDependencies" : [
-                "truffle:ASM_9.7.1",
+                "truffle:ASM_9.8",
             ],
             "class" : "ShadedLibraryProject",
             "shade" : {
--- a/truffle/mx.truffle/suite.py
+++ b/truffle/mx.truffle/suite.py
@@ -109,38 +109,38 @@ suite = {
       "license": ["MIT"],
     },
 
-    "ASM_9.7.1": {
-      "digest": "sha512:4767b01603dad5c79cc1e2b5f3722f72b1059d928f184f446ba11badeb1b381b3a3a9a801cc43d25d396df950b09d19597c73173c411b1da890de808b94f1f50",
-      "sourceDigest": "sha512:d7c0de5912d04949a3d06cad366ff35a877da2682d9c74579625d62686032ea9349aff6102b17f92e9ec7eb4e9b1cd906b649c6a3ac798bfb9e31e5425de009d",
+    "ASM_9.8": {
+      "digest": "sha512:cbd250b9c698a48a835e655f5f5262952cc6dd1a434ec0bc3429a9de41f2ce08fcd3c4f569daa7d50321ca6ad1d32e131e4199aa4fe54bce9e9691b37e45060e",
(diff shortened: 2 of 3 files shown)


oubidar-Abderrahim:
Thank you for your contribution, we'll take a look at this shortly

@wirthi could you please assign the appropriate person to review? thank you

espresso/mx.espresso/suite.py
wirthi, line 469: I think this code needs to be updated, as we shadow that library there. I don't think this happend as part of your PR?

gilles-duboscq (approved):

espresso/mx.espresso/suite.py
porunov: @wirthi I'm not sure what parts of the code should be updated. As far as I know ASM 9.8 didn't have any breaking changes, so I doubt there would be necessity to update particular code, but I might just miss something that's not obvious to me.
From ASM [changelog](https://asm.ow2.io/versions.html):

> 29 March 2025: ASM 9.8 (tag ASM_9_8)
> 
>     new Opcodes.V25 constant for Java 25
>     bug fixes
>         Fix one more copy operation on DUP2
>         318015: Valid bytecode for jvm, but failed to pass the CheckClassAdapter.
>         `ASMifier` should print calls to `valueOf` instead of deprecated constructors of primitive wrappers

Do you think there could be some places where these fixes were previously patched and I need to remove those patches or you refer to some other changes that should be done during version updates of shadowed libraries?

wirthi: @gilles-duboscq is this update of the dependency alone enough to update the shadowed library? Or does @porunov need to update any code in the repo for this?

gilles-duboscq: I think the update to the library alone should be enough.

woess (approved):

espresso/mx.espresso/suite.py
porunov: @wirthi let me know please if there are any other changes needed in this PR or if the current changes are sufficient. Thank you!

porunov: @wirthi I just wander if you had a chance to take another look into this PR. Do you see any obstacles or issues that could prevent this PR from being merged? Happy to resolve those if there are any. Thank you.

wirthi: sorry for the delay. I'll put the PR in our internal CI, will also require internal approvals.

I am on it.

porunov: Thank you! Let me know if any changes are needed from my side and I will try to resolve them if there are any.

wirthi:
this was merged in #13367

PR #13037, opened 2026-02-24
[GR-73556] Remove deprecated AutomaticFeature annotation

graalvmbot:
This annotation has been deprecated for a while. It is time to remove it and simplify code movement.


fniephaus:
FYI @zakkak, @sdeleuze, @graemerocher, @tomas-langer

It seems all frameworks have removed the use of this annotations a long time ago.

PR #13080, opened 2026-03-03
JNI Validation tests

asselyam:
A few tests were added to test the correct behaviour of the system to the misuse cases


asselyam:
Dear Dr. Häubl @christianhaeubl here is the updated version of the code

PR #13085, opened 2026-03-04
[GR-71079] Running Tests in GraalVM CI doc.

graalvmbot:


vm/TESTING.md
```diff
@@ -0,0 +1,149 @@
+# Testing GraalVM: Understanding CI and Running Tests Locally
```
olyagpl, line 1: @simonis, 
This is the continuation of https://github.com/oracle/graal/pull/12924 where we agree that some documentation on running tests  would be necessary. 
Could you please have a look at this doc? Maybe you'd have some suggestions.

vm/TESTING.md
```diff
@@ -0,0 +1,149 @@
...
+## Local Test Environment
+
+You can run the same core test commands locally that are used in GraalVM CI.
+Start by preparing your local environment.
+
+1. Ensure your development environment meets the prerequisites: [prepare the environment](https://github.com/oracle/graal/blob/master/vm/README.md#prepare-the-environment).
+2. Install `mx` as described [here](README.md#setting-mx).
+3. Clone your fork of the Graal repository and enter the component directory you want to test (for example, `compiler`, `substratevm`, or `truffle`).
+4. Make sure you are using a JDK version supported by CI (for example, `jdk21`, `jdk25`, or `latest`).
+5. Run tests using `mx`. For example:
```
simonis, line 105: Which build does this test? The previous steps only describe how to checkout Graal but not what needs to be build and which build artifacts will be tested.

E.g if just do the above, it will simply not work.

Even if I first build Native Image with `mx --primary-suite=substratevm --components=ni build` and then run `mx --primary-suite=substratevm native-unittest` it will fail. In order for it to work, I would first have to build `nju` in the build step (i.e. `mx --primary-suite=substratevm --components=ni,nju build`). 

Even after building `nju`, `mx --primary-suite=substratevm native-unittest` works, but `mx --primary-suite=substratevm unittest` fails with:
```
Error occurred during initialization of boot layer
java.lang.module.FindException: Module org.graalvm.sdk not found
```

This is all completely mysterious. I also don't understand why this doc is in the `ci/` folder? What we really need is a comprehensive developer documentation which includes local testing. E.g. https://github.com/oracle/graal/tree/master/substratevm/docs recently added some documentation, but it is incomplete and inaccurate. E.g. it mentions to run `mx --primary-suite=substratevm gate --dry-run` in order to get the tag names required for `gate --tags` but `--dry-run` will only print the verbose task names. In order to get the tags, we have to use `mx -v gate --dry-run`.

Issue #13096, opened 2026-03-05
[Native Image] GroovySubstitutions are Groovy 4 specific (break with Groovy 5).


jerboaa:
### Describe the Issue

This was originally filed under quarkus but it turns out the rest-assured update from 5.x to 6.0.0 caused native build fails for some of the quarkus native integration tests. The underlying issue is that the 5.x to 6.x update implicitly upgraded Groovy to version 5. In groovy 5 the method referenced in the `GroovySubstitutions.java` class does no longer exist and, thus, potentially breaks apps which have groovy 5 code on the classpath.

https://github.com/oracle/graal/blob/28bd824861942178a62658ee79ba574cbcfc74e5/substratevm/src/com.oracle.svm.polyglot/src/com/oracle/svm/polyglot/groovy/GroovySubstitutions.java#L52-L55

This results in quarkus native build failures of this sort:
```
[...]

PR #12986, opened 2026-02-16
Fix Groovy substitutions for Apache Groovy /5.x compatibility

lissam1:
### Problem

Native Image builds fail when Apache Groovy  5.x is present on the classpath with the following error during the initialization phase:
```
Error: Could not find target method: protected static void 
com.oracle.svm.polyglot.groovy.Target_org_codehaus_groovy_vmplugin_v7_IndyInterface_invalidateSwitchPoints.invalidateSwitchPoints()
```

This affects projects using:
- Apache Groovy 5.x (`org.apache.groovy:groovy`)
- Micronaut

Reported in issue #10200.

### Investigation & Root Cause

Through investigation of the Groovy source code and GraalVM substitution mechanism, we identified the following:

#### 1. Architecture Change in Groovy /5.x

Groovy changes:

**Old Groovy :**
- `org.codehaus.groovy.vmplugin.v7.IndyInterface` contains `invalidateSwitchPoints()` method

**New Groovy (5.x):**
- `org.codehaus.groovy.vmplugin.v8.IndyInterface` is the primary interface (contains `invalidateSwitchPoints()`)
- `org.codehaus.groovy.vmplugin.v7.IndyInterface` exists only as a **deprecated bridge class** for backward compatibility
- v7 **no longer contains** `invalidateSwitchPoints()` - it only delegates to v8

**Evidence from Groovy source code:**

v7/IndyInterface (Groovy 5.x):
```java
@Deprecated
public class IndyInterface {
    // All methods delegate to v8:
    public static CallSite bootstrap(...) {
        return org.codehaus.groovy.vmplugin.v8.IndyInterface.bootstrap(...);
    }
    
    // NO invalidateSwitchPoints() method!
}
```

v8/IndyInterface (Groovy 5.x):
```java
public class IndyInterface {
    protected static void invalidateSwitchPoints() {  // ← Method is HERE now
        if (LOG_ENABLED) {
            LOG.info("invalidating switch point");
        }
        synchronized (IndyInterface.class) {
            SwitchPoint old = switchPoint;
            switchPoint = new SwitchPoint();
            SwitchPoint.invalidateAll(new SwitchPoint[]{old});
        }
    }
}
```

#### 2. Current GraalVM Substitution Issue

The existing `GroovyIndyInterfaceFeature` has :

**Only targets v7**
```java
@Override
public boolean isInConfiguration(IsInConfigurationAccess access) {
[...]

PR #13097, opened 2026-03-05
Fix groovy substitutions for Groovy 5+

jerboaa:
The feature originally expects the `org.codehaus.groovy.vmplugin.v7.IndyInterface` class to be present for it to be "in config". This is no longer sufficient since Groovy 5 has the substituted method moved to the `v8` package. See:

https://docs.groovy-lang.org/docs/groovy-5.0.0/html/api/org/codehaus/groovy/vmplugin/v8/IndyInterface.html

Instead, check that the method is present as well for the feature to be in config. Use the version specific substitution depending on the Groovy version.

*Testing:*
- [x] Quarkus native integration test referenced in #13096 passes after this change, fails before.

Closes: #13096

### DIFF
--- a/substratevm/src/com.oracle.svm.polyglot/src/com/oracle/svm/polyglot/groovy/GroovySubstitutions.java
+++ b/substratevm/src/com.oracle.svm.polyglot/src/com/oracle/svm/polyglot/groovy/GroovySubstitutions.java
@@ -32,23 +32,71 @@ import org.graalvm.nativeimage.hosted.Feature;
 import com.oracle.svm.core.annotate.Substitute;
 import com.oracle.svm.core.annotate.TargetClass;
 
-final class GroovyIndyInterfaceFeature implements Feature {
+final class GroovyV4IndyInterfaceFeature extends GroovyIndyInterfaceFeature {
+
+    // Groovy 4 has the invalidateSwitchPoints method in the v7 package
+    static final String INDY_INTERFACE_CLASS_NAME = "org.codehaus.groovy.vmplugin.v7.IndyInterface";
 
     static final class IsEnabled implements BooleanSupplier {
         @Override
         public boolean getAsBoolean() {
-            return ImageSingletons.contains(com.oracle.svm.polyglot.groovy.GroovyIndyInterfaceFeature.class);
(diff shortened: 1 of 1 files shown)


jerboaa:
@vjovanov Could you please help get a fix like this one integrated? Thank you!

vjovanov:
Yes, ideally we want to move features like this into the language or library. However, this feature seems to be valid only for earlier versions of Groovy so we don't need to move it.

This should be also backported to 25 until we get the feature into Groovy and for Groovy previous versions.

jerboaa:
I can cover the GraalVM 25 community backport once this fix is integrated.

oubidar-Abderrahim:
I mirrored the PR and should proceed with it once the gates are all green

jerboaa:
> I mirrored the PR and should proceed with it once the gates are all green

Thanks!

lissam1:
`invalidateSwitchPoints()` was moved to v8 package for groovy 5, i had a draft pr before my idea was to check in under v7 or v8 so to have backward compatibility. see https://github.com/oracle/graal/pull/12986

jerboaa:
The updated patch checks both `v7` and `v8`.

jerboaa:
GHA failure seems unrelated. I see the same math stubs failure in https://github.com/oracle/graal/pull/13207

jerboaa:
@vjovanov @oubidar-Abderrahim Could we please move forward with this? It's causing test noise in our CI for testing quarkus on graal/master (which we run daily). Thank you!

vjovanov:
@jerboaa sorry for the delay. We had a miscommunication about who is managing this PR. 

We have replaced the reflection acces with `ReflectUtil` and now it is being tested.

jerboaa:
OK, thanks!

jerboaa:
Yes. https://github.com/oracle/graal/commit/7e522b59d23a0d360ba94aa6a1db87d19d93b2c8 fixes this in a less Groovy version specific way. So this PR would have to be re-done as a follow-up to https://github.com/oracle/graal/commit/7e522b59d23a0d360ba94aa6a1db87d19d93b2c8 which isn't on my list of priorities currently.

PR #13136, opened 2026-03-12
[GR-73969] Add Project Terminus Migration Guide

graalvmbot:
This add the **Project Terminus Migration Guide**. It is intended for Native Image developers migrating build-time code away from host reflection and toward Terminus-safe guest-context APIs.


zakkak:
@zapster I asssume GraalVM Features should follow a similar migration. Is that right? Is there any other "graalvm-external" code that should be migrated?

zapster:
@zakkak what exaclty do you mean with "GraalVM Features"? If it is features that are using only the public APIs, those don't need to be migrated. We will run them in the guest context. We also have "guest features" internally, e.g., https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.guest/src/com/oracle/svm/guest/features/jdk/hosted/JavaNetHttpFeature.java. Those can (and must) continue use core reflection just as they did before. It is only "internal" features that heavily interact with the builder that need to switch to JVMCI reflection.

zakkak:
> If it is features that are using only the public APIs, 

Correct, thank you for the clarification @zapster !

PR #13174, opened 2026-03-19
[GR-71501] Ristretto deopt support.

graalvmbot:


vjovanov:
Would be cool to leave a description on this PR with some insight into the key things that were done.

PR #13217, opened 2026-03-26
[GR-74344] Add workflow to check GitHub Actions with Macaron.

graalvmbot:


.github/workflows/macaron-check-github-actions.yml
```diff
@@ -38,3 +38,26 @@
...
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/**'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  id-token: write
```
behnazh-w: Since this workflow is running on pull requests, let's remove `id-token: write` and attestation upload to avoid giving untrusted runs signing arbitrary reports.

.github/workflows/macaron-check-github-actions.yml
```diff
@@ -38,3 +38,26 @@
...
+  analyze:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - name: Run Macaron action
+        uses: oracle/macaron@fda4dda04aa7228fcaba162804891806cf5a1375  # v0.22.0
+        with:
+          repo_path: ./
+          policy_file: check-github-actions
+          policy_purl: pkg:github.com/oracle/graal@.*
```
behnazh-w, line 62:
```suggestion
          policy_purl: pkg:github.com/oracle/graal@.*
    - name: Upload Macaron reports
      if: ${{ always() }}
      uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
      with:
        name: macaron-reports
        path: |
          output/reports/github_com/oracle/graal/graal.html
          output/macaron.db
        if-no-files-found: warn
        retention-days: 90

    - name: Check policy verification passes
      if: ${{ always() }}
      run: |
        if [ ! -f output/vsa.intoto.jsonl ]; then
          echo "The check-github-actions policy failed. Check the uploaded reports."
          exit 1
        fi
```

PR #13223, opened 2026-03-27
Chrome Inspector Tool: Fix info messages written to error stream

### LINKED ISSUE #13222: [GR-74436] Chrome Inspector Tool writes info messages to error stream
 - GraalVM version or commit id if built from source: 25.0.1
 - CE or EE: CE
 - JDK version: JDK21
 - OS and OS Version: Linux
  - Architecture: amd64
 - The output of `java -Xinternalversion`: 
```
 OpenJDK 64-Bit Server VM (21.0.10+7-LTS) for linux-amd64 JRE (21.0.10+7-LTS), built on 2026-01-20T00:00:00Z by "admin" with gcc 11.3.0
```

Yes, the latest code still includes the bug.

**Describe the issue**
The GraaVM Chrome Inspector Tool logs its info output to the error stream, one example of this:

https://github.com/oracle/graal/blob/86297a5f72182866688ab6593b87108aaa070602/tools/src/com.oracle.truffle.tools.chromeinspector/src/com/oracle/truffle/tools/chromeinspector/instrument/InspectorInstrument.java#L380

**Expected behavior**
Info output is logged to the output stream, i.e. change the above to:

```java
 PrintWriter info = new PrintWriter(env.out(), true); 
```

florian-h05:
Fixes #13222.

### DIFF
--- a/tools/src/com.oracle.truffle.tools.chromeinspector.test/src/com/oracle/truffle/tools/chromeinspector/test/MultiEngineTest.java
+++ b/tools/src/com.oracle.truffle.tools.chromeinspector.test/src/com/oracle/truffle/tools/chromeinspector/test/MultiEngineTest.java
@@ -136,8 +136,9 @@ public class MultiEngineTest extends EnginesGCedTest {
         AtomicInteger port = new AtomicInteger(0);
         verifySerialDebug(port, isUp, error);
         ByteArrayOutputStream out = new ByteArrayOutputStream();
+        ByteArrayOutputStream err = new ByteArrayOutputStream();
         for (int i = 0; i < sources.length; i++) {
-            runEngine(sources[i], port, out, isUp[i]);
+            runEngine(sources[i], port, out, err, isUp[i]);
         }
         if (error.get() != null) {
             throw new AssertionError(error.get());
@@ -178,6 +179,7 @@ public class MultiEngineTest extends EnginesGCedTest {
         }
         List<Throwable> errors = Collections.synchronizedList(new LinkedList<>());
         ByteArrayOutputStream out = new ByteArrayOutputStream();
+        ByteArrayOutputStream err = new ByteArrayOutputStream();
         Thread[] threads = new Thread[sources.length];
         int port = getFreePort();
         verifyParallelDebug(port, isUp, errors);
@@ -185,7 +187,7 @@ public class MultiEngineTest extends EnginesGCedTest {
             int index = i;
             Thread t = new Thread(() -> {
                 try {
-                    runEngine(sources[index], new AtomicInteger(port), out, isUp[index]);
+                    runEngine(sources[index], new AtomicInteger(port), out, err, isUp[index]);
                 } catch (Throwable thr) {
                     thr.printStackTrace();
                     errors.add(thr);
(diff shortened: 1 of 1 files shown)


fniephaus:
Thanks for the fix, and thank you for contributing this.

I took a close look at the change and ran the relevant chrome inspector tests locally with `mx`.

What I found:
- The behavioral change looks correct: informational inspector output now goes to `stdout`, while the actual error paths in `open(...)` still write to `stderr`.
- `TimeoutTest` passes.
- `InspectorMessageTransportTest` passes.
- `InspectorAddressTest` now fails in the cases that expect the inspector banner on `stderr` (`testHostPortDefault`, `testHostPortEnabled`, `testPort`, `testPort0`, and `testPath`).

The failures line up with the PR change: `InspectorAddressTest` still captures only `.err(...)` and parses the WebSocket banner from that stream, so it needs to be updated to read/assert `stdout` for the banner and keep `stderr` for real errors only.

So overall this looks like the right fix to me, but I do think the test update should be included before merging. Thanks again for the contribution. 🤖

florian-h05:
@fniephaus Thanks for your review!
I've updated the tests, but unfortunately I haven't been able to run them locally, cannot get the mx build working on Fedora 43.
Where can I see the test results in CI?

fniephaus:
>I've updated the tests, but unfortunately I haven't been able to run them locally, cannot get the mx build working on Fedora 43. Where can I see the test results in CI?

We don't run the tests on GitHub CI, too bad you cannot run them on Fedora. I will share any problems caught by our internal CI.

tools/src/com.oracle.truffle.tools.chromeinspector.test/src/com/oracle/truffle/tools/chromeinspector/test/InspectorAddressTest.java
```diff
@@ -38,51 +38,56 @@
...
+        String out = outOutput.toString();
         assertTrue(out, out.isEmpty());
+        String err = errorOutput.toString();
+        assertTrue(err, err.isEmpty());
     }
 
     @Test
     public void testPort() {
-        try (Context context = Context.newBuilder().option("inspect", "2992").err(errorOutput).build()) {
+        try (Context context = Context.newBuilder().option("inspect", "2992").info(outOutput).build()) {
```
fniephaus: `Context.newBuilder().info(outOutput)` does not compile: `Context.Builder` exposes `out(...)` and `err(...)`, but not `info(...)`. This should be `Context.newBuilder().option("inspect", "2992").out(outOutput).build()`. 🤖

fniephaus:
`MultiEngineTest` still captures inspector startup messages from stderr at `Engine.newBuilder()...err(out)`, but this PR moves those messages to stdout. That breaks `runEngine(...)` because `InspectorAddressTest.parseWSPort(out.toString())` no longer sees the `Debugger listening on ...` line, and it also weakens the same-path assertions. This test should be updated to read stdout as well, likely with `.out(out)` or `.out(out).err(out)`. I am leaving this as a general PR comment because `MultiEngineTest` is not part of the changed diff, so GitHub will not accept it as an inline review comment. 🤖

florian-h05:
I built me a Docker container based on Ubuntu to run mx and it works fine.
I've now been able to run the tests myself and fix all remaining test failures, `mx gate` now succeeds.

PR #13229, opened 2026-03-27
Register allocator correctness verifier

danocmx:
Register allocation verifier, inspired by cranelift's regalloc checker. Works by saving symbols before allocation - variables, constants, registers and stack slots, then matches these to locations inserted by allocator(s) - registers and stack slots. Outputs of instructions then store these symbols to those locations in block's verification state. Spills, register moves and reloads move these around. Before any checking is done, initial allocation state (map of form location → symbol) is calculated for each block based on it's predecessors, when same location with different symbols meet, a conflict is created - can be resolved by writing to the location. Afterwards, checking is done for each block, we verify that symbols before allocation are stored at the locations inserted by allocator(s), taking into account reloads and spills that the allocator created.

Implemented as a phase (`RegAllocVerifierPhase`) that wraps around both register and stack allocator, collects symbol information before allocation and matches this information to instructions after allocation to create verifier IR that it then uses to update state allocation maps and to verify operands.

When a violation is detected by the verifier, a `RAVException` is thrown and if dumping is enabled a `.rav.txt` file is created in `graal_dumps` directory with the exception details and verifier IR.

Implemented as `jdk.graal.compiler.lir.alloc.verifier` package, modifies `AllocationStage` to insert the verification phase when enabled and also changes modifier to public for `build` method in `LocationMarker` to access it in the verifier package. No other modification to existing was needed, variable locations stripped by the allocator from label/jump instructions are recovered by finding their first usage in `FromUsageGlobalResolver` class to not mess with the existing register allocator code.

Some functionality tested in `RegAllocVerifierTest`, runs valid methods with the verifier, then injects faults and tests if the verifier can detect them. To be expanded.

New compiler flags:
- `EnableRAVerifier` - enables the verification for the compilation
- `VerifyStackAllocator` - verifies that stack allocator output together with register allocator
- `CollectReferences` - uses `LocationMarker` class to collect GC reference sets before final code analysis stage to verify and invalidate old references

Other functionality
- Handle rematerialised constants
- Check kinds between original variables and current locations, as well as the allocation state at said location
- Check kinds between moved values
- Make sure callee-saved registers are recovered
- Check correspondence to operand flags
- Make sure `alive` location is not overwritten by the instruction as `temp` or `output`
- Remove old references and check if location marked as reference actually stores one
- Make sure LIR reference and `JavaKind` Object match


danocmx:
cc @d-kozak @gergo-

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/ConflictedAllocationState.java
```diff
@@ -0,0 +1,150 @@
...
+    public void addConflictedValue(ValueAllocationState state) {
+        if (hasConflictedValue(state)) {
+            return;
+        }
+
+        this.conflictedStates.add(state);
+    }
+
+    public boolean hasConflictedValue(ValueAllocationState valueAllocationState) {
+        for (var state : this.conflictedStates) {
```
gergo-, line 64: It seems unfortunate that `conflictedStates` is a set but you need to do a linear search in it. Maybe instead of a set of states, it should be an `RAValue -> ValueAllocationState` map?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/AllocationState.java
```diff
@@ -0,0 +1,81 @@
...
+        return false;
+    }
+
+    /**
+     * Create a copy of this state, necessary for state copies made over program graph edges.
+     *
+     * @return Newly copied state
+     */
+    @Override
+    public abstract AllocationState clone();
```
gergo-, line 67: My understanding is that you should implement the `Cloneable` interface if you override `Object.clone()`.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/AllocationStateMap.java
```diff
@@ -0,0 +1,206 @@
...
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Mapping between a location and allocation state that stores one of these: -
+ * {@link UnknownAllocationState unknown} - our null state, nothing was stored yet -
+ * {@link ValueAllocationState value} - symbol that is stored at said location -
+ * {@link ConflictedAllocationState conflicted} - set of Values that are supposed to be at same
+ * location
```
gergo-, line 42: Automated formatting destroyed your list, please use proper HTML list syntax in Javadoc comments.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/AllocationStateMap.java
```diff
@@ -0,0 +1,206 @@
...
+    public boolean mergeWith(AllocationStateMap source) {
+        boolean changed = false;
+        for (var entry : source.internalMap.entrySet()) {
+            if (!this.internalMap.containsKey(entry.getKey())) {
+                changed = true;
+
+                this.putWithoutRegCheck(entry.getKey(), UnknownAllocationState.INSTANCE);
+            }
+
+            var currentValue = this.internalMap.get(entry.getKey());
```
gergo-, line 176: This method would be a bit cleaner if you did this `get` once and then used the result everywhere, including `currentValue == null` instead of the `containsKey` call.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/RAVException.java
gergo-, line 1: I think it would make sense for this exception class and all of its subclasses to live in a separate subpackage.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/RematerializedConstantSourceMissingError.java
```diff
@@ -22,25 +22,23 @@
...
-
-public class JfrOptions {
-    @Option(help = "Determines if a signal handler is used for execution sampling.")//
-    public static final HostedOptionKey<Boolean> SignalHandlerBasedExecutionSampler = new HostedOptionKey<>(false, JfrOptions::validateSamplerOption);
+/**
+ * Re-materialized constant has a wrong source (not a {@link RAVInstruction.ValueMove}), but either
+ * undefined or some different Operation or Move.
+ */
+@SuppressWarnings("serial")
+public class RematerializedConstantSourceMissingError extends RAVError {
```
gergo-, line 32: Why is this an (internal) error rather than an exception?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/CalleeSaveMap.java
```diff
@@ -0,0 +1,128 @@
...
+ * <p>
+ * Virtual moves assigning to the same registers need to also be handled, because both values could
+ * be retrieved.
+ * </p>
+ */
+public class CalleeSaveMap {
+    protected RegisterConfig registerConfig;
+    protected List<Register> calleeSaveRegisters;
+
+    protected Map<RARegister, RAValue> virtualValues;
```
gergo-, line 47: Generally speaking, fields only assigned in the constructor should be declared `final`.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/CalleeSaveMap.java
```diff
@@ -0,0 +1,128 @@
...
+ * Make sure callee-saved register values are retrieved at an exit block.
+ *
+ * <p>
+ * Virtual moves assigning to the same registers need to also be handled, because both values could
+ * be retrieved.
+ * </p>
+ */
+public class CalleeSaveMap {
+    protected RegisterConfig registerConfig;
+    protected List<Register> calleeSaveRegisters;
```
gergo-, line 45: If this is always just a reference to `registerConfig.getCalleeSaveRegisters()`, then you shouldn't use a separate field for it.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/CalleeSaveMap.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.List;
+import java.util.Map;
+
+import jdk.graal.compiler.util.EconomicHashMap;
+import jdk.vm.ci.code.Register;
+import jdk.vm.ci.code.RegisterConfig;
+import jdk.vm.ci.code.RegisterValue;
+
+/**
+ * Make sure callee-saved register values are retrieved at an exit block.
```
gergo-, line 36: This class doesn't really "make sure" that this happens, it stores the relevant data. Please focus more on what this class stores, and you can `@link` to the place where the relevant properties are actually checked.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/CalleeSaveMap.java
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * <p>
+ * Virtual moves assigning to the same registers need to also be handled, because both values could
+ * be retrieved.
```
gergo-, line 40: This sentence is not clear to me. Maybe an example would help, or a reference to the place where this constraint is actually checked.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/CalleeSaveMap.java
```diff
@@ -0,0 +1,128 @@
...
+
+    protected Map<RARegister, RAValue> virtualValues;
+
+    public CalleeSaveMap(RegisterConfig registerConfig) {
+        this.registerConfig = registerConfig;
+        calleeSaveRegisters = registerConfig.getCalleeSaveRegisters();
+        virtualValues = new EconomicHashMap<>();
+    }
+
+    public class CalleeSavedRegister extends RARegister {
```
gergo-, line 55: I think this should be called `CalleeSavedRAVRegister`, and `RARegister` would also be better as `RAVRegister`.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/ConflictedAllocationState.java
```diff
@@ -0,0 +1,150 @@
...
+                newlyConflictedState.addConflictedValue(otherConfState);
+            }
+        }
+
+        if (other instanceof UnknownAllocationState) {
+            // Unknown state creates an Illegal ValueAllocationState inside it, because
+            // the unknown state is coming from a different predecessor to the same block,
+            // and it means that this location was not defined there, but it was defined in a
+            // different predecessor block, meaning it's now in a conflicted state, where
+            // it either is defined or it - should not be used in further blocks.
```
gergo-, line 112: As a general style tip, `/* block comments */` are a better fit for multi-line comments because IDEs format them nicely. Personally I also prefer block comment syntax even for single-line comments.

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/core/test/RegAllocVerifierTest.java
```diff
@@ -0,0 +1,1725 @@
...
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static jdk.graal.compiler.lir.LIRInstruction.OperandFlag.REG;
+import static jdk.graal.compiler.lir.LIRInstruction.OperandFlag.STACK;
```
gergo-, line 99: We tend to put static imports at the top. Please run `mx eclipseformat --primary`, let's see if it has an opinion on this. (Make sure you're based on a `master` from the last week or two, that way you should get an Eclipse dependency downloaded automatically. Previously you had to provide a local Eclipse installation, and different versions had different formatting preferences.)

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/core/test/RegAllocVerifierTest.java
```diff
@@ -0,0 +1,1725 @@
...
+        }
+
+        protected String[] getAllocationRestrictedToArray(LIR lir, RegisterAllocationConfig cfg) {
+            var regAllocatedTo = findAllocatedRegister(lir);
+            var allocatableRegs = cfg.getAllocatableRegisters();
+            String[] restrictedTo = new String[allocatableRegs.size() - 1];
+            int i = 0;
+            for (var reg : allocatableRegs) {
+                if (i >= allocatableRegs.size() - 1) {
+                    continue;
```
gergo-, line 222: I think this should be a break. And maybe it would be clearer to check it at the end of the loop body, after incrementing `i`.

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/core/test/RegAllocVerifierTest.java
```diff
@@ -0,0 +1,1725 @@
...
+                }
+
+                restrictedTo[i++] = reg.toString();
+            }
+
+            ignoredReg = regAllocatedTo;
+            return restrictedTo;
+        }
+
+        protected Register findAllocatedRegister(LIR lir) {
```
gergo-, line 236: Could use some Javadoc. This just finds the first allocated register we see in the program?

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/core/test/RegAllocVerifierTest.java
```diff
@@ -0,0 +1,1725 @@
...
+                        public Value doValue(Value value, LIRInstruction.OperandMode mode, EnumSet<LIRInstruction.OperandFlag> flags) {
+                            if (value instanceof RegisterValue regValue) {
+                                allocatedTo[0] = regValue.getRegister();
+                            }
+                            return value;
+                        }
+                    });
+
+                    if (allocatedTo[0] != null) {
+                        break;
```
gergo-, line 261: Could return here, or at least label the outermost loop and do `break outer;`.

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/core/test/RegAllocVerifierTest.java
```diff
@@ -0,0 +1,1725 @@
...
+                        regUsage.put(register, regUsage.getOrDefault(register, 1) + 1);
+                    }
+
+                    return value;
+                }
+            });
+        }
+
+        protected Register getUnusedAllowedRegister(LIR lir, RegisterAllocationConfig cfg) {
+            Map<Register, Integer> usedRegisters = new EconomicHashMap<>();
```
gergo-, line 370: What is the integer used for? It looks to me like this could just be a set of registers. Or is this meant to implement the "only used once as output" semantics? I don't see that. I also don't understand the relationship to the very similar operations on a very similar map in an instance field.

This whole class is unclear to me, it could use more explicit documentation on the class itself, its fields, and the individual operations.

compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/core/test/RegAllocVerifierTest.java
```diff
@@ -0,0 +1,1725 @@
...
+
+            return null;
+        }
+
+    }
+
+    /**
+     * Change the kind of operand to trigger a KindsMismatchException, very simply, find the first
+     * instruction that is not a label and look through its operand array to find the first variable
+     * and change its type to Illegal.
```
gergo-, line 405: Several of these phase wrapper classes use "find the first relevant operand in the first relevant instruction" semantics, maybe that could be factored out as a general helper?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/BlockVerifierState.java
```diff
@@ -0,0 +1,861 @@
...
+        this.block = block;
+    }
+
+    public AllocationStateMap getValues() {
+        return values;
+    }
+
+    /**
+     * Merge states of block and it's predecessor. This process creates a new state based on the
+     * contents of the predecessor, creating conflicts where current locations do not match.
```
gergo-, line 99: Does this create a new state? It seems like it merges the other state into the current one, updating the current one in place.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/BlockVerifierState.java
```diff
@@ -0,0 +1,861 @@
...
+     * Check state values stored in LIRFrameState same way other operands, except we skip those
+     * where the state values are incorrectly stored, after stack allocation, because some values
+     * are no longer present.
+     *
+     * @param op Operation for which we are checking state values
+     */
+    protected void checkStateValues(RAVInstruction.Op op) {
+        if (!op.hasCompleteState()) {
+            // Some values are null after allocation because of stack slot allocator
+            // because it is skipped when iteration (StackLockValue).
```
gergo-, line 118: Where does this skipping of `StackLockValue`s happen? Could you check more specifically for this situation?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/BlockVerifierState.java
```diff
@@ -0,0 +1,861 @@
...
+            // keeps its own registers before and after allocation, but those
+            // can also contain different variable symbols.
+            return;
+        }
+
+        if (ValueUtil.isStackSlot(curr.getValue()) && LIRValueUtil.isVirtualStackSlot(orig.getValue())) {
+            // TestCase: IntegerDivRemCanonicalizationTest
+            // instruction r10|QWORD = STACKLEA slot: stack:80|ILLEGAL[*] in B0
+            // had vstack:0, which is not mentioned in first label or elsewhere
+            // so symbol vstack:0 won't be found
```
gergo-, line 192: So basically you're not checking if the allocation of virtual stack slots to concrete stack slots is consistent in the same way that you check the consistency of the allocation of virtual registers to concrete registers?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/BlockVerifierState.java
```diff
@@ -0,0 +1,861 @@
...
+        }
+
+        AllocationState state = this.values.get(curr);
+        if (orig.equals(curr)) {
+            // For these cases we do not consider checking state taking the original
+            // register as a symbol, because there are too many cases when this does
+            // not work, for example, RETURN with rax tends to contain the actual
+            // generated variable instead of rax symbol, or NEAR_FOREIGN_CALL
+            // keeps its own registers before and after allocation, but those
+            // can also contain different variable symbols.
```
gergo-, line 184: It's not clear to me what "RETURN with rax tends to contain the actual generated variable instead of rax symbol" means. I'm probably missing some general context. You should write a `package-info.java` as an entry point that explains the big picture of how the verifier works and how the main classes work together. See `vectorapi/package-info.java` as an example.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/ConflictResolver.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+
+/**
+ * Resolve {@link ConflictedAllocationState} occurrences based on an internal set of rules.
+ *
+ * <p>
+ * In-case comparison of {@link ValueAllocationState} fails, it also might get resolved by this.
+ * </p>
+ */
+public interface ConflictResolver {
```
gergo-, line 40: This interface only seems to have one implementation. Do you expect there to be more, or could the interface be removed?

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/RegAllocVerifierPhase.java
```diff
@@ -0,0 +1,607 @@
...
+ * the order of spills, reloads, and moves is correct and that variables before allocation are
+ * actually stored in current locations chosen by the allocator.
+ *
+ * <p>
+ * Needs to extend the RegisterAllocationPhase to not throw an exception.
+ * </p>
+ */
+public class RegAllocVerifierPhase extends RegisterAllocationPhase {
+    public static class Options {
+        @Option(help = "Verify that register allocation is indeed, correct", type = OptionType.Debug) public static final OptionKey<Boolean> EnableRAVerifier = new OptionKey<>(true);
```
gergo-, line 73: Please put the `@Option` annotation on a separate line. The formatter will try to put it on the same line again, we usually wrap the whole block of option declarations in `// @formatter:off`/`// @formatter:on`. Alternatively, you can put an empty `//` comment at the end of the line to prevent merging.

gergo- (commented):
I've done a general pass over about half of the code, adding some style and comment remarks. From my point of view, the most important thing for a thorough review would be to have documentation of the "big picture": What is the overall thing we are doing, and which classes cooperate in which way to achieve that goal.

The code seems clean overall, it's clear that you understand what you are doing, now we just need to understand it too :-)

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/exceptions/JavaKindReferenceMismatchException.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public JavaKindReferenceMismatchException(AllocatableValue orig, AllocatableValue curr, JavaKind kind, RAVInstruction.Base instruction, BasicBlock<?> block) {
+        super(getMessage(orig, curr, kind), instruction, block);
+        this.orig = orig;
+        this.curr = curr;
+        this.kind = kind;
+    }
+
+    public static String getMessage(AllocatableValue orig, AllocatableValue curr, JavaKind kind) {
+        if (JavaKind.Object.equals(kind)) {
```
d-kozak: `kind.isObject()` would be a bit cleaner.

compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/lir/alloc/verifier/AllocationState.java
```diff
@@ -0,0 +1,79 @@
...
+ * questions.
+ */
+package jdk.graal.compiler.lir.alloc.verifier;
+
+import jdk.graal.compiler.core.common.cfg.BasicBlock;
+
+/**
+ * Interface for state concrete location is in, stored in {@link AllocationStateMap}.
+ */
+public abstract class AllocationState implements Cloneable {
```
d-kozak: Looks like this could be a `sealed` class?

PR #13301, opened 2026-04-09
Use DebugScope.getName() for language top scope label

The-Alchemist:
### Summary
The Debug Adapter Protocol server always exposed Truffle’s language top scope (from `DebuggerSession.getTopScope`) under the fixed name **`"Global"`**, ignoring each language’s own scope display name.

This change uses **`DebugScope.getName()`** (which already reflects the guest scope’s `toDisplayString` via interop) when building the DAP `Scope` entry. If the name is missing, empty, or `getName()` fails with `DebugException`, the label falls back to **`"Global"`**, preserving previous behavior when a usable name is not available.

### Motivation
Guest languages can already provide a human-readable scope name through their scope object, but the DAP layer discarded it. Showing that name in the variables UI (e.g. namespace- or project-specific labels) improves clarity without changing how variables are resolved.

### Use Case
For [Clojure-on-Truffle–style languages](https://github.com/The-Alchemist/cloffle-clojure), “globals” are usually namespace def / Var bindings, not Java-style globals. The language already exposes that through `TruffleLanguage.getScope()`, and the guest scope object’s `toDisplayString` is a natural place to say which namespace you’re looking at (for example `Namespace: user` or `Namespace: my.app`).

Today the DAP server never showed that string: it always labeled that scope Global in the Variables view, so Clojure users saw a generic bucket name even when the implementation already knew the real namespace.

With this change, `DebugScope.getName()` (which is driven by that toDisplayString) becomes the DAP scope title when it’s non-empty, so the same breakpoint UI can show a namespace-specific label. If the name can’t be resolved, behavior stays Global, matching the old default.

### Risk / compatibility
**Low.** Clients still receive a string scope name; only the default label changes when `getName()` returns a non-empty value. Fallback remains `"Global"` on failure or empty name.B

### DIFF
--- a/tools/src/com.oracle.truffle.tools.dap/src/com/oracle/truffle/tools/dap/server/StackFramesHandler.java
+++ b/tools/src/com.oracle.truffle.tools.dap/src/com/oracle/truffle/tools/dap/server/StackFramesHandler.java
@@ -42,6 +42,9 @@ import java.util.List;
 
 public final class StackFramesHandler {
 
+    /** DAP variables-view label for a guest top scope when no usable guest name is available. */
+    public static final String DEFAULT_TOP_SCOPE_NAME = "Global";
+
     private final ExecutionContext context;
     private final DebuggerSession debuggerSession;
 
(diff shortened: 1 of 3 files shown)


tools/src/com.oracle.truffle.tools.dap/src/com/oracle/truffle/tools/dap/server/StackFramesHandler.java
```diff
@@ -143,7 +143,7 @@ public List<Scope> getScopes(ThreadsHandler.SuspendedThreadInfo info, int frameI
             while (dscope != null) {
                 if (dscope.isFunctionScope() || dscope.getDeclaredValues().iterator().hasNext()) {
                     // provide only scopes that have some variables
-                    scopes.add(Scope.create("Global", info.getId(dscope), true));
+                    scopes.add(Scope.create(topScopeDapName(dscope), info.getId(dscope), true));
```
fniephaus, line 149: 🤖 This changes observable DAP output, but the PR does not update the existing regression coverage. `SimpleLanguageDAPTest.testScopes()` still asserts `Global` at several scope responses, and there is no new DAP test for a language-defined top-scope display name. Please update that test and add a custom-language regression that asserts a non-default `DebugScope.getName()` value flows through here, plus the empty/error fallback.

tools/src/com.oracle.truffle.tools.dap/src/com/oracle/truffle/tools/dap/server/StackFramesHandler.java
```diff
@@ -152,7 +152,24 @@ public List<Scope> getScopes(ThreadsHandler.SuspendedThreadInfo info, int frameI
...
-    public static Variable evaluateOnStackFrame(ThreadsHandler.SuspendedThreadInfo info, int frameId, String expression) throws DebugException {
+    /**
+     * Label for the guest language top scope in the DAP variables view. Prefer
+     * {@link DebugScope#getName()} (guest {@code toDisplayString}) so languages can surface a
+     * meaningful name; fall back to {@code "Global"} if the name is unavailable or empty.
+     */
+    private static String topScopeDapName(DebugScope dscope) {
+        try {
+            String name = dscope.getName();
+            if (name != null && !name.isEmpty()) {
```
fniephaus: 🤖 This only falls back for `null` and the empty string. If a guest scope returns whitespace-only text from `toDisplayString()`, the DAP UI will get a blank label instead of the legacy `Global` fallback. I would trim here, or otherwise reject blank names before returning them.

fniephaus (commented):
Thank you for the contribution and for improving the DAP experience here. I hope you don't mind that I used AI for reviewing and the agent left two inline comments. Please take a look.

The-Alchemist:
> Thank you for the contribution and for improving the DAP experience here. I hope you don't mind that I used AI for reviewing and the agent left two inline comments. Please take a look.

I don't mind at all.  I used AI to help me with the feature itself and the PR description.

I think I can address the comments fairly easily with some additional tests and better verification.

The-Alchemist:
@fniephaus : let me know you think with the latest changes I pushed

fniephaus (approved):
LGTM, will integrate via https://github.com/oracle/graal/pull/13358

PR #13391, opened 2026-04-23
fix(posix): map pthread_key_tPointer to pthread_key_t instead of size_t

### LINKED ISSUE #13386: [Native Image] pthread_setspecific: wrong arguments crash on shutdown — riscv64 only, root cause in PosixPlatformThreads.createUnmanagedThreadLocal()
### Describe the Issue

Running a native-image binary on riscv64, the program runs fine but crashes on the way out:

```
Hello, riscv64!
pthread_setspecific(key, value): wrong arguments
```

The `Hello, riscv64!` line appears, so the binary ran to completion. Shutdown is what fails. Exit code is non-zero; glibc aborts from inside `pthread_setspecific`.

This seems to be riscv64-only. The same binary compiles and exits cleanly on amd64 and aarch64. I spent a while assuming I'd built something wrong before the disassembly pointed somewhere more specific.

### Using the latest version of GraalVM can resolve many issues.

- [ ] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

(No riscv64 distribution exists yet; this issue is part of the effort to enable one.)

### GraalVM Version

Built from `gounthar/graal` fork at commit `3575f11303fd` (tracking 25.1.0-dev), `JAVA_HOME` pointing to labs-openjdk jvmci-25.1-b17 built natively on riscv64. I checked `PosixPlatformThreads.java` in the upstream `oracle/graal` tree and the same code is there.

### Operating System and Version

```
Linux banana-pi-f3 6.6.66-3-deepin-riscv64 #1 SMP RISC-V GNU/Linux
Debian 13 (Trixie), rv64gc, SpacemiT K1
```

### Troubleshooting Confirmation

- [ ] I tried the suggestions in the [troubleshooting guide](https://www.graalvm.org/latest/reference-manual/native-image/guides/troubleshoot-run-time-errors/).

### Run Command

```bash
./helloworld   # native-image binary compiled with --tool:llvm-backend
```

### Expected Behavior

Clean exit, code 0.

### Actual Behavior

```
Hello, riscv64!
pthread_setspecific(key, value): wrong arguments
```

Non-zero exit (SIGABRT from glibc's `pthread_setspecific` key validation).

### Steps to Reproduce

1. Build substratevm for riscv64 using labs-openjdk jvmci-25.1-b17 on a native riscv64 machine
2. Compile a minimal Hello World: `native-image --tool:llvm-backend -cp . HelloWorld -o helloworld`
3. Run `./helloworld`

### Additional Context

**What I think is happening**

As best I can tell, the crash path is: `IsolateThreadCache.clear()` -> `PosixPlatformThreads.setUnmanagedThreadLocalValue()` -> `pthread_setspecific()` with a garbage key.

I think the issue is in `PosixPlatformThreads.createUnmanagedThreadLocal()`. It calls `StackValue.get(Pthread.pthread_key_tPointer.class)` to allocate stack space for the key. Because `pthread_key_tPointer` is `@CPointerTo(nameOfCType = "size_t")`, `StackValue` allocates 8 bytes. `pthread_key_create()` then writes 4 bytes (the actual `pthread_key_t`, which is `unsigned int` on Linux). The subsequent `key.read()` does an 8-byte load, and the upper 32 bits are whatever happened to be on the stack.

Where amd64 and riscv64 seem to diverge: amd64 glibc validates the key with a 32-bit `cmpl`, so the garbage upper bits are never examined. riscv64 glibc uses a 64-bit `bltu`, so a key value like `0x????????00000002` looks >= 1024, `pthread_setspecific` returns `EINVAL`, and glibc aborts.

The disassembly seems consistent with this (from `objdump -d`):

```asm
addi a0,s0,-48          # a0 = &stack_slot (8 bytes, from StackValue.get)
jalr  (pthread_key_create@plt)   # writes 4 bytes into the slot
ld    s2,-48(s0)        # 8-byte load, upper 4 bytes are stack garbage
...
mv    a0,s2             # key with garbage upper bits
jalr  (pthread_setspecific@plt)  # EINVAL on riscv64, silent pass on amd64
```

**What I tried**

Zeroing the stack slot before calling `pthread_key_create`. In `PosixPlatformThreads.java`, `createUnmanagedThreadLocal()`:

```java
Pthread.pthread_key_tPointer key = StackValue.get(Pthread.pthread_key_tPointer.class);
// pthread_key_t is unsigned int (4 bytes) but StackValue allocates size_t (8 bytes).
// Zero the slot so the upper 4 bytes are clean on riscv64.
((WordPointer) key).write(Word.zero());
PosixUtils.checkStatusIs0(Pthread.pthread_key_create(key, Word.nullPointer()), ...);
```

With this change, the binary ran and exited cleanly on the F3. I'm not confident this is the right fix though. The deeper issue may be that `pthread_key_tPointer` is typed as `size_t` when the underlying C type is `unsigned int`; fixing the typedef might be cleaner, though I haven't worked out whether that ripples elsewhere. I'd rather flag it and be wrong than stay quiet.

Worth checking other `StackValue.get(Pthread.pthread_key_tPointer.class)` call sites too; if this pattern appears elsewhere, the same truncation issue would be there.

I'm not a SubstrateVM contributor and this is my first real read of this part of the codebase. I came at it from the riscv64 side, trying to get native-image working on the Banana Pi F3 as part of getting GraalVM onto riscv64 at all. I've signed (or will sign before this is acted on) the OCA.

If the fix is in the wrong place or the wrong approach, I'm happy to test patches on hardware; the F3 is available.

gounthar:
## Summary

`StackValue.get(pthread_key_tPointer.class)` allocates `size_t` bytes (8 on 64-bit) because `pthread_key_tPointer` is `@CPointerTo(nameOfCType = "size_t")`. `pthread_key_create` writes only the 4-byte `pthread_key_t` value (`unsigned int` on Linux). Then `key.read()` does an 8-byte load, so whatever was sitting on the stack in the upper 32 bits gets folded into the returned `ThreadLocalKey`.

On amd64 this is silent: glibc validates the key with a 32-bit comparison and never touches the upper word. On riscv64, glibc uses a 64-bit `bltu` against `PTHREAD_KEYS_MAX` (1024). A key like `0xdeadbeef00000002` fails that check, `pthread_setspecific` returns `EINVAL`, and glibc aborts during `IsolateThreadCache.clear()` at shutdown:

    pthread_setspecific(key, value): wrong arguments

Zeroing the slot before the call fixes this on my setup. I'm not sure it's the right approach though. Zeroing at allocation time might be cleaner, or maybe `pthread_key_tPointer` should be sized to reflect the actual 4-byte type rather than `size_t`. Happy to take direction from anyone who knows this area better.

## Related Issues

Fixes #13386

## Testing

Tested on a single Banana Pi F3 (SpacemiT K1, rv64gc, Debian). With the fix, a HelloWorld native binary runs and exits cleanly. Without it, it prints output and then aborts with the error above.

I haven't tested other riscv64 boards or glibc versions, so there may be cases I haven't hit. On amd64 the change should be a no-op since the 32-bit glibc comparison never reads the upper bytes, but I haven't run a full test suite to confirm nothing regresses there. A proper regression test would need a riscv64 CI runner, which I don't have access to.

## Documentation

Nothing to update. Runtime crash fix, no API change.

## Contributor Checklist

### DIFF
--- a/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/Pthread.java
+++ b/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/headers/Pthread.java
@@ -29,6 +29,7 @@ import org.graalvm.nativeimage.c.constant.CConstant;
 import org.graalvm.nativeimage.c.function.CFunction;
 import org.graalvm.nativeimage.c.function.CFunction.Transition;
 import org.graalvm.nativeimage.c.function.CLibrary;
+import org.graalvm.nativeimage.c.struct.AllowWideningCast;
 import org.graalvm.nativeimage.c.struct.CPointerTo;
 import org.graalvm.nativeimage.c.struct.CStruct;
 import org.graalvm.nativeimage.c.type.VoidPointer;
@@ -78,8 +79,9 @@ public class Pthread {
     public interface pthread_key_t extends UnsignedWord {
     }
 
-    @CPointerTo(nameOfCType = "size_t")
+    @CPointerTo(nameOfCType = "pthread_key_t")
     public interface pthread_key_tPointer extends PointerBase {
+        @AllowWideningCast
         pthread_key_t read();
     }
 


oubidar-Abderrahim:
Thanks for the contribution!

We'll review this shortly. @wirthi, could you please assign a reviewer?

gounthar:
Appreciate the quick acknowledgment, looking forward to the review.

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -2703,6 +2703,9 @@ def suppress_gnu_property_type_5_warnings(line):
                 mx.run(symbol_dump_command.split() + [staticlib_path], out=collect_symbols_fn('JVM_'), err=suppress_gnu_property_type_5_warnings)
 
             if len(symbols) == 0:
+                if not self.staticlibs:
+                    mx.warn('No static JDK libraries found for riscv64 - generating empty JvmFuncsFallbacks.c')
```
Zeavee: I don't think this warning should mention `riscv64`.

substratevm/mx.substratevm/mx_substratevm.py
```diff
@@ -2814,9 +2817,15 @@ def mx_register_dynamic_suite_constituents(register_project, register_distributi
     else:
         lib_prefix = mx.add_lib_prefix('')
         lib_suffix = mx.add_static_lib_suffix('')
-        layout = {
-            './': ['file:' + join(base_jdk_home, 'lib', lib_prefix + '*' + lib_suffix)]
-        }
+        from glob import glob as _glob
```
Zeavee: Please put this at the top of the file.

substratevm/mx.substratevm/mx_substratevm.py
Zeavee, line 1: I'm quite curious why you needed that. What jdk did you use to build native image?

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/thread/PosixPlatformThreads.java
```diff
@@ -275,6 +275,10 @@ public boolean supportsUnmanagedThreadLocal() {
     @Uninterruptible(reason = "Called from uninterruptible code.", mayBeInlined = true)
     public ThreadLocalKey createUnmanagedThreadLocal() {
         Pthread.pthread_key_tPointer key = StackValue.get(Pthread.pthread_key_tPointer.class);
+        // StackValue allocates size_t (8 bytes) but pthread_key_t is unsigned int (4 bytes).
+        // Zero the slot so the upper 32 bits are clean; otherwise a 64-bit key.read() picks
+        // up stack garbage, and riscv64 glibc rejects the resulting key value via pthread_setspecific.
+        ((WordPointer) key).write(Word.zero());
```
Zeavee: Instead of zeroing the value, maybe it would make more sense to have the `nameOfCType` of `pthread_key_tPointer` be `pthread_key_t`.

substratevm/mx.substratevm/mx_substratevm.py
gounthar: Fixed. Replaced the hardcoded `'riscv64'` with `mx.get_arch()` so the warning isn't architecture-specific.

substratevm/mx.substratevm/mx_substratevm.py
gounthar: Fixed. Removed the local import. `glob` is already at the top of the file; I clearly didn't look far enough.

substratevm/mx.substratevm/mx_substratevm.py
gounthar: Initially I used `openjdk-25-jdk-headless` from Debian apt on a Banana Pi F3 runner. That JDK ships with no static `.a` libraries, which is exactly why the `JvmFuncsFallbacks` build task hits that branch and why I changed `mx.abort()` to `mx.warn()`. Later I rebuilt `labs-openjdk` for riscv64 from source (tracked in labs-openjdk#34), which produced 38 static `.a` libs. With those in place the build gets further.

Should have put this in the PR description. My mistake.

substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/thread/PosixPlatformThreads.java
gounthar: Agreed, cleaner. Changing `nameOfCType` from `"size_t"` to `"pthread_key_t"` fixes it at the type level rather than patching around it in `createUnmanagedThreadLocal`. I want to test this on the F3 first since that's exactly where the `@CPointerTo`/`StackValue`/`pthread_setspecific` interaction breaks. Will update once verified. You probably know better than I do whether this could affect other platforms too.

substratevm/mx.substratevm/mx_substratevm.py
Zeavee: No worries!

I don't know if we really want to support building graal without the static `.a` libraries though. At least this change would be out of the scope of this PR, and would be touching all the architectures, so maybe it's better to keep this one focused on the `pthread` issue.

gounthar: Agreed, pulled it out. I force-pushed so the branch is now just the pthread_key_t mapping and nothing else: nameOfCType = "pthread_key_t" with @AllowWideningCast on read(), which is what you suggested, and it drops the stack-slot zeroing completely.

The static-lib handling and the libffi arch mapping were really separate riscv64 build concerns, so I will move those to their own PRs if they turn out to be worth it. You know the substratevm side better than I do, so tell me if you would rather I split things differently. Thanks for the steer.

Zeavee (approved):
Looks good to me, thank you!

PR #13454, opened 2026-05-01
Fix incorrect OtherContextException wrapper when the exception comes back to its creator context

### LINKED ISSUE #13453: Exception from inner context is incorrectly wrapped, causing AssertionError and infinite cause chains
 - GraalVM version or commit id if built from source: **[e.g. 19.3]**
 - CE or EE: EE
 - JDK version: 25.0.2
 - OS and OS Version: Linux
 - Architecture: amd64

Yes, because the code still has the same missing logic.

**Describe the issue**

From https://github.com/truffleruby/truffleruby/issues/4233

Exceptions from an inner/other context are wrapped in `OtherContextException`.
The bug is when the exception comes back to the context that created it it's not wrapped, it's still in a `OtherContextException`.
That both triggers Truffle assert's:
```
no need for foreign value if contexts match (java.lang.AssertionError)
	from com.oracle.truffle.polyglot.OtherContextGuestObject$OtherContextException.<init>(OtherContextGuestObject.java:342)
	from com.oracle.truffle.polyglot.OtherContextGuestObject.migrateException(OtherContextGuestObject.java:187)
	from com.oracle.truffle.polyglot.EngineAccessor$EngineImpl.evalBoundary(EngineAccessor.java:944)
	from com.oracle.truffle.polyglot.EngineAccessor$EngineImpl.evalInternalContext(EngineAccessor.java:920)
	from com.oracle.truffle.api.TruffleContext.evalPublic(TruffleContext.java:358)
	from org.truffleruby.interop.PolyglotNodes$InnerContextEvalNode.eval(PolyglotNodes.java:282)
	from org.truffleruby.interop.PolyglotNodes$InnerContextEvalNode.evalCached(PolyglotNodes.java:253)
	from org.truffleruby.interop.PolyglotNodesFactory$InnerContextEvalNodeFactory$InnerContextEvalNodeGen.executeAndSpecialize(PolyglotNodesFactory.java:1037)
	from org.truffleruby.interop.PolyglotNodesFactory$InnerContextEvalNodeFactory$InnerContextEvalNodeGen.execute(PolyglotNodesFactory.java:953)
	from org.truffleruby.language.control.SequenceNode.execute(SequenceNode.java:36)
	from org.truffleruby.language.RubyMethodRootNode.execute(RubyMethodRootNode.java:66)
```

and more subtle and messy problems like stack overflows due to infinite cause chains because these etra wrappers break identity (`InteropLibrary#isIdentical`).

* Create an inner context
* Pass a lambda to the inner context, the lambda should throw an exception of the outer context
* Call that lamba in the inner context
* Catch the exception in the outer context, the exception object must be the initial outer exception, but it's not, that's the bug

eregon:
## Summary

Fixes https://github.com/oracle/graal/issues/13453

## Testing

I wrote a test for this in `truffle/src/com.oracle.truffle.api.instrumentation.test/src/com/oracle/truffle/api/instrumentation/test/TruffleContextTest.java` with the help of Codex. It passes with the fix and fails without.

I also have tests in TruffleRuby for this: https://github.com/truffleruby/truffleruby/pull/4235/changes
I confirm this PR fixes https://github.com/oracle/graal/issues/13453, as in those tests fail without and pass with this PR.

## Documentation

No documentation updates are needed.

## Contributor Checklist

### DIFF
--- a/truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/OtherContextGuestObject.java
+++ b/truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/OtherContextGuestObject.java
@@ -338,13 +338,13 @@ final class OtherContextGuestObject implements TruffleObject {
         }
 
         @TruffleBoundary
-        OtherContextException(PolyglotContextImpl thisContext, Exception delegate, PolyglotContextImpl delegateContext) {
+        OtherContextException(PolyglotContextImpl receiverContext, Exception delegate, PolyglotContextImpl delegateContext) {
             super(delegate.getMessage());
             assert !(delegate instanceof OtherContextException) : "recursive host foreign value found";
-            assert thisContext != null && delegateContext != null : "Must have associated contexts.";
-            assert thisContext != delegateContext : "no need for foreign value if contexts match";
+            assert receiverContext != null && delegateContext != null : "Must have associated contexts.";
+            assert receiverContext != delegateContext : "no need for foreign value if contexts match";
             this.delegate = delegate;
-            this.receiverContext = thisContext;
+            this.receiverContext = receiverContext;
             this.delegateContext = delegateContext;
         }
 
--- a/truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/OtherContextGuestObject.java
+++ b/truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/OtherContextGuestObject.java
@@ -181,10 +181,16 @@ final class OtherContextGuestObject implements TruffleObject {
     @TruffleBoundary
     static <T extends Throwable> RuntimeException migrateException(PolyglotContextImpl receiverContext, Throwable e, PolyglotContextImpl valueContext) throws T {
         if (e instanceof OtherContextException) {
+            // Same logic as in migrateValue()
             OtherContextException other = (OtherContextException) e;
             if (other.receiverContext == receiverContext && other.delegateContext == valueContext) {
+                // reuse wrapper it is already wrapped
                 throw other;
+            } else if (other.receiverContext == valueContext && other.delegateContext == receiverContext) {
+                // unpack foreign value it belongs to that context
+                throw (T) other.delegate;
             } else {
+                // Preserve original context of the delegate when forwarding through third context
                 throw new OtherContextException(receiverContext, other.delegate, other.delegateContext);
             }
         } else if (InteropLibrary.getUncached().isException(e)) {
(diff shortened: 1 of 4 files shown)


truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/OtherContextGuestObject.java
```diff
@@ -181,10 +181,16 @@ static Object sendImpl(Node node, PolyglotSharingLayer layer, Object receiver, M
...
             OtherContextException other = (OtherContextException) e;
             if (other.receiverContext == receiverContext && other.delegateContext == valueContext) {
+                // reuse wrapper it is already wrapped
                 throw other;
+            } else if (other.receiverContext == valueContext && other.delegateContext == receiverContext) {
+                // unpack foreign value it belongs to that context
+                throw (T) other.delegate;
             } else {
+                // TODO: Should the last arg be valueContext like in migrateValue?
                 throw new OtherContextException(receiverContext, other.delegate, other.delegateContext);
```
eregon: migrateValue() does:
https://github.com/oracle/graal/blob/3c0e0b07214f3150d340d34206cfbd3263cfd497/truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/PolyglotContextImpl.java#L1731
(in migrateValue(), this == receiverContext)

Which is correct?
I'm not sure what the `receiverContext` is for, wouldn't only tracking the `delegateContext` be enough?
Is it because the OtherContextGuestObject/OtherContextException objects need to belong to one context and not many? (if so, is that needed?)

oubidar-Abderrahim:
Thank you for the PR @eregon 
@chumer could you please review, thanks

truffle/src/com.oracle.truffle.polyglot/src/com/oracle/truffle/polyglot/OtherContextGuestObject.java
eregon: I (and Codex) have figured this out, it was another bug, this time in migrateValue().
Pushed a commit with a fix + test and verified the test passes with the fix and fails without.

eregon:
@chumer @tzezula I wrote a test now with the help of Codex (and reviewed it), and also addressed the comment above which turned out to be a second bug, so now this PR fixes 2 bugs and adds a test for each. Please review.

eregon:
Ah during rebasing I noticed @iamstolis already fixed the second bug in 5b5c93ad6f8127c26d503ae47695fe85fe4f7f8c / #13522.
I'll drop my last commit then as that's already fixed & tested.

tzezula (approved):
Looks good to me.
It unwraps the object originated in the targetContext rather than creating a decorator chain.

chumer:
lgtm. inbounding the PR. sorry for the delay.

chumer (approved):

chumer:
PR is merged here: https://github.com/oracle/graal/pull/13599
(will reuse your branch next time, to track this properly, sorry)

Issue #13351, opened 2026-04-17
native-image: add riscv64 (linux-riscv_64) platform support


gounthar:
**Feature request**

**Is your feature request related to a problem? Please describe.**

Building native executables for linux-riscv64 is not possible with GraalVM Native Image. No GraalVM distribution exists for riscv64, and native-image cannot cross-compile to riscv64 from other hosts.

Concrete impact: JReleaser recently merged support for recognizing `linux-riscv_64` as a valid platform ([jreleaser/jreleaser#2062](https://github.com/jreleaser/jreleaser/pull/2062)), but cannot ship its own native riscv64 binary release because GraalVM does not support that target. The JVM-based standalone packages work, but the native binary distribution gap remains.

**Describe the solution you'd like.**

A GraalVM JDK distribution for `linux-riscv64` with `native-image` support, equivalent to the existing `linux-aarch64` offering. This would allow projects using GraalVM to build and distribute native binaries for riscv64 targets.

**Describe who do you think will benefit the most.**

- Java projects producing native binaries (JReleaser, Quarkus native, Micronaut, etc.) that target riscv64
- The growing riscv64 ecosystem where native startup time and memory footprint matter
- CI pipelines running on native riscv64 hardware (runners available today via RISE: `ubuntu-24.04-riscv`)

**Describe alternatives you've considered.**

- Mandrel: same gap, no riscv64 distribution as of 25.0.2.0-Final
- Cross-compilation: native-image does not support cross-compilation in the general case
- JVM-based fallback: works but loses the native binary performance benefits

**Additional context.**

Native RISC-V CI runners are available today via the RISE project (`ubuntu-24.04-riscv` label, GitHub App: https://github.com/apps/rise-risc-v-runners). Infrastructure is not the blocker - the GraalVM substrate port is.

riscv64 adoption is accelerating: Debian, Ubuntu, and Fedora all have stable riscv64 ports. Hardware like the Banana Pi F3 (SpacemiT K1) and server-class RISC-V boards are increasingly used for development and CI.

**Express whether you'd like to help contributing this feature**

I can help with providing access to RISE riscv64 runners for testing once development is underway.

wirthi:
[...]

PR #13826, opened 2026-06-18
fix(riscv64): implement fixed-parameter and return-saving calling convention

gounthar:
## Summary

Master restored the LLVM backend with a riscv64 target, but `SubstrateRISCV64RegisterConfig.getCallingConvention` still throws "Fixed parameter assignments and return saving are not yet supported on this platform" for custom-ABI calls. The first foreign call during image generation (`throwNewNullPointerException`) trips it, so native-image cannot get past "Compiling methods" on riscv64.

This implements the fixed-parameter-assignment and return-saving path for riscv64, mirroring the existing AArch64 register config: the return-buffer placeholder parameter is parked in x28 (t3) where AArch64 uses r8, and custom-ABI parameters are assigned to their fixed register/stack locations. The standard-ABI path is unchanged in behaviour, only moved under the `!type.customABI()` branch.

## Related Issues

Part of reviving the riscv64 LLVM backend, discussed in #13351.

## Testing

Built substratevm for linux-riscv64 on a Banana Pi F3 (SpaceMiT K1) and ran native-image on a HelloWorld with `--tool:llvm-backend --no-fallback`. Before this change the build aborted in `getCallingConvention` at the first foreign call; after it, the build proceeds through all method compilation. The image does not yet link on riscv64 due to a separate, independent issue (the in-process object reader still uses the LLVM 13 bytedeco preset against LLVM 20.1.4 objects, discussed in #13351).

## Documentation

No documentation changes needed.

## Contributor Checklist

- [ ] If I used a coding assistant, I remain responsible for the entire contribution and have reviewed it accordingly.

### DIFF
--- a/substratevm/src/com.oracle.svm.core.graal.riscv64/src/com/oracle/svm/core/graal/riscv64/SubstrateRISCV64RegisterConfig.java
+++ b/substratevm/src/com.oracle.svm.core.graal.riscv64/src/com/oracle/svm/core/graal/riscv64/SubstrateRISCV64RegisterConfig.java
@@ -69,6 +69,7 @@ import static jdk.vm.ci.riscv64.RISCV64.x24;
 import static jdk.vm.ci.riscv64.RISCV64.x25;
 import static jdk.vm.ci.riscv64.RISCV64.x26;
 import static jdk.vm.ci.riscv64.RISCV64.x27;
+import static jdk.vm.ci.riscv64.RISCV64.x28;
 import static jdk.vm.ci.riscv64.RISCV64.x3;
 import static jdk.vm.ci.riscv64.RISCV64.x8;
 import static jdk.vm.ci.riscv64.RISCV64.x9;
@@ -76,10 +77,12 @@ import static jdk.vm.ci.riscv64.RISCV64.x9;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.graalvm.collections.EconomicSet;
 import org.graalvm.nativeimage.Platform;
 
 import com.oracle.svm.core.ReservedRegisters;
 import com.oracle.svm.core.config.ObjectLayout;
+import com.oracle.svm.core.graal.code.AssignedLocation;
 import com.oracle.svm.core.graal.code.SubstrateCallingConvention;
 import com.oracle.svm.core.graal.code.SubstrateCallingConventionKind;
 import com.oracle.svm.core.graal.code.SubstrateCallingConventionType;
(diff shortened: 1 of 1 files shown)


substratevm/src/com.oracle.svm.core.graal.riscv64/src/com/oracle/svm/core/graal/riscv64/SubstrateRISCV64RegisterConfig.java
```diff
@@ -237,54 +258,86 @@ public CallingConvention getCallingConvention(Type t, JavaType returnType, JavaT
...
+                    }
+
+                }
+                if (register != null) {
+                    boolean useJavaKind = isEntryPoint && Platform.includedIn(Platform.LINUX.class);
+                    locations[i] = register.asValue(valueKindFactory.getValueKind(useJavaKind ? kind : kind.getStackKind()));
+                } else {
+                    if (type.nativeABI()) {
+                        if (Platform.includedIn(Platform.LINUX.class)) {
+                            ValueKind<?> valueKind = valueKindFactory.getValueKind(type.outgoing ? kind.getStackKind() : kind);
```
Zeavee: Please create the `linuxNativeStackParameterAssignment` method since this is also mirrored on `AArch64`

Zeavee (approved):
Looks good to me, thank you!

Please address the small comment and I will merge the PR.

substratevm/src/com.oracle.svm.core.graal.riscv64/src/com/oracle/svm/core/graal/riscv64/SubstrateRISCV64RegisterConfig.java
gounthar: Done in ee19375 — pulled the inline Linux native block into `linuxNativeStackParameterAssignment`, matching the AArch64 method. No behaviour change, and substratevm still builds clean on riscv64. Thanks for the review.

substratevm/src/com.oracle.svm.core.graal.riscv64/src/com/oracle/svm/core/graal/riscv64/SubstrateRISCV64RegisterConfig.java
Zeavee, line 357: After looking more, can you please also copy the `AArch64` logic for the `returnLocation` here. This way we will also support “return-saving” custom ABI support.

gounthar: Done in 6591803. I mirrored the AArch64 returnLocation block, so a custom ABI with a single-register returnSaving slot sends the return value to that register, and falls back to the standard return register otherwise. substratevm still compiles clean on the F3 (riscv64). Thanks for catching it.

PR #13854, opened 2026-06-23
[GR-77370] Clean up more defensively after SVM helloworld test to avoid masking the real test Error

simonis:
## Summary

- While running the gate tests for the Shenandoah port, I run into this issue. The problem is that if loading the helloworld shared library fails, there will be an error in the final block which will mask the original test failure. With these changes, the original error will be reported.

## Related Issues

- N/A

## Testing

- Ran `mx --primary-suite=substratevm gate --strict-mode --tags helloworld` with both, Serial and Shenandoah GC

## Documentation

- Test infra changes - no documentation updates.

## Contributor Checklist


wirthi:
has been merged. 

https://github.com/oracle/graal/commit/5bbbbc395bcbcb9a5a7fda46884232b47cc8b01a

PR #13902, opened 2026-07-01
[GR-76772] Support AMD64 pairwise vector multiply add

graalvmbot:


farnoy:
@gergo- I'm curious: is there a separate suite of tests that are not part of the repo? How was this tested and reviewed? I'm guessing there's an internal review tool & process that's not public?

gergo-:
Yes, the relevant tests are internal, as is the relevant review process. If you're very interested in reviewing compiler code, I can mirror pull requests to GitHub earlier.

farnoy:
No need, I wouldn't be able to meaningfully review it either way. But if I did try to contribute something smaller to get more familiar, I'm at a disadvantage not having the test suite.

gergo-:
I understand; this decision is above my pay grade. For whatever it's worth, if you do ever want to try contributing in this area, you can use tests in the style of https://github.com/oracle/graal/blob/master/compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/core/test/VectorAPISmokeTest.java.

PR #14013, opened 2026-07-20
[GR-65437] [GR-77499] Use containing method and bci to make lambda name stable

graalvmbot:
Make lambda proxy names stable across builds by deriving their identity from the lambda capture site (capturing method@bci) rather than the VM-generated lambda class name.

For layered images, persist that capture site in the layer snapshot. The application layer uses it to resolve and load only the requested lambda type, instead of loading every lambda from the capturing class.

The lambda parser caches capture-site metadata and, when needed by a layer load, parses the relevant capture method once and maps its invokedynamic BCIs to lambda classes.


jerboaa:
@Zeavee This PR causes: https://github.com/oracle/graal/issues/14314 Please take a look. Thanks!

PR #14195, opened 2026-08-07
[Web Image] Fix unhandled UNSTRUCTURED_LOCKING and an order-dependent single-abstract-method failure

### LINKED ISSUE #14192: [Web Image] --tool:svm-wasm: "Incorrect single abstract method logic" for a non-functional interface reaching the JS boundary
### Describe the Issue

Building with --tool:svm-wasm fails during analysis as soon as the reachable graph contains an interface with two abstract methods, even though that interface is never passed to JavaScript and is never instantiated.

InterceptJSInvokeTypeFlow.onObservedUpdate computes the single abstract method of every type observed flowing into a JS invoke, and ReflectUtil.singleAbstractMethodForInterface guarantees each such interface has exactly one. The type it fails on is com.intellij.util.containers.RefHashMap.Key, reported as instantiated: false, reachable: true, which declares get() and hashCode(). The type flow appears to be wider than the set of types that can actually cross the boundary. The program exports exactly one function.

Which interface is named varies between runs. I have also seen ConcurrentRefHashMap$KeyReference with the same two candidates, consistent with the analysis being parallel and the first offender winning.

Ruled out before filing:

Not the exported type being too generic. The export is BiFunction<JSString, JSString, JSString>, which erases to apply(Object, Object). Replacing it with a purpose-declared @FunctionalInterface whose method is JSString apply(JSString, JSString), with no Object in the erased signature, does not help. The build fails the same way on the next such interface.

Not avoidable by exporting another way. The Web Image API guide lists @JS.Export as "not implemented yet" and JSObject subclasses as "not supported yet", so the @JS helper pattern that triggers this is the only documented way to expose a function to JavaScript.

Not reproducible in the small. A standalone program that exports the same BiFunction while keeping a reachable, uninstantiated two-abstract-method interface in the graph builds fine. The trigger seems flow-sensitive rather than "any such interface is reachable", which is why the reproducer uses a real dependency rather than a ten-line file. Happy to keep minimizing if that would help.

### Using the latest version of GraalVM can resolve many issues.

- [x] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

### GraalVM Version

java version "25.0.4" 2026-07-21 LTS
Java(TM) SE Runtime Environment Oracle GraalVM 25.2.4+7.1 (build 25.0.4+7-LTS-jvmci-25.2-b20)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 25.2.4+7.1 (build 25.0.4+7-LTS-jvmci-25.2-b20, mixed mode, sharing

### Operating System and Version

Linux 6.18.5 #1 SMP PREEMPT_DYNAMIC x86_64 GNU/Linux

### Build Command

native-image --tool:svm-wasm
-cp "classes:ni-classes:$NI/resources:ktfmt.jar"
-o kotlin_fmt --no-fallback
-H:+UnlockExperimentalVMOptions -H:+ReportExceptionStackTraces
--initialize-at-build-time=com.facebook
--initialize-at-build-time="$(rl $NI/initialize-at-build-time.txt)"
--initialize-at-run-time="$(rl $NI/initialize-at-run-time.txt)"
--initialize-at-run-time=com.facebook.ktfmt.format.Parser
-J--add-opens=java.base/java.util=ALL-UNNAMED
KotlinFmt

### Expected Behavior

Analysis completes and a WasmGC module is produced, rather than failing on an internal collection interface that the program never exposes to JavaScript.

### Actual Behavior

Analysis aborts after about 30 seconds with a failed guarantee. No image is produced.

com.oracle.graal.pointsto.util.ParallelExecutionException: failed guarantee: Incorrect single abstract method logic for interface AnalysisType<RefHashMap$Key -> HotSpotType<Lorg/jetbrains/kotlin/com/intellij/util/containers/RefHashMap$Key;, resolved>, instantiated: false, reachable: true>. Found at least two candidate methods: AnalysisMethod<RefHashMap$Key.get -> HotSpotMethod<RefHashMap$Key.get()>, invoked: false, implInvoked: false, intrinsic: false, inlined: false> and AnalysisMethod<RefHashMap$Key.hashCode -> HotSpotMethod<RefHashMap$Key.hashCode()>, invoked: false, implInvoked: false, intrinsic: false, inlined: false>

The failing guarantee is in ReflectUtil.singleAbstractMethodForInterface, reached from InterceptJSInvokeTypeFlow.onObservedUpdate. Full trace in the Build Log Output field below.

### Steps to Reproduce

1. Download ktfmt 0.64 and strip the jline metadata from it. The fat jar carries jline native-image.properties files naming reflection configs that the shading dropped; this is unrelated to the bug, but native-image aborts at [1/8] Initializing without it.

curl -fsSLO https://repo1.maven.org/maven2/com/facebook/ktfmt/0.64/ktfmt-0.64-with-dependencies.jar
cp ktfmt-0.64-with-dependencies.jar ktfmt.jar
zip -q -d ktfmt.jar 'META-INF/native-image/org.jline/*'

2. Fetch ktfmt's own native-image metadata from upstream main. It is newer than the 0.64 release and supplies the reachability metadata, the initialization lists, and a substitution class.

git clone --depth 1 --filter=blob:none --sparse https://github.com/facebook/ktfmt
git -C ktfmt sparse-checkout set core/src/main/native-image
NI=ktfmt/core/src/main/native-image

3. Write the entry point. Web Image has no @JS.Export, so this is the documented helper pattern: pass a functional interface into a native @JS method whose body parks it on globalThis.

cat > KotlinFmt.java <<'EOF'
import com.facebook.ktfmt.format.Formatter;
import java.util.function.BiFunction;
import org.graalvm.webimage.api.JS;
import org.graalvm.webimage.api.JSString;

public final class KotlinFmt {
@JS(args = {"format"}, value = "globalThis.__f = format;")
private static native void exportFormat(BiFunction<JSString, JSString, JSString> format);

exportFormat((source, options) -> JSString.of(Formatter.format(source.asString())));
}
}
EOF

4. Compile the entry point and the substitution class.

javac --add-modules org.graalvm.webimage.api -cp ktfmt.jar -d classes KotlinFmt.java
javac -cp "ktfmt.jar:$JAVA_HOME/lib/svm/builder/svm.jar" -d ni-classes $NI/java/com/facebook/ktfmt/nativeImage/*.java

5. Run the build command from the Build Command field, with this helper defined first for the initialization lists.

rl() { grep -v '^#' "$1" | grep -v '^$' | tr '\n' ',' | sed 's/,$//'; }

The build fails in about 30 seconds. Note that --initialize-at-run-time=com.facebook.ktfmt.format.Parser is required because ktfmt 0.64 builds its KotlinCoreEnvironment eagerly, which --initialize-at-build-time=com.facebook would otherwise place in the image heap. It is unrelated to this failure.

### Additional Context

Trying to create a wasm version of the formatter 

### Build Log Output and Error Messages

_No response_

### LINKED ISSUE #14193: [Web Image] --tool:svm-wasm: unhandled UNSTRUCTURED_LOCKING in WebImageImplicitExceptionsFeature aborts code generation
### Describe the Issue

WebImageImplicitExceptionsFeature.getSupportMethodName maps thirteen BytecodeExceptionKinds to runtime support methods and falls through to shouldNotReachHere for UNSTRUCTURED_LOCKING, so any image whose reachable graph produces a BytecodeExceptionNode of that kind cannot be built.

The thirteen handled kinds are createNewArrayStoreExceptionWithArgs, createNegativeArraySizeException, createNewClassCastExceptionWithArgs, createNewDivisionByZeroException, createNewArithmeticException, createNewNullPointerException, createNewOutOfBoundsExceptionWithArgs, createNewOutOfBoundsException, createNewIncompatibleClassChangeError, createNewNegativeLengthException, createNewArgumentIsNotArrayException, createNewAssertionErrorObject and createNewAssertionErrorNullary.

This looks like a simply unimplemented case rather than a configuration problem. Disassembling lib/svm/tools/svm-wasm/builder/svm-wasm.jar shows the switch directly, and no build flag avoids it. Something in the reachable graph, the Kotlin compiler frontend or the intellij-core it is built on, has monitor structure the parser cannot recover, so a node of that kind is created and the Wasm backend has no lowering for it.

One request beyond the fix. The generated diagnostic report (svm_err_b_*.md) does not name the method that produced the node. The feature fails while iterating registered bytecode exceptions, with no enclosing-method context anywhere in the error. If there is a flag that prints the owning method for each registered BytecodeExceptionNode, that would make it possible to work around this from outside by substituting the offending method. As it stands there is nothing actionable to go on.

Related: https://github.com/oracle/graal/issues/14192

### Using the latest version of GraalVM can resolve many issues.

- [x] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

### GraalVM Version

Where the trace comes from:

java version "25.0.4" 2026-07-21 LTS
Java(TM) SE Runtime Environment Oracle GraalVM 25.0.4+7.1 (build 25.0.4+7-LTS-jvmci-b01)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 25.0.4+7.1 (build 25.0.4+7-LTS-jvmci-b01, mixed mode, sharing)

Also checked, and still carrying the unhandled case:

java version "25.0.4" 2026-07-21 LTS
Java(TM) SE Runtime Environment Oracle GraalVM 25.2.4+7.1 (build 25.0.4+7-LTS-jvmci-25.2-b20)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 25.2.4+7.1 (build 25.0.4+7-LTS-jvmci-25.2-b20, mixed mode, sharing)

### Operating System and Version

Linux 6.18.5 #1 SMP PREEMPT_DYNAMIC x86_64 GNU/Linux

### Build Command

ative-image --tool:svm-wasm
-cp "classes:ni-classes:$NI/resources:ktfmt.jar"
-o kotlin_fmt --no-fallback
-H:+UnlockExperimentalVMOptions -H:+ReportExceptionStackTraces
--initialize-at-build-time=com.facebook
--initialize-at-build-time="$(rl $NI/initialize-at-build-time.txt)"
--initialize-at-run-time="$(rl $NI/initialize-at-run-time.txt)"
--initialize-at-run-time=com.facebook.ktfmt.format.Parser
-J--add-opens=java.base/java.util=ALL-UNNAMED
KotlinFmt

### Expected Behavior

Code generation completes and a WasmGC module is produced, with UNSTRUCTURED_LOCKING lowered like the other BytecodeExceptionKinds.

### Actual Behavior

Analysis completes: 12,320 types, 17,128 fields and 57,270 methods reachable, 4,291 types registered for reflection. Code generation then aborts after about a minute, and no image is produced.

com.oracle.svm.core.util.VMError$HostedError: InternalFeature defined by com.oracle.svm.hosted.webimage.codegen.WebImageImplicitExceptionsFeature unexpectedly failed with a(n) jdk.graal.compiler.debug.GraalError

Caused by: jdk.graal.compiler.debug.GraalError: should not reach here: unexpected value: UNSTRUCTURED_LOCKING

### Steps to Reproduce

Identical to the sibling issue except for step 0. Run all of it with GraalVM 25.0.4 LTS from https://download.oracle.com/graalvm/25/archive/graalvm-jdk-25.0.4_linux-x64_bin.tar.gz — on 25.2.4 the build stops earlier and never reaches this 

1. Download ktfmt 0.64 and strip the jline metadata from it. The fat jar carries jline native-image.properties files naming reflection configs that the shading dropped; unrelated to this bug, but native-image aborts at [1/8] Initializing without it.

curl -fsSLO https://repo1.maven.org/maven2/com/facebook/ktfmt/0.64/ktfmt-0.64-with-dependencies.jar
cp ktfmt-0.64-with-dependencies.jar ktfmt.jar
zip -q -d ktfmt.jar 'META-INF/native-image/org.

2. Fetch ktfmt's own native-image metadata from upstream main. It is newer than the 0.64 release and supplies the reachability metadata, the initialization lists, and a substitution class.

git clone --depth 1 --filter=blob:none --sparse https://github.com/facebook/ktfmt
git -C ktfmt sparse-checkout set core/src/main/native-image
NI=ktfmt/core/src/main/native-image

3. Write the entry point. Web Image has no @JS.Export, so this is the documented helper pattern.

cat > KotlinFmt.java <<'EOF'
import com.facebook.ktfmt.format.Formatter;
import java.util.function.BiFunction;
import org.graalvm.webimage.api.JS;
import org.graalvm.webimage.api.JSString;

public final class KotlinFmt {
@JS(args = {"format"}, value = "globalThis.__f = format;")
private static native void exportFormat(BiFunction<JSString, JSString, JSString> format);

exportFormat((source, options) -> JSString.of(Formatter.format(source.asString())));
}
}
EOF

4. Compile the entry point and the substitution class.

javac --add-modules org.graalvm.webimage.api -cp ktfmt.jar -d classes KotlinFmt.java
javac -cp "ktfmt.jar:$JAVA_HOME/lib/svm/builder/svm.jar" -d ni-classes $NI/java/com/facebook/ktfmt/nativeImage/*.java

5. Run the build command above, with this helper defined first for the initialization lists.

rl() { grep -v '^#' "$1" | grep -v '^$' | tr '\n' ',' | sed 's/,$//'; }

Analysis takes about a minute, then code generation fails.

--initialize-at-run-time=com.facebook.ktfmt.format.Parser is required because ktfmt 0.64 builds its KotlinCoreEnvironment eagerly, which --initialize-at-build-time=com.facebook would otherwise place in the image heap. Going the other way, initializing the environment at build time instead, is not an alternative route to this point: it cascades through intellij-core, eighteen classes deep without converging, and starts freezing ResolveCache, PsiManagerImpl and weak-value maps into the image.

### Additional Context

The goal is shipping ktfmt, a real Kotlin formatter, as a wasm module so it runs on Node with no JVM installed. The equivalent Web Image build of google-java-format works today and produces a 12MB WasmGC module, so the pipeline is sound; ktfmt gets all the way through analysis and stops at this switch.

Kotlin bytecode has a history of hitting Graal's structural assumptions, irreducible loops from kotlinc being the familiar case, so it would not be surprising if the monitor structure here comes from the same source.

### Build Log Output and Error Messages

com.oracle.svm.core.util.VMError$HostedError: InternalFeature defined by com.oracle.svm.hosted.webimage.codegen.WebImageImplicitExceptionsFeature unexpectedly failed with a(n) jdk.graal.compiler.debug.GraalError
at org.graalvm.nativeimage.builder/com.oracle.svm.core.util.VMError.shouldNotReachHere(VMError.java:86)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.FeatureHandler.handleFeatureError(FeatureHandler.java:304)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.FeatureHandler.forEachFeature(FeatureHandler.java:95)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.lambda$runPointsToAnalysis$1(NativeImageGenerator.java:844)
at org.graalvm.nativeimage.pointsto/com.oracle.graal.pointsto.AbstractAnalysisEngine.runAnalysis(AbstractAnalysisEngine.java:179)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.runPointsToAnalysis(NativeImageGenerator.java:841)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:569)
at org.graalvm.extraimage.builder/com.oracle.svm.hosted.webimage.WebImageGenerator.doRun(WebImageGenerator.java:116)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:534)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:548)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:761)
at org.graalvm.extraimage.builder/com.oracle.svm.hosted.webimage.NativeImageWasmGeneratorRunner.build(NativeImageWasmGeneratorRunner.java:194)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.start(NativeImageGeneratorRunner.java:163)
at org.graalvm.extraimage.builder/com.oracle.svm.hosted.webimage.NativeImageWasmGeneratorRunner.main(NativeImageWasmGeneratorRunner.java:78)
at org.graalvm.extraimage.builder/com.oracle.svm.hosted.webimage.NativeImageWasmGeneratorRunnerProvider.run(NativeImageWasmGeneratorRunnerProvider.java:42)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:117)
Caused by: jdk.graal.compiler.debug.GraalError: should not reach here: unexpected value: UNSTRUCTURED_LOCKING
at jdk.graal.compiler/jdk.graal.compiler.debug.GraalError.shouldNotReachHereUnexpectedValue(GraalError.java:63)
at org.graalvm.extraimage.builder/com.oracle.svm.hosted.webimage.codegen.WebImageImplicitExceptionsFeature.getSupportMethodName(WebImageImplicitExceptionsFeature.java:111)
at org.graalvm.extraimage.builder/com.oracle.svm.hosted.webimage.codegen.WebImageImplicitExceptionsFeature.resolveSupportMethod(WebImageImplicitExceptionsFeature.java:116)
at org.graalvm.extraimage.builder/com.oracle.svm.hosted.webimage.codegen.WebImageImplicitExceptionsFeature.registerBytecodeException(WebImageImplicitExceptionsFeature.java:174)
at org.graalvm.extraimage.builder/com.oracle.svm.hosted.webimage.codegen.WebImageImplicitExceptionsFeature.duringAnalysis(WebImageImplicitExceptionsFeature.java:163)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.lambda$runPointsToAnalysis$2(NativeImageGenerator.java:844)
at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.FeatureHandler.forEachFeature(FeatureHandler.java:93)

### LINKED ISSUE #14194: [Web Image] --tool:svm-wasm: AtomicReferenceFieldUpdater.newUpdater throws NullPointerException at run time
### Describe the Issue

An AtomicReferenceFieldUpdater created with newUpdater throws a NullPointerException when the image runs. Because the updater is almost always a static final field, this surfaces as an ExceptionInInitializerError from the declaring class, and the class is unusable. The image builds without a warning, so there is nothing at build time to act on.

This is not a reachability-metadata gap. Registering the target field for reflection makes no difference, with or without allowUnsafeAccess, and the minimal reproducer below has no metadata at all and no third-party code.

Web Image appears to have no substitution for the field updaters: grep -r AtomicReferenceFieldUpdater web-image/src/ finds nothing, and there is no objectFieldOffset support in the suite. SubstrateVM handles these through recomputed field offsets (RecomputedFields.java), which does not carry over to WasmGC, where objects are structs with typed fields rather than bytes at an offset. So this looks less like a bug in existing code than a piece that is not there yet. Please read the title as descriptive rather than a claim about the cause.

Impact is broad, because field updaters are common in concurrency-aware libraries. I hit it via IntelliJ's ExtensionPointImpl through kotlin-compiler-embeddable; kotlinx.coroutines, Netty and Reactor all use them too. Any of those makes a Web Image build unusable at run time even when it compiles cleanly.

AtomicIntegerFieldUpdater and AtomicLongFieldUpdater are presumably in the same position, though I only tested the reference one.

### Using the latest version of GraalVM can resolve many issues.

- [x] I tried with the [latest version](https://www.graalvm.org/downloads/) of GraalVM.

### GraalVM Version

java version "25.0.4" 2026-07-21 LTS
Java(TM) SE Runtime Environment Oracle GraalVM 25.2.4+7.1 (build 25.0.4+7-LTS-jvmci-25.2-b20)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 25.2.4+7.1 (build 25.0.4+7-LTS-jvmci-25.2-b20, mixed mode, sharing)

### Operating System and Version

Linux 6.18.5 #1 SMP PREEMPT_DYNAMIC x86_64 GNU/Linux

### Troubleshooting Confirmation

- [x] I tried the suggestions in the [troubleshooting guide](https://www.graalvm.org/latest/reference-manual/native-image/guides/troubleshoot-run-time-errors/).

### Run Command

node --experimental-wasm-exnref afu.js

### Expected Behavior

The updater resolves the field and the program prints:

get=initial
after set=updated
cas=true
final=cased

which is what the same class files print on a JVM.

### Actual Behavior

The class initializer fails and nothing else runs:

Exception in thread "main" java.lang.ExceptionInInitializerError
Caused by: java.lang.RuntimeException: java.lang.NullPointerException

Frames are wasm indices only. Resolved against the .wat name section, the failure is in AtomicReferenceFieldUpdater$AtomicReferenceFieldUpdaterImpl.<init>(Class, Class, String, Class), called from the declaring class's initializer.

### Steps to Reproduce

1. Save this as Afu.java. No dependencies, no reflection metadata, no @JS.

import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;

public class Afu {
static class Holder {
volatile String value = "initial";
}

private static final AtomicReferenceFieldUpdater<Holder, String> UPDATER =
AtomicReferenceFieldUpdater.newUpdater(Holder.class, String.class, "value");

Holder h = new Holder();
System.out.println("get=" + UPDATER.get(h));
UPDATER.set(h, "updated");
System.out.println("after set=" + UPDATER.get(h));
System.out.println("cas=" + UPDATER.compareAndSet(h, "updated", "cased"));
System.out.println("final=" + h.value);
}
}

2. Build it. The build succeeds in about 20 seconds with no diagnostics.

javac -d ac Afu.java
native-image --tool:svm

### Additional Context

_No response_

### Run-Time Log Output and Error Messages

_No response_

amritk:
## Summary

Two independent Web Image build failures, both found compiling ktfmt (the Kotlin
formatter, which embeds the Kotlin compiler frontend and IntelliJ's
intellij-core) with `--tool:svm-wasm`.

**Unhandled `UNSTRUCTURED_LOCKING`.**
`WebImageImplicitExceptionsFeature.getSupportMethodName` maps thirteen
`BytecodeExceptionKind`s to support methods in Web Image's `ImplicitExceptions`
and falls through to `shouldNotReachHere` for `UNSTRUCTURED_LOCKING`, so an image
whose reachable graph produces such a node cannot be built at all. This adds the
missing nullary creator, following the shape of the other kinds that carry a
fixed message (`createNewNegativeLengthException`,
`createNewArgumentIsNotArrayException`), and maps the kind to it.

A nullary creator rather than one taking the message: the regular SVM backend
maps this kind to `CREATE_ILLEGAL_MONITOR_STATE_EXCEPTION`, which takes a
`String`, and `NonSnippetLowerings` appends `getExceptionMessage()` as a constant
argument during lowering. Web Image's feature compares `node.getArguments()`
against the support method's parameter count directly with no such append, so a
one-parameter creator would fail the arity check.

**`java.lang.Object` filter silently disabled by initialization order.**
`ReflectUtil.singleAbstractMethodForInterface` skips abstract methods matching a
public `java.lang.Object` method, per JLS 9.8, so a `@FunctionalInterface` may
legally redeclare `hashCode()`. It built that filter set from
`ResolvedJavaType.getDeclaredMethods(false)`, which returns an empty array for a
type that is not linked yet and is never allowed to force linking
(`AnalysisType.getDeclaredMethods` guarantees `!forceLink`). The result is cached
in a static field for the rest of the build, so an empty result silently disables
the filter for every later query.

The symptom is a failed guarantee ("Incorrect single abstract method logic")
naming an interface that is a legitimate functional interface — IntelliJ's
`RefHashMap.Key` declares `get()` and `hashCode()`. It is order dependent: two
identical builds of the same application named two different interfaces,
`RefHashMap$Key` and `ConcurrentRefHashMap$KeyReference`, because which query
happens to populate the cache first depends on the parallel analysis.

Building the set by reflecting over the host `java.lang.Object` removes the
linkage dependency entirely, and a guarantee makes an empty set loud rather than
silently wrong.

## Related Issues

- Fixes #14193 (unhandled `UNSTRUCTURED_LOCKING`)
- Fixes #14192 (single abstract method failure). Note the cause given in that
  report is wrong — the `Object`-method filter already exists and is correct; it
  is the cached empty set that disables it. Corrected in a comment on the issue.
- Doesnt fix #14194

## Testing

No automated test is included, and I would appreciate direction on where one
belongs.

I did write a unit test against `singleAbstractMethodForInterface` and then
removed it: called with host JVMCI types it returns empty for every interface,
because `GuestAnnotationAccess` cannot read `@FunctionalInterface` off them, so
it asserted nothing. It also would not have caught the original bug, which
depends on analysis ordering rather than on the shape of the interface. A
meaningful regression test looks like a real image build, and I would rather
follow your lead than guess at the harness.

Verified manually:

- `mx build` in the `web-image` suite, against labsjdk
  `ce-25.0.4+7-jvmci-25.2-b20` (the `labsjdk-ce-latest` pin in `common.json`),
  Linux x64.
- Compiled ktfmt 0.64 with `--tool:svm-wasm` through the built `native-image`.
  Before: analysis fails on #14192; with only that fixed, code generation fails
  on #14193. After both: the build completes and produces a 33MB WasmGC module
  (12,320 types and 57,270 methods reachable).
- Ran the resulting module under Node 22 to confirm it instantiates and its
  exported entry point is reachable.
- Re-ran the ktfmt build repeatedly to confirm the order-dependent failure no
  longer appears.

Scope note: the module builds and boots but cannot yet do useful work, because
`AtomicReferenceFieldUpdater.newUpdater` throws at run time under Web Image.
That is unrelated to either fix here, reproduces on a released 25.2.4 with a
15-line example, and is filed separately — nothing in this PR is intended to
address it.

## Documentation

No documentation updates. Both changes are internal to the Web Image builder and
alter no documented behaviour, option, or API.

## Contributor Checklist

- [ ] I have updated tests and documentation where appropriate.

### DIFF
--- a/web-image/src/com.oracle.svm.hosted.webimage.test/src/com/oracle/svm/hosted/webimage/test/WebImageUnitTests.java
+++ b/web-image/src/com.oracle.svm.hosted.webimage.test/src/com/oracle/svm/hosted/webimage/test/WebImageUnitTests.java
@@ -29,6 +29,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Suite;
 
+import com.oracle.svm.hosted.webimage.test.unit.ImplicitExceptionsSupportMethodTest;
 import com.oracle.svm.hosted.webimage.test.unit.MoveResolverTest;
 import com.oracle.svm.hosted.webimage.test.unit.WasmActiveDataTest;
 
@@ -36,6 +37,7 @@ import junit.framework.TestSuite;
 
 @RunWith(Suite.class)
 @Suite.SuiteClasses({
+                ImplicitExceptionsSupportMethodTest.class,
                 MoveResolverTest.class,
                 WasmActiveDataTest.class
 })
--- a/web-image/src/com.oracle.svm.hosted.webimage.test/src/com/oracle/svm/hosted/webimage/test/unit/ImplicitExceptionsSupportMethodTest.java
+++ b/web-image/src/com.oracle.svm.hosted.webimage.test/src/com/oracle/svm/hosted/webimage/test/unit/ImplicitExceptionsSupportMethodTest.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2026, 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.oracle.svm.hosted.webimage.test.unit;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import com.oracle.svm.hosted.webimage.codegen.WebImageImplicitExceptionsFeature;
+import com.oracle.svm.webimage.functionintrinsics.ImplicitExceptions;
+
+import jdk.graal.compiler.nodes.extended.BytecodeExceptionNode.BytecodeExceptionKind;
+
+/**
+ * Checks that every {@link BytecodeExceptionKind} has a usable support method in
+ * {@link ImplicitExceptions}.
+ * <p>
+ * {@link WebImageImplicitExceptionsFeature#getSupportMethodName} throws for unmapped kinds, and
+ * mismatched arities are only detected once a
+ * {@link jdk.graal.compiler.nodes.extended.BytecodeExceptionNode} of that kind actually shows up in
+ * the analyzed graphs. Both failure modes therefore only surface for applications that happen to
+ * contain the offending node, which is how the unmapped
+ * {@link BytecodeExceptionKind#UNSTRUCTURED_LOCKING} went unnoticed. Going over the enum here
+ * catches a new or changed kind at test time instead.
+ * <p>
+ * The methods are only inspected reflectively, never invoked, so that
+ * {@link ImplicitExceptions}'s class initializer (which registers foreign calls) does not have to
+ * run outside of an image build.
+ */
+@RunWith(Parameterized.class)
+public class ImplicitExceptionsSupportMethodTest {
+
+    @Parameterized.Parameters(name = "{0}")
+    public static BytecodeExceptionKind[] data() {
+        return BytecodeExceptionKind.values();
+    }
+
+    @Parameterized.Parameter public BytecodeExceptionKind kind;
+
+    @Test
+    public void supportMethodIsUsable() {
+        String methodName = WebImageImplicitExceptionsFeature.getSupportMethodName(kind);
+        Assert.assertNotNull("No support method name for " + kind, methodName);
+
+        List<Method> candidates = new ArrayList<>();
+        for (Method method : ImplicitExceptions.class.getDeclaredMethods()) {
+            if (method.getName().equals(methodName)) {
+                candidates.add(method);
+            }
+        }
+
+        Assert.assertEquals(ImplicitExceptions.class.getName() + "." + methodName + " for " + kind + " must exist exactly once", 1, candidates.size());
+
+        Method supportMethod = candidates.get(0);
+        Assert.assertTrue(supportMethod + " must be public", Modifier.isPublic(supportMethod.getModifiers()));
+        Assert.assertTrue(supportMethod + " must be static", Modifier.isStatic(supportMethod.getModifiers()));
+        Assert.assertTrue(supportMethod + " must return a Throwable for " + kind, Throwable.class.isAssignableFrom(supportMethod.getReturnType()));
+        Assert.assertEquals(supportMethod + " must take the arguments of " + kind, kind.getNumArguments(), supportMethod.getParameterCount());
+    }
+}
--- a/web-image/src/com.oracle.svm.hosted.webimage/src/com/oracle/svm/hosted/webimage/codegen/WebImageImplicitExceptionsFeature.java
+++ b/web-image/src/com.oracle.svm.hosted.webimage/src/com/oracle/svm/hosted/webimage/codegen/WebImageImplicitExceptionsFeature.java
@@ -91,6 +91,7 @@ public final class WebImageImplicitExceptionsFeature implements InternalFeature 
             case ILLEGAL_ARGUMENT_EXCEPTION_ARGUMENT_IS_NOT_AN_ARRAY -> "createNewArgumentIsNotArrayException";
             case ASSERTION_ERROR_OBJECT -> "createNewAssertionErrorObject";
             case ASSERTION_ERROR_NULLARY -> "createNewAssertionErrorNullary";
+            case UNSTRUCTURED_LOCKING -> "createNewIllegalMonitorStateException";
 
             default -> throw GraalError.shouldNotReachHereUnexpectedValue(exceptionKind);
         };
--- a/web-image/src/com.oracle.svm.hosted.webimage/src/com/oracle/svm/hosted/webimage/util/ReflectUtil.java
+++ b/web-image/src/com.oracle.svm.hosted.webimage/src/com/oracle/svm/hosted/webimage/util/ReflectUtil.java
@@ -25,6 +25,7 @@
 
 package com.oracle.svm.hosted.webimage.util;
 
+import java.lang.reflect.Method;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Optional;
@@ -57,11 +58,18 @@ public class ReflectUtil {
     private static synchronized UnmodifiableEconomicSet<String> getObjectMethodDescriptors(MetaAccessProvider metaAccess) {
         if (objectMethodDescriptors == null) {
             EconomicSet<String> descriptors = EconomicSet.create();
-            for (ResolvedJavaMethod declaredMethod : metaAccess.lookupJavaType(Object.class).getDeclaredMethods(false)) {
-                if (declaredMethod.isPublic()) {
-                    descriptors.add(fullDescriptor(declaredMethod));
-                }
+            /*
+             * Reflection on the host java.lang.Object rather than
+             * ResolvedJavaType.getDeclaredMethods, because that returns an empty array for a type
+             * that is not linked yet and is never allowed to force linking. This set is computed
+             * once and cached for the rest of the build, so picking up an empty result would
+             * silently disable the java.lang.Object filter below for every later query, and which
+             * query lands here first depends on the order the parallel analysis happens to run in.
+             */
+            for (Method declaredMethod : Object.class.getMethods()) {
+                descriptors.add(fullDescriptor(metaAccess.lookupJavaMethod(declaredMethod)));
             }
+            GraalError.guarantee(!descriptors.isEmpty(), "Found no public methods on java.lang.Object");
             objectMethodDescriptors = descriptors;
         }
 
--- a/web-image/src/com.oracle.svm.webimage/src/com/oracle/svm/webimage/functionintrinsics/ImplicitExceptions.java
+++ b/web-image/src/com.oracle.svm.webimage/src/com/oracle/svm/webimage/functionintrinsics/ImplicitExceptions.java
@@ -104,6 +104,11 @@ public class ImplicitExceptions {
     }
 
     @NeverInline("outlining for smaller code size")
+    public static IllegalMonitorStateException createNewIllegalMonitorStateException() {
+        return new IllegalMonitorStateException("Unstructured locking encountered. Native Image enforces structured locking (JVMS 2.11.10)");
+    }
+
+    @NeverInline("outlining for smaller code size")
     public static ArithmeticException createNewArithmeticException() {
         return new ArithmeticException();
     }


xhon-pelushi (commented):
Verified both halves against `master` at source level. They look correct to me, and the `UNSTRUCTURED_LOCKING` mapping matches the canonical definition exactly. I also found **two more kinds with the same problem** that this PR doesn't cover, plus a cheap way to stop this class of failure recurring.

## `UNSTRUCTURED_LOCKING`: arity, type and message all match the canonical definition

`BytecodeExceptionNode.java:159` declares:

```java
UNSTRUCTURED_LOCKING(0, IllegalMonitorStateException.class,
    "Unstructured locking encountered. Native Image enforces structured locking (JVMS 2.11.10)");
```

with the doc comment *"No additional arguments are allowed."* So all three of the PR's choices are pinned by the enum rather than guessed:

- **arity 0** — a nullary creator is the only correct shape.
- **`IllegalMonitorStateException`** — matches, and matches the main backend, which uses the same type for this case (`ImplicitExceptions.CACHED_ILLEGAL_MONITOR_STATE_EXCEPTION` at line 70, plus `CREATE_ILLEGAL_MONITOR_STATE_EXCEPTION`).
- **the message** — character-for-character identical to the enum's `exceptionMessage`.

Hardcoding the literal rather than reading `BytecodeExceptionKind.UNSTRUCTURED_LOCKING.exceptionMessage` follows the file's existing convention — `"Argument is not an array"` is already duplicated there from `ILLEGAL_ARGUMENT_EXCEPTION_ARGUMENT_IS_NOT_AN_ARRAY` — so no objection from me, just noting the field exists if de-duplicating ever seems worthwhile.

## `ReflectUtil`: the descriptor-format trap is avoided

My first concern was that swapping `ResolvedJavaMethod` for a host `java.lang.reflect.Method` would change the descriptor strings and silently break the `java.lang.Object` filter below — turning a loud empty-set bug into a quiet mismatch bug. It doesn't, because the PR still funnels through the same overload:

```java
descriptors.add(fullDescriptor(metaAccess.lookupJavaMethod(declaredMethod)));
```

Only the *source* of the method list changes, not the formatting. Worth stating explicitly since it's the non-obvious reason this is safe.

The set is also equivalent. `Object.class.getMethods()` versus the old `getDeclaredMethods(false)` filtered by `isPublic()`, on JDK 21:

```
Object.class.getMethods()                = 9  [equals, getClass, hashCode, notify, notifyAll, toString, wait(), wait(long), wait(long,int)]
getDeclaredMethods() filtered to public  = 9  [same]
SETS EQUAL: true
```

`clone` and `finalize` are `protected` and excluded by both; `wait0` is `private` and excluded by both. Dropping the explicit `isPublic()` check is right because `getMethods()` is public-only, and since `Object` has no superclass there is nothing inherited for `getMethods()` to add.

The `GraalError.guarantee(!descriptors.isEmpty(), ...)` is the right shape for this bug — it converts a silent, order-dependent misbehaviour into a deterministic build failure.

One incidental improvement worth noting: `objectMethodDescriptors` is a `static` field with no reset hook, and the old value was derived from a build-scoped `metaAccess.lookupJavaType(...)` result. The new value comes from host reflection, so it's genuinely build-independent and safe to keep in a static across multiple images in one JVM. That's a second latent problem this change removes.

## Two more kinds are still unmapped

Comparing the enum against the switch on current `master`, three kinds fall through to `shouldNotReachHereUnexpectedValue`:

```
BytecodeExceptionKind values (15):  ARRAY_STORE ASSERTION_ERROR_NULLARY ASSERTION_ERROR_OBJECT CLASS_CAST
  DIVISION_BY_ZERO ILLEGAL_ARGUMENT_EXCEPTION_ARGUMENT_IS_NOT_AN_ARRAY
  ILLEGAL_ARGUMENT_EXCEPTION_NEGATIVE_LENGTH INCOMPATIBLE_CLASS_CHANGE INTEGER_EXACT_OVERFLOW
  INTRINSIC_OUT_OF_BOUNDS LONG_EXACT_OVERFLOW NEGATIVE_ARRAY_SIZE NULL_POINTER OUT_OF_BOUNDS
  UNSTRUCTURED_LOCKING

handled by getSupportMethodName (12): ... (as above, minus the three below)

unhandled: INTEGER_EXACT_OVERFLOW  LONG_EXACT_OVERFLOW  UNSTRUCTURED_LOCKING
```

This PR fixes the third. The other two are the same failure mode — an image whose reachable graph contains an `Math.addExact`-style overflow check simply cannot be built with `--tool:svm-wasm`. Both are cheap:

```java
INTEGER_EXACT_OVERFLOW(0, ArithmeticException.class),
LONG_EXACT_OVERFLOW(0, ArithmeticException.class),
```

Arity 0, `ArithmeticException`, and no declared message — and Web Image's `ImplicitExceptions` already has a nullary `createNewArithmeticException()`, so mapping both to it needs no new methods. The main backend distinguishes them (`createIntegerOverflowException` → `"integer overflow"`, `createLongOverflowException` → `"long overflow"`), so separate creators would give better messages if you'd rather match that. Either way it's a small addition, and up to you whether it belongs here or in a follow-up.

## Suggestion: make the next missing kind a test failure, not a build failure

The reason `UNSTRUCTURED_LOCKING` shipped unhandled is that nothing checks the mapping is total — the gap only surfaces when someone compiles a program that happens to produce that node. An exhaustive test over the enum would catch all of these at once and keep catching them as kinds are added:

```java
@Test
public void everyBytecodeExceptionKindHasASupportMethod() {
    for (BytecodeExceptionNode.BytecodeExceptionKind kind : BytecodeExceptionNode.BytecodeExceptionKind.values()) {
        Assert.assertNotNull(kind.toString(), WebImageImplicitExceptionsFeature.getSupportMethodName(kind));
    }
}
```

That would have flagged all three unmapped kinds today, and turns a "cannot build this program at all" into a failing unit test. Neither half of this PR currently has a regression test, and this one is cheap enough to be worth adding.

## What I could not test

I did not build GraalVM or Web Image — no `mx` toolchain here, and the build is well beyond this machine's budget — so I could not reproduce either ktfmt `--tool:svm-wasm` failure or run the Web Image suite. Everything above is source verification against `master` plus a JDK 21 check of the two `java.lang.Object` method sets. In particular I have not confirmed that `metaAccess.lookupJavaMethod` on the nine host `Object` methods succeeds at every point in the build where `getObjectMethodDescriptors` can first be called; that's the one step in the new path I'd want a real build to exercise.

xhon-pelushi:
**Correction to my earlier review, first — I got one part wrong.**

I claimed `INTEGER_EXACT_OVERFLOW` and `LONG_EXACT_OVERFLOW` were also unmapped and that this PR was fixing "the third" of three gaps. That is not right. `master` already maps both, in a combined arm I missed because it does not follow the one-constant-per-case shape of the lines around it:

```java
case INTEGER_EXACT_OVERFLOW, LONG_EXACT_OVERFLOW -> "createNewArithmeticException";
```

So there was exactly **one** unmapped kind, `UNSTRUCTURED_LOCKING`, and this PR covers it. Apologies if that sent anyone looking for work that did not need doing.

**The new test looks right, and stronger than what I sketched.**

`ImplicitExceptionsSupportMethodTest` checks more than the non-null name I proposed — that the method exists exactly once, is public and static, returns a `Throwable`, and that its parameter count equals `kind.getNumArguments()`. Doing it reflectively so `ImplicitExceptions`'s class initializer never runs outside an image build is a nice touch I would not have thought of.

Since it iterates the whole enum and `getSupportMethodName` throws on an unmapped kind, I checked all 15 constants against `ImplicitExceptions` at `0ddb8ccd` to confirm it will actually be green:

| Kind | Arity | Support method | Params |
| ---- | ----- | -------------- | ------ |
| NULL_POINTER | 0 | `createNewNullPointerException` | 0 |
| OUT_OF_BOUNDS | 2 | `createNewOutOfBoundsExceptionWithArgs(int, int)` | 2 |
| INTRINSIC_OUT_OF_BOUNDS | 0 | `createNewOutOfBoundsException` | 0 |
| CLASS_CAST | 2 | `createNewClassCastExceptionWithArgs(Object, Class<?>)` | 2 |
| ARRAY_STORE | 1 | `createNewArrayStoreExceptionWithArgs(Object)` | 1 |
| INCOMPATIBLE_CLASS_CHANGE | 0 | `createNewIncompatibleClassChangeError` | 0 |
| ASSERTION_ERROR_NULLARY | 0 | `createNewAssertionErrorNullary` | 0 |
| ASSERTION_ERROR_OBJECT | 1 | `createNewAssertionErrorObject(Object)` | 1 |
| ILLEGAL_ARGUMENT_EXCEPTION_NEGATIVE_LENGTH | 0 | `createNewNegativeLengthException` | 0 |
| ILLEGAL_ARGUMENT_EXCEPTION_ARGUMENT_IS_NOT_AN_ARRAY | 0 | `createNewArgumentIsNotArrayException` | 0 |
| NEGATIVE_ARRAY_SIZE | 1 | `createNegativeArraySizeException(int)` | 1 |
| DIVISION_BY_ZERO | 0 | `createNewDivisionByZeroException` | 0 |
| INTEGER_EXACT_OVERFLOW | 0 | `createNewArithmeticException` | 0 |
| LONG_EXACT_OVERFLOW | 0 | `createNewArithmeticException` | 0 |
| UNSTRUCTURED_LOCKING | 0 | `createNewIllegalMonitorStateException` | 0 |

All 15 are mapped, each name resolves to exactly one `public static` method returning a `Throwable` subclass, and every arity matches. Nothing in the enum is left to trip the `default -> throw` arm.

Same caveat as before: this is source verification only — I still have no `mx` toolchain here, so I have not run the suite or built Web Image.

Issue #13975, opened 2026-07-14
[GR-77783][Native Image][Tracing Agent] Reflective Field.get/set is over 100x slower on GraalVM 25


rotilho:
### Describe the Issue

The Native Image tracing agent in Oracle GraalVM 25 introduces severe runtime overhead for reflective field access when the default `track-reflection-metadata=true` behavior is enabled. Reflection-heavy applications and integration-test suites that complete normally on GraalVM 21 and 24 can take several times longer on GraalVM 25, eventually exceeding test-level or CI time limits.

This is particularly visible in Spring and Kotlin workloads, where serializers, dependency injection, binders, proxies, and test infrastructure repeatedly access fields through reflection. In an existing full application test matrix, the same agent-instrumented suite took:

- Oracle GraalVM 21: 9m12s
- Oracle GraalVM 24: 8m53s
- Oracle GraalVM 25: 35m56s

The underlying per-operation regression is much larger than the full-suite ratio. With GraalVM 25.0.3, repeated `Field.get` calls take approximately 22 ms without the agent and 2,540 ms with the default agent: a 113x slowdown. The same operation on GraalVM 24 has almost no additional agent overhead.

The regression appears to align with #11839 / commit `dc5c3a8a32fc0349d61803088de6626f7917da34`, backported to 25.0 in #11894. That change moved reflection field registration from query time to access time and added JVMTI breakpoints to the `Field.get*/set*` methods. Consequently, every reflective field read or write appears to enter the breakpoint/JNI/trace path, including repeated accesses to a field that has already been observed.

There is currently no complete configuration workaround. Disabling `track-reflection-metadata` restores performance, but GraalVM 25 then omits the accessed field from reachability metadata. A native image built with that output succeeds at build time but fails at runtime with `MissingReflectionRegistrationError` for the omitted field.

A minimal dependency-free probe is provided below only to isolate and measure the problem. Three-run medians for 100,000 reflective field reads show that the regression starts with GraalVM 25 and remains present in both the latest tested 25.0 maintenance release and the tested 25.1 Innovation release:

| Oracle GraalVM | No agent | Default agent | Agent / no-agent | Agent with `track-reflection-metadata=false` |
|---|---:|---:|---:|---:|
| 21.0.11 | 39.109 ms | 46.986 ms | 1.20x | 40.954 ms |
| 24.0.2 | 36.983 ms | 37.682 ms | 1.02x | 40.390 ms |
[...]

PR #14204, opened 2026-08-08
Improve native-image agent ZIP resource tracing

### LINKED ISSUE #10352: Significant execution time degradation with agent used during tests
**Describe the issue**
When running Spring tests with the GraalVM agent, there is significant execution time degradation compared to running without the agent. With the agent enabled, test execution takes over 1 hour versus 4 minutes without it for a project with ~120 MockMvc integration tests.

Originally I described issue in gradle native plugin https://github.com/graalvm/native-build-tools/issues/655 but was suggested to move it here.

1. The issue can be partially demonstrated with spring-petclinic, though with smaller impact:
   - Without agent: ~1m 28s
   - With standard agent: ~2m 16s  
   - With conditional agent: ~3m 43s

However, my case shows much more severe degradation. I should note that a few beans were mocked using @MockBean (switching to @MockitoBean didn't change much), which might lead to context refresh. Even considering this behavior, I'm wondering why it would slow execution so dramatically. Additionally, pet-clinic also uses @MockitoBean but doesn't show comparable degradation.

**Environment:**
OS: Windows
GraalVM Version: GraalVM 22.3.1 Java 19 C
Java Version: 19.0.2+7-jvmci-22.3-b12

OS: WSL
GraalVM Version: CE 23.0.1+11.1
Java Version: 23.0.1

**More details:**

I attached Flight Recorder to worker at it shows long pauses. 
image

rotilho:
Place the `ZipFile.getEntry` breakpoint at its private `getZipEntry` invocation so it runs only after a successful lookup. This avoids JVMTI events for failed lookups and for stream or enumeration paths while preserving direct resource metadata and caller origins.

Add real-agent coverage for successful, missing, and streamed ZIP entries.

## Summary

The Native Image tracing agent currently places a breakpoint at the entry of `ZipFile.getEntry(String)`. Consequently, every lookup crosses a JVMTI breakpoint, including unsuccessful probes that do not produce resource metadata. Framework workloads can perform many thousands of such probes.

This change dynamically locates the `getZipEntry(String, int)` invocation inside public `ZipFile.getEntry(String)` and installs the breakpoint at that bytecode location. The JDK reaches this invocation only after finding the requested entry.

This means:

- successful direct `ZipFile.getEntry(String)` calls retain the existing resource trace operation, metadata, caller attribution, and origin stack;
- failed lookups do not cross the breakpoint;
- `ZipFile.stream()`, `JarFile.stream()`, and enumeration paths do not cross the breakpoint or produce incidental resource metadata;
- the bytecode location is resolved through JVMTI bytecode and constant-pool data rather than hardcoding a JDK-specific BCI.

## Related Issues

- Fixes https://github.com/oracle/graal/issues/10352
- Related performance investigation: https://github.com/oracle/graal/issues/13975
- This change addresses the distinct ZIP lookup overhead identified while investigating that report. It does not change ordinary reflective `Field.get` or `Field.set` tracing.

## Testing

The following checks passed:

- Built `SVM_AGENT` and `com.oracle.svm.configure.test`.
- Built `SVM_TESTS`.
- `mx gate --tags condconfig`
- `mx checkstyle`
- `git diff --check`
- Real-agent tests covering:
  - a successful direct classpath ZIP lookup is recorded;
  - a missing ZIP entry is not recorded;
  - streamed ZIP entries are not recorded.

Performance validation with 25,000 operations:

- Failed lookups, default configuration: 805.86 ms → 4.75 ms.
- Failed lookups, configuration with origins: 856.88 ms → 5.57 ms.
- Streaming, default configuration: 2.84 ms → 2.75 ms.
- Streaming, configuration with origins: 2.85 ms → 2.85 ms.

An end-to-end Spring Petclinic tracing-agent comparison improved from 324.82 seconds to 122.47 seconds, a 62.3% reduction. Generated resource metadata remained exactly equal at 840 entries in both runs.

A focused origins test confirmed that successful direct lookup metadata remains attributed to the application caller through public `ZipFile.getEntry`, without exposing the private helper frame.

## Documentation

No documentation updates are needed. This changes an internal tracing-agent interception point while preserving the existing resource tracing protocol and user-visible configuration.

## Contributor Checklist


rotilho:
Hey @vjovanov! Sorry for the ping, but with the next CPU release approaching, I was hoping this fix could make it into the release.

This change reduced agent time by 62% in https://github.com/rotilho/spring-petclinic, so the impact is quite significant.

PR #14239, opened 2026-08-12
Generatioanl Shenandoah fix: Skip card mark for off-heap stores

JohnTortugo:
Generational Shenandoah crashed with SIGSEGV/SIGBUS in Graal-compiled code when storing an oop into the contents of an OopHandle, such as the store performed by the Thread.setScopedValueCache intrinsic.

`HotSpotShenandoahBarrierSet.writeBarrierType()` answers `BarrierType.FIELD` for an `OopHandleLocationIdentity` so that the SATB pre barrier is emitted, but `ShenandoahBarrierSet.addWriteBarriers()` then also attached a **card barrier unconditionally**. An OopHandle's contents live in a native OopStorage rather than in the Java heap, so computing `card_table_base + (address >> card_shift)` lands outside the card table and the card write corrupts unrelated memory or faults.

Gate the card barrier on a new overridable `isInHeap()` predicate, mirroring HotSpot, which gates on the `IN_HEAP` decorator via `ShenandoahBarrierSet::need_card_barrier` and splits `oop_store_in_heap` from `oop_store_not_in_heap`. The SATB pre barrier is retained. G1 was unaffected because it does not override `writeBarrierType(LocationIdentity)` and so emits no barrier at all here; ZGC was unaffected because it carries the distinction down to its LIR generators as `StoreKind.Native`.

Also pass 32 rather than 8 as the `cbz` operand size in the AArch64 card barrier op. `cbz` has no 8-bit form and asserted on every conditional card mark. With assertions disabled the encoder falls back to a 32-bit `cbz`, which is accidentally correct because the preceding `ldrb` zero extends, so this is benign in a product build but makes the path impossible to exercise in an assertion-enabled build.  It is a prerequisite for the in-heap case of the test added here.

Added `ShenandoahOopHandleBarrierTest`, which asserts that an `OopHandle` oop store keeps its SATB pre barrier and is not card marked, and that an ordinary in-heap oop store still is.

Testing (aarch64, jargraal): a ScopedValue stress reproducer driving first-time ScopedValue.get() calls on freshly created threads crashes within a second unpatched; patched it survives 271M virtual-thread and 1.8M platform-thread iterations, including under `-XX:+ShenandoahVerify`, which also validates remembered set integrity. The new test fails without the `isInHeap()` gate ("OopHandle oop store must not be card marked") and passes with it.

## Contributor Checklist

### DIFF
--- a/compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/hotspot/test/ShenandoahOopHandleBarrierTest.java
+++ b/compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/hotspot/test/ShenandoahOopHandleBarrierTest.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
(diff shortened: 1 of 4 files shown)


simonis:
Congratulations, great finding @JohnTortugo (especially without a real reproducer!). The change looks good to me, but I'm not a compiler expert, so I defer the real review to one of the compiler experts (cc @dougxc ).

JohnTortugo:
ping @tkrodriguez ; Just bringing this on your radar again.

tkrodriguez:
Sorry I took so long to put this up.  https://github.com/oracle/graal/pull/14300

JohnTortugo:
Thank you @tkrodriguez

Issue #9763, opened 2024-09-26
[GR-44523] Implement Tail Call Proposal for WebAssembly


fniephaus:
### TL;DR

We plan to implement [the Tail Call proposal](https://github.com/WebAssembly/tail-call) in GraalWasm.

### Details

This proposal adds new call instructions for tail calls. Tail calls behave like regular function calls, except that they pop the current stack frame before executing the call. Support for such calls is necessary for a correct and efficient implementation of languages that rely on tail call elimination. It also facilitates implementing languages with control constructs like coroutines or continuations efficiently.
Implementing this proposal will allow us to run such languages efficiently on GraalWasm.

More details at https://github.com/WebAssembly/tail-call/blob/main/proposals/tail-call/Overview.md

PR #14245, opened 2026-08-13
[GR-44523] Implementation of the WebAssembly Tail Calls Proposal

flohuemer:
## Summary

This PR implements the WebAssembly Tail Calls Proposal in GraalWasm. Because the JVM lacks native tail-call support, the implementation uses two strategies.

1. Direct self-recursion

For functions that tail-call themselves, the parser rewrites the tail call into a branch targeting the function's entry point, effectively forming a loop. When the interpreter executes this branch, it:
- Moves the call arguments to the first locals,
- Resets all other locals,
- Unwinds the stack,
- Jumps back to the start of the function.

This achieves tail-call behavior without the need for calls.

2. All other tail calls (including indirect and ref calls)

In a chain of consecutive tail calls, the first call is executed as a conventional call. Each subsequent tail call returns a special tail-call value that carries the next call target and its arguments, rather than invoking the call itself. This causes the stack to unwind back to the first caller, which then invokes the returned call target with the provided arguments. As a result, all later calls in the chain reuse the same stack space.

To avoid allocations, the call target and arguments are stored in a thread-local value. To distinguish the first call in a tail-call chain from subsequent ones, the second argument of each call is used as a tail-call marker. Conventional calls ignore this marker.

Additionally, this PR:

- Moves the i32 variants of local-variable access bytecodes into the `misc` section of the custom bytecode. Because functions with more than 256 locals are expected to be rare, this should not affect interpreter performance.
- Moves extraction of the final return value from the root node into the interpreter's return statement.

## Related Issues

- [Implement Tail Call Proposal for WebAssembly](https://github.com/oracle/graal/issues/9763)

## Testing

- Added 9 tail-call-related tests covering all three call forms (direct, indirect, ref)
- Ran the gate locally to ensure existing tests still pass. 
- Note: I was not able to run the official WebAssembly test suite (including the test for tail calls) on this PR.

## Documentation

- No user-facing documentation changes. Code comments and JavaDoc were updated where appropriate.

## Contributor Checklist


wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmReturnCallNode.java
```diff
@@ -0,0 +1,108 @@
...
+import com.oracle.truffle.api.CallTarget;
+import com.oracle.truffle.api.dsl.Cached;
+import com.oracle.truffle.api.dsl.GenerateCached;
+import com.oracle.truffle.api.dsl.GenerateInline;
+import com.oracle.truffle.api.dsl.NeverDefault;
+import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.api.nodes.DenyReplace;
+import com.oracle.truffle.api.nodes.UnadoptableNode;
+
+@GenerateInline(false)
```
flohuemer, line 56: I am not entirely sure if this is the correct annotation for this class, as it has an instance field.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -1818,20 +1790,88 @@ private static int callIndirectI32Handler(int offset, State state, VirtualState
...
+        final WasmFunctionInstance functionInstance = state.thiz.loadRefCallFunctionInstance(frame, --virtualState.stackPointer);
+        // Invoke the resolved function.
+        int paramCount = state.module.symbolTable().functionTypeParamCount(expectedFunctionTypeIndex);
+        Object[] args = state.thiz.createArgumentsForCall(frame, expectedFunctionTypeIndex, paramCount, virtualState.stackPointer);
+        return state.thiz.executeIndirectReturnCall(frame, functionInstance, callNodeIndex, args);
+    }
+
+    @EarlyInline
+    private static Object returnCallRefI32(int offset, State state, VirtualState virtualState, FrameWithoutBoxing frame) {
+        final int callNodeIndex = rawPeekU8(state.bytecode, offset + 1);
```
jirkamarsik: Should be `rawPeekI32` on this line and the next.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -1818,20 +1790,88 @@ private static int callIndirectI32Handler(int offset, State state, VirtualState
         return offset + 13;
     }
 
+    @EarlyInline
+    private static Object returnCallU8(int offset, State state, VirtualState virtualState, FrameWithoutBoxing frame) {
```
jirkamarsik, line 1799: Minor: Looking at the surrounding code, the style for bytecode handler method names is `{bytecodeName}Handler`, so we should have `returnCallU8Handler`, `returnCallI32Handler`...

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4792,6 +4833,71 @@ private Object executeIndirectCallNode(int callNodeIndex, CallTarget target, Obj
...
+        return false;
+    }
+
+    private Object executeIndirectReturnCall(VirtualFrame frame, WasmFunctionInstance functionInstance, int callNodeIndex, Object[] args) {
+        WasmArguments.setModuleInstance(args, functionInstance.moduleInstance());
+        return executeReturnCall(frame, functionInstance.target(), callNodeIndex, args, WasmLanguage.get(this));
+    }
+
+    private Object executeReturnCall(VirtualFrame frame, CallTarget target, int callNodeIndex, Object[] args, WasmLanguage language) {
+        WasmArguments.setReturnCallMarker(args);
```
jirkamarsik: In the first branch of the `if`, this is not needed. Let's move this into the second branch, where we start the tail-call chain.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4792,6 +4833,71 @@ private Object executeIndirectCallNode(int callNodeIndex, CallTarget target, Obj
...
+            WasmReturnCallNode callNode = (WasmReturnCallNode) callNodes[callNodeIndex];
+            WasmArguments.setModuleInstance(args, instance);
+            target = callNode.directTarget();
+        }
+        return executeReturnCall(frame, target, callNodeIndex, args, WasmLanguage.get(this));
+    }
+
+    private boolean assertDirectReturnCall(WasmInstance instance, WasmFunction function, int callNodeIndex) {
+        WasmFunctionInstance functionInstance = instance.functionInstance(function.index());
+        // functionInstance may be null for calls between functions of the same module.
```
jirkamarsik: Can `functionInstance` be null? I'm looking at the `functionInstance` getter in `WasmInstance` and whenever it would return `null`, it always allocates a new `WasmFunctionInstance` .

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmReturnCallNode.java
jirkamarsik: We are not inlining `WasmCallNode`s, so disabling the generation of the inline variant here looks like a good call!

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
jirkamarsik: I see it is the same in `assertDirectCall`. In that case, we can leave it as is and I'll do a clean-up pass later where I update these `assert` methods.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/WasmContextOptions.java
```diff
@@ -118,6 +122,9 @@ private void checkOptionDependencies() {
         if (gc && !typedFunctionReferences) {
             failDependencyCheck("GC", "TypedFunctionReferences");
         }
+        if (tailCallLoops && !tailCalls) {
+            failDependencyCheck("TailCallLoops", "TailCalls");
```
jirkamarsik: We should either remove this dependency check or, better yet, default `TailCallLoops` to `false` when `TailCalls` are disabled. Otherwise, you will get an error if you try `--wasm.TailCalls=false`.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4636,6 +4612,71 @@ private static int miscHandler(int offset, State state, VirtualState virtualStat
...
+                nextOffset = offset + 5;
+                break;
+            }
+            case Bytecode.BR_RETURN_CALL: {
+                int paramCount = state.module.symbolTable().functionTypeParamCount(state.thiz.codeEntry.functionIndex());
+                unwindStack(frame, virtualState.stackPointer, 0, paramCount);
+                dropStack(frame, virtualState.stackPointer, virtualState.stackPointer - paramCount);
+                nextOffset = state.thiz.bytecodeStartOffset;
+                virtualState.stackPointer = state.thiz.codeEntry.stackBase();
+                WasmFrame.initializeLocals(frame, paramCount, state.thiz.codeEntry);
```
jirkamarsik, line 4661: When resetting the interpreter state, we should also reset `state.activeLegacyCatchCount`.

jirkamarsik: NB: We should also clear the reserved stack slots used to hold the active legacy catch stack.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4636,6 +4612,71 @@ private static int miscHandler(int offset, State state, VirtualState virtualStat
...
+            case Bytecode.LOCAL_TEE_OBJ_I32: {
+                final int index = rawPeekI32(state.bytecode, offset + 1);
+                local_tee_obj(frame, virtualState.stackPointer - 1, index);
+                nextOffset = offset + 5;
+                break;
+            }
+            case Bytecode.BR_RETURN_CALL: {
+                int paramCount = state.module.symbolTable().functionTypeParamCount(state.thiz.codeEntry.functionIndex());
+                unwindStack(frame, virtualState.stackPointer, 0, paramCount);
+                dropStack(frame, virtualState.stackPointer, virtualState.stackPointer - paramCount);
```
jirkamarsik: I think this should be `dropStack(frame, virtualState.stackPointer, paramCount)` so that we drop the whole stack and keep only the parameters.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4636,6 +4612,71 @@ private static int miscHandler(int offset, State state, VirtualState virtualStat
...
+                WasmFrame.initializeLocals(frame, paramCount, state.thiz.codeEntry);
+                /*
+                 * As this return call forms a loop without explicit header,
+                 * we report the loop count here.
+                 */
+                if (CompilerDirectives.hasNextTier()) {
+                    int count = CompilerDirectives.inCompiledCode() ? ++state.backEdgeCounter.count : ++state.interpreterBackEdgeCounter;
+                    if (CompilerDirectives.injectBranchProbability(0.001, count >= REPORT_LOOP_STRIDE)) {
+                        TruffleSafepoint.poll(state.thiz);
+                        LoopNode.reportLoopCount(state.thiz, REPORT_LOOP_STRIDE);
```
jirkamarsik: Can we add OSR support here the same way we have in the `LOOP` instruction?

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4792,6 +4833,71 @@ private Object executeIndirectCallNode(int callNodeIndex, CallTarget target, Obj
...
+        } else {
+            WasmReturnCallNode callNode = (WasmReturnCallNode) callNodes[callNodeIndex];
+            Object result = callNode.execute(target, args);
+            if (result == WasmConstant.RETURN_CALL_VALUE) {
+                // We only load and clean up the thread local when we start a tail-call chain.
+                final WasmLanguage.ReturnCallData data = language.returnCallData();
+                do {
+                    result = callNode.execute(data.target, data.args);
+                } while (result == WasmConstant.RETURN_CALL_VALUE);
+                data.target = null;
```
jirkamarsik: It might be better to clear `data.target` and `data.args` in a `finally` block to be robust w.r.t. exceptions.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4636,6 +4612,71 @@ private static int miscHandler(int offset, State state, VirtualState virtualStat
...
+                break;
+            }
+            case Bytecode.LOCAL_TEE_OBJ_I32: {
+                final int index = rawPeekI32(state.bytecode, offset + 1);
+                local_tee_obj(frame, virtualState.stackPointer - 1, index);
+                nextOffset = offset + 5;
+                break;
+            }
+            case Bytecode.BR_RETURN_CALL: {
+                int paramCount = state.module.symbolTable().functionTypeParamCount(state.thiz.codeEntry.functionIndex());
```
jirkamarsik: `functionTypeParamCount` expects a type index, not a function index. We should pass in `state.thiz.codeEntry.function().typeIndex` or just use `state.thiz.codeEntry.function().paramCount()` directly.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/BinaryParser.java
```diff
@@ -1119,6 +937,65 @@ private CodeEntry readFunction(int functionIndex, int[] locals, int sourceCodeSt
...
+                    } else {
+                        state.addReturnCall(callNodes.size(), callFunctionIndex);
+                        callNodes.add(new CallNode(bytecode.location(), callFunctionIndex, true));
+                    }
+                    state.setUnreachable();
+                    break;
+                }
+                case Instructions.RETURN_CALL_INDIRECT: {
+                    checkTailCallSupport(opcode);
+                    final int expectedFunctionTypeIndex = readUnsignedInt32();
```
jirkamarsik: We can use `readFunctionTypeIndex` here.

jirkamarsik:
The PR fails the official spec tests that try to mix exceptions and tail calls. This is caused by the fact that this approach treats the first tail call in a chain of tail calls as a regular call. This means that the frame, along with its exception handlers, of the first tail caller is kept on the stack and its exception handlers fire while those handlers should have been dropped already.

A quick and dirty fix is to include one new bit of state which is set whenever a function makes the first tail call in a tail-call chain. That bit is then checked when triggering exception handlers and if the bit is set, no exception handlers fire. However, this is yet another new piece of state that has to be managed (e.g. reset in `BR_RETURN_CALL`).

jirkamarsik:
I have finished my review.

Thanks to both Simon and to you for contributing the patch! Let me know if you'd prefer to clean up the PR yourself of if you'd like me to take over.

jirkamarsik:
> A quick and dirty fix is to include one new bit of state which is set whenever a function makes the first tail call in a tail-call chain. That bit is then checked when triggering exception handlers and if the bit is set, no exception handlers fire. However, this is yet another new piece of state that has to be managed (e.g. reset in `BR_RETURN_CALL`).

In an ideal solution, we would always implement `return_call` instructions as tail calls. This would require setting up the trampoline and passing the return call marker to any function which could do an immediate tail call. Whether a function uses tail calls can be detected during module parsing. For direct calls (`call` with a non-imported function), we could be using the trampoline only if the callee uses tail calls. For indirect calls (`call` with an imported function or `call_indirect` or `call_ref`), we would have to use it always. I imagine that would lead to performance regressions but maybe they could be removed by profiling on whether a given call node ever produces a `RETURN_CALL_VALUE`.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/WasmConstant.java
```diff
@@ -51,7 +51,7 @@ public class WasmConstant implements TruffleObject {
     public static final WasmConstant VOID = new WasmConstant("wasm-void-result", WasmType.VOID, false);
     public static final WasmConstant NULL = new WasmConstant("wasm-ref-null", WasmType.NULL, true);
     public static final WasmConstant MULTI_VALUE = new WasmConstant("wasm-multi-value-result", WasmType.MULTI_VALUE, false);
-    public static final Object RETURN_VALUE = new Object();
+    public static final Object RETURN_CALL_VALUE = new Object();
```
jirkamarsik: Maybe we should use a `WasmConstant` here as we do in `MULTI_VALUE` to make the new object readable in case it leaks to some instrumentation tool.

jirkamarsik:
Other than the issues outlined above, all the relevant official spec tests are now passing. I also ran all the Wasm-related benchmarks in our CI and performance is stable.

jirkamarsik (changes requested):

flohuemer:
> > A quick and dirty fix is to include one new bit of state which is set whenever a function makes the first tail call in a tail-call chain. That bit is then checked when triggering exception handlers and if the bit is set, no exception handlers fire. However, this is yet another new piece of state that has to be managed (e.g. reset in `BR_RETURN_CALL`).
> 
> In an ideal solution, we would always implement `return_call` instructions as tail calls. This would require setting up the trampoline and passing the return call marker to any function which could do an immediate tail call. Whether a function uses tail calls can be detected during module parsing. For direct calls (`call` with a non-imported function), we could be using the trampoline only if the callee uses tail calls. For indirect calls (`call` with an imported function or `call_indirect` or `call_ref`), we would have to use it always. I imagine that would lead to performance regressions but maybe they could be removed by profiling on whether a given call node ever produces a `RETURN_CALL_VALUE`.

Identifying whether a callee uses tail calls would potentially require a second pass over the binary during parsing, or some form of fixup. So I would first try this without the special-casing in the parser, but with profiles in the call nodes. The performance impact would be a single return value check. I can try this and run the benchmarks to see if this impacts performance.

My plan would be to move the trampoline intoo the call nodes and the InteropCallAdapterNode - to support tail calls in the main/entry function - and add a profile to all these locations.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: I wasn't sure what counts as a handler, as all other handlers return the next opcode and not an object. I will change my code accordingly.

jirkamarsik: You make a good point! Feel free to keep the current name.

jirkamarsik:
> Identifying whether a callee uses tail calls would potentially require a second pass over the binary during parsing, or some form of fixup.

I was thinking of keeping the same bytecodes and not doing any fixups, instead storing the information that function `i` uses tail calls as a final field in either `WasmModule` or `WasmFunction`. This could be set during the first pass while parsing function `i`. At runtime, this could be queried in `WasmFunctionNode#executeDirectCall` to check whether we can get rid of the trampoline and result check.

flohuemer:
I implemented the trampoline in the call nodes with profiles and a tail-call flag in `WasmFunction`. Most peak-performance benchmarks are stable. However, `digitron` is 6% slower, as it has a lot of indirect calls.

| benchmark | main | tail-calls |
| --- | --- | --- |
| cdf | 5.400 | 5.400 |
| digitron | 3.856 | 3.630 |
| event-sim | 2.056 | 2.047 |
| fft | 8.087 | 8.103 |
| hash-join | 9.108 | 9.160 |
| merge-join | 9.925 | 9.844 |
| phong | 6.037 | 6.036 |
| qsort | 0.945 | 0.955 |
| strings | 0.543 | 0.545 |

Another idea would be to use exceptions instead of returning a tail-call result. However, I think this would make tail calls very slow in both the interpreter and in compiled code.

A solution to this would be that the first tail call would be initiated by an exception and all following would use a return value.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4792,6 +4833,71 @@ private Object executeIndirectCallNode(int callNodeIndex, CallTarget target, Obj
...
+        WasmArguments.setReturnCallMarker(args);
+        if (WasmArguments.getReturnCallMarker(frame.getArguments())) {
+            final WasmLanguage.ReturnCallData data = language.returnCallData();
+            data.target = target;
+            data.args = args;
+            return WasmConstant.RETURN_CALL_VALUE;
+        } else {
+            WasmReturnCallNode callNode = (WasmReturnCallNode) callNodes[callNodeIndex];
+            Object result = callNode.execute(target, args);
+            if (result == WasmConstant.RETURN_CALL_VALUE) {
```
woess: The trampoline preserves the wrong exception-handling context once a tail-call chain contains more than one `return_call`.

Consider:
```wat
(module
  (type $empty (func))
  (tag $e (type $empty))

  (func $c (result i32)
    throw $e)

  (func $b (result i32)
    block $caught
      try_table (result i32) (catch_all $caught)
        return_call $c
      end
      return
    end
    i32.const 2)

  (func $a (result i32)
    block $caught
      try_table (result i32) (catch_all $caught)
        return_call $b
      end
      return
    end
    i32.const 1)

  (func (export "_main")
    call $a
    i32.const 2
    i32.ne
    if unreachable end))
```
Expected: `$b` catches `$c`’s exception, so `$a` returns 2.
Actual: `$a` catches it, returning 1, and `_main` traps.
```
wat2wasm --enable-exceptions --enable-tail-call tail-exception.wat -o tail-exception.wasm
mx wasm --experimental-options --wasm.TailCalls=true tail-exception.wasm
```

jirkamarsik: I think that this expectation is wrong. Tail calls should drop the frame, including any exception handlers. So once `$a` tail calls into `$b`, the handler in `$a` should be dropped. Similarly, when `$b` tail calls into `$c`, the handler in `$b` should be dropped. When `$c` throws, there should be no handler left on the stack and the exception should leave `$a` and be raised to `_main`.

I modified your example:
```
(module
  (type $empty (func))
  (tag $e (type $empty))

  (func $c (result i32)
    throw $e)

  (func $b (result i32)
    block $caught
      try_table (result i32) (catch_all $caught)
        return_call $c
      end
      return
    end
    i32.const 2)

  (func $a (result i32)
    block $caught
      try_table (result i32) (catch_all $caught)
        return_call $b
      end
      return
    end
    i32.const 1)

  (func (export "_main")
    block $caught
      try_table (result i32) (catch_all $caught)
        call $a
      end
      return
    end
    unreachable))
```
and tested with `wasmtime`:

```
jirka@JMARSIK:~$ wat2wasm --enable-exceptions --enable-tail-call tail-exception.wat -o tail-exception.wasm
jirka@JMARSIK:~$ wasmtime run -Wexceptions --invoke _main tail-exception.wasm
Error: failed to run main module `tail-exception.wasm`

Caused by:
    0: failed to invoke `_main`
    1: error while executing at wasm backtrace:
    0:     0x61 - <unknown>!<wasm function 3>
    2: wasm trap: wasm `unreachable` instruction executed
```

woess: I stand corrected. Thanks!

woess:
Comments aside, LGTM, although I'm not a big fan of the return call marker argument; would prefer if we could get rid of it (eventually).

jirkamarsik:
> Comments aside, LGTM, although I'm not a big fan of the return call marker argument; would prefer if we could get rid of it.

The more elegant fix to the semantic issue with exceptions that moves the trampoline one frame lower also removes the need for the return call marker, since all tail calls are now guaranteed to be supported by trampolines. However, it incurs an extra check on every indirect function call, since you can never know if the callee won't suddenly decide to make a tail call. @flohuemer already implemented this approach and reported a 6% peak slowdown on `digitron`. It might be worth it to avoid these hacks but then again I'm not happy to see every indirect call having to pay the price for tail calls.

flohuemer:
I will have another look into the performance numbers tomorrow. Might actually be a different problem I am seeing.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: Fixed

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/WasmContextOptions.java
flohuemer: Fixed

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: Fixed

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: You are correct. This also drops the legacy catch slots. Fixed.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: Added

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: Fixed

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: Fixed

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/WasmConstant.java
flohuemer: Fixed

flohuemer:
I updated the PR and tried to address all your requests.
It seems that `digitron` is pretty unstable on my local machine. Can you please check if there are any regressions on your benchmark infrastructure.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4653,28 +4654,22 @@ private static int miscHandler(int offset, State state, VirtualState virtualStat
...
-                nextOffset = state.thiz.bytecodeStartOffset;
-                virtualState.stackPointer = state.thiz.codeEntry.stackBase();
+                // clear the entire stack (including legacy catch slots), except for the parameters
+                dropStack(frame, virtualState.stackPointer, paramCount);
+                // reinitialize all remaining locals
                 WasmFrame.initializeLocals(frame, paramCount, state.thiz.codeEntry);
+
+                nextOffset = state.thiz.bytecodeStartOffset;
+                virtualState.stackPointer = state.stackBase;
+                state.activeLegacyCatchCount = state.startActiveLegacyCatchCount;
```
jirkamarsik: We don't need `startActiveLegacyCatchCount`. We are jumping to the top of the function, outside of any catch clauses. `executeBodyFromOffset` is only executed with a non-zero `startActiveLegacyCatchCount` when resuming mid-function via OSR.

jirkamarsik: For cases when `startActiveLegacyCatchCount` is non-zero (we are invoked from OSR), it would actually be wrong to reset to `startActiveLegacyCatchCount`, since the legacy-catch depth was only valid at the point of OSR transfer and not at the top of the function.

jirkamarsik:
When calling the start function in `Linker#runStartFunction`, we should install a trampoline in case the start function tail calls. I think other calls into Wasm should be handled now (internal calls use the call nodes from `WasmFunctionNode` and external calls use the call nodes from `InteropCallAdapter`).

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
```diff
@@ -4636,6 +4613,65 @@ private static int miscHandler(int offset, State state, VirtualState virtualStat
                 }
                 break;
             }
+            case Bytecode.LOCAL_GET_I32: {
```
jirkamarsik, line 4614: We need to increment all of the `offset + X` expressions in the `LOCAL_*_I32` clauses by 1 to account for the `MISC` prefix.

jirkamarsik:
We should add a changelog entry for this new feature (I expect this to land in 25.4.4).

jirkamarsik (changes requested):
I ran benchmarks on the CI and everything looks stable. 👏

The PR's looking great now! I posted a few more review comments and then we should be good to merge.

flohuemer:
> When calling the start function in `Linker#runStartFunction`, we should install a trampoline in case the start function tail calls. I think other calls into Wasm should be handled now (internal calls use the call nodes from `WasmFunctionNode` and external calls use the call nodes from `InteropCallAdapter`).

I added the relevant calls in the linker.

> We should add a changelog entry for this new feature (I expect this to land in 25.4.4).

I added a changelog entry.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: Fixed

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: Thanks for pointing that out. Fixed.

wasm/src/org.graalvm.wasm/src/org/graalvm/wasm/nodes/WasmFunctionNode.java
flohuemer: I added a test to make sure exceptions work as expected.

jirkamarsik (approved):
Looks great! Thank you very much!

PR #14284, opened 2026-08-20
[GR-79516] Add GC-agnostic garbage collector tests

JohnTortugo:
Please, review this PR to add add three test classes to `com.oracle.svm.test` to verify garbage collector correctness in Native Image.

Before this change, the GC-related tests in `com.oracle.svm.test` covered JFR event emission (`jfr/TestGCEvents`, `jfr/TestGCHeapSummaryEvent`, `jfr/TestGarbageCollectionEvents` and friends), the JMX management interface (`jmx/JmxTest`, via `GarbageCollectorMXBean` and `MemoryMXBean`) and native memory accounting (`nmt/NativeMemoryTrackingTests`). Those verify that GC information is reported, not that the collector behaves correctly: for example `JmxTest` asserts `getCollectionCount() >= 0`, which holds even if collection is completely broken. Nothing verified that references are cleared and enqueued, that the object graph is intact after a collection, or that atomic reference updates behave correctly while collecting.

The new tests close that gap:

- **`GCReferenceTest`** (7 tests) — weak and phantom references must be cleared and enqueued once their referent becomes unreachable; a strongly reachable referent must *never* be cleared; a `Cleaner` action must run. Also scales to 1000 references to catch a collector that only processes some of the references it discovers.

- **`GCObjectGraphTest`** (6 tests) — linked chains (10k nodes), object arrays (20k elements) and `HashMap` contents (5k entries) must be unchanged after collections; reference fields must hold the last value written to them; cyclic garbage must not disturb live data. This is where a missing or incorrect reference update after the collector moves an object becomes visible immediately, instead of as an unrelated crash much later.

- **`GCAtomicReferenceTest`** (6 tests) — compare-and-set and get-and-set on reference fields must behave correctly while collections happen, via `AtomicReference`, `VarHandle`, `AtomicReferenceArray`, and from four concurrent threads. Atomic reference updates combine a hardware atomic with GC bookkeeping, so they are covered separately from plain stores. Notably, a single-threaded compare-and-set must always succeed; a spurious failure would mean the VM reported a mismatch that cannot exist in the program.

The tests are deliberately **GC-agnostic**: collections are provoked only with `System.gc()` and allocation pressure — never with GC-specific options, VM internals or the WhiteBox API — so they are valid for any current or future Native Image collector, including concurrent ones. Assertions that depend on a collection having happened retry until a timeout rather than assuming a single `System.gc()` suffices, which keeps them correct for a collector that finishes its work on background threads.

## Testing

- `mx --primary-suite=substratevm native-unittest com.oracle.svm.test.GCReferenceTest com.oracle.svm.test.GCObjectGraphTest com.oracle.svm.test.GCAtomicReferenceTest` → `OK (19 tests)`, exit 0.
- Full suite: `mx --primary-suite=substratevm native-unittest` → `OK (280 tests)`, exit 0. The new classes are picked up automatically by the harness; no regressions.
- `mx --primary-suite=substratevm --components=ni,nju build` → clean, including checkstyle.
- Platform tested: `linux-aarch64` with the default (Serial) GC. Not yet run on `linux-amd64` or other platforms; the tests contain no platform- or collector-specific code.

## Documentation

No documentation updates needed — this change only adds tests. The intent of each test, and the specific bug class it is designed to catch, is documented in Javadoc on the test classes and methods.

## Contributor Checklist

### DIFF
--- a/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCAtomicReferenceTest.java
+++ b/substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCAtomicReferenceTest.java
@@ -1,0 +1,261 @@
+/*
+ * Copyright (c) 2026, 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package com.oracle.svm.test;
+
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Tests atomic operations on reference fields while the garbage collector is active.
+ *
+ * Ordinary reference stores and atomic ones (compare-and-set, get-and-set) are handled by different
+ * code in the VM: an atomic operation has to both perform the hardware atomic instruction and do the
+ * GC bookkeeping, without breaking either. That combination is easy to get subtly wrong, and when a
+ * collector moves objects at the same time it gets harder still, because the value sitting in the
+ * field may be the old location of an object that has just been moved. A buggy implementation can
+ * then make a compare-and-set fail when it should have succeeded, or leave a stale reference behind.
+ *
+ * Each test therefore performs a lot of atomic reference traffic while allocating and collecting, and
+ * verifies afterwards that the values are exactly what the atomic operations should have produced.
+ * The tests work with any garbage collector and use only standard Java APIs.
+ */
+public class GCAtomicReferenceTest {
+
+    /** Written to a volatile field so the compiler cannot delete the allocations. */
+    private static volatile Object sink;
+
+    private static void allocateGarbage() {
+        sink = new byte[16 * 1024];
+        sink = null;
+    }
+
+    /** A value object that remembers which iteration created it. */
+    private static final class Value {
+        private final int id;
+
+        Value(int id) {
+            this.id = id;
+        }
+    }
+
+    /** Target of the {@link VarHandle} test below. */
+    private static final class Holder {
+        @SuppressWarnings("unused") private volatile Value field;
+    }
+
+    /**
+     * Repeatedly replaces the contents of an {@link AtomicReference} with
+     * {@code compareAndSet}, allocating and collecting along the way.
+     *
+     * Every step must succeed, because the test is single-threaded and always passes the value it
+     * just read as the expected value. A failed compare-and-set here would therefore mean the VM
+     * reported a mismatch that cannot exist in the program itself, which is exactly the symptom of a
+     * GC that moved the referenced object without the atomic operation accounting for it.
+     */
+    @Test
+    public void testCompareAndSetAlwaysSucceedsWhileCollecting() {
+        final int iterations = 20_000;
+        AtomicReference<Value> ref = new AtomicReference<>(new Value(0));
+
+        for (int i = 1; i <= iterations; i++) {
+            Value expected = ref.get();
+            Value updated = new Value(i);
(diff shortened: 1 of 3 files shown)


JohnTortugo:
Can anyone take a look please?

oubidar-Abderrahim:
@wirthi  could you please assign a reviewer to this PR? thank you

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCAtomicReferenceTest.java
```diff
@@ -0,0 +1,261 @@
...
+ * then make a compare-and-set fail when it should have succeeded, or leave a stale reference behind.
+ *
+ * Each test therefore performs a lot of atomic reference traffic while allocating and collecting, and
+ * verifies afterwards that the values are exactly what the atomic operations should have produced.
+ * The tests work with any garbage collector and use only standard Java APIs.
+ */
+public class GCAtomicReferenceTest {
+
+    /** Written to a volatile field so the compiler cannot delete the allocations. */
+    private static volatile Object sink;
```
christianhaeubl, line 54: Please either use `GraalDirectives.blackhole(...)` or something like the pattern that I posted below. Please also do the same for the `sink` fields in the other classes.
```
@NeverInline("Prevent optimizations.")
private static byte[] allocateGarbage() {
    return new byte[16 * 1024];
}
```

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCAtomicReferenceTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+    private static void allocateGarbage() {
```
christianhaeubl, line 56: We typically try to move helper functions and classes to the bottom so that the most relevant parts are visible right away when someone opens a file.

substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCReferenceTest.java
```diff
@@ -0,0 +1,261 @@
...
+    /**
+     * Keeps asking for a garbage collection until {@code condition} becomes true, then returns. If
+     * the condition is still false when the timeout expires, the test fails.
+     *
+     * Why the loop is necessary: a single {@link System#gc()} is only a hint. Some collectors need
+     * allocation to happen before they start a cycle, and a concurrent collector does its work on
+     * background threads, so the result may only be visible a little later. Allocating garbage and
+     * sleeping briefly on each attempt covers both cases.
+     */
+    private static void awaitGC(BooleanSupplier condition, String message) {
```
christianhaeubl, line 89: For this helper method (and maybe a small number of others), it might be useful to move them to a dedicated class in case that more tests are added later on.

christianhaeubl (commented):
Thanks for working on those tests, I added some minor comments. Additionally, some AI feedback is below:

1. [Medium] Reference tests fail with the supported Epsilon GC
   graal: substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCReferenceTest.java:89-106

   Problem: Reference-processing tests cannot succeed with the supported Epsilon collector.
   
   Evidence: `awaitGC` fails unless an unreachable referent is processed within 30 seconds. Five tests depend on that behavior, while Epsilon uses the never-collect policy and makes explicit collection requests no-ops. The class contains no collector restriction despite stating that it works with any collector.
   
   Impact: An Epsilon run produces five unavoidable failures and can spend roughly 150 seconds waiting for conditions that cannot become true.

2. [Low] Cyclic-garbage test never verifies cycle collection
   graal: substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCObjectGraphTest.java:195-210

   Problem: The cyclic-garbage test never observes whether any cycle is reclaimed.
   
   Evidence: Each cycle is discarded immediately, and every assertion examines only the unrelated live list. The test therefore passes when none of the cycles is collected. Because the cycle allocations do not escape, compiler optimization can also remove them entirely.
   
   Impact: Regressions in tracing or reclaiming cyclic garbage can pass a test whose name and documentation claim to cover that behavior.

3. [Low] CAS completes before the collection it is meant to cross
   graal: substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCAtomicReferenceTest.java:89-98

   Problem: The single-threaded CAS variants do not guarantee a collection between loading the expected reference and executing the CAS.
   
   Evidence: Every explicit collection request occurs after the CAS has completed. Constructing the replacement value could trigger an allocation collection, but the test does not ensure that it does. The VarHandle and array-element variants have the same ordering.
   
   Impact: These variants can complete without exercising a CAS whose expected referent spans a moving collection, allowing defects in that interaction to remain undetected.

4. [Low] GC completion at the timeout boundary is reported as failure
   graal: substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCReferenceTest.java:89-106

   Problem: Successful completion near the timeout boundary can be reported as failure.
   
   Evidence: The deadline is checked before the condition. If the condition becomes true after the final check but before the deadline, the subsequent sleep can carry the loop past the deadline and the method fails without observing the result. The deadline also uses wall-clock time, making it sensitive to clock adjustments.
   
   Impact: Slow or concurrent reference processing can cause intermittent failures even when it completes within the nominal timeout.

5. [Low] Workers can finish before any collection is requested
   graal: substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCAtomicReferenceTest.java:250-257

   Problem: The concurrent CAS test can skip every explicit collection request.
   
   Evidence: Workers are released before the collection loop evaluates `done.getCount()`. If the test thread is descheduled after releasing them, the workers can finish all updates and reduce the latch to zero before that evaluation, preventing the loop body from running while the final counter assertion still passes.
   
   Impact: Some executions cover ordinary concurrent CAS behavior without covering atomic reference operations interacting with an explicitly requested collection.

6. [Low] Non-escaping atomic target can eliminate the tested CAS
   graal: substratevm/src/com.oracle.svm.test/src/com/oracle/svm/test/GCAtomicReferenceTest.java:85-101

   Problem: Escape analysis can eliminate the atomic operation being tested.
   
   Evidence: The `AtomicReference` is local and never passed to another thread or opaque consumer. Graal supports virtualizing compare-and-set operations on non-escaping virtual instances, so the reference can be scalar-replaced and its CAS reduced without emitting the underlying atomic operation.
   
   Impact: This method can pass without executing the hardware atomic operation or its GC barriers, so it does not establish the implementation-path coverage claimed by its documentation.

PR #14319, opened 2026-08-26
FIX: Record fast accessor method dependency

JohnTortugo:
### Summary

Record accessor methods handled by BytecodeParser.tryFastInlineAccessor in the graph’s method dependency list.

### Problem

The fast accessor inlining path emits a field load directly instead of creating a nested BytecodeParser. As a result, it bypasses the graph.recordMethod(method) call normally performed by BytecodeParser.build().

When JVMTI class redefinition or retransformation is enabled, HotSpot converts the recorded methods into evol_method dependencies. Without the accessor in this list, retransformation of its declaring class does not invalidate compiled
callers containing its inlined body.

This can leave stale compiled code active. The issue was reproduced using Mockito’s inline mock maker:

1. Compile and inline a simple getter into a hot caller.
2. Retransform the getter’s declaring class while creating a mock.
3. Direct calls observe the transformed implementation, but the compiled caller continues executing the original inlined field read.

### Fix

Call graph.recordMethod(targetMethod) when the fast accessor path successfully inlines a getter. This restores the dependency that would have been recorded by the nested parser.

Add a regression test verifying that a fast-inlined accessor is present in StructuredGraph.getMethods().

### Testing

- New regression test fails before the fix and passes afterward.
- Full InliningTest: 37 tests passed, 2 ignored.
- Checkstyle passed for both modified files.
- Mockito/JUnit reproducer passed 10 consecutive runs with the rebuilt libgraal.
- Post-fix LogCompilation output includes both:

<dependency type='evol_method' x0='...Victim getValue ()Ljava/lang/String;'/>
<dependency type='evol_method' x0='...Caller call (...)'/>

### Contributor Checklist

### DIFF
--- a/compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/core/test/inlining/InliningTest.java
+++ b/compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/core/test/inlining/InliningTest.java
@@ -69,6 +69,29 @@ import jdk.vm.ci.meta.TriState;
 
 public class InliningTest extends GraalCompilerTest {
 
+    private static final class FastAccessor {
+        private final int value;
+
+        FastAccessor(int value) {
+            this.value = value;
+        }
+
+        int getValue() {
+            return value;
+        }
+    }
+
+    public static int fastAccessorSnippet(FastAccessor accessor) {
+        return accessor.getValue();
+    }
+
+    @Test
+    public void testFastAccessorRecordsMethodDependency() {
+        StructuredGraph graph = assertInlined(getGraph("fastAccessorSnippet", false));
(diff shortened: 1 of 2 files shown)


JohnTortugo:
Any concerns here? @dougxc

dougxc:
Looks good to me - I'll integrate it.

JohnTortugo:
Thank you @dougxc

JohnTortugo:
@dougxc - feel free to close this PR once you integrate the commits in your PR. TY.

dougxc:
Will do. The integration is making its way through our internal gates...

dougxc:
Merged as [d4b847f55ea2fe951d75dc6f873e49e80edcffa](https://github.com/oracle/graal/commit/fd4b847f55ea2fe951d75dc6f873e49e80edcffa)

PR #14394, opened 2026-09-07
[GR-79489] Use EVEX for vcvtph2ps/vcvtps2ph when an operand is xmm16-31

mukel:
**Human context:**

I started hitting intermittent compile errors with Native Image, which means no image generated, but only on my AVX512-capable workstation. The compiler error was clear, some high-register xmm16+ was being allocated for the `float <-> float16` conversions, which only supports VEX (xmm0-15).

I reviewed the changes carefully, I understand them fully, and I deemed that the fix is clean and simple; it also include regression tests.  

_For reference: the AI first attempt and recommended fix was a whole refactor of the register allocator ..._

**AI PR description (CC):**

`Float.float16ToFloat` and `Float.floatToFloat16` are intrinsified to `vcvtph2ps`/`vcvtps2ph`. Since #24464 (GR-43051) the register allocator can place their operands in xmm16-31, but both macros were always VEX-encoded, which cannot address those registers and fails during code emission with:

```
GraalError: failed guarantee: instruction VCVTPH2PS illegal operand xmm22 for category XMM
```

This picks the EVEX variant (`EVCVTPH2PS`/`EVCVTPS2PH`, already defined) when an operand is a high register and keeps the shorter VEX encoding for xmm0-15, same approach as `AMD64MacroAssembler.movflt`/`movdbl`.

Tests:
- `AMD64Float16HighRegisterEncodingTest`: both macros encode as EVEX for xmm16/xmm22/xmm31 and stay VEX for low registers. Host-independent (synthetic AVX-512 target, encoding only).
- `AMD64Float16ConversionCompilationTest`: compiles kernels that keep 20 conversion operands live so the allocator uses xmm16-31, and asserts a conversion op actually landed in a high register. Requires full AVX-512, skipped otherwise.

Without the fix, 4 of the 5 tests fail with the error above.

### DIFF
--- a/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/asm/amd64/AMD64Assembler.java
+++ b/compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/asm/amd64/AMD64Assembler.java
@@ -6528,11 +6528,11 @@ public class AMD64Assembler extends AMD64BaseAssembler implements MemoryReadInte
     }
 
     public final void vcvtph2ps(Register dst, Register src) {
-        VexRMOp.VCVTPH2PS.emit(this, AVXSize.XMM, dst, src);
+        VexRMOp.VCVTPH2PS.encoding(avxEncoding).emit(this, AVXSize.XMM, dst, src);
     }
 
     public final void vcvtps2ph(Register dst, Register src, int imm8) {
-        VexMRIOp.VCVTPS2PH.emit(this, AVXSize.XMM, dst, src, imm8);
+        VexMRIOp.VCVTPS2PH.encoding(avxEncoding).emit(this, AVXSize.XMM, dst, src, imm8);
     }
 
     public final void vmovdqu(Register dst, AMD64Address src) {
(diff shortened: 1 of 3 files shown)


compiler/src/jdk.graal.compiler/src/jdk/graal/compiler/asm/amd64/AMD64Assembler.java
```diff
@@ -6527,12 +6527,24 @@ public final void unpcklpd(Register dst, Register src) {
...
+     * The F16C conversions keep the shorter VEX encoding for xmm0-15 and only switch to EVEX when an
+     * operand is an AVX-512 high register (xmm16-31), which VEX cannot address.
+     */
     public final void vcvtph2ps(Register dst, Register src) {
-        VexRMOp.VCVTPH2PS.emit(this, AVXSize.XMM, dst, src);
+        if (isAVX512Register(dst) || isAVX512Register(src)) {
+            VexRMOp.EVCVTPH2PS.emit(this, AVXSize.XMM, dst, src);
+        } else {
+            VexRMOp.VCVTPH2PS.emit(this, AVXSize.XMM, dst, src);
+        }
```
gergo-: Our general invariant is to make the choice of instruction based on the assembler's general configuration, so that we never mix VEX and EVEX encodings. Even if a particular instance of an operation could indeed be VEX-encoded and save some bytes. I think the following should do it:
```suggestion
        VexRMOp.VCVTPH2PS.encoding(avxEncoding).emit(this, AVXSize.XMM, dst, src);
```

mukel: Done! Fixed for both ops and updated the tests e.g. there is a test that checked VEX was used for low registers, removed.

gergo- (approved):
Looks good, thanks for this contribution. I'll get this through the internal review/merge dance.

PR #14396, opened 2026-09-07
Preserve floating guards during empty-loop removal

### LINKED ISSUE #14395: [GR-79463] RemoveEmptyLoopsPhase removes loops containing loop-variant floating guards
**Describe the issue**

`RemoveEmptyLoopsPhase.tryRemoveEmptyLoop` determines whether a loop is empty by walking the fixed nodes starting at `loopBegin().next()`. This walk does not see floating guards.

A guard anchored at the loop header or at a dominating block before the loop can still belong to the loop when its condition depends on a loop value. The current implementation may remove such a loop and replace its induction variable with the exit value. This skips the per-iteration guard evaluation and also changes frame states that use the induction variable.

This was observed in `java.util.jar.Attributes.read`, where a valid manifest was rejected with:
```stacktrace
java.io.IOException: invalid header field (line 1)
```
The compiled code tested the byte at `len - 1` instead of scanning the line. A later deoptimization resumed at the loop header with `len - 1` as the scan index.

1. Parse a counted loop whose fixed body only updates its induction variable.
2. Add a floating `GuardNode` with no usages and a condition depending on the loop's `ValuePhiNode`.
3. Anchor the guard either at the loop header or at a block dominating the loop.
4. Run `RemoveEmptyLoopsPhase` while floating guards are allowed.
5. Check whether the loop and guard remain alive.

Expected result:

The loop remains because the guard must be evaluated using each iteration's value.

Actual result:

The loop is removed and the induction variable is replaced with its exit value.

A phase-level reproducer is included in `NekoCaffeine/graal@37bc8f02ba0e`.

- GraalVM version: Oracle GraalVM 25.1.3+9.1 (`25.0.3+9-LTS-jvmci-25.1-b19`)
- Source revision used for patch validation: `8d3ee1a162d0646c1a8a9f957e738ed801cacf96`
- JDK major version: 25
- OS: Fedora Linux 44
- Architecture: AMD64

**More details**

Immediately before empty-loop removal, the guard condition in the failing compilation depended on the loop through the following chain:

ValuePhi -> Pi -> ZeroExtend -> Add -> OffsetAddress
          -> FloatingRead -> IntegerEquals -> Guard

The guard's anchor dominated the loop. Immediately after `RemoveEmptyLoopsPhase`, the loop was gone and the Phi had been replaced by its exit value.

PRs #10770 and #10842 handle the corresponding floating-guard scheduling constraint for loop peeling. They do not cover the empty-loop removal decision.

NekoCaffeine:
## Summary

`RemoveEmptyLoopsPhase` currently determines whether a loop is empty by walking its fixed control flow. This misses floating guards whose anchors are outside the loop but whose conditions depend on loop values.

Check the complete loop fragment for floating guards before removing a loop that may be entered. The existing loop-fragment membership rules account for both the guard anchor and its condition.

The new tests cover guards anchored at the loop header and before the loop, including a guard depending on a floating read. They also verify that genuinely empty loops and loops with invariant guards outside the loop are still removed.

## Related Issues

- Fixes #14395
- Related to #10770 and #10842.

## Testing

Tested from source revision `8d3ee1a162d0646c1a8a9f957e738ed801cacf96` with Labs JDK `25.4-b23`.

- Built the GraalJDK CE and compiler test distributions.
- Ran `ConvertDeoptimizeToGuardPrioritiesTest`, `GuardPrioritiesTest`, `LoopFragmentTest`, `LoopsDataTest`, and `RemoveEmptyLoopsPhaseTest`: 15 tests passed.
- Ran Checkstyle on both changed Java files.

On the pre-fix implementation, `testLoopHeaderGuard`, `testPreheaderGuard`, and `testFloatingReadGuard` fail because the loop is removed.

## Documentation

No documentation changes are needed. This is an internal compiler correctness fix and does not change a public API or supported workflow.

## Contributor Checklist

### DIFF
--- a/compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/loop/test/RemoveEmptyLoopsPhaseTest.java
+++ b/compiler/src/jdk.graal.compiler.test/src/jdk/graal/compiler/loop/test/RemoveEmptyLoopsPhaseTest.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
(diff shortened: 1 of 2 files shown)


gergo-:
Done as https://github.com/oracle/graal/pull/14462.

PR #14398, opened 2026-09-07
Fix Ruby flavor extended-mode comments to always end at a newline

### LINKED ISSUE #4435: Regexp /x mode: comment line ending in backslash swallows the next line (MRI parity; breaks metanorma-standoc)
## Summary

In extended (`/x`) mode, a `#` comment line that ends with a backslash has the newline treated as escaped: the comment continues onto the next line and swallows it. MRI ends `/x` comments at the newline unconditionally. When the swallowed line carries a closing paren, the pattern dies at match time with `missing ), unterminated subpattern`.

## Minimal repro

```ruby
# The regexp compiler sees a literal backslash+newline inside the comment
# (Regexp.new keeps it honest — a /.../x literal would be line-continued
# by ruby's own lexer on both implementations, which is NOT what this is about).
src = "(a) # c\\\n(b)"
m = Regexp.new(src, Regexp::EXTENDED).match("ab")
p(m && m[2])
```

- MRI 3.4.8: `"b"` (the comment ends at the newline; `(b)` is live)
- TruffleRuby 34.0.1: `nil` (the comment swallowed the `(b)` line)

And the shape that errors — the swallowed line carries the group's closing paren:

```ruby
src2 = "(x # c\\\n)"
Regexp.new(src2, Regexp::EXTENDED).match("x")
```

- MRI 3.4.8: matches
- TruffleRuby 34.0.1: `Polyglot::ForeignException: missing ), unterminated subpattern` (raised lazily at match time via `Truffle::RegexpOperations.match_from`; the literal itself "compiles")

## Environment

`truffleruby 34.0.1 (like ruby 3.4.9), GraalVM CE Native, arm64-darwin23` (truffleruby-community-34.0.1-macos-aarch64 standalone).

## Real-world impact

Found dogfooding metanorma under TruffleRuby: `metanorma-standoc` 3.4.9's `LinkProtectPreprocessor` builds its pass-inline regexp from a heredoc whose last comment line ends with a literal backslash (`# [ ... ] not preceded by \\` → the pattern string ends that line with a single `\`):

- `lib/metanorma/converter/macros_nosub.rb:27` and `:134` (two classes share the construct)

Under TruffleRuby the closing `)` on the next line is swallowed, so `String#split`/`Regexp#match?` raise and every compile aborts during boilerplate processing. Under MRI this has worked forever.

A gem-side reword of the comment (so it no longer ends in a backslash) works around it, but the engine divergence looks unintended — MRI's `/x` treats the newline as the comment terminator regardless of a trailing backslash.

eregon:
## Summary

A backslash at the end of a comment line must not escape the newline: Onigmo ends extended-mode comments at a newline unconditionally. The escape handling was inherited from duplicating the Python flavor.

cc @djoooooe 

## Related Issues

Fixes https://github.com/truffleruby/truffleruby/issues/4435

## Testing

Added test.

## Documentation

- Describe any documentation updates included in this pull request.
- If no documentation updates are needed, state that explicitly.

## Contributor Checklist

### DIFF
--- a/regex/src/com.oracle.truffle.regex.test/src/com/oracle/truffle/regex/tregex/test/RubyTests.java
+++ b/regex/src/com.oracle.truffle.regex.test/src/com/oracle/truffle/regex/tregex/test/RubyTests.java
@@ -384,6 +384,13 @@ public class RubyTests extends RegexTestBase {
     }
 
     @Test
+    public void github4435() {
+        // In extended mode, a newline always ends a comment, even when preceded by a backslash.
+        test("(a) # c\\\n(b)", "x", "ab", 0, true, 0, 2, 0, 1, 1, 2);
+        test("(x # c\\\n)", "x", "x", 0, true, 0, 1, 0, 1);
+    }
+
+    @Test
     public void beginningAnchor() {
         test("\\Ga", "", "a", 0, true, 0, 1);
         test("\\Ga", "", "ba", 0, false);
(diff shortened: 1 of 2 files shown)


djoooooe (approved):
lgtm

PR #14405, opened 2026-09-08
[GR-NONE] [sulong] Execute AVX/AVX2/FMA intrinsics emitted by Eigen/Ceres under -mavx2 -mfma

masariello:
## Summary

Sulong cannot currently execute much of the bitcode clang emits for AVX/AVX2/FMA
code. Compiling a real numerical workload (Eigen 3.4, Ceres Solver 2.2) with
`-mavx2 -mfma` and running it under `lli` fails on missing intrinsics, missing
256-bit vector paths, and unsupported Eigen inline asm. This PR closes those
gaps and adds an end-to-end validation harness for them.

Three things are being fixed:

**1. Coverage of the generic (non-`llvm.x86.*`) intrinsics clang emits.**
Most AVX/FMA code from clang lowers to width-generic LLVM IR, not to x86
builtins, so the fix is to make the width-generic dispatch paths complete
rather than to add more `llvm.x86.*` special cases.

- `minnum`/`copysign`: width-generic vector nodes.
- Rounding: added `trunc`/`nearbyint`/`roundeven`; `ceil`/`round` routed through
  the generic vector path. `LLVMRound` corrected to round-half-away-from-zero
  (`round(-2.5) == -3`).
- `minimum`/`maximum`: IEEE-754-2019 scalar + vector nodes.
- `vector.reduce.fmin` / `.fmul` added to the vector-op matcher.
- `ctpop`/`ctlz`/`cttz`: width-generic vector nodes.
- `{s,u}{add,sub}.sat`: one length-generic factory replaces the hardcoded
  scalar + `v8i16` cases.
- `llvm.fmuladd.vN{f32,f64}` for arbitrary `N`.
- `llvm.masked.load/store` for `v4f64` (LLVM 21+ signature) and `v<N>i32`.
- Wide-vector bitcasts (`<4 x double>` <-> `i256`, etc.).

**2. Numerical-correctness fixes surfaced by running Eigen and Ceres.**

- Bitcast `float`/`double` -> integer now uses `Float.floatToRawIntBits` /
  `Double.doubleToRawLongBits`. The non-raw variants canonicalise every NaN to
  a positive NaN and drop the sign bit, so `signbit()` and `copysign()` on a NaN
  diverged from x86 (`copysign(NaN, -1)` must stay a negative NaN). A bitcast is
  a bit-preserving reinterpret, so raw bits are correct for all values.
- `llvm.ldexp` uses `Math.scalb` instead of `value * Math.pow(2, exp)`, which
  produced `0 * +inf = NaN` for `value == 0` with a large exponent and failed
  Eigen's `packetmath` `pldexp` "zero stays zero" check.
- `llvm.log2` (and the constrained variant) decomposes finite positive `x` as
  `m * 2^e`, `m` in `[0.5, 1)`, returning `e + log(m)/log(2)`. The naive
  `log(x)/log(2)` was ~1 ulp off (`log2(100)` gave `6.6438561897747253` instead
  of `6.6438561897747244`) and failed Ceres' `Jet.Log2` finite-difference check.
- `llvm.experimental.constrained.fma` routes to the existing fused `LLVMFma`
  node (single-rounding `Math.fma`), ignoring the FP-environment metadata args
  like the other constrained builtins. Fixes Ceres' `Jet.Fma`.

**3. x86 specifics that remain genuinely x86.**

- `llvm.x86.avx.{round,rsqrt,min,max}` 256-bit intrinsics.
- Eigen inline asm: `vfmadd231ps/pd` (fused), `vminps/pd`, `vmaxps/pd`, plus
  AT&T `#` comments so `EIGEN_ASM_COMMENT` parses.
- `xgetbv` support, needed by the runtime CPU-feature detection Eigen/Ceres do.

## Related Issues

None filed — this came out of trying to run Eigen and Ceres under `lli`.

## Testing

- **SulongSuite**: new `avx2fma` test suite with 12 self-checking differential
  tests (`bitcount256`, `eigen_matmul_smoke`, `fma`, `fmuladd_vec`,
  `fp_copysign256`, `fp_minmax256`, `fp_round256`, `minmax256`, `reduce256`,
  `round256`, `rsqrt256`, `sat256`), plus `xgetbv001` in the existing
  `inlineasm` suite. Each checks Sulong's result against what the same source
  produces natively.
- **End-to-end, in CI**: a new GitHub Actions workflow
  (`.github/workflows/sulong-avx-fma-tests.yml`) builds Sulong, then compiles
  Eigen 3.4.0 and Ceres Solver 2.2.0 from source with the Sulong bootstrap
  toolchain and runs their own unit-test suites under `lli`. The superbuild
  lives in `sulong/tests/avx-fma-eigen-ceres/` and runs two variants:
  - `bitcode`: pure-bitcode Eigen + Ceres, which is what exercises the 256-bit
    AVX2/FMA packet paths (Eigen `packetmath`, `vectorization_logic`) and Ceres
    dense linear algebra via Eigen.
  - `mkl`: Ceres re-run with `LAPACK=ON` binding Intel MKL's `libmkl_rt.so`
    from bitcode through Sulong's native interface. Skipped cleanly when MKL is
    not available on the runner.

  Latest run: **all green** — Eigen 77/77 and Ceres 84/84 under `lli` in the
  `bitcode` variant, Ceres 84/84 again in the `mkl` variant (MKL was present on
  the runner, so that path was exercised rather than skipped). Wall time
  2h03m56s on `ubuntu-latest`.

  A CMake toolchain file declares a cross-compile whose
  `CROSSCOMPILING_EMULATOR` is `lli`, so every `add_test` runs on the Sulong
  LLVM runtime; an `add_test` shim covers the two CMake signatures that
  otherwise suppress the emulator prepend. `ctest --output-junit` XMLs are
  uploaded as workflow artifacts.
- Output cross-checked against `clang` 22 for the `linux` triple.

## Documentation

- `sulong/docs/contributor/AVX-INTRINSICS.md` added — reframed around
  generic-IR completeness rather than a list of x86 builtins, and corrects the
  stale claim that `llvm.fma` was unsupported.

## Contributor Checklist


masariello:
I am covered by the Oracle Contributor Agreement — I signed it before opening this pull request.

The `oca/oracle` check is still red here. Leaving this comment as the [GraalVM contributor guide](https://www.graalvm.org/community/contributors/) directs:

> If there is a delay in showing you have signed OCA, or the bot leaves a message on your pull request that you have not signed the OCA, leave a comment on the pull request stating that you are now covered by the OCA.

In case the check database needs a manual update, the two identities the bot flagged are:

- GitHub username: `masariello`
- Commit author email: `masariello@gmail.com`

Happy to provide anything else that helps resolve it.

PR #14407, opened 2026-09-09
[GR-NONE] [truffle] [sulong] Make native library loading on Windows behave like POSIX

masariello:
## Summary

Loading and calling native libraries from Sulong works on Linux and macOS but
degrades badly on Windows, in two independent ways.

**1. The `default` library resolved almost nothing.**

`dlopen(NULL)` / `RTLD_DEFAULT` on POSIX searches the main executable *and every
shared object loaded into the process*. The Windows implementation used
`GetProcAddress(GetModuleHandleA(NULL), name)`, which searches only the export
table of the main executable — for a JVM launcher that table is essentially
empty. `NFIContextExtension` uses the default library handle as the last-resort
fallback for every unresolved native symbol, so on Windows that fallback
silently resolved nothing.

The NFI test suite had already worked around this rather than fixing it:
`NFITest.lookupAndBindDefault` redirected to `reexport_`-prefixed copies inside
the test library when running on Windows, so the real path was never covered on
any platform.

`lookup_win32.c` now walks every module loaded into the process, giving the
default library the same meaning it has on POSIX. `EnumProcessModules` is
resolved dynamically as `K32EnumProcessModules` from `kernel32.dll`, so this
adds no link-time dependency.

**2. A DLL's own dependencies were not found next to it.**

Sulong always loaded native libraries with no flags, so `LoadLibraryEx` got
`dwFlags == 0`. Windows then resolves the library's dependencies against the
standard search order, which does *not* include the directory of the DLL being
loaded. On POSIX the loader handles this via RPATH/RUNPATH, commonly `$ORIGIN`.
A library such as Intel MKL, whose entry-point DLL pulls in siblings from its
own directory, therefore fails to load on Windows while loading fine on Linux.

`NFIContextExtension` now passes
`LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR|LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` when loading
by absolute path on Windows. The NFI already accepted these flag names — nothing
ever passed them. Deliberately **not** `LOAD_WITH_ALTERED_SEARCH_PATH`: that is
the legacy scheme and it puts the process working directory on the dependency
search path, which is a DLL-planting vector.

**Also fixed along the way**

- `throwError` in `lookup_win32.c` could pass an uninitialised pointer to
  `ThrowNew` and then `LocalFree` it, because `FormatMessageA` leaves its output
  pointer untouched when it fails and has no message for every error code.
- `NFIContextExtension` carried its own `getNativeLibrarySuffix` /
  `getNativeLibrarySuffixVersioned` returning `"so"` on Windows, hiding the
  correct Windows-aware versions in the `NativeContextExtension` parent. They
  were unreachable — Java statics bind lexically — but a trap for the next
  caller. Removed; the parent's platform checks are now computed once into
  constants with an `isWindows()` accessor.

## Related Issues

None filed — this came out of trying to bind a native library from Sulong on
Windows the way it works on Linux.

## Testing

- `NFITest.lookupAndBindDefault` no longer takes the Windows shortcut, so every
  platform exercises the real default library. The `reexport_strdup` /
  `reexport_free` shims existed only for that workaround and are removed.
- **`StringNFITest` was the caller that workaround existed for.** It now names
  the symbol per platform: the C runtime on Windows spells it `_strdup`, since
  the POSIX name is deprecated there and is not exported by `ucrtbase.dll`.
  `free` is spelled the same on both. That `free` resolves through the default
  library on Windows is itself evidence the fix works.
- New **`DefaultLibraryNFITest`** resolves and calls a libc function through the
  default library, and checks that an undefined symbol fails rather than
  resolving to garbage. Nothing anywhere exercised the default library directly
  before.
- **`LoadNFILibraryTest`**: the `RTLD_*` cases now skip on Windows. They were
  vacuous there — the Windows flag parser ignores names it does not know, so
  each request collapsed to a plain `load` that passed while asserting nothing
  about the flag. Added the Windows `LOAD_LIBRARY_SEARCH_*` counterparts,
  including the combination Sulong now uses.
- New GitHub Actions workflow runs these tests on Windows **and** Linux from one
  matrix, so the two cannot silently drift apart again.

Results:

| | Windows | Linux |
|---|---|---|
| `mx unittest com.oracle.truffle.nfi.test` | **OK (560 tests)** | **OK (570 tests)** |
| `DefaultLibraryNFITest` | passed | passed |
| `loadSearchDllLoadDir*` | passed | skipped, Windows-only |

Also verified locally on Windows 11 / MSVC 14.51 / labsjdk
25.0.4.1.1-jvmci-25.4-b23, where Sulong's `nativemode` and `runtime` projects
build clean.

## Documentation

No user-facing documentation changes. The reasoning that is easy to get wrong
later — why the modern `LOAD_LIBRARY_SEARCH_*` flags rather than
`LOAD_WITH_ALTERED_SEARCH_PATH`, and why the default-library lookup must walk
loaded modules — is recorded as comments at both sites.

## Not addressed here

`--llvm.libraryPath` still does not reach the OS loader on Windows. A DLL's
dependencies are now found next to it, but not in other `libraryPath` entries;
`LD_LIBRARY_PATH` covers that on Linux. Closing it needs `AddDllDirectory` and
therefore new NFI native API, so it is left for a separate change.

## Contributor Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Y7QHXmVpEqBM5eJsQRj5cN

### DIFF
--- a/.github/workflows/nfi-native-library-tests.yml
+++ b/.github/workflows/nfi-native-library-tests.yml
@@ -1,0 +1,216 @@
+#
+# Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+#
+# The Universal Permissive License (UPL), Version 1.0
+#
+# Subject to the condition set forth below, permission is hereby granted to any
+# person obtaining a copy of this software, associated documentation and/or
+# data (collectively the "Software"), free of charge and under any and all
+# copyright rights in the Software, and any and all patent rights owned or
+# freely licensable by each licensor hereunder covering either (i) the
+# unmodified Software as contributed to or provided by such licensor, or (ii)
+# the Larger Works (as defined below), to deal in both
+#
+# (a) the Software, and
+#
+# (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+# one is included with the Software each a "Larger Work" to which the Software
+# is contributed by such licensors),
+#
+# without restriction, including without limitation the rights to copy, create
+# derivative works of, display, perform, and distribute the Software and make,
+# use, sell, offer for sale, import, export, have made, and have sold the
+# Software and the Larger Work(s), and to sublicense the foregoing rights on
+# either these or other terms.
+#
+# This license is subject to the following condition:
+#
+# The above copyright notice and either this complete permission notice or at a
+# minimum a reference to the UPL must be included in all copies or substantial
+# portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+# ---------------------------------------------------------------------------
+# Truffle NFI native-library tests, run on Windows *and* Linux.
+#
+# The point of running both is that the Windows behaviour is supposed to match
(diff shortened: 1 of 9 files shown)


masariello:
I am covered by the Oracle Contributor Agreement — I signed it before opening this pull request.

The `oca/oracle` check is still red here. Leaving this comment as the [GraalVM contributor guide](https://www.graalvm.org/community/contributors/) directs:

> If there is a delay in showing you have signed OCA, or the bot leaves a message on your pull request that you have not signed the OCA, leave a comment on the pull request stating that you are now covered by the OCA.

In case the check database needs a manual update, the two identities the bot flagged are:

- GitHub username: `masariello`
- Commit author email: `masariello@gmail.com`

Happy to provide anything else that helps resolve it.

PR #14467, opened 2026-09-17
Truffle DSL: fix varargs warning

The-Alchemist:
## Problem

```
warning: non-varargs call of varargs method with inexact argument type for last parameter;
    cached.add(Arrays.<Object>asList(this.fused_assumptions_));
                                          ^
  cast to java.lang.Object for a varargs call
  cast to java.lang.Object[] for a non-varargs call and to suppress this warning
```

## Cause

The introspection data generated by `FlatNodeGenFactory` passes all cached values of a specialization to a single `Arrays.<Object>asList(...)` call. When a specialization has exactly one cache and that cache is a reference array, javac resolves the call as a *non-varargs* invocation, because the array is assignable to the `Object[]` parameter.

The array contents are spread across the cached data list, contradicting the one-entry-per-cache contract of `Introspection.SpecializationInfo.getCachedData(int)`. The same cache in a specialization with two or more caches is wrapped as a single element, so the behavior was inconsistent. The returned list also aliased the live cache array.


chumer:
Thanks for the fix. Currently in internal review. Give it 1-2 days until it gets merged. Thanks!

The-Alchemist:
@chumer np!

